PR #17, opened 2022-01-03
Logger middleware

### LINKED ISSUE #11: "Logger" builtin middleware
We want to have "Logger" middleware as a *builtin* middleware. We have to write middleware with no dependency on any other npm modules. To know how to write middleware, see `Powered By` middleware: 

```js
const poweredBy = async (c, next) => {
 await next()
 await c.res.headers.append('X-Powered-By', 'Hono')
}
```

* <https://github.com/yusukebe/hono/blob/master/src/middleware/poweredBy.js>

After we wrote middleware, add the method to `Middleware` object:

```js
Middleware.poweredBy = poweredBy
```

* <https://github.com/yusukebe/hono/blob/master/src/middleware.js>

And then, we can use middleware like this:

```js
const { Hono, Middleware } = require('hono')
...
app.use('*', Middleware.poweredBy)
```

metrue:
This PR is to close #11 by  implementing the logger middleware without any dependencies (heavily inspired by https://github.com/koajs/logger).


yusukebe:
@metrue 
Thank you for PR. This is great.

`logger` return *async function* such as below is good idea for passing the args. I think we have to implement middleware in a similar way. 

```js
const logger = (fn = console.log) => {
   return async (c, next) => {
```

PR #38, opened 2022-01-10
feat(body-parse): a body parse middleware

metrue:
Hi @yusukebe , not sure it's good idea or not if *Hono* provides a built-in `body-parse` middleware to handle the `json`, `xml` and `form` payload for `POST` method.  
This is just a simple PR to explain what it maybe like,  which is not ready for review, just for a better discussion.

### DIFF
--- a/src/hono.ts
+++ b/src/hono.ts
@@ -13,6 +13,7 @@ declare global {
   interface Request {
     params: (key: string) => string
     query: (key: string) => string | null
+    parsedBody: any
   }
 }
 
--- a/src/middleware/body-parse/body-parse.test.ts
+++ b/src/middleware/body-parse/body-parse.test.ts
@@ -1,0 +1,68 @@
+import makeServiceWorkerEnv from 'service-worker-mock'
+import { Hono, Middleware } from '../../hono'
+// eslint-disable-next-line node/no-extraneous-require
+const FormData = require('form-data')
+
+declare let global: any
+Object.assign(global, makeServiceWorkerEnv())
+
+describe('Parse Body Middleware', () => {
+  const app = new Hono()
+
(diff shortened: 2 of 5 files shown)


src/hono.ts
```diff
@@ -13,6 +13,7 @@ declare global {
   interface Request {
     params: (key: string) => string
     query: (key: string) => string | null
+    body: any
```
metrue, line 16: The tricky part is the `body` of original fetch event is `read-only`,  so I'm thinking maybe we can extend this interface to have `body` to hold the parsed payload?

yusukebe:
@metrue 

I’ve been wondering about it, but now I think `body-parse` middleware is good.

The points:

* `body` property of Request is readable by default. That might be better to leave it be, but that's not the case.
* Request instance has `text()`, `json()`, and `formData()` methods, so `body-parse` is not absolutely necessary. But, it's handy to handle request bodies of any format transparently.

So, please write `body-parse` middleware as builtin. Thank you for your proposal.

src/middleware/body-parse/body-parse.test.ts
```diff
@@ -0,0 +1,68 @@
...
+      headers: new Headers({ 'Content-Type': 'application/text' }),
+    })
+    const res = await app.dispatch(req)
+    expect(res).not.toBeNull()
+    expect(res.status).toBe(200)
+    expect(req.parsedBody).toEqual(payload)
+    expect(await res.text()).toEqual(payload)
+  })
+
+  // NOTE the service-worker-mock does not support req.formData()
```
yusukebe, line 51: I see.  I will think about another way.

yusukebe (approved):

yusukebe:
Thank you! I'll merge it.

PR #44, opened 2022-01-12
fix: correct typo for change directory

justinnoel:


yusukebe:
@justinnoel Thank you!

PR #59, opened 2022-01-21
refactor(utils): better management for utils modules to avoid kitchen sink issue

metrue:
Refactor `utils` with better naming.


yusukebe (approved):
Good suggestion. Thank you!

PR #62, opened 2022-01-24
feat: Add new shortcuts for request/response

yusukebe:
Add new shortcuts for Requests:

```js
c.req.header('Header-Name')
c.req.param('q') // instead of c.req.params
```

Add new shortcuts for Response:

```js
c.header('Header-Name', 'Header-Value')
c.status(201)
c.statusText('This status text')
return c.text('This is Body text')
/*
new Response('This is Body text', {
  status: 201,
  statusText: 'This status text',
  headers: { 'Header-Name': 'Header-Value', 'Content-Type': 'plain/text' },
})
*/
```

### DIFF
--- a/src/middleware/default.ts
+++ b/src/middleware/default.ts
@@ -6,7 +6,6 @@ export const defaultMiddleware = async (c: Context, next: Function) => {
     const url = new URL(c.req.url)
     return url.searchParams.get(key)
   }
-  c.req.params = c.req.param // alias
   c.req.header = (name: string): string => {
     return c.req.headers.get(name)
   }
--- a/src/middleware/default.ts
+++ b/src/middleware/default.ts
@@ -6,6 +6,10 @@ export const defaultMiddleware = async (c: Context, next: Function) => {
     const url = new URL(c.req.url)
     return url.searchParams.get(key)
   }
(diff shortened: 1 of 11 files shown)


src/middleware/default.ts
```diff
@@ -6,6 +6,10 @@ export const defaultMiddleware = async (c: Context, next: Function) => {
     const url = new URL(c.req.url)
     return url.searchParams.get(key)
   }
+  c.req.params = c.req.param // alias
```
metrue, line 9: Why we just make `ctx.req.params` at first place to avoid this alias?

yusukebe: I see, that is better to make the alias after this: https://github.com/yusukebe/hono/pull/62/files#diff-fef0c99deaa99c63dde9a9af2451ae5281a68e92c338218d644a08b358878245R131

metrue: Why we have both `params` and `param` in a `Request`, is possible to make it only `params` or `param`?

yusukebe: Yes, we make it `param` only because `params` is just for backward compatibility and that we have both may cause confusion.

PR #84, opened 2022-02-02
fix: use `atob` for basic auth

yusukebe:


src/middleware/basic-auth/basic-auth.ts
```diff
@@ -31,26 +31,16 @@ const auth = (req: Request) => {
   return { username: userPass[1], password: userPass[2] }
 }
 
-let b: BufferConstructor
-
 function decodeBase64(str: string) {
-  return b.from(str, 'base64').toString()
+  if (atob && btoa) {
+    return atob(str)
```
danielweck, line 36: hello, btoa (and its counterpart atob) will crash with non-ASCII Unicode characaters that don't fit in one byte of binary data (aka binary strings with exactly 8 bit code units).
See: https://developer.mozilla.org/en-US/docs/Glossary/Base64#solution_1_%E2%80%93_escaping_the_string_before_encoding_it

danielweck: unit test: https://github.com/yusukebe/hono/blob/7dd486ad7b4557d263aa5977831f76ea587b3512/src/middleware/basic-auth/basic-auth.test.ts#L14-L15

yusukebe: Thank you!

Fixed this issue on #85

PR #85, opened 2022-02-02
fix: Unicode problem about atob/btoa

yusukebe:
https://github.com/yusukebe/hono/pull/84#pullrequestreview-870144435

Thank you @danielweck

### DIFF
--- a/src/middleware/basic-auth/basic-auth.test.ts
+++ b/src/middleware/basic-auth/basic-auth.test.ts
@@ -11,20 +11,32 @@ describe('Basic Auth by Middleware', () => {
   })
 
   const app = new Hono()
+
   const username = 'hono-user-a'
   const password = 'hono-password-a'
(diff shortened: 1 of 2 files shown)


danielweck:
Hello 😃 

Have you considered using a well-tested micro-lib like `@cfworker/base64url`? See source code:

https://github.com/cfworker/cfworker/tree/main/packages/base64url/src

yusukebe:
HI!

I didn't know `@cfworker/base64url` , thank you to tell me 😄 

Our policy is that Hono should not depend other libraries. So, we should not use `@cfworker/base64url` directly. 

`@cfworker/base64url` is implemented with `decodeURIComponent/encodeURIComponent` . This PR is using `TextDecoder`. I think it may not matter which one. If we should write with another way, I will refer to these libralies 👍 

Thank you!

PR #102, opened 2022-02-14
Added type to c.req.param key.

usualoma:
Hi,

Thank you for publishing such a good product.

This PR adds the ability to complete the type of `c.req.param`.

I think this is a nice feature, but it also makes type definitions more complex, so I think it is reasonable to decide not to adopt it.

## Pros

We can get support from the editor for the type of c.req.param. Also, if there is no possibility of a value being returned from c.req.param, the parameter type will be set to `never` and cannot be specified.

https://user-images.githubusercontent.com/30598/153787066-92e7f97b-32e9-4387-add5-166b2cf302c9.mov

image

image

image

image

### Cons

The simplicity of the type display will be lost.
The following image is before the PR was applied. For those who are not used to reading conditional type specifications, this may be easier to read.

image

### DIFF
--- a/src/hono.ts
+++ b/src/hono.ts
@@ -116,6 +116,8 @@ export class Hono {
   connect
   */
 
+  all<Path extends string>(arg: Path, ...args: Handler<ParamKeys<Path>>[]) : Hono;
+  all(arg: Handler<never>, ...args: Handler<never>[]): Hono;
   all(arg: string | Handler, ...args: Handler[]): Hono {
     return this.addRoute('all', arg, ...args)
   }
--- a/src/hono.ts
+++ b/src/hono.ts
@@ -8,16 +8,32 @@ import type { Env } from './context'
 const METHOD_NAME_OF_ALL = 'ALL'
 
 declare global {
-  interface Request {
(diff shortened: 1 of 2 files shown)


src/hono.ts
```diff
@@ -47,24 +63,38 @@ export class Hono {
   }
 
   /* HTTP METHODS */
```
yusukebe, line 65: Also, please write the features on [`all`](https://github.com/yusukebe/hono/pull/102/files#diff-fef0c99deaa99c63dde9a9af2451ae5281a68e92c338218d644a08b358878245R119)

SS

yusukebe (changes requested):

yusukebe:
Hi @usualoma !

Thank you for your PR. This is great feature. We will accept this PR even though the code may be more complex.
I commented one thing for your code. Please check it.

yusukebe (approved):

src/hono.ts
usualoma: Oops, also added to "all” by 2d2726a

usualoma:
Thanks! 😄

PR #109, opened 2022-02-20
Introduce RegExpRouter

usualoma:
Hi,

I tried to improve the performance of the router using a different strategy.

### Strategy

All the routes into one RegExp object.

* /help
* /:user_id/followees
* /:user_id/followers
* /:user_id/posts
* /:user_id/posts/:post_id
* /:user_id/posts/:post_id/likes

into `^/(?:help$()|([^/]+)/(?:followe(?:es$()|rs$())|posts(?:/([^/]+)(?:/likes$()|$())|$())))`

It will be a large regular expression, but since common prefixes can be grouped together by trie, I think it will grow slowly in the general case.
Also, the return value of `match` will be larger in proportion to the number of routes, but it will be an array where most of the elements are `undefined`, so I think it is still an acceptable size in the general case.

#### Existing modules adopting the same strategy

The following Perl modules use large regular expression objects for high performance.

* [Router::Boom](https://github.com/tokuhirom/Router-Boom)
* [Router::Assemble](https://github.com/sixapart/Router-Assemble)

### Benchmark

```
hono x 410,442 ops/sec ±5.62% (72 runs sampled)
hono with RegExpRouter x 460,937 ops/sec ±6.52% (65 runs sampled)
itty-router x 73,523 ops/sec ±3.63% (81 runs sampled)
sunder x 131,994 ops/sec ±0.80% (84 runs sampled)
worktop x 81,120 ops/sec ±4.16% (82 runs sampled)
Fastest is hono with RegExpRouter
```

### Compatibility

I ran a test by changing it to use RegExpRouter by default and got green, so I believe it is compatible enough.
https://github.com/usualoma/hono/compare/regexp-router..regexp-router-as-default

#### Known differences

* The current router can include capture in `{}` notation (e.g. `/:action{(update|delete)}`), but RegExpRouter can't.  We need to use `(?:...)` instead of `(...)`.
* The current router can `add` route after call `match`, but RegExpRouter can't.
* The current router can not handle multiple capture in same position (e.g. `/path/to/:decimal{[0-9]+}`, `/path/to/:hexadecimal{0x[0-9A-F]+}`), but RegExpRouter can.

### DIFF
--- a/benchmarks/handle-event/index.js
+++ b/benchmarks/handle-event/index.js
@@ -1,6 +1,7 @@
 import Benchmark from 'benchmark'
 import { makeEdgeEnv } from 'edge-mock'
 import { Hono } from '../../dist/index'
+import { RegExpRouter } from '../../dist/router/reg-exp-router'
 import itty from 'itty-router'
 const { Router: IttyRouter } = itty
 import { Router as SunderRouter, Sunder } from 'sunder'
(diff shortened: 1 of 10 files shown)


yusukebe:
Hi @usualoma 

I tried the benchmark on my PC. Results are...

```plain
hono x 723,504 ops/sec ±6.76% (63 runs sampled)
hono with RegExpRouter x 934,401 ops/sec ±5.49% (68 runs sampled)
itty-router x 160,676 ops/sec ±3.23% (88 runs sampled)
sunder x 312,128 ops/sec ±4.55% (85 runs sampled)
worktop x 209,345 ops/sec ±4.52% (78 runs sampled)
Fastest is hono with RegExpRouter
```

SUGEEEEEEE!

*RegExpRouter* is pretty fast compared to other *node* routers not only for Cloudflare Workers such as `find-my-way`.  Ref: <https://github.com/delvedor/router-benchmark>.

This is great. I'll merge this PR. After a while, I may make RegExpRoute as a default router if there is no problem.

Thank you!!

usualoma:
@yusukebe Thank you!

> After a while, I may make RegExpRoute as a default router if there is no problem.

👍 

Feel free to make changes ( or fix bugs ) to the RegExpRouter as needed.

PR #113, opened 2022-02-22
Make ParamMap Array<[key, value]>.

usualoma:
In this PR, I will add a very small optimization to RegExpRouter.

It turns out that iterating through the results of Object.keys is little bit expensive, so I change ParamMap to a list of keys and values (a.k.a Association List).

https://gist.github.com/usualoma/6d22fdb1b64e7ae8f4d6fe2d204b9208

I don't think this optimization will make any visible difference to the router benchmarks.

### FYI

I also tried optimize by using lastIndexOf() instead of indexOf(), but failed. Because lastIndexOf() is slow.
https://blog.taaas.jp/tips/v8lastindexof/


yusukebe:
Thank you!

PR #135, opened 2022-03-11
Optimize RegExpRouter

usualoma:
Hi,

I tried to optimize RegExpRouter.

### Summary

I think most middleware is applied to paths like '\*' or '/path/to/\*' that take no parameters.
In that case, parameter handling is not needed, so `match` can be optimized as follows.
https://github.com/yusukebe/hono/compare/master...usualoma:optimize-reg-exp-router?expand=1#diff-06f981941f95eecd6e249629fd9d5d4badaf8dfbe96ade6219e67691359df679R35-R38

And by optimizing the '\*' commonly used in middleware separately as described above, we can make it simpler for normal routing as well.
https://github.com/yusukebe/hono/compare/master...usualoma:optimize-reg-exp-router?expand=1#diff-06f981941f95eecd6e249629fd9d5d4badaf8dfbe96ade6219e67691359df679R43

### Benchmark

Since TrieRouter and RegExpRouter are already fast enough, the ranking may not be stable due to minute environmental influences, but the following results were obtained in my environment.

```
hono x 413,307 ops/sec ±3.71% (78 runs sampled)
hono with RegExpRouter x 445,681 ops/sec ±6.82% (66 runs sampled)
hono with RegExpRouter in this PR x 479,019 ops/sec ±6.31% (66 runs sampled)
itty-router x 67,901 ops/sec ±3.32% (88 runs sampled)
sunder x 115,183 ops/sec ±2.73% (80 runs sampled)
worktop x 72,526 ops/sec ±4.84% (80 runs sampled)
Fastest is hono with RegExpRouter in this PR
```

However, this is more than just a speedup on this benchmark.
* Speeds up middleware routing.
* Hopefully, even the following cases may be faster than TrieRouter 😛
    * https://twitter.com/yusukebe/status/1502069834712219650

### DIFF
--- a/src/router/reg-exp-router/node.ts
+++ b/src/router/reg-exp-router/node.ts
@@ -42,9 +42,7 @@ function compareKey(a: string, b: string): number {
 export class Node {
   index?: number
   varIndex?: number
-  children: {
-    [key: string]: Node
-  } = {}
+  children: Record<string, Node> = {}
 
   insert(tokens: readonly string[], index: number, paramMap: ParamMap, context: Context): void {
     if (tokens.length === 0) {
--- a/src/router/reg-exp-router/router.test.ts
+++ b/src/router/reg-exp-router/router.test.ts
@@ -1,4 +1,5 @@
 import { RegExpRouter } from './router'
+import { METHOD_NAME_OF_ALL } from '../../router'
 
 describe('Basic Usage', () => {
   const router = new RegExpRouter<string>()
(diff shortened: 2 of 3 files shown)


yusukebe:
Hi @usualoma 

> I think most middleware is applied to paths like '*' or '/path/to/*' that take no parameters.

I'm thinking about same things. This PR looks good to me!

The score of `RegExpRouter` has been about the same as a score of `TrieRouter` in this case.

SS

I'll merge it!

PR #169, opened 2022-04-24
feat(jwt): jwt middleware

### LINKED ISSUE #168: JWT middleware

metrue:
This MR is to add JWT middleware to hono.

Close #168


yusukebe:
Hi @metrue, thank you for PR.

I have two suggestions:

1. How about using `encodeBase64` / `decodeBase64` in [utils/crypto.ts](https://github.com/yusukebe/hono/blob/master/src/utils/crypto.ts) for encoding and decoding Base64.
2. How about making JWT core functions `sign` and `verify`... as utility methods in the `utils` directory. For example, make `utils/jwt.ts` or `utils/jwt/index.ts`.... If we do it this way, we can reuse it in our handlers or middleware.

metrue:
Hi @yusukebe 

> How about using encodeBase64 / decodeBase64 in [utils/crypto.ts](https://github.com/yusukebe/hono/blob/master/src/utils/crypto.ts) for encoding and decoding Base64.
We may not be able to use utils/crypto.ts directly, since JWT encodes the header and payload with base64url which is a bit difference with base64. 
But maybe we can implement it into utils/crypto.ts

> How about making JWT core functions sign and verify... as utility methods in the utils directory. For example, make utils/jwt.ts or utils/jwt/index.ts.... If we do it this way, we can reuse it in our handlers or middleware.
I think we could do that, although I haven't thought of a use case now.

yusukebe:
@metrue 

I see. 

First, as much as possible, I want you to use base64 in `utils/crypto.ts` as inside logic for base64url. Because base64 in `utils/crypto.ts` is simply not only made for Cloudflare Workers but also for such as Compute@Edge. Compute@Edge does not have `atob` / `btoa`, but the base64 will work without `atob` and `btoa` if we polyfill `buffer` ( although not fully tested).

Second, 

> I think we could do that, although I haven't thought of a use case now.

Methods in `utils/*.ts` are [exported in packages.json](https://github.com/yusukebe/hono/blob/79e4794126d8b7da3da617ea563cc268cfdf2dab/package.json#L33). I haven't said it publicly, but we can use these utility functions in our application. If we want to do encode/decode base64, we can use the `encodeBase64` / `decodeBase64` in `utils/cyrpt.ts`.

```js
import { encodeBase64, decodeBase64 } from 'hono/utils/crypto'
```

Or we can use methods for Cloudflare Workers.

```js
import { getContentFromKVAsset } from 'hono/utils/cloudflare'
```

So, if methods for JWT are placed under the `utils` and exported, these are useful for us. For example, we issue a JWT token in a login handler ( This code is just POC ).

```js
app.post('/login', (c) => {
  const username = c.req.parsedBody.username
  const password = c.req.parsedBody.password
  if (isOK(username, password)) {
    const token = JWT.sign({ username: username }, secret)
    res.json({
      token: token,
    })
  }else {
    //...
  }
})
```

These are my thoughts. What do you think?

metrue:
Good points, @yusukebe .

Hono's not only designed for Cloudflare Worker, but also for Fastly Compute@Edge, it's important to make the codes compatible for both of them as much as possible. 
And yes there're cases we may use JWT functionality outside of middleware.

I will refactor the codes.

yusukebe:
Great! Thank you!

julianpoma:
The fact that Hono provides a sign method should be documented somewhere! Saved me from installing a library like `jose` for this purpose!

yusukebe:
Hi @julianpoma 

Are you referring to the `sign` method in `utils/jwt/jwt.ts`? We don't officially provide the methods in `utils/*` for public use. They are meant for internal use only, primarily because there's the possibility they might undergo breaking changes without announcement.

But, most APIs within the utils are stable, so it might not be a bad idea to consider exporting methods from `utils/jwt/jwt.ts`.

julianpoma:
Exactly @yusukebe. I am doing an API server and I need to issue JWTokens, not just verify them. 

Understood that this methods could change without notice, I can live with that :) 

I think it's worth considering exposing a small set of helpers to handle JWTs - `sign`, `verify` and `decode`. Just like Hono includes [methods to manage cookies.](https://hono.dev/helpers/cookie)

yusukebe:
@julianpoma 

> I think it's worth considering exposing a small set of helpers to handle JWTs - sign, verify and decode. Just like Hono includes [methods to manage cookies.](https://hono.dev/helpers/cookie)

This is a awesome idea! Could you create the issue for the helpers as a feature request? Thanks!

PR #180, opened 2022-04-27
fix(jwt): the alg option should be same as the one in options

metrue:
This PR is to,
* fix the algorithm not used in Jwt middleware when it's defined.
* add more unit tests.


yusukebe:
OK. Thank you!

PR #188, opened 2022-04-30
feat(middleware): enable middleware on http method

metrue:


yusukebe:
Hi @metrue 

I want to know your use cases for this PR.

In some cases, it's good that we can write `app.use('GET', '*', ...)`. But I think this feature is not necessary in most cases. For example, the `CORS` middleware must handle some methods like `OPTIONS` for preflight. Not only one method. So, I should not specify the method by`app.use`. These are just my thought.

metrue:
Hi @yusukebe 

I think there're middlewares only work for a specific `HTTP` method on same resource (`PATH`),  take a `/posts` of blog application API as an example,  

```
POST    /posts    ---> create a blog post
GET      /posts    --->  fetch the blog posts
```
We may enable a `authorization` middleware for `POST /posts`, but open access for `GET /posts`, or we may enable a `bodyParse` middleware for `POST /posts` to construct the post payload, but disable it for `GET /posts`.

metrue:
I marked this PR as a draft is because it's not elegant not for the codes but for the Hono interface for user,  and that's why I made [another proposal](#189)

yusukebe:
OK! Let's discuss it in the #189

PR #190, opened 2022-05-01
fix(env): Property 'xxx' does not exist on type 'Env'.

metrue:
make the Env interface [string:string], to fix the issue `Property 'xxx' does not exist on type 'Env'.` when we're trying to get the env 'xxx' from `ctx.env` with `ctx.env.xxx`


yusukebe:
Hi @metrue 

I've written the same thing. This issue is fixed in #191 .

Issue #200, opened 2022-05-05
How to structure large apis using hono?


OultimoCoder:
Traditionally using express I would have this file structure:

```
── routes
    ├── index.js
    ├── users.route.js
── index.js  

```

And then in `users.route.js` I would define all my routes relating to users e.g.

```
const router = express.Router();

router
  .route('/')
  .post(auth('manageUsers'), validate(userValidation.createUser), userController.createUser)
  .get(auth('getUsers'), validate(userValidation.getUsers), userController.getUsers);

router
  .route('/:userId')
  .get(auth('getUsers'), validate(userValidation.getUser), userController.getUser)
  .patch(auth('manageUsers'), validate(userValidation.updateUser), userController.updateUser)
  .delete(auth('manageUsers'), validate(userValidation.deleteUser), userController.deleteUser);

module.exports = router;
```

In `routes/index.js` I would apply the base urls to all my defined routes and then in the top level `index.js` I would import and enable them:

```
// v1 api routes
app.use('/v1', routes);
```

What is the recommended way of doing this in Hono?

yusukebe:
Hi @OultimoCoder 

Currently, we can write *routing* using `route` like this:

```ts
export const app = new Hono()

const root = app.route('/v1')
root.get('/', (c) => c.text('getUsers'))
root.post('/', (c) => c.text('createUser'))

```

And, This syntax may be available in the next release.

```ts
root.post('/', basicAuth(), cors(), (c) => c.text('getUsers')) // <--- this syntax is coming soon!
```

I think this is not such a smart design for large APIs, so I want to add another way for routing, such as below:

```ts
import { Hono, Route } from 'hono'

const app = new Hono()
const route = new Route()

route.get('/entry', handler, handler).post(handler, handler)

app.use('/v1', route)
app.fire()
```

This is still just the idea.

Conclusion, please use `app.route` method. Thank you.

yusukebe:
@OultimoCoder 

This feature is already implemented #203 . 
We are planning to emit this feature in the next release v1.2.0. Maybe its' soon!

yusukebe:
@OultimoCoder 

We released v1.2.0. See: <https://github.com/honojs/hono/releases/tag/v1.2.0>.

yusukebe:
Hi @OultimoCoder 

`Route` is deprecated. Please use the `Hono` instance!

https://github.com/honojs/hono#grouping

PR #203, opened 2022-05-06
feat: `app.route(path, Route)`

### LINKED ISSUE #201: Proposal: Stacked routes as middleware
I've noticed that we don't have a smart API for the application with a large structure. It's related to #200 
So, I propose yet another way for routing. That is stacking routes and using these as middleware.

```ts
import { Hono, Route } from 'hono'

const app = new Hono()
const route = new Route()

route.get('/entry', handler, handler).post(handler, handler)

app.use('/v1', route)
app.fire()
```

The routers `TrieRouter` and `RegExpRouter` do not have routes like a *stack* but *tree*. So, we can't use the ways directly, and we have to make the object to stack routes. `Route` object is for stacking in the above code.

**It's just the idea**. If you have comments, please tell me.

yusukebe:
Using `Route` object with `app.route` method, we can make the large structured applications more easily.

For example:

```ts
// route.ts
import { Route } from 'hono'
import { cors } from 'hono/cors'
import { poweredBy } from 'hono/powered-by'

export const route = new Route()

route
  .get('/', (c) => c.text('List posts'))
  .post(cors(), poweredBy(), (c) => {
    return c.text('Create post')
  })
  .get('/:id{[0-9]+}', (c) => {
    const id = c.req.param('id')
    return c.text(`Your id is ${id}`)
  })
```

```ts
// index.ts
import { route } from './route'

import { Hono } from 'hono'
const app = new Hono()

app.route('/v1', route)

app.fire()
```

Related #200 
Close #201

### DIFF
--- a/src/hono.test.ts
+++ b/src/hono.test.ts
@@ -1,7 +1,7 @@
 import { poweredBy } from './middleware/powered-by'
 import type { Context } from '@/context'
 import type { Next } from '@/hono'
-import { Hono } from '@/hono'
+import { Hono, Route } from '@/hono'
 
 describe('GET Request', () => {
   const app = new Hono()
@@ -265,7 +265,7 @@ describe('Middleware', () => {
       return c.text(`${message}`)
     })
 
-    app.get('/error', (c) => {
+    app.get('/error', () => {
       throw new Error('Error!')
     })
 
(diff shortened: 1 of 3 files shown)


metrue:
Hi @yusukebe 

I know the idea of `router.route` is trying to make routes management easier for large structure, but I think I may need a little more time to think about it, mainly about two things,
* the interface of hono (the app object)
* the interface of router

but please make your call to merge it as you want.

yusukebe:
@metrue OK! Please think about it.

yusukebe:
Hi @metrue 

Thank you for telling your thinking. Something is the same as my thought, something is not the same.

This lines:

> hono is like a connector
> router is just another middleware

I think half right but half wrong. In the first place, I've made Hono as a **router**. The points:

1. Hono is just a **router** actually.
2. But, Hono is thin web framework that have `Context` to access Request/Response easily.
3. *Router* what you call such like `TrieRouter` and `RegExpRouter` is just like a **engine** using inside of Hono
4. To register the handler/middleware, use **app.HTTP_METHOD** like `app.get` `app.post` `app.delete`...

So, I can say that "*Router*" may not be a middleware. For another way, I made `Route` and `app.route()` in this PR to have a bunch of APIs.

```ts
const route = new Route()
route.get('/', (c) => c.text('foo')) // <--- stack the handler

const app = new Hono()
app.route('/users', route) // <--- add stacked handlers to app
```

Isn't this good?

metrue:
Hey @yusukebe 

Thank you so much for sharing, it may take me some time to fully understand, but that's fine, let's keep delivering and I think everything will be clear.

PR #220, opened 2022-05-11
feat: multiple matching routes for RegExpRouter

usualoma:
#### Middleware pre-bundling

```
router.add('ALL', '*', 'middleware a')
router.add('GET', '*', 'middleware b')
router.add('GET', '/entry/*', 'middleware c')
router.add('GET', '/entry', 'get entries')
router.add('GET', '/entry/:id', 'get entry')
```

Pre-bundle these routings as follows ( for GET request)

| path | bundled targets |
| ------------- | ------------- |
| `*`  | `['middleware a', 'middleware b']`  |
| `/entry/*`  | `['middleware a', 'middleware b', 'middleware c']`  |
| `/entry`  | `['middleware a', 'middleware b', 'middleware c', 'get entries']`  |
| `/entry/:id`  | `['middleware a', 'middleware b', 'middleware c', 'get entry']`  |

This way, all middleware and handlers can be executed with a single regular expression matching.

#### Primary matchers and secondary matchers

Also in this case, the final match is likely to be the one that does not end up with a `*` and does not have routing lower than itself.

* Target for first match (in primaryMatchers)
    * `/entry`
    * `/entry/:id`
* Fall back on failure (in secondaryMatchers)
    * `*`
    * `/entry/*`

This allows for faster discovery of more likely routes.

#### Ambiguous routes

The following cases cannot be completely determined in advance, so when this case is detected, the matching routine is changed to a multiple matching. When switched to this one, the speed is slower. (Still faster than a typical router, but slower than a TrieRouter.)

However, I do not believe that most real-world routing will be determined to be "ambiguous".

```
router.add('GET', '/:user/entries', 'get user entries')
router.add('GET', '/entry/:name', 'get entry')
```

### TODO

The test passes (although some test data needs to be modified) even after replacing the default with RegExpRouter. Thus, I think it works fine in the general case. However, RegExpRouter is quite complex now and we will need to add more tests in the future.

### Benchmark

The following results were obtained in my environment.

```
hono x 359,641 ops/sec ±4.01% (75 runs sampled)
hono with RegExpRouter x 691,034 ops/sec ±5.07% (65 runs sampled)
itty-router x 73,662 ops/sec ±3.07% (81 runs sampled)
sunder x 133,993 ops/sec ±1.66% (90 runs sampled)
worktop x 80,886 ops/sec ±2.40% (81 runs sampled)
Fastest is hono with RegExpRouter
```


usualoma:
Oops, I will revise the content `${''}` later to be more appropriate.

https://github.com/honojs/hono/pull/220/files#diff-06f981941f95eecd6e249629fd9d5d4badaf8dfbe96ade6219e67691359df679R330

usualoma:
`${''}` issue has been fixed by 8c939c6

usualoma:
The work I intend to do on this PR is complete and I have no plans to add any more commits.

yusukebe:
SS

SUGEEEEEEEE. @usualoma 

Thank you for your hard work! I'll merge it!

PR #223, opened 2022-05-12
feat(reg-exp-router): error handling for duplicate param name

usualoma:
The actual check feature was added by bd42c418100a28bba2cd584ad43424e48325baf5;  f969924a144705663d6be42294e77af40135b01a and 2baa25302d63250af7155212471d98a844eb86d7 are optional.

I made one PR because it has to do with the timing of the check. If we cannot agree on the implementation of an optional part, I will reconsider.

### Check for duplicate parameter names bd42c418100a28bba2cd5

I think it's OK because it's implemented honestly.

### Add router.prepare and invoke #router.prepare() in hono.fire f969924a144705663d6be42294e77af40135b01a and 2baa25302d63250af7155212471d98a844eb86d7

The RegExpRouter architecture makes it difficult to perform detailed routing checks during the `add()` phase, so we would like to perform the checks later, when all routing information is available. This would be a different behavior than the TrieRouter, but I think that the different timing of detection would be permissible with respect to this issue.

While this check can do on the "initial request" as it has been in the current master branch, I think it would be better if it could do on prior phase.

As such, I hope to have an API that allows the router to know that `hono.fire()` has been executed.

This API also allows control over when to `delete this.routeData`, so that RegExpRouter, like TrieRouter, can call `match()` and then `add()`. This makes it easier to write tests.

```typescript
router.add('GET', '/:type/:action', 'foo')
const res = router.match('GET', '/posts/123')
expect(res).not.toBeNull()
router.add('GET', '/posts/:id', 'bar') // error on the current branch. API would allow us to do this.
```

### DIFF
--- a/src/hono.ts
+++ b/src/hono.ts
@@ -241,6 +241,7 @@ export class Hono<E = Env, P extends string = ''> extends defineDynamicClass()<E
   }
 
   fire(): void {
+    this.#router.prepare()
     addEventListener('fetch', (event: FetchEvent): void => {
       event.respondWith(this.handleEvent(event))
     })
--- a/src/router.ts
+++ b/src/router.ts
@@ -4,6 +4,9 @@ export const METHOD_NAME_ALL_LOWERCASE = 'all' as const
 export abstract class Router<T> {
   abstract add(method: string, path: string, handler: T): void
   abstract match(method: string, path: string): Result<T> | null
+  prepare(): void {
+    // preprocessing, validity checks, etc.
+  }
 }
 
 export class Result<T> {
--- a/src/router/reg-exp-router/router.test.ts
+++ b/src/router/reg-exp-router/router.test.ts
@@ -77,7 +77,7 @@ describe('Order independent', () => {
   })
 
   it('abstract -> concrete', async () => {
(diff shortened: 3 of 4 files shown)


yusukebe:
@usualoma Thank you!

> Check for duplicate parameter names

Great! This feature is really wanted.

> Add router.prepare and invoke #router.prepare() in hono.fire

Actually, I don't agree with this. Because whether calling`fire()` is not promised.
With [Cloudflare Module Worker format](https://developers.cloudflare.com/workers/learning/migrating-to-module-workers/), we don't use `addEventListener`. We export `fetch` like this:

```ts
export default {
  fetch(request: Request, env: Env, event: FetchEvent) {
    return app.fetch(request, env, event)
  },
}
```

or just like this:

```ts
export default app
```

So we can't determine the timing that the user has prepared to set the routings.

> While this check can do on the "initial request" as it has been in the current master branch, I think it would be better if it could do on prior phase.

I think it's OK to go with the way for the `master branch`.

usualoma:
Oh, I didn't study enough.

I will close this PR and rebuild it with the following feature.

> Check for duplicate parameter names

PR #234, opened 2022-05-15
Tweaks return values of functions for request parameters.

usualoma:
If no arguments are specified, returns an object containing all parameters.
This is just an idea, but what about being able to write the following.

```typescript
app.get('/posts/:id', (c) => {
  const { id } = c.req.param()
  const { q, lang } = c.req.query()
  const {
    'Content-Type': contentType,
    'X-Custom': custom,
   } = c.req.header()
}))
```

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -7,7 +7,7 @@ type Data = string | ArrayBuffer | ReadableStream
 
 export type Env = Record<string, any>
 
-export class Context<RequestParamKeyType = string, E = Env> {
+export class Context<RequestParamKeyType extends string = string, E = Env> {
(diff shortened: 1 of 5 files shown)


usualoma:
@yusukebe 

I think it would be useful to have it.
What do you think?
If it looks good, I'll write tests.

yusukebe:
Hi @usualoma 

Good feature! Let’s go with it.
I made a `develop` brunch. Please merge into the `develop` branch instead of `master`.

usualoma:
My work on this PR is done!

yusukebe:
Thank you!

PR #266, opened 2022-05-24
fix(jwt): ensure the request does not continue if the jwt token is invalid or expired

cdrx:
The current JWT middleware checks the token only after the request has run (because `await next()` is run first). I think this means that whatever code is in your handler is called, _even if_ the JWT token is invalid or expired. 

This PR changes the behaviour to bail out of the request if the token validation fails.


yusukebe:
Hi @cdrx !

You're right. I'll merge the PR, and after merging, I will add some tests for this.
Thank you!

PR #268, opened 2022-05-24
refactor(router): Make Router an interface, not a class

usualoma:
This PR includes BREAKING changes.

Currently, there is one router per hono instance. With this specification, I think it would be more convenient to simply pass the object rather than the class.

I would hope to be able to write the following

```typescript
const app1 = new Hono({ router: new MyCustomRouter() })
const app2 = new Hono({ router: new MyCustomRouter(initOptions) })

function newStaticRouter<T>() {
  const routes = {}
  return {
    add(method: string, path: string, handler: T) {
      routes[`${method}:${path}`] = { handlers: [handler], params: {} }
    }
    match(method: string, path: string) {
      return routes[`${method}:${path}`] || null
    }
  }
}
const app3 = new Hono({ router: newStaticRouter() })
```

On the other hand, however, I think there are advantages to being a class, and I would agree with you even if you choose not to adopt this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -518,14 +518,14 @@ test('GET /hello is ok', async () => {
 })
 ```
 
-## routerClass
+## router
 
-The `routerClass` option specify which router is used inside. The default router is `TrieRouter`. If you want to use `RexExpRouter`, write like this:
+The `router` option specify which router is used inside. The default router is `TrieRouter`. If you want to use `RexExpRouter`, write like this:
 
 ```ts
 import { RegExpRouter } from 'hono/router/reg-exp-router'
 
-const app = new Hono({ routerClass: RegExpRouter })
+const app = new Hono({ router: new RegExpRouter() })
 ```
 
 ## Cloudflare Workers with Hono
--- a/src/compose.test.ts
+++ b/src/compose.test.ts
@@ -164,8 +164,8 @@ describe('compose with Context - 401 not authorized', () => {
     c.res = new Response('Not authorized', { status: 401 })
   }
 
-  middleware.push(handler)
   middleware.push(mHandler)
+  middleware.push(handler)
 
   it('Should return 401 Response', async () => {
     const composed = compose<Context>(middleware, onError)
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -32,7 +32,6 @@ export const compose = <C>(
           if (res && context instanceof Context) {
             context.res = res
             context.res.finalized = true
-            await dispatch(i + 1) // <--- Call next()
           }
           return context
         })
--- a/src/hono.test.ts
+++ b/src/hono.test.ts
@@ -598,13 +598,15 @@ describe('Hono with `app.route`', () => {
     const app = new Hono()
     const api = new Hono()
     const middleware = new Hono()
-    api.get('/posts', (c) => c.text('List'))
-    api.post('/posts', (c) => c.text('Create'))
-    api.get('/posts/:id', (c) => c.text(`GET ${c.req.param('id')}`))
+
     api.use('*', async (c, next) => {
       await next()
       c.res.headers.append('x-custom-a', 'a')
     })
+
+    api.get('/posts', (c) => c.text('List'))
+    api.post('/posts', (c) => c.text('Create'))
+    api.get('/posts/:id', (c) => c.text(`GET ${c.req.param('id')}`))
     app.route('/api', api)
 
     app.get('/foo', (c) => c.text('bar'))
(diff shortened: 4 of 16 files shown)


yusukebe:
Hi @usualoma 

Great!

It's good we can pass the `initOptions` and, `router` option is easy to understand compared to `routerClass`.
I'll merge it!

yusukebe:
@usualoma 

We have another problem. In this commit https://github.com/honojs/hono/commit/7cc9644b6067a9ac07151989fa2a7d924acb6bed, I wrote some code to reset `_router` property in `route` method. It's necessary for implementing like `api = new Hono(); app.route('/api', api)`. We have to reset routing in the router.
https://github.com/honojs/hono/blob/7cc9644b6067a9ac07151989fa2a7d924acb6bed/src/hono.ts#L151

In this branch, can we reset `router` property?  How about make `reset` method on the Router or do you have a good idea?

```ts
export interface Router<T> {
  add(method: string, path: string, handler: T): void
  match(method: string, path: string): Result<T> | null
  reset(): void
}
```

usualoma:
@yusukebe 
Oh, I have an idea. Give me a minute.

usualoma:
@yusukebe 
Uh, in my opinion, the previous implementation is the expected behavior.

```diff
diff --git a/src/hono.ts b/src/hono.ts
index f4c15cb..958b587 100644
--- a/src/hono.ts
+++ b/src/hono.ts
@@ -81,7 +81,6 @@ interface Route<E extends Env> {
   path: string
   method: string
   handler: Handler<string, E>
-  order: number
 }
 
 export class Hono<E = Env, P extends string = '/'> extends defineDynamicClass()<E, P, Hono<E, P>> {
@@ -139,26 +138,14 @@ export class Hono<E = Env, P extends string = '/'> extends defineDynamicClass()<
   }
 
   route(path: string, app?: Hono<any>): Hono<E, P> {
-    if (!app) {
-      this._tempPath = path
-      return this
+    this._tempPath = path
+    if (app) {
+      app.routes.map((r) => {
+        this.addRoute(r.method, r.path, r.handler)
+      })
+      this._tempPath = null
     }
 
-    app.routes.map((r) => {
-      r.path = mergePath(path, r.path)
-    })
-    const routes = [...this.routes, ...app.routes]
-    this._router = new this.routerClass()
-
-    routes
-      .sort((a, b) => {
-        return a.order - b.order
-      })
-      .map((r) => {
-        const path = r.path
-        this.addRoute(r.method, path, r.handler)
-      })
-
     return this
   }
 
@@ -192,7 +179,7 @@ export class Hono<E = Env, P extends string = '/'> extends defineDynamicClass()<
       path = mergePath(this._tempPath, path)
     }
     this._router.add(method, path, handler)
-    const r: Route<E> = { path: path, method: method, handler: handler, order: this.routes.length }
+    const r: Route<E> = { path: path, method: method, handler: handler }
     this.routes.push(r)
   }
 
```

I think that the following results are confusing.

```typescript
app.use('/api/*', function a(){ })
app.get('/api/*', function b(){ })
api.use('*', function c(){ })
api.use('*', function d(){ })
app.route('/api', api)

await app.request('http://localhost/api')

// => a() => c() => b() => d()
```

I think it is easier to control if they are called `a => b => c => d` in the registration order.

yusukebe:
Hi @usualoma 

In the previous implementation, the test on this line fails. Right?
https://github.com/honojs/hono/blob/e4b29baa18305a2a215c1d3214653a4534b07931/src/hono.test.ts#L640

I'd like to write like this. But in the previous one, it will fail.

```ts
describe.only('`route`', () => {
  const app = new Hono()
  const api = new Hono()
  const middleware = new Hono()

  api.use('*', async (c, next) => {
    await next()
    c.res.headers.append('a', 'a')
  })

  api.get('/posts', (c) => c.text('posts'))

  middleware.use('*', async (c, next) => {
    await next()
    c.res.headers.append('b', 'b')
  })

  app.route('/api', api)
  app.route('/api', middleware)

  it('Should return Response', async () => {
    const res = await app.request('http://localhost/api/posts')
    expect(res.status).toBe(200)
    expect(await res.text()).toBe('posts')
    expect(res.headers.get('a')).toBe('a') // fail! <--- expected `a` but null
    expect(res.headers.get('b')).toBe('b') // fail !<--- expected `b` but null
  })
})
```

yusukebe:
@usualoma 

Sorry, this may be a bug in the router.
Scoring is wired...

```
    [
      {
        handler: [AsyncFunction (anonymous)],
        order: 2.03,
        name: 'ALL /api/*'
      },
      {
        handler: [AsyncFunction (anonymous)],
        order: 2.03,
        name: 'ALL /api/*'
      },
      {
        handler: [Function (anonymous)],
        order: 2.02,
        name: 'GET /api/posts'
      }
    ]
```

yusukebe:
@usualoma 

That was a bug of trie-router. I've fixed it #271 and merged it into master!

yusukebe:
Hi @usualoma 

Can you fix this error?
https://github.com/honojs/hono/runs/6600495718?check_suite_focus=true#step:6:9

usualoma:
@yusukebe 
Oh sorry, I fixed it with 4dd5cc2.

PR #288, opened 2022-05-31
[POC] Performance tuning

usualoma:
`cd benchmarks/handle-event && npm run start` in my environment with this PR.

```
hono - trie-router(default) x 102,452 ops/sec ±5.76% (85 runs sampled)
hono - regexp-router x 145,484 ops/sec ±2.92% (85 runs sampled)
itty-router x 72,637 ops/sec ±3.09% (85 runs sampled)
sunder x 134,340 ops/sec ±2.42% (87 runs sampled)
worktop x 81,620 ops/sec ±4.09% (84 runs sampled)
Fastest is hono - regexp-router
```

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -1,4 +1,3 @@
-import { HonoResponse } from './response'
 import type { StatusCode } from './utils/http-status'
 import { isAbsoluteURL } from './utils/url'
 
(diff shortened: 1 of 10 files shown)


src/context.ts
```diff
@@ -37,16 +38,19 @@ export class Context<RequestParamKeyType extends string = string, E = Env> {
...
     this.notFoundHandler = notFoundHandler
+    this.finalized = false
+  }
 
-    if (!this.res) {
-      const res = new HonoResponse(null, { status: 404 })
-      res._finalized = false
-      this.res = res
-    }
+  get res(): Response {
```
yusukebe, line 41: By this implementation, `HonoResponse` is not necessary. Right?

usualoma: I thought so.

yusukebe: That's good!

Issue #291, opened 2022-06-01
JWT Middleware seems to expose secret


sdarnell:
The JWT middleware validates the token against the secret, but if the token is not present or there's an error, the secret is exposed to the client in the WWW-Authenticate header.

For example:
```
    if (parts.length !== 2) {
      ctx.res = new Response('Unauthorized', {
        status: 401,
        headers: {
          'WWW-Authenticate': 'Basic ${options.secret}',
        },
      })
      return
    }
```

There's a bit more info about what WWW-Authenticate should include: https://www.rfc-editor.org/rfc/rfc6750#section-3
Note the error fields.

metrue:
Thanks a lot raising this out @sdarnell , it's a terrible bug, just creating a quick PR to fix it. https://github.com/honojs/hono/pull/294

PR #294, opened 2022-06-02
fix(token): correct 'WWW-Authenticate' header in response

metrue:
This PR is to fix https://github.com/honojs/hono/issues/291

FYI: https://www.rfc-editor.org/rfc/rfc2617#page-8

### DIFF
--- a/src/middleware/jwt/index.test.ts
+++ b/src/middleware/jwt/index.test.ts
@@ -33,7 +33,7 @@ describe('Jwt Auth by Middleware', () => {
     const req = new Request('http://localhost/auth/a')
     const res = await app.request(req)
     expect(res).not.toBeNull()
-    expect(res.status).toBe(401)
+    expect(res.status).toBe(400)
     expect(await res.text()).toBe('Unauthorized')
     expect(res.headers.get('x-foo')).toBeFalsy()
   })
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @metrue 

Thank you for fixing it!

sdarnell:
Thanks for making the change @metrue.

Looking at it though, I'd argue that the return code should be 401 Unauthorized not 400 Bad request, whenever there is missing or invalid authentication headers.

https://www.rfc-editor.org/rfc/rfc2617#section-3.2.1
> 3.2.1 The WWW-Authenticate Response Header
> 
>    If a server receives a request for an access-protected object, and an
>    acceptable Authorization header is not sent, the server responds with
>    a "401 Unauthorized" status code, and a WWW-Authenticate header...

Also, and this is much less critical, I notice that the realm in the WWW-Authenticate header for the basic auth middleware can be specified. I suspect all auth middleware providers should do something similar.
The additional error info in your JWT change seems a good improvement.

See
https://github.com/honojs/hono/blob/master/src/middleware/basic-auth/index.ts#L44
https://github.com/honojs/hono/blob/master/src/middleware/basic-auth/index.ts#L70

metrue:
@sdarnell 
> Looking at it though, I'd argue that the return code should be 401 Unauthorized not 400 Bad request, whenever there is missing or invalid authentication headers.

Yeah, it's a little hesitation between `400` and `401`,  and I agree it totally a `401` here. PR: https://github.com/honojs/hono/pull/296 

> I notice that the realm in the WWW-Authenticate header for the basic auth middleware can be specified. I suspect all auth middleware providers should do something similar.
The additional error info in your JWT change seems a good improvement.
I think it's good idea, let me think about and can do the improvement.

PR #305, opened 2022-06-08
Export `md5` function in src/utils/crypto.ts

erhlee-bird:
Export an additional `md5` function in the crypto utility.

Useful when working with R2 and providing a MD5 hash for verification.

```
import { md5 } from "hono/utils/crypto";

...
const md5Digest = md5(body);
const object = await bucket.put(key, body, {md5: md5Digest});
```

It is worth noting, however, that "MD5 is not part of the WebCrypto standard but is supported in Cloudflare Workers for interacting with legacy systems that require MD5." <sup>[1]</sup>

[1]: https://developers.cloudflare.com/workers/runtime-apis/web-crypto/


yusukebe:
Hi @erhlee-bird !

Thank you for your PR. It's OK to export the `md5` function. Could you write the test for `md5` on `src/utils/crypto.test.ts`? Then, I'll merge it.

PR #306, opened 2022-06-09
feat: jsx middleware

### LINKED ISSUE #300: JSX middleware
We want to have ***JSX middleware***. This was discussed in this issue #289 .

I want to make the following:

* Just support for Sever-Side Rendering. Don't support the browser side. Render it as strings, not virtual DOM
* Make it as middleware
* Zero-dependencies
* Use `c.render` method
* Support `props.children`

Code example:

```ts
import { Hono } from 'hono'
import { jsx } from 'hono/jsx'
import { Header, Footer } from './component'

const app = new Hono()

const Layout = (props) => {
  <html>
    <body>{props.children}</body>
  </html>
}

type Props = {
  message: string
}

const Top = <Props>(props) => {
  return (
    <Layout>
      <Header />
        <h1>{ props.message }</h1>
      <Footer />
    </Layout>
  )
}

app.get('/', (c) => {
  const message = 'Hello Hono!'
  return c.render(<Top message={message} />)
})

app.fire()
```

yusukebe:
Close #300

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,7 @@ Built-in middleware make _"**Write Less, do more**"_ in reality. You can use a l
 - [CORS](https://github.com/honojs/hono/tree/master/src/middleware/cors/)
 - [ETag](https://github.com/honojs/hono/tree/master/src/middleware/etag/)
 - [GraphQL Server](https://github.com/honojs/hono/tree/master/src/middleware/graphql-server/)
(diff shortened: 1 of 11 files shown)


usualoma:
I would prefer that #309 be included, but even if it is not, I too think that this PR already meets the criteria for  a Minimum Viable Product.

PR #307, opened 2022-06-09
perf(utils/html): Improve performance `escape()`

usualoma:
* Replace all in one `replace()` call.
* We do not need to escape "'".


yusukebe:
Thank you!

PR #308, opened 2022-06-10
feat(middleware/jsx): Support dangerouslySetInnerHTML and some special spec

usualoma:
I have implemented the basic JSX specification using the following as a reference.

https://en.reactjs.org/docs/jsx-in-depth.html


usualoma:
@yusukebe 

> I was going to implement the same feature

Uh, sorry.


> I don't think we need any more features, what do you think?

There is one feature that would be nice to have...wait a minute.

yusukebe:
> Uh, sorry.

Do not worry!

> There is one feature that would be nice to have...wait a minute.

All right.

PR #309, opened 2022-06-10
feat(middleware/jsx): Introduce memo.

usualoma:
This PR provide a memo that can be used in much the same way as React.memo.

I think the React.memo is "do not recalculate the Virtual DOM unless there are changes in the props".
There is no Virtual DOM in hono's jsx engine, but I think we can give a meaning to memo that says "return a calculated EscapedString if there are no changes in props".

This can be used to avoid recalculating the following JSX on every render
https://github.com/yusukebe/iekei/blob/nano-jsx/src/components/footer.tsx

### DIFF
--- a/src/middleware/jsx/index.test.tsx
+++ b/src/middleware/jsx/index.test.tsx
@@ -1,5 +1,5 @@
 import { Hono } from '../../hono'
-import { h, jsx } from '.'
+import { h, jsx, memo } from '.'
 
 describe('JSX middleware', () => {
   const app = new Hono()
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @usualoma !

I think it is a very nice feature because it can increase the performance with a few code changes. I love it.

I've tried to use `memo` in this PR with my small application including the below component.

```ts
export const Footer = memo(() => {
  return (
    <footer>
      <address>
        powered by <a href='https://github.com/honojs/hono'>Hono</a>
      </address>
    </footer>
  )
})
```

It got faster.

Before:

スクリーンショット 2022-06-10 12 49 34

After:

スクリーンショット 2022-06-10 12 49 16

Let's go with it.

usualoma:
🎉 
I don't intend to add anything to this PR, so please merge if you like.
I think shallowEqual can be improved in some edge cases, but I think it is mostly OK, so I think it is fine for now.

PR #310, opened 2022-06-10
feat(middleware/jsx): We need Fragment.

usualoma:
I think we also need Fragment.

### Why accept key?

Although `key` is not used in hono, it would be useful to be able to bring the template with React as is, so this was added for compatibility.

> key is the only attribute that can be passed to Fragment. 
https://reactjs.org/docs/fragments.html#keyed-fragments


yusukebe:
Thank you for your PR. Yes, we need Fragment!

Issue #289, opened 2022-05-31
micro-template.js middleware


yusukebe:
We want to have template engines for server-side rendering. But, almost all are using `eval` or depending on node `fs`, so it does not work in Cloudflare Workers. *Mustache* is the only one to work. Another way, **micro-temple.js** may work. I'll try to make *micro-template middleware*.

ref: <https://github.com/cho45/micro-template.js>

usualoma:
@yusukebe If we can return HTML from JSX syntax, can we solve some of the demands of the template engine?

https://github.com/honojs/hono/compare/master...usualoma:tsx-feature

yusukebe:
@usualoma 

Exactly. JSX is one of the good ideas. When I tried to make a web page for *Sever-Side Rendering* with Hono, I've used **[nano-jsx](https://nanojsx.io)**. You can see the source code here: 

https://github.com/yusukebe/iekei/tree/nano-jsx/src

I want to do that. **nano-jsx** is great!. It's simple and easy to use. But I've noticed that it's pretty good to provide a more tiny JSX function on Hono.

Points:

* Just support for Sever-Side Rendering (Don't support browser side)
* Make it as middleware
* Zero-dependencies
* Use `c.render` method

I'm assuming a code like this:

```tsx
import { Hono } from 'hono'
import { jsx } from 'hono/jsx'
import { Header, Footer } from './component'

const app = new Hono()

const Layout = (props) => {
  <html>
    <body>{props.children}</body>
  </html>
}

type Props = {
  message: string
}

const Top = <Props>(props) => {
  return (
    <Layout>
      <Header />
        <h1>{ props.message }</h1>
      <Footer />
    </Layout>
  )
}

app.get('/', (c) => {
  const message = 'Hello Hono!'
  return c.render(<Top message={message} />)
})

app.fire()
```

usualoma:
@yusukebe 

Thank you!
I think we can optimize for speed by focusing on server-side rendering.

We can memoize `<Header />` and `<Footer />` that don't take props (as strings, not virtual DOM), like the example you gave.

Let me think about this in another issue.

yusukebe:
I think this issue could be fulfilled with *jsx* middleware. So, close this.

PR #312, opened 2022-06-10
feat: Introduce html template literals tag.

usualoma:
This PR is a small addition but one that I believe has great potential.
What we will be able to do is explained below.

## Advance information

There is a precedent for a library that handles HTML template literals with a tag named "html", and I refer to this as an idea.

* https://github.com/lit/lit/tree/main/packages/lit-html
* https://github.com/developit/htm

Thanks to these libraries, VisualStudioCode and vim also interprets template literals as HTML, allowing syntax highlighting and formatting to be applied.

* https://marketplace.visualstudio.com/items?itemName=bierner.lit-html
* https://github.com/MaxMEllon/vim-jsx-pretty

### Render HTML

```typescript
app.get('/:username', (c) => {
  const { username } = c.req.param()
  return c.html(
    html`<!DOCTYPE html>
      <h1>Hello! ${username}!</h1>`
  )
})
```

### Insert snippet into JSX

```typescript
const snippet = html`
  <script async src="https://www.googletagmanager.com/gtag/js?id=MEASUREMENT_ID"></script>
  <script>
    window.dataLayer = window.dataLayer || []
    function gtag() {
      dataLayer.push(arguments)
    }
    gtag('js', new Date())

    gtag('config', 'MEASUREMENT_ID')
  </script>
`

app.get('/', (c) => {
  return c.render(
    <html>
      <head>
        <title>Test Site</title>
        {snippet}
      </head>
      <body>Hello!</body>
    </html>
  )
})
```

### Insert inline script into JSX

```typescript
app.get('/', (c) => {
  return c.render(
    <html>
      <head>
        <title>Test Site</title>
        {html`
          <script>
            // No need to use dangerouslySetInnerHTML. If you write it here, it will not be escaped.
          </script>
        `}
      </head>
      <body>Hello!</body>
    </html>
  )
})
```

### Act as functional component

Since `html` returns an HtmlEscapedString, it can act as a fully functional component without using JSX.

#### Use `html` to speed up the process instead of `memo`

```typescript
const Footer = () => html`
  <footer>
    <address>My Address...</address>
  </footer>
`
```

#### Receives props and embeds values

```typescript
interface SiteData {
  title: string
  description: string
  image: string
  children?: any
}
const Layout = (props: SiteData) => html`
<html>
<head>
  <meta charset="UTF-8">
  <title>${props.title}</title>
  <meta name="description" content="${props.description}">
  <head prefix="og: http://ogp.me/ns#">
  <meta property="og:type" content="article">
  <!-- More elements slow down JSX, but not template literals. -->
  <meta property="og:title" content="${props.title}">
  <meta property="og:image" content="${props.image}">
</head>
<body>
  ${props.children}
</body>
</html>
`

const Content = (props: { siteData: SiteData; name: string }) => (
  <Layout {...props.siteData}>
    <h1>Hello {props.name}</h1>
  </Layout>
)

app.get('/', (c) => {
  const props = {
    name: 'World',
    siteData: {
      title: 'Hello <> World',
      description: 'This is a description',
      image: 'https://example.com/image.png',
    },
  }
  return c.render(<Content {...props} />)
})
```

image

I understand that using multiple template engines within a single project is generally a bad option.
Nevertheless, it doesn't look too bad, and I think it is possible to both improve execution speed and the developer experience by using it differently as follows.

* Layout-like component with many HTML elements, simply filling in values => template literals with `html` tag
* Call other components or perform repetitive processing => JSX

### DIFF
--- a/src/middleware/jsx/index.ts
+++ b/src/middleware/jsx/index.ts
@@ -115,6 +115,6 @@ export const memo = <T>(
   }) as FC<T>
 }
 
-export const Fragment = (props: { key?: string; children?: any }): EscapedString => {
+export const Fragment = (props: { key?: string; children?: any }): HtmlEscapedString => {
   return h('', {}, ...(props.children || []))
 }
--- a/src/middleware/jsx/index.ts
+++ b/src/middleware/jsx/index.ts
@@ -1,6 +1,7 @@
 import type { Context } from '../../context'
 import type { Next } from '../../hono'
 import { escape } from '../../utils/html'
+import type { HtmlEscapedString } from '../../utils/html'
 
 declare global {
   // eslint-disable-next-line @typescript-eslint/no-namespace
@@ -21,13 +22,11 @@ export const jsx = () => {
   }
 }
 
-type EscapedString = string & { isEscaped: true }
-
 export const h = (
   tag: string | Function,
   props: Record<string, any>,
-  ...children: (string | EscapedString)[]
-): EscapedString => {
+  ...children: (string | HtmlEscapedString)[]
+): HtmlEscapedString => {
   if (typeof tag === 'function') {
     return tag.call(null, { ...props, children: children.length <= 1 ? children[0] : children })
   }
@@ -42,7 +41,7 @@ export const h = (
         throw 'Can only set one of `children` or `props.dangerouslySetInnerHTML`.'
       }
 
-      const escapedString = new String(v.__html) as EscapedString
+      const escapedString = new String(v.__html) as HtmlEscapedString
       escapedString.isEscaped = true
       children = [escapedString]
       continue
@@ -69,13 +68,13 @@ export const h = (
 
   result += `</${tag}>`
 
-  const escapedString = new String(result) as EscapedString
+  const escapedString = new String(result) as HtmlEscapedString
   escapedString.isEscaped = true
 
   return escapedString
 }
 
-type FC<T = Record<string, any>> = (props: T) => EscapedString
+type FC<T = Record<string, any>> = (props: T) => HtmlEscapedString
 
 const shallowEqual = (a: Record<string, any>, b: Record<string, any>): boolean => {
   if (a === b) {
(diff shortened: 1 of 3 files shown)


usualoma:
By making the return value of `h` a result of `new String()` rather than a data structure that strongly depends on JSX, we can now seamlessly work with other functions beyond JSX, as in this case. I think it was a good choice to use `new String()`.

----

And yes, as in the very first example, `html` works as a simple template engine, so I think it would also solve the problem of #289.

usualoma:
### What are this PR not doing?

I could have made `html` more sophisticated so that it could be written as follows (like [htm](https://github.com/developit/htm)), but I didn't do it.

```typescript
html`
<${OtherComponent} ...${props}>
  ...
</${OtherComponent}>
`
```

I decided to keep `html` extremely simple and let JSX do the hard stuff.

yusukebe:
Hi @usualoma !

It's a good feature. It's powerful to use with JSX middleware.

I found a bug or a thing where is not implemented. If write like these:

```ts
const Layout = (props) => html`<body>
  ${props.children}
</body>``

const Top = (props) => {
  return (
    <Layout>
      <h1>Hello</h1>
      <p>{props.name}</p>
    </Layout>
  )
}

app.get('/', (c) => {
  const name = 'Hono'
  return c.html(<Top name={name} />)
})
```

But the output is not expected.

```html
<body>
  &lt;h1&gt;Hello&lt;/h1&gt;,&lt;p&gt;Hono&lt;/p&gt;
</body>
```

usualoma:
@yusukebe Thanks!

I would like to fix it with the following commit. Please review.
61a5586

yusukebe:
I've confirmed. It works well!

Thank you! I'll merge it.

usualoma:
I am wondering if I should prepare something like `mark_raw`.

https://metacpan.org/pod/Text::Xslate#mark_raw($str-:Str)-:RawStr

Rarely, it may be necessary.

### When to implement

#### Overload `html`

```
const name = 'John &quot;Johnny&quot; Smith' // already escaped
html`I am ${html(name)}`
```

### Introduce `markRaw` ( or a better name such as `raw` or `escaped`. )

```
const name = 'John &quot;Johnny&quot; Smith' // already escaped
html`I am ${markRaw(name)}`
```

usualoma:
( I Created #314 )

PR #324, opened 2022-06-17
refactor(reg-exp-router): Routes are returned in the registration order.

### LINKED ISSUE #322: Re-think route order
Hmm... I got in trouble.

I was writing an application using Hono, and I realized I couldn't do what I wanted to do. I wrote like that:

```ts
app.get('/foo/bar', (c) => {
  return c.text('/foo/bar')
})

app.get('*', (c) => {
  return c.text('*')
})
```

And access `/foo/bar`.

```http
GET /foo/bar
```

I got `*`, but I want `/foo/bar`!! However, this is a natural part of the specification. **It's expected**. Because we scored as a following:

```ts
app.get('/foo/bar', a()) // 2.1
app.get('*', b()) // 0.2
```

I found this is not friendly to use. If `b()` is a *middleware*, this behavior is good, but if *both are handlers*, it's not good that the `b()` is executed. Should we re-think the route order?

usualoma:
Fixes #322

### DIFF
--- a/src/router/reg-exp-router/router.test.ts
+++ b/src/router/reg-exp-router/router.test.ts
@@ -106,6 +106,22 @@ describe('Registration order', () => {
     expect(res).not.toBeNull()
     expect(res?.handlers).toEqual(['bar', 'foo'])
   })
+
+  it('middleware -> handler', async () => {
+    router.add('GET', '*', 'bar')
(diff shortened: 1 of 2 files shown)


yusukebe:
@usualoma Thank you!

PR #326, opened 2022-06-17
feat(context): add executionCtx to Context

hansottowirtz:
When using Module syntax, there is no FetchEvent, only ExecutionContext (FetchEvent implements ExecutionContext's functions as well, `waitUntil` and `passThroughUntilException`).

This PR adds `c.executionCtx`, which is set on both Service Worker syntax and Module syntax. `c.event` is only used on Service Worker syntax.

### DIFF
--- a/README.md
+++ b/README.md
@@ -483,10 +483,22 @@ app.use('/', async (c, next) => {
 })
 ```
 
+### c.executionCtx
+
+```ts
+// ExecutionContext object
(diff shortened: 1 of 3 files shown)


yusukebe:
Hi @hansottowirtz !

You are right. It would be better to have `c.executeCtx`. And the naming is good.

I'll merge it!

PR #378, opened 2022-07-14
Skip closing tag if it is an empty element

tani:
Hi. According to the web standard,
we should skip to render the closing tag if it is an empty tag. This pull request adds the validation for the empty element. Cheers.

### DIFF
--- a/src/middleware/jsx/index.test.tsx
+++ b/src/middleware/jsx/index.test.tsx
@@ -27,6 +27,11 @@ describe('render to string', () => {
     expect(template.toString()).toBe('<p><span>a</span><span>b</span></p>')
   })
 
+  it('Empty elements are rended withtout closing tag', () => {
+    const template = (<input />)
+    expect(template.toString()).toBe('<input>')
+  })
+
   it('Props value is null', () => {
     const template = <span data-hello={null}>Hello</span>
     expect(template.toString()).toBe('<span>Hello</span>')
--- a/src/middleware/jsx/index.ts
+++ b/src/middleware/jsx/index.ts
@@ -10,6 +10,24 @@ declare global {
   }
 }
 
+const emptyTags = [
+  'area',
+  'base',
+  'br',
+  'col',
+  'embed',
+  'hr',
+  'img',
+  'input',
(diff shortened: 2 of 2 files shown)


yusukebe:
Hi @tani ! cc: @usualoma 

That's a good point! I want to merge this PR, but one thing concerns me.

We may need "`single U+002F SOLIDUS character (/)`" for self-closing.

> Then, if the element is one of the [void elements](https://www.w3.org/TR/2010/WD-html5-20101019/syntax.html#void-elements), or if the element is a [foreign element](https://www.w3.org/TR/2010/WD-html5-20101019/syntax.html#foreign-elements), then there may be a single U+002F SOLIDUS character (/). This character has no effect on [void elements](https://www.w3.org/TR/2010/WD-html5-20101019/syntax.html#void-elements), but on [foreign elements](https://www.w3.org/TR/2010/WD-html5-20101019/syntax.html#foreign-elements) it marks the start tag as self-closing.
> https://www.w3.org/TR/2010/WD-html5-20101019/syntax.html

In this PR, the `<br></br>` JSX tag will be rendered to:

```html
<br>
```

But for compatibility with XHTML it should be with `/`:

```html
<br />
```

There are some opinions on which is better like this:

<https://github.com/prettier/prettier/issues/5246>

I am not sure which is better. However, *React* will render `<br />`. So, I think it's good we follow *React* and add `/` for self-closing.

```tsx
import { Hono } from 'hono'
import React from 'react'
import { renderToString } from 'react-dom/server'

const app = new Hono()

app.get('/', (c) => {
  const hello = (
    <p>
      Hello
      <br></br>
    </p>
  )
  return c.text(renderToString(hello))
})

export default app
```

This code will render:

```html
<p>Hello<br/></p>
```

@usualoma , do you have an opinion?

usualoma:
@tani @yusukebe 
I think it's a good improvement!
I don't have a specific opinion, but I too think it is a good way to follow React.

yusukebe:
@tani 

Thank you for fixing it.

@usualoma 

Thank you for your comment.

I'll merge it!

Issue #175, opened 2022-04-25
Measure coverage


yusukebe:

metrue:
It'd be nice if we can 
* show the coverage diff when a PR issued, I found [jest-coverage-report-action](https://github.com/ArtiomTr/jest-coverage-report-action) may help.
* show the coverage status in README.md, and I found [jest-coverage-badges](https://github.com/pamepeixinho/jest-coverage-badges).

PR #383, opened 2022-07-15
Add Jest coverage GitHub action to show Coverage Report in Pull Requests

thebearingedge:
This PR partially addresses #175. I'm not sure how you'd like to make the badge, but this will show coverage for new PRs.

PS: Your project is awesome! 🤓

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,14 @@ jobs:
       matrix:
         node-version: [16.x, 18.x]
         # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
-
+        include:
+          - node-version: 18.x
+            env:
+              COVERAGE: 1
+    permissions:
+      checks: write
+      pull-requests: write
+      contents: write
     steps:
       - uses: actions/checkout@v2
       - name: Use Node.js ${{ matrix.node-version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -23,4 +30,11 @@ jobs:
       - run: yarn install --frozen-lockfile
       - run: npm run lint
       - run: npm run build
-      - run: npm test
+      - name: Run Tests on Node.js ${{ matrix.node-version }}
+        if: ${{ matrix.env.COVERAGE != '1' }}
+        run: npm test
+      - name: Run Tests w/Coverage on Node.js ${{ matrix.node-version }}
+        if: ${{ matrix.env.COVERAGE == '1' }}
```
metrue, line 37: I think coverage can be enabled by default, any concerns?

thebearingedge: I guess I just didn't expect the outcome of running coverage for both versions of Node.js (redundant info).

metrue: I see, coverage on one version should be enough.

yusukebe:
Hi @thebearingedge !

Thank you! That's what I want. But, this GitHub Actions does not work well.
Can you fix it?

PS.
To make the badge on README, the command seems good:
https://github.com/jaywcjlove/coverage-badges-cli

thebearingedge:
@yusukebe, hello! I should have known, but it looks like we will always have the same problem with Pull Requests from forks. 

I should have realized it, but write access for fork Pull Request actions is dangerous, so maximum token permissions will be **read**.

https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

I will see if there is another way some time tomorrow, but I think an external service or proxy app with be required.

😕 sorry
Screen Shot 2022-07-15 at 8 26 36 AM

yusukebe:
@thebearingedge 

I see, but don't be sorry!
This is a complex problem. I think maybe we don't have to force it.

thebearingedge:
closing for now

PR #384, opened 2022-07-15
Fix documentation : Add proper casting to `Post` object using `json()` method

foyzulkarim:
### Problem

Using the code is causing build error as below.
image

### Reason
In the definition, we can see the `.json` method don't take any type nor any parameter. Instead it is needed to cast the response object manually.

image

`json(): Promise<JSON>;`

### Solution 

Manually cast the response object as `Post` like below

`const post = await c.req.json() as Post`

It will solve the build error.

### Special note

We also need to create the `model` file and implement methods to properly solve the other build errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -743,7 +743,7 @@ api.post(
     await auth(c, next)
   },
   async (c) => {
-    const post = await c.req.json<Post>()
+    const post = await c.req.json() as Post
     const ok = createPost({ post })
     return c.json({ ok })
   }
--- a/deno_dist/README.md
+++ b/deno_dist/README.md
@@ -743,7 +743,7 @@ api.post(
     await auth(c, next)
   },
   async (c) => {
-    const post = await c.req.json<Post>()
+    const post = await c.req.json() as Post;
     const ok = createPost({ post })
     return c.json({ ok })
   }


yusukebe:
Hi @foyzulkarim !

According to Web Standard, `json()` can not have generics. But, in Cloudflare Workers environments, we could pass the generics like `json<T>()` for adding types of the body.

```ts
const post = c.req.json<Post>()
```

With `@cloudflare/workers-types`, type hints are enabled. The definition of `json<T>()` is here:
https://github.com/cloudflare/workers-types/blob/master/overrides/http.d.ts#L2

So, it's OK to write down `c.req.json<Post>()`.

> We also need to create the model file and implement methods to properly solve the other build errors.

I know it, the code is just for explanation, so I think it does not have to work well.

---

**BUT**, it is indeed difficult to understand. I will make the website for Hono next time, so I will try to explain it clearly then. I'll close this PR. Thank you for noticing them.

foyzulkarim:
@yusukebe thanks for your explanation. The title is mentioning as `Practical example`, not the `Cloudflare example`. If it is not too much trouble, please add this small note in the readme somewhere. Otherwise whenever a new visitor will come,  everyone of them will face this error.

yusukebe:
Yes, I'll do it. Thanks again for your concern.

foyzulkarim:
Thanks for your awesome work. 🥇

PR #386, opened 2022-07-15
Add compress middleware

tani:
`compress` middleware solves the performance related to the transferring speed on the internet.
This pull request depends on the `CompressStream` [^1]. It might not work on the Bun because Safari cannot use no such a class. However, we can run it on Node and Deno. Please consider to employ this middleware in your project. Cheers.

[^1]: https://developer.mozilla.org/en-US/docs/Web/API/CompressionStream/CompressionStream

### DIFF
--- a/src/middleware/compress/index.test.ts
+++ b/src/middleware/compress/index.test.ts
@@ -2,13 +2,6 @@ import { Hono } from '../../hono'
 import { compress } from '.'
 
 describe('Parse Compress Middleware', () => {
-  
-  if (parseInt(process.versions.node.split('.')[0]) < 17) {
-    it('node version', () => {
-      expect(parseInt(process.versions.node.split('.')[0])).toBeLessThan(17)
-    })
-    return
-  }
 
   const app = new Hono()
 
--- a/src/middleware/compress/index.test.ts
+++ b/src/middleware/compress/index.test.ts
@@ -1,0 +1,62 @@
+import { Hono } from '../../hono'
+import { compress } from '.'
+
+describe('Parse Compress Middleware', () => {
+  
+  if (parseInt(process.versions.node.split('.')[0]) < 17) {
+    it('node version', () => {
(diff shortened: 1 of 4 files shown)


yusukebe:
Hi @tani !

On Cloudflare Workers, the response body is compressed automatically. So, we don't care about making *compress middleware*. But, thinking about Deno, indeed we need this middleware for better performance. Nice work.

When the CI is passed, I'll merge this PR!

yusukebe:
@tani 

On `jest-environment-miniflare`, CompressionStream is implemented at v@2.6.0.

https://github.com/cloudflare/miniflare/issues/206

So, maybe we have to update `jest-environment-miniflare` by this command:

```
yarn add -D jest-environment-miniflare
```

src/middleware/compress/index.ts
```diff
@@ -0,0 +1,24 @@
+import * as WebStreams from 'node:stream/web'
```
yusukebe, line 1: @tani 

Hey, we must not use **node** module. Only use web standard API available on Cloudflare and Deno.

tani: Noted. Cheers.

src/middleware/compress/index.test.ts
```diff
@@ -0,0 +1,62 @@
+import { Hono } from '../../hono'
+import { compress } from '.'
+
+describe('Parse Compress Middleware', () => {
+  
+  if (parseInt(process.versions.node.split('.')[0]) < 17) {
```
yusukebe, line 6: @tani 

I think this version check is not needed. We don't use node `node:stream/web`.

tani: Noted. Thanks

yusukebe:
@tani 

Thank you for fixing them.

I'll merge `next` branch for **v2.0.0**. It' will be released soon!

tani:
Hi, I am really excited to hear that.
I am looking forward to the new release. Cheers :beers:

PR #399, opened 2022-07-17
refactor(context): executionCtx always returns ExecutionContext.

### LINKED ISSUE #385: How to modify the definition of executionCtx in Context interface?
CleanShot 2022-07-16 at 00 01 27@2x

I'm using the js module syntax, but the tsc tell me that executionCtx maybe undefined. so how can I resolve this problem?

I know I can use `c.executionCtx!.waitUntil`, is there has another solutions ?

usualoma:
This PR fixes #385.

### Summary

This PR will replace executionCtx() with a getter method to check the type to runtime.

#### May executionCtx() return a FetchEvent?

Since the FetchEvent object satisfies the ExecutionContext interface, how about returning a FetchEvent from executionCtx() even if we have a FetchEvent? I think users will be able to use executionCtx().waitUntil() without being aware of whether it is Service Worker syntax or new Module Worker format.

#### Checking the type with instanceof

In this PR, I am checking the type in `this._executionCtx instanceof FetchEvent`.
I am not familiar with the performance differences, but is there anything that makes it less desirable than `'respondWith' in this._executionCtx`?

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -11,8 +11,8 @@ type Env = Record<string, any>
 export interface Context<RequestParamKeyType extends string = string, E = Env> {
   req: Request<RequestParamKeyType>
   env: E
-  event: FetchEvent | undefined
-  executionCtx: ExecutionContext | undefined
+  event: FetchEvent
+  executionCtx: ExecutionContext
   finalized: boolean
 
   get res(): Response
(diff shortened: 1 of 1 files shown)


bytemain (approved):

yusukebe:
Hi @usualoma !

Thank you for your implementation.

> how about returning a FetchEvent from executionCtx() even if we have a FetchEvent?

It's good that the `executionCxt()` returns FetchEvent if it is not ExecutionContext.

> I think users will be able to use executionCtx().waitUntil() without being aware of whether it is Service Worker syntax or new Module Worker format.

I agree with you.

> Checking the type with instanceof

I think `this._executionCtx instanceof FetchEvent` is simpler and better than the previous implementation, it's OK.

All right, let's go this.

PR #400, opened 2022-07-17
Pass path to serve-static middleware when using cloudflare module workers

raytracer:
I had a problem using the `{path: "..."}` option with the serve-static middleware when using cloudflare module workers. This should hopefully fix it.


yusukebe:
Hi @raytracer !

Ah, I forgot it. Thank you for your help!
I'll release v2.0.2 including this fix later.

PR #402, opened 2022-07-17
Update context.ts

bkniffler:
Allow typecast in context.get(), e.g.

```ts
const user = context.get<User>('user');
```


yusukebe:
Nice! Thank you!

PR #424, opened 2022-07-27
Create middleware

winston0410:


yusukebe:
Hi @winston0410 !

Thank you for your PR, but we can't accept this.
Because the middleware in this repository should not depend on other libraries.
This middleware depends on `@cfworker/json-schema`.

So, it should be a "*third-party*" middleware. We can make the validator middleware together.
However, the code in this PR has room for improvement.
We have to write tests, we should consider error handling.

Let's discuss this in the another place.

Issue #414, opened 2022-07-20
Extending Context


yusukebe:
There is a request to extend Context. Discussed at #410 and https://github.com/honojs/honojs.dev/pull/1
The possibility of adding `c.afunction()` by the middleware is a good feature, I think.

But I don't have any idea to implement it. One idea is using `declare module`:

```
declare module 'hono' {
   interface Context {
     afunction()
   }
 }
```

It seems to be not smart. Is there someone have other ideas?

RichiCoder1:
A recent pattern in typescript is something called the "builder" pattern, in which you expose methods which return the base type plus some additional type information:

https://stackoverflow.com/questions/69853720/using-builder-pattern-to-create-a-strongly-typed-diverse-record-like-redux-too

A possibility is something like [this playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAwg9gO2BAHsKBeKBvAvgWACgjRIoBZASwBNqAbCAdwEMAnCAMQFcEBjAHgAq8JKnRYRyNAD5MUABS9gKAFxRhiKcAA0UBGLXyAlJlkAFVnAC2lAM4R+ANzg1pJjOcs37-AEoRbMER7KAAfKGdXAG4ScGgqWgYWdm4+AHVKYAALSTEhXLQ5Ap11YoBBWkzKYLk8WSxsIgBIXk0DUra0Cuoq4JjCJqzmBHoIVjUE0eTOHgENUUKAMg6F4G7ehFtpftx+2LIACUQ4MrAwfM7xWEv6nGbW1bV5rX6mrnt5Sy5kNVtgVkoCAA5robIkmGwIBMaFNIak5sU3GojggTmcLqtts13g5nmJ1sBqptarhpJ84N8oVA-gDgaCYUlIdDwdN4RlssUMVpdHiupVCcEkVAUWjzrz0MtxQSiVsdntCA8-lBmGdkcdTmA5AguHQ6MrbMqECB5SqwAA6HHyADkACorbpGgMHlo1I6mk1kH81Farc1cNpmkMRgxxvqQHwFO5ZHgA4RcEYLR9bfawxHFMpdPo0FG7k7gnAGGa6HAgfJSBA4AAzKBKFBmz3iDBYABENMBQObRiiUAA9D2oP8uBBmswWJk9GJjDsjEQiEA) though that's a very simplified implementation.

migueloliveiraz:
It must be something good thinking. it implies in all design philosophy of project

[...]

PR #429, opened 2022-07-28
feat(context): Introduce ContextVariableMap.

usualoma:
Allows the user to type the value of `ctx.set('variable', var)` or `ctx.get('variable')`.

This is one of the ideas to solve the #414 problem, but even if this were not used to solve #414, I think this would still be a useful feature.

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -1,4 +1,4 @@
-import type { NotFoundHandler } from './hono'
+import type { ContextVariableMap, NotFoundHandler } from './hono'
 import type { CookieOptions } from './utils/cookie'
 import { serialize } from './utils/cookie'
 import type { StatusCode } from './utils/http-status'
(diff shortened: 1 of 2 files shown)


src/hono.ts
```diff
@@ -7,6 +7,8 @@ import { METHOD_NAME_ALL, METHOD_NAME_ALL_LOWERCASE } from './router'
 import { TrieRouter } from './router/trie-router' // Default Router
 import { getPathFromURL, mergePath } from './utils/url'
 
+export interface ContextVariableMap {}
```
RichiCoder1, line 10: I forget the specifics, but this should probably extend `Record<string, any>`

src/context.ts
```diff
@@ -20,7 +20,10 @@ export interface Context<RequestParamKeyType extends string = string, E = Env> {
   header: (name: string, value: string) => void
   status: (status: StatusCode) => void
   set: (key: string, value: any) => void
```
RichiCoder1, line 22: Set should probably also attempt to enforce the type of `value` if it's specified on `ContextVariableMap` for extra type safety.

usualoma: Thanks!
Fixed in 1852968.

src/hono.ts
usualoma: In this PR, there is no need to extend `Record<string, any>` since `ContextVariableMap` is a type for type definition and no object is created. (This is like [HTMLElementTagNameMap](https://github.com/microsoft/TypeScript/blob/949fffb1e0b683a3b4fceb915e7fd4ddb4b54798/lib/lib.dom.d.ts#L18029-L18147).)
Also, not extending `Record<string, any>` will give better completion results.

yusukebe:
Thank you! Looks good!

PR #430, opened 2022-07-28
fix(jsx): Fix for JSX checked and selected props

### LINKED ISSUE #427: `checked` in JSX middleware
When I want to set the value of attributes according to the condition which is true or false, it is to be:

```tsx
<input value={true ? 'foo' : 'bar'} />
```

But how can I make the checkbox to be unchecked?
Can I render `checked` if it is true, or **do not** render the `checked` if it is false:

I want to make the below tag:

```tsx
<input type='checkbox' checked /> <!-- If true -->
<input type='checkbox' /> <!-- If false -->
```

But, I can't:

```tsx
<input type='checkbox' checked={false}/> <!-- NG: It will be checked -->
<input type='checkbox' {true ? 'checked' : ''}/> <!-- NG: Syntax error, and it will be checked -->
```

How can I remove `checked` if it is false? I think we should implement it.

Ref: [html - What values for checked and selected are false? - Stack Overflow](https://stackoverflow.com/questions/4228658/what-values-for-checked-and-selected-are-false)

digitalmio:
Fixes #427:
- `checked` and `selected` will not be rendered when set to false
- `checked="checked"` and `selected="selected"` will be rendered when no value set or set to `true`

### DIFF
--- a/src/middleware/jsx/index.test.tsx
+++ b/src/middleware/jsx/index.test.tsx
@@ -75,19 +75,49 @@ describe('render to string', () => {
     })
   })
 
-  describe('Special case for input "checked" and "selected" params', () => {
+  describe('Special case for input boolean params', () => {
     it('default prop value for checked', () => {
       const template = <input type='checkbox' checked />
-      expect(template.toString()).toBe('<input type="checkbox" checked="checked"/>')
+      expect(template.toString()).toBe('<input type="checkbox" checked=""/>')
     })
 
     it('default prop value for checked={true}', () => {
       const template = <input type='checkbox' checked={true} />
-      expect(template.toString()).toBe('<input type="checkbox" checked="checked"/>')
+      expect(template.toString()).toBe('<input type="checkbox" checked=""/>')
     })
 
     it('no prop for checked={false}', () => {
       const template = <input type='checkbox' checked={false} />
+      expect(template.toString()).toBe('<input type="checkbox"/>')
+    })
+
+    it('default prop value for disabled', () => {
+      const template = <input type='checkbox' disabled />
+      expect(template.toString()).toBe('<input type="checkbox" disabled=""/>')
+    })
+
+    it('default prop value for disabled={true}', () => {
+      const template = <input type='checkbox' disabled={true} />
+      expect(template.toString()).toBe('<input type="checkbox" disabled=""/>')
+    })
+
+    it('no prop for disabled={false}', () => {
+      const template = <input type='checkbox' disabled={false} />
+      expect(template.toString()).toBe('<input type="checkbox"/>')
+    })
+
+    it('default prop value for readonly', () => {
+      const template = <input type='checkbox' readonly />
+      expect(template.toString()).toBe('<input type="checkbox" readonly=""/>')
+    })
+
+    it('default prop value for readonly={true}', () => {
+      const template = <input type='checkbox' readonly={true} />
+      expect(template.toString()).toBe('<input type="checkbox" readonly=""/>')
+    })
+
+    it('no prop for readonly={false}', () => {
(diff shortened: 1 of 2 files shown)


src/middleware/jsx/index.ts
```diff
@@ -54,6 +54,11 @@ const jsxFn = (
       continue
     } else if (v === null || v === undefined) {
       continue
+    } else if (['checked', 'selected'].includes(propsKeys[i]) && v === false) {
```
yusukebe, line 57: I've investigated HTML specs. There are other boolean attributes, `disabled`, `readonly` and`multiple`.
See: https://stackoverflow.com/questions/706384/boolean-html-attributes
Lets' add them.

src/middleware/jsx/index.test.tsx
```diff
@@ -75,6 +75,50 @@ describe('render to string', () => {
     })
   })
 
+  describe('Special case for input "checked" and "selected" params', () => {
+    it('default prop value for checked', () => {
+      const template = <input type='checkbox' checked />
+      expect(template.toString()).toBe('<input type="checkbox" checked="checked"/>')
```
yusukebe, line 81: I think it's better to render the attribute with no value like the below:

```
<input type="checkbox" checked/>'
```

yusukebe:
Hi @digitalmio !

Thank you so much! I really want this!
I've commented on some things. Please check out.

src/middleware/jsx/index.test.tsx
usualoma: I think the following result is better here

```
<input type="checkbox" checked="" />
```

In #378, we made it possible to parse a rendered string as XML, just like in React, so I think it is better to adopt empty string for boolean attributes as well.

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#boolean_attributes

usualoma:
Hi, @digitalmio, @yusukebe.

Thank you!
I also commented on one point.
Best regards!

src/middleware/jsx/index.test.tsx
yusukebe: Alright! Let's go with `<input type="checkbox" checked=""/>`.
@digitalmio 
Is it OK?

digitalmio:
Hey, no worries, I'll add those 3 more to the list and can amend the value to empty string today lunchtime or after work.

digitalmio:
Hey @usualoma  @yusukebe, thinking out loud here: shall we also add list for tags where this rule should be applied (`input`, `button`, `select`, `option` etc) where on `div` it should render 'as normal'?

usualoma:
Hi, @digitalmio 

I understand that hono's JSX  implementation has the following policy

* Keep it as simple as possible.
* The output HTML will be React friendly.
    * Better to have less confusion for developers than to have Hono's own colors when it comes to JSX.

And...

* React does not distinguish by the element name.
    * `<div autoPlay={true}>text</div>` => `<div autoplay="">text</div>`
    * https://github.com/facebook/react/blob/2c7dea73659da283fa433f5fd657a5a29797fd67/packages/react-dom/src/shared/DOMProperty.js

Therefore, I think it is better to treat an attribute whose "Value" is a "Boolean attribute" in the table on the following page as a boolean attribute without distinguishing the element name.
https://html.spec.whatwg.org/#attributes-3

digitalmio:
Makes sense. Pushed code with more props added, but without distinction of the key.

usualoma (approved):
Thank you!

I'm sure we can optimize it a bit more, but the optimization can be done in another PR, and this PR has enough tests along the use cases and they are successful, and I think this is good enough to merge.

yusukebe:
Yes, we can merge.

@digitalmio 
Thank you for doing what I wanted to do!

Issue #282, opened 2022-05-30
Is not Hono fastest...?


yusukebe:
Shame on me, the benchmark scores may not be correct.

First, I say Hono is the fastest compared to other routers. The score of [the benchmark](https://github.com/honojs/hono/blob/16967676a303bf79a9e7bcd45a963dec60ad0c8b/benchmarks/handle-event/index.js) is below:

```
hono - trie-router(default) x 724,143 ops/sec ±3.63% (80 runs sampled)
hono - regexp-router x 1,236,810 ops/sec ±6.77% (72 runs sampled)
itty-router x 161,786 ops/sec ±2.28% (97 runs sampled)
sunder x 312,262 ops/sec ±2.59% (85 runs sampled)
worktop x 224,979 ops/sec ±1.13% (96 runs sampled)
Fastest is hono - regexp-router
✨  Done in 95.05s.
```

But, isn't it too fast? I wrote the minimal handler to compare it. It should be the fastest.

```ts
const minimalHandler = async () => {
  return new Response('foo')
}
```

The score is below:

```
minimal handler x 392,854 ops/sec ±3.23% (90 runs sampled)
hono - trie-router(default) x 720,882 ops/sec ±3.23% (84 runs sampled)
hono - regexp-router x 1,195,783 ops/sec ±6.53% (70 runs sampled)
itty-router x 160,515 ops/sec ±2.68% (91 runs sampled)
sunder x 303,981 ops/sec ±3.09% (86 runs sampled)
worktop x 211,622 ops/sec ±3.71% (83 runs sampled)
Fastest is hono - regexp-router
✨  Done in 101.34s.
```

In this benchmark, the score of *minimal handler* is not the fastest, and `hono - regexp-router` is still the fastest. Why??

As an experiment, I remove `async/await` for `matchRoute` in `hono.ts`.

```diff
diff --git a/src/hono.ts b/src/hono.ts
index f3bc3d8..41a0ba7 100644
--- a/src/hono.ts
+++ b/src/hono.ts
@@ -173,10 +173,7 @@ export class Hono<E = Env, P extends string = '/'> extends defineDynamicClass()<
     this.routes.push(r)
   }

-  private async matchRoute(
-    method: string,
-    path: string
-  ): Promise<Result<Handler<string, E>> | null> {
+  private matchRoute(method: string, path: string): Result<Handler<string, E>> | null {
     return this.router.match(method, path)
   }

[...]

PR #435, opened 2022-07-30
perf(jsx): Performance Tuning for JSX

usualoma:
### Benchmark script

I have added a benchmark script, albeit a simplified one, to provide a clue for discussion.

#### before

```
render x 55,038 ops/sec ±4.28% (94 runs sampled)
```

#### after

```
render x 66,330 ops/sec ±3.28% (96 runs sampled)
```


yusukebe:
Hi @usualoma !

It's very good to add the benchmarks, and it really became faster.

Perhaps, adding "React" app benchmarks using `renderToString()` in `react-dom/server` will be a good guide? But maybe the comparison is nonsense.

I think all the rest is OK

usualoma:
@yusukebe

Oh, indeed. I too think it is important to compare with other products. (Not necessarily to be the fastest, but to see where we are.)

I added a comparison with React, Preact and Nano JSX in bc35351.

```
Hono x 65,868 ops/sec ±1.04% (96 runs sampled)
React x 94,587 ops/sec ±1.40% (95 runs sampled)
Preact x 78,812 ops/sec ±0.86% (91 runs sampled)
Nano x 20,701 ops/sec ±0.98% (96 runs sampled)
Fastest is React
```

It was unexpected that React turned out to be the fastest. However, I think hono is also doing well with its small code.
We will need to do more research to determine if this is a benchmarking issue or an actual speed issue, as in #282, but for now we will accept these results as reality and hope to make further improvements in the future.

yusukebe:
@usualoma 

Great work!

I was surprised that React is the fastest.

> but for now we will accept these results as reality and hope to make further improvements in the future.

Yes!
This time, I'lll merge once. Thank you!

Issue #437, opened 2022-07-31
JWT Middleware to Support Cookies?


nathanclevenger:
Today the JWT middleware only works with authorization headers and not cookies
```
const credentials = ctx.req.headers.get('Authorization')
```

What if an optional param to the options:
```
export const jwt = (options: { secret: string; alg?: string; cookie?: string }) => {
```

Then the above code could be modified to something like:
```
const credentials = ctx.req.headers.get('Authorization') ?? ctx.req.cookie[cookie]
```

What do you think?

yusukebe:
Hi! I think it's OK.

@metrue How about you?

metrue:
@yusukebe @nathanclevenger 

Yes, I think it's a great proposal,  let me check it.

PR #441, opened 2022-08-01
feat(jwt): enable token in cookie

metrue:
Close https://github.com/honojs/hono/issues/437


yusukebe:
Thank you for your quick response!

PR #444, opened 2022-08-01
perf(jsx): JSX performance improvement

### LINKED ISSUE #439: JSX performance improvement
In #435, we found that JSX in hono is not that fast, so we will improve performance.

#### Make `escape` faster

We can make `escape` faster by avoiding using `String.prototype.replace`, as we do in react.

https://github.com/facebook/react/blob/3f3b46c845333be5d8efcd9acb06e3ff57358189/packages/react-dom/src/server/escapeTextForBrowser.js#L51-L94

#### Reduce generation of string objects

Adding strings to the buffer reduces the generation of string objects.

#### Avoid `Array.prototype.flat(Infinity)`

`Array.prototype.flat(Infinity)` is very useful, but it is slow and should not be used.

### POC

https://github.com/honojs/hono/compare/main...usualoma:hono:poc-pref-jsx

#### Benchmark

```
Hono x 246,570 ops/sec ±1.04% (88 runs sampled)
React x 132,325 ops/sec ±9.23% (82 runs sampled)
Preact x 100,303 ops/sec ±4.16% (83 runs sampled)
Nano x 24,140 ops/sec ±6.48% (82 runs sampled)
Fastest is Hono
```

I think we ought to get this done!

usualoma:
This PR fixes #439.

### Changes

There are no changes to the external specification in this PR, and we can continue to stringify templates with `(<p>text</p>).toString()`. However, there are some changes in the internal structure.

#### `jsx` returns `JSXNode` instead of `HtmlEscapedString`

Previously, upon calling `jsx`, basic stringification was completed and an `HtmlEscapedString` object was returned, but with this PR, a data-wrapped `JSXNode` is returned upon calling `jsx`. The stringification is done at `toString()` call.

### Discussion

With this PR, `escape` defined in utils/html is no longer used inside the hono repository. (Only `escapeToBuffer` is used.)

The `escape` function is fast enough and convenient for hono users to use in their applications, but depending on policy, it could be removed as unnecessary.

### Benchmark

```
Hono x 173,197 ops/sec ±2.16% (93 runs sampled)
React x 84,569 ops/sec ±2.81% (91 runs sampled)
Preact x 61,307 ops/sec ±3.21% (86 runs sampled)
Nano x 16,871 ops/sec ±5.46% (78 runs sampled)
Fastest is Hono
```

### DIFF
--- a/benchmarks/jsx/index.tsx
+++ b/benchmarks/jsx/index.tsx
@@ -25,11 +25,11 @@ const buildPage = ({ jsx, Fragment }: { jsx: any; Fragment: any }) => {
 
   const Form = () => (
     <form>
-      <input type='text' value='a' readonly tabindex={1} />
-      <input type='checkbox' value='b' checked={true} tabindex={2} />
-      <input type='checkbox' value='c' checked={true} tabindex={3} />
-      <input type='checkbox' value='d' checked={false} tabindex={4} />
-      <input type='checkbox' value='e' checked={false} tabindex={5} />
+      <input type='text' value='1234567890 < 1234567891' readonly tabindex={1} />
+      <input type='checkbox' value='1234567890 < 1234567891' checked={true} tabindex={2} />
+      <input type='checkbox' value='1234567890 < 1234567891' checked={true} tabindex={3} />
+      <input type='checkbox' value='1234567890 < 1234567891' checked={false} tabindex={4} />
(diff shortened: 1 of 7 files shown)


yusukebe:
Hi @usualoma !

Great! I confirmed that it also works on Bun.

> The escape function is fast enough and convenient for hono users to use in their applications, but depending on policy, it could be removed as unnecessary.

Yes, I think it's better to remove `utils/html`. I'll do it another PR.

I'll merge this PR!

PR #446, opened 2022-08-01
refactor(jsx): Support all the boolean attributes.

usualoma:
I think it is better to support all boolean attirbutes listed in the table below.
https://html.spec.whatwg.org/#attributes-3

And, supporting all of them will not affect performance.


usualoma:
I forgot to apply prettier, so I applied it and force pushed.

yusukebe:
Your right! Thank you.

PR #469, opened 2022-08-15
fix(compose): Support a handler that non-async and returning a promise.

usualoma:
#466 is awesome!

However, we should also support asynchronous functions that are not 'AsyncFunction'.

### DIFF
--- a/deno_dist/compose.ts
+++ b/deno_dist/compose.ts
@@ -31,11 +31,8 @@ export const compose = <C>(
       let isError: boolean = false
 
       try {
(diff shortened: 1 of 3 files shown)


yusukebe:
Hi @usualoma 

You are right! Thank you!

And, could you do `yarn denoify`? We have to generate the code for Deno into `deno_dist` manually.

usualoma:
@yusukebe 
Oh sorry, I ran `yarn denoify`.

yusukebe:
Thanks!

Issue #472, opened 2022-08-19
Types for `c.set` and `c.get`


yusukebe:
By #429, we can set the Type for `c.set` and `c.get` functions. This is a fine way to set/get the specific property used in the Middleware.

```ts
type WebClient = {
  user: string
  pass: string
}

declare module 'hono' {
  interface ContextVariableMap {
    client: WebClient
  }
}

//...

app.use('*', async (c, next) => {
  const client: WebClient = { user: 'foo', pass: 'bar' }
  c.set('client', client)
  await next()
})

app.get('/foo', (c) => {
  const client = c.get('client') // client is `WebClient`
  return c.text('foo')
})
```

![SS](https://user-images.githubusercontent.com/10682/185519526-5700022b-ca41-49a2-aaa4-14dac5f4bfe4.png)

It's a great feature, thanks to @usualoma .

Now I want that this could be made more portable and easy to use for users, not the library author. Specifically, it's good if the following code is all needed to complete the project. The point is "without `declare module`".

```ts
app.use('*', async (c, next) => {
  const client: WebClient = { user: 'foo', pass: 'bar' }
  c.set('client', client)
  // OR
  c.set<WebClient>('client', client)

  await next()
})

app.get('/foo', (c) => {
  // show the "type hint" like above in VSCode
  const client = c.get('client') // client is `WebClient`
  return c.text('foo')
})
```

This is very short, and every user can use `set/get` with Types.

**BUT**

I think this is too difficult to implement, or **no way** to implement it exists. We can't treat "Type" this way. I know this is an optimistic thought. But I want to do that. Does anyone know how to make it come true, or do you know other similar API designs to ?

mglikesbikes:
I'm solving this problem with well-named middleware and variables, if that's helpful:

```
// src/handlers/stytch.ts
import type { User } from "stytch/types/lib/shared";

export const useStytchSession = () => {
  return async (c: Context, next: Next) => {
    try {
      const authResponse = await client.sessions.authenticate({ … });
      c.set("stytchUser", authResponse.user as User);
      await next();
    } catch (e) {}
  };
};

// src/handlers/hello
import type { User } from "stytch/types/lib/shared";

app.post('/private', useStytchSession());
app.post('/private', (c:Context) {
  const user: User = c.get('stytchUser');
}
```

yusukebe:
Hi @mglikesbikes !
Thanks, but I still want a Types 🙂

yusukebe:
[...]

PR #478, opened 2022-08-23
feat(types): enable adding Types for variables used in `c.set`/`c.get`

yusukebe:
This PR is for enabling adding Types for variables used in `c.set` and `c.get`. Related to #472 

`c.set()` and `c.get()` are useful to store variables for middleware and handlers. For example, you can set the "client" object in middleware and use it later in a handler. But, these key-value do not have Types.

SS

In this PR, Types will be available for `c.set` and `c.get` by passing the Types as generics into Hono class.

SS

SS

Before this PR, the generics is used for "bindings" on Cloudflare Workers.

```ts
type Bindings = {
  KV: KVNamespace
  Storage: R2Bucket
}

const app = new Hono<Bindings>()
```

After this PR, the generics will be used for "bindings" and "variables":

```ts
type Bindings = {
  KV: KVNamespace
  Storage: R2Bucket
}

type WebClient = {
  user: string
  pass: string
}

type Variables = {
  client: WebClient
}

const app = new Hono<{ Variables: Variables; Bindings: Bindings }>()

app.get('/foo', (c) => {
  const client = c.get('client') // client is WebClient
  const kv = c.env.KV // kv is KVNamespace
  //...
})
```

I think this is a cool feature for fine DX, but it has a small breaking change that is should be told the user well.

### DIFF
--- a/deno_dist/hono.ts
+++ b/deno_dist/hono.ts
@@ -9,13 +9,30 @@ import { getPathFromURL, mergePath } from './utils/url.ts'
 
 export interface ContextVariableMap {}
 
-type Env = Record<string, any>
-export type Handler<RequestParamKeyType extends string = string, E = Env> = (
+export type Bindings = Record<string, any> // For Cloudflare Workers
+export type Variables = Record<string, any> // For c.set/c.get functions
+export type Environment = {
+  Bindings: Bindings
+  Variables: Variables
+}
+
+export type Handler<
+  RequestParamKeyType extends string = string,
+  E extends Partial<Environment> = Environment
+> = (
   c: Context<RequestParamKeyType, E>,
   next: Next
 ) => Response | Promise<Response> | Promise<void> | Promise<Response | undefined>
-export type NotFoundHandler<E = Env> = (c: Context<string, E>) => Response | Promise<Response>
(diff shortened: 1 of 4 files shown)


deno_dist/hono.ts
```diff
@@ -166,7 +186,7 @@ export class Hono<E extends Env = Env, P extends string = '/'> extends defineDyn
   private async dispatch(
     request: Request,
     eventOrExecutionCtx?: ExecutionContext | FetchEvent,
-    env?: E
+    Environment?: E['Bindings']
```
usualoma, line 189: Looks like a typo.

```suggestion
    environment?: E['Bindings']
```

usualoma (approved):
Great WORK!

I have also considered this approach but could not achieve a good result at the time. I think this PR is turning out to be a good one!

While ContextVariableMap is still useful for middleware developers, I think this PR approach is easier to use when writing individual applications.

deno_dist/hono.ts
yusukebe: Yes, it's typo.

yusukebe:
Thank you for your review. I'll merge this!

PR #487, opened 2022-08-27
[BREAKING] fix: make that `c.req.parseBody` parses only `FormData`

### LINKED ISSUE #484: Should `parseBody` only parse FormData?
Currently, `req.parseBody()` parses any objects according to content type.

<https://github.com/honojs/hono/blob/main/src/utils/body.ts>

But, I think it would be better just to restrict it to FormData / `req.formData()`.

The first reason, the type we receive will be obscured; thus, it will be `any`. `parseBody` definition is below:

```ts
parseBody<T = any>(r: Request | Response): Promise<T>
```

If it only parses FormData, it will be restricted:

```ts
Record<string, string | File>
```

The second reason, it has the possibility of opening Security Vulnerability because we receive unexpected values. "*Plack*" rejected the PR for the same reason: <https://github.com/plack/Plack/pull/678>.

Consider the following code referred from the PR comment:

```ts
app.post('/foo', async (c) => {
  const body = await c.req.parseBody()
  const id = body.itemId
  const item = shopQuery(id)
  // ~> ['SELECT * FROM shop_item WHERE item_id = ?', id]
  // ...
```

In this case, `body.itemId` is expected `string` but, if receive Request that the body is JSON like below:

```ts
new Request(JSON.stringify({
  itemId: [1, 2, 3],
}), {
  headers: {
  },
})
```

`body.itemId` will be `[1, 2, 3]`. If `shopQuery` behaves differently when it receives a string than when it receives an array, this could be a problem.

```ts
// ~> ['SELECT * FROM shop_item WHERE item_id IN (?,?,?)', id ]
```

This would not happen if `shopQuery` does the type check on the input value, but it would be better to reduce the possibility of this happening. Fortunately, the Request object has a `req.json()`, `req.text()`,  and an `req.arrayBuffer()` method in addition to the `req.formData`. Users can explicitly body parse using those methods. So, `parseBody` should support only FormData. This fix includes breaking changes, but I think we should do it.

---

@metrue @usualoma 
I want to know your opinion.

yusukebe:
Make that `c.req.parseBody` parses only FormData and does not parse JSON, text, and arrayBuffer. This fix includes breaking changes.

Before this PR, this function returns the parsed body according to Content-Type, despite the type of body. And, it will be `any`:

```ts
parseBody<T = any>(r: Request | Response): Promise<T>
```

Even though the user was expecting a `Record<string, string>`, this may return `Object`. This behavior could potentially cause a security vulnerability. See #484 

In this PR, `c.req.parseBody` only parse `FormData`. If the user wants to parse others than FormData, then the user should explicitly use `json()`, `text()`, and `arrayBuffer()`.

```ts
const data = c.req.json()
```

`parseBody` will only return `Record<string, string | File>`.

```ts
parseBody(r: Request | Response): Promise<Record<string, string | File>>
```

Fix #484

### DIFF
--- a/deno_dist/request.ts
+++ b/deno_dist/request.ts
@@ -25,9 +25,9 @@ declare global {
       (name: string): string
       (): Cookie
     }
-    parsedBody?: Promise<any>
+    parsedBody?: Promise<Record<string, string | File>>
     parseBody: {
-      <T = any>(): Promise<T>
+      (): Promise<Record<string, string | File>>
     }
   }
 }
@@ -98,10 +98,10 @@ export function extendRequestPrototype() {
     }
   } as InstanceType<typeof Request>['cookie']
 
-  Request.prototype.parseBody = function <T = any>(this: Request): Promise<T> {
-    let body: Promise<T>
+  Request.prototype.parseBody = function (this: Request): Promise<Record<string, string | File>> {
+    let body: Promise<Record<string, string | File>>
     if (!this.parsedBody) {
-      body = parseBody<T>(this)
+      body = parseBody(this)
       this.parsedBody = body
     } else {
       body = this.parsedBody
--- a/deno_dist/utils/body.ts
+++ b/deno_dist/utils/body.ts
@@ -1,28 +1,16 @@
-export async function parseBody<T = any>(r: Request | Response): Promise<T> {
-  let body: any
-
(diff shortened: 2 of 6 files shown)


mliakos:
This should probably be reflected in the docs.

PR #489, opened 2022-08-28
fix(context): fix type error

yusukebe:
Fixed the type error below:

SS


koistya:
Yet another TypeScript error:

image

PR #490, opened 2022-08-29
fix app.HTTP_METHOD type

azukiazusa1:
Hi, Thank you for publishing such a good product!

This PR improve type of app.HTTP_METHOD.

If the return type of the handler function is wrong, as in the example below, an error is reported for "path".

スクリーンショット 2022-08-29 17 34 04

This is not ideal. Because it should be the handler function that should report the error like below.

スクリーンショット 2022-08-29 17 36 39

### DIFF
--- a/src/hono.ts
+++ b/src/hono.ts
@@ -53,17 +53,17 @@ interface HandlerInterface<
   E extends Partial<Environment> = Environment,
   U = Hono<E, T>
 > {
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @azukiazusa1 !

This is a good fix! It will be more user friendly than before. Thank you, I'll merge it.

PR #493, opened 2022-09-01
perf: do not `compose` if it has only one handler

yusukebe:
Made it does not `compose` if it has only one handler. `await composed` is slow and could be speed up by avoiding `compose`.

## Benchmark

Will be faster if it returns only "Hello World".

### handle-event

17%  Faster!!

```
hono - trie-router(default) x 421,632 ops/sec ±4.36% (79 runs sampled)
hono - regexp-router x 452,187 ops/sec ±5.90% (69 runs sampled)
hono - trie-router(default) - New x 506,280 ops/sec ±3.92% (85 runs sampled)
hono - regexp-router - New x 610,975 ops/sec ±4.52% (83 runs sampled)
```

### Deno - flash

Not much has changed.

```
v2.1.3
Requests/sec: 185881.3210
New
Requests/sec: 188679.7531
```

### Bun

11% Faster!!

```
v2.1.3
Requests/sec: 133396.6236
New
Requests/sec: 149045.7514
```

### DIFF
--- a/deno_dist/hono.ts
+++ b/deno_dist/hono.ts
@@ -199,9 +199,26 @@ export class Hono<
     const result = this.matchRoute(method, path)
     request.paramData = result?.params
 
-    const handlers = result ? result.handlers : [this.notFoundHandler]
-
     const c = new HonoContext<string, E>(request, env, eventOrExecutionCtx, this.notFoundHandler)
+
+    // Do not `compose` if it has only one handler
+    if (result && result.handlers.length === 1) {
+      const handler = result.handlers[0]
+      try {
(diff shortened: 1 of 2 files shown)


usualoma:
@yusukebe 

Awesome! But I think the logger is called twice in the following cases.

```ts
app.use('*', logger())
app.get('/', (c) => c.text('Hono!!'))
```

```
GET /not-found
```

yusukebe:
Hi @usualoma !

Thanks for pointing that out. It is indeed a bug. I have fixed it on this PR, can you review it?

#494

PR #494, opened 2022-09-01
fix: fixed the issue logger called twice

yusukebe:
By this PR #493 , the logger is called twice if it's not found. And the first log shows "200". The following case:

```ts
app.use('*', logger())
app.get('/', (c) => c.text('Hono!!'))
```

```
GET /not-found
```

SS

This problem is caused by two reasons.

1. It did not return 404 response immediately If the response is undefined. 
2. Default `c.res` is 200 response. It should be 404 response.

This PR fixed this problem.

This is the issue pointed out in the comments of this PR.
https://github.com/honojs/hono/pull/493#issuecomment-1234769263

### DIFF
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -1,4 +1,5 @@
 import type { Environment, NotFoundHandler, ContextVariableMap, Bindings } from './hono.ts'
+import { defaultNotFoundMessage } from './hono.ts'
 import type { CookieOptions } from './utils/cookie.ts'
 import { serialize } from './utils/cookie.ts'
 import type { StatusCode } from './utils/http-status.ts'
@@ -91,7 +92,7 @@ export class HonoContext<
   }
 
   get res(): Response {
-    return (this._res ||= new Response())
+    return (this._res ||= new Response(defaultNotFoundMessage, { status: 404 }))
   }
 
   set res(_res: Response) {
--- a/deno_dist/deno/serve-static.ts
+++ b/deno_dist/deno/serve-static.ts
@@ -13,7 +13,7 @@ const DEFAULT_DOCUMENT = 'index.html'
 export const serveStatic = (options: ServeStaticOptions = { root: '' }) => {
   return async (c: Context, next: Next): Promise<Response | undefined> => {
     // Do nothing if Response is already set
-    if (c.res && c.finalized) {
+    if (c.finalized) {
       await next()
     }
 
--- a/deno_dist/hono.ts
+++ b/deno_dist/hono.ts
@@ -35,6 +35,8 @@ export type ErrorHandler<E extends Partial<Environment> = Environment> = (
 
 export type Next = () => Promise<void>
 
+export const defaultNotFoundMessage = '404 Not Found'
+
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 type ParamKeyName<NameWithPattern> = NameWithPattern extends `${infer Name}{${infer _Pattern}`
   ? Name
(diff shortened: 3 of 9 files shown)


usualoma:
@yusukebe Thanks!

I think I also need to apply errorHandler, but can we do without it?
```diff
diff --git a/src/hono.ts b/src/hono.ts
index d1ba641..e804644 100644
--- a/src/hono.ts
+++ b/src/hono.ts
@@ -217,8 +217,12 @@ export class Hono<
           }
         }
         return this.notFoundHandler(c as Context)
-        // Do nothing
-      } catch {}
+      } catch (err) {
+        if (err instanceof Error) {
+          return this.errorHandler(err, c as Context<string, Environment>)
+        }
+        throw err
+      }
     }
 
     const handlers = result ? result.handlers : [this.notFoundHandler]
```

PS

I understand that this PR approach is the fastest. What worries me a little is that it seems to sacrifice maintainability for the sake of optimizing to a benchmark.

I am concerned that the inclusion of this code will reduce the motivation to "optimize compose.ts" and slow it down in the real world.

However, I also understand the importance of the "fastest framework on the benchmark" identity. (It's only a benchmark, but it's a benchmark)

If we are going to go further with the "optimization by removing await", I would suggest the following in addition to what is in this PR. This branch also performs quite well (though not as well as this PR). This change would also optimize performance in cases where middleware is used.
https://github.com/honojs/hono/compare/main...usualoma:hono:perf-remove-await-composed

yusukebe:
Hi @usualoma !

> I think I also need to apply errorHandler, but can we do without it?

This is what I have been struggling with, and I have been skipping the steps.

It does not catch the error here:

https://github.com/honojs/hono/blob/63c7e5ada1ee5a463b9ee0315217dd3fde934c4e/src/hono.ts#L221

but the error will be caught below:

https://github.com/honojs/hono/blob/63c7e5ada1ee5a463b9ee0315217dd3fde934c4e/src/hono.ts#L235

So I decided that it is not necessary. I know this is not a good approach because the handler which occurs the error will be executed twice. However, to keep the code short, I wrote this.

Hmmm. Do you think it is bad?

---

> sacrifice maintainability for the sake of optimizing to a benchmark

This is a very difficult matter.

We have to keep the code simple to improve for evolution. But more than that, I'm still thinking the identity of Hono is to be fast and must be at the top of the benchmark. For example, Hono is used in official benchmarks on Deno, and it is expected to avoid `await` and speed up the process on Bun. Both maintainability and optimization are necessary, but in this case I would choose optimization.

> If we are going to go further with the "optimization by removing await", I would suggest the following in addition to what is in this PR.
> This change would also optimize performance in cases where middleware is used.

This is good!! I have also been concerned about the performance of middleware.

If this PR is merged, will you create a new PR for `perf-remove-await-composed` ?

usualoma:
@yusukebe Thanks for the answer.

Understood! I accept that choice favorably.

> We have to keep the code simple to improve for evolution. But more than that, I'm still thinking the identity of Hono is to be fast and must be at the top of the benchmark. For example, Hono is used in official benchmarks on Deno, and it is expected to avoid await and speed up the process on Bun. Both maintainability and optimization are necessary, but in this case I would choose optimization.

I think it should be avoided that the handler is executed twice.
(Although most of the time it is not a problem if it is executed twice because of an error.) I think that unexpected problems can occur when a POST or PUT request is executed twice.

I will create a PR for perf-remove-await-composed later!

usualoma (approved):
👍

PR #495, opened 2022-09-02
perf(compose): Remove `await composed()` from hono.ts.

usualoma:
### What is included?

* Reduced use of `await`
* Reduced use of `context instanceof HonoContext`
    * I think this code did not limit the type of `context` for abstraction, but since it will always be `HonoContext` in production environment, we have limited the type of `context` so that we do not have to compare.

There are many changes to the test code, but no changes to Hono's API.

### Benchmark

These are the results in my environment.
#493 reflects the results of optimization.
```
hono - trie-router(default) x 220,343 ops/sec ±5.68% (83 runs sampled)
hono - regexp-router x 261,953 ops/sec ±5.37% (82 runs sampled)
itty-router x 82,331 ops/sec ±3.38% (85 runs sampled)
sunder x 114,834 ops/sec ±2.25% (83 runs sampled)
worktop x 66,525 ops/sec ±3.41% (87 runs sampled)
Fastest is hono - regexp-router
```

Apply the following changes to this PR.

```diff
diff --git a/src/hono.ts b/src/hono.ts
index c4c0810..4a2da5e 100644
--- a/src/hono.ts
+++ b/src/hono.ts
@@ -203,24 +203,6 @@ export class Hono<
 
     const c = new HonoContext<string, E>(request, env, eventOrExecutionCtx, this.notFoundHandler)
 
-    // Do not `compose` if it has only one handler
-    if (result && result.handlers.length === 1) {
-      const handler = result.handlers[0]
-      try {
-        const res = handler(c, async () => {})
-        if (res) {
-          const awaited = res instanceof Promise ? await res : res
-          if (awaited) return awaited
-        }
-        return this.notFoundHandler(c as Context)
-      } catch (err) {
-        if (err instanceof Error) {
-          return this.errorHandler(err, c as Context)
-        }
-        throw err
-      }
-    }
-
     const handlers = result ? result.handlers : [this.notFoundHandler]
 
     const composed = compose<HonoContext<string, E>>(handlers, this.notFoundHandler)
```

This version is a bit slower, but this is doing quite well, too.

```
hono - trie-router(default) x 218,861 ops/sec ±6.06% (79 runs sampled)
hono - regexp-router x 265,468 ops/sec ±6.16% (79 runs sampled)
itty-router x 83,826 ops/sec ±3.77% (86 runs sampled)
sunder x 120,162 ops/sec ±2.26% (85 runs sampled)
worktop x 66,387 ops/sec ±5.05% (84 runs sampled)
Fastest is hono - regexp-router
```

### DIFF
--- a/deno_dist/compose.ts
+++ b/deno_dist/compose.ts
@@ -1,14 +1,22 @@
 import { HonoContext } from './context.ts'
 import type { NotFoundHandler } from './hono.ts'
 
+interface ComposeContext {
+  finalized: boolean
+  res: any
+}
+
 // Based on the code in the MIT licensed `koa-compose` package.
-export const compose = <C>(middleware: Function[], onNotFound?: NotFoundHandler) => {
+export const compose = <C extends ComposeContext>(
+  middleware: Function[],
+  onNotFound?: NotFoundHandler
+) => {
   const middlewareLength = middleware.length
   return (context: C, next?: Function) => {
     let index = -1
     return dispatch(0)
 
-    async function dispatch(i: number): Promise<C> {
+    function dispatch(i: number): C | Promise<C> {
       if (i <= index) {
         throw new Error('next() called multiple times')
       }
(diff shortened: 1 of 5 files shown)


yusukebe:
Hi @usualoma !

Thank you for the PR! I've taken the benchmark for the application just returning "Hello World" on Bun.

SS

The code with no #493 applied (the code you used in the second benchmark) will be much faster than v2.1.3! This will speed up handlers with some middleware. Nice! I'll merge it.

yusukebe:
Hi @usualoma !

Sorry to get back on topic. Once again, I benchmarked on Bun. This PR will be effective when multiple middleware is applied to the handler. So, I wrote the following script:


```ts
import { Hono } from '../../src/index'

  app.use('/with-middleware', async (c, next) => {
    c.set('count', i)
    await next()
    c.get('foo')
  })
}
app.get('/with-middleware', (c) => c.text('GET(/with-middleware)'))
export default app
```

Then, run the scripts on Bun 0.1.11 latest canary build, with each versions:

1. Current `main` branch
2. [fix/logger-called-twice branch](https://github.com/honojs/hono/tree/fix/logger-called-twice/src), not applied this PR yet.
3. v2.1.3

Results are below:

SS

SS

Current main branch is slower than the others. Maybe the changes in this PR are having a worse effect.

And, I've notice that all handlers except main handler are `async`, because they are almost middlware. So they have to `await`, and perhaps what this PR is doing is not much different from the previous version. 

What do you think about this? Sorry if I am wrong.

usualoma:
@yusukebe Thanks!

Your point is spot on.

I did not correctly consider in this PR that "middleware is an `AsyncFunction` and `await` is used".

I think the following PR would solve this problem.
https://github.com/honojs/hono/pull/496

PR #499, opened 2022-09-05
Fix regexp ambigous route

### LINKED ISSUE #497: [Bug] returns 404 using RegExpRouter
The following code using RegExpRouter, `POST /posts` will return "404 Not Found" response, the Hono version is `v2.1.3`. And it wIll be "Internal Server Error" In the current tree https://github.com/honojs/hono/tree/e174b7085821164877e46b62cc3f56dcf81b4b77.

```ts
import { Hono } from 'hono'

app.get('/posts', (c) => c.text('foo'))
app.post('/posts', (c) => c.text('foo'))
app.get('/posts/page/:page', (c) => c.text('foo'))
app.get('/posts/:id/comments', (c) => c.text('foo'))
```

```
$ curl -X POST http://localhost:8787/posts
404 Not Found%
```

The same problem does not occur with TrieRouter.

usualoma:
Fixes #497

The performance of this code is not that good, but it is a rare case and has little impact. Performance improvements will be considered later.

### DIFF
--- a/deno_dist/router/reg-exp-router/router.ts
+++ b/deno_dist/router/reg-exp-router/router.ts
@@ -327,7 +327,10 @@ export class RegExpRouter<T> implements Router<T> {
     Record<string, AnyMatcher<T>[]>,
     boolean
   ] {
-    // @ts-ignore
+    if (!this.routeData) {
+      throw new Error('Can not call the buildAllMatchers since the matcher is already built.')
+    }
+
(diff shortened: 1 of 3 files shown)


yusukebe:
@usualoma Thank you!

Could you do `denoify`?

usualoma:
@yusukebe Fixed! c9ad185

yusukebe:
Looks Good! I'll merge it!

PR #501, opened 2022-09-06
Introduce StaticRouter and SmartRouter

usualoma:
### Motivation

I think it would be difficult for users to take advantage of the "hono has two routers and the optional RegExpRouter is faster."
This PR changes this to "By default, the best router is selected based on the routing of the application."

### StaticRouter

StaticRouter is a new alternative. Works fast when there is no path to receive parameters.
I think that this is not a router for benchmarking purposes, and that it could be used in the real world as follows

```ts
app.use('*', logger(), cors())
app.use(
  '/graphql',
  graphqlServer({
    schema,
    rootResolver,
  })
)
```

However, when benchmarked, there did not appear to be a significant difference with RegExpRouter in nodejs.
If it is a bun, there may be a little difference, but there may not be.

I think this StaticRouter is a good example of the use of SmartRouter, but may not be necessary for hono.

### SmartRouter

This is a router that finds the best one for routing from several possible routers and invoke it when `match` is called.

### Side effects of merging this

* All routers are bundled by default, which increases the bundle size.
* Slightly longer spinup time.
* Makes debugging routers more difficult.

#### with path parameters

https://gist.github.com/usualoma/6d8db87c9f40614073b0517dc3e8cee6

```
hono - smart x 267,322 ops/sec ±5.61% (77 runs sampled)
hono - smart - trie-router x 230,474 ops/sec ±5.72% (81 runs sampled)
hono - smart - regexp-router x 267,080 ops/sec ±5.65% (68 runs sampled)
hono - trie-router x 223,486 ops/sec ±5.05% (84 runs sampled)
hono - regexp-router x 262,344 ops/sec ±6.61% (77 runs sampled)
Fastest is hono - smart,hono - smart - regexp-router,hono - regexp-router
```

#### without path parameters

https://gist.github.com/usualoma/0b86e02ee9f3a16dd8e5af77f479624c

```
hono - smart - trie-router x 276,017 ops/sec ±6.15% (77 runs sampled)
hono - smart - regexp-router x 298,234 ops/sec ±6.67% (75 runs sampled)
hono - smart - static-router x 296,272 ops/sec ±6.79% (63 runs sampled)
hono - trie-router x 267,239 ops/sec ±6.06% (81 runs sampled)
hono - regexp-router x 303,253 ops/sec ±5.80% (82 runs sampled)
Fastest is hono - regexp-router,hono - smart - regexp-router,hono - smart - static-router
```

#### bun - RegExpRouter

```ts
import { Hono } from './src/index'
import { RegExpRouter } from './src/router/reg-exp-router'

app.get('/posts', (c) => c.text('foo'))

export default app
```

```
$ ./bombardier -c 200 -d 20s http://10.1.0.139:3000/posts
Bombarding http://10.1.0.139:3000/posts for 20s using 200 connection(s)
[==========================================================================] 20s
Done!
Statistics        Avg      Stdev        Max
  Reqs/sec     72705.89    1742.91   75809.44
  Latency        2.75ms    86.69us    10.99ms
  HTTP codes:
    1xx - 0, 2xx - 1454123, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:    10.40MB/s
```

#### bun - SmartRouter - StaticRouter

```ts
import { Hono } from './src/index'
import { SmartRouter } from './src/router/smart-router'
import { StaticRouter } from './src/router/static-router'

const app = new Hono({
  router:
    new SmartRouter({
      routers: [new StaticRouter()],
    })
})

app.get('/posts', (c) => c.text('foo'))

export default app
```

```
$ ./bombardier -c 200 -d 20s http://10.1.0.139:3000/posts
Bombarding http://10.1.0.139:3000/posts for 20s using 200 connection(s)
[==========================================================================] 20s
Done!
Statistics        Avg      Stdev        Max
  Reqs/sec     74691.57    1911.30   80312.78
  Latency        2.68ms    97.25us    10.77ms
  HTTP codes:
    1xx - 0, 2xx - 1493610, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:    10.68MB/s
```

### DIFF
--- a/deno_dist/hono.ts
+++ b/deno_dist/hono.ts
@@ -3,8 +3,12 @@ import type { Context } from './context.ts'
 import { HonoContext } from './context.ts'
 import { extendRequestPrototype } from './request.ts'
 import type { Router } from './router.ts'
+import { METHODS } from './router.ts'
 import { METHOD_NAME_ALL, METHOD_NAME_ALL_LOWERCASE } from './router.ts'
-import { TrieRouter } from './router/trie-router/index.ts' // Default Router
+import { RegExpRouter } from './router/reg-exp-router/index.ts'
+import { SmartRouter } from './router/smart-router/index.ts'
+import { StaticRouter } from './router/static-router/index.ts'
+import { TrieRouter } from './router/trie-router/index.ts'
 import { getPathFromURL, mergePath } from './utils/url.ts'
 
 export interface ContextVariableMap {}
@@ -68,8 +72,7 @@ interface HandlerInterface<
   (path: string, ...handlers: Handler<string, E>[]): U
 }
 
-const methods = ['get', 'post', 'put', 'delete', 'head', 'options', 'patch'] as const
-type Methods = typeof methods[number] | typeof METHOD_NAME_ALL_LOWERCASE
+type Methods = typeof METHODS[number] | typeof METHOD_NAME_ALL_LOWERCASE
 
 function defineDynamicClass(): {
   new <E extends Partial<Environment> = Environment, T extends string = string, U = Hono>(): {
@@ -89,7 +92,9 @@ export class Hono<
   E extends Partial<Environment> = Environment,
   P extends string = '/'
 > extends defineDynamicClass()<E, P, Hono<E, P>> {
-  readonly router: Router<Handler<string, E>> = new TrieRouter()
+  readonly router: Router<Handler<string, E>> = new SmartRouter({
+    routers: [new StaticRouter(), new RegExpRouter({ allowAmbiguous: false }), new TrieRouter()],
+  })
   readonly strict: boolean = true // strict routing - default is true
   private _tempPath: string = ''
   private path: string = '/'
@@ -101,7 +106,7 @@ export class Hono<
 
     extendRequestPrototype()
 
-    const allMethods = [...methods, METHOD_NAME_ALL_LOWERCASE]
+    const allMethods = [...METHODS, METHOD_NAME_ALL_LOWERCASE]
     allMethods.map((method) => {
       this[method] = <Path extends string = ''>(
         args1: Path | Handler<ParamKeys<Path>, E>,
--- a/deno_dist/router.ts
+++ b/deno_dist/router.ts
@@ -1,5 +1,6 @@
 export const METHOD_NAME_ALL = 'ALL' as const
 export const METHOD_NAME_ALL_LOWERCASE = 'all' as const
+export const METHODS = ['get', 'post', 'put', 'delete', 'head', 'options', 'patch'] as const
 
 export interface Router<T> {
   add(method: string, path: string, handler: T): void
@@ -10,3 +11,5 @@ export interface Result<T> {
   handlers: T[]
   params: Record<string, string>
 }
+
+export class UnsupportedPathError extends Error {}
--- a/deno_dist/router/reg-exp-router/router.ts
+++ b/deno_dist/router/reg-exp-router/router.ts
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/ban-ts-comment */
 import type { Router, Result } from '../../router.ts'
-import { METHOD_NAME_ALL } from '../../router.ts'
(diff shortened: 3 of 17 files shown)


yusukebe:
Wow, crazy!

I'm going to look at the implementation and consider it well.

> All routers are bundled by default, which increases the bundle size.

This is what I am concerned about.

Again!

usualoma:
@yusukebe Thanks for the comment.

I have force pushed with minor fixes, but the essence has not changed since the initial push.

usualoma:
There are no more commits to be added.

I think it's an interesting and useful idea, but I'm not sure I really want to merge this.

yusukebe:
Hi @usualoma !

This idea is ultra great! I like it! But as you are thinking, we have to consider well whether to merge or not. So, I don't make conclusions right away.

I investigated "bundle size". `wrangler` command will bundle the scripts and shows the "upload size". Below are the sizes of "Hello World" program on each framework for Cloudflare Workers.

```
Hono v2.1.4
Total Upload: 27.04 KiB / gzip: 6.09 KiB

This PR / with SmartRouter
Total Upload: 50.35 KiB / gzip: 10.32 KiB

Sunder
Total Upload: 47.21 KiB / gzip: 11.08 KiB

Worktop
Total Upload: 7.02 KiB / gzip: 2.70 KiB
```

This is just "information", and I would still like to think about it.

yusukebe:
In actuality, there is a way to reduce the bundle size.

Currently, we build the project as "CommonJS". If we make it ESM, the size will be smaller. I've tried to change the compiler option on `tscofing.json` and build it as ESM. Then, the bundle size built by `wrangler` has been smaller!

```
Total Upload: 36.10 KiB / gzip: 8.64 KiB
```

I will create another Issue on the matter of migrating from CommonJS to ESM. But, it's good news.

usualoma:
@yusukebe Good news!

```
Hono this PR / with SmartRouter
Total Upload: 36.10 KiB / gzip: 8.64 KiB
```

Hmmm, but RegExpRouter is so huge.

yusukebe:
Hi @usualoma !

I want to merge this Static Router and Smart Router! I love this feature.

> I think it would be difficult for users to take advantage of the "hono has two routers and the optional RegExpRouter is faster." This PR changes this to "By default, the best router is selected based on the routing of the application."

You are right about this. Users have to use the fastest router without having to do anything else.

I love it because this is **really "smart"**. It will be more difficult to debug, but it will be more "fun" with this way.

I think the issue of the bundle size could be improved a little by migrating ComonJS to ESM. I don't know if it will work, but I would like to try it.

How about we go with this!

usualoma:
@yusukebe 
Sorry for the delay in answering.

Thanks! I agree with merging.

I've been thinking the past few days that if we assume SmartRouter, we could create a lightweight version of RegExpRouter with some features removed from RegExpRouter. (I haven't gotten around to starting it, though.)

I think we can represent RegExpRouter with less code by deciding not to support it in the lightweight version of RegExpRouter, since there are rarely routings that have different variable names in the same place, such as the following.

```
app.get('/path/:var1/path')
app.get('/path/:var2/path')
```

I would like to try such a refactoring along with migrating ComonJS to ESM.

Shall we fix this branch so that the test is green and merge it into main?

yusukebe:
Hi @usualoma !

> I think we can represent RegExpRouter with less code by deciding not to support it in the lightweight version of RegExpRouter, since there are rarely routings that have different variable names in the same place, such as the following.

This is a great idea! Just fall back to TrieRouter If it hits such rare cases.

> Shall we fix this branch so that the test is green and merge it into main?

I've created the `develop` branch <https://github.com/honojs/hono/tree/develop>

Let's work on the `develop`" branch for the new features such as migration to SmartRouter and ESM, will be included in v2.2.0 new version.  So, fix this branch and merge it into the `develop` branch.

yusukebe:
@usualoma 

Ah, we don't really need to use the "develop" branch anyway, so let's merge it into the "main" branch.

usualoma:
OK!
Shall we proceed with 0b0e113?

yusukebe:
Yes, merged!

PR #505, opened 2022-09-08
feat(middleware): introduce "built-in" Validator Middleware

yusukebe:
This PR introduces "built-in" validator middleware.

There is third-party validator middleware `@honojs/validator`. I like the API and think it is well thought out. And the middleware has many validation patterns, such as email addresses, IP addresses, etc. Instead, it's too fat because of using "validator.js" and  "jsonpath-plus".

However, we want to have validator middleware; validation is often required when building web applications.

This middleware is the port of `@honojs/validator`, making it more simple, lightweight, and with zero dependencies. By reducing the rules for validation. Usage is compatible with `@honojs/validator`:

```ts
import { Hono } from 'hono'
import { validation } from 'hono/validator'

const app = new Hono()

app.post(
  '/post',
  validation((v, message) => ({
    body: {
      // Validation rules
      title: [v.required, [v.isLength, { max: 100 }]],
      // Custom error message
      body: [v.required, message('Body is wrong!!')],
      tags: v.optional,
    },
    additionalProperties: false,
  }))
)

app.post('/post', async (c) => {
  return c.text('Valid')
})
```

And, this bundle size built by Wrangler is very smaller than third-party.

```
@honojs/validator:
Total Upload: 269.52 KiB / gzip: 52.51 KiB

hono/validator:
Total Upload: 37.27 KiB / gzip: 7.96 KiB
```

The use case is that the user uses this built-in middleware first and will use the third-party if the user wants to validate variable patterns.

I think it's good this validator is distributed as built-in.


usualoma:
@yusukebe 

I think this is a good API!

I have an idea about validator.
You may have a different opinion, but I would like you to take a look at the idea once you do.
I will summarize later.

usualoma:
## I would consider the following.

### Validated and exactly the same value should be returned

Since 0, null, and undefined can have other meanings, especially in SQL, it should be avoided that their values are validated as empty strings and returned as undefined in c.req.json().

### I think it would be better to make the additional properties not error-proof but not easily retrievable by the user

I would prefer that additional properties not be obtained by default.
However, it would be inconvenient for `query` or `header` to error out additional properties.

## An idea

So I thought it would be better to use something like the following PR: "perform validation and return the result".
https://github.com/honojs/hono/pull/508

Since it overrides the `c.req.query` result, I am aware that this is a rather aggressive process and may be controversial.
However, since it is to be built in as middleware, I think that this much is acceptable as long as it is safe with a concise API.

yusukebe:
Hi @usualoma !

Thank you for sharing and writing your idea! I think that's good!!

However, let me think some more about whether to merge or not. I'll consider whether this is helpful or not from the user's point of view.

ThatOneBro:
Hey @yusukebe ! I got busy with a new API and so I haven't been able to work much on documentation yet but I will try again this weekend! 

I did see this PR and I have some thoughts about the validation middleware. I am actually using it heavily for the API I am working on and had to make some modifications to validate arrays of objects. 

So we have this syntax with json-path: 
```js
{
  "patients[*].firstName": [v.trim, v.required],
  "patients[*].lastName": [v.trim, v.required],
  ...
}
```

But to actually use this for useful validation, you want to return the index of invalid objects along with the fields:
Input:
```json
{ 
  "patients": [
    {
      "firstName": "Bob",
      "lastName": "Doe"
    },
    {
      "firstName": "Alice",
    }
  ]
}

```
Result:
```
patients[2].lastName is invalid -- required
```

In order to do this, I temporarily made a quick patch... This weekend I will try to make a PR if this is something we would like to add to the main middleware. This is what the main part looks like:

```diff
@@ -117,10 +121,23 @@ const validatorMiddleware = (validatorFunction) => {
             const req = c.req.clone();
             const json = (await req.json());
             Object.keys(field).map(async (key) => {
-                const data = (0, jsonpath_plus_1.JSONPath)({ path: key, json });
-                const value = `${data[0]}`; // Force converting to string
-                const message = (name) => `Invalid Value: the JSON body "${key}" is invalid - ${name}`;
-                validate(field[key], value, message);
+                if (key.includes("[*]")) {
+                    // TODO: Make this recursive?
+                    const [arrPath, keyInEle] = key.split("[*].");
+                    const arrElements = (0, jsonpath_plus_1.JSONPath)({ path: arrPath, json })[0];
+                    const keyElements = arrElements.map(ele => ele[keyInEle]);
+                    for (const [i, val] of keyElements.entries()) {
+                        const value = val ? `${val}` : undefined;
+                        const specificKey = key.replace(/\*/, i);
+                        const message = (name) => `Invalid Value: the JSON key "${specificKey}" with value "${value}" is invalid - ${name}`;
+                        validate(field[key], specificKey, value, message);
+                    }
+                } else {
+                    const data = (0, jsonpath_plus_1.JSONPath)({ path: key, json }); 
+                    const value = data[0] ? `${data[0]}` : undefined;
+                    const message = (name) => `Invalid Value: the JSON key "${key}" with value "${value}" is invalid - ${name}`;
+                    validate(field[key], key, value, message); 
+                }
             });
         }
         if (v.done) {
 ```
 
 Basically I split parts of the subpath to treat each member of the array as a separate object to validate. This could be applied recursively too probably, but I didn't get that far to think of the edge cases. But I think this kind of validation will be common in APIs dealing with CRUD operations, so it will probably be important. I do like this syntax though, because it still creates a flat validation object as an API.

**Comment by yusukebe** [comment]
Hi @usualoma @ThatOneBro !

I have considered this matter and have concluded.

## Summary

* Accept the idea "*Pass only validated properties*" by @usualoma #508.
* Add the option `additionalProperties`, default is `true`. If the user does not set this option explicitly, "Pass all properties".
* Make `@honojs/validator` **obsolete**. If the user wants to use more rules, import the function from "validator.js".
* Add some rules which are not been implemented yet. Such `duplication`, `in`.

## Pass only validated properties

I agree that you @usualoma said it is not a practical way to throw errors for all the additional properties of "headers" and "query". And if the target is JSON, it is difficult to decide which properties should be considered as additional properties. Implementing this feature will avoid these concerns. Then, above all, it is secure.

But I think it's better that it does not work if the `additionalProperties` option is not explicitly set to `false`.

```ts
app.post(
  '/post',
  validation((v) => ({
    json: {
      title: v.required,
      body: [v.isLength, { max: 400 }],
      'author.name': v.required,
    },
    additionalProperties: false, // default is `true`
  })),
```

Otherwise, it is difficult for users to understand "why missing some properties" when doing `c.req.json`. And @ThatOneBro  is right, it would be better for users who are used to 3rd party validators. If they does not want "pass the additional properties", just set the property as `false`.

## About `@honojs/validator`

At first, I thought this middleware and "@honojs/validator" are two different products. But, I've noticed that the user can add the rule, just import the function from `validator.js`.

```ts
import isEmail from 'validator/lib/isEmail'
import { Hono } from 'hono'
import { validation } from 'hono/validator'

//...

app.post(
  '/post',
  validation((v) => ({
    json: {
      title: v.required,
      body: [v.isLength, { max: 400 }],
      'author.name': v.required,
      'author.email': isEmail,
```

Since there is no need for two middleware to exist, I think it is better to make @honojs/validator obsolete. This is big change, but we have to do it for the future.

## Remaining things

I want to add some rules such `duplication` or `in`, it's not implemented yet. In this PR or another PR after merged, will be implemented.

## Shipping

Include this feature in v2.2.0, which will be shipped soon.

---

If there are no big different opinions, we would like to do this.

ThatOneBro:
I think this makes sense, and of course greatly simplifies the middleware.

There is one use case I am concerned about though, as it is one I am working with now.

Will the custom json path support array notation too? I think this could be a common use case for bulk routes where users create an array of resources.

For example, using this:
```js
{
  json: {
    "patients[*].firstName": [v.required],
    "patients[*].lastName": [v.required],
    ...
  },
}
```
to validate an array of patient resources.

Do you think this support is worth the complexity? I can help to add this feature if needed, just wanted to know your opinion first.

usualoma:
@yusukebe 

Sorry for the late response here as well.
I have a different view on several points.


### "why missing some properties"  issue

> Otherwise, it is difficult for users to understand "why missing some properties" when doing c.req.json. 

OK, I understand this very well. I may waste a day of my time trying to solve "why missing some properties". I think this is very bad for DX.


### Default value of additionalProperties

I believe that in the #508 implementation, `additionalProperties` makes sense only because it is `false`.
I'm wondering if the default for `additionalProperties` is `true`, we can leave the #505 implementation as is. I think that "raise an error if there are additional properties" is not a bad idea, but rather preferable, if the user opts in to enable the `additionalProperties` check.

### 0 / null / undefined problem

Since 0, `null`, and `undefined` are converted to empty strings, they will be validated as empty strings, but you will get 0, `null`, and `undefined` when you retrieve the value.
https://github.com/honojs/hono/pull/505/files#diff-531c2433675cbaf70e386c1bd153061aab1e15575134abd5d65b79882f9c1424R72

I think this is the only problem with #505, so if this can be fixed, we can do without #508 and leave the #505 implementation as it is.

yusukebe:
Hi @usualoma !

Thank you for the response!

I think we have another thing to consider, that is about validating JSON.  It's related to the implementation.

In this #505 PR, treating only values in the same hierarchy are "additional properties".

https://github.com/honojs/hono/blob/b7195ff4ea82d025415cd13ef21e5f4a195e2d50/src/middleware/validator/middleware.ts#L206-L219

For example, the following test will fail:

```ts
app.post(
  '/json',
  validation((v) => ({
    json: {
      'post.author.name': v.required,
    },
    additionalProperties: false,
  })),
  (c) => {
    return c.text('Valid')
  }
)

it('Should return 400 response - JSON', async () => {
  const json = {
    post: {
      title: 'Hello', // this property is not treat as "additional property"
      author: {
        name: 'abcdef',
      },
    },
  }
  const req = new Request('http://localhost/json', {
    method: 'POST',
    body: JSON.stringify(json),
  })
  expect(res.status).toBe(400) // will be 200
})
```

If we want to check for `post.title`, I think we have to check the whole JSON structure recursively. It's a lot of work.
// Or do you know of a good implementation way?

So, "checking additional properties is difficult".

In comparison, with the #508 implementation, we only need to pass the validated values to  `c.req.query` or `c.req.json`, so we don't have to check for the additional properties to exist. Even though we have to change request.ts, #508 is simpler to implement. So I like #508. 

However, I think you are also right. So how about the following plan?

* Going with #508 
* If `removeAddtional` option is `false`, pass through all properties.
* Default value of `removeAdditional` is `true`.

I use `removeAdditional` instead of `additonalProperty`, I think it is easier to understand.

What do you think about this plan?

usualoma:
@yusukebe 
Okay, I understand more about the background of the issue.
I vote for this plan!

> * Going with [Feat/builtin validator middleware another idea #508](https://github.com/honojs/hono/pull/508)
> * If `removeAddtional` option is `false`, pass through all properties.
> * Default value of `removeAdditional` is `true`.

yusukebe:
@usualoma 

Thank you!! Is there anything left to do on #508 ? If not, I'll merge it into this `feat/builtin-validator-middleware` branch.


@ThatOneBro 

Could you create the issue about "*json path supporting array notation*" later?

usualoma:
@yusukebe 

Thank you!
I have no plans to add anything to #508.

ThatOneBro:
@yusukebe 
> Could you create the issue about "_json path supporting array notation_" later?
Yes of course, it should be a separate issue. I'll make it now.

PR #507, opened 2022-09-08
Feat: data passing in context

Skye-31:
This allows users to type and create their own data, which is passed through middleware. This can be useful for things like https://npmjs.com/package/d1-orm, where the library is initialised in the first middleware, and then can't be used as easily in routes past that without doing something like editing the Env object.

### DIFF
--- a/src/context.test.ts
+++ b/src/context.test.ts
@@ -143,6 +143,11 @@ describe('Context', () => {
     expect(res.headers.get('foo')).not.toBeNull()
     expect(res.headers.get('foo')).toBe('bar')
   })
+
+  it('should support data passing', async () => {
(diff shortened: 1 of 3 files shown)


yusukebe:
Hi @Skye-31 ! Thank you for the PR!

The current Hono has features similar to those of your PR, you can use `c.set`/`c.get`:

```ts
// For adding types
type Variables = {
  message: string
}

const app = new Hono<{ Variables: Variables }>()

  c.set('message', 'Hono is cool!!')
  await next()
})

app.get('/', (c) => {
  const message = c.get('message')
  return c.text(`The message is "${message}"`)
})
```

I think we have to write the document more, as this is not well documented.

Thank you!

Skye-31:
Ah, I thought this was for headers, thanks for clarifying!

Issue #503, opened 2022-09-07
[Proposal] Allow cors multiple origins


Tc-001:
This has been one of the pain points of cors, because without special logic it is usually impossible to allow multiple origins. 

My proposal is to allow:
 ```js
app.use(
  '*',
  cors({
    origin: ['https://example.com', 'https://example.org'],
    //...
  })
)
```

And for the server to respond with the correct entry based on the [`Referrer`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referer) header. (Falling back to `origin[0]` if empty/none apply)

yusukebe:
Hi @Tc-001 !

Thank you for your proposal! I think this is a good idea and we are able to implement it as the specification you tell us.

I'll try to make it later.

PR #532, opened 2022-09-15
Check in Origin header instead of Referer

usualoma:
I think we should check by the Origin header, not the Referer header.


yusukebe:
You are right! Thank you.

This PR really Close #503

PR #533, opened 2022-09-15
feat(cors): Enable to check origin header by a function.

usualoma:
I would also like to be able to dynamically check Origin by a function.


yusukebe:
Hi @usualoma !

Good feature! I'll merge it. Thank you!

PR #535, opened 2022-09-18
feat: another idea of Validator Middleware

yusukebe:
I like the current validator middleware API:

```ts
app.post(
  '/json',
  validation((v) => ({
    json: {
      'post.author.name': v.isAlpha,
    },
  })),
  (c) => {
    return c.text('Valid')
  }
)
```

But there are times when I want "Types":

```ts
  async (c) => {
    // How can I access `post.author.name` ?
    return c.text('Valid')
  }
```

So I implemented the new idea of Validator Middleware that support "Types". This is the approach of "declaring the keys of properties first".

```ts
app.post(
  '/posts',
  validator((v) => ({
    id: v.json('post.id').asNumber().isRequired(),
    title: v.json('post.title').isRequired().isLength(400),
    body: v.json('post.body').isOptional(),
  })),
  (c) => {
    const data = c.req.valid()
    return c.text(`ID is ${data.id}, Title is ${data.title}`)
  }
)
```

SS

Wow, we got "Types"!

There are other good points to this approach as well.

* Easy to understand the result data because "declaring property name first".
* Easy to writing rules. `v.isLength(400)` is better than `[v.isLength, 400]` .
* Getting only "declared properties". Additional properties are always ignored. It's safe.

I liked the current Validator Middleware, but I like this one even better.

One thing that bothers me is the implementation of `hono.ts` has become more complex.

How do you think about this?

### DIFF
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -1,4 +1,10 @@
-import type { Environment, NotFoundHandler, ContextVariableMap, Bindings } from './hono.ts'
+import type {
+  Environment,
+  NotFoundHandler,
+  ContextVariableMap,
+  Bindings,
+  ValidatedData,
+} from './hono.ts'
 import { defaultNotFoundMessage } from './hono.ts'
 import type { CookieOptions } from './utils/cookie.ts'
 import { serialize } from './utils/cookie.ts'
@@ -10,9 +16,10 @@ export type Data = string | ArrayBuffer | ReadableStream
 
 export interface Context<
   RequestParamKeyType extends string = string,
-  E extends Partial<Environment> = any
+  E extends Partial<Environment> = any,
+  D extends ValidatedData = ValidatedData
 > {
-  req: Request<RequestParamKeyType>
+  req: Request<RequestParamKeyType, D>
   env: E['Bindings']
   event: FetchEvent
   executionCtx: ExecutionContext
@@ -45,10 +52,11 @@ export interface Context<
 
 export class HonoContext<
   RequestParamKeyType extends string = string,
-  E extends Partial<Environment> = Environment
-> implements Context<RequestParamKeyType, E>
+  E extends Partial<Environment> = Environment,
+  D extends ValidatedData = ValidatedData
+> implements Context<RequestParamKeyType, E, D>
 {
-  req: Request<RequestParamKeyType>
+  req: Request<RequestParamKeyType, D>
   env: E['Bindings']
   finalized: boolean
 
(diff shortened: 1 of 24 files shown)


usualoma:
@yusukebe 
That's wonderful!
The following points are also very good

> * Easy to understand the result data because "declaring property name first".
> * Easy to writing rules. v.isLength(400) is better than [v.isLength, 400] .
> * Getting only "declared properties". Additional properties are always ignored. It's safe.

I think the implementation of this PR is good, but let me confirm one point.

### Is it best to insert it as middleware?

For example, we could define it as a function like the following.
* Define rules
* Applying the rule

I think this way can be implemented only in the middleware without complicating hono.ts.

I also think this approach would allow multiple rules to be applied in a single request.

However, I understand the opinion that "inserting as middleware is more declarative to the request".

```typescript
import { validationRule, validate } from 'middleware/validator'

const rule = validationRule((v) => ({
  id: v.json('post.id').asNumber().isRequired(),
  title: v.json('post.title').isRequired().isLength(400),
  body: v.json('post.body').isOptional(),
}))
app.post('/posts', (c) => {
  const data = validate(c, rule)
  return c.text(`ID is ${data.id}, Title is ${data.title}`)
})
```

yusukebe:
Hi @usualoma !

Thank you for the comment.

> However, I understand the opinion that "inserting as middleware is more declarative to the request".

Yes. I think it's better to validate before dispatching the handler. Besides, the API is prettier if it's as Middleware.

I would like to go with this if no one else has an opinion.

usualoma:
I think we can go ahead with the PR as is!

usualoma:
Hi @yusukebe!
Is the merge result of this PR currently working correctly? I got the following results in my environment.

image

I made the following changes and it worked.

```ts
diff --git a/src/middleware/validator/middleware.ts b/src/middleware/validator/middleware.ts
index 5e9d059..d168c47 100644
--- a/src/middleware/validator/middleware.ts
+++ b/src/middleware/validator/middleware.ts
@@ -7,14 +7,12 @@ type ValidationFunction<T> = (v: Validator, c: Context) => T
 
 type Schema = Record<string, VString | VNumber | VBoolean | VObject>
 type SchemaToProp<T> = {
-  [K in keyof T]: T[K] extends VBase
-    ? T[K]['type'] extends 'number'
-      ? number
-      : T[K]['type'] extends 'boolean'
-      ? boolean
-      : T[K]['type'] extends 'string'
-      ? string
-      : never
+  [K in keyof T]: T[K] extends VNumber
+    ? number
+    : T[K] extends VBoolean
+    ? boolean
+    : T[K] extends VString
+    ? string
     : never
 }
 
```

Alternatively, the following changes also worked.

```ts
diff --git a/src/middleware/validator/validator.ts b/src/middleware/validator/validator.ts
index d12bba6..54ca8e9 100644
--- a/src/middleware/validator/validator.ts
+++ b/src/middleware/validator/validator.ts
@@ -163,9 +163,10 @@ export abstract class VBase {
 }
 
 export class VString extends VBase {
+  type: 'string' = 'string'
+
   constructor(options: VOptions) {
     super(options)
-    this.type = 'string'
   }
 
   isEmpty = (
@@ -212,9 +213,10 @@ export class VString extends VBase {
 }
 
 export class VNumber extends VBase {
+  type: 'number' = 'number'
+
   constructor(options: VOptions) {
     super(options)
-    this.type = 'number'
   }
 
   isGte = (min: number) => {
@@ -227,9 +229,10 @@ export class VNumber extends VBase {
 }
 
 export class VBoolean extends VBase {
+  type: 'boolean' = 'boolean'
+
   constructor(options: VOptions) {
     super(options)
-    this.type = 'boolean'
   }
 
   isTrue = () => {
```

yusukebe:
@usualoma 

You are right. I have fixed the problem with the following PR, please review it.

#545

PR #540, opened 2022-09-19
feat(router): support optional route

yusukebe:
This PR introduces "optional parameter":

```ts
app.get('/api/animals/:type?',() => {})
```

Will match:

```http
GET /api/animals
GET /api/animals/dog
```

Reference:
https://router.vuejs.org/guide/essentials/route-matching-syntax.html#sensitive-and-strict-route-options


yusukebe:
@usualoma !

If you agree on this specification and implementation, could you make it for RegExpRouter? That it wouldn't be hard to follow this implementation.

usualoma:
@yusukebe 
How about #542?

yusukebe:
Hi @usualoma !

Looks good! I will merge it.

PR #545, opened 2022-09-20
fix(validator): make "Types" work well

yusukebe:
Fixed the bug pointed here: <https://github.com/honojs/hono/pull/535#issuecomment-1252210438>

PR #546, opened 2022-09-20
feat(validator): Enable verification results to be retrieved as structured data.

usualoma:
Hi @yusukebe!

How about this additional feature?


usualoma:
@yusukebe 
Sorry, I had created this PR against fix/validator-types and it seems to have been closed with #545, so please change it to main.

yusukebe:
@usualoma 

It seems that I do not have the privilege to do that; can't you do it?

usualoma:
I can't either.

Just forgot denoify also, so I will recreate another PR.

PR #547, opened 2022-09-20
feat(validator): Enable verification results to be retrieved as structured data.

usualoma:
Same as #546 

How about this additional feature?


yusukebe:
Hi @usualoma !

Great feature! I love it. I'd like to merge it into the main, is there anything else left to do?

usualoma:
@yusukebe 
Thank you. That's all I'm trying to do!

Issue #552, opened 2022-09-24
[validator] types of `c.env.` will disappear when using validator


yusukebe:

yusukebe:
Fixed by #553

PR #553, opened 2022-09-25
fix(types): enable types for `c.env` when using validator

### LINKED ISSUE #551: [validator] Context inside `done` has not types

yusukebe:
Now, you can do like this:

SS

Close #551 #552


dn-l:
@yusukebe could you please release this fix?

PR #555, opened 2022-09-26
fix(validator): make `isOptional` rule works correctly

yusukebe:


ThatOneBro:
Great! I was actually going to write an issue about this, I appreciate this change 👍

Issue #561, opened 2022-09-27
Feature request: Custom error responses for hono/validator


Skye-31:
Currently, when using the validator middleware, there's no (nice) way of providing your own error responses when the request passes validation, resulting in a plaintext response that looks something like ```Invalid Value: the JSON body "name" is invalid - undefined```. This isn't ideal for an API where the main format is JSON.

This feature request is to allow people to deal with formatting the error themselves.

ThatOneBro:
Hey @Skye-31 , I agree, this is a necessary feature! I actually have patch, however it's for the old external validator middleware and will need to be ported to the new built-in middleware. I have been planning on creating a PR with this but this issue gives me a good cue that it is time! 

The system I made allows you to pass a `defaultMessageFunc` to the middleware like this:
```ts
{
  ...
  defaultMessageFunc: (ruleName: string, key: string, value: string) =>
          createApiErrorObject(key, getErrorByRuleName(ruleName), value),
}
```

Then in the done function, I'm able to do something like this to return the errors as a json response:
```ts
{
  ...
  done: (result, c) => {
     if (result.hasError) {
       return c.json({ errors: result.messages }, 400);
     }
  },
}
```
I also have thought of allowing this system to be extended at each field, but this would be able to overwrite the default validation error message directly (at least for the json type, maybe separate functions could be specified for each type of validated input source for incoming requests).

@yusukebe What do you think of such a solution? 😄 

If you think it's good, I can modify my existing patch for the new middleware and create a PR for feedback.

yusukebe:
Hi!
I have a few ideas on this matter. Please wait a moment, as I will introduce them later.

Skye-31:
The syntax of this looks good to me 🙂

Skye-31:
Any news on this? 🙂 
I currently have to do a really hacky workaround to get this working for production

yusukebe:
Hi @Skye-31 

Sorry for the late response! Sorry for the late response. I'm just implementing a new way to show the error messages.

By the way, do you know the "done" option? Possibly it will help you get what you want to do.

```ts
app.post(
  '/posts',
  validator(
    (v) => ({
      title: v.body('title').isRequired(),
      body: v.body('body').isRequired(),
    }),
    {
      done: (res, c) => {
        if (res.hasError) {
          return c.json({
            errors: res.messages,
          })
        }
      },
    }
  ),
  (c) => {
    return c.text('Valid!')
  }
)
```

If it has errors, this will return like this:

```json
[...]

PR #563, opened 2022-09-28
feat(validator): add support for JSON array path validation

ThatOneBro:
This PR adds support for JSON array paths in the JSONPath util, and also support for validating JSON array paths from a validator context.

The update doesn't break backwards compatibility with the current API (besides the previous less standard `.i` syntax, eg. `posts.4` instead of `posts[4]`) and provides the ability to validate arbitrary array paths, including complex paths of nested arrays and objects.

For example this:
```ts
const jsonBody = {
  posts: [
    {
       title: 'New Post 1',
       tags: ['new-ish', false],
    },
    {
       title: 'New Post 2',
       tags: ['newest', true],
    },
  ],
}
```
Can have arbitrary validations like this:
```ts
validator((v) => ({
  title: v.json('posts[*].title').isRequired(),
  secondTag: v.json('posts[*].tags[1]').asBoolean().isRequired(),
}))
```

One thing I would like to do in the future is to provide messages to the API consumer about which specific values in an array path failed validation, eg. `posts[2].title is required -- undefined`. I have a working version of this in a patch for the old middleware using `validator.js`, but it would need to be modified, as mentioned in #561. However, that is for another time. As of now, I think I am ready for comment on this PR.

Let me know if there is anything that needs to be changed with this PR before merge, or if you think this is a good approach at all.

### DIFF
--- a/src/middleware/validator/validator.ts
+++ b/src/middleware/validator/validator.ts
@@ -128,7 +128,11 @@ export abstract class VBase {
         result.message = this._message
       } else {
         const valToStr = Array.isArray(value)
-          ? `[${value.map((val) => (val === undefined ? 'undefined' : val)).join(', ')}]`
+          ? `[${value
+              .map((val) =>
+                val === undefined ? 'undefined' : typeof val === 'string' ? `"${val}"` : val
+              )
+              .join(', ')}]`
           : value
         switch (this.target) {
           case 'query':
--- a/src/middleware/validator/validator.ts
+++ b/src/middleware/validator/validator.ts
@@ -307,18 +307,21 @@ export class VObject extends VBase {
 }
 
 export class VNumberArray extends VNumber {
+  isArray: true
   constructor(options: VOptions) {
     super(options)
     this.isArray = true
   }
 }
 export class VStringArray extends VString {
+  isArray: true
   constructor(options: VOptions) {
     super(options)
     this.isArray = true
   }
 }
 export class VBooleanArray extends VBoolean {
+  isArray: true
   constructor(options: VOptions) {
     super(options)
     this.isArray = true
--- a/src/middleware/validator/validator.ts
+++ b/src/middleware/validator/validator.ts
@@ -1,9 +1,10 @@
 import { JSONPath } from '../../utils/json'
+import type { JSONObject, JSONPrimative, JSONArray } from '../../utils/json'
 import { rule } from './rule'
 import { sanitizer } from './sanitizer'
 
 type Target = 'query' | 'header' | 'body' | 'json'
-type Type = string | number | boolean | object | null | undefined
+type Type = JSONPrimative | JSONObject | JSONArray | File
 type Rule = (value: Type) => boolean
 type Sanitizer = (value: Type) => Type
 
@@ -126,18 +127,21 @@ export abstract class VBase {
       if (this._message) {
         result.message = this._message
       } else {
+        const valToStr = Array.isArray(value)
+          ? `[${value.map((val) => (val === undefined ? 'undefined' : val)).join(', ')}]`
+          : value
         switch (this.target) {
           case 'query':
-            result.message = `Invalid Value: the query parameter "${this.key}" is invalid - ${value}`
+            result.message = `Invalid Value: the query parameter "${this.key}" is invalid - ${valToStr}`
             break
           case 'header':
-            result.message = `Invalid Value: the request header "${this.key}" is invalid - ${value}`
+            result.message = `Invalid Value: the request header "${this.key}" is invalid - ${valToStr}`
             break
           case 'body':
-            result.message = `Invalid Value: the request body "${this.key}" is invalid - ${value}`
+            result.message = `Invalid Value: the request body "${this.key}" is invalid - ${valToStr}`
             break
           case 'json':
-            result.message = `Invalid Value: the JSON body "${this.key}" is invalid - ${value}`
+            result.message = `Invalid Value: the JSON body "${this.key}" is invalid - ${valToStr}`
             break
         }
       }
@@ -147,26 +151,51 @@ export abstract class VBase {
 
   private validateValue = (value: Type): boolean => {
     // Check type
-    if (typeof value !== this.type) {
-      if (this._optional && typeof value === 'undefined') {
-        // Do nothing.
-        // The value is allowed to be `undefined` if it is `optional`
-      } else {
-        return false
+    if (Array.isArray(value)) {
+      if (!value[0] || typeof value[0] !== this.type) {
+        if (this._optional && typeof value[0] === 'undefined') {
+          // Do nothing.
+          // The value is allowed to be `undefined` if it is `optional`
+        } else {
+          return false
+        }
       }
-    }
 
-    // Sanitize
-    for (const sanitizer of this.sanitizers) {
-      value = sanitizer(value)
-    }
-
-    for (const rule of this.rules) {
-      if (!rule(value)) {
-        return false
+      // Sanitize
+      for (const sanitizer of this.sanitizers) {
+        value = value.map((innerVal) => sanitizer(innerVal)) as JSONArray
       }
+
+      for (const rule of this.rules) {
+        for (const val of value) {
+          if (!rule(val)) {
+            return false
+          }
+        }
+      }
+      return true
+    } else {
+      if (typeof value !== this.type) {
+        if (this._optional && typeof value === 'undefined') {
+          // Do nothing.
+          // The value is allowed to be `undefined` if it is `optional`
+        } else {
+          return false
+        }
+      }
+
+      // Sanitize
+      for (const sanitizer of this.sanitizers) {
+        value = sanitizer(value)
+      }
+
+      for (const rule of this.rules) {
+        if (!rule(value)) {
+          return false
+        }
+      }
+      return true
     }
-    return true
   }
 }
 
--- a/src/middleware/validator/validator.ts
+++ b/src/middleware/validator/validator.ts
@@ -58,7 +58,7 @@ export abstract class VBase {
 
   isRequired = () => {
     return this.addRule((value: unknown) => {
-      if (value) return true
+      if (value !== undefined && value !== null && value !== '') return true
       return false
     })
   }
@@ -115,7 +115,7 @@ export abstract class VBase {
       value = body[this.key]
     }
     if (this.target === 'json') {
-      const obj = await req.json()
+      const obj = (await req.json()) as JSONObject
       value = JSONPath(obj, this.key)
     }
 
--- a/src/middleware/validator/validator.ts
+++ b/src/middleware/validator/validator.ts
@@ -1,10 +1,10 @@
 import { JSONPath } from '../../utils/json'
-import type { JSONObject, JSONPrimative, JSONArray } from '../../utils/json'
+import type { JSONObject, JSONPrimitive, JSONArray } from '../../utils/json'
 import { rule } from './rule'
 import { sanitizer } from './sanitizer'
 
 type Target = 'query' | 'header' | 'body' | 'json'
-type Type = JSONPrimative | JSONObject | JSONArray | File
+type Type = JSONPrimitive | JSONObject | JSONArray | File
 type Rule = (value: Type) => boolean
 type Sanitizer = (value: Type) => Type
 
--- a/src/middleware/validator/validator.ts
+++ b/src/middleware/validator/validator.ts
@@ -3,7 +3,7 @@ import { rule } from './rule'
 import { sanitizer } from './sanitizer'
 
 type Target = 'query' | 'header' | 'body' | 'json'
-type Type = string | number | boolean | object | undefined
+type Type = string | number | boolean | object | null | undefined
 type Rule = (value: Type) => boolean
 type Sanitizer = (value: Type) => Type
 
--- a/src/middleware/validator/validator.ts
+++ b/src/middleware/validator/validator.ts
@@ -152,12 +152,11 @@ export abstract class VBase {
   private validateValue = (value: Type): boolean => {
     // Check type
     if (Array.isArray(value)) {
-      if (!value[0] || typeof value[0] !== this.type) {
-        if (this._optional && typeof value[0] === 'undefined') {
-          // Do nothing.
-          // The value is allowed to be `undefined` if it is `optional`
-        } else {
-          return false
+      for (const val of value) {
+        if (typeof val !== this.type) {
+          // Value is of wrong type here
+          // If not optional, or optional and not undefined, return false
+          if (!this._optional || typeof val !== 'undefined') return false
         }
       }
 
--- a/src/middleware/validator/validator.ts
+++ b/src/middleware/validator/validator.ts
@@ -27,6 +27,7 @@ type VOptions = {
   target: Target
   key: string
   type?: 'string' | 'number' | 'boolean' | 'object'
+  isArray?: boolean
 }
 
 export abstract class VBase {
@@ -35,6 +36,7 @@ export abstract class VBase {
   key: string
   rules: Rule[]
   sanitizers: Sanitizer[]
+  isArray: boolean
   private _message: string | undefined
   private _optional: boolean
   constructor(options: VOptions) {
@@ -43,6 +45,7 @@ export abstract class VBase {
     this.type = options.type || 'string'
     this.rules = []
     this.sanitizers = []
+    this.isArray = options.isArray || false
     this._optional = false
   }
 
@@ -122,7 +125,7 @@ export abstract class VBase {
     result.value = value
     result.isValid = this.validateValue(value)
 
-    if (result.isValid == false) {
+    if (result.isValid === false) {
       if (this._message) {
         result.message = this._message
       } else {
@@ -154,7 +157,11 @@ export abstract class VBase {
 
   private validateValue = (value: Type): boolean => {
     // Check type
-    if (Array.isArray(value)) {
+    if (this.isArray) {
+      if (!Array.isArray(value)) {
+        return false
+      }
+
       for (const val of value) {
         if (typeof val !== this.type) {
           // Value is of wrong type here
@@ -300,11 +307,20 @@ export class VObject extends VBase {
 }
 
 export class VNumberArray extends VNumber {
-  isArray = true
+  constructor(options: VOptions) {
+    super(options)
+    this.isArray = true
+  }
 }
 export class VStringArray extends VString {
-  isArray = true
+  constructor(options: VOptions) {
+    super(options)
+    this.isArray = true
+  }
 }
 export class VBooleanArray extends VBoolean {
-  isArray = true
+  constructor(options: VOptions) {
+    super(options)
+    this.isArray = true
+  }
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,10 +1,6 @@
 {
-  "eslint.validate": [
-    "javascript",
-    "javascriptreact",
-    "typescript",
-    "typescriptreact"
-  ],
+  "deno.enable": false,
+  "eslint.validate": ["javascript", "javascriptreact", "typescript", "typescriptreact"],
   "editor.codeActionsOnSave": {
     "source.fixAll.eslint": true
   }
--- a/deno_dist/middleware/validator/middleware.ts
+++ b/deno_dist/middleware/validator/middleware.ts
@@ -1,14 +1,37 @@
 import type { Context } from '../../context.ts'
 import type { Environment, Next, ValidatedData } from '../../hono.ts'
 import { VBase, Validator } from './validator.ts'
-import type { VString, VNumber, VBoolean, VObject, ValidateResult } from './validator.ts'
+import type {
+  VString,
+  VNumber,
+  VBoolean,
+  VObject,
+  VNumberArray,
+  VStringArray,
+  VBooleanArray,
+  ValidateResult,
+} from './validator.ts'
 
 type Schema = {
-  [key: string]: VString | VNumber | VBoolean | VObject | Schema
+  [key: string]:
+    | VString
+    | VNumber
+    | VBoolean
+    | VObject
+    | VStringArray
+    | VNumberArray
+    | VBooleanArray
+    | Schema
 }
 
 type SchemaToProp<T> = {
-  [K in keyof T]: T[K] extends VNumber
+  [K in keyof T]: T[K] extends VNumberArray
+    ? number[]
+    : T[K] extends VBooleanArray
+    ? boolean[]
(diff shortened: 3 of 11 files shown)


yusukebe:
Hi @ThatOneBro !

Wow, great! This is one of the things I wanted to do. It seems to me that implementation is OK.
I want to merge it.

@usualoma 
What about do you think?

> One thing I would like to do in the future is to provide messages to the API consumer about which specific values in an array path failed validation, eg. posts[2].title is required -- undefined.

It would be better if we could do this.

BTW, as I commented in this PR #561 , I am in the progress of implementing a new way to handle error messages.

ThatOneBro:
> Wow, great! This is one of the things I wanted to do. It seems to me that implementation is OK. I want to merge it.

Great! Glad to hear it. I'm glad it's not too broken then 😆 

> BTW, as I commented in this PR https://github.com/honojs/hono/issues/561 , I am in the progress of implementing a new way to handle error messages.

Oh yeah great! I'll wait for that change then! 😄 

Actually, about this PR, what do you think about these line where type is enforced? Are these checks too lax? Should it check the whole array, or will this be a problem?
```ts
@@ -147,26 +151,51 @@ export abstract class VBase {

  private validateValue = (value: Type): boolean => {
    // Check type
    if (typeof value !== this.type) {
      if (this._optional && typeof value === 'undefined') {
        // Do nothing.
        // The value is allowed to be `undefined` if it is `optional`
      } else {
        return false
    if (Array.isArray(value)) {
      if (!value[0] || typeof value[0] !== this.type) { // Is this too naive about other array elements?
        if (this._optional && typeof value[0] === 'undefined') { // Is this too naive about other array elements?
          // Do nothing.
          // The value is allowed to be `undefined` if it is `optional`
        } else {
          return false
        }
      }
    }
```

yusukebe:
@ThatOneBro 

> Actually, about this PR, what do you think about these line where type is enforced? Are these checks too lax? Should it check the whole array, or will this be a problem?

Ah, I think we should check all the values in the whole array.

usualoma:
Hi @ThatOneBro !
I think this PR is a great idea and a great implementation!
Would be nice if we could merge it!

### We should be aware of whether the resulting value is an array or not

The request in the following code will be 400 in the main branch but 200 in this PR branch.
This is not the result the user expects, so we should be aware of whether the resulting value is an array or not.

```ts
  app.post(
    '/post',
    validator((v) => ({
      post: {
        title: v.json('post.title').isAlpha(),
      }
    })),
    (c) => {
      const post = c.req.valid().post
      return c.json({post})
    }
  )

  it('Should return 400 response', async () => {
    const json = {
      post: {
        title: ['Hello'], // <- array
      },
    }
    const req = new Request('http://localhost/post', {
      method: 'POST',
      body: JSON.stringify(json),
    })
    expect(res.status).toBe(400)
  })
```

src/middleware/validator/validator.ts
```diff
@@ -1,9 +1,10 @@
 import { JSONPath } from '../../utils/json'
+import type { JSONObject, JSONPrimative, JSONArray } from '../../utils/json'
 import { rule } from './rule'
 import { sanitizer } from './sanitizer'
 
 type Target = 'query' | 'header' | 'body' | 'json'
-type Type = string | number | boolean | object | undefined
+type Type = JSONPrimative | JSONObject | JSONArray | File
```
usualoma, line 7: I think this is a typo.

```suggestion
type Type = JSONPrimitive | JSONObject | JSONArray | File
```

ThatOneBro:
@yusukebe 
> Ah, I think we should check all the values in the whole array.

Added some new commits to check all the types in an array, and also some more tests 😄 Let me know what you think!

src/middleware/validator/validator.ts
ThatOneBro: You're right! I fixed it

ThatOneBro:
> ### We should be aware of whether the resulting value is an array or not
> The request in the following code will be 400 in the main branch but 200 in this PR branch. This is not the result the user expects, so we should be aware of whether the resulting value is an array or not.

Good point, so we should either make it explicit in the JSONPath like this:
```ts
validator((v) => ({
  post: {
    title: v.json('post.title[*]').isAlpha(),
  }
})),
```
Or as a a method like this:
```ts
validator((v) => ({
  post: {
    title: v.json('post.title').isArrayOf().isAlpha(),
    published: v.json('post.published').isArrayOf().asBoolean().isAlpha(),
  }
})),
```
Or maybe, we need to just a VArray class, which contains VBase subclasses?
```ts
validator((v) => ({
  post: {
    title: v.json('post.title').asArray('string').isAlpha(),
  }
})),
```

What do you think @yusukebe @usualoma ?

yusukebe:
@ThatOneBro @usualoma 

I think this is a matter of whether or not to allow "Array" in the validated result values. If it supports "Array" values and adds "Array" type, the test code will be like the below:

```ts
  app.post(
    '/post',
    validator((v) => ({
      post: {
        tags: v.json('post.title').asArray('string').isAlpha(),
      },
    })),
    (c) => {
      const res = c.req.valid()
      const tags = res.post.tags // <== string[]
      const tag1 = tags[0] // <== "Workers"
      return c.json({ tag1: tag1 })
    }
  )

  it('Should return 400 response', async () => {
    const json = {
      post: {
        tags: ['Workers', 'Deno', 'Bun'],
      },
    }
    const req = new Request('http://localhost/post', {
      method: 'POST',
      body: JSON.stringify(json),
    })
    expect(res.status).toBe(200)
    expect(await res.json()).toEqual({ tag1: 'Workers' })
  })
```

And we should make "VArray".

We have to consider how well it will support the type, but yes, "asArray" is useful.

usualoma:
I haven't confirm if it is possible, but it would be nice to be able to write like this.

```ts
app.post(
  '/post',
  validator((v) => ({
    post: {
      tags: v.json('post.title').asArray((v) => v.asString().isAlpha()),
    },
  })),
  (c) => {
    const res = c.req.valid()
  }
)

it('Should return 200 response', async () => {
  const json = {
    post: {
      tags: ['Workers', 'Deno', 'Bun'],
    },
  }
  const req = new Request('http://localhost/post', {
    method: 'POST',
    body: JSON.stringify(json),
  })
  expect(res.status).toBe(200)
})
```

yusukebe:
Ah,  

```ts
v.json('post.title').asArray((v) => v.asString().isAlpha())
```

I like that "callback" style, but it is used too much, so I think users may be confused. So it's good for me to be like below:

```ts
v.json('post.title').asArray('string').isAlpha()
```

I haven't thought about implementation, but it is not impossible.

usualoma:
In the meantime, I will note what I felt.
However, if we decide to go with the `asArray('string')` style, I agree with that decision.


### asArray('string')

This seemed inconsistent with `asNumber()` and `asBoolean()`.
If this writing style (`asArray('string')`) is adopted, it would be consistent if there is `as('string')` or `as('boolean')` and then `asArrayOf('string')`.

Yes, of course, I understand that `asNumber()` and `asBoolean()` are already released feature, so changing them is not desirable.


### asNumber().asArray()

This makes me wonder if I should write `asNumber().asArray()` or `asArray().asNumber()`.

It might be possible to make a specification that either is acceptable.

yusukebe:
> It might be possible to make a specification that either is acceptable.

Yes. I think `asNumber().asArray()` and `asArray().asNumber()` should both behave the same.

usualoma:
@ThatOneBro 

I think if we change it to something like the following PR, it would be more compatible with https://jsonpath.com/ and allow for a variety of specifications.
https://github.com/ThatOneBro/hono/pull/1

ThatOneBro:
> @ThatOneBro
> 
> I think if we change it to something like the following PR, it would be more compatible with https://jsonpath.com/ and allow for a variety of specifications. [ThatOneBro#1](https://github.com/ThatOneBro/hono/pull/1)

Great work! Yes I think this is better. I did notice two failing tests from the change, I will merge the PR and attempt to fix them. 
After that I'll propose my implementation of `VArray` and `asArray` .

ThatOneBro:
> I've implemented what I think about `asArray`. This is a verbose code, but you can refer to it.
Great! I think there are a few broken tests in the latest version, should I change them on this PR, or will you resolve them?

Currently there are two that are broken:
```
● Handle array paths › Should provide error with invalid array paths

    expect(received).toBe(expected) // Object.is equality

    Expected: "Invalid Value: the JSON body \"posts[*].rating[3]\" is invalid - [undefined, undefined, undefined]"
    Received: "Invalid Value: the JSON body \"posts[*].rating[3]\" is invalid - undefined"
    
● Handle array paths › Should allow optional array paths

    expect(received).toEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 0

      Array [
        "cool",
    -   undefined,
        "lame",
      ]
```

I think test one *could* be modified but test 2 probably *should* still pass with the new implementation.

yusukebe:
@ThatOneBro 

Let's do below:

* I'll create PR to your branch: https://github.com/ThatOneBro/hono/tree/add-jsonpath-array
* Please resolve the test problem above.
* I think my PR lacks some tests. Please add more tests if needed.
* Feel free to make changes if you are not satisfied with the implementation.

usualoma:
@ThatOneBro 

Thanks for merging my PR.

Sorry I didn't mention it in that PR. I think that in validator.test.ts the expected result should be modified.
```diff
diff --git a/src/middleware/validator/validator.test.ts b/src/middleware/validator/validator.test.ts
index 60381ef..32673e4 100644
--- a/src/middleware/validator/validator.test.ts
+++ b/src/middleware/validator/validator.test.ts
@@ -257,7 +257,7 @@ describe('Handle array paths', () => {
     const res = await validator.validate(req)
     expect(res.isValid).toBe(true)
     expect(res.message).toBeUndefined()
-    expect(res.value).toEqual(['cool', undefined, 'lame'])
+    expect(res.value).toEqual(['cool', 'lame'])
   })
 
   it('Should provide error with invalid array paths', async () => {
@@ -265,10 +265,10 @@ describe('Handle array paths', () => {
     const res = await validator.validate(req)
     expect(res.isValid).toBe(false)
     const messages = [
-      'Invalid Value: the JSON body "posts[*].rating[3]" is invalid - [undefined, undefined, undefined]',
+      'Invalid Value: the JSON body "posts[*].rating[3]" is invalid - undefined',
     ]
     expect(res.message).toBe(messages.join('\n'))
-    expect(res.value).toEqual([undefined, undefined, undefined])
+    expect(res.value).toEqual(undefined)
   })
 
   it('Should prevent invalid types within array', async () => {
```

The reason is that the following results are obtained at https://jsonpath.com/, and I think it is better to have the same behavior at hono.

image

image

usualoma:
I think the added `asArray` implementation is a good one!
I also think that the following additional modifications are needed in my opinion.

```diff
diff --git a/src/middleware/validator/validator.ts b/src/middleware/validator/validator.ts
index c4372b8..e92f9bb 100644
--- a/src/middleware/validator/validator.ts
+++ b/src/middleware/validator/validator.ts
@@ -35,6 +35,7 @@ export abstract class VBase {
   key: string
   rules: Rule[]
   sanitizers: Sanitizer[]
+  isArray: boolean = false
   private _message: string | undefined
   private _optional: boolean
   constructor(options: VOptions) {
@@ -154,7 +155,11 @@ export abstract class VBase {
 
   private validateValue = (value: Type): boolean => {
     // Check type
-    if (Array.isArray(value)) {
+    if (this.isArray) {
+      if (!Array.isArray(value)) {
+        return false
+      }
+
       for (const val of value) {
         if (typeof val !== this.type) {
           // Value is of wrong type here
```

ThatOneBro:
Hey @usualoma!
> The reason is that the following results are obtained at https://jsonpath.com/, and I think it is better to have the same behavior at hono

Yes I agree we should keep the implementation in line with jsonpath. So the `undefined` instead of `[undefined, undefined, undefined]` makes a lot of since, and I don't think it affects validation. 

However, returning `["cool", "lame"]` instead of `["cool", undefined, "lame"]` will break the validation as implemented, right? You could check if all array lengths match the expected length, but you cannot easily know which indices are invalid without doing a diff I guess.

How should we handle this? I would like to be able to return some error message like this:
`posts[*].rating[1] is invalid -- undefined`
to the user. How could we accomplish this?

> I think the added asArray implementation is a good one!
I also think that the following additional modifications are needed in my opinion.

You're right! I was just writing tests for this functionality. I'll add your change too!

ThatOneBro:
Actually there's still one question about arrays I have:
Should an array type be differentiated from deeply nested values? For example:
```ts
const req = {
  posts: [
    { 
       title: 'Post 1',
       tags: ['post', 'long'],
    },
    { 
       title: 'Post 2',
       tags: ['post', 'short'],
    },
  ],
}
```
```ts

v => { 
  // Should these two be treated the same? 
  title: v.json('posts[*].name').asArray(), // Should `asArray()` be required here?
  tags: v.json('posts[*].tags').asArray(),
}
```
In the example, the first item `title` is not an array value being evaluated, just it happens that the path returns an array of these values which should be evaluated with the same validators. However, the second item `tags` is an actual array. Should these be treated the same or differently? Personally, I think the first one shouldn't require an `asArray()` call. What do you think?

yusukebe:
Hi @ThatOneBro 

I think this requires `asArray()`:

```ts
title: v.json('posts[*].title').asArray()
```

`asSomeType` means "expect SomeType and will cast to SomeType". Since `VString` is the default in this Validator, if we expect an array, we should use `asArray()` to make it a `VStringArray`. And the "actual value that is evaluated" should be used. That should be the value actually used by the user.

If it is a nested array, we can write it like this:

```ts
tags: v.json('post[*].tags[*]').asArray()
```

What do you think?

usualoma:
Hmmm, sorry if I'm wrong. I thought about it, and I think what @ThatOneBro needs should be accomplished in this way. (Not a problem that should be solved with `*`, in my opinion.)

```ts
describe('validate nested structure by asArray', () => {
  const app = new Hono()
  app.post(
    '/post',
    validator((v) => ({
      posts: v.json('posts').asArray((nest) => ({
        title: nest.json('title').isAlpha(),
        tags: nest.json('tags').asArray().isRequired(),
        ids: nest.json('ids').asNumber().asArray(),
      }))
    })),
    (c) => {
      const res = c.req.valid()
      return c.json({ tag1: res.posts[0].tags[0] })
    }
  )

  it('Should return 200 response', async () => {
    const json = {
      post: {
        title: 'foo',
        tags: ['Workers', 'Deno', 'Bun'],
        ids: [1, 3, 5],
      },
    }
    const req = new Request('http://localhost/post', {
      method: 'POST',
      body: JSON.stringify(json),
    })
    expect(res.status).toBe(200)
    expect(await res.json()).toEqual({ tag1: 'Workers' })
  })
})
```

I think we can assign a type to this.

image

image

image

I think this would be a good way to return a thoughtful error message (if we try hard enough).

ThatOneBro:
@usualoma You know what, this makes a lot of sense. I thought about doing this in a similar way before, but because the original middleware used JSONPath syntax, I tried to make it work with that. This actually may be better for nested objects inside of arrays. I'll wait for @yusukebe to make a comment before we start work on this though.

yusukebe:
Ah, yes! It would be best if we could do the following that @usualoma mentioned:

```ts
validator((v) => ({
  posts: v.json('posts').asArray((nest) => ({
    title: nest.json('title').isAlpha(),
    tags: nest.json('tags').asArray().isRequired(),
    ids: nest.json('ids').asNumber().asArray(),
  })),
}))
```

It's good!!

@ThatOneBro 

Can you do it this way?

usualoma:
I think the following

### `props[*]` and `asArray()` work very well!

```ts
title: v.json('posts[*].title').asArray()
```

```ts
tags: v.json('post[*].tags[*]').asArray()
```

these are very nice features to be able to write in this way.

It would be a rather large change if we were to do nested validation, so I think it would be better to merge them once in this PR (sorry if you don't get the result you originally intended), which is the addition of these (`*` and asArray() ) features.

I think it would be a good idea to have a separate PR for nested validation.

ThatOneBro:
> these are very nice features to be able to write in this way.

True, I think this is good. I will edit some tests to make sure that they are testing the desired functionality.

> I think it would be a good idea to have a separate PR for nested validation.

Oh yeah, you're probably right. We'll save it for another PR.

There is one last thing I have a concern about though, is the isRequired method for arrays. If a post is missing a title, it still validates successfully with the current method. For example:
```ts
  const v = new Validator()
  const json = {
    post: {
      title: ['Hello'],
      flags: [true, false],
      published: [true],
      comment: [
        {
          title: 'abc',
          flags: [true, false],
        },
        {
          title: 'def',
          flags: [false, true],
        },
        {
          flags: [true],
        },
      ],
    },
  }

  const req = new Request('http://localhost/', {
    method: 'POST',
    body: JSON.stringify(json),
  })

  it('Should fail validation if `isRequired` and missing entry for path', async () => {
    const validator = v.json('post.title').asArray().isRequired()
    const res = await validator.validate(req)

    expect(res.isValid).toBe(false)
  })
// Result:
// expect(received).toBe(expected) // Object.is equality
// Expected: false
// Received: true

Do we think this is okay?

**Comment by yusukebe** [comment]
Is this the problem pointed out in this comment?

https://github.com/honojs/hono/pull/563#issuecomment-1262822204

I think it should be `["cool", undefined, "name"]` rather than `["cool", "name"]`. it is important to keep the `https://jsonpath.com` specification, but it is more important to be able to validate undefined.

**Comment by ThatOneBro** [comment]
> Is this the problem pointed out in this comment?
> 
> [#563 (comment)](https://github.com/honojs/hono/pull/563#issuecomment-1262822204)
> 
> I think it should be `["cool", undefined, "name"]` rather than `["cool", "name"]`. it is important to keep the `https://jsonpath.com` specification, but it is more important to be able to validate `undefined`.

Yes that's right. I think that we may have to violate the spec in this case, as much as we don't want to do that. What do you think about this @usualoma ? Does it make sense in this case?

**Comment by usualoma** [comment]
OK.

```ts
["cool", undefined, "name"].
```

If we were to adopt this, I think the other problem should also be as follows.

```ts
[undefined, undefined, undefined]
```

In that case, the code in json.ts could be simpler with the following changes.

```diff
diff --git a/src/utils/json.test.ts b/src/utils/json.test.ts
index 03fc1e3..5586f6f 100644
--- a/src/utils/json.test.ts
+++ b/src/utils/json.test.ts
@@ -109,7 +109,7 @@ describe('JSONPath', () => {
   })
 
   it('Should return undefined if there is no matching path.', () => {
-    expect(JSONPath(data, 'releases[*].i386')).toBe(undefined)
+    expect(JSONPath(data, 'releases[*].i386')).toEqual([undefined, undefined, undefined])
   })
 
   it('Should return value from nested array', () => {
diff --git a/src/utils/json.ts b/src/utils/json.ts
index f4d9613..cd79dfb 100644
--- a/src/utils/json.ts
+++ b/src/utils/json.ts
@@ -3,8 +3,6 @@ export type JSONArray = (JSONPrimitive | JSONObject | JSONArray)[]
 export type JSONObject = { [key: string]: JSONPrimitive | JSONArray | JSONObject }
 export type JSONValue = JSONObject | JSONArray | JSONPrimitive
 
-const noMatch = Symbol('no match')
-
 const JSONPathInternal = (data: JSONValue, parts: string[]): JSONValue => {
   const length = parts.length
   for (let i = 0; i < length && data !== undefined; i++) {
@@ -14,31 +12,15 @@ const JSONPathInternal = (data: JSONValue, parts: string[]): JSONValue => {
     }
 
     if (typeof data !== 'object' || data === null) {
-      throw noMatch
+      return undefined
     }
 
     if (p === '*') {
       const restParts = parts.slice(i + 1)
-      const values = Object.values<JSONValue>(data).map((v): JSONValue | typeof noMatch => {
-        try {
-          return JSONPathInternal(v, restParts)
-        } catch (e) {
-          if (e === noMatch) {
-            return noMatch
-          } else {
-            throw e
-          }
-        }
-      })
-      if (values.every((v) => v === noMatch)) {
-        throw noMatch
-      }
-      const matches = values.filter((v): v is JSONValue => v !== noMatch)
-      return restParts.indexOf('*') === -1 ? matches : matches.flat()
-    } else if (p in data) {
-      data = (data as JSONObject)[p] // `data` may be an array, but accessing it as an object yields the same result.
+      const values = Object.values(data).map((v) => JSONPathInternal(v, restParts))
+      return restParts.indexOf('*') === -1 ? values : values.flat()
     } else {
-      throw noMatch
+      data = (data as JSONObject)[p] // `data` may be an array, but accessing it as an object yields the same result.
     }
   }
   return data

```

ThatOneBro:
So all functionality related to validator and JSONPath is working, all tests passing. Looks like it should be good!

However, one last thing before the PR can go through, it seems there is a type error coming from the update to SchemaToProp with the addition of the Array types to the generic...

We are getting:
Screen Shot 2022-09-29 at 11 11 44 PM

For some reason it thinks that `data.value` is always a `string[]` instead of `string`. Any quick suggestions?

usualoma:
Ah, I see, TypeScript is very clever and considers the same type if there is no change in the property definition 😢 
After reverting b8ddef8, I think it can be fixed by adding the following changes.

```ts
diff --git a/src/middleware/validator/validator.ts b/src/middleware/validator/validator.ts
index 32802c0..4705cdc 100644
--- a/src/middleware/validator/validator.ts
+++ b/src/middleware/validator/validator.ts
@@ -307,18 +307,21 @@ export class VObject extends VBase {
 }
 
 export class VNumberArray extends VNumber {
+  isArray: true
   constructor(options: VOptions) {
     super(options)
     this.isArray = true
   }
 }
 export class VStringArray extends VString {
+  isArray: true
   constructor(options: VOptions) {
     super(options)
     this.isArray = true
   }
 }
 export class VBooleanArray extends VBoolean {
+  isArray: true
   constructor(options: VOptions) {
     super(options)
     this.isArray = true
```

ThatOneBro:
> Ah, I see, TypeScript is very clever and considers the same type if there is no change in the property definition 😢

Good catch! After that fix, I think I am ready to request final comments before merge! Let me know if there is anything else you see that should be changed or fixed first.

usualoma:
Great work!!!
I have no additional comments.

ThatOneBro:
> Great work!!! I have no additional comments.

Thank you! Your feedback and refactoring was great! This PR went very quickly thanks to you both. I am glad to work with such responsive and skilled people 😄

yusukebe:
@ThatOneBro 

Great!!
Have you done it already? I'll check again this later and merge it.

ThatOneBro:
> Great!! Have you done it already? I'll check again this later and merge it.

Yes, I've committed everything and tests all seem to pass! I've finalized everything we have talked about it seems 👍

yusukebe:
I've checked, and it's great! Thank you, @ThatOneBro , @usualoma !

I'll merge it now.

PR #571, opened 2022-09-30
Fix Trie router multiple matching path parameters

### LINKED ISSUE #570: Path templates with parameter prefix
Sort of a follow up from #569 but figured it was a subtly different issue. Having renamed the parameter, I've found the router doesn't match a longer path template if it first matches a shorter one.

I've put together a failing test here: https://github.com/honojs/hono/compare/main...gatesn:hono:ngates/prefix-params?expand=1

I believe this can be fixed by removing this `break`: https://github.com/honojs/hono/blob/13ce2eef950f8eb11fe219280b0d5f61bd4fbec1/src/router/trie-router/node.ts#L190

But I'm not familiar enough with the code base to know whether that's correct, or just happens to work in this case.

Many thanks for this project, it's incredibly useful and a pleasure to use :)

gatesn:
Fixes #570

### DIFF
--- a/deno_dist/router/trie-router/node.ts
+++ b/deno_dist/router/trie-router/node.ts
@@ -187,7 +187,6 @@ export class Node<T> {
             if (typeof name === 'string' && !matched) {
               params[name] = part
             }
-            break
           }
         }
       }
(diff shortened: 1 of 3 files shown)


yusukebe:
@gatesn 

Thank you for your PR.

One request. Could you do `yarn denoify` ? The `denoify` command generates code for the Deno. And please commit & push them.

gatesn:
Yup, done 🙂

PR #573, opened 2022-10-01
fix(cache): export cache middleware for Deno

ThatOneBro:
We should export the Cache middleware for the Deno distribution so that we can use it in Deno 1.26+ now that it is available.


yusukebe:
Hi @ThatOneBro !

You are right. I'm glad you noticed. Thank you!

Issue #575, opened 2022-10-02
Error handling x middleware regression since 2.1.3


MattIPv4:
👋 In versions 2.1.3 and beyond, there is a regression in how errors interact with middleware...

In version 2.1.2 and prior, if an error occurred in a route and `app.onError` was invoked, middleware would still be processed for the error response. This would allow middleware such as the built-in `logger` and built-in `cors` to still process the error response, logging the response or injecting CORS headers.

In version 2.1.3 and since, if an error occurs and `app.onError` responds, middleware does not have a chance to process the response, so middleware such as `logger` cannot log the response, and `cors` cannot inject any CORS headers.

```js
const { Hono } = require('hono');
const { logger } = require('hono/logger');
const { cors } = require('hono/cors');

const app = new Hono();
app.use('*', logger());
app.use('*', cors({ origin: '*' }));

app.get('/error', () => {
  throw new Error('Test error');
});

app.onError((err, ctx) => {
  return ctx.json({
    error: true,
    status: 500,
    message: err.message,
  });
});

const main = async () => {
  const res = await app.request('http://localhost/error');
  console.log([ ...res.headers.entries() ]);
};

main();
```

[RunKit demo: hono@latest](https://runkit.com/mattipv4/6338d5a353c9c20008210cfa)

[RunKit demo: hono@2.1.2](https://runkit.com/mattipv4/6338d71a084d5a0008dd6ed0)

This is pretty breaking for my project, as I would like to have error responses logged, and I'd like them to contain CORS headers like any other response.

yusukebe:
You are right; we have changed the way of error handling. However, this is not a good approach, and I would like to revert back to it, with other changes.

Please wait!

yusukebe:
Hi @MattIPv4 !

In this PR, we make it behaves the same as 2.1.2 and prior. The CORS and logger middleware will work even after the Handler throws an Error.

https://github.com/honojs/hono/pull/576

The new version, including this fix, will be released soon!

MattIPv4:
Awesome, thank you! I look forward to testing it once that release is cut 💙

yusukebe:
Hi! new version has been released:

https://github.com/honojs/hono/releases/tag/v2.2.4

MattIPv4:
Hey! It looks like it's been fixed for synchronous routes :tada:, but the issue still remains for any route that is asynchronous 😢 

[2.1.2](https://runkit.com/mattipv4/633b6d2e74e576000865b3c1)

[...]

PR #576, opened 2022-10-02
fix(compose): don't throw Error; set error response into `c.res`

yusukebe:
**This PR has a small breaking change**

Currently, if an error occurred in the Handler, it would immediately throw an Error. Therefore, the Middleware process after `await next()` will not be executed. When handling errors in Middleware, it was necessary to wrap them in try-catch.

```ts
  try {
    await next()
  } catch (_) {
    //...
  }
})
```

This works well for Sentry Middleware, but has a bad effect on Logger Middleware and CORS. This is mentioned #575 

This way of error handling was introduced in #491.

However, referring above, it is not so good, so in this PR, reverting back to the way and makinging it possible to handle Error objects in the Middleware.

If the Handler throws an error, there is no need to try-catch in the Middleware, and the error resopnse is passed into `c.res`.

```ts
  await next()
  // c.res is error reponse
})

app.get('/error', () => {
  throw new Error('Test error')
})
```

If you wanted to handle the original Error object in the Middleware, it would be in `c.error`.

```ts
  await next()
  const originalMessage = c.error.message
  //...
})
```

For example, for Sentry Middleware, the following:

https://github.com/honojs/sentry/blob/ca742a2ea35808cf2b70a19407a93481b4fbe8a0/src/index.ts#L51-L56

Write like this:

```ts
await next()
if (c.error) {
  sentry.captureException(c.error)
  throw error
}
```

### DIFF
--- a/deno_dist/compose.ts
+++ b/deno_dist/compose.ts
@@ -1,5 +1,5 @@
 import { HonoContext } from './context.ts'
-import type { Environment, NotFoundHandler } from './hono.ts'
+import type { Environment, NotFoundHandler, ErrorHandler } from './hono.ts'
 
 interface ComposeContext {
   finalized: boolean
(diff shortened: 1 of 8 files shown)


ThatOneBro:
Looks good to me. I think this makes a lot of sense, allows for middleware to be aware of errors from other middleware before it 👍

Zerebokep:
So with this change we don't need to call captureException inside app.onError when using sentry?

yusukebe:
Hi @Zerebokep !

Do you use "Sentry Middleware" or not? If you are **not** using Middleware, it must continue to call `captureException` in onError.

PR #578, opened 2022-10-02
fix(validator): Malformed JSON body now returns HTTP 400 Bad Request

### LINKED ISSUE #562: Hono/validator: Unexpected end of JSON input
When using something like the following, and there being an invalid request body (e.g. malformed json, or no body at all), the library throws an unexpected end of json error, with status 500.
```js
validator((v) => ({
  email: v
    .json("email")
    .isRequired()
}))
```

It should probably throw a 400 error instead 🙂

ThatOneBro:
This PR adds a new type of error that will be caught by the default error handler: `HTTPError`. It can be passed any valid `StatusCode` and it will return this code along with the `StatusText` from `getStatusText()` as the response body. 

The validator middleware now catches when `Request.json()` fails and throws `new HTTPError(400)` to be caught and returned to the client. This should probably close #562 if I'm not mistaken.

### DIFF
--- a/deno_dist/hono.ts
+++ b/deno_dist/hono.ts
@@ -9,6 +9,8 @@ import { RegExpRouter } from './router/reg-exp-router/index.ts'
 import { SmartRouter } from './router/smart-router/index.ts'
 import { StaticRouter } from './router/static-router/index.ts'
 import { TrieRouter } from './router/trie-router/index.ts'
+import { getStatusText, HttpError } from './utils/http-status.ts'
+import type { StatusCode } from './utils/http-status.ts'
 import { getPathFromURL, mergePath } from './utils/url.ts'
 
 export interface ContextVariableMap {}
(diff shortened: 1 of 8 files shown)


yusukebe:
Hi @ThatOneBro !

Sorry, this PR is not acceptable for the following reasons:

* The policy is not to create the object extending "Error"  as possible. (But we will consider it if necessary.)
* If it throws an error in the middleware, other middleware cannot handle the error. This matter is related to #576 #575 
* It would be good to create an Issue about Malformed JSON body error handling separately from the HttpError object issue.

If the Middleware returns an error, e.g., 400 response, it is better to assign a Response to `c.res`.

```ts
      try {
        result = await validator.validate(c.req)
      } catch (err) {
        c.res = c.text('400', 400)
        await next()
        return
      }
```

This is not done in the Validator Middleware yet, but is implemented in the following Basic Auth Midldeware.

https://github.com/honojs/hono/blob/main/src/middleware/basic-auth/index.ts#L56

ThatOneBro:
> Hi @ThatOneBro !

Hello @yusukebe !

> * The policy is not to create the object extending "Error"  as possible. (But we will consider it if necessary.)

> * It would be good to create an Issue about Malformed JSON body error handling separately from the HttpError object issue.

Okay, so I won't create `HttpError`, at least not one that extends `Error` 👍 

> If the Middleware returns an error, e.g., 400 response, it is better to assign a Response to `c.res`.

Sure that makes sense, I didn't see your PR until I already finished this PR 😅 But this definitely makes more sense. I'll create a new PR with the new strategy.

Do you think just adding:

```ts
try {
  const obj = (await req.json()) as JSONObject
  value = JSONPath(obj, this.key)
} catch (e) {
  throw Error('Malformed JSON in request')
}
```

and

```ts
try {
  result = await validator.validate(c.req)
} catch (e) {
  return c.text(getStatusText(400), 400)
}
```
Will cover this PR? We can forget about the `HttpError` for now probably.

yusukebe:
> I didn't see your PR until I already finished this PR

Sorry!!

> Do you think just adding:

Yes, it's pretty good!!

ThatOneBro:
Closed in favor of #579.

PR #579, opened 2022-10-02
fix(validator): return HTTP 400 on malformed JSON request

### LINKED ISSUE #562: Hono/validator: Unexpected end of JSON input

ThatOneBro:
This PR should close #562. The Validator middleware will now return a HTTP 400 Bad Request response whenever JSON in the request body fails to parse, or if the body is empty.

### DIFF
--- a/deno_dist/middleware/validator/middleware.ts
+++ b/deno_dist/middleware/validator/middleware.ts
@@ -1,5 +1,6 @@
 import type { Context } from '../../context.ts'
 import type { Environment, Next, ValidatedData } from '../../hono.ts'
+import { getStatusText } from '../../utils/http-status.ts'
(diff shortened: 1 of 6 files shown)


yusukebe:
Hi @ThatOneBro !

This is good! One request. Please do `denoify`?

ThatOneBro:
Hey @yusukebe!
Sorry about that, totally forgot! 😅 Should be good to go now!

yusukebe:
Thank you!!

PR #580, opened 2022-10-02
fix(reg-exp-router): Register path for all the methods if method === METHOD_NAME_ALL.

### LINKED ISSUE #574: Regression of `.all()` between 2.1.4 and 2.2.0
Using Cloudflare Workers

Example Code:
```typescript
app.get('/cf', CFEndpoint)
app.get('/version', VersionEndpoint)
app.all(`/cors`, CORSHandle)

app.all('/', async (c) => {
    if (c.env.PRODUCTION === 'true') {
        return c.redirect(
            'https://github.com/Cyb3r-Jak3/workers-serverless-api',
            301
        )
    } else return await c.notFound()
})
app.all('*', (c) => c.notFound())

export default app
```
On 2.1.4 a request to `/cors` goes to the `/cors` endpoint. On 2.2.0 it will return a 404. If `app.all(`/cors`, CORSHandle)` is replaced with `app.get(`/cors`, CORSHandle)` then it gets handled.

usualoma:
fixes #574

@yusukebe I think this PR will fix #574!

### DIFF
--- a/deno_dist/router/reg-exp-router/router.ts
+++ b/deno_dist/router/reg-exp-router/router.ts
@@ -119,17 +119,17 @@ export class RegExpRouter<T> implements Router<T> {
       const path = paths[i]
 
       routes[method] ||= {}
-      routes[method][path] ||= [
-        ...(routes[METHOD_NAME_ALL][path] ||
-          findMiddleware(middleware[method], path) ||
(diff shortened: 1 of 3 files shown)


yusukebe:
Thank you! I'll check it and merge it later!

yusukebe:
Hi @usualoma !

Thank you! Looks good! I'll merge it.

PR #589, opened 2022-10-07
perf: remove unrequired cloning of ctx.res

JakeChampion:
Response constructor's 2nd argument does not require a cloned response as all it does is read these fields and does not modify anything:
- status
- statusText
- headers

> https://fetch.spec.whatwg.org/#initialize-a-response


yusukebe:
Hi @JakeChampion !

Thank you! You are right! I'll merge it.

Issue #584, opened 2022-10-04
Feature: nested validation


ThatOneBro:
In #563, we talked about nested validation with arrays. @usualoma mentioned an interesting way to handle nested validations like this:

```ts
describe('validate nested structure by asArray', () => {
  const app = new Hono()
  app.post(
    '/post',
    validator((v) => ({
      posts: v.json('posts').asArray((nest) => ({
        title: nest.json('title').isAlpha(),
        tags: nest.json('tags').asArray().isRequired(),
        ids: nest.json('ids').asNumber().asArray(),
      }))
    })),
    (c) => {
      const res = c.req.valid()
      return c.json({ tag1: res.posts[0].tags[0] })
    }
  )

  it('Should return 200 response', async () => {
    const json = {
      post: {
        title: 'foo',
        tags: ['Workers', 'Deno', 'Bun'],
        ids: [1, 3, 5],
      },
    }
    const req = new Request('http://localhost/post', {
      method: 'POST',
      body: JSON.stringify(json),
    })
    const res = await app.request(req)
    expect(res.status).toBe(200)
    expect(await res.json()).toEqual({ tag1: 'Workers' })
  })
})
```

I think we can assign a type to this.

<img width="558" alt="image" src="https://user-images.githubusercontent.com/30598/193165303-dbf14b39-1d97-429b-8874-07ccfb6e27eb.png">

<img width="625" alt="image" src="https://user-images.githubusercontent.com/30598/193165323-95a14dff-9e94-4c05-95ca-564d958ed1b3.png">

<img width="645" alt="image" src="https://user-images.githubusercontent.com/30598/193165344-924cba2b-1ed8-4e45-a69c-dbc0d1c9c968.png">

_Originally posted by @usualoma in https://github.com/honojs/hono/issues/563#issuecomment-1262975936_

I think this coupled with some extension of the `asObject()` method, would allow clean and proper validation of any complex valid JSON structure! Hono users could then do things like this:
```ts
    validator((v) => ({
      posts: v.json('posts').asArray((nest) => ({
        title: nest.json('title').isAlpha(),
        meta: nest.json('meta').asObject((nest) => ({
          // A user should probably not be able to create posts like this, but it's an example ;)
          relatedPosts: nest.json('relatedPosts').asArray((nest) => ({
            title: nest.json('title').isAlpha(),
            tags: nest.json('tags').asArray().isRequired(),
            ids: nest.json('ids').asNumber().asArray(),
          })),
[...]

PR #592, opened 2022-10-09
feat(validator): introduce nested validation

yusukebe:
This PR introduces "nested validation" for Validator Middleware.

In this PR, we can validate nested JSON structured data using `v.array()` and `v.object()` functions. After much thought and consideration, the result is this API.

For example, there is a nested object like below:

```ts
const data = {
  posts: [
    {
      id: 123,
      title: 'JavaScript',
      tags: ['Workers', 'Deno', 'Bun'],
    },
  ],
  pager: {
    prev: true,
    next: false,
  },
}
```

It will be validated with `v.array()` and `v.object()`:

```ts
app.post(
  '/posts',
  validator((v) => ({
      id: v.json('id').asNumber().isRequired(),
      title: v.json('title'),
      tags: v.json('tags').asArray(),
    })),
    pager: v.object('pager', (v) => ({
      prev: v.json('prev').asBoolean(),
      next: v.json('next').asBoolean(),
    })),
  })),
  (c) => {
    return c.text('Valid!')
  }
)
```

And it also has types:

SS

Isn't it cool? I want to know your opinion.

## Implementation

It has become a little bit complex. There may be something missing or wrong points in this implementation. If you notice any, please review them.

The code for `JSONPathCopy` is based on @ThatOneBro's  and @usualoma 's code.

https://github.com/honojs/hono/pull/508/files#diff-91414c404fd51d6fb8e47ba511fd1a5c410e5fe3e732254f0f6d621175fcd647

This PR related: #584

### DIFF
--- a/deno_dist/middleware/validator/middleware.ts
+++ b/deno_dist/middleware/validator/middleware.ts
@@ -1,7 +1,8 @@
 import type { Context } from '../../context.ts'
 import type { Environment, Next, ValidatedData } from '../../hono.ts'
 import { getStatusText } from '../../utils/http-status.ts'
-import { VBase, Validator } from './validator.ts'
+import { mergeObjects } from '../../utils/object.ts'
+import { VBase, Validator, VObjectBase } from './validator.ts'
 import type {
   VString,
   VNumber,
@@ -11,18 +12,20 @@ import type {
   VStringArray,
   VBooleanArray,
   ValidateResult,
+  VArray,
 } from './validator.ts'
 
-type Schema = {
+export type Schema = {
   [key: string]:
     | VString
     | VNumber
     | VBoolean
-    | VObject
     | VStringArray
     | VNumberArray
     | VBooleanArray
     | Schema
+    | VObject<Schema>
+    | VArray<Schema>
 }
 
 type SchemaToProp<T> = {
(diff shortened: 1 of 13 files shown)


yusukebe:
The tests were passed on my environment, but these failed on the CI. I know it's about copying objects. I'll investigate to solve it.

ThatOneBro:
Looks good to me so far! I will look again tomorrow to make sure everything makes sense, but from first glances it looks great. I was able to use nested arrays successfully on CFW 👍

usualoma:
Awesome job!
I agree with these changes!

### Minor issues

I don't think we are testing `asArray` here. Do we need to correct the name of the test?

https://github.com/honojs/hono/pull/592/files#diff-69cfe3c04c26baba3e5078f8d04c7434c55883045ab8bb255a27f09383330f5fR406

yusukebe:
> I don't think we are testing asArray here. Do we need to correct the name of the test?

I forgot to remove them! Thank you!

And, this test fails on the CI, do you know the solution?
https://github.com/honojs/hono/actions/runs/3216097795/jobs/5257654578#step:7:292

It does not fail on my environment. I think this is a problem related to copying objects, but I don't know how to fix it.

ThatOneBro:
> It does not fail on my environment. I think this is a problem related to copying objects, but I don't know how to fix it.

I just tested it in a Docker container, it seems it may be an issue with Ubuntu or Linux in general. On my Mac, the tests do not fail. I will continue looking into it more, hopefully we can figure it out easily.

ThatOneBro:
I created a PR (#598) for this feature branch that should fix the CI issues, it introduces an immutable version of the mergeObjects.

yusukebe:
Hi @ThatOneBro !

Wow cooooool! Tests have been passed. Thank you very much!

I'll merge this PR to the main branch later.

PR #603, opened 2022-10-16
feat(validator): has the error message per rule

yusukebe:
In this PR, the validator will have multiple result objects, so it will have an error message per rule.

For example, one error message will become more friendly.

```
'Invalid Value: the request body "title" is invalid - abcdefg'
```

will be:

```
'Invalid Value [abcdef]: the request body "title" is invalid - isLength'
```

Internally, it has error objects as follows:

```json
[
  {
      "isValid": true,
      "target": "body",
      "key": "title",
      "value": "abcdefg",
      "ruleName": "should be \"string\"",
      "ruleType": "type"
  },
  {
      "isValid": false,
      "message": "Invalid Value [abcdefg]: the request body \"title\" is invalid - isLength",
      "target": "body",
      "key": "title",
      "value": "abcdefg",
      "ruleName": "isLength",
      "ruleType": "value"
  }
]
```

If you want to do more complex error handling, you can refer to this result objects in the `done` option.

```ts
const vHandler = validator(
  (v) => ({
    title: v.body('title').isLength({ max: 5 }),
  }),
  {
    done: (res, c) => {
      if (res.hasError) {
        // ... do something
        // console.log(res.results)
      }
    },
  }
)
```

This PR is related to #561

### DIFF
--- a/deno_dist/middleware/validator/middleware.ts
+++ b/deno_dist/middleware/validator/middleware.ts
@@ -96,28 +96,39 @@ export const validatorMiddleware = <T extends Schema, Env extends Partial<Enviro
     let data: any = {}
 
     for (const [keys, validator] of validatorList) {
-      let result: ValidateResult
+      let results: ValidateResult[]
       try {
-        result = await validator.validate(c.req)
+        results = await validator.validate(c.req)
       } catch (e) {
         // Invalid JSON request
         return c.text(getStatusText(400), 400)
       }
-      if (result.isValid) {
-        // Set data on request object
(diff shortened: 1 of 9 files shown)


yusukebe:
Hi @ThatOneBro !

Could you see this PR?

ThatOneBro:
Hi @yusukebe , sorry it took so long! I meant to already review this PR, I'll look at it right now! 😄

ThatOneBro:
I did some testing and I think this PR is good.

I added a test in #609 in case you want to merge the test in too.

PR #608, opened 2022-10-18
fix(bearer-auth, basic-auth): handler should not be executed when unauthorized

NOBLES5E:
This is a huge security concern. Imagine securing a nuclear launch handler with it :scream: 

I would suggest yanking all previous versions with this middleware

### DIFF
--- a/src/middleware/basic-auth/index.test.ts
+++ b/src/middleware/basic-auth/index.test.ts
@@ -4,11 +4,15 @@ import { basicAuth } from '.'
 
 describe('Basic Auth by Middleware', () => {
   const crypto = global.crypto
+  let handlerExecuted: boolean
   beforeAll(() => {
     global.crypto = require('crypto').webcrypto
   })
   afterAll(() => {
     global.crypto = crypto
+  })
+  beforeEach(() => {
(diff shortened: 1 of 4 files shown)


yusukebe:
Hi @NOBLES5E !
I'll do it soon. Thank you!!

yusukebe:
@NOBLES5E Thanks again!

You are right, this is bad. As your implementation, we must "return" the Response.

>  // To Hono maintainers: I don't think we can support nested middleware when the middleware returns early
>  // since `next()` is never called in this case. Feel free to correct me

We need to support "nested" middleware because there is a case only the context has "username/password". For example, Cloudflare Workers / Modules mode.

I think the best way is to return `auth` as follows, what do you think? This is a breaking change, but it's  OK because it is a security fix.

```ts
app.use('/auth/*', async (c, next) => {
  const auth = basicAuth({ username: c.env.USERNAME, password: c.env.PASSWORD })
  return auth(c, next)
})
```

However, this is an issue that should be resolved as quickly as possible, so I may merge or I may fix it without your response.

NOBLES5E:
Thank you for your prompt action : )

PR #629, opened 2022-10-25
Fix ESM support in NodeJS

pavi2410:
It's because of missing file extension in import paths

https://github.com/honojs/hono/blob/d8da569299ff85fb0a2923d4359892f1c171134d/src/hono.ts#L1


yusukebe:
Hi @pavi2410 !

You are right, the supporting ESM is not working well. I think your way is good, but I'm investigating another way. Please wait.

pavi2410:
Thank you @yusukebe ! I appreciate that

yusukebe:
Hi @pavi2410 !

I've made PR #631 , and I'll release the new version, including this fix, later. If you try using the released version and still have problems, please feel free to create an Issue. Thank you!

PR #631, opened 2022-10-27
fix(build): fix ESM support in Node.js / use `esbuild`

yusukebe:
This PR fixes ES Modules support in Node.js is not working well. Previously, if we use Hono package in Node.js directly, the error would throw.

SS

I've done the following to fix this while keeping support for CommonJS.

* Declare `type: "module"` in "package.json".
* Use "esbuild" and make the script "build.ts" to build the projects instead of `tsc`.
* Add extensions for importing modules, in ESM files that are emitted by esbuild, eg:`import { compose } from "./compose.js"`.
* Copy "package.json" for CommonJS into `dist/cjs/package.json` (a little tricky).

This has enabled proper support for both ESM and CommonJS. However, this isn't easy to test and may not work well in your environment. If there is a problem, please let me know. We will fix it.

This PR is inspired by #629 . Thanks @pavi2410 .


yusukebe:
I'll merge it!

PR #634, opened 2022-10-29
fix(typo) package.cjs.json target to type

taishinaritomi:
The "type" in package.cjs.json was set to "target".
The bug did not occur because the default "type" in package.json is "commonjs".
I have not checked another property but it may have been overridden by default.

This package.cjs.json also worked
```
{}
```

### DIFF
--- a/package.cjs.json
+++ b/package.cjs.json
@@ -1,3 +1,3 @@
 {
-  "target": "CommonJS"
+  "type": "commonjs"
 }


yusukebe:
Hi @taishinaritomi !

You are right! I was confused with `tsconfig.json`. Thanks for fixing the typo that was difficult to realize!

PR #635, opened 2022-10-29
fix(validator): enable handling "Bad Request" in validator

yusukebe:
Currently, with Validator middleware, if the JSON body is empty, it will return "Malformed JSON in request body" response error immediately. This behavior is good but it's not possible to handle in the validator middleware `done` method. This is mentioned in #561.

To handle "Malformed JSON in request body" error we have to write another middleware to get an error object. It's very verbose. So, in this PR, I made it we can handle "Malformed JSON in request body" in validator middleware.

If we have handler like below:

```ts

app.post(
  '/posts',
  validator(
    (v) => ({
      title: v.json('title'),
    }),
    {
      done: (results, c) => {
        return c.json(results, 400)
      },
    }
  ),
  (c) => {
    return c.json({
      message: 'Valid!',
    })
  }
)
```

Then, post the empty body to the endpoint, the result response will be this:

```json
{
  "hasError": true,
  "messages": [
    "Malformed JSON in request body"
  ],
  "results": [
    {
      "isValid": false,
      "message": "Malformed JSON in request body",
      "target": "unknown",
      "key": null,
      "value": null,
      "ruleName": "Malformed JSON in request body",
      "ruleType": "value"
    }
  ]
}
```

This will make it easier to handle cases where the JSON body content is empty.

### DIFF
--- a/deno_dist/middleware/validator/middleware.ts
+++ b/deno_dist/middleware/validator/middleware.ts
@@ -49,7 +49,15 @@ export const validatorMiddleware = <
         results = await validator.validate(c.req as Request)
       } catch (e) {
         // Invalid JSON request
-        return c.text(getStatusText(400), 400)
+        if (e instanceof Error) {
+          const result = getErrorResult(e)
+          resultSet.hasError = true
+          resultSet.messages = [result.message || '']
+          resultSet.results = [result]
+          break
+        } else {
+          return c.text(getStatusText(400), 400)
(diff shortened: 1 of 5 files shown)


ThatOneBro:
Looks good to me! Tested it out, works as expected. Good fix 👍

yusukebe:
Yeah, I'll merge it!

Issue #625, opened 2022-10-24
How to get full type safety with extracted handler function?


yujong-lee:
Below is minimal example that I am curious about.

```ts
app.get(`/api/:id`, (c) => {
  const key = c.req.param().id; // This is auto-completed
});

app.get(`/api/:id`, handler);
const handler = (c: Context<string, Environment>) => {
  const key = c.req.param().id; // This is NOT auto-completed
}

```

Could you guide me how to deal with this? Thanks.

yusukebe:
Hi @yujong-lee !

You can pass the literal `id` to Context as generics:

```ts
const handler = (c: Context<'id'>) => {
  const key = c.req.param('id')
}
```

![SS](https://user-images.githubusercontent.com/10682/197545125-5c41534b-4c98-4104-91ca-3eff5692277d.png)

It's a little bit tricky :)

yujong-lee:
@yusukebe Thank you.

yusukebe:
We have to consider two things.

First, we can use `Handler` instead of `Context` to define the handler.

```ts
const handler: Handler = (c) => {
  return c.text('foo')
}
```

It's better because `Handler` will warn if the response is not returned.

<img width="351" alt="SS" src="https://user-images.githubusercontent.com/10682/197658477-8747bbef-93ad-493f-bbc0-64dd204b6b31.png">

Second. Generics are complex. If the users want to declare the handler as their own, they must use Generics for type inference:

```ts
type Env = {
  Variables: {
    foo: string
  }
}

const schema = (v: Validator) => ({
  query: v.query('query').isRequired(),
  page: v.query('page').isNumeric().isOptional(),
})

const handler: Handler<'id', Env, ReturnType<typeof schema>> = (c) => {
  const id = c.req.param('id')
  const { query, page } = c.req.valid()
  const foo = c.get('foo')
  return c.text('foo')
}
```

This is an extreme example, but writing Generics is complicated for the user. If at all possible, I would like the users to write a handler in `app.get()`.

```ts
const app = new Hono<Env>()

// ...

app.get('/foo/:id', validator(schema), (c) => {
  return c.text('foo')
})
```

However, there are cases where users want to define handlers separately, as @yujong-lee did.

If we could overload Interface or type, we could write it with the following syntax, but it seems impossible.

<img width="653" alt="SS" src="https://user-images.githubusercontent.com/10682/197660045-23e7b294-634b-4478-8eda-1ea20d7d639c.png">

[...]

PR #637, opened 2022-10-30
feat(types): introduce `CustomHandler` interface

yusukebe:
In this PR, introduce `CustomHandler` interface that is exported for the users as `Handler`.

If we want to declare a handler outside of `app.get()`, we write it such like this:

```ts
const handler: Handler = (c) => {
  return c.text('Hi!')
}
```

Then, in this `handler`, how can we make `c.env` or `c.get('foo')` have a type? Pass `string` as the first argument of the Generics for `Handler` and write the following:

```ts
type Env = {
  Bindings: { TOKEN: string }
  Variables: { post: Post }
}

const handler: Handler<string, Env> = (c) => { // <---
  const post = c.get('post')
  return c.json(post)
}
```

This is redundant to write `string` on purpose. So, in this PR, we can write just like this:

```ts
const handler: Handler<Env> = (c) => {
  const post = c.get('post')
  return c.json(post)
}
```

Of course, it keep to have the types:

SS

In actually, we can do something similar for Validator schema:

```ts
const schema = (v: Validator) => ({
  query: v.query('q'),
})

type Schema = ReturnType<typeof schema>

const handler: Handler<Schema> = (c) => {
  const { query } = c.req.valid()
  return c.json({ query: query })
}
```

SS

I think this is helpful to declare the handler independently for the user.

### DIFF
--- a/deno_dist/compose.ts
+++ b/deno_dist/compose.ts
@@ -1,5 +1,5 @@
 import { HonoContext } from './context.ts'
-import type { Environment, NotFoundHandler, ErrorHandler } from './hono.ts'
+import type { Environment, NotFoundHandler, ErrorHandler } from './types.ts'
 import type { Schema } from './validator/schema.ts'
 
 interface ComposeContext {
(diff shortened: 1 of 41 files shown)


yujong-lee:
This closes #625. 

Great feature. Looking forward to this being published.

yusukebe:
@yujong-lee 

Thanks, I'll merge this and a new version will be released soon.

It is tested well, but it is difficult to notice problems with "Types", so please feel free to create an Issue if you notice any.

PR #640, opened 2022-10-31
test(routing): add test cases for route grouping order

yujong-lee:
```ts
it('only works with correct order', async () => {
  three.get('/hi', (c) => c.text('hi'))
  two.route('/three', three)
  one.route('/two', two)

  const { status } = await one.request('http://localhost/two/three/hi', { method: 'GET' })
  expect(status).toBe(200)
})
```

Since this is kind-of-mentioned in [Routing Priority](https://honojs.dev/docs/api/routing/#routing-priority), it is easy to order routes incorrectly when grouping them. 
(I did. 😄 )

I made some test cases in Hono's repo to understand what's happening in my app. I think it is good to add here too. And eventually, we better have some more detailed examples [here](https://github.com/honojs/honojs.dev/blob/main/content/docs/api/routing.md).

### DIFF
--- a/deno_dist/compose.ts
+++ b/deno_dist/compose.ts
@@ -1,5 +1,5 @@
 import { HonoContext } from './context.ts'
-import type { Environment, NotFoundHandler, ErrorHandler } from './hono.ts'
+import type { Environment, NotFoundHandler, ErrorHandler } from './types.ts'
 import type { Schema } from './validator/schema.ts'
 
 interface ComposeContext {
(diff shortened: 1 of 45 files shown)


yusukebe:
Hi @yujong-lee !

Thank you for the PR. You are right, this leads to mistakes. Sorry for wasting your time;(

We want to merge it, so please fix the test to pass the CI.

yusukebe:
@yujong-lee 

Added the description about it: https://honojs.dev/docs/api/routing/#grouping-ordering

It's a little bit difficult to explain, then if you have the more helpful description, please give us the PR :)

yusukebe:
@yujong-lee 

Thanks!

Issue #413, opened 2022-07-19
`Context.platform`


yusukebe:
I want to know which platform the application works on. Is it on Cloudflare? Fastly? Bun? Deno?

If we can get the name of the platform by `Context.platform`, it's useful to make middleware or something.

```ts
const name = c.platform
```

yujong-lee:
This can be closed due to #644.

PR #644, opened 2022-11-02
feat(context): Introducing c.runtime

yujong-lee:
As mentioned in #413, it might be useful to have way to know on which platform the code is running. This PR introduces `c.runtime`, which is getter of `Context` interface.

# Node, Deno, and Bun
```ts
const global = globalThis as any

if (global?.process?.title === 'node') {
  return 'node'
}

if (global?.Deno !== undefined) {
  return 'deno'
}

if (global?.Bun !== undefined) {
  return 'bun' 
}
```
This is straightforward. 

For `Deno`, `global?.process?.title === 'deno'` is possible,  but I just used shorter one.

# Cloudflare and Fastly
```ts
if (typeof global?.WebSocketPair === 'function') {
  return 'cloudflare'
}

return 'fastly'
```

[WebSocketPair](https://developers.cloudflare.com/workers/runtime-apis/websockets/) is Cloudflare Worker's runtime API. I picked this because this is only available in CF. (For example, [cache](https://developers.cloudflare.com/workers/runtime-apis/cache/) exists in other platform with similar API.)

For `Fastly`, I don't know good way to determine it at runtime yet. (Resources: [1](https://github.com/honojs/compute-starter-kit), [2](https://developer.fastly.com/reference/compute/ecp-env/))

But it is almost safe to say it is `Fastly` if it is not others.
```ts
type Platform = 'node' | 'deno' | 'bun' | 'cloudflare' | 'fastly'
```

### DIFF
--- a/bun_test/index.test.tsx
+++ b/bun_test/index.test.tsx
@@ -25,9 +25,9 @@ describe('Basic', () => {
     expect(res.headers.get('x-query')).toBe('bar')
   })
 
-  it('returns current platform (bun)', async () => {
+  it('returns current runtime (bun)', async () => {
     const c = new HonoContext(new Request('http://localhost/'))
-    expect(c.platform).toBe('bun')
+    expect(c.runtime).toBe('bun')
   })
 })
 
--- a/bun_test/index.test.tsx
+++ b/bun_test/index.test.tsx
@@ -1,4 +1,5 @@
 import { describe, expect, it } from 'bun:test'
+import { HonoContext } from '../src/context'
 import { Hono } from '../src/index'
 import { basicAuth } from '../src/middleware/basic-auth'
 import { jwt } from '../src/middleware/jwt'
@@ -22,6 +23,11 @@ describe('Basic', () => {
     expect(await res.text()).toBe('Hello Deno!')
     expect(res.headers.get('x-param')).toBe('foo')
     expect(res.headers.get('x-query')).toBe('bar')
+  })
+
+  it('returns current platform (bun)', async () => {
+    const c = new HonoContext(new Request('http://localhost/'))
+    expect(c.platform).toBe('bun')
   })
 })
 
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -7,6 +7,7 @@ import type { Schema, SchemaToProp } from './validator/schema.ts'
 type HeaderField = [string, string]
 type Headers = Record<string, string | string[]>
 export type Data = string | ArrayBuffer | ReadableStream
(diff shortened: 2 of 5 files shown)


yusukebe:
Hi @yujong-lee !

Wow great!! This is exactly what I wanted!

One thing:

> But it is almost safe to say it is Fastly if it is not others.

I think it is better not to judge this as `fastly`. There are other platforms (not just runtime) that Hono runs on. For example, it works on Edge Runtime by Vercel.

* https://zenn.dev/yusukebe/articles/f3b130f69287fc

It also may work on [Vercel Edge Functions](https://vercel.com/features/edge-functions) (Although, Vercel's backend is powered by Cloudflare Workers!).

There is another platform/runtime that has not yet been released called Lagon.

* https://lagon.app

It is possible that there will be another platform on which the web standard API will work, so it is better to set `other` if it is difficult to judge.

Or do you have a better idea than `other`?

yujong-lee:
`other` seems fine to me. I was thinking `unknown`, but `unknown` feels like something got wrong. 

`other` implies that it is just something that we don't support judging it yet.

yujong-lee:
1. I found way to determine `Fastly`
[Fastly compute serve](https://developer.fastly.com/reference/cli/compute/serve/) minify code by default and have no options to turn it off. So I way struggling to find `Fastly-specific global variable`. But I found this [document](https://js-compute-reference-docs.edgecompute.app/index.html#Request) and solved it.

2. I added `Vercel`
[Addressing the runtime](https://edge-runtime.vercel.sh/features/available-apis#addressing-the-runtime) from official doc.

yusukebe:
@yujong-lee 

Adding "fastly" and "vercel" is good!

> So I started to think that c.runtime is better than c.platform.

Agree. I think `c.runtime` is better too.

> If we agree to do so, we need to decide how to name return values.

I think the name should be short and all lowercase because it can be just a "short name". So, I want to go with the following:

```ts
type runtime = 'node' | 'deno' | 'bun' | 'vercel' | 'cloudflare' | 'fastly' | 'other'
```

For example. It is unlikely that Cloudflare will have a Runtime other than "Cloudflare Workers" in he future. So, it's ok to be `cloudflare`. Of course, Cloudflare could have a different Runtime than Workers for Cloudflare Pages. However, that is a rare case, and in that case, we can add cloudflare-pages.

yujong-lee:
@yusukebe Ok. I agree.

 I changed the code like below.
```ts
get runtime(): Runtime
type Runtime = 'node' | 'deno' | 'bun' | 'vercel' | 'cloudflare' | 'fastly' | 'other'
```

yusukebe:
Thank you! I'll merge it!

PR #650, opened 2022-11-03
feat(app): introduce `app.showRoutes()`

yusukebe:
This PR includes a new feature `app.showRoutes()`.

Sometimes it is difficult to see what routes are registered. I've created this feature to prevent us from losing time on that. Using `app.showRoutes`, will show the routes that are registered:

SS


yujong-lee:
This look nice. Will this be called by default when the app starts?

PR #651, opened 2022-11-03
fix(context): check cloudflare before checking node in `c.runtime`

yusukebe:
In my environment the test in `context.test.ts` fails:

SS

I don't know why it is be passed on the CI, but I think we have to fix the test and `contex.ts`.

Actually, the test environment with `miniflare` is not only "node" but also "cloudflare". There are both `global.process.title` and `global.WebSocketPair` in my environment.

SS

In this case, we have to check first whether it is `cloudflare` or not and check it later for `node`.

### DIFF
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -198,10 +198,6 @@ export class Context<
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const global = globalThis as any
 
-    if (global?.process?.title === 'node') {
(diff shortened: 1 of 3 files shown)


yujong-lee:
I was running `npx jest` rather than `yarn test`. So I think `miniflare` env is not applied. I have no idea about by CI passed though.

Anyway, I agree with the change you made. Thank you for the quick fix.

PR #653, opened 2022-11-05
feat(app): introduce `app.on(method, path, handler)`

yusukebe:
This PR introduces `app.on()`.

Currently, the HTTP methods we can use are specified in the code:

https://github.com/honojs/hono/blob/fadce526736a610248ec5175ca66ddcc5a84c883/src/router.ts#L3

The `app.on()` function will enable we set the custom HTTP method as we like:

```ts
app.on('PURGE', '/foo', async (c) => {
  await purgeCache('foo')
  return c.redirect('/')
})
```

### DIFF
--- a/deno_dist/hono.ts
+++ b/deno_dist/hono.ts
@@ -140,6 +140,15 @@ export class Hono<
     return this
   }
 
+  on(method: string, path: string, ...handlers: Handler<P, E, S>[]) {
+    if (!method) return this
+    this.path = path
+    handlers.map((handler) => {
+      this.addRoute(method.toUpperCase(), this.path, handler)
+    })
+    return this
+  }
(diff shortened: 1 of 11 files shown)


yusukebe:
Hi @usualoma !

If you can, please review this PR (I think this is a good feature).

ThatOneBro:
Oooh I really like this. I'm going to test it tonight 🔥 Looks like a great feature for extending the language of HTTP. (Purge I think is actually pretty common already.) I approve this feature 👍

usualoma:
@yusukebe 
Yes, I too think it would be a very nice feature!
I think some refactoring (#655) needs to be added to accommodate edge cases, but this PR is a good way to go!

PR #655, opened 2022-11-06
Refactoring of feat/on branche

usualoma:
* Since the first argument of `Router<T>#add` is expected to be uppercase, and existing code will not work correctly if lowercase is passed, I thought `toUpperCase` was unnecessary.
* In StaticRouter, when adding a new method, the middleware in METHOD_ALL must be copied.


yusukebe:
@usualoma 

Thank you!! I'll merge it later.

PR #658, opened 2022-11-10
fix(type): change the type definition of `req` in Context

### LINKED ISSUE #657: relax the constraint to allow custom `Schema` type
The change introduced in the https://github.com/honojs/hono/pull/641 currently breaks my existing code, as I wrote a custom request validator instead of using the built-in one.

More specifically, I'm suggesting changing the

https://github.com/honojs/hono/blob/b82547f70af539e1ec37393bb3f1c1c3753b9cb3/src/context.ts#L17

into something like

```ts
{
  // ..
  req: Request<P, S extends Schema ? SchemaToProp<S> : S>
}
```

And this would then allow

https://github.com/honojs/hono/blob/b82547f70af539e1ec37393bb3f1c1c3753b9cb3/src/types.ts#L19-L23

the middleware, to be able to pass any typed data to the next handler.

---

P.S., it seems kinda weird that a _middleware_ would require the core code to do extra stuff for it, but anyway, I think it should always be a good idea to allow users to pick up whatever they want to for request deserialization.

yusukebe:
This will fix #657

### DIFF
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -12,9 +12,9 @@ export type Data = string | ArrayBuffer | ReadableStream
 export class Context<
   P extends string = string,
   E extends Partial<Environment> = Environment,
-  S extends Partial<Schema> | unknown = Schema
+  S extends Partial<Schema> = Schema
 > {
-  req: Request<P, S extends Schema ? SchemaToProp<S> : any>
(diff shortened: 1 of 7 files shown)


equt:
LGTM, thx for the quick fix :)

Issue #660, opened 2022-11-11
Sanitizer in Validator middleware does not work


Toanzzz:
I tried to make some params always have lowercase, and figure that using the `addSanitizer` func in the validation pipe would be the most intuitive, except that it doesn't work at all 😅 (tried the `trim` function, but it also does nothing)

You can check the code below to reproduce it:

### Reproduce code:

```ts
import { Hono } from 'https://deno.land/x/hono@v2.5.0/mod.ts'
import { serve } from 'https://deno.land/std@0.163.0/http/server.ts'
import { validator } from 'https://deno.land/x/hono@v2.5.0/middleware.ts'

export const lowCase = (s: unknown) => `${s}`.toLowerCase()

const app = new Hono()
app.post(
  '/json',
  validator(v => ({
    trim: v.json('trim').trim().isOptional(),
    low: v.json('low').addSanitizer(lowCase).isOptional(),
  })),
  ctx => ctx.json(ctx.req.valid())
)

await serve(app.fetch)
```

### Sample request (using [Httpie](https://httpie.io/))

```sh
http POST localhost:8000/json 'trim==    b    ' low==CDF --raw '{
  "trim": "   a   ",
  "low":"ABC"
}'
```

### Expected result

```json
{
  "trim": "a",
  "low": "abc"
}
```

### Actual result 🚫

```json
{
  "trim": "   a   ",
  "low": "ABC"
}
```

yusukebe:
Hi @Toanzzz ! 

Thank you for creating the issue.

Currently, the result values are not sanitized as you said. Some codes have to implement are lacking. I'll try to write them. Please wait.

yusukebe:
Hi @Toanzzz !

This is fixed with #661 by @yujong-lee , we'll release the new version including this fix soon. Thank you!

yusukebe:
It's released! I'll close this issue.

PR #661, opened 2022-11-12
fix(sanitizer)

yujong-lee:
This address #660.

### DIFF
--- a/deno_dist/validator/validator.ts
+++ b/deno_dist/validator/validator.ts
@@ -164,13 +164,6 @@ export abstract class VBase {
     return this
   }
 
-  sanitizeValue = (value: Type) => (
-    this.sanitizers.reduce(
-      (acc, sanitizer) => sanitizer(acc),
-      value
-    )
-  )
-
   message = (text: string) => {
     const len = this.rules.length
     if (len >= 1) {
@@ -279,6 +272,13 @@ export abstract class VBase {
     const prefix = 'should be'
     return this.isArray ? `${prefix} "${this.type}[]"` : `${prefix} "${this.type}"`
   }
+
+  private sanitizeValue = (value: Type) => (
+    this.sanitizers.reduce(
+      (acc, sanitizer) => sanitizer(acc),
+      value
+    )
+  )
 
   private validateRule(rule: Rule, value: Type): ValidateResult {
     let isValid: boolean = false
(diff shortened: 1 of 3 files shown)


yujong-lee:
https://github.com/honojs/hono/blob/d3a6f80d7c70db5e48b3658c6dbbc228c58a3328/src/validator/validator.ts#L249-L257

This part is not called. Maybe we should `addRule` in `addSanitizer`? I am not sure what will be a good fix.

yusukebe:
Hi @yujong-lee !

You may already know this, but sanitizers are currently used only for verification and are not reflected in the verification results. If you are able to correct it, it would be very helpful you could.

yusukebe:
@yujong-lee 

I think that the sanitized value in the `validateValue` is not reflected in the result.

https://github.com/honojs/hono/blob/d3a6f80d7c70db5e48b3658c6dbbc228c58a3328/src/validator/validator.ts#L330

So we need to sanitize it before passing it to `validateValue` or return the value from `validateValue`.

yujong-lee:
@yusukebe Thank you for help! 

Yes, that is something we should implement.

But another point I found is that in my test case, `validateRule` is called only once with `rule.type = 'type'`, not `rule.type = 'value'`.

https://github.com/honojs/hono/blob/d3a6f80d7c70db5e48b3658c6dbbc228c58a3328/src/validator/validator.ts#L274-L283

So `validateValue` is not even called.

Is there something that I missed?

yusukebe:
Ah, yes, you are right! In your test case, the `validateRule` is not called with `rule.type === 'value'`, so the sanitizer is also not called.

I think we don't have to include sanitizers in `validateValue`. For example, how about creating a `sanitizeValue` function and calling it before `validateValue` and not sanitize in the `validateValue`.

deno_dist/validator/validator.ts
```diff
@@ -164,6 +164,13 @@ export abstract class VBase {
     return this
   }
 
+  sanitizeValue = (value: Type) => (
```
yusukebe, line 167: How about making it "private"? I think it's better that users do not use `sanitizeValue` directly. And in that case, it would be better to write this function much later. Like above the `validateRule`.

yusukebe:
@yujong-lee Thanks! I'll merge it.

Issue #671, opened 2022-11-20
New worker- types and ambient types


stefanmaric:
Cloudflare has released a new major version of their worker types: https://blog.cloudflare.com/improving-workers-types/

This new version adds exported types, so using ambient types are not required. See: https://github.com/cloudflare/workers-types/issues/195

Right now Hono pulls `@cloudflare/worker-types@3`'s ambient types into its own types, and since references are added globally in TypeScript's compilation context, consumers of Hono get their context polluted with Cloudflare's types as well.

This is a bit of problem if either your project doesn't target workers as a runtime or if you want to create re-usable Hono middlewares.

Would it be an option for Hono to upgrade to the `v4` of `@cloudflare/worker-types`, use imports types to reference Cloudflare's types, and remove the global Hono `Request` declaration?

The caveat is that this would be a breaking change for those creating handlers/middlewares without explicitly using Hono's Handler/Middleware types. That's it, those setting the types for their functions like:

```typescript
export const myHandler = (ctx: { req: Request }, next: () => Promise<void>): Response => {
  const thing = req.param('thing')
  //----------------^
  // TS would complain here since browser/workers Request doesn't have Hono's methods.
  ...
}
```

I wouldn't expect many to be using this syntax as it is very verbose.

yusukebe:
Hi @stefanmaric !

Thanks for telling us. This is a serious issue. I am thinking about which is the best way while creating prototypes.

PR #673, opened 2022-11-21
refactor: support new `@cloudflare/worker-types`

yusukebe:
The new version of `@cloudflare/worker-types` has been released:

https://blog.cloudflare.com/improving-workers-types/

This PR will support the new version and do some refactorings.

* Fixed the type mismatch of `Request` object that is caused by using the new *workers-types*.
* The new *workers-types* exports some types like `Request`, and we can import and use it. But we will continue to extend `Request` on the global without doing it.
* We no longer use *workers-types* in the core Hono project. Instead, use [lib.webworker.d.ts](https://raw.githubusercontent.com/microsoft/TypeScript/main/lib/lib.webworker.d.ts). This means it does not depend on Cloudflare Workers.
* However, only `KVNamespace` refers to *workers-types*.

If it looks OK, I'll merge it.

Related #671


yusukebe:
By the way, I got this error.

SS

deno_dist/utils/cloudflare.ts
```diff
@@ -1,3 +1,5 @@
+/// <reference types="@cloudflare/workers-types" />
```
stefanmaric, line 1: Wouldn't type import work as well:

```typescript
import type { KVNamespace } from '@cloudflare/workers-types'
```

It would prevent Cloudflare types from being added to the TS compilation context.

src/middleware/serve-static/serve-static.ts
```diff
@@ -1,3 +1,5 @@
+/// <reference types="@cloudflare/workers-types" />
```
stefanmaric, line 1: Same comment here about type imports.

deno_dist/utils/cloudflare.ts
yusukebe: I see. But this file in `deno_dist` is used for Deno. So, it is not needed. I've removed it.

src/middleware/serve-static/serve-static.ts
yusukebe: OK, I've fixed it.

yusukebe:
@stefanmaric 

Thank you for reviewing.

I'll merge it now.

yusukebe:
This closed #671

Issue #668, opened 2022-11-19
Type error when using "moduleResolution": "node16" in tsconfig.


universse:
Hi, TypeSript is throw missing type error with the above config.
```
Could not find a declaration file for module 'hono'. '.../hono@2.5.1/node_modules/hono/dist/cjs/index.js' implicitly has an 'any' type.
Try `npm i --save-dev @types/hono` if it exists or add a new declaration (.d.ts) file containing `declare module 'hono';`ts(7016)
```
Related: https://github.com/microsoft/TypeScript/issues/49160

yusukebe:
Hi @universse !

Thank you for creating the issue. I'll investigate to solve it.

yusukebe:
@universse 

It's been fixed and we've released the new version including this patch.

https://github.com/honojs/hono/releases/tag/v2.5.3

Please try it and feel free to reopen the issue.

koistya:
@yusukebe  `"moduleResolution": "NodeNext"` (or, `"NodeNext"`) doesn't seem to work with `2.6.x`

yusukebe:
Hi @koistya !

I've tried it, but that does not throw the errors. I did:

1. Edit `edge/tscondig.json`. Added `"moduleResolution": "NodeNext"`
2. Update Hono to `2.6.2`.
3. `yarn workspace edge build`

Did I do it wrong?

koistya:
@yusukebe run `yarn tsc --build` or, open `edge/index.ts` — VS Code is not be able to load types from `"hono"`.

```
yarn tsc --build

edge/index.ts:17:19 - error TS7006: Parameter 'ctx' implicitly has an 'any' type.

17 app.get("/echo", (ctx) => {
                     ~~~

edge/index.ts:27:23 - error TS7031: Binding element 'req' implicitly has an 'any' type.

27 app.get("*", async ({ req, executionCtx, env }) => {
                         ~~~

edge/index.ts:27:28 - error TS7031: Binding element 'executionCtx' implicitly has an 'any' type.

27 app.get("*", async ({ req, executionCtx, env }) => {
                              ~~~~~~~~~~~~

edge/index.ts:27:42 - error TS7031: Binding element 'env' implicitly has an 'any' type.

27 app.get("*", async ({ req, executionCtx, env }) => {
                                            ~~~


Found 4 errors.
```

yusukebe:
Thank you. Reproduced. I'll investigate.

yusukebe:
@koistya 

[...]

PR #679, opened 2022-11-26
fix(package.json): add "type": "module"

taishinaritomi:
hono did not work with ESM project.
Added type field to package.json.

```
import { Hono } from "hono";
         ^^^^
SyntaxError: Named export 'Hono' not found. The requested module 'hono' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'hono';
const { Hono } = pkg;
```

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,4 +1,4 @@
-module.exports = {
+export default {
   testMatch: ['**/test/**/*.+(ts|tsx|js)', '**/src/**/(*.)+(spec|test).+(ts|tsx|js)'],
   transform: {
     '^.+\\.(ts|tsx)$': 'ts-jest',
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "2.5.5",
   "description": "Ultrafast web framework for Cloudflare Workers, Deno, and Bun.",
   "main": "dist/cjs/index.js",
+  "type": "module",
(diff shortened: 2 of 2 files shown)


yusukebe:
Hi @taishinaritomi !

This line has been removed in this PR #669 , but you are right, it is needed. 

Thank you. I'll merge it.

yusukebe:
Hi @taishinaritomi !

If you can figure it out, I would like to know.

Since accepting this PR, and adding `"type": "module"`, the errors below occur. If we set `"moduleResolution": "NodeNext"` in `tsconfg.json`, VScode can't load types of Hono.

https://github.com/honojs/hono/issues/668#issuecomment-1365355811

I'm also looking into this, but if you know how to solve it, please let us know how.

taishinaritomi:
I will check now

yusukebe:
Thank you!

taishinaritomi:
We have found ways to improve it, but it's a lot of work!

"tsconfig.json.compilerOptions.moduleResolution = NodeNext" option seems to require that .d.ts files also have the extension

## Before


## After

yusukebe:
Ah, extension! We have a lot of work to do, but it's good to know we've found a solution.
I'll try it soon. Thank you very much!

Issue #677, opened 2022-11-24
Validator: v.array() weird experience


Skye-31:
<details>

```js
import { Hono } from "hono";
import { validator } from "hono/validator";

const app = new Hono();

app.post(
	"/",
	validator(
		(v) => ({
			response: v.object("response", (v1) => ({
				content: v1.json("content").isOptional().isLength({ max: 2000 }),
				embeds: v1
					.array("embeds", (v2) => ({
						title: v2.json("title").isOptional(),
						description: v2.json("description").isOptional(),
						url: v2.json("url").isOptional(),
						timestamp: v2.json("timestamp").isOptional(),
						color: v2.json("color").asNumber().isOptional(),
						author: v2
							.object("embeds.[*].author", (v3) => ({
								name: v3.json("name").isOptional(),
								url: v3.json("url").isOptional(),
								icon_url: v3.json("icon_url").isOptional(),
							}))
							.isOptional(),
						footer: v2
							.object("embeds.[*].footer", (v3) => ({
								text: v3.json("text").isOptional(),
								icon_url: v3.json("icon_url").isOptional(),
							}))
							.isOptional(),
						image: v2
							.object("embeds.[*].image", (v3) => ({
								url: v3.json("url").isOptional(),
							}))
							.isOptional(),
						thumbnail: v2
							.object("embeds.[*].thumbnail", (v3) => ({
								url: v3.json("url").isOptional(),
							}))
							.isOptional(),
						fields: v2
							.array("embeds.[*].fields", (v3) => ({
								name: v3.json("name"),
								value: v3.json("value"),
								inline: v3.json("inline").isOptional().asBoolean(),
							}))
							.isOptional(),
					}))
					.isOptional(),
				flags: v1.json("flags").isOptional().asNumber(),
			})),
		}),
		{
			done: (res) => {
				if (res.hasError) {
					return new Response(res.messages.join("\n"), { status: 400 });
				}
			},
		}
	),
	(c) => {
		return Response.json(c.req.valid());
	}
);

app.notFound(() => {
	return new Response("Not found", { status: 404 });
})

export default app;
```
</details>

Issues:
- [x] 1. I have to use `fields: v2.array("embeds.[*].fields")` despite the fact it's already under `embeds` - if you don't it incorrectly looks under `response`. #683 
- [ ] 2. `v.array()` won't accept an empty array - even with it being marked as isOptional, the validator will error. This also occurs when `null` is provided. #690 
[...]

PR #680, opened 2022-11-27
fix(validator): fix jsons validation after arrays

dario-piotrowicz:
fix the bug of `v.json()` calls made after a `v.array()` to be mistakenly considered as still being inside the array

fixes one of the issues mentioned in #677

____

This is the result of the test I've added before applying the change:
before

(the rest of the checks passed already, but I've included them for completeness and to cover all cases for `v.json()` :slightly_smiling_face: )

### DIFF
--- a/src/validator/validator.test.ts
+++ b/src/validator/validator.test.ts
@@ -730,6 +730,76 @@ describe('Nested objects', () => {
       expect(results[0].message).toBe(`${messages[i]}`)
     }
   })
+
+  it('Should threat `v.json()` as a nested array type only inside an array', async () => {
+    const json = {
+      blog: {
+        totalPosts: '3',
+        posts: [
+          {
+            id: 123,
+            title: 'JavaScript',
+            author: {
+              name: 'foo',
+            },
+          },
+          {
+            id: '456',
+            title: 'Framework',
+            author: {
+              name: 'bar',
+            },
+          },
+          {
+            id: 789,
+            title: 'Hono',
+            author: {
+              name: 'baz',
+            },
+          },
+        ],
+        highlightedPostId: '123',
+      }
+    }
+  
+    const req = new Request('http://localhost/', {
+      method: 'POST',
+      body: JSON.stringify(json),
+    })
+
+    const v = new Validator()
+    const vObject = v.object('blog', v => ({
+      totalPosts: v.json('totalPosts').asNumber(),
+      posts: v.array('posts', v => ({
+        id: v.json('id').asNumber(),
+        title: v.json('title'),
+        author: v.object('author', v => ({
+          name: v.json('name'),
+        }))
+      })),
+      highlightedPostId: v.json('highlightedPostId').asNumber(),
+    }))
+
+    const validators = vObject.getValidators()
+
+    const totalPostsResult = await validators[0].validate(req)
+    expect(totalPostsResult[0].isValid).toBe(false)
+    expect(totalPostsResult[0].message).toBe('Invalid Value [3]: the JSON body "blog.totalPosts" is invalid - should be "number"')
+
+    const postsIdResult = await validators[1].validate(req)
+    expect(postsIdResult[0].isValid).toBe(false)
+    expect(postsIdResult[0].message).toBe('Invalid Value [123, "456", 789]: the JSON body "blog.posts.[*].id" is invalid - should be "number[]"')
+
+    const postsTitleResult = await validators[2].validate(req)
+    expect(postsTitleResult[0].isValid).toBe(true)
+
+    const postAuthorResult = await validators[3].validate(req)
+    expect(postAuthorResult[0].isValid).toBe(true)
+
+    const highlightedPostResult = await validators[4].validate(req)
+    expect(highlightedPostResult[0].isValid).toBe(false)
+    expect(highlightedPostResult[0].message).toBe('Invalid Value [123]: the JSON body "blog.highlightedPostId" is invalid - should be "number"')
+  })
 })
 
 describe('Custom message', () => {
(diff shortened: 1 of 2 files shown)


src/validator/validator.test.ts
```diff
@@ -730,6 +730,63 @@ describe('Nested objects', () => {
       expect(results[0].message).toBe(`${messages[i]}`)
     }
   })
+
+  it('Should threat `v.json()` as a nested array type only inside an array', async () => {
+    const json = {
+      blog: {
```
yusukebe, line 736: If we add the `author.name` such as below:

```ts
      blog: {
        totalPosts: '3',
        posts: [
          {
            id: 123,
            title: 'JavaScript',
            author: {
              name: 'foo',
            },
          },
          {
            id: '456',
            title: 'Framework',
            author: {
              name: 'bar',
            },
          },
          {
            id: 789,
            title: 'Hono',
            author: {
              name: 'baz',
            },
          },
        ],
        highlightedPostId: '123',
      },
```

and write the validator below:

```ts
    const v = new Validator()
    const vObject = v.object('blog', (v) => ({
        author: v.object('author', (v) => ({
          name: v.json('name'),
        })),
      })),
    }))
```

then, we expect it to be valid, but it will no be valid. So, I think we have to remove this line:

https://github.com/honojs/hono/blob/8ed7763c38e153ee78860724582ffd0f0d96a845/src/validator/validator.ts#L111

yusukebe:
Hi @dario-piotrowicz !

Thank you for the PR. I'm just trying to fix this issue. I've added the comment, please check it!

src/validator/validator.test.ts
dario-piotrowicz: good point, I guess it was added there to reset the `isArray` field? that shouldn't be necessary anymore right?

dario-piotrowicz:
> Hi @dario-piotrowicz !
> 
> Thank you for the PR. I'm just trying to fix this issue. I've added the comment, please check it!

Cool, good point thanks :slightly_smiling_face:
I've applied your suggested change and updated the unit test, all seems to be working fine :slightly_smiling_face::+1:

src/validator/validator.test.ts
yusukebe: Yes, this line is not needed.

src/validator/validator.ts
```diff
@@ -104,10 +104,10 @@ export class Validator {
     this.isArray = true
     const res = validator(this)
     const arr = new VArray(res, path)
+    this.isArray = false
```
yusukebe, line 107: I reconsidered and think it would be better to create a new instance of `Validator`.

```ts
  array = <T extends Schema>(path: string, validator: (v: Validator) => T): VArray<T> => {
    const newValidator = new Validator()
    newValidator.isArray = true
    const res = validator(newValidator)
    const arr = new VArray(res, path)
    return arr
  }
  object = <T extends Schema>(path: string, validator: (v: Validator) => T): VObject<T> => {
    const newValidator = new Validator()
    newValidator.isArray = this.isArray
    const res = validator(newValidator)
    const obj = new VObject(res, path)
    return obj
  }
```

In this case, a different object from `this` should be invoked in `validator`.

src/validator/validator.test.ts
yusukebe: The tests you write look good!

src/validator/validator.ts
dario-piotrowicz: yes I totally agree!

To be honest I thought of this but I wasn't sure there was a specific reason why the same instance of `Validator` was being used (I guess it is more efficient, although I can't imagine it making a noticeable impact on performances).

I'll make the change :slightly_smiling_face::+1:

dario-piotrowicz: @yusukebe I've changed the implementation, I soft of followed your suggestion but changed things up a bit, I thinks (please let me know if I'm wrong) that `isArray` should not be accessible from the outside of the class (as the user can access that as well).
So I made the `isArray` (I've renamed it to `inArray` as that makes more sense to me, please let me know if you want me to revert that) a private field of the class which you pass in the constructor. We also don't want the user to do that so I renamed the class and removed its export, what we are exporting then is a `Validator` class which wraps the old class but it removes the `inArray` parameter so that the user doesn't see that at all (since it is an implementation details I think).

What do you think? :slightly_smiling_face:

dario-piotrowicz: Actually... I don't think that the `Validator` class is accessible from user code is it? in that case I think that the extra class is not really necessary...

dario-piotrowicz: extra class removed

yusukebe: @dario-piotrowicz 

Renaming `isArray` to `inArray` and changing its accessibility of it are great! The `Validator` class must be accessed from `middleware/validator/middleware.ts`, so I think it's OK to keep it.

I want to go with this current code, OK?

dario-piotrowicz: Yes, if you're ok with it I think it should be good to get merged :smiley:

yusukebe: OK!

In addition, the most important reason to export `Validator` is there are use cases that the developer can write validation schema outside of the `validator` function.

```ts
const schema = (v: Validator) => ({
  title: v.body('title').isRequired(),
  body: v.body('body').isRequired(),
})

app.post('/posts', validator(schema), (c) => {
  const { title, body } = c.req.valid()
  //...
})
```

dario-piotrowicz: ah I see, thanks for clarifying :slightly_smiling_face::+1:

yusukebe (approved):

PR #683, opened 2022-11-27
fix(validator): fix subsequent nested objects names

dario-piotrowicz:
fix the bug of having subsequent objects inside a nested structure causing the validator to forget the previous keys in the scheme's path

fixes point 1 of #677

### DIFF
--- a/src/validator/validator.ts
+++ b/src/validator/validator.ts
@@ -46,8 +46,6 @@ export abstract class VObjectBase<T extends Schema> {
 
   getValidators = (): VBase[] => {
     const validators: VBase[] = []
-    const thisKeys: string[] = []
-    Object.assign(thisKeys, this.keys)
 
     const walk = (container: T, keys: string[], isOptional: boolean) => {
       for (const v of Object.values(container)) {
@@ -55,11 +53,7 @@ export abstract class VObjectBase<T extends Schema> {
           // eslint-disable-next-line @typescript-eslint/ban-ts-comment
           // @ts-ignore
           isOptional ||= v._isOptional
-          keys.push(...v.keys)
-          walk(v.container as T, keys, isOptional)
-          const tmp: string[] = []
-          Object.assign(tmp, thisKeys)
-          keys = tmp
+          walk(v.container as T, [...keys, ...v.keys], isOptional)
         } else if (v instanceof VBase) {
           if (isOptional) v.isOptional()
           v.baseKeys.push(...keys)
--- a/deno_dist/validator/validator.ts
+++ b/deno_dist/validator/validator.ts
@@ -46,8 +46,6 @@ export abstract class VObjectBase<T extends Schema> {
 
   getValidators = (): VBase[] => {
     const validators: VBase[] = []
-    const thisKeys: string[] = []
-    Object.assign(thisKeys, this.keys)
 
     const walk = (container: T, keys: string[], isOptional: boolean) => {
       for (const v of Object.values(container)) {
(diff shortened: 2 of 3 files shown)


src/validator/validator.ts
```diff
@@ -58,7 +56,8 @@ export abstract class VObjectBase<T extends Schema> {
           keys.push(...v.keys)
           walk(v.container as T, keys, isOptional)
           const tmp: string[] = []
-          Object.assign(tmp, thisKeys)
+          const lastObjectInArray = v instanceof VArray && keys[keys.length - 2] === '[*]'
+          Object.assign(tmp, keys.slice(0, keys.length - (lastObjectInArray ? 3 : 1)))
```
dario-piotrowicz, line 60: @yusukebe the issue here was that we were forgetting the keys after finishing the check of an object, for example in:
```ts
    const vObject = v.object('blog', v => ({
      posts: v.array('posts', v => ({
        author: v.object('author', v => ({
          name: v.json('name'),
        })),
        publisher: v.object('publisher', v => ({
          name: v.json('name'),
        }))
      })),
    }))
```
after having walked through the author, the keys would be reset to `['blog']` (instead of `['blog', 'posts']`) so I figured that we shouldn't always use the same `thisKeys` but always use the keys and just push and pop values as needed.

the removal is for 1 element if the last element was not the last in an array (for example in `blog.posts.[*].publisher.name` we want to remove `name`) otherwise it is 3 because we also need to remove the `[*]` and the name of the array (for example in `blog.posts.[*].publisher` we want to remove `publisher`, `[*]` and `posts`) 

What do you think? does this solution make sense to you? :thinking:

yusukebe: Hi. If we fix the positions of the array, it seems the test below failed. Or not?

```ts
describe.only('Nested objects and arrays', () => {
  const blogJson = {
    blog: {
      meta: {
        foo: {
          bar: 'bar',
        },
      },
      posts: [
        {
          author: {
            name: 'foo',
          },
        },
      ],
    },
  }

  it('Should correctly handle values inside objects and arrays', async () => {
    const req = new Request('http://localhost/', {
      method: 'POST',
      body: JSON.stringify(blogJson),
    })

    const v = new Validator()
    const vObject = v.object('blog', (v) => ({
      meta: v.object('meta', (v) => ({
        foo: v.object('foo', (v) => ({
          bar: v.json('bar'),
        })),
      })),
        author: v.object('author', (v) => ({
          name: v.json('name'),
        })),
      })),
    }))
    for (const validator of vObject.getValidators()) {
      const results = await validator.validate(req)
      expect(results[0].isValid).toBe(true)
    }
  })
})
```

dario-piotrowicz: @yusukebe my solution was a bit silly, I've updated it to what I believe to be a correct solution, I'll mark the PR as ready to review, please have a look :slightly_smiling_face:

yusukebe: Wow, cool implementation! Looks good!

yusukebe:
Hi @dario-piotrowicz !

Almost done. There is one thing I would like you to do. Run `yarn denoify` and commit & push the generated code. The files are used for Deno.

yusukebe (approved):

dario-piotrowicz:
> Hi @dario-piotrowicz !
> 
> Almost done. There is one thing I would like you to do. Run `yarn denoify` and commit & push the generated code. The files are used for Deno.

Awesome thanks :smile: 
Done :slightly_smiling_face::+1:

yusukebe:
Thanks! I'll merge it!

PR #684, opened 2022-11-28
fix(reg-exp-router): Allows registering a star after a long prefix.

### LINKED ISSUE #681: Middleware is bypassed when get("*") is used
Hello! I'm not sure if this is expected behavior or not, but I'm having an issue with a middleware not executing for a specific path:

```typescript
app.use("/v1/p/*", async (c, next) => {
  console.log("Middleware called");
  return c.text("OK");
});

app.get("*", async (c) => {
  return c.text("GET called");
})
```

What happens?
Calling `/` -> `GET called`, expected
Calling `/v1/p` -> `GET called`, expected
Calling `/v1/p/test` -> `GET called`, expected value: `OK` + console.log

If I remove the get route, the middleware works for the last two cases

usualoma:
I believe this change will fix #681.

### DIFF
--- a/deno_dist/router/reg-exp-router/router.ts
+++ b/deno_dist/router/reg-exp-router/router.ts
@@ -25,6 +25,8 @@ function buildMatcherFromPreprocessedRoutes<T>(routes: [string, T[]][]): Matcher
     return nullMatcher
   }
 
+  routes = routes.sort(([a], [b]) => a.length - b.length)
+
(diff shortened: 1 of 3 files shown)


yusukebe:
@usualoma 

I have checked, and it works fine. Thank you very much!

One thing, could you do `denoify`?

usualoma:
Ah, I did it again...

Fixed eb7be7d.

yusukebe:
Thanks, I'll merge it!

PR #686, opened 2022-11-29
fix: Cache middleware throws Internal Server Error on Cloudflare Workers

brn:
Cache middleware throws Internal Server Error when return a cached response.
I've investigate some logs but I can't locate the problem.
In this investigation, I found that response wrapped by new Response will resolve problems, so for now I've added code which wrap response with new Response.

And I've added code that will avoid caching response with status >= 400.

### DIFF
--- a/src/middleware/cache/index.ts
+++ b/src/middleware/cache/index.ts
@@ -19,6 +19,9 @@ export const cache = (options: {
     const response = await cache.match(key)
     if (!response) {
       await next()
+      if (!c.res.ok) {
+        return
+      }
       addHeader(c.res)
       const response = c.res.clone()
       if (options.wait) {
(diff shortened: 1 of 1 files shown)


brn:
@yusukebe Sorry I've not testing with minimum samples yet.
So I'll investigate the problem again and rewrite this PR.

dimik:
@brn  Error probably happens when response.body is a readable stream.
 in your handler you should consume body otherwise it won't be cached

```typescript
new Hono().get('...', cache({ cacheName: 'myCache' }), async ctx => {
  const response = await fetch(...);
  const body = await response.arrayBuffer(); // Consume ReadableStream body for caching
  return new Response(body, response);
});

```

@yusukebe i'd drop this cache bcs body should be consumed in a single place

yusukebe:
We can write it shortly:

```ts
return new Response(response.body, response)
```

yusukebe:
This can be closed. Thanks!

PR #690, opened 2022-11-29
Fix Validation of optional empty/undefined arrays

dario-piotrowicz:


src/validator/validator.test.ts
```diff
@@ -372,6 +372,32 @@ describe('Handle array paths', () => {
     expect(results[0].isValid).toBe(true)
     expect(results[0].value).toEqual([false, true, undefined])
   })
+
+  ;[undefined, []].forEach(value => {
```
Skye-31, line 376: Not sure if your `;` is intended at the start here?

dario-piotrowicz: Yeah it is :sweat_smile:, I needed to finish the previous statement, since this codebase doesn't use semicolons if I didn't put the `;` there would be a syntax error, or I could have put the semicolon at the end of line 374

I am not used to not using semicolons so I am not sure if this is preferred in this codebase or if a semicolon on line 374 would be better :thinking: 

In general I __think__ the prefixed semicolon is preferred, for example:

[Screenshot at 2022-11-29 22-14-57](https://blog.izs.me/2010/12/an-open-letter-to-javascript-leaders-regarding/)

(PS: I actually prefer using semicolons at the end of all js/ts statements :cry:)

dario-piotrowicz: Ah yeah there's another prefixed one here: https://github.com/honojs/hono/blob/b963d6e7cbbefd89bc28e12809da0ec271383b7b/src/router/static-router/router.ts#L13
so maybe I did make the right call :smiley:

Skye-31: Interesting - not used to seeing that syntax 😅 
Thanks for enlightening!

yusukebe:
@dario-piotrowicz 

Great work! Can we go with it?

dario-piotrowicz:
@yusukebe thanks :smiley: 

Yeah it should be good to go :slightly_smiling_face:

Issue #688, opened 2022-11-29
`all` will not be called in `RegExpRouter`


yusukebe:
This PR #684 fixed issue #681, thanks to @usualoma . But since this fix, `all` doesn't work the situation below:

```ts
import { RegExpRouter } from '../../../src/router/reg-exp-router'

const app = new Hono({ router: new RegExpRouter() })

app.all('/foo', (c) => {
  return c.text('foo')
})

app.get('/bar', (c) => {
  return c.text('bar')
})
```

```
GET /foo => 404 Not Found // Not expected
```

And the following test will fail:

```ts

describe('ALL and GET', () => {
  const router = new RegExpRouter<string>()

  router.add('ALL', '/foo', 'foo')
  router.add('GET', '/bar', 'bar')

  it('get /foo', () => {
    const res = router.match('GET', '/foo')
    expect(res?.handlers).toEqual(['foo'])
  })
})
```

I'm investigating to fix it.

@usualoma 

I'm sorry. I did not check it carefully enough. If you have time, it would be very helpful if you could take a look too.

yusukebe:
This is fixed by #691

PR #691, opened 2022-11-29
fix(reg-exp-router): Copy from METHOD_NAME_ALL for new methods

usualoma:
Hi @yusukebe !

I don't think #684 is the cause of this problem, as #688 reproduces in v2.5.6 as well.
However, the cause is in a similar area and it is my fault for not noticing this problem when fixing #684.

In the current RegExpRouter architecture, handlers added with METHOD_NAME_ALL are copied to each method, so I believe that copying all existing handlers the first time they (new method) are added will correct the problem, including #681.

### DIFF
--- a/deno_dist/router/reg-exp-router/router.ts
+++ b/deno_dist/router/reg-exp-router/router.ts
@@ -97,7 +97,13 @@ export class RegExpRouter<T> implements Router<T> {
     }
 
     if (/\*$/.test(path)) {
-      middleware[method] ||= {}
+      if (!middleware[method]) {
+        middleware[method] = {}
+        Object.keys(middleware[METHOD_NAME_ALL]).forEach((p) => {
+          middleware[method][p] = [...middleware[METHOD_NAME_ALL][p]]
+        })
+      }
+
       const re = buildWildcardRegExp(path)
       middleware[method][path] ||=
         findMiddleware(middleware[method], path) ||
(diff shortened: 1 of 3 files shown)


yusukebe:
Hi @usualoma !

Great work! It is fixed.

However, this may not be related to the issue. The following test fails:(

```ts
describe('ALL and Star', () => {

  router.add('ALL', '/x', '/x')
  router.add('GET', '*', 'star')

  it('Should return /x and star', async () => {
    const res = router.match('GET', '/x')
    expect(res).not.toBeNull()
    console.log(res?.handlers)
    expect(res?.handlers).toEqual(['/x', 'star'])
  })
})
```

This does not fail with `TrieRouter` and is expected behavior in the specification.
If you can respond, it would be helpful if you could either fix it in this PR or create another one.
(I tried to fix it too, but I haven't gotten to it.)

usualoma:
Indeed...sorry.

I believe d81ed23 is the appropriate fix.

yusukebe:
@usualoma 

Sorry to repeat this. How about the following cases? It may fail.

```ts
router.add('GET', '*', 'star1')
router.add('ALL', '/x', '/x')
router.add('GET', '*', 'star2')
router.add('GET', '*', 'star3')

it('Should return /x and star', async () => {
  const res = router.match('GET', '/x')
  expect(res).not.toBeNull()
  expect(res?.handlers).toEqual(['star1', '/x', 'star2', 'star3'])
})
```

usualoma:
@yusukebe 
How about 5d3adf0!

yusukebe:
Perfect! Thank you! I'll merge it!

Issue #692, opened 2022-11-30
cors middleware vulnerability


dagnelies:
Hi, 

Currently, the cors middleware is vulnerable and has a few culprits.

- the `next` callback is directly invoked, even before the cors checks if the request is allowed or not
- the `next` callback is invoked twice, once for the preflight request and once for the actual request
- by default, all methods are allowed, which differs from the standard which allows only get/head/post by default
- some other defaults are overridden

As a practical example, it would quite be disastrous to first apply `POST /delete-account` before noticing that the origin/method is not allowed. 

Now, this *should* not happen because the preflight request already forbids the browser to send further requests. However, if the preflight request itself triggers some "any method" route (including `OPTIONS`) due to the `next` callback, you may unintentionally trigger the corresponding code instead of having the request blocked.

yusukebe:
This is fixed by #693

PR #693, opened 2022-11-30
Updated CORS

dagnelies:
Do not call `next` for OPTIONS pre-flight requests to avoid side effects and more sensible defaults.

### DIFF
--- a/src/middleware/cors/index.ts
+++ b/src/middleware/cors/index.ts
@@ -12,7 +12,7 @@ type CORSOptions = {
 export const cors = (options?: CORSOptions): MiddlewareHandler => {
   const defaults: CORSOptions = {
     origin: '*',
-    allowMethods: ['GET', 'HEAD'],
+    allowMethods: ['GET', 'HEAD', 'PUT', 'POST', 'DELETE', 'PATCH'],
     allowHeaders: [],
     exposeHeaders: [],
   }
--- a/src/middleware/cors/index.ts
+++ b/src/middleware/cors/index.ts
@@ -13,8 +13,8 @@ export const cors = (options?: CORSOptions): MiddlewareHandler => {
   const defaults: CORSOptions = {
     origin: '*',
     allowMethods: ['GET', 'HEAD'],
-    allowHeaders: ['Content-Type'],
-    exposeHeaders: ['Content-Length'],
+    allowHeaders: [],
+    exposeHeaders: [],
   }
   const opts = {
     ...defaults,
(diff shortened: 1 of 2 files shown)


dagnelies:
Fixes #692

yusukebe:
Hi @dagnelies !

Thank you for the issue and quick fix. It looks good. One thing, could you fix the test to pass the CI?

dagnelies:
Hmmm... replying with the default `Access-Control-Allow-Headers` from the request makes more sense, I restored that. I restored the default empty `exposedHeaders` too. Probably better that way.

src/middleware/cors/index.ts
```diff
@@ -12,7 +12,7 @@ type CORSOptions = {
 export const cors = (options?: CORSOptions): MiddlewareHandler => {
   const defaults: CORSOptions = {
     origin: '*',
-    allowMethods: ['GET', 'HEAD', 'PUT', 'POST', 'DELETE', 'PATCH'],
+    allowMethods: ['GET', 'HEAD'],
```
yusukebe, line 15: I'm reconsidering. How about this line? I think the current implementation is OK.

```ts
['GET', 'HEAD', 'PUT', 'POST', 'DELETE', 'PATCH']
```

Indeed, it is more secure to limit `GET` and `HEAD`, but this is breaking changes. And it may be common to allow the above six methods by default. For example, Express cors implementation:

https://github.com/expressjs/cors/blob/f038e7722838fd83935674aa8c5bf452766741fb/lib/index.js#L10

yusukebe:
@dagnelies 

I've added the comment. Please check it.

src/middleware/cors/index.ts
dagnelies: You are right, it would be a breaking change.

The most important part IMHO is not calling the `next` during the pre-flight request anyway.

dagnelies: While leaving it like this is good to avoid a breaking change, I think it is imperative to indicate in the docs that the allowed HTTP methods are "all" by default.

yusukebe:
@dagnelies 

Thanks! Finally, could you run `yarn denoify` and commit&push the generated file? It will be used for Deno. Please!

dagnelies:
Hmmmm ....I guess I did something wrong ...It generated all test files in `deno_dist` too. 🤔 I Picked it manually.

yusukebe:
> It generated all test files in deno_dist too.

Weird... It should generate only `deno_dist/middleware/cors/index.ts`.

yusukebe:
Complete! Thanks a lot! I'll merge it.

PR #694, opened 2022-11-30
chore: fix lint warnings

dario-piotrowicz:
This fixes all the lint warnings we get in PRs:
Screenshot at 2022-11-30 20-07-23

### DIFF
--- a/src/types.test.ts
+++ b/src/types.test.ts
@@ -27,6 +27,7 @@ describe('Test types of CustomHandler', () => {
   test('No arguments', async () => {
     const handler: Handler = (c) => {
       const data = c.req.valid()
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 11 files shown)


src/types.test.ts
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
dario-piotrowicz, line 2: @yusukebe in this file there are lots of unused variables and incorrect types, I wasn't sure if there is a specific reason for all of this so I just added this eslint-disable comment

If there isn't a specific reason for that let me know and I can do some cleanup here

yusukebe:
Hi @dario-piotrowicz 

Wow great!!! We ignored it, but I was thinking it was not good.
Thank you. I'll merge it.

PR #696, opened 2022-12-01
Fix detecting node runtime

### LINKED ISSUE #685: `c.runtime === 'node'` does not work well
`c.runtime === 'node'` does not work well. It is because the `global.process.title` is such as`/Users/yusuke/.n/bin/node` in my environment. We have to detect which is `node` or not in another way.

yujong-lee:
This resolves #685.

### DIFF
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -219,7 +219,7 @@ export class Context<
       return 'vercel'
     }
 
-    if ((typeof process !== 'undefined') && (process.release.name === 'node')) {
+    if (global?.process?.release?.name === 'node') {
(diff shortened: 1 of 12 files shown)


yusukebe:
Hi @yujong-lee !

```ts
    if (typeof global?.EdgeRuntime !== 'string') {
       return 'vercel'
     }
```

I think this should be:

```ts
if (typeof global?.EdgeRuntime === 'string') {
```

yujong-lee:
@yusukebe  Good catch. Thank you.

yusukebe:
It works fine in my environment. I'll merge it!

PR #700, opened 2022-12-04
feat(handler): "schema" allow normal types

### LINKED ISSUE #689: Handler `Schema` does not allow normal types
Middleware must extend the validator Schema type to provide type safety for middleware specific properties or data. This requires middleware to map normal types to the Hono validator types:
https://github.com/honojs/hono/blob/07efdb5103c4cdb37585724945e26d93f33de383/src/validator/schema.ts#L13-L24

For instance, the following middleware results in a type error.
```typescript
const validateMe = () => {
  const handler: MiddlewareHandler<
    string,
    { Bindings: Bindings },
    { post: { id: string } } // <--- Type '{ post: { id: string; }; }' does not satisfy the constraint 'Partial<Schema>'.
> = async (c, next) => {
   // validation ... 
    await next();
  };
  return handler;
};
```

If one were to implement a Zod based validator one would need to map to the validator specific types which seems unnecessary.

yusukebe:
This PR allows normal types, instead of only allowing `Schema` in the Handler.

```ts
type User = {
  id: string
  name: string
  age: number
}

const handler: Handler<User> = (c) => {
  const user = c.req.valid() // <--- user is User
  // ...
}
```

This change also makes it possible to use the 3rd party library such as "[Zod](https://zod.dev)" for validation.

```ts
import { z } from 'zod'
import type { ZodType } from 'zod'
import { Hono } from '../../../src/hono'
import type { Handler } from '../../../src/index'

const app = new Hono()

const schema = z.object({
  id: z.string(),
  name: z.string(),
  age: z.number(),
})

// You can write a custom validator with Zod:
const zValidtor = <T extends ZodType>(schema: T): Handler<z.infer<T>> => {
  return async (c, next) => {
    const parsed = schema.safeParse(await c.req.json())
    if (!parsed.success) {
    }
    c.req.valid(parsed.data)
    await next()
  }
}

app.post('/post', zValidtor(schema), (c) => {
  const user = c.req.valid() // <--- user will be `User`
  return c.json(user)
})
```

Builtin Validator is recommended, but if developers want to use Zod, they can.

This may resolve #689

### DIFF
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -13,7 +13,7 @@ export type Data = string | ArrayBuffer | ReadableStream
 export class Context<
   P extends string = string,
   E extends Partial<Environment> = Environment,
-  S extends Partial<Schema> = Schema
+  S = Schema
 > {
   req: Request<unknown, P, S extends Schema ? SchemaToProp<S> : S>
   env: E['Bindings']
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -13,7 +13,8 @@ export type Data = string | ArrayBuffer | ReadableStream
 export class Context<
   P extends string = string,
   E extends Partial<Environment> = Environment,
-  S = Schema
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  S = any
 > {
   req: Request<unknown, P, S extends Schema ? SchemaToProp<S> : S>
   env: E['Bindings']
--- a/deno_dist/hono.ts
+++ b/deno_dist/hono.ts
@@ -10,7 +10,6 @@ import { TrieRouter } from './router/trie-router/index.ts'
 import type { ExecutionContext } from './types.ts'
 import type { Handler, Environment, ParamKeys, ErrorHandler, NotFoundHandler } from './types.ts'
 import { getPathFromURL, mergePath } from './utils/url.ts'
-import type { Schema } from './validator/schema.ts'
 
 interface HandlerInterface<
   P extends string,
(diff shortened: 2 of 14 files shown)


yusukebe:
Hi @usualoma @ThatOneBro and others!

I've made the cool feature. If you have time, could you check this PR and share your opinion?

If there are no problems after sometime, I'll merge it.

aulneau:
this is a great starting point. something I've been wondering is if we should adopt a similar position as tRPC in their api, so rather than this
```ts
import { z } from 'zod'
import type { ZodType } from 'zod'
import { Hono } from '../../../src/hono'
import type { Handler } from '../../../src/index'

const app = new Hono()

const schema = z.object({
    id: z.string(),
    name: z.string(),
    age: z.number(),
})

// You can write a custom validator with Zod:
const zValidtor = <T extends ZodType>(schema: T): Handler<z.infer<T>> => {
    return async (c, next) => {
        const parsed = schema.safeParse(await c.req.json())
        if (!parsed.success) {
        }
        c.req.valid(parsed.data)
        await next()
    }
}

app.post('/post', zValidtor(schema), (c) => {
    const user = c.req.valid() // <--- user will be `User`
    return c.json(user)
})
```
We could do something like:

```ts
import { z } from 'zod'
import type { ZodType } from 'zod'
import { Hono } from '../../../src/hono'
import type { Handler } from '../../../src/index'

const app = new Hono()

const schema = z.object({
    id: z.string(),
    name: z.string(),
    age: z.number(),
})

app.post('/post')
    .input(schema)
    .output(schema)
    .return((c) => {
       // TBD, not usre
    })
```

ThatOneBro:
I like this API a lot. I need to test it some but looks good to me. I just started using zod recently so this is a great change!

usualoma:
This is an awesome PR 👍

ryan-mars:
@yusukebe Apologies for taking so long to review. This resolves #689 exactly as I'd hoped. Thank you!!! Please merge. 

Also, I have no personal need for the following suggestion, so I can't comment. 

```typescript 
app.post('/post')
    .input(schema)
    .output(schema)
    .return((c) => {
       // TBD, not usre
    })
```

yusukebe:
@ryan-mars 

> Also, I have no personal need for the following suggestion, so I can't comment.

This relates to the feature that will be introduced in the future. It has nothing to do with this PR. I won't go into details here, but you might like it.

Okay, I'll merge!

PR #706, opened 2022-12-08
fix(reg-exp-router): bug of "star and all" routing

### LINKED ISSUE #699: Bug for "star and all" in RegExpRouter
This bug was revealed in Issue #687 (Thanks a lot, @dagnelies !)

In RegExpRouter, the following test will fail.

```ts
describe.only('ALL and Star', () => {

  router.add('GET', '*', 'star1')
  router.add('GET', '/a', 'a')
  router.add('ALL', '/b/*', 'star2')
  router.add('GET', '/b/bar', 'bar')

  it.only('Should return star1, star2, and bar', async () => {
    const res = router.match('GET', '/b/bar')
    expect(res).not.toBeNull()
    expect(res?.handlers).toEqual(['star1', 'star2', 'bar']) // <--- ['star1', 'bar']
  })
})
```

"star and all" problems are difficult...

yusukebe:
This PR may fix #699.

### DIFF
--- a/deno_dist/router/reg-exp-router/router.ts
+++ b/deno_dist/router/reg-exp-router/router.ts
@@ -136,11 +136,10 @@ export class RegExpRouter<T> implements Router<T> {
 
       Object.keys(routes).forEach((m) => {
         if (method === METHOD_NAME_ALL || method === m) {
-          routes[m][path] ||= [
-            ...(findMiddleware(middleware[m], path) ||
-              findMiddleware(middleware[METHOD_NAME_ALL], path) ||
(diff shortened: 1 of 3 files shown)


yusukebe:
Hi @usualoma !

I'm trying to fix bug #699, and this PR may fix it. Could you review it?

usualoma:
@yusukebe
I am very sorry. I had missed #699.
I am even more sorry, but I would like to modify it in a slightly different way, so let me create another PR. I will get started on it right away.

My apologies for the time you have spent.

PR #709, opened 2022-12-08
`cookie` should return `string | undefined`

unworthyEnzyme:
if a cookie with a name isn't present, `cookie` function returns undefined although this isn't reflected at the type level.

### DIFF
--- a/deno_dist/request.ts
+++ b/deno_dist/request.ts
@@ -33,7 +33,7 @@ declare global {
       (): Record<string, string>
     }
     cookie: {
-      (name: string): string
+      (name: string): string | undefined
       (): Cookie
     }
     bodyData?: BodyData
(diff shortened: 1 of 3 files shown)


yusukebe:
Hi @unworthyEnzyme !

Thank you for the PR. You are right, it should return `string | undefined`.
I want to merge it, but CI failed. Could you fix the tests? I think we can write like below:

```ts
if (yummyCookie) res.headers.set('Yummy-Cookie', yummyCookie)
```

If we should rewrite the tests, it means "breaking changes". We'll write it down clearly in the release.

unworthyEnzyme:
@yusukebe sorry i should've run the tests before opening a pr. Let me check it out tonight.

unworthyEnzyme:
@yusukebe  updated the test. Hope it works now although i am seeing tracing output after i changed the test, is this intended?

yusukebe:
@unworthyEnzyme 

Thank you for updating. 

> Hope it works now although i am seeing tracing output after i changed the test, is this intended?

Yes, no problem! It has been shown before. I'll merge it!

PR #722, opened 2022-12-13
perf(context): `c.text()` returns a response immediately if header is empty

yusukebe:
`c.text()` returns `Response` immediately if headers are empty.

It does not set `Content-Type` intentionally, but `text/plain` is added automatically, so it is not a breaking change. This is mainly for benchmarking, but benchmarking is important.

`GET /` will become 13% faster on Bun!

SS

### DIFF
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -163,7 +163,14 @@ export class Context<
     return this.newResponse(data, status, headers)
   }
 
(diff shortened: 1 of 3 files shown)


ThatOneBro:
Good work! Looks like a nice perf win

yusukebe:
Alright, I'll merge it.

Issue #712, opened 2022-12-09
`HonoRequest`


yusukebe:
Currently, Hono extends`Request` object to handle requests from a client.

```ts
Request.prototype.param = function (this: Request, key?: string) {
//...
}
```

This is efficient, but it is not welcome to extend the Web Standard's Request object. Because it's "standard". So I think it's good to make `HonoReuqest` for wrapping `Request` object.

```ts
class HonoRequest {
  request: Request
  constructor(request: Request) {
    this.request = request
  }

  header(key: string) {
    return this.request.headers.get(key)
  }
  
  //...
}
```

This may make implementation simpler. It may also improve performance.

But this is breaking changes such `c.req.headers.get()` will be obsolete. So we have to include this change in `v3.0.0`.

PR #732, opened 2022-12-17
POC: Internal HonoRequest class

stefanmaric:
This prevents the prototype pollution of the global Request class and prevents issues with type clashes in TypeScript.

Pending to verify:

- [ ] Compat with the diff runtimes and platforms
- [ ] Performance degradation

I simply extended the Request class, but if it doesn't work in all platforms, there's the alternative to wrap the original request object with a Proxy which shouldn't be problematic, but probbly performance will be lesser of that of the current extend approach.


stefanmaric:
Hey @yusukebe this is a total coincidence. :smile: 

I didn't know about #712 until after I sent this PR, while looking for #671 to reference it in this PR.

I've thought about this issue with the Request pollution since that day and today I was upgrading Hono to take advantage of the new worker-types introduced with #673.

This is my last standing Typescript issues with Hono and Cloudflare Workers. Let me know your thoughts!

yusukebe:
Hi @stefanmaric !

So, I am also considering the same things!

I think the current "extending Request object with prototype" is not a good way, to be honest. Then, there are two ways to create "HonoRequest" in my head:

1. Create `HonoRequest` class that extends `Request`, as you wrote in this PR
2. Create `HonoRequest` as a wrapper class of Request. Proposed in #712 

We need to consider the performance.

It still needs to be verified, but "1" may be slower because of the cost of generating `HonoRequest`.

With the "2" I am considering, there is no performance degradation. However, it would be a "big breaking change" because it would be completely different from the current Request object.

```ts
// Just PoC
class HonoRequest {
  originalRequest: Request
  constructor(request: Request) {
    this.originalRequest = request
  }

  header(key: string) {
    return this.originalRequest.headers.get(key)
  }
  
  //...
}
```

As now, I would go with "2". But I'm still not sure if this is the best. If we do it with "2", we have to include it in the next major version "v3.0.0".

yusukebe:
This is the benchmark result that I ran the "Hello World" application with the current version and this PR.

SS

There may be improvement potential, but it's slow.

stefanmaric:
@yusukebe yeah, I expected it to be slower, not that slow tho. :laughing: 

Could you provide instructions on how to run the benchmarks? - I would like to explore different approaches.

yusukebe:
@stefanmaric 

Yes. Code:

```ts
import { Hono } from '../../src/hono'

const app = new Hono()
app.get('/', (c) => c.text('Hii'))

export default app
```

I'm doing benchmarking on Bun.

```
bun run index.ts
```

And using [bombardier](https://github.com/codesenberg/bombardier).

```
bombardier -d 10s --fasthttp 'http://localhost:3000/'
```

---

I'll create the PR for the "2" pattern to create HonoRequest soon.

stefanmaric:
@yusukebe it turns out using a Proxy is faster than extending `Request` (at least in my testing): https://github.com/stefanmaric/hono/compare/poc/hono-request...stefanmaric:hono:poc/hono-request-proxy

I'm getting a ~10% penalty with Proxies vs the ~40% of the extend `Request` approach. I think this would be acceptable.

yusukebe:
Hi @stefanmaric !

Thank you for trying it. I have created #733 with the implementation "2" pattern and mentioned your implementation also. Please check it out.

stefanmaric:
Niiice @yusukebe! I actually prefer the approach in #733 but didn't know breaking changes were an option. Will follow up in the other PR.

yusukebe:
This is no longer necessary. Closing.

Issue #727, opened 2022-12-14
[Proposal] RPC


yusukebe:
I'll introduce an ultra cool feature. It will improve our developer experience. This issue is a continuation of #602.

"[tRPC](https://trpc.io)" is helpful for sharing the types between the server and the client. This means it makes the types an API specification. There is a request to make tRPC adapter #582 , I think it's good to make the adapter as 3rd party middleware.

So, tPRC is great, but we can do it in a better way with Hono because we can make it "integrated".

What we want to do are below:

* Share the request format inferred from the validation schema.
* Share the response format inferred from JSON objects that is returned in a handler.
* Implement "client" not only on the server-side.

Then, I have created those prototypes. They are not fully implemented, but they work. See the screencast:

![output](https://user-images.githubusercontent.com/10682/207732300-2e30a611-cead-4bee-8dd5-2ebdbbe732e6.gif)

Server-side code:

```ts
const api = new Hono()

const getRoute = api
  .get('/hello', (c) => {
    return c.jsonT({
      messages: 'hello',
    })
  })
  .build()

const postRoute = api
  .post(
    '/posts',
    validator((v) => ({
      id: v.json('id'),
      title: v.json('title'),
    })),
    (c) => {
      const post = c.req.valid()
      return c.jsonT({ title: post.title })
    }
  )
  .build()

export type AppType = typeof postRoute & typeof getRoute
```

Client-side code:

```ts
import type { AppType } from './server'

const client = new Client<AppType>('http://127.0.0.1:8787/api')

const res = await client.get('/hello')

console.log(res.messages)

const res2 = await client.json('/posts', {
  id: '123',
  title: 'Hello!'
})

console.log(res2.title)
```

The client-side needs to be considered, but this is a good idea for the server-side API. And, if developers want to use 3rd party validator like a "Zod", they may use it.

Notes, we need to brush up on our data structures to handle types with Generics.

Isn't it cool? I may create the PR later.

cleaton:
Some thoughts I’ve had wile working with my “type-knit” project that might be relevant here as well:

1. Cloudflare Durable Objects & services are exposed as stubs with a “fetch” method. This stub could have a client wrapper (that does not need url) to be used on workers (micro service RPC)
2. It might be best if the client returns a composable type instead of directly retiring the api return type. Ex Result<T> . This way response object could be passed from DO to worker to web client without serializing in worker. Ex:
‘’’worker
return DoClient(dostub).json(“/increment”, {id: 1234})

[...]

PR #733, opened 2022-12-18
feat: introduce HonoRequest with "wrapper pattern"

yusukebe:
This PR introduces "HonoRequest" class to handle a request object with a "wrapper pattern". This is told in #712 . For the same purpose @stefanmaric is implementing "HonoRequest" with "proxy pattern". #727 .

The wrapper pattern has two features.

1. There are no "big" breaking changes, but access to environment-specific properties such as `c.req.cf` must be changed to `c.req.original.cf`.
2. There is no performance degradation with this change. It may even improve it.

Detail. First, HonoRequest will no longer be compatible with `Request`. So, if we want to use the raw `Request` object, we should write like this:

```ts
const cacheKey = c.req.original
```

But, all standard methods and properties of `Request` are ported. We can use `c.req` the same way we have always used it.

```ts
const value = c.req.header('key')
const value2 = c.req.headers.get('key2')
const url = c.req.url
const method = c.req.method
```

If we give enough notice, it could be released as a minor update such "v2.7.0" instead of a major update "v3.0.0".

Second, about performance. Using the implementation in #732 with "proxy pattern", we can not avoid that it will be slower.

https://github.com/stefanmaric/hono/compare/poc/hono-request...stefanmaric:hono:poc/hono-request-proxy

This is the benchmark score with [the scripts](https://github.com/SaltyAom/bun-http-framework-benchmark):

SS

For Hono, being fast is of great importance. So, I believe that the implementation in this PR is an advantage over #732 (However, I am very thankful to @stefanmaric).

---

This matter is still open for discussion and the PR will not be merged immediately.

### DIFF
--- a/deno_dist/request.ts
+++ b/deno_dist/request.ts
@@ -9,7 +9,7 @@ export class HonoRequest<
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   Data = any
 > {
-  original: Request
+  raw: Request
 
   private paramData: Record<string, string> | undefined
   private headerData: Record<string, string> | undefined
@@ -20,7 +20,7 @@ export class HonoRequest<
   private data: Data | undefined
 
   constructor(request: Request, paramData?: Record<string, string> | undefined) {
-    this.original = request
+    this.raw = request
     this.paramData = paramData
   }
 
@@ -85,7 +85,7 @@ export class HonoRequest<
   header(name?: string) {
     if (!this.headerData) {
       this.headerData = {}
-      this.original.headers.forEach((value, key) => {
+      this.raw.headers.forEach((value, key) => {
         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
         this.headerData![key] = value
       })
@@ -100,7 +100,7 @@ export class HonoRequest<
   cookie(key: string): string | undefined
   cookie(): Cookie
   cookie(key?: string) {
-    const cookie = this.original.headers.get('Cookie')
+    const cookie = this.raw.headers.get('Cookie')
     if (!cookie) return
     const obj = parse(cookie)
     if (key) {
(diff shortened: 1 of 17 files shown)


stefanmaric:
> If we give enough notice, it could be released as a minor update such "v2.7.0" instead of a major update "v3.0.0".

I would vote for making a V3, and apply any other breaking changes that might be useful in the process since, even tho most of the Request methods are available as they were before, referential integrity at runtime has been broken (`instanceof`, `.apply()`, `.call()`, `.bind()`, etc won't work the same) along with type integrity at buid time for those relying on Request's type either implicitly (like `const req = new Request(); req.cookies();`) or explicitly (like `(req: Request) => req.cookies();`).

With almost 3K stars, I would expect at least a handful of Hono users to be surprised by their pipeline being broken or worse by bugs in production after a minor update.

deno_dist/request.ts
```diff
@@ -117,52 +77,129 @@ export function extendRequestPrototype() {
...
+    return await this.original.arrayBuffer()
+  }
+
+  async blob() {
+    return await this.original.blob()
+  }
+
+  async formData() {
+    return await this.original.formData()
+  }
```
stefanmaric, line 151: I would avoid the `await`s here to prevent unnecessary microtasks.

Performance impact is probably negligible because engines can optimize for these scenarios (but still, maybe not all of them), but it also changes the execution flow in subtle ways.

See: https://v8.dev/blog/fast-async

yusukebe:
@stefanmaric 

Thanks for your feedback. You are right, I have reconsidered that we should release it as "v3", I will create a separate branch for "v3" like `next` and merge it there. However, I will leave this PR open a little longer.

cleaton:
What about 'raw' instead of 'original'? It's my understanding usage of original request is discouraged unless necessary to access platform specific things like .cf. For me raw communicates this better and is also much shorter.

deno_dist/request.ts
yusukebe: > I would avoid the awaits here to prevent unnecessary microtasks.

This is good. And we can also avoid the await inside `c.json()`.

yusukebe:
@cleaton

> What about 'raw' instead of 'original'?

Looks good! I'll replace `original` to `raw`.

yusukebe:
Alright, I'll merge this into the `next` branch!

Issue #734, opened 2022-12-18
Lagon runtime support


QuiiBz:
Hello! I would like to discuss how we can land [Lagon](https://github.com/lagonapp/lagon) support on Hono. It's an open-source runtime and platform specifically made for the edge, similar to CF Workers / Deno Deploy, bootstrapped and driven by the community.

Some notes:
- We can detect the runtime (https://github.com/honojs/hono/pull/644) using `!!globalThis.__lagon__`
- I have already created [an issue](https://github.com/lagonapp/lagon/issues/97) on Lagon repo to track the progress
- It might be a good idea to add an example on both [Hono examples repo](https://github.com/honojs/examples) and [Lagon examples directory](https://github.com/lagonapp/lagon/tree/main/examples)
- Lagon implements the native Web APIs and follows WinterCG conventions

You can already test locally using the CLI (`npm i -g @lagon/cli`, and feel free to DM me if you want to try the cloud version), here is an example of the "Hello World" showcased on the website and adapter for Lagon:

```ts
import { Hono } from 'hono'

const app = new Hono()
app.get('/', (c) => c.text('Hello! Hono!'))

export const handler = app.fetch
```

yusukebe:
Hi @QuiiBz !

Ultra cool!! It works on my machine!! Yes, let's go ahead.

<img width="446" alt="SS" src="https://user-images.githubusercontent.com/10682/208305431-75e6f446-1ef6-40da-9dcd-10ff13323638.png">

<img width="287" alt="SS" src="https://user-images.githubusercontent.com/10682/208305433-8482c508-7d16-4c73-87b8-009e30bde713.png">

QuiiBz:
Cool! Do you have a test suite that we could use to run Lagon on? I would happily create a PR at https://github.com/honojs/honojs.dev to add some documentation.

QuiiBz:
Currently, there is no `lagon test` command built-in. We can't use a jest environment, because the code will still run inside Node. I see two options for this:

- Create a `lagon test` command that will wrap your code inside a handler function and run it (but I'm not sure if we can bundle Jest inside a non-node environment)
- Use `lagon dev` and perform HTTP queries on the dev server to check for proper responses (similar to the test in the example mentioned above)

I'll try each built-in middleware and see if there are any issues.

QuiiBz:
Update on middlewares, I've tested most of the feature and middlewares built-in into Hono, using `lagon dev` and running Jest to make HTTP requests on the Lagon dev server:

[...]

PR #742, opened 2022-12-24
feat: Lagon support with runtime detection & tests

QuiiBz:
Closes https://github.com/honojs/hono/issues/734
Closes https://github.com/lagonapp/lagon/issues/97

This PR follows https://github.com/honojs/honojs.dev/pull/27 and adds the following:
- Runtime detection using the `__lagon__` global, returning the `lagon` key as specified in WinterCG Runtime Keys
- `test:lagon` command that runs `lagon_test/index.ts` inside Lagon (using `lagon dev`) and runs `lagon_test/index.test.ts` against it
- Add "Lagon" mentions after mentions of Cloudflare Workers, Deno and Bun

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,5 +53,6 @@ jobs:
         with:
           node-version: 18.x
       - run: yarn install --frozen-lockfile
-      - run: npm i -g @lagon/cli esbuild
+      - run: npm run build
+      - run: npm i -g @lagon/cli@latest esbuild@latest
       - run: npm run test:lagon
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,3 +44,14 @@ jobs:
       - run: curl -fsSL https://bun.sh/install | bash
       - run: echo "${HOME}/.bun/bin" >> $GITHUB_PATH
       - run: npm run test:bun
+
+  lagon:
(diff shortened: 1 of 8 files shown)


yusukebe:
Hi @QuiiBz !

Great!! We can test with `test:lagon`!!

> Add "Lagon" mentions after mentions of Cloudflare Workers, Deno and Bun

This is OK. However, with all due respect, "Lagon" is still in the "Dev" phase and may not specify here in some cases. But, now, it's OK. It is good to support many platforms for Hono.

QuiiBz:
Agree, I can remove those mentions if you want while it's still in dev. I've just released a new version of Lagon CLI to add support for ETag, so I'll remove the `.skip` on the ETag test.

yusukebe:
CI failed. I think you should do a `build` before `jest`.

yusukebe:
> Agree, I can remove those mentions if you want while it's still in dev.

Thanks for understanding!

QuiiBz:
> CI failed. I think you should do a `build` before `jest`.

Yes, I'm just waiting for the latest release of the CLI to push the fix :)

QuiiBz:
Just removed the mentions of Lagon, enabled the ETag test, and added a build step before running the tests. Now the CI should work!

yusukebe:
Thank you! I'll merge it!

PR #743, opened 2022-12-24
feat: new types for "RPC" support & built-in  validator obsolete

yusukebe:
This RP includes big changes. If we accept it, we have to merge it into [next branch](https://github.com/honojs/hono/tree/next), and it will be released in the next major version, "v3".

In this PR, I introduce new types for supporting the "RPC" feature. And make the built-in validator obsolete.

## Built-in validator obsolete

Perhaps this means "heartbroken".  I really like the current Built-in Validator, it can easily handle JSON, Forms, etc., and is easy and intuitive to write. But the more we try to make it better, the more we have to do. It is very difficult to make the Validator correct. So I began to think that there was no need for us to make our own validators. I am very sorry to the contributor who worked with, but with this PR, I have removed the built-in validator completely.

## Recommend using 3rd party validators

So I would recommend using 3rd validators, such [Zod](https://zod.dev), [TypeBox](https://github.com/sinclairzx81/typebox). They provide type safety as well as value verification.

## New types for RPC

On the other hand, I would like to add an "RPC" feature like [tPRC](https://trpc.io). This means that server-side API definitions are shared as "Types" and can be used by the Client.

https://github.com/honojs/hono/issues/727

So, I added new types and methods to provide "RPC" for 3rd-party validators.

The implementation just adds `app.build()` and `c.jsonT()`.  And I just added a new type.

```ts
describe('`c.jsonT()` and `app.build`', () => {
  const app = new Hono()
  const route = app
    .get('/posts', (c) => {
      return c.jsonT({
        id: 123,
        title: 'Hello',
        flag: true,
      })
    })
    .build()

  it('Should return correct types with c.jsonT() and app.build()', () => {
    type ExpectType = {
      output: {
        json: {
          id: number
          title: string
          flag: boolean
        }
      }
    }
    type verify = Expect<Equal<ExpectType, typeof route>>
  })
})
```

Then, we can use these to create middleware to do the following (This middleware and client are not included in this PR.):

https://user-images.githubusercontent.com/10682/209447905-bf1afe60-c0fd-45bb-881a-9cf00a352d08.mov

## Make these under `@hono` namespace

For now, I plan to create an HTTP cline that wraps `fetch` and the validator using "Zod". We will not include these in the core package, but will distribute them under `@hono` namespace. For example, `@hono/client` and `@hono/zod`, will be managed in [the monorepo for middleware](https://github.com/honojs/middleware).

## TODO

If we accept this PR, there are a few things we must do.

* Refactor the code.
* Deprecate the current built-in validator.
* Release the next branch as "pre-release" and test it well.
* Create `@hono/client` and `@hono/zod`.
* Try validators other than "Zod".

---

This is one of the big, painful decisions, but one I want to make in v3.


yusukebe:
I got a better idea. So close this PR and will create another one.

PR #745, opened 2022-12-25
feat: new validator middleware using 3rd-party & current middleware obsolete

yusukebe:
This RP includes big changes. If we accept it, we have to merge it into [next branch](https://github.com/honojs/hono/tree/next), and it will be released in the next major version, "v3".

In this PR, I introduce new validator middleware supporting the "RPC" feature. And make the current built-in validator obsolete.

## Current built-in validator obsolete

Perhaps this means "heartbroken".  I really like the current Built-in Validator, it can easily handle JSON, Forms, etc., and is easy and intuitive to write. But the more we try to make it better, the more we have to do. It is very difficult to make the Validator correct. So I began to think that there was no need for us to make our own validators. I am very sorry to the contributor who worked with, but with this PR, I have removed the current built-in validator completely.

## New "generic" validator middleware for using 3rd-party

So I would recommend using 3rd validators, such [Zod](https://zod.dev), [TypeBox](https://github.com/sinclairzx81/typebox). They provide type safety as well as value verification.

Therefore, I created a thin Validator middleware that can handle such 3rd-party Validators in a generic way. 

## New types for RPC

On the other hand, I would like to add an "RPC" feature like [tPRC](https://trpc.io). This means that server-side API definitions are shared as "Types" and can be used by the Client.

https://github.com/honojs/hono/issues/727

So, I added new types and methods `app.build()` and `c.jsonT()` to provide "RPC" for 3rd-party validators.

## Example

For example, we can write the validator with "Zod" like this:

```ts
const v =
  <T extends ZodSchema>(schema: T) =>
  (value: unknown, c: Context) => {
    const parsed = schema.safeParse(value)
    if (!parsed.success) {
    }
    const data = parsed.data as z.infer<T>
    return data
  }

const api = new Hono()

const postRoute = api
  .post(
    '/author',
    validator(
      'json',
      v(
        z.object({
          name: z.string(),
          age: z.string(),
          nickname: z.string().optional(),
        })
      )
    ),
    (c) => {
      const author = c.req.valid()
      return c.jsonT({
        youare: {
          name: author.name,
          age: author.age,
        },
      })
    }
  )
  .build()

export type AppType = typeof postRoute
```

This is a screencast for the demo. It's a little different from the implementation in this PR, but it works the same.

https://user-images.githubusercontent.com/10682/209447905-bf1afe60-c0fd-45bb-881a-9cf00a352d08.mov

## Make the client

For now, I plan to create an HTTP cline that wraps `fetch`. We will not include this in the core package, but will distribute it under `@hono` namespace `@hono/client`. And will be managed in [the monorepo for middleware](https://github.com/honojs/middleware).

## `@hono/zod` etc.

It is also a good idea to create a module like `@hono/zod` to write it simpler and shorter.

## TODO

If we accept this PR, there are a few things we must do.

* Refactor the code.
* Deprecate the current built-in validator.
* Release the next branch as "pre-release" and test it well.
* Create `@hono/client`.
* Try validators other than "Zod".


ThatOneBro:
Makes sense to me. It was fun to work on the validator but you are right, it's a lot of work to make it work well and zod does a lot of the good work for us. I think this makes sense 👍

yusukebe:
Multiple Validator support and a large refactoring on Type. All done. Merging.

PR #747, opened 2022-12-27
fix: Support tsconfig.json "moduleResolution" : "NodeNext"

taishinaritomi:
To support "moduleResolution" : "NodeNext", output .d.ts to the dist/types directory and add dist/types/package.json { "type": "commonjs" }.

#679 
#668


yusukebe:
Hi @taishinaritomi !

I've checked, and this works well in my environment! It's very helpful. Thank you!

PR #768, opened 2023-01-02
feat(trie-router): `regexp` support path including slashes

yusukebe:
"TrieRouter" did not support regexp routing with slash `/`. This means that it will match paths in the same depth but cannot match paths in a dynamic depth. 

```ts
// GET `/js/chunk/abc.js` is not matched
app.get('/:filename{.+.js$}', async (c, next) => {
  console.log(`You access ${c.req.param('filename')}`)
  await next()
})
```

In this PR, make it supports slash.

Will match `/js/chunk/abc.js` not only `/abc.js`.

P.S.

Currently, "RegExpRouter" also does **not** support regexp routing with slashes. It would be good to support it.

### DIFF
--- a/deno_dist/router/trie-router/node.ts
+++ b/deno_dist/router/trie-router/node.ts
@@ -172,6 +172,15 @@ export class Node<T> {
           // Named match
           // `/posts/:id` => match /posts/123
           const [key, name, matcher] = pattern
+
+          // `/js/:filename{[a-z]+.js}` => match /js/chunk/123.js
+          const restPathString = parts.slice(i).join('')
+          if (matcher instanceof RegExp && matcher.test(restPathString)) {
+            handlerSets.push(...this.getHandlerSets(node.children[key], method))
+            params[name] = restPathString
+            continue
+          }
+
           if (matcher === true || (matcher instanceof RegExp && matcher.test(part))) {
             if (typeof key === 'string') {
               if (isLast === true) {
--- a/src/router/trie-router/node.test.ts
+++ b/src/router/trie-router/node.test.ts
@@ -359,6 +359,30 @@ describe('Multi match', () => {
       expect(res?.handlers).toEqual(['Middleware A', 'Middleware B'])
     })
   })
+
+  describe('Including slashes', () => {
+    const node = new Node()
+    node.insert('get', '/js/:filename{[a-z0-9]+.js}', 'any file')
+    node.insert('get', '/js/main.js', 'main.js')
+
+    it('get /js/main.js', () => {
+      const res = node.search('get', '/js/main.js')
+      expect(res).not.toBeNull()
+      expect(res?.handlers).toEqual(['any file', 'main.js'])
+    })
+
+    it('get /js/chunk/123.js', () => {
+      const res = node.search('get', '/js/chunk/123.js')
+      expect(res).not.toBeNull()
+      expect(res?.handlers).toEqual(['any file'])
+    })
+
+    it('get /js/chunk/nest/123.js', () => {
+      const res = node.search('get', '/js/chunk/nest/123.js')
+      expect(res).not.toBeNull()
+      expect(res?.handlers).toEqual(['any file'])
+    })
+  })
 })
 
 describe('Duplicate param name', () => {
--- a/src/router/trie-router/node.test.ts
+++ b/src/router/trie-router/node.test.ts
@@ -362,25 +362,46 @@ describe('Multi match', () => {
 
   describe('Including slashes', () => {
     const node = new Node()
-    node.insert('get', '/js/:filename{[a-z0-9]+.js}', 'any file')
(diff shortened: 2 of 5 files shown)


yusukebe:
Hi @usualoma ! Happy new year!

I think this PR is good, will merge it. And, I hope RegExpRouter supports this too. Could you do it?

The tests below fail now. What we have to do is make it pass.

```ts
describe.only('RegExp including slashes', () => {
  router.add('GET', '/js/:filename{[a-z0-9]+.js}', 'js file')

  it('/js/main.js', () => {
    const res = router.match('GET', '/js/main.js')
    expect(res).not.toBeNull()
    expect(res?.handlers).toEqual(['js file'])
  })

  it('/js/chunk/123.js', () => {
    const res = router.match('get', '/js/chunk/123.js')
    expect(res).not.toBeNull()
    expect(res?.handlers).toEqual(['js file'])
  })
})
```

usualoma:
@yusukebe 
Sorry for the delay!

With the current changes, I think it will not pass if the REST API routing is defined as follows.
Also, I think that params should have a value with a slash in it.

```diff
diff --git a/src/router/trie-router/node.test.ts b/src/router/trie-router/node.test.ts
index 2d50d04..f5054ac 100644
--- a/src/router/trie-router/node.test.ts
+++ b/src/router/trie-router/node.test.ts
@@ -362,25 +362,46 @@ describe('Multi match', () => {
 
   describe('Including slashes', () => {
     const node = new Node()
-    node.insert('get', '/js/:filename{[a-z0-9]+.js}', 'any file')
+    node.insert('get', '/js/:filename{[a-z0-9/]+.js}', 'any file')
     node.insert('get', '/js/main.js', 'main.js')
 
     it('get /js/main.js', () => {
       const res = node.search('get', '/js/main.js')
       expect(res).not.toBeNull()
       expect(res?.handlers).toEqual(['any file', 'main.js'])
+      expect(res?.params).toEqual({ filename: 'main.js' })
     })
 
     it('get /js/chunk/123.js', () => {
       const res = node.search('get', '/js/chunk/123.js')
       expect(res).not.toBeNull()
       expect(res?.handlers).toEqual(['any file'])
+      expect(res?.params).toEqual({ filename: 'chunk/main.js' })
     })
 
     it('get /js/chunk/nest/123.js', () => {
       const res = node.search('get', '/js/chunk/nest/123.js')
       expect(res).not.toBeNull()
       expect(res?.handlers).toEqual(['any file'])
+      expect(res?.params).toEqual({ filename: 'chunk/nest/123.js' })
+    })
+  })
+
+  describe('REST API', () => {
+    const node = new Node()
+    node.insert('get', '/users/:username{[a-z]+}', 'profile')
+    node.insert('get', '/users/:username{[a-z]+}/posts', 'posts')
+
+    it('get /users/hono', () => {
+      const res = node.search('get', '/users/hono')
+      expect(res).not.toBeNull()
+      expect(res?.handlers).toEqual(['profile'])
+    })
+
+    it('get /users/hono/posts', () => {
+      const res = node.search('get', '/users/hono/posts')
+      expect(res).not.toBeNull()
+      expect(res?.handlers).toEqual(['posts'])
     })
   })
 })
```

yusukebe:
@usualoma 

You are right! My implementation is not enough. 

I think we have to rewrite `splitPath()`. This causes a huge performance loss. There are some requests for this feature. But we may not have to support this feature, as slowing it down anymore is not a good idea.

What do you think about it?

If we do not so, for example, we can write like this:

```ts
app.get('*', async (c, next) => {
  if (!/.+\.js$/.test(c.req.url)) {
    return await next()
  }
  const mw = cache({ cacheName: 'foo', cacheControl: 'max-age=3600' })
  await mw(c, next)
})
```

yusukebe:
Or another idea, we can add another router that supports all patterns, although it may be slow. Then, users can chose this router if they want to use "regexp with slashes". For instance, it may be interesting to use `URLPattern`.

However, it may not be necessary to go that far.

usualoma:
@yusukebe 
Thanks for the reply!

I think RegExpRouter can handle it without any performance degradation, and TrieRouter can also handle it without that much degradation by adjusting the splitPath() on insert.

I don't have time to work on it today, but I'll try to come up with a plan after tomorrow.

usualoma:
As one idea, how about something like this?
https://github.com/honojs/hono/pull/781

yusukebe:
@usualoma 

> As one idea, how about something like this?

Using `@`! I was also thinking if we can do it with "special character". Looks good! I want to go with it.

> I think RegExpRouter can handle it without any performance degradation

This is good news. I'll be waiting for whenever you will be ready.

yusukebe:
@usualoma 

I want to include this feature into "[next branch](https://github.com/honojs/hono/tree/next)" for the next major version, "v3". If you create the PR implementing it with RegExpRouter, please point to `next` as the base branch.

usualoma:
@yusukebe 
I have created "support path including slashes" feature for RegExpRouter. #789

PR #776, opened 2023-01-04
fix(package.json): typo exports static-router

taishinaritomi:
There was a mistake in #747 , sorry.
I was able to find it in publint.
``npx publint`` or https://publint.dev/hono


yusukebe:
Hi @taishinaritomi !

Thank you!

> npx publint or https://publint.dev/hono

I didn't know `publint`, it was difficult to check `package.json` etc. works well. This is what I want.

I'll merge it.

PR #782, opened 2023-01-06
perf: optimize parsing queries and headers / 19% faster than 2.7.1

yusukebe:
This PR includes many changes for optimization.

* Avoid using `URLSearchParams`.
* Created original function to parse query params.
* Optimize handling headers.
* Reduced the number of headers passed to `Response`.
* Decode URL based on whether it contains `%` or not.
* Pass static value to `split` instead of regexp.
* Use `indexOf` instead of `slice`.

The result, 19% faster than 2.7.1!

SS


yusukebe:
This will be merged into the "next" branch.

PR #790, opened 2023-01-08
refactor(utils): splitRoutingPath, allow `@` to be safely used in parameter names.

usualoma:
I noticed that `@` is not part of the path, but is available as part of the parameter name.
I don't expect many people to use @ in parameter names, but I would like to make @ available because fewer restrictions due to internal implementation methods is better.
I don't think it is quintessential to use `\` as a parameter name, so I don't think there is any conflict if it is `@\\{i}`.

### For another discussion

To simplify future discussions, it may be possible to decide in some version that "only `[a-zA-Z_][a-zA-Z0-9_]+` can be used for parameter names".

### DIFF
--- a/src/utils/url.test.ts
+++ b/src/utils/url.test.ts
@@ -38,6 +38,12 @@ describe('url', () => {
 
     ps = splitRoutingPath('/users/:name{[0-9a-zA-Z_-]{3,10}}')
     expect(ps).toStrictEqual(['users', ':name{[0-9a-zA-Z_-]{3,10}}'])
+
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @usualoma !

> but I would like to make @ available because fewer restrictions due to internal implementation methods is better.

I also think it's better and a good implementation!

> To simplify future discussions, it may be possible to decide in some version that "only [a-zA-Z_][a-zA-Z0-9_]+ can be used for parameter names".

Good idea, but if it works, I don't think we need to do it.

---

Thanks a lot!! Some developers want this "regexp with slashes" feature, and they will be pleased. I'll merge all PR related to the topic.

PR #796, opened 2023-01-10
feat: introduce `HTTPException`

yusukebe:
In this PR, I will introduce `HTTPException`.

Until now, we have not dared to create an `Exception` class that extends `Error`. Because we can handle it with the middleware looking up status codes.

```ts
app.post('/validator', async (c, next) => {
  await next()
  if (c.res.status === 400) {
    return c.text('Validation error!', 400)
  }
})
```

But, if 401 error occurs in Basic Auth or Bearer Auth middleware, another middleware cannot handle it because Basic Auth and Bearer Auth middleware immediately returns the Response. We must not fire the next middleware handler for a security issue.

So, when a fatal error occurs, such as an authentication error, we can throw this `HTTPException` and handle it with `app.onError`.

```ts
app.onError((e, c) => {
  if (e instanceof HTTPException && e.status === 401) {
    return c.json(
      {
        message: 'Custom error message!',
      },
      401
    )
  }
  return c.text('Internal Server Error', 500)
})
```

It is not usually used, but it is a necessary feature to customize the error message for such an "Unauthorized" error.

This will be merged into "next" branch.

### DIFF
--- a/deno_dist/hono.ts
+++ b/deno_dist/hono.ts
@@ -17,6 +17,7 @@ import type {
   Environment,
   Route,
 } from './types.ts'
+import { HTTPException } from './utils/http-exception.ts'
 import { getPathFromURL, mergePath } from './utils/url.ts'
 
 type Methods = typeof METHODS[number] | typeof METHOD_NAME_ALL_LOWERCASE
@@ -87,6 +88,9 @@ export class Hono<
   }
 
   private errorHandler: ErrorHandler<E> = (err: Error, c: Context<E>) => {
+    if (err instanceof HTTPException) {
+      return err.getResponse()
+    }
     console.trace(err)
     const message = 'Internal Server Error'
     return c.text(message, 500)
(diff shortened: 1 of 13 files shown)


yusukebe:
Hi @metrue @usualoma @ThatOneBro and others

If you have any opinion about this feature, please share it. I'll merge it after a while.

ThatOneBro:
Looks good to me! I think it's a good change 👍 I thought about something like this before

hclj37:
I'm not sure if `e instanceof UnauthorizedException` is better than `e instanceof HTTPException && e.status === 401`.

Say `HTTPException` is the base `Exception` class, and bundled with some built-in `Exception`s inherited from it.

Reference: https://docs.nestjs.com/exception-filters#built-in-http-exceptions

yusukebe:
Of course, we can create an `UnauthorizedException` etc., but there will be more code. Also, it is more flexible to do it with only `HTTPException`.

yusukebe:
I'll merge.

hclj37:
> 



> Of course, we can create an `UnauthorizedException` etc., but there will be more code. Also, it is more flexible to do it with only `HTTPException`.

Thank you for the explanation.

subhendukundu:
In the latest package, `HTTPException` not exported properly? I am not able to import it. :(

yusukebe:
Hi @subhendukundu !

It's not documented. Can you import it like the following?

```ts
```

PR #805, opened 2023-01-15
remove unnecessary individual strict mode family options

hclj37:
The [TypeScript documentation](https://www.typescriptlang.org/tsconfig#strict) said:

> The `strict` flag enables a wide range of type checking behavior that results in stronger guarantees of program correctness. **Turning this on is equivalent to enabling all of the strict mode family options**, which are outlined below. You can then turn off individual strict mode family checks as needed.

Because there is already a `strict` flag, so `strictPropertyInitialization` and `strictNullChecks` are redundant.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -8,8 +8,6 @@
     "forceConsistentCasingInFileNames": true,
     "strict": true,
     "skipLibCheck": true,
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @hclj37 !

You are right! I'll merge this. Thanks!

PR #807, opened 2023-01-16
feat: [BREAKING] introduce adapters

yusukebe:
This PR introduces "adapters" and includes breaking changes.

For example, if we want to use serve-static middleware, we have to specify which runtime we want to write like this:

```ts
import { serveStatic } from 'hono/serve-static.module' // <--- Cloudflare Workers module-mode
```

This means that the implementation depends on the environment. I believe that runtime-specific implementations are not appropriate as middleware. Implementations that use runtime-specific features should be separated explicitly. And implementations that use "not" Web standard API should not be builtin-middleware.

So, I'll introduce "adapters". With the adapter, we can use "serve statice" with writing like the following:

```ts
import { serveStatice } from 'hono/cloudflare-workers'
```

You can now see that this is a Cloudflare Workers specific feature.

This PR introduces the following four adapters.

* Cloudflare Workers
* Cloudflare Pages
* Bun
* Deno

Cloudflare Workers, Bun, and Deno have the serve-static feature the same as the current. Cloudflare Pages is new! It exports the `handle` function. We can use it within [`/functions` mode](https://developers.cloudflare.com/pages/platform/functions/get-started/).

```ts
import { Hono } from 'hono'
import { handle } from 'hono/cloudflare-pages'

const app = new Hono()

app.get('/api/hello', async (c) => {
  return c.text('Hello Pages!')
})

export const onRequest = handle(app)
```

This will be helpful for developers who want to use Hono in Cloudflare Pages. Related: https://github.com/honojs/hono/discussions/802

This is a big breaking change, but we can do it with the major version for "v3". If this PR is merged, we could test its usability as "RC" for v3.

## Naming

> Shouldn't it be `hono/adapter/cloudflare-workers`?

I know what you want to say. But I prefer short names, and that's what we've tried to keep in Hono. So, it's good to be `hono/cloudflare-workers`.

> `handler` is suitable for "Cloudflare Pages"?

I think it's better. This is a matter of taste, but would not be `serve`. This is because `serve` is used in Deno, and `node-server` is used in the sense of "starting a server". So, I think `handle` is a good choice.

## Notice

However, we should minimize the functionality of the adapters because one of the good things fo Hono is running on many runtimes with a platform-independent implementation.

## UPDATE

Also added "Next.js" adapter. We can write Edge API Routes with this simple syntax:

```ts
import { Hono } from 'hono'
import { handle } from 'hono/nextjs'

export const config = {
  runtime: 'edge',
}

const app = new Hono()

app.get((c) => {
  return c.text('Hello from Hono!')
})

export default handle(app)
```


src/adapter/cloudflare-pages/handler.ts
```diff
@@ -0,0 +1,13 @@
+// @denoify-ignore
+import type { EventContext } from '@cloudflare/workers-types'
+import type { Hono } from '../../hono'
+import type { Env } from '../../types'
+
+export const handle = <E extends Env>(app: Hono<E>) => {
+  return (eventContext: EventContext<{}, string, {}>) => {
```
yudai-nkt, line 7: Shouldn't the type annotation for `eventContext` be `EventContext<E["Bindings"], string, {}>`? I'm not familiar with what `Env["Variables"]` is meant for, so `E["Bindings"]` is not the correct type argument either. Don't care about this suggestion if I'm wrong.

yusukebe: Thank you for the comment! I think `<{}, string, {}>` should be kept. This is OK because `Env` in Pages and `E['Bindings']`are different things, and users do not touch this part.

yudai-nkt: OK, then let's keep that part as is!

yusukebe:
Okay, I'll merge it.

PR #824, opened 2023-01-19
build: validate package exports after build

yudai-nkt:
This PR checks if Hono has proper package exports using publint. See #822 for details.


yusukebe:
Thankyou. I'll merge it.

PR #825, opened 2023-01-20
fix(request): `param()` return `''` instead of `undefined`

### LINKED ISSUE #821: Optional Route parameter has wrong type
Given the following route `/api/animal/:type?`, the `type` parameter is optional and can be undefined. However, the type returned from `req.param().type` is `string` instead of an optional string (i.e `type?: string`)

yusukebe:
This will fix #821 and includes a "little" breaking change.

If define the route with an optional parameter:

```ts
  app.get('/api/animal/:type?', (c) => {
    const type = c.req.param('type')
    return c.json({
      type: type,
    })
  })
```

And then get access to `/api/animal`. Currently, the `type` will be `undefined`.

In this PR, that will be `''` - blank string.

Does `type` return `string` or `string | undefined`? was ambiguous, so we decided always to return `string` in this PR.

### DIFF
--- a/deno_dist/request.ts
+++ b/deno_dist/request.ts
@@ -58,7 +58,7 @@ export function extendRequestPrototype() {
     if (this.paramData) {
       if (key) {
         const param = this.paramData[key]
-        return param ? decodeURIComponent(param) : undefined
+        return param ? decodeURIComponent(param) : ''
       } else {
         const decoded: Record<string, string> = {}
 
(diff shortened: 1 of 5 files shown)


deno_dist/request.ts
```diff
@@ -58,7 +58,7 @@ export function extendRequestPrototype() {
     if (this.paramData) {
       if (key) {
         const param = this.paramData[key]
-        return param ? decodeURIComponent(param) : undefined
```
pmbanugo, line 61: I think the previous code is fine. We just need to make the type for when it's undefined be `string | undefined` or `string | null`.

PR #826, opened 2023-01-21
fix(request): make optional param type correct

### LINKED ISSUE #821: Optional Route parameter has wrong type

yusukebe:
This PR fix #821 

In the current version, if `/api/animal/:type?` is defined as a route, `type` will be `string` though it is "optional". It's the wrong type. In this PR, it will be `string | undefined`.

SS

I've made PR #825  which makes it a blank string `''` not `string | undefined`. But, after getting advice from @pmbanugo, I rethought it should be `string | undefined`.

### DIFF
--- a/deno_dist/hono.ts
+++ b/deno_dist/hono.ts
@@ -82,10 +82,9 @@ export class Hono<
 
     const allMethods = [...METHODS, METHOD_NAME_ALL_LOWERCASE]
     allMethods.map((method) => {
-      this[method] = <Env extends Environment, Data>(
-        args1: string | Handler<string, Env, Data>,
-        ...args: Handler<string, Env, Data>[]
-      ) => {
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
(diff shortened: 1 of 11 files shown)


yusukebe:
Hi @pmbanugo, how about this change?

pmbanugo:
Yes, this type is what I think is better.  I looked at the new tests and they also certify what I would expect. 👍🏽 

Although there were some @ts-ignore this has caused you 🙈 I'm wondering if there's a different way to implement it to avoid most of those. I don't know the codebase so I have no suggestion for an alternate approach. But it does get the job done 👍🏽

yusukebe:
@pmbanugo 

Thank you for reviewing!

Of course, I think we don't have to use `@ts-ignore`, but the current version has a complicated code about "Types". So we plan to refactor it and make it a bit cleaner in the next major version.

Okay, I'll merge it later!

pmbanugo:
I see. Thanks!

PR #827, opened 2023-01-21
Missing mime types

Hoodgail:
Missing many common mime types such as: GLTF, GLB, MP4, M4P etc


yusukebe:
Hi @Hoodgail !

Thank you for the PR.

It's difficult to accept it because if we do so, the package size will be very large. The one of the core concepts of Hono is to be small. Of course, I also think it's not good there is no `mp4` in the mime type utils, but, I think it is not necessary, such as `oa2: "application/vnd.fujitsu.oasys2"`. The best thing to do is to select only the mime types that we really need carefully.

Hoodgail:
Thank you for your response, makes sense, i don't know what i was thinking adding all of those there, i'll remove the unnecessary ones

Hoodgail:
Done, maybe this change makes more sense

yusukebe:
Thank you! But one more thing. Can you sort them alphabetically?

Hoodgail:
Done, but could you also check out https://github.com/honojs/node-server/pull/18 ?

yusukebe:
Thanks! I'll check it later.

yusukebe:
@Hoodgail 

Could you do `yarn lint:fix` and `yarn denoify`?

tomByrer:
Are the office docs (doc_ xls_ ppt_) really used in apps that often to be hard-coded?  I don't see `.wav` used much anymore for new sites.

Add [`avif`](https://caniuse.com/avif) for certain

Unsure about these media codex, even tough I like them
? [`AV1`](https://caniuse.com/av1)  default for iPhone vids, YouTube starting to serve
? [`jpgxl`](https://caniuse.com/jpegxl) still in beta, pushed by Cloudinary IIRC

Hoodgail:
@tomByrer I think AVI is a good mime to add, i've seen people use it before, But i also added jpegx1 it might be in beta but some people might find it usefull still

yusukebe:
Thanks, @Hoodgail and @tomByrer !

Looks good! I'll merge it now.

PR #829, opened 2023-01-22
perf: use `substring` and `//.test` ...

yusukebe:
This PR makes Hono faster with small improvements.

Note:

* `substring` is faster than `slice`
* `indexOf ...` if faster than `//.test`


yusukebe:
Buns' tests failed may be because of Bun self. I'll merge.

PR #843, opened 2023-01-27
chore: benchmark of routers

yusukebe:
Added benchmark scripts for measuring the speed of HTTP routers on Node.js and Bun.

In conclusion, in my environment, RegExpRouter is the third fastest on Node.js and the first or second fastest on Bun. Unfortunately, TrieRouter is slow (It's unavoidable because it has a sorting process). The difference in results between Bun and Node.js is probably due to Bun's regular expressions being faster.

`@medley/router` used in [Elysia](https://elysiajs.com) and `koa-tree-router` are fast. Our advantage is that the file sizes are smaller than anywhere else. And routings are also flexible for users' use-cases. So it is not a comparison in general, but the result is the result.

## Node.js

```
---------------------------------------------------------------------------- -----------------------------
• short static - GET /user
---------------------------------------------------------------------------- -----------------------------
Hono RegExpRouter                     140.13 ns/iter (127.78 ns … 204.16 ns) 141.62 ns  191.8 ns 204.08 ns
Hono TrieRouter                       212.64 ns/iter (208.33 ns … 287.68 ns) 212.97 ns 268.56 ns 271.18 ns
@medley/router                        102.52 ns/iter   (97.63 ns … 165.8 ns) 105.03 ns 148.52 ns 152.68 ns
find-my-way                            92.51 ns/iter  (85.26 ns … 105.31 ns)  95.32 ns 100.29 ns 102.32 ns
koa-tree-router                        89.93 ns/iter    (83.4 ns … 105.3 ns)  92.71 ns 101.17 ns 101.98 ns
trek-router                           118.23 ns/iter  (109.71 ns … 129.4 ns) 121.31 ns 126.18 ns 127.26 ns
express (WARNING: includes handling)  650.04 ns/iter  (639.19 ns … 708.6 ns) 652.52 ns  708.6 ns  708.6 ns
koa-router                                 2 µs/iter     (1.92 µs … 2.07 µs)   2.04 µs   2.07 µs   2.07 µs

summary for short static - GET /user
  koa-tree-router
   1.03x faster than find-my-way
   1.14x faster than @medley/router
   1.31x faster than trek-router
   1.56x faster than Hono RegExpRouter
   2.36x faster than Hono TrieRouter
   7.23x faster than express (WARNING: includes handling)
   22.23x faster than koa-router
```

```
• static with same radix - GET /user/comments
---------------------------------------------------------------------------- -----------------------------
[...]

PR #845, opened 2023-01-28
feat(reg-exp-router): Lookup static path by using Object

usualoma:
Optimize paths that do not contain variables by doing a lookup via Object.
This optimization will also reduce the size of the regular expression generated when static paths are included, so there will be no negative impact on the overall result.


usualoma:
### TODO

* [ ] Need to check for ambiguous paths. (does not affect performance)

usualoma:
Completed.

* [x] static path ambiguous check
* [x] merge next branch
* [x] denoify

yusukebe:
@usualoma 

Looks good! It becomes really ultra fast! You already know this, but for reference, here are the results in my environment with the changes. It is the fastest not only in "all together" but in most others.

SS

SS

`find-my-way` is used in Fastify, and the other routers are all top-class among the many JavaScript routers. So, winning them means that RegExpRouter is now the fastest router in the JavaScript world!! You can be proud of this work.

Great!! Thanks!!

usualoma:
@yusukebe Thanks!

Thanks to Router::Boom, thanks to hono's project, thanks to the points made by the benchmark script in #843, I think we got this result.
I too am happy to be involved with the hono's RegExpRouter that is the fastest in the world.

PR #850, opened 2023-01-29
Rename runtime keys

### LINKED ISSUE #844: Use Runtime Keys by WinterCG
`c.runtime` returns which runtime is currently running on. The current keys are below:

```ts
type Runtime = 'node' | 'deno' | 'bun' | 'cloudflare' | 'fastly' | 'vercel' | 'lagon' | 'other'
```

Recently the WinterCG is [defining the "Runtime Keys"](https://github.com/wintercg/runtime-keys/pull/2). We can see the draft in the below link:

* <https://runtime-keys.proposal.wintercg.org>

According to this document, there are other names for the keys than the ones we have defined. For example, we defined the Cloudflare Workers as `cloudflare`, but WinterCG named it `workerd`.

Since Hono should be WinterCG compliant, we should use these keys (`workerd` etc.). The only thing is that Fastly has not joined WinterCG, so that should stay as it is.

This change will be a Breaking Change, so let's add it to the [next branch](https://github.com/honojs/hono/tree/next).

yujong-lee:
Resolves #844.


yusukebe:
Hi @yujong-lee !

Thanks. One thing. We have to change the test, failed at this line:

https://github.com/honojs/hono/actions/runs/4034617841/jobs/6935965327#step:7:331

yusukebe:
Thanks, I'll merge it.

PR #853, opened 2023-01-30
feat: rewnew output types

yusukebe:
This PR renews output types emitted from "app".

Before this PR, it will output types with `app.build()` and the client such as `hc` use it. But in this PR, we don't have to use `app.build()`, and emitted types are changed:

For example, write the endpoint with the validator middleware:

```ts
const route = app.post(
  '/api/v2/posts',
  validator(
    'json',
    validatorFunc(
      z.object({
        id: z.number(),
        title: z.string(),
      })
    )
  ),
  (c) => {
    return c.jsonT({
      success: true,
    })
  }
)
```

And we can get the type:

```ts
export type AppType = typeof route
type Schema = ExtractSchema<typeof route>
```

The extracted types are the followings. `Schema` will be:

```ts
type Schema = {
  '/api/v2/posts': {
    $post: {
      input: {
        json: {
          id: number
          title: string
        }
      }
      output: {
        success: boolean
      }
    }
  }
}
```

I think this format is very useful for writing the client.  Just a simple use of this types and the client looks like this.

https://user-images.githubusercontent.com/10682/215485265-a74e85e1-ea8a-44f3-abf3-423b08051222.mov

This client has no implementation but is more type-safe.

I think this is good PR.

BTW:

I am now thinking of rewriting the client and creating it in the core project as `hono/client`. This is because it is easier to handle and develop.

### DIFF
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -1,6 +1,6 @@
 import { HonoRequest } from './request.ts'
 import type { TypeResponse } from './types.ts'
-import type { Env, NotFoundHandler } from './types.ts'
+import type { Env, NotFoundHandler, Input } from './types.ts'
 import type { CookieOptions } from './utils/cookie.ts'
 import { serialize } from './utils/cookie.ts'
 import type { StatusCode } from './utils/http-status.ts'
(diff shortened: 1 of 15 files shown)


yusukebe:
Use Comment Block Generator <https://blocks.jkniest.dev> for formatting the comment blocks.
Pretty good.

Issue #582, opened 2022-10-03
trpc adapter


aulneau:
hi again :)

so in exploring trpc, they have some [adapters](https://github.com/trpc/trpc/tree/next/packages/server/src/adapters), I took a stab at modifying their fetch adapter to work with hono, but I'd love some feedback on if this looks okay:

```ts
import { AnyRouter, inferRouterContext, resolveHTTPResponse } from '@trpc/server';
import { HTTPBaseHandlerOptions, HTTPRequest } from '@trpc/server/dist/http/internals/types';
import type { Context,  } from 'hono';
import { StatusCode } from 'hono/utils/http-status';

export type HonoCreateContextFn<TRouter extends AnyRouter> = (opts: {
  ctx: Context;
}) => inferRouterContext<TRouter> | Promise<inferRouterContext<TRouter>>;

export type HonoCreateContextOption<TRouter extends AnyRouter> =
  unknown extends inferRouterContext<TRouter>
    ? {
        /**
         * @link https://trpc.io/docs/context
         **/
        createContext?: HonoCreateContextFn<TRouter>;
      }
    : {
        /**
         * @link https://trpc.io/docs/context
         **/
        createContext: HonoCreateContextFn<TRouter>;
      };

export type HonoHandlerOptions<TRouter extends AnyRouter> = HTTPBaseHandlerOptions<
  TRouter,
  Request
> &
  HonoCreateContextOption<TRouter>;
export type HonoHandlerRequestOptions<TRouter extends AnyRouter> = {
  ctx: Context;
  endpoint: string;
} & HonoHandlerOptions<TRouter>;

export async function honoRequestHandler<TRouter extends AnyRouter>(
  opts: HonoHandlerRequestOptions<TRouter>
): Promise<Response> {
  const createContext = async () => {
    return opts.createContext?.({ ctx: opts.ctx });
  };

  const url = new URL(opts.ctx.req.url);
  const path = url.pathname.slice(opts.endpoint.length + 1);
  const req: HTTPRequest = {
    query: url.searchParams,
    method: opts.ctx.req.method,
    headers: Object.fromEntries(opts.ctx.req.headers),
    body: await opts.ctx.req.text(),
  };

  const result = await resolveHTTPResponse({
    req,
    createContext,
    path,
    router: opts.router,
    batching: opts.batching,
    responseMeta: opts.responseMeta,
    onError(o) {
      opts?.onError?.({ ...o, req: opts.ctx.req });
    },
  });

[...]

PR #862, opened 2023-02-02
feat: `hc` - Hono Client

yusukebe:
This PR introduces a new feature, "Hono Client". "hc" for the same purpose was built as `@hono/hc`, but making it obsolete, includes it in the core package.

## Usage

First, create the API server. Write the endpoint and export the type of route:

```ts
// server.ts
import { Hono } from 'hono'
import { z } from 'zod'
import { zValidator } from '@hono/zod-validator'

const app = new Hono()

const route = app.post(
  '/api/v2/posts',
  zValidator(
    'json',
    z.object({
      id: z.number(),
      title: z.string(),
    })
  ),
  (c) => {
    const { id, title } = c.req.valid('json')
   // Use `c.jsonT()` to emit types
    return c.jsonT({
      message: `${id} is ${title}`,
      success: true,
    })
  }
)

export type AppType = typeof route // export the type
```

Second, start writing the `client.ts`, import the type from `server.ts` and pass it to the `hc` function as Generics:

```ts

import type { AppType } from './server'

const client = hc<AppType>('http://localhost:8787/')
```

Then, time for magic:

https://user-images.githubusercontent.com/10682/216456881-7c9d4bad-e669-4cfb-a027-86ce98074973.mov

It will suggest the code based on the path and schema of the API server endpoint.

And handle the response:

https://user-images.githubusercontent.com/10682/216457919-1484ed48-b1d3-4428-b6be-565f7a9ee879.mov

`data` has the appropriate type. Great!

The final code is here:

```ts
const res = await client.api.v2.posts.$post({
  json: {
    id: 123,
    title: 'Hello!!',
  },
})

const data = await res.json()

console.log(data.message)
```

### Don't you use `@hono/hc`?

Separating core packages and `hc` is good, but it makes development difficult. For example, if we want to refer to the same `Hono` or `ValidationTypes` types, but the projects are separated, it is difficult to do so, and type errors occur frequently. Even if we include it in the core package, it is independent of `hono`, so the bundle size will not change unless we use it. In addition, it is very small and does not depend on an external library.

### DIFF
--- a/deno_dist/client/client.ts
+++ b/deno_dist/client/client.ts
@@ -1,0 +1,122 @@
+import type { Hono } from '../hono.ts'
+import type { ValidationTypes } from '../types.ts'
+import type { UnionToIntersection } from '../utils/types.ts'
+import type { Callback, Client, RequestOption } from './types.ts'
+import { replaceUrlParam, mergePath, removeIndexString } from './utils.ts'
+
+const createProxy = (callback: Callback, path: string[]) => {
+  const proxy: unknown = new Proxy(() => {}, {
+    get(_obj, key) {
+      if (typeof key !== 'string') return undefined
+      return createProxy(callback, [...path, key])
+    },
+    apply(_1, _2, args) {
+      return callback({
+        path,
+        args,
+      })
+    },
+  })
+  return proxy
+}
+
+class ClientRequestImpl {
+  private url: string
+  private method: string
+  private queryParams: URLSearchParams | undefined = undefined
+  private pathParams: Record<string, string> = {}
+  private rBody: BodyInit | undefined
+  private cType: string | undefined = undefined
+
+  constructor(url: string, method: string) {
+    this.url = url
+    this.method = method
+  }
+  fetch = (
+    args?: ValidationTypes & {
+      param?: Record<string, string>
+    },
+    opt?: RequestOption
+  ) => {
+    if (args) {
+      if (args.query) {
+        for (const [k, v] of Object.entries(args.query)) {
+          this.queryParams ||= new URLSearchParams()
+          this.queryParams.set(k, v)
+        }
+      }
+
+      if (args.queries) {
+        for (const [k, v] of Object.entries(args.queries)) {
+          for (const v2 of v) {
+            this.queryParams ||= new URLSearchParams()
+            this.queryParams.append(k, v2)
+          }
+        }
+      }
+
+      if (args.form) {
+        const form = new FormData()
+        for (const [k, v] of Object.entries(args.form)) {
+          form.append(k, v)
+        }
+        this.rBody = form
+      }
+
+      if (args.json) {
+        this.rBody = JSON.stringify(args.json)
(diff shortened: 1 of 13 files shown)


yusukebe:
Hi, @usualoma @ThatOneBro @cleaton and others!

How about this? You don't have to do a detailed review, I'd like to hear your brief thoughts.

cleaton:
@yusukebe over all I think it looks good, the client api is easy to use and I like that it's possible to pass additional request parameters. 

I think hiding the typed result behind a `json()` method togeter with all the existing `Response` methods makes it difficult to discover (Result has so many methods/attributes...). I would suggest giving another name than `json` maybe `ht()` for `HonoType` or something else. Can also consider using an interface to split the `Result` methods, ex:
```ts
interface HonoResponse<T> {
   response(): Response
   ht(): TypedResponse<T>
```

This kind of interface could also help make the client composable with server endpoints in a way where it's possible to forward responses from nested Hono Servers without unwrapping the response. For example in Cloudflare, have endpoint defined in a worker, inside the endpoint use hono client to call DurableObject instance and directly return the result (making the worker endpoint and DO endpoint return type identical). Current implementation which requires returning `c.jsonT()` requires extracting the DO response in the worker endpoint.

not directly related to this PR, but i happened to see
```ts
 export type TypeResponse<T = unknown> = {
   response: Response | Promise<Response>
   data: T
   format: 'json' // Currently, support only `json` with `c.jsonT()`
 }
 ```
 it does not look like `data` is used anywhere so I think it can be removed (it's already encoded inside `response`). A typescript type can contain type parameters that are not used in any field/value (very useful for passing type info forward without incuring runtime cost of passing the actual object)

**Comment by yusukebe** [comment]
Hi @cleaton ! 

> I think hiding the typed result behind a json() method togeter with all the existing Response methods makes it difficult to discover (Result has so many methods/attributes...).

Is this referring to the client's `res.json()`?

```ts
const res = await client.api.v2.posts.$get()

const data = await res.json() // <--- this?
```

You are right `json()` is a hard keyword to discover. But, this is because I want to keep compatibility with the `Response` object, which is the return value of the `fetch`. I thought about wrapping the `fetch`'s `Response` like `HonoResponse`, but decided against it because it would be too complicated.

ThatOneBro:
@yusukebe I think this is great! Everything looks really nice to me. I haven't used it yet but once I use it I will be sure to give more feedback. For now, I think this is a great decision 😄

usualoma:
@yusukebe
This is some really great magic!
I think this PR is well implemented with lean and less code, and is highly maintainable.

yusukebe:
Hi @cleaton !

I'll merge this PR soon.


This feature - Hono Client - is 
Yoru comment https://github.com/honojs/hono/issues/582#issuecomment-1279598527 
https://github.com/honojs/hono/issues/582#issuecomment-1279598527

cleaton:
@yusukebe thanks for your hard work! Looking forward to using it ~

yusukebe:
I'll merge it now!

PR #863, opened 2023-02-04
Fix 404 error on Serve Static Middleware

BioniCosmos:
There will be a 404 error when non-standard characters appear in URL. Add `decodeURI()` to fix it.


yusukebe:
Hi @BioniCosmos !

Thanks! It's a good fix. I'll merge it.

PR #866, opened 2023-02-06
feat: use added app's errorHandler and notFoundHandler for fallback.

usualoma:
### about `errorHandler`

As for errorHandler, I think it would be the expected result for any app.

### about `notFoundHandler`

As for notFoundHandler, I think most apps will get the expected results, but cb6eff90f99cdce38d69e15dc79a7cacb8b8c812 commitments will lead to unexpected results in some cases.

#### unexpected results ?

```typescript
// prepare
const app = new Hono()
const api = new Hono()
const middleware = new Hono()

api.get('/posts', (c) => c.text('List'))

middleware.use('*', async (c, next) => {
  await next()
  c.res.headers.append('x-custom-b', 'b')
})
middleware.notFound((c) => c.text('Not Found', 404))

app.route('/api', middleware)
app.route('/api', api)

// request
app.request('http://localhost/api/posts') // => not found
```

However, this is a very rare usage, and if you don't add cb6eff90f99cdce38d69e15dc79a7cacb8b8c812, notFound will be ignored, so it's still cb6eff90f99cdce38d69e15dc79a7cacb 8b8c812 should be added, in my opinion.

### DIFF
--- a/deno_dist/hono.ts
+++ b/deno_dist/hono.ts
@@ -61,6 +61,16 @@ interface Route<
   handler: Handler<P, E, S>
 }
 
+const notFoundHandler = <E extends Partial<Environment>>(c: Context<string, E>) => {
+  return c.text('404 Not Found', 404)
+}
+
+const errorHandler = <E extends Partial<Environment>>(err: Error, c: Context<string, E>) => {
+  console.trace(err)
+  const message = 'Internal Server Error'
+  return c.text(message, 500)
+}
+
 export class Hono<
   E extends Partial<Environment> = Environment,
   P extends string = string,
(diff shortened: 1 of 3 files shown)


yusukebe:
Hi @usualoma ! Thank you for the PR.

"errorHandler" is good. It works as expected. Great.

But about "notFoundHanlder", it may have problems. There are situations we want to write followings:

```ts
book.get('/books', (c) => c.text('Books'))
book.notFound((c) => c.text('Book not found', 404))

author.get('/authors', (c) => c.text('Authors'))

app.route('/api', book)
app.route('/api', author)
```

Then the routes will be:

SS

So `GET /api/authos` will be not found.

I have reconsidered that we don't need to support `notFound` in `app.route`.

usualoma:
@yusukebe
Ah, I see, I understand. Thank you.

If we could only target errorHandler, we could write it shorter like #871. How about this one?

yusukebe:
We decided to go with #871 . No longer needed.

yusukebe:
@usualoma !

For #871, as I wrote https://github.com/honojs/hono/pull/871#issuecomment-1422647262, the tests failed in "one handler pattern".

```ts
describe('onError', () => {
  const app = new Hono()
  const sub = new Hono()

  sub.get('/ok', (c) => c.text('OK'))

  sub.get('/error', () => {
    throw new Error('This is Error')
  })

  sub.onError((err, c) => {
    return c.text('onError by sub', 500)
  })

  app.route('/sub', sub)

  it('handled by app', async () => {
    const res = await app.request('https://example.com/sub/ok')
    expect(res.status).toBe(200)
  })
})
```

So I revert the commit of #871 . Then, I want to go with this PR without "notFoundHandler". Sorry for the bother, but can you support this?

usualoma:
OK.
I will start over with another PR.

PR #867, opened 2023-02-06
perf(reg-exp-router): Cache Result<T> instead of T[] if there is no parameter.

usualoma:
I would like to change it to cache the constructed `Result<T>` for static paths.

I could not see a clear change in the benchmark, but I believe there is a slight performance benefit, which can reduce object creation at request time.


yusukebe:
Hi @usualoma !

This seems to be a good change, and there is no side effect. Okay, I'll merge it!

Issue #868, opened 2023-02-06
Remove StaticRouter from default candidates ( or remove StaticRouter completely )


usualoma:
### Reasons for removing StaticRouter from the default list of candidates

With the optimization by #845 #867, I believe RegExpRouter can now perform as well as StaticRouter in a static path only application.

With https://github.com/usualoma/hono/commit/866a84ba2688367cb1b12ce270299971c451c1c6, The result is as follows.

```
$  npm run bench-static:node
...
summary for all together
  Hono RegExpRouter
   1.14x faster than Hono StaticRouter
   2.78x faster than koa-tree-router
   2.97x faster than @medley/router
   4.07x faster than trek-router
   4.24x faster than find-my-way
   8.91x faster than Hono TrieRouter
   34.19x faster than express (WARNING: includes handling)
   75.49x faster than koa-router

$ npm run bench-static:bun
...
summary for all together
  Hono RegExpRouter
   1.16x faster than Hono StaticRouter
   2.86x faster than koa-tree-router
   3.11x faster than @medley/router
   4.27x faster than trek-router
   9.83x faster than Hono TrieRouter
   34.08x faster than express (WARNING: includes handling)
   79.63x faster than koa-router
```

Whether RegExpRouter or StaticRouter is the winner can change if you swap the order in https://github.com/usualoma/hono/commit/866a84ba2688367cb1b12ce270299971c451c1c6#diff-1454672c9673414d6a1726ce738b4f2dddae2f99ad5a806d69185ff7a813f2a1R12-R20. Based on these results, I would consider the two to be of equal performance in static path routing.

The bundle size can be reduced by removing StaticRouter as the default candidates.

### If so, should we even remove StaticRouter from repository?

If an application that uses static paths wants to reduce the size of the application to the extreme, specifying a StaticRouter is an effective way to do so. It is worth leaving it if we recommend this type of use.

```typescript
const app = new Hono({ router: new StaticRouter() })
```

However, since Hono is small enough by default and is almost never used in such a way, I feel that StaticRouter can be removed from the repository.

yudai-nkt:
This might be a bit off-topic, but what are the trie router's advantages over regexp router other than bundle size (2.8kB v.s. 3.9kB minified)? If regexp router is most performant and fattest but we don't seriously care about size, then it seems Hono only needs to have it instead of the other two.

[...]

PR #869, opened 2023-02-06
feat: Remove the StaticRouter from default candidates.

usualoma:
#868


yusukebe:
We will create a separate PR to drop StaticRouter from the package, so I will merge this PR. Thank you.

PR #871, opened 2023-02-07
feat: use added app's errorHandler

usualoma:
Derived from #866, version for errorHandler only

### DIFF
--- a/deno_dist/compose.ts
+++ b/deno_dist/compose.ts
@@ -9,8 +9,8 @@ interface ComposeContext {
 // Based on the code in the MIT licensed `koa-compose` package.
 export const compose = <C extends ComposeContext, E extends Partial<Environment> = Environment>(
   middleware: Function[],
-  onNotFound?: NotFoundHandler<E>,
-  onError?: ErrorHandler<E>
+  onError?: ErrorHandler<E>,
+  onNotFound?: NotFoundHandler<E>
 ) => {
   const middlewareLength = middleware.length
   return (context: C, next?: Function) => {
--- a/deno_dist/hono.ts
+++ b/deno_dist/hono.ts
@@ -124,11 +124,7 @@ export class Hono<
         const handler =
           app.errorHandler === errorHandler
             ? r.handler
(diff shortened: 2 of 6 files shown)


yusukebe:
Hi @usualoma 

Using `compose`! This is a good idea!

It is not we should do usually, but how about changing the order of the `compose` arguments?

```ts
const compose = (
  middleware: Function[],
  onNotFound?: NotFoundHandler,
  onError?: ErrorHandler
) => Function
```

To:

```ts
const compose = (
  middleware: Function[],
  onError?: ErrorHandler, 
  onNotFound?: NotFoundHandler
) => Function
```

By doing so, we can remove the following `notFoundHandler`:

https://github.com/honojs/hono/pull/871/commits/0707d87d0e5b4b662e6b7bf258d5c7f61477e9c4#diff-a6d396782ec1222678901e1d0ac0bf9bb5b63f6560fd26fd81694dc2b4de5a89R129

`compose` is only used in `hono.ts` and it is not difficult to rewrite `compose.test.ts`. I think it's good to keep code short.

usualoma:
> It is not we should do usually, but how about changing the order of the compose arguments?

Yes, I agree.

Swapped and pushed.
ec3a21b

yusukebe:
Thanks! I'll merge it.

yusukebe:
Ah, wait. The following tests "one handler pattern" will fail:

```ts
describe('onError', () => {
  const app = new Hono()
  const sub = new Hono()

  sub.get('/ok', (c) => c.text('OK'))

  sub.get('/error', () => {
    throw new Error('This is Error')
  })

  sub.onError((err, c) => {
    return c.text('onError by sub', 500)
  })

  app.route('/sub', sub)

  it('handled by app', async () => {
    const res = await app.request('https://example.com/sub/ok')
    expect(res.status).toBe(200)
  })
})
```

SS

Using the `compose` is a good idea, but it does not support "one handler pattern".

usualoma:
@yusukebe Oh, sorry, you're right. I'll reconsider.

PR #873, opened 2023-02-08
feat(router) [breaking]: StaticRouter obsolete

### LINKED ISSUE #868: Remove StaticRouter from default candidates ( or remove StaticRouter completely )

yusukebe:
We've discussed in #868, we decided to remove StaticRouter from default candidates and remove it also from the packages. This means StaticRouter will become obsolete. After merging the PR, we can't use StaticRouter.

Thanks to @usualoma for creating StaticRouter. And goodbye StaticRouter.

This will fix #868.


yusukebe:
Hi @usualoma !

If you are OK, I will merge this PR with you as co-author.

usualoma:
@yusukebe Thanks, LGTM!

PR #874, opened 2023-02-08
feat: use added app's errorHandler with comose()

usualoma:
A shorter version that solves the problem of https://github.com/honojs/hono/pull/871#issuecomment-1422647262


yusukebe:
Hi @usualoma

Looks good!

I think it is a matter of taste whether we choose this (`compose()`) or PR #875 (try/catch), but the shorter is better.

Let's go with this.

usualoma:
@yusukebe OK, 076e075 denoified.

yusukebe:
Thanks!

PR #881, opened 2023-02-11
feat(adapter)!: simplify HandleInterface and reduce bundle size

yudai-nkt:
***This PR breaks compatibility with rc.12 but is still compatible with rc.11.*** Since rc.12 is only 9 hours old and adapters aren't in the stable releases yet, I hope this breaking change doesn't have any significant impact.

With this PR, type definitions for `handle` in Cloudflare Pages and Next.js become simpler and parameter names are more descriptive. It also saves the bundle size; minified ESM for Cloudflare Pages adapter is reduced from 825B to 503B and Next.js adapter from 561B to 406B.

Here are the pros/cons I can think of:

<dl>
<dt>Pros</dt>
  <dd>

  - Bundle size is smaller.
  - Interface is simpler.
  </dd>
<dt>Cons</dt>
  <dd>

  - Parameter order is different from `Hono.prototype.route`. Some people might feel awkward.
  - A new Hono app is always instantiated and this has a slight overhead. That said, I think it won't be a problem because:
    - The overhead is negligible compared to other heavy work user's app does.
    - I guess people want to mount `subApp` to a dedicated path like `/api` when they use these adapters (at least I do). So `new Hono()` is called anyway in rc.12 as well.
</dl>

Feel free to close if you disagree to this change.

### DIFF
--- a/src/adapter/cloudflare-pages/handler.test.ts
+++ b/src/adapter/cloudflare-pages/handler.test.ts
@@ -31,7 +31,7 @@ describe('Adapter for Cloudflare Pages', () => {
     app.get('/foo', (c) => {
       return c.text('/api/foo')
     })
-    const handler = handle('/api', app)
+    const handler = handle(app, '/api')
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
     // @ts-ignore
     const res = await handler({ request })
--- a/src/adapter/cloudflare-pages/handler.ts
+++ b/src/adapter/cloudflare-pages/handler.ts
@@ -9,23 +9,14 @@ type EventContext = {
 }
 
 interface HandleInterface {
-  <E extends Env>(app: Hono<E>): (eventContext: EventContext) => Response | Promise<Response>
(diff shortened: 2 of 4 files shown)


yudai-nkt:
By the way, I really appreciate #878. My `functions/api/[[route]].ts` looks like the following and this gets even simpler with the new `handle`, which internally does the same job.

```ts
import { Hono } from "hono";
import { handle } from "hono/cloudflare-pages";

const api = new Hono<{ Bindings: Env }>();
// requests handling here.
const app = new Hono();
app.route("/api", api);

export const onRequest: PagesFunction = handle(app);
```

Thanks for making this feature built-in!

yusukebe:
Hi @yudai-nkt !

You are right! Your way is better.

> Parameter order is different from Hono.prototype.route. Some people might feel awkward.

This was my concern as well, but it is more important to simplify the code and reduce the bundle size.

> By the way, I really appreciate #878.

Glad you liked it :)

Okay, let's go with it.

PR #882, opened 2023-02-11
feat(`app`): support multiple methods with `app.on()`

yusukebe:
This PR enables handling multiple methods with `app.on()` method.

```ts
app.on(['PUT', 'DELETE'], '/posts/:id', (c) => {
  //...
  return c.text(`Your method is ${c.req.method}`)
})
```

### DIFF
--- a/src/hono.ts
+++ b/src/hono.ts
@@ -86,12 +86,21 @@ export class Hono<E extends Env = Env, S = {}> extends defineDynamicClass()<E, S
     })
 
     // Implementation of app.on(method, path, ...handlers[])
-    this.on = (method: string, path: string, ...handlers: H[]) => {
+    this.on = (method: string | string[], path: string, ...handlers: H[]) => {
       if (!method) return this
+      let methods: string[] = []
+      if (Array.isArray(method)) {
+        methods = method
+      } else {
+        methods = [method]
+      }
       this.path = path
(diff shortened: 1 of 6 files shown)


src/hono.ts
```diff
@@ -86,12 +86,21 @@ export class Hono<E extends Env = Env, S = {}> extends defineDynamicClass()<E, S
     })
 
     // Implementation of app.on(method, path, ...handlers[])
-    this.on = (method: string, path: string, ...handlers: H[]) => {
+    this.on = (method: string | string[], path: string, ...handlers: H[]) => {
       if (!method) return this
+      let methods: string[] = []
```
yudai-nkt, line 91: You can do `[method].flat()` to convert a string into a tuple containing it while keeping an array of strings untouched ([example](https://www.typescriptlang.org/play?target=8#code/MYewdgzgLgBAZiEMC8MBECRoNwChSSwBGAhgE4BCJAXijANpqlloA06p1aAurvuBBAAbAKYA6ISADmACnqZuYuEJJQZASnV4Cg0ROlzmVaouWqNW3EA)). I don't know which version is more performant but the performance won't be so different because the length of methods array is very small in this use case.

yusukebe: Thanks! I'll use `flat()`. As for performance, it is not important since it's for just registering routes.

PR #884, opened 2023-02-11
header may return undefined

AlexErrant:
See [Line 94](https://github.com/honojs/hono/pull/884/files#diff-b1a2ab03a3672e105afeaab6615ad625365cd8459d592a840d925f132af62b35R94).

### DIFF
--- a/deno_dist/request.ts
+++ b/deno_dist/request.ts
@@ -36,7 +36,7 @@ declare global {
     }
     headerData?: Record<string, string>
     header: {
-      (name: string): string
+      (name: string): string | undefined
       (): Record<string, string>
     }
     cookie: {
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -1,6 +1,6 @@
 # Contribution Guide
 
-Contributions Welcome! We will be grad for your help.
+Contributions Welcome! We will be glad for your help.
 You can contribute in the following ways.
 
 - Create an Issue - Propose a new feature. Report a bug.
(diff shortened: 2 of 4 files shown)


yusukebe:
Hi @AlexErrant !

Thank you for creating the PR.

You are right, that value could be `undefined`. But we want to return `''` in the case of `undefined`. So, what we should do is not change the return type but assign `''` in the case of `undefined`.

```ts
return this.headerData[name.toLowerCase()] || ''
```

Then the tests don't fail.

AlexErrant:
> But we want to return `''` in the case of `undefined`

I think returning an empty string is incorrect behavior. For example, [consider a custom request header used to prevent CSRF attacks](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#use-of-custom-request-headers):

> ...you can simply verify the presence of this header and value on all your server side AJAX endpoints in order to protect against CSRF attacks

I follow OWASP's advice [here](https://github.com/AlexErrant/Pentive/blob/7cc709f70de45cbfce46c66fe19349a21ce5a3e0/mediaRouter/src/util.ts#L71), where I check for the presence of a custom header - if it doesn't exist, I return early. The value of my custom header [also happens to be an empty string](https://github.com/AlexErrant/Pentive/blob/7cc709f70de45cbfce46c66fe19349a21ce5a3e0/app/src/apiClient.ts#L11). (For the purposes of CSRF, the value of the custom header is irrelevant.)

What you're suggesting has two disadvantages:
1. Making it return `''` for `undefined` is API breaking behavior. If anyone else were to implement CSRF security like me, you would be breaking their security!
2. It's impossible to test for the (non)existence of a header using this function - `''` is a valid header value.

yusukebe:
@AlexErrant 

Thanks for the detailed explanation. I've reconsidered. You are right. "We should not return `''`".

So, I want to accept your PR, but some tests fail because of type mismatches. Could you fix the tests?

yusukebe:
@AlexErrant 

And, please do `yarn denoify`. This is needed to generate the code for Deno.

AlexErrant:
LMK if you'd rather _not_ have that last commit.

yusukebe:
Thanks!

PR #908, opened 2023-02-19
feat(client): Inject common headers

hagishi:
Enabled passing the commonly used site header as an argument during HonoClient instance generation.

### DIFF
--- a/deno_dist/client/client.ts
+++ b/deno_dist/client/client.ts
@@ -1,8 +1,8 @@
 import type { Hono } from '../hono.ts'
 import type { ValidationTargets } from '../types.ts'
 import type { UnionToIntersection } from '../utils/types.ts'
-import type { Callback, Client, RequestOption } from './types.ts'
-import { replaceUrlParam, mergePath, removeIndexString } from './utils.ts'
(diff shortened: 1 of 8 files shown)


yusukebe:
Hi @hagishi !

Thanks. This is a good feature. I'd like to accept and merge it.

One thing. How about making the second argument "RequestOptions" (I think it's better to rename `RequestOption` to `RequestOptions`) ?

```ts
// RequestOption is renamed to `RequestOptions`
hc = <T extends Hono>(baseUrl: string, options?: RequestOptions)
```

Then, we can pass the args.

```ts
const client = hc<AppType>('http://localhost', {
  headers: {
    'x-hono': 'hono',
  },
})
```

This would make it easier to handle if there are more options.

hagishi:
Hi @yusukebe 
We made changes to allow accepting RequestOptions as an argument.

yusukebe:
@hagishi 

Great work! Implementing `deepMerge` is also good.

Finally, could you run `yarn denoify` and commit & push the emitted file? It's needed for Deno.

yusukebe:
Thanks! I'll merge it!

PR #909, opened 2023-02-19
feat(types): keep schema info with `app.route()`

yusukebe:
This PR will fix the matter of #907 .

When using `app.route()`, the schema information is missing but made it so that it is kept.

SS


yudai-nkt:
Thanks! I checked this PR locally and confirmed it works as expected (with multiple `route`s)!

スクリーンショット 2023-02-19 22 08 51

PR #918, opened 2023-02-22
Expose HonoRequest Type

KeKs0r:
Congrats on the v3, launch. Really liking Hono as a clean and fast framework. 

In my apps, I sometimes build utility functions that are used by multiple handlers and they interact with the Request. Since the request has changed to `HonoRequest`, I think it is valuable to also expose its type, so ppl can write functions and define `HonoRequest` as a parameter.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -13,7 +13,6 @@ export type {
 export type { Context, ContextVariableMap } from './context'
 export type { HonoRequest } from './request'
 
-
 declare module './hono' {
   interface Hono {
     fire(): void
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @KeKs0r !

You are right. The type of `HonoRequest` should be exposed.

We'd accept this PR, but there is one request. Edit `src/mod.ts` and export `HonoRequest` like `src/index.ts`. This file is used by Deno.

yusukebe:
@KeKs0r 

I`ll do the `denoify` after I merge this PR! Thanks!

PR #919, opened 2023-02-22
fix(types): `app.route` with multiple endpoints returns types correctly

### LINKED ISSUE #915: Incorrect type inference for `route`ing Hono instance with different `jsonT`-returning endpoints
Suppose we have the following code, which doesn't compile:

```ts
import { Hono } from "hono";

const api = new Hono()
  .get("/foo", (c) => c.jsonT({ foo: "" }))
  .post("/bar", (c) => c.jsonT({ bar: 0 }));
const app = new Hono().route("/api", api);

const client = hc<typeof app>("/");
const res = await client
  .api
  .foo
  .$get();
// ^^^^
// Property '$get' does not exist on type 'ClientRequest<{ $get: { input: {}; output: { foo: string; }; }; } | { $post: { input: {}; output: { bar: number; }; }; }>'.
// Property '$get' does not exist on type 'ClientRequest<{ $post: { input: {}; output: { bar: number; }; }; }>'.ts(2339)
```

In the above case, `app` has a type of 

```ts
Hono<Env, Record<"/api/foo", {
  $get: {
    input: {};
    output: {
      foo: string;
    };
  };
} | {
  $post: {
    input: {};
    output: {
      bar: number;
    };
  };
}> | Record<"/api/bar", {
  $get: {
    input: {};
    output: {
      foo: string;
    };
  };
} | {
  $post: {
    input: {};
    output: {
      bar: number;
    };
  };
}>>
```

which isn't what I want. The expected type should be

```ts
Hono<Env, Record<"/api/foo", {
  $get: {
    input: {};
    output: {
      foo: string;
    };
  };
}> | Record<"/api/bar", {
  $post: {
    input: {};
    output: {
      bar: number;
    };
  };
}>>
```

In terms of unit testing, the following test should pass but it currently fails.

```ts
test('MergeSchemaPath with different endpoints', () => {
  type Sub = Schema<
    'get',
    '/foo',
    {},
    {
      foo: string
    }
  > &
    Schema<
      'get',
      '/bar',
      {},
      {
        bar: number
      }
    >

  type Actual = MergeSchemaPath<Sub, '/api'>

  type Expected = {
    '/api/foo': {
      $get: {
        input: {}
        output: {
          foo: string
        }
      }
    }
  } & {
    '/api/bar': {
      $get: {
        input: {}
        output: {
          bar: number
        }
      }
    }
  }

})
```

My guess is that union distribution is doing something harm. I am trying to fix this but haven't succeeded yet. I'll keep investigation.

yusukebe:
As pointed out in #915, the type was incorrectly inferred when mounting an app with multiple `jsonT()` endpoints using `app.route()`.

SS

This PR will fix #915

### DIFF
--- a/deno_dist/hono.ts
+++ b/deno_dist/hono.ts
@@ -120,9 +120,9 @@ export class Hono<E extends Env = Env, S = {}> extends defineDynamicClass()<E, S
   private notFoundHandler: NotFoundHandler = notFoundHandler
   private errorHandler: ErrorHandler = errorHandler
 
-  route<SubPath extends string, SubSchema>(
+  route<SubPath extends string, SubEnv extends Env, SubSchema>(
     path: SubPath,
-    app?: Hono<E, SubSchema>
+    app?: Hono<SubEnv, SubSchema>
   ): Hono<E, RemoveBlankRecord<MergeSchemaPath<SubSchema, SubPath> | S>> {
     this._tempPath = path
     if (app) {
--- a/deno_dist/hono.ts
+++ b/deno_dist/hono.ts
@@ -83,7 +83,8 @@ export class Hono<E extends Env = Env, S = {}> extends defineDynamicClass()<E, S
             this.addRoute(method, this.path, handler)
           }
         })
-        return this
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        return this as any
       }
     })
 
@@ -96,7 +97,8 @@ export class Hono<E extends Env = Env, S = {}> extends defineDynamicClass()<E, S
           this.addRoute(m.toUpperCase(), this.path, handler)
         })
       }
-      return this
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      return this as any
     }
 
     // Implementation of app.use(...handlers[]) or app.get(path, ...handlers[])
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -330,17 +330,15 @@ describe('Merge path with `app.route()`', () => {
     const client = hc<typeof app>('http://localhost')
 
     it('Should return correct types - GET /api/foo', async () => {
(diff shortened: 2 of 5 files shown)


deno_dist/hono.ts
```diff
@@ -83,7 +83,8 @@ export class Hono<E extends Env = Env, S = {}> extends defineDynamicClass()<E, S
             this.addRoute(method, this.path, handler)
           }
         })
-        return this
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        return this as any
```
yusukebe, line 87: It's difficult to avoid type mismatch in this section, so using `any`, but it's OK because this is the implementation part.

src/client/client.test.ts
```diff
@@ -305,4 +320,27 @@ describe('Merge path with `app.route()`', () => {
...
+  describe('Multiple endpoints', () => {
+    const api = new Hono()
+      .get('/foo', (c) => c.jsonT({ foo: '' }))
+      .post('/bar', (c) => c.jsonT({ bar: 0 }))
+    const app = new Hono().route('/api', api)
+    type AppType = typeof app
+    const client = hc<typeof app>('http://localhost')
+
+    it('Should return correct types - GET /api/foo', async () => {
+      type verify1 = Expect<NotEqual<typeof client.api.foo.$get, any>>
```
yudai-nkt, line 333: Why would you assert that `typeof client.api.foo.$get` is `NotEqual` to `any`? In the next line, you get `res` with the correct type, so I suppose you can compare `typeof client.api.foo.$get` with that type, can't you?

yusukebe: Ah, you are right, it's not necessary. At first, I've been checked only `typeof client.api.foo.$get`, so I didn't write the assert for `typeof data.foo`. But if we assert the `typeof data.foo`, we don't need to check `typeof client.api.foo.$get`.

yusukebe: Removed. Thanks for pointing.

Issue #917, opened 2023-02-21
Fetch a Cloudflare service binding with Hono Client


JonasHiltl:
It would be great if the Hono client could be used to send a fetch request to a service binding. 
With service bindings you have to call the `fetch` method on the binded `Fetcher` like this:
```toml
services = [
  { binding = "auth", service = "authentication" }
]
```
```javascript
export default {
  async fetch(request, env) {
    const authResponse = await env.auth.fetch(request.clone());
    ...
  }
};
```
Since Hono client uses the default `fetch` implementation this does not work, it would be great if we could pass our own fetch function which could default to the normale environment `fetch` implementation and the Hono Client would just call this fetch parameter.

yusukebe:
Hi @JonasHiltl !

I am not familiar with Service Bindings, but it is not necessary to use Hono Client. How about the following code?

```ts
import type { Fetcher } from '@cloudflare/workers-types'
import { Hono } from 'hono'

type Bindings = {
  auth: Fetcher
  application: Fetcher
}

const app = new Hono<{ Bindings: Bindings }>()

app.all('*', async (c) => {
  const authResponse = await c.env.auth.fetch(c.req.raw.clone())
  if (authResponse.status !== 200) {
    return authResponse
  }
  return await c.env.application.fetch(c.req.raw)
})

export default app
```

JonasHiltl:
Hi @yusukebe,
yes it is not necessary to use Hono Client since I could just use the normal fetch call, but I want the advantages of the full route types that come with Hono Client. 

The service I want to call has validation built in with the honos `zValidator`, so I can get the types for the route. But since the service is a Cloudflare Worker I can only call it through a service binding from a different worker. This currently does not work with Hono Client since it uses the default fetch method. I would be able to call the service binding if the [fetch](https://github.com/honojs/hono/blob/main/src/client/client.ts#L95) method could be passed as an argument to the client (it could default to the normal fetch API). 
That way I can have the full types that come with Hono Client while still being able to call service bindings through their fetch method.

PR #920, opened 2023-02-22
Custom fetch method for Hono Client options

JonasHiltl:
As described in my issue https://github.com/honojs/hono/issues/917 it would be great to be able to pass a custom fetch method to Hono Client to be able to call Service Bindings through Hono Client.

I have just added an optional `fetch` method to the `RequestOptions` so that the `fetch` method of a Service Binding can be used instead of the default `fetch` API when it's set in the options.

See this example on how it would be used:
```
services = [
  { binding = "AUTH", service = "auth-service" },
]
```
```typescript
const client = hc<CreateProfileType>('/', { fetch: c.env. AUTH.fetch });
...
// use client normally
```

### DIFF
--- a/deno_dist/client/client.ts
+++ b/deno_dist/client/client.ts
@@ -92,7 +92,7 @@ class ClientRequestImpl {
     setBody = !(methodUpperCase === 'GET' || methodUpperCase === 'HEAD')
 
     // Pass URL string to 1st arg for testing with MSW and node-fetch
-    return fetch(url, {
+    return (opt?.fetch || fetch)(url, {
       body: setBody ? this.rBody : undefined,
       method: methodUpperCase,
       headers: headers,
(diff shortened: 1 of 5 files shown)


yusukebe:
Hi @JonasHiltl !

Great work! Now I fully understand what you want to do.

I'd like to accept this PR, but there is one request. Could you do `yarn denoify` and `git add` the generated file & git commit & push? The file is needed for the Deno. Thanks.

JonasHiltl:
@yusukebe Done!

yusukebe:
Thanks. I'll merge it!

yusukebe:
Hi @ronkeiser 

This is caused by a type mismatch. The Request object referred to differs between Hono's and `@cloudflare/workers-types`. I think resolving this is complex or impossible. The best workaround is ignoring the type check.

ronkeiser:
@yusukebe Sounds good. Thank you for your responsiveness and your work on this wonderful project!

PR #921, opened 2023-02-23
feat: allow pathname argument

hagishi:
In order to avoid having to write the full URL every time when writing tests, 
we implemented a feature to implicitly format the URL and generate the request object when the pathname argument is entered.


hagishi:
Hi @yusukebe 

Thank you for your response!!

1. I agree to add a leading slash using mergePath if it's missing from the path.
2. I also agree as I didn't anticipate that the path could include "http" in the pathname. 👍

yusukebe:
@hagishi 

Thanks. This feature will make it easier to write tests. I'll merge it!

PR #929, opened 2023-02-25
feat: Improve route()

usualoma:
### What is this PR?

This PR implements the following.
* Reduce side-effects of `route()` calls on the parent app
    * https://github.com/honojs/hono/compare/main...usualoma:hono:feat/basepath-2?expand=1#diff-8ac13809c9886e994d1db33943de82df4d6c5d88b73fd0270c0189804ff565c2R182
* Support nested `route()` calls
    * https://github.com/honojs/hono/compare/main...usualoma:hono:feat/basepath-2?expand=1#diff-8ac13809c9886e994d1db33943de82df4d6c5d88b73fd0270c0189804ff565c2R179-R180

### What will this PR allow us to do

`route()` can be used to specify the base path.

```typescript
const hono = new Hono().route('/api')
const api = hono.get('/search', (c) => c.jsonT({ ok: true }))
type AppType = typeof api
```

image

### Why can't we specify our basePath in the constructor in this PR?

I also wrote the following code to try to be able to specify it in the constructor, but I just could not get it to infer correctly.
https://github.com/honojs/hono/compare/main...usualoma:hono:feat/basepath

image

I know that the following utility method can be prepared and used in place of `new Hono` in the same way as specifying it in the constructor.

```typescript
export const hono = <E extends Env = Env, S = {}, BasePath extends string = ''>(
  init: {
    router?: Router<H>
    strict?: boolean
    basePath?: BasePath
  } = {}
) => {
  return new Hono<E, S, BasePath>(init as any)
}
```

image

However, I see no need for such a utility method, as it is short enough to be written as follows

```typescript
const hono = new Hono().route('/api')
```

### DIFF
--- a/deno_dist/client/client.ts
+++ b/deno_dist/client/client.ts
@@ -101,7 +101,7 @@ class ClientRequestImpl {
 }
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export const hc = <T extends Hono<any>>(baseUrl: string, options?: RequestOptions) =>
+export const hc = <T extends Hono<any, any, any>>(baseUrl: string, options?: RequestOptions) =>
   createProxy(async (opts) => {
     const parts = [...opts.path]
 
--- a/deno_dist/client/types.ts
+++ b/deno_dist/client/types.ts
@@ -52,7 +52,7 @@ type PathToChain<
     }
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type Client<T> = T extends Hono<any, infer S>
+export type Client<T> = T extends Hono<any, infer S, any>
   ? S extends Record<infer K, Endpoint>
     ? K extends string
       ? PathToChain<K, S>
--- a/deno_dist/hono.ts
+++ b/deno_dist/hono.ts
@@ -18,6 +18,7 @@ import type {
   NotFoundHandler,
   OnHandlerInterface,
   TypedResponse,
+  MergePath,
   MergeSchemaPath,
   RemoveBlankRecord,
 } from './types.ts'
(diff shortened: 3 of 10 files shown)


yusukebe:
Hi @usualoma !

This is great! I also don't want to create a utility method like `hono()`. Looks good to me because using the existing `route()` method, two improvements have been made, and the types have been kept.

@yudai-nkt @ThatOneBro and others
What do you think about it?

usualoma:
Hi @yusukebe,
Thanks for the review.
I added two small refactoring commits and then denoify.

yusukebe:
Ah, these have been cleaned up by refactoring. Good.

ThatOneBro:
Looks good! Great work @usualoma , this is some pretty crazy Typescript magic you added

boehs:
I'd say, this is exciting and I'd like to see it in, but I'm very curious to see if it can be cleaned up, `basePath` seems much more intuitive

usualoma:
Hi @boehs,

I get that "basePath is more intuitive."
But when we allow it to be specified in our constructor, the type is inevitably not well inferred.

In this PR, we only improve the behavior of route() without changing hono's API. And with the improved behavior of route(), route() can be used in the same way as specifying basePath.

In the future, if we find a better way to specify it in the constructor, we can rewrite it that way.

Even leaving the constructor discussion aside, I think this PR is a useful change, so I think it is a good idea to merge it first.


### Other Ideas

If route() is not intuitive, it might be possible to create an alias to route().

```typescript
const hono = new Hono().basePath('/api')
```

or

```typescript
const hono = new Hono().under('/api')
```

But, well, I guess I don't need it.

yusukebe:
Hi @boehs @usualoma !

I also hope to use `basePath` in the constructor, but it's impossible now, as @usualoma said (perhaps we can do it if we master TypeScript more and more). And we don't need an alias method for `route()`.

This PR is a very good PR that improves `route()`. I'll merge this.

PR #930, opened 2023-02-25
feat(client): introduce "after" hook

### LINKED ISSUE #928: [Feature suggestion] "after" hook for Hono Client
From discord. About Hono Client.

Sometimes we want to write a common handling when receiving responses.

```ts
const res = await client.api.hello.$get()
if (res.status === 404) {
  window.location.href = '/'
}

// ...

const res = await client.api.morning.$get()
if (res.status === 404) {
  window.location.href = '/'
}

// ...
```

Since it is tedious to write the same process every time, how about having a hook for when the response is received?

```ts
const client = hc<AppType>('/api', {
  after: (res) => {
    if (res.status === 404) {
      window.location.href = '/'
      return
    }
  }
})
```

It can also be used to output logs, for example.

```ts
const client = hc<AppType>('/api', {
  after: (res) => {
    console.log(`Response status is ${res.status}`)
  }
})
```

yusukebe:
This PR introduces the new feature "after" hook for Hono Client.

As mentioned in #928, writing the same steps for each response is tedious.

In this PR, `after` hook allows us to write common actions after getting the response.For example, to output logs, we can write the following:

```ts
const client = hc<AppType>('http://localhost:8787', {
  after: (res) => {
    console.log(`${Date.now()}: Response status is ${res.status}`)
  },
})
```

I would like to keep the Hono Client to a minimum, but this `after` Hook is good because it allows for so much with so little implementation.

This will resolve #928

### DIFF
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -32,7 +32,7 @@ class ClientRequestImpl {
     this.url = url
     this.method = method
   }
-  fetch = (
+  fetch = async (
     args?: ValidationTargets & {
       param?: Record<string, string>
     },
@@ -92,11 +92,21 @@ class ClientRequestImpl {
     setBody = !(methodUpperCase === 'GET' || methodUpperCase === 'HEAD')
 
     // Pass URL string to 1st arg for testing with MSW and node-fetch
-    return (opt?.fetch || fetch)(url, {
+    const res = (opt?.fetch || fetch)(url, {
       body: setBody ? this.rBody : undefined,
       method: methodUpperCase,
       headers: headers,
     })
+
+    if (opt?.after) {
+      const hookedResponse = await opt.after(await res)
+      if (hookedResponse) {
+        return hookedResponse
+      } else {
+        return res
+      }
+    }
+    return res
   }
 }
 
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -100,11 +100,7 @@ class ClientRequestImpl {
 
     if (opt?.after) {
       const hookedResponse = await opt.after(await res)
-      if (hookedResponse) {
-        return hookedResponse
-      } else {
-        return res
-      }
+      return hookedResponse ?? res
     }
     return res
   }
--- a/src/client/types.ts
+++ b/src/client/types.ts
@@ -12,12 +12,12 @@ type Data = {
   output: {}
 }
 
-type AfterHook = (res: Response) => Promise<Response | void> | Response | void
+type AfterResponseHook = (res: Response) => Promise<Response | void> | Response | void
 
 export type RequestOptions = {
   headers?: Record<string, string>
   fetch?: typeof fetch
-  after?: AfterHook
+  after?: AfterResponseHook
 }
 
 type ClientRequest<S extends Data> = {
--- a/src/client/types.ts
+++ b/src/client/types.ts
@@ -12,9 +12,12 @@ type Data = {
   output: {}
 }
 
+type AfterHook = (res: Response) => Promise<Response | void> | Response | void
+
 export type RequestOptions = {
   headers?: Record<string, string>
   fetch?: typeof fetch
+  after?: AfterHook
 }
 
 type ClientRequest<S extends Data> = {
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -369,3 +369,30 @@ describe('Use custom fetch method', () => {
     expect(res).toEqual(returnValue)
   })
 })
+
+describe('After Hook', () => {
+  const server = setupServer(
+    rest.get('http://localhost/hello', (_, res, ctx) => {
+      return res(ctx.text('OK'))
+    })
+  )
+
+  beforeAll(() => server.listen())
+  afterEach(() => server.resetHandlers())
+  afterAll(() => server.close())
+
+  const app = new Hono().get('/hello', (c) => c.text('OK'))
+  type AppType = typeof app
+  const client = hc<AppType>('http://localhost', {
+    after: (res) => {
+      res.headers.append('X-Custom', 'Hono!')
+    },
+  })
+
+  it('Should return Response with the X-Custom header', async () => {
+    const res = await client.hello.$get()
+    expect(res.status).toBe(200)
+    expect(res.headers.get('X-Custom')).toBe('Hono!')
+    expect(await res.text()).toBe('OK')
+  })
+})


src/client/client.ts
```diff
@@ -92,11 +92,21 @@ class ClientRequestImpl {
...
       headers: headers,
     })
+
+    if (opt?.after) {
+      const hookedResponse = await opt.after(await res)
+      if (hookedResponse) {
+        return hookedResponse
+      } else {
+        return res
+      }
```
yudai-nkt, line 107: This could be `return hookedResponse ?? res`, couldn't it? Line 101&ndash;110 could actually be reduced to `return (opt?.after && await opt.after(await res)) ?? res`, but it's a bit too extreme. 😅

src/client/types.ts
```diff
@@ -12,9 +12,12 @@ type Data = {
   output: {}
 }
 
+type AfterHook = (res: Response) => Promise<Response | void> | Response | void
```
yudai-nkt, line 15: Similar to Ky's inferface, `AfterResponse` or `AfterResponseHook` would be more descriptive. Just a personal preference though.

src/client/client.ts
yusukebe: Ah, yes! `return hookedResponse ?? res` is good, but `return (opt?.after && await opt.after(await res)) ?? res` is too  extreme :)

yudai-nkt: Glad you liked the first suggestion! I noticed I can be even more extreme with `return opt?.after?.(await res) ?? res` lol.

yusukebe: Haha

src/client/types.ts
yusukebe: Descriptive is important. Let's make it `AfterResponseHook`.

koistya:
Alternatively:

```ts
const client = hc<AppType>('http://localhost:8787', {
  hooks: {
    afterResponse: [
      async (res, retry) => {
        if (res.status === 401) {
          const accessToken = await getAccessToken();
          return retry({ headers: { authorization: `Bearer ${accessToken}` }});
        }
        return res;
      }
    ]
  }
});
```

src/client/client.ts
```diff
@@ -92,11 +92,17 @@ class ClientRequestImpl {
...
     // Pass URL string to 1st arg for testing with MSW and node-fetch
-    return (opt?.fetch || fetch)(url, {
+    const res = (opt?.fetch || fetch)(url, {
       body: setBody ? this.rBody : undefined,
       method: methodUpperCase,
       headers: headers,
     })
+
+    if (opt?.after) {
+      const hookedResponse = await opt.after(await res)
```
usualoma, line 102: I would also like to receive the request object with a hook. What do you think?

The code is a bit long, but if it looks like the following, we won't have to worry about the order of the parameters if we add retry options later.


```diff
diff --git a/src/client/client.test.ts b/src/client/client.test.ts
index b2c8fde..52e0331 100644
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -384,8 +384,8 @@ describe('After Hook', () => {
   const app = new Hono().get('/hello', (c) => c.text('OK'))
   type AppType = typeof app
   const client = hc<AppType>('http://localhost', {
-    after: (res) => {
-      res.headers.append('X-Custom', 'Hono!')
+    after: ({ response }) => {
+      response.headers.append('X-Custom', 'Hono!')
     },
   })
 
diff --git a/src/client/client.ts b/src/client/client.ts
index 01543b6..2814860 100644
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -92,14 +92,18 @@ class ClientRequestImpl {
     setBody = !(methodUpperCase === 'GET' || methodUpperCase === 'HEAD')
 
     // Pass URL string to 1st arg for testing with MSW and node-fetch
-    const res = (opt?.fetch || fetch)(url, {
+    const req = new Request(url, {
       body: setBody ? this.rBody : undefined,
       method: methodUpperCase,
       headers: headers,
     })
+    const res = (opt?.fetch || fetch)(req)
 
     if (opt?.after) {
-      const hookedResponse = await opt.after(await res)
+      const hookedResponse = await opt.after({
+        request: req,
+        response: await res,
+      })
       return hookedResponse ?? res
     }
     return res
diff --git a/src/client/types.ts b/src/client/types.ts
index ca26125..f94ffb1 100644
--- a/src/client/types.ts
+++ b/src/client/types.ts
@@ -12,7 +12,10 @@ type Data = {
   output: {}
 }
 
-type AfterResponseHook = (res: Response) => Promise<Response | void> | Response | void
+type AfterResponseHook = (arg: {
+  request: Request
+  response: Response
+}) => Promise<Response | void> | Response | void
 
 export type RequestOptions = {
   headers?: Record<string, string>

```

usualoma:
I agree with the current code except where I commented.

yusukebe:
Thank you for the comments. I'm going on a trip and will reply later.

yudai-nkt:
(more of an idea rather than review or suggestion, so leaving comments here)

After some thought, I suppose `opt?.hooks?.after` is better organized than `opt?.after` if you accept more hooks after receiving feature requests in the future. Examples of such hooks include but are not limited to `beforeRequest`, `onFailure`, and `onTimeout` (these are what I can come up with right now).

It depends on how minimal we want to keep `hc`.

koistya:
@yudai-nkt potential hook names to consider for future development ([`options.hooks`](https://github.com/honojs/hono/pull/930#issuecomment-1445206677)):

`init`, `beforeRequest`, `beforeRedirect`, `beforeRetry`, `afterResponse`, `beforeError`.

yudai-nkt:
@koistya oops I overlooked your comment, sorry. I think we are basically along the same lines.

yusukebe:
Hi @yudai-nkt @usualoma @koistya !

Thank you for the review and comment. This is a matter of how far `hc` is minimal, as @yudai-nkt says. It's not an easy decision to make so let's take some time to think about it.

yusukebe:
I made this PR a draft because it is still incomplete and will not be developed soon.

dimik:
[KY](https://github.com/sindresorhus/ky#hooks) uses the same hooks naming as @koistya suggested, would be great to use the same here. Another alternative is to have a Middleware instead of hook, It's obvious that beforeRequest/afterResponse looks exactly like middleware, and so we have single concept for client and server

PR #933, opened 2023-02-26
fix: let the app.request behave the same as fetch

tangye1234:
This pr is to extend `app.request` usage range.

We test app by fetching directly via `app.request` without a setup of an app server, but the request parameters can only accept string or request, while fetch's can accept string, request or URL for the first parameter.

Moreover, when the input is request, then the second paramter `RequestInit` dose not worked as fetch dose.

### DIFF
--- a/deno_dist/hono.ts
+++ b/deno_dist/hono.ts
@@ -264,10 +264,14 @@ export class Hono<E extends Env = Env, S = {}> extends defineDynamicClass()<E, S
     return this.dispatch(request, executionCtx, Env)
   }
 
-  request = async (input: Request | string, requestInit?: RequestInit) => {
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @tangye1234 !

Thanks! This is a good fix. I want to merge. So, for the last step. could you do `yarn denoify`, `git add` the generated file,  and commit&push it? It will be needed for Deno.

tangye1234:
> Hi @tangye1234 !
> 
> Thanks! This is a good fix. I want to merge. So, for the last step. could you do `yarn denoify`, `git add` the generated file, and commit&push it? It will be needed for Deno.

Thanks for your advice. I have already pushed new generated deno source files.

yusukebe:
Okay! I'll merge it. Thank you!

PR #937, opened 2023-02-27
fix(client): show type error if `$get()` needs args

yusukebe:
Made it throws the type error if `client.path.$method()` needs arguments:

SS

This fix will make the Client more type-safety.


yudai-nkt:
This is what I found inconvenient but I haven't filed an issue yet, thanks! I'll check later.

yusukebe:
Hi @yudai-nkt !

Have you done check it? I want to merge it.

yudai-nkt:
I'm afraid I won't have check this on my local machine anytime soon, but basically LGTM!

yusukebe:
Okay, I'll merge it!

PR #939, opened 2023-02-27
feat(`jsonT`): check JSON type

yusukebe:
`jsonT()` returns a JSON object composed of primitive types. For example, a `Date` object is converted to a string. But currently, `jsonT()` allows the `Date` type.

```ts
app.get('/hello', (c) => {
  return c.jsonT({
    messge: 'Hello',
    datetime: new Date(), // does not show the error
  })
})
```

And the Client will get the `Date` type even though the actual value is a string.

SS

Allowing the types of objects other than primitives to be loose can be confusing.

In this PR, I've made it provides a JSON type and limits the object that `jsonT()` would receive, and if they are not JSON type, it throws an error.

SS

On the client side, the return type of `res.json()` is `never`.

SS

This feature was inspired by the following article.

<https://zenn.dev/kosei28/articles/f4bac1ed2b64a7#date%E5%9E%8B%E3%81%AF%E6%96%87%E5%AD%97%E5%88%97%E3%81%AB%E3%81%AA%E3%82%8B>

By restricting the types that `jsonT()` accepts to JSON types composed of primitives, we can make it more type-safe.

### DIFF
--- a/deno_dist/utils/types.ts
+++ b/deno_dist/utils/types.ts
@@ -10,3 +10,18 @@ export type UnionToIntersection<U> = (U extends any ? (k: U) => void : never) ex
 ) => void
   ? I
   : never
+
+export type JSONPrimitive = string | boolean | number | null | undefined
+export type JSONArray = (JSONPrimitive | JSONObject | JSONArray)[]
+export type JSONObject = { [key: string]: JSONPrimitive | JSONArray | JSONObject }
+export type PrettyJSON<T> = T extends true
+  ? boolean
+  : T extends false
+  ? boolean
+  : T extends JSONPrimitive
+  ? T
+  : T extends JSONArray
+  ? PrettyJSON<T[number]>
+  : T extends JSONObject
+  ? { [K in keyof T]: PrettyJSON<T[K]> }
+  : never
--- a/deno_dist/utils/types.ts
+++ b/deno_dist/utils/types.ts
@@ -15,12 +15,14 @@ export type JSONPrimitive = string | boolean | number | null | undefined
 export type JSONArray = (JSONPrimitive | JSONObject | JSONArray)[]
 export type JSONObject = { [key: string]: JSONPrimitive | JSONArray | JSONObject }
 export type JSONValue = JSONObject | JSONArray | JSONPrimitive
-export type PrettyJSON<T> = T extends true
-  ? boolean
-  : T extends false
-  ? boolean
-  : T extends JSONPrimitive
-  ? T
+
+// `boolean` will be `true` or `false` because it's an alias of them.
+// See: https://github.com/microsoft/TypeScript/issues/22596
+// This type converts `true | false` to `boolean` that we expect.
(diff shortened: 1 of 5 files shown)


yusukebe:
Hi @usualoma @yudai-nkt !

What do you think about this feature? Could you look at it with PR #937?

usualoma:
I think this change is reasonable.

src/utils/types.ts
```diff
@@ -10,3 +10,18 @@ export type UnionToIntersection<U> = (U extends any ? (k: U) => void : never) ex
 ) => void
   ? I
   : never
+
+export type JSONPrimitive = string | boolean | number | null | undefined
+export type JSONArray = (JSONPrimitive | JSONObject | JSONArray)[]
+export type JSONObject = { [key: string]: JSONPrimitive | JSONArray | JSONObject }
```
yudai-nkt, line 16: Top-level primitives and arrays can be valid JSONs. So the correct JSON type definition would be as follows IIUC.

```ts
export type JSONObject = JSONPrimitive | { [key: string]: JSONObject } | JSONObject[]
```

src/utils/types.ts
```diff
@@ -10,3 +10,18 @@ export type UnionToIntersection<U> = (U extends any ? (k: U) => void : never) ex
...
+export type PrettyJSON<T> = T extends true
```
yudai-nkt, line 17: I'm afraid I'm not sure what `PrettyJSON` does (I'm just reviewing on Web UI right now). A small comment or unit tests describing its functionality would be appreciated.

yudai-nkt:
The motivation behind this PR sounds great and I left two small comments. Please check them when you have time.

src/utils/types.ts
yusukebe: You are right. These values can be JSON. I'll make `JSONValue` type like the following and use it.

```ts
type JSONValue = JSONObject | JSONArray | JSONPrimitive
```

deno_dist/utils/types.ts
```diff
@@ -10,3 +10,21 @@ export type UnionToIntersection<U> = (U extends any ? (k: U) => void : never) ex
...
+
+export type JSONPrimitive = string | boolean | number | null | undefined
+export type JSONArray = (JSONPrimitive | JSONObject | JSONArray)[]
+export type JSONObject = { [key: string]: JSONPrimitive | JSONArray | JSONObject }
+export type JSONValue = JSONObject | JSONArray | JSONPrimitive
+
+// `boolean` will be `true` or `false` because it's an alias of them.
+// See: https://github.com/microsoft/TypeScript/issues/22596
+// This type converts `true | false` to `boolean` that we expect.
+type TrueAndFalseToBoolean<T> = T extends true ? boolean : T extends false ? boolean : T
```
yusukebe, line 22: @yudai-nkt 

I added the comment and made `TrueAndFalseToBoolean` that is more descriptive. As noted here, `boolean` is distributed `true | false`, so I'm fixing it. I'm sure there are other ways to do it, but this is how I did it.

yusukebe:
I'll merge it now!

PR #941, opened 2023-03-01
feat(app): add `basePath` option

yusukebe:
With this PR, we can specify the `basePath` in the constructor of `Hono`:

```ts
const app = new Hono({ basePath: '/api' })
```

Also, it has the correct type.

SS

SS

Related to #929

### DIFF
--- a/deno_dist/hono.ts
+++ b/deno_dist/hono.ts
@@ -33,7 +33,7 @@ interface RouterRoute {
 }
 
 function defineDynamicClass(): {
-  new <E extends Env = Env, S = {}, BasePath extends string = ''>(): {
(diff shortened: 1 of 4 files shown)


yusukebe:
Hi @usualoma !

Maybe I could have done it. Could you see it if you have time?

usualoma:
Hi @yusukebe,

Thanks!
I could get that far too, but when I specified `<Env>` as the type parameter, the basePath inference did not work as expected.

image

We could proceed with the restriction that "if <Env> is specified, route() is the only way to go."

yusukebe:
Hmmm...Type is a hard thing.

yusukebe:
@usualoma Thanks for the comment!

Let us give up `basePath` and use `app.route()`. Closing the PR.

PR #942, opened 2023-03-01
fix(validator): can't get json value

super-yolin:
fix(validator): can't get json value.

### DIFF
--- a/src/validator/validator.ts
+++ b/src/validator/validator.ts
@@ -25,7 +25,7 @@ export const validator = <
     switch (target) {
       case 'json':
         try {
-          value = await c.req.raw.clone().json()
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @super-yolin !

We need to clone the raw request object, and it is working correctly least with the tests on Cloudflare Workers.

I want to see your code which does not work well. Could you paste the snippets as a comment?

super-yolin:
> Hi @super-yolin !
> 
> 
> 
> We need to clone the raw request object, and it is working correctly least with the tests on Cloudflare Workers.
> 
> 
> 
> I want to see your code which does not work well. Could you paste the snippets as a comment?

it doesn't work in bun

PR #943, opened 2023-03-02
fix(pages/nextjs): don't use `route()` if path is not passed

### LINKED ISSUE #938: Capturing Hono exceptions with the Cloudflare Pages Sentry Plugin
Hi there, i'd like to be able to use the Cloudflare Pages Sentry Plugin (https://developers.cloudflare.com/pages/platform/functions/plugins/sentry) to capture exceptions from my Hono API. I've got the following 2 files:

`functions/hello.ts`
```
import { Hono } from 'hono'
import { handle } from 'hono/cloudflare-pages'

const app = new Hono()

app.get('/hello', (c) => {
  throw new Error('Capture this error in Sentry!')
})

export const onRequest = handle(app, '/api')
```

`functions/_middleware.ts`
```
import sentryPlugin from "@cloudflare/pages-plugin-sentry";

export const onRequest: PagesFunction = sentryPlugin({
  dsn: "https://sentry.io/welcome/xyz",
});
```

When I trigger the error from the `hello.ts` function.. it doesn't see to be properly capturing it in the Sentry Pages Plugin. **I'd like to use the Sentry Pages Plugin so I can capture all my exceptions (not just from my Hono API's).** 

Any idea's how I could get the exception to properly trigger the Sentry Pages Plugin? Thanks!

yusukebe:
If we use `handle` in the Cloudflare Pages adapter with no second argument:

```ts
export const onRequest = handle(app)
```

With the current implementation,  `app` will be passed to `route()`

https://github.com/honojs/hono/blob/b1c50597088c2e4e1d3666f53f3f9de171cbd506/src/adapter/cloudflare-pages/handler.ts#L17-L22

If we do so, the error thrown from `app` can't be re-thrown:

```ts
const app = new Hono().route('/api')

app.get('/hello', (c) => {
  throw new Error('Capture this error in Sentry!')
})

app.onError((e) => {
  throw e // <--- This will not be a real "error", will return 500 response.
})

export const onRequest = handle(app)
```

This PR fixes this issue.

This will resolve #938

### DIFF
--- a/src/adapter/cloudflare-pages/handler.test.ts
+++ b/src/adapter/cloudflare-pages/handler.test.ts
@@ -38,4 +38,22 @@ describe('Adapter for Cloudflare Pages', () => {
     expect(res.status).toBe(200)
     expect(await res.text()).toBe('/api/foo')
   })
+
+  it('Should not use `route()` if path argument is not passed', async () => {
+    const request = new Request('http://localhost/api/error')
+    const app = new Hono().route('/api')
+
+    app.onError((e) => {
+      throw e
+    })
+    app.get('/error', () => {
+      throw new Error('Custom Error')
+    })
+
+    const handler = handle(app)
+    // It does throw the error if app is NOT "subApp"
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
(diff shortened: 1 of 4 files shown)


yusukebe:
This is a good fix, and there are no side effects. Merge it.

Issue #947, opened 2023-03-03
Using multiple or nested `Hono.prototype.route`


usualoma:
I, too, think it would be better to be able to specify this in the method chaining.

### Discussed in https://github.com/orgs/honojs/discussions/945

<div type='discussions-op-text'>

<sup>Originally posted by **kissge** March  3, 2023</sup>
I've tried to use `route` like this, but couldn't make it work:

```ts
// api/photos/photos.routes.ts
const routes = new Hono().get('/list', (c) => c.jsonT({ photos: 0 }))
export default routes
```

```ts
// api/videos/videos.routes.ts
const routes = new Hono().get('/list', (c) => c.jsonT({ videos: 0 }))
export default routes
```

```ts
// api/routes.ts
import photos from './photos/photos.routes.ts'
import videos from './photos/photos.routes.ts'
const routes = new Hono().route('/photos', photos).route('/videos', videos)
export default routes
```

```ts
// api/index.ts
import routes from './routes'
const app = new Hono()
    .route('/', routes)
    .use(...)
    .use(...)
    .use(...)
```

Is this usage expected? Does hono support it? Any hint would be appreciated.

https://github.com/kissge/hono/commit/7e35ff5817b92c87a97276e6cea20edb57aa94c0</div>

PR #948, opened 2023-03-03
feat: Returns the parent's app from `route(path, subApp)`.

usualoma:
#947

As for `route()`, in general, I think the expected type depends on the arguments given.

If no subApp is given, then of course an app with path as basePath should be returned.
```
const subApp = app.route('/api')
```

However, if a subApp is given, the app that has it in the basePath is usually not necessary.
```
const probablyNeverBeUsed = app.route('/api', subApp)
```
If so, it would be more convenient to return `this` when a subApp is specified. This PR includes such a change.

### Is this a good API?

However, it is generally not a good API to have the return type change depending on the argument. I think we can avoid this confusion by renaming the single argument `route()` as `basePath()` and make the return value intuitive as well.
https://github.com/honojs/hono/compare/main...usualoma:hono:feat/basePath-method

### Other options

I don't think the current behavior of main branch is so bad, so I think there is an option not to change it.

@yusukebe 
What do you think?

### DIFF
--- a/src/adapter/cloudflare-pages/handler.ts
+++ b/src/adapter/cloudflare-pages/handler.ts
@@ -9,14 +9,18 @@ type EventContext = {
 }
 
 interface HandleInterface {
-  <E extends Env>(app: Hono<E>, path?: string): (
-    eventContext: EventContext
-  ) => Response | Promise<Response>
+  <E extends Env, S extends {}, BasePath extends string>(
+    app: Hono<E, S, BasePath>,
+    path?: string
+  ): (eventContext: EventContext) => Response | Promise<Response>
 }
 
 export const handle: HandleInterface =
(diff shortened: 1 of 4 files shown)


yusukebe:
Hi @usualoma !

I've checked and I'm considering it. Give me a minute. Thanks!

yusukebe:
Hi @usualoma !

Thank you for the PR.

This looks great to me. It was bad that we can not "chain" because `app.route()`  returning "subApp". It is better to returns `this`.

I agree with creating a `basePath()`. It is easier for users to understand.

---

If we are merging, please direct it to the [`next` branch](https://github.com/honojs/hono/tree/next). This will be used for the next minor version "v3.1".

yusukebe:
This PR is not needed now. Closing.

PR #949, opened 2023-03-03
feat: `env` support enviroment variables for multi runtimes

yusukebe:
This PR introduces `env` function to get environment variables for any runtimes, not only variables of Cloudflare Workers' Bindings.

We can write the code with the same syntax at any runtime:

```ts

app.get('/env', (c) => {
  const { NAME } = env<{ NAME: string }>(c)
  return c.text(NAME)
})
```

Supported runtimes:

* Cloudflare Workers
* Deno
* Bun
* Lagon ( It works on dev but can't resolve the path of `.env` on CI )
* Node.js
* Vercel ( No CI )
* Fastly ( Users may not define the values... )

## Why don't we use `c.env`?

Already we are using `c.env` for Cloudflare Workers' Bindings, so we can use it for other runtimes. But, if we do so, we have to write the code to `context.ts`. It's bad to write that things depend on each environment on the common component. Separating `env` as `import { env } from 'hono/adapter'` is a better API.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
       - uses: denoland/setup-deno@v1
         with:
           deno-version: v1.x
-      - run: deno test --allow-read test_deno
+      - run: deno test --allow-read --allow-env test_deno
 
   bun:
     name: 'Bun'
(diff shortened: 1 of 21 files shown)


yusukebe:
@yujong-lee @metrue @ThatOneBro and others:

Hi, What do you think about this feature? I think this is necessarily, and API design is better.

yujong-lee:
Great work! It makes sense to write it as a function.

yusukebe:
Thanks. Let's go with it.

PR #951, opened 2023-03-03
fix(type): remove unnecessary distribution

### LINKED ISSUE #950: Returning `Array<T>` response makes `jsonT` think it as `T` type
https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgCQgOwnAvnAZlCEOAcgAt0JiBuAKFElkTlIGNs8CiyKB6FgG2ABTNDGo0aLdAGd4AQzBg4AXjhohAdxQUAFAEoAdAHMhMHcR64IlADRwdLPSoB8cFgYBW09ABUdSYAATAC44AG1iK1s4ABYAJgBdbD09Wh4eOEy4AD0AfgkpNFk3QRF4VVYAHhgATzAhCFw4BTBncx5iVMkZeCghaRVmjTlgeAFhUQMogwASEzMuwuLAuRg5Qblh0bg+6U9vNH00jKy8iSA

image

yusukebe:
Do not use `PrettyJSON<T>` because it causes unnecessary distributions. As a side effect, `boolean` is inferred to `true | false`. But I think there is no problem in practical use.

This will fix #950

### DIFF
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -283,7 +283,7 @@ describe('Infer the response/request type', () => {
       const req = client.index.$get
 
       type Actual = InferResponseType<typeof req>
-      type Expected = { ok: boolean }
+      type Expected = { ok: true }
       type verify = Expect<Equal<Expected, Actual>>
     })
 
@@ -340,7 +340,7 @@ describe('Merge path with `app.route()`', () => {
     const client = hc<AppType>('http://localhost')
     const res = await client.api.search.$get()
     const data = await res.json()
-    type verify = Expect<Equal<boolean, typeof data.ok>>
+    type verify = Expect<Equal<true, typeof data.ok>>
     expect(data.ok).toBe(true)
   })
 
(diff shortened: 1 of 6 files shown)


yudai-nkt:
I don't really look at this in detail (I'm on the phone right now), but can't this be fixed by wrapping a problematic type parameter into a tuple to suppress union distribution?

yusukebe:
Hi  @yudai-nkt !

What I want to do is "do not expand `boolean` to `true | false`", so I made the `PrettyJSON`:

https://github.com/honojs/hono/pull/939#discussion_r1119813580

But, this makes unexpected behavior:

https://github.com/honojs/hono/issues/950

My implementation may be bad, however, we can decide don't use `PrettyJSON` and use `true | false` without forcing `boolean` to be them.

yudai-nkt:
I commented without really understanding whether union distribution is the cause of the problem (so deleted it and will check later). thanks for the explanation anyway!

yudai-nkt:
Yes, I confirmed I was misunderstanding and we should go with this PR! Sorry for the disturbance.

yusukebe:
Merging it into the "next" branch!

PR #956, opened 2023-03-05
docs(readme): add creator name of RegExpRouter and SmartRouter

yusukebe:
I have wanted to write this for a while; this is important.

One of the features of Hono is the RegExpRouter, but I must not take credit for that.


yusukebe:
@usualoma I'll merge this, if that's okay:)

usualoma:
Thank you!

PR #957, opened 2023-03-05
feat(context): allow passing `unknown` to `executionCtx.waitUntil`

yusukebe:
This is suggested by @koistya. Thanks!

PR #958, opened 2023-03-05
Allow any instead of void for waitUntil.

Manouchehri:
This matches up with what Cloudflare uses.

https://github.com/Cloudflare-Mining/Cloudflare-Datamining/blob/87ef982bb7bbf9f8f1468f38de1dc1701e4a04cf/data/workers-types/experimental/index.d.ts#L291


yusukebe:
Hi @Manouchehri !

Thank you!

We just implemented the same feature and merged it into the "next" branch that is used for the next minor version v3.1.

https://github.com/honojs/hono/pull/957

The PR uses `unknown`, which is better than `any`.

Issue #952, opened 2023-03-03
Basic auth not working when using app.onError


JustJoostNL:
Since version 3.0.0 the basic auth doesn't work anymore when also using `app.onError`

Example:

```js

app.use('/test-auth', basicAuth(
    {
        username: 'test',
        password: 'test',
    }
))

app.get('/test-auth', async (c) => {
    return c.json({
        message: 'You are authorized!',
        status: 200
    })
})

app.onError((err, c) => {
    console.log("An error has been thrown: " + err)
    return c.json({
        message: 'Internal Server Error',
        status: 500
    }, 500);
})

```

In this example, when you use the /test-auth route, it just returns an "Internal Server Error" and the error it returns is: "Error: Unauthorized"

When I am not using the `app.onError` (I comment that out), it does work normally.
This issue occurs since version 3.0.0

I hope someone can help 🙂
Thanks in advance!

yusukebe:
Hi @JustJoostNL !

Since version 3.0.0, a basic auth middleware throws the `HTTPException` if it does not authorize. This makes our appliation more secure. So, we have to handle it in `app.onError()`. For example, to customize the message, write like the following:

```ts
app.onError((err, c) => {
  if (err instanceof HTTPException) {
    const res = err.getResponse()
    return new Response(
      JSON.stringify({
        message: 'Unauthorized',
        status: 401,
      }),
      res
    )
  }
  return c.text('Internal Server Error', 500)
})
```

This is a bit confusing. If you can work it fine, I will put it in the documentation. Thank you.

P.S:

@usualoma @ThatOneBro and others

If you have any smart ideas to handle the exception, please share these. But I think this way is better. Or it's trouble to make a response from the `Response` object, so, I think it will be convenient like this API that overrides `c.json()`:

```ts
if (err instanceof HTTPException) {
  const res = err.getResponse()
  return c.json(
    {
      message: 'Unauthorized',
      status: 401,
    },
    res
  )
}
```

But it may be necessary because the cord getting fat.

JustJoostNL:
Hey @yusukebe! Thanks for your answer!

I assume I need to use an `import` to get the `HTTPException`?

I did try:

 ```js
import { HTTPException } from "hono/dist/types/http-exception";
``` 
But I get an error when using that.

yudai-nkt:
[...]

PR #959, opened 2023-03-06
feat(context): allow passing `RequestInit` to `c.json()` etc.

yusukebe:
This PR proposes allowing us to pass `InitResponse` as the second argument to `c.text()` or `c.json()`. It will overload these methods.

```ts
interface TextRespond {
  (text: string, status?: StatusCode, headers?: HeaderRecord): Response
  (text: string, init?: ResponseInit): Response
}
```

This is inspired by Issue #952. Now, it is very tedious to customize Basic Auth errors.

```ts
app.onError((e, c) => {
  if (e instanceof HTTPException) {
    const res = e.getResponse()
    res.headers.set('content-type', 'application/json')
    return new Response(
      JSON.stringify({
        message: 'Unauthorized!',
      }),
      res
    )
// ...
```

So I thought it would be easier if we can pass `ResponseInit`, i.e. meta information such as headers, to `c.json()`.

```ts
app.onError((e, c) => {
  if (e instanceof HTTPException) {
    return c.json(
      {
        message: Unauthorized!',
      },
      e.getResponse()
    )
  }
```

This feature is not only used in the above cases, but can also be used when we want to modify and return the content retrieved from the outside using `fetch`.

```ts
app.get('/mirror', async (c) => {
  const res = await fetch('http://example.com')
  return c.json({ foo: 'bar' }, res)
})
```

## Side effects

* The code size will increase. But they are almost types definitions.
* Overload is often confusing to users. But, I think it is intuitive to pass `ResponseInit` in this case.
* The implementation becomes more complex. There is no way around this :(

---

It would be a good feature, except that the code would be complicated. In addition, I think it's not a bad idea to allow overloads since one of the big features of Hono is flexible to write code.

If you have any other good ideas, please let me know.

### DIFF
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -16,6 +16,39 @@ export interface ExecutionContext {
 }
 export interface ContextVariableMap {}
 
+interface NewResponse {
+  (data: Data | null, status?: StatusCode, headers?: HeaderRecord): Response
+  (data: Data | null, init?: ResponseInit): Response
+}
+
+interface BodyRespond extends NewResponse {}
+
+interface TextRespond {
+  (text: string, status?: StatusCode, headers?: HeaderRecord): Response
+  (text: string, init?: ResponseInit): Response
+}
+
+interface JSONRespond {
+  <T = JSONValue>(object: T, status?: StatusCode, headers?: HeaderRecord): Response
+  <T = JSONValue>(object: T, init?: ResponseInit): Response
+}
+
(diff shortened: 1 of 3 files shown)


ThatOneBro:
I like this change a lot, makes sense to be able to modify responses like this to me.

yusukebe:
Hi @ThatOneBro @usualoma 

Thanks for the comments! I'm going to merge it!

PR #969, opened 2023-03-09
feat(validator): support type transformation

### LINKED ISSUE #946: Using zod's `z.transform` makes hc client `unknown` type
// Please allow me to make several posts one after another..

[https://www.typescriptlang.org/play?#code/JYWwDg9gTgL....](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgLwGoEMA2wAm6bRwC+cAZlBCHAOQACAFhAHYQD0yEOAtAG5a75o1ANwAoUJFiI4ACWYRiZClWqMWI8eGjwkyReUo0OODRO3T6AY33KaatpewBTJjA2iEouN7iXmAZ3h0MDA4AF44JicAd1l5AAoASgA6AHMnGHivHxzqVlIICGoAGmyc7zR+PAIoLPL6mgBHAFcnKABPErKGiuSIACMAKydLTM8entwALhRkwKhgJlSk5KgndIAPeNYAPQBtAAYuAE4AXQBqABJWRO6JhuSYKHQmfwKoEHiAOWaQfrbEqV7t4iIC7j4wT14pZEuEAHzg+qWZKDfzMAAq8XGwJy018qycjWSfGwOHi1BabU6KVwQJxPlYrHp9x2AH5EeVQXTvIkxN0-K94I5gC54BErAAeGDtMBOCCkODBMBw8msai87qMnJs0REDz8gJBELhSIxOIsFbpTIcvIFIrc8qVUmCWoc3KUjpdHHIPpDEZjN0NPE++aLZYpNabbb7I5nK43QP1LWPZ6vd6fH5-AEOnpcjmQhrQ2FhBE45GojFYxPlPHItZEkm4cke6nJWnVnJa5k9HU4vM5DUGwW+ZyuE2S6Wy+WKkIqvLqsSdpk+HV60RAA)

image

yusukebe:
This PR provides supporting "type transformation" for the validator.

For example, Zod has a "transform" feature.

```ts
z.string()
  .regex(/^[0-9]+$/)
  .transform(Number)
```

This means that the input type(`string`) and output type(`number`) are different. But, Hono does not support different input and output types. If we use "transform" in Zod Validator Middleware, the value from `c.req.valid()` will be "input type" not "output type":

```ts
const app = new Hono().post(
  '/foo',
  zValidator(
    'json',
    z.object({
      id: z
        .string()
        .regex(/^[0-9]+$/)
        .transform(Number),
    })
  ),
  (c) => {
    c.req.valid('json').id // expected to be `number`
```

SS

This is mentioned at #946 

With this PR, the validator treats input and output types as different, so it will suport Zod's transform.

SS

Without this feature, Validation using Zod and others would be very inconvenient to use. So, if there are no problems, would like to include it in the next minor version upgrade "v3.1.0.

This will fix #946

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -22,8 +22,8 @@ export type Env = {
 export type Next = () => Promise<void>
 
 export type Input = {
-  input?: Partial<ValidationTargets>
-  output?: Partial<{ [K in keyof ValidationTargets]: unknown }>
+  in?: Partial<ValidationTargets>
+  out?: Partial<{ [K in keyof ValidationTargets]: unknown }>
 }
 
 ////////////////////////////////////////
@@ -71,12 +71,12 @@ export interface HandlerInterface<
   // app.get(handler, handler)
   <I extends Input = {}, O = {}>(
     ...handlers: [H<E, ExtractKey<S>, I, O>, H<E, ExtractKey<S>, I, O>]
-  ): Hono<E, RemoveBlankRecord<S | Schema<M, ExtractKey<S>, I['input'], O>>, BasePath>
+  ): Hono<E, RemoveBlankRecord<S | Schema<M, ExtractKey<S>, I['in'], O>>, BasePath>
 
   // app.get(handler x 3)
   <P extends string, O = {}, I extends Input = {}, I2 extends Input = I, I3 extends Input = I & I2>(
     ...handlers: [H<E, ExtractKey<S>, I, O>, H<E, ExtractKey<S>, I2, O>, H<E, ExtractKey<S>, I3, O>]
-  ): Hono<E, RemoveBlankRecord<S | Schema<M, ExtractKey<S>, I3['input'], O>>, BasePath>
+  ): Hono<E, RemoveBlankRecord<S | Schema<M, ExtractKey<S>, I3['in'], O>>, BasePath>
 
   // app.get(handler x 4)
   <
@@ -93,7 +93,7 @@ export interface HandlerInterface<
       H<E, ExtractKey<S>, I3, O>,
       H<E, ExtractKey<S>, I4, O>
     ]
-  ): Hono<E, RemoveBlankRecord<S | Schema<M, ExtractKey<S>, I4['input'], O>>, BasePath>
+  ): Hono<E, RemoveBlankRecord<S | Schema<M, ExtractKey<S>, I4['in'], O>>, BasePath>
 
   // app.get(handler x 5)
   <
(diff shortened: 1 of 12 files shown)


yusukebe:
Hi @yudai-nkt !

Let me know your opinion for this feature. I think the RPC will be brushed up with the addition of this.

src/types.ts
```diff
@@ -21,7 +21,10 @@ export type Env = {
 
 export type Next = () => Promise<void>
 
-export type Input = unknown
+export type Input = {
```
yudai-nkt, line 24: `Input["input"]` and `Input["output"]` might be confusing, so there would be a better name for this (I'm afraid I can't come up with a nice suggestion now). Ideas from other people are very much appreciated.

yudai-nkt:
I left a small comment and below is an overall one.

Would it be possible to turn this into a `@hono/zod-validator` feature? As far as I read TypeBox's readme, it doesn't have such type transformation. So it may well be said that this PR is currently a Zod-specific feature. If we add such features that are specific to a certain validation library into Hono's core, it will bloat the bundle size (most of the addition in this PR are only type-level though). If future contributors make other validation middlewares such as `@hono/ajv-validator` or `@hono/yup-validator`, it is more likely to happen such situations because each validation library has its own features.

In short, I suppose `@hono/zod-validator` is a better place for this PR to live in. I don't know about technical feasibility of implementing this feature in the middleware, so please tell me if it's difficult.

yusukebe:
Hi @yudai-nkt !

Thank you for your comment.

> Would it be possible to turn this into a @hono/zod-validator feature?

This feature is not only for the Zod Validator, but for general purposes. I think we should implement it in the core. I used the word "transform" because it is a useful.

For example, if we want to validate a string input and receive it as a number type in the handler, we can do so without using Zod as follows.

```ts
const validationFunction: ValidationFunction<{ id: string }, { id: number }> = (
  v
) => {
  // Validate
  return {
    id: Number(v.id),
  }
}

const route = app.get('/post', validator('json', validationFunction), (c) => {
  const { id } = c.req.valid('json') // `id` is number
  return c.jsonT({
    id 
  })
})

const client = hc<typeof route>('/')
client.post.$get({
  json: {
    id: '123', // `id` have to be string
  },
})
```

The input type and output type of `@hono/typebox-validator` are the same, but users may write validators with different input and output types using TypeBox.

It is natural for doing type conversions within the "validation function" and this PR is necessary to support it. Of course, it depends on the scope of the "validation function" role, but users may expect type conversions to be available in the "validation function".

---

I'll consider the naming of `Input["input"]` and `Input["output"]` later!

src/types.ts
yusukebe: I've renamed `input` to `in` and `output` to `out`. It's not that different, but I think it's easier to understand than before.

yusukebe:
Alright, I'll merge this!

PR #971, opened 2023-03-11
fix(types): make `MergePath<'/api', '/'>` work well

### LINKED ISSUE #962: Hono Client incorrectly handles root endpoints
Defining root endpoints makes the hono client to add a `index` route, which make the requests fail because it doesn't exist.

```ts
const app = new Hono()
  .get('/', (c) => c.text('Pretty Blog API'))

type App = typeof app;

const client = hc<App>("http://localhost:8787/");
```

### Current Behavior
```ts
client.index.$get()
```

### Expected Behavior
```ts
client.$get()
```

yusukebe:
This PR fixes the `MergePath` type definition.

Before this PR if you write the following:

```ts
type path = MergePath<'/api', '/'>
```

`path` will be `/api/`. But this is not correct because the actual value with `mergePath()` is the different. The trailing slash is removed:

https://github.com/honojs/hono/blob/fea78f29bd681350ce61dd71b33042aa6ec7db5c/src/utils/url.test.ts#L113

So, I've fixed it:

```ts
type path = MergePath<'/api', '/'>
// path will be /api
```

Whit this PR, the matters mentioned at #962 may be resolved.

SS

### DIFF
--- a/deno_dist/types.ts
+++ b/deno_dist/types.ts
@@ -338,7 +338,9 @@ export type MergePath<A extends string, B extends string> = A extends ''
     ? `${P}/${Q}`
     : `${P}/${B}`
   : B extends `/${infer Q}`
-  ? `${A}/${Q}`
+  ? Q extends ''
+    ? A
+    : `${A}/${Q}`
   : `${A}/${B}`
 
 ////////////////////////////////////////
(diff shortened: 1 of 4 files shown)


yusukebe:
Hi @usualoma !

I think this is a good fix. What do you think about it?

PR #973, opened 2023-03-12
fix(etag): returns correct headers

### LINKED ISSUE #972: `etag` middleware returns incorrect headers
The [**IETF** standards](https://datatracker.ietf.org/doc/html/rfc9110#name-304-not-modified) say the following[^1] about `304 Not Modified` responses...

> The server generating a 304 response MUST generate any of the following header fields that would have been sent in a [200 (OK)](https://datatracker.ietf.org/doc/html/rfc9110#status.200) response to the same request:
> * [Content-Location](https://datatracker.ietf.org/doc/html/rfc9110#field.content-location), [Date](https://datatracker.ietf.org/doc/html/rfc9110#field.date), [ETag](https://datatracker.ietf.org/doc/html/rfc9110#field.etag), and [Vary](https://datatracker.ietf.org/doc/html/rfc9110#field.vary)
> * Cache-Control and Expires (see [[CACHING](https://datatracker.ietf.org/doc/html/rfc9111)])

However, the `etag` middleware does not return the `ETag` header for 304 responses.

https://github.com/honojs/hono/blob/fea78f29bd681350ce61dd71b33042aa6ec7db5c/src/middleware/etag/index.ts#L20-L30

[^1]: https://datatracker.ietf.org/doc/html/rfc9110#section-15.4.5-2

yusukebe:
In this PR, I've made the Etag Middleware returns the `Etag` header if the response is `304`.

This will fix #972

### DIFF
--- a/deno_dist/middleware/etag/index.ts
+++ b/deno_dist/middleware/etag/index.ts
@@ -22,6 +22,9 @@ export const etag = (options: ETagOptions = { weak: false }): MiddlewareHandler 
       c.res = new Response(null, {
         status: 304,
         statusText: 'Not Modified',
+        headers: {
+          Etag: etag,
+        },
       })
       c.res.headers.delete('Content-Length')
     } else {
--- a/deno_dist/middleware/etag/index.ts
+++ b/deno_dist/middleware/etag/index.ts
@@ -23,7 +23,7 @@ export const etag = (options: ETagOptions = { weak: false }): MiddlewareHandler 
         status: 304,
         statusText: 'Not Modified',
         headers: {
-          Etag: etag,
+          ETag: etag,
         },
       })
       c.res.headers.delete('Content-Length')
--- a/src/middleware/etag/index.test.ts
+++ b/src/middleware/etag/index.test.ts
@@ -86,6 +86,7 @@ describe('Etag Middleware', () => {
     })
     res = await app.request(req)
     expect(res.status).toBe(304)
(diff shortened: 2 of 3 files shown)


src/middleware/etag/index.test.ts
```diff
@@ -86,6 +86,7 @@ describe('Etag Middleware', () => {
     })
     res = await app.request(req)
     expect(res.status).toBe(304)
+    expect(res.headers.get('Etag')).toBe(etag)
```
navtoj, line 89:
```suggestion
    expect(res.headers.get('ETag')).toBe(etag)
```

I know http headers are case-insensitive but just for consistency in the codebase.

deno_dist/middleware/etag/index.ts
```diff
@@ -22,6 +22,9 @@ export const etag = (options: ETagOptions = { weak: false }): MiddlewareHandler
       c.res = new Response(null, {
         status: 304,
         statusText: 'Not Modified',
+        headers: {
+          Etag: etag,
```
navtoj, line 26: https://github.com/honojs/hono/pull/973/files#r1133260865

navtoj (approved):
Everything looks good! 👍🏽

src/middleware/etag/index.test.ts
yusukebe: Wow, that's my mistake!

yusukebe:
@navtoj 

Thank you very much!

PR #982, opened 2023-03-14
Allow validator to apply validation to path param.

usualoma:
How about allowing the validator to also apply validation to the path param?

We can still write the following, which is a helpful way to do it.
```
app.get('/users/:id{[0-9]+}/books/:title', handler);
```

However, validating with a validator in the following cases would be good.
* Should use 400 instead of 404.
* Should be validated by something that regular expressions cannot express.
* Should be changed to Number type.

### Type?

I tried the following changes but couldn't get the https://github.com/usualoma/hono/blob/1a3f2f62d6dece59e4c5e90b5ed21f0388d8b4d9/src/validator/validator.test.ts#L96-L150 test to pass and gave up. Although we cannot reference the path parameter type (generated from `:id`) in ValidationFunction, this PR change would be a useful enough feature.

```diff
diff --git a/src/validator/validator.ts b/src/validator/validator.ts
index c38ad05..b5d27a5 100644
--- a/src/validator/validator.ts
+++ b/src/validator/validator.ts
@@ -7,9 +7,9 @@ type ValidationTargetByMethod<M> = M extends 'get' | 'head' // GET and HEAD requ
   ? Exclude<keyof ValidationTargets, ValidationTargetKeysWithBody>
   : keyof ValidationTargets
 
-export type ValidationFunction<InputType, OutputType, E extends Env = {}> = (
+export type ValidationFunction<InputType, OutputType, E extends Env = {}, P extends string = any> = (
   value: InputType,
-  c: Context<E>
+  c: Context<E, P>
 ) => OutputType | Response | Promise<Response>
 
 export const validator = <
@@ -29,7 +29,7 @@ export const validator = <
   E extends Env = any
 >(
   target: U,
-  validationFunc: ValidationFunction<InputType, OutputType, E>
+  validationFunc: ValidationFunction<InputType, OutputType, E, P>
 ): MiddlewareHandler<E, P, V> => {
   return async (c, next) => {
     let value = {}
```

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -366,6 +366,7 @@ export type ValidationTargets = {
   form: Record<string, string | File>
   query: Record<string, string>
   queries: Record<string, string[]>
+  param: Record<string, string>
 }
 
 ////////////////////////////////////////
--- a/src/validator/validator.test.ts
+++ b/src/validator/validator.test.ts
@@ -327,6 +327,43 @@ describe('Validator middleware with Zod validates queries params', () => {
   })
 })
 
+describe('Validator middleware with Zod validates param', () => {
+  const app = new Hono()
+
+  const schema = z.object({
+    id: z
+      .string()
+      .regex(/^[0-9]+$/)
+      .transform((v) => {
+        return Number(v)
+      }),
+    title: z.string(),
+  })
+  app.get('/users/:id/books/:title', zodValidator('param', schema), (c) => {
+    const param = c.req.valid('param')
+    return c.jsonT({
+      param: param,
+    })
+  })
+
+  it('Should validate Form data and return 200 response', async () => {
+    const res = await app.request('http://localhost/users/123/books/Hello')
+    expect(res.status).toBe(200)
+    expect(await res.json()).toEqual({
+      param: {
+        id: 123,
+        title: 'Hello',
+      },
+    })
+  })
+
+  it('Should validate Form data and return 400 response', async () => {
+    const res = await app.request('http://localhost/users/0.123/books/Hello')
+    expect(res.status).toBe(400)
+    expect(await res.text()).toBe('Invalid!')
+  })
+})
+
 describe('Validator middleware with Zod multiple validators', () => {
   const app = new Hono<{ Variables: { id: number } }>()
   const route = app.post(
--- a/src/validator/validator.ts
+++ b/src/validator/validator.ts
@@ -58,6 +58,9 @@ export const validator = <
       case 'queries':
         value = c.req.queries()
         break
+      case 'param':
+        value = c.req.param() as Record<string, string>
+        break
     }
 
     const res = validationFunc(value as InputType, c)


yusukebe:
Hi @usualoma !

Good suggestion! But there were some bugs about inferring types. I've fixed these bugs in PR #983 

Shall we discuss this PR after merging the current "next" branch?

usualoma:
@yusukebe Thanks!
I have merged "next" in ad00c7d.

However, keeping this PR as a small change is better because the current code that does not pass the path as a template parameter works well when a validation function is defined separately, as shown below.

https://github.com/usualoma/hono/blob/1a3f2f62d6dece59e4c5e90b5ed21f0388d8b4d9/src/validator/validator.test.ts#L99-L103

yusukebe:
@usualoma 

How about this though it will be verbose?:

```diff
diff --git a/src/validator/validator.ts b/src/validator/validator.ts
index 0ed7065..60f524e 100644
--- a/src/validator/validator.ts
+++ b/src/validator/validator.ts
@@ -7,10 +7,12 @@ type ValidationTargetByMethod<M> = M extends 'get' | 'head' // GET and HEAD requ
   ? Exclude<keyof ValidationTargets, ValidationTargetKeysWithBody>
   : keyof ValidationTargets

-export type ValidationFunction<InputType, OutputType, E extends Env = {}> = (
-  value: InputType,
-  c: Context<E>
-) => OutputType | Response | Promise<Response>
+export type ValidationFunction<
+  InputType,
+  OutputType,
+  E extends Env = {},
+  P extends string = string
+> = (value: InputType, c: Context<E, P>) => OutputType | Response | Promise<Response>

 export const validator = <
   InputType,
@@ -18,6 +20,7 @@ export const validator = <
   M extends string,
   U extends ValidationTargetByMethod<M>,
   OutputType = ValidationTargets[U],
+  P2 extends string = P,
   V extends {
     in: { [K in U]: unknown extends InputType ? OutputType : InputType }
     out: { [K in U]: OutputType }
@@ -32,7 +35,8 @@ export const validator = <
   validationFunc: ValidationFunction<
     unknown extends InputType ? ValidationTargets[U] : InputType,
     OutputType,
-    E
+    E,
+    P2
   >
 ): MiddlewareHandler<E, P, V> => {
   return async (c, next) => {
@@ -67,7 +71,7 @@ export const validator = <
         break
     }

-    const res = validationFunc(value as never, c)
+    const res = validationFunc(value as never, c as never)

     if (res instanceof Response || res instanceof Promise) {
       return res
```

usualoma:
@yusukebe 
Oh wow, this is great! This looks good to me.

If we include this change, you should be the author; what should we do? Would you like to add it in another PR or add the commit to this branch?

yusukebe:
@usualoma 

No. The author of this PR is you! So apply the above diff. Then, I will make myself as "Co-author" when I merge it.

usualoma:
@yusukebe 
Thank you, I understand.
Applied the patch and denoified.

yusukebe:
Thank you for understanding. I've merged!

PR #984, opened 2023-03-14
feat(validator): `query` supports array params

yusukebe:
This PR enables the validator to handle array params with `query`, not using `queries`.

For example, before this PR, we could not validate multiple params like `tag` using `query`:

```
http://localhost/search?page=123&tag=a&tag=b
```

Or we have to use `queries` but if we do so, they all will be `array`.

With this PR, the values we could get from `query` will be `Record<string, string | string[]>`. So we can write the followings:

```ts
const app = new Hono()

const schema = z.object({
  q: z.string(),
  tag: z.array(z.string()),
})

const route = app.get('/post', zValidator('query', schema), (c) => {
  const { q, tag } = c.req.valid('query')
  return c.jsonT({
    queryString: q,
    tags: tag,
  })
})
```

This issue is mentioned by @equt in our Discord. Thanks.

### DIFF
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -116,7 +116,7 @@ describe('Basic - query, queries, form, and path params', () => {
     .get(
       '/search',
       validator('query', () => {
-        return {} as { q: string }
+        return {} as { q: string; tag: string[] }
       }),
       (c) => {
         return c.jsonT({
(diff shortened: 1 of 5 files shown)


yusukebe:
Thank you for reviewing. If you have any further suggestions, please let us know anytime!

PR #498, opened 2022-09-05
perf(req): improve `c.req.query` performance

yusukebe:
I've improved `c.req.query` performance. Currently, `c.req.query` uses `new URL` for parsing query parameters strings. It is inefficient. In this PR, `c.req.query` uses just `URLSearchParams`, it becomes to be faster.

## Benchmarks

Scripts:

```ts
import { run, bench } from 'mitata'
import { getQueryStringFromURL } from '../../src/utils/url'

const urlString = 'http://localhost?foo=bar&foo2=bar2&foo3=bar3'

{
  bench('URL', () => {
    const url = new URL(urlString)
    url.searchParams.get('foo')
  })
}

{
  bench('getQueryStringFromURL + URLSearchParams', () => {
    const queryStirng = getQueryStringFromURL(urlString)
    const searchParams = new URLSearchParams(queryStirng)
    searchParams.get('foo')
  })
}

await run()
```

Results:

```plain
MacBook@yusuke $ bun run ./url-params.ts
cpu: Apple M1 Pro
runtime: bun 0.1.11 (arm64-darwin)

benchmark                                    time (avg)             (min … max)       p75       p99      p995
------------------------------------------------------------------------------- -----------------------------
URL                                        1.18 µs/iter   (933.55 ns … 3.21 µs)   1.14 µs   3.21 µs   3.21 µs
getQueryStringFromURL + URLSearchParams  802.39 ns/iter   (696.65 ns … 2.35 µs) 795.49 ns   2.35 µs   2.35 µs

---

MacBook@yusuke $ deno run -A ./deno.ts
cpu: Apple M1 Pro
runtime: deno 1.25.1 (aarch64-apple-darwin)

benchmark                                    time (avg)             (min … max)       p75       p99      p995
------------------------------------------------------------------------------- -----------------------------
URL                                        2.62 µs/iter     (2.53 µs … 2.67 µs)   2.64 µs   2.67 µs   2.67 µs
getQueryStringFromURL + URLSearchParams    1.33 µs/iter      (1.3 µs … 1.34 µs)   1.33 µs   1.34 µs   1.34 µs
```

In the real world:

```ts
import { Hono } from '../../src/hono'

const app = new Hono()

app.get('/', (c) => {
  const q = c.req.query('q')
  return c.text(q)
})

export default app
```

Results:

Bun

```plain
---
v2.1.3

MacBook@yusuke $ bombardier -c 200 -d 10s 'http://localhost:3000/?q=search-query&foo=bar&foo2=bar2'
[...]

PR #988, opened 2023-03-16
feat(req): introduce `c.req.path`

yusukebe:
This PR introduces the property `path` for HonoRequest.

This simple feature allows you to get the URL's path by accessing `c.req.path`.

```ts
app.get('/search', (c) => c.text(c.req.path)) // "search"
```

This feature is what many users wanted. This is good because there is no performance degradation, only a slight increase in file size. There are some API changes, but they are used internally, so there are no problems.

### DIFF
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -59,6 +59,7 @@ type ContextOptions<E extends Env> = {
   env: E['Bindings']
   executionCtx?: FetchEvent | ExecutionContext | undefined
   notFoundHandler?: NotFoundHandler<E>
+  path?: string
   paramData?: Record<string, string>
 }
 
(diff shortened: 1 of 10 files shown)


usualoma:
Hi @yusukebe!

I think `c.req.path` is a nice feature!

As I mentioned in my comment below, I believe that "usually clients do not send fragments" and "even if an evil client sends them, there is no need to cut them," but should we cut them?
https://github.com/honojs/hono/pull/498#pullrequestreview-1096123598

yusukebe:
@usualoma 

Thanks for your comment.

You are right! We don't have to cut fragments, and we can remove `removeFragment` from `url.ts`. Great!

yusukebe:
PS.

Previously `getQueryParam()` cut fragments, but there is no need for that, and `URLSearchParams` does not cut fragments either.

yusukebe:
I'll merge it.

usualoma:
👍

PR #998, opened 2023-03-19
feat: Enable to add subApp with basePath.

### LINKED ISSUE #996: `basePath` incompatible with `app.route()`
The following deno script throws a typescript error (works fine without `basePath`)

```ts
import { Hono } from "[hono/mod.ts](https://deno.land/x/hono@v3.1.0/mod.ts)";

const mainApp = new Hono();
mainApp.get("/foo", (c) => c.text("bar"));

const subApp = new Hono().basePath("/sub");
subApp.get("/bar", (c) => c.text("baz"));

mainApp.route("/sub", subApp);
/*
Argument of type 'Hono<Env, {}, "/sub">' is not assignable to parameter of type 'Hono<Env, {}, "">'.
  The types of 'route(...).use' are incompatible between these types.
    Type 'MiddlewareHandlerInterface<Env, Record<string, unknown>, "/sub">' is not assignable to type 'MiddlewareHandlerInterface<Env, Record<string, unknown>, "">'.
      Types of parameters 'handlers' and 'handlers' are incompatible.
        Types of parameters 'c' and 'c' are incompatible.
          Type 'Context<Env, `/sub/${string}`, {}>' is not assignable to type 'Context<Env, string, {}>'.
            Types of property 'req' are incompatible.
              Type 'HonoRequest<`/sub/${string}`, unknown>' is not assignable to type 'HonoRequest<string, unknown>'.
                Type 'string' is not assignable to type '`/sub/${string}`'.deno-ts(2345)
*/
```

This is not that big of a deal (maybe even not intended to work?), but would be useful when navigating my code and for RPC across files 
(Will probably have to see if the client even supports it, as I don't know if it was built with multiple `app`s in mind, but can cross that bridge later!)

---

If this is something that could be made to work, it should also be possible to enforce a correct `basePath` when using `.route`
```ts
const mainApp = new Hono();

const subApp = new Hono().basePath("/sub");

/* TS error */
mainApp.route("/notsub", subApp);

/* no error */
mainApp.route("/sub", subApp);
```

usualoma:
This PR fixes #996.

### DIFF
--- a/deno_dist/hono.ts
+++ b/deno_dist/hono.ts
@@ -134,18 +134,18 @@ export class Hono<
   private notFoundHandler: NotFoundHandler = notFoundHandler
   private errorHandler: ErrorHandler = errorHandler
 
-  route<SubPath extends string, SubEnv extends Env, SubSchema>(
+  route<SubPath extends string, SubEnv extends Env, SubSchema, SubBasePath extends string>(
     path: SubPath,
-    app: Hono<SubEnv, SubSchema>
+    app: Hono<SubEnv, SubSchema, SubBasePath>
   ): Hono<E, RemoveBlankRecord<MergeSchemaPath<SubSchema, SubPath> | S>, BasePath>
   /** @deprecated
    * Use `basePath` instead of `route` with one argument.
    * The `route` with one argument has been removed in v4.
    */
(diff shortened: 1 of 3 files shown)


yusukebe:
Hi @usualoma !

Looks good! I'll merge it soon and ship a patch release that includes this fix.

Thanks!

PR #1008, opened 2023-03-22
perf: Reduce await.

usualoma:
This change will improve the results of the following benchmarks
https://github.com/SaltyAom/bun-http-framework-benchmark

### DIFF
--- a/deno_dist/hono.ts
+++ b/deno_dist/hono.ts
@@ -239,36 +239,33 @@ export class Hono<
 
       try {
         res = handler(c, async () => {})
+
         if (!res) {
           return this.notFoundHandler(c)
         }
+
+        if (res instanceof Response) return res
+
+        if ('response' in res) {
+          res = res.response
+        }
+
+        if (res instanceof Response) return res
+
+        return res.then((resolvedRes) => {
+          if (resolvedRes !== undefined && 'response' in resolvedRes) {
+            resolvedRes = resolvedRes['response'] as Response
+          }
+          if (!resolvedRes) {
+            return this.notFoundHandler(c)
+          }
+          return resolvedRes
+        }).catch((err) => {
+          return this.handleError(err, c)
+        })
       } catch (err) {
         return this.handleError(err, c)
       }
-
-      if (res instanceof Response) return res
-
-      if ('response' in res) {
-        res = res.response
-      }
-
-      if (res instanceof Response) return res
-
-      return (async () => {
-        let awaited: Response | TypedResponse | void
-        try {
-          awaited = await res
-          if (awaited !== undefined && 'response' in awaited) {
-            awaited = awaited['response'] as Response
-          }
-          if (!awaited) {
-            return this.notFoundHandler(c)
-          }
-        } catch (err) {
-          return this.handleError(err, c)
-        }
-        return awaited
-      })()
     }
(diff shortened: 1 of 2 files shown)


usualoma:
Oh, did I do something wrong?
I don't have a good reproduction environment at hand right now, so please wait a day or so.

yusukebe:
> Oh, did I do something wrong?

Hmm, it looks right to me. But `hono-next` may win against `hono-3.1.2`, but never against Elysia.

By the way, for `hono-next` shown above, it is importing the `Hono` using the following method.

* Pull this PR into my local Hono project.
* `yarn build`
* Edit the source in the benchmark project.
* Import from the `dist` directory.

```ts
import { Hono } from "../../honojs/hono/dist/hono"
import { RegExpRouter } from "../../honojs/hono/dist/router/reg-exp-router"
```

Importing `.ts` files in `src` directly is very slow. There may be a transpile overhead.

> I don't have a good reproduction environment at hand right now, so please wait a day or so.

Of course! Anytime is OK!

usualoma:
@yusukebe 

I am very sorry.
After trying several patterns, I have concluded that this PR does not improve performance.
Merging this PR would reduce "await" but not simplify the code, so closing it is a good idea.

### benchmark

#### 3.0.5 vs. 3.1.2

The bun-http-framework-benchmark results have improved from 3.0.5 -> 3.1.2.

3.0.5 had the following line
https://github.com/honojs/hono/blob/v3.0.5/src/context.ts#L230

The following changes resulted in the same level of performance as in 3.1.2.

```
index a928f04..ebf5560 100644
--- a/src/context.ts
+++ b/src/context.ts
@@ -227,7 +227,7 @@ export class Context<
 
   json = <T = object>(
     object: T,
-    status: StatusCode = this._status,
+    status: StatusCode,
     headers?: HeaderRecord
   ): Response => {
     const body = this._pretty
```

The default values for the arguments were causing the performance degradation, and since 3.1.2 no longer has these default values, the benchmark results are better.

#### elysia, v3.1.2 and this PR

Here are the results I got by running the bench.mjs included in bun-http-framework-benchmark with the following changes

```
	const frameworks = ['stricjs', 'hono', 'hono-with-pr', 'elysia', 'hono-with-pr', 'hono']
```

|  Framework       |  Get (/)    |  Params, query & header | Post JSON  |
| ---------------- | ----------- | ----------------------- | ---------- |
| stricjs | 172,007.58 | 116,555.76 | 98,109.85 |
| hono | 168,853.04 | 126,484.1 | 91,333.92 |
| hono-with-pr | 166,639.34 | 125,667.53 | 90,332.28 |
| elysia | 168,042.83 | 124,095.01 | 88,349.93 |
| hono-with-pr | 167,502.24 | 116,297.98 | 87,698.79 |
| hono | 166,862.86 | 116,142.15 | 87,505.12 |

This PR will not improve performance. And there appears to be no difference between v3.1.2 and elysia.

The bun-http-framework-benchmark benchmark is measured by v3.0.x, which may be why it is losing to elysia in POST.
https://github.com/SaltyAom/bun-http-framework-benchmark/blob/main/package.json#L36

yusukebe:
Great job. However, we are closing it due to no improvement in performance. Thanks!

PR #1009, opened 2023-03-23
feat(adapter): Added aws-handler support for APIGatewayProxyEventV2

schonert:
Super exicted about the slim adapter! Got the chance to start using it yesterday and ran into some missing mappings when calling lambda directly through the function URL.

- Renamed `APIGatewayEvent` to `APIGatewayProxyEvent` to follow the convention from `@types/aws-lambda`
- Added `APIGatewayEventV2`. The type used by function urls
- Added support for sending files through streams. Ran into issues while using with `serveStatic`. Files will be converted to `base64`.
- Broke the handler into smaller functions. It became quite bloated when accounting for base64 and event mapping

Submitted directly through github.com - will catch up on any red flags 👍


schonert:
All tests are green 👍

src/adapter/aws-lambda/handler.ts
```diff
@@ -24,41 +39,94 @@ interface APIGatewayProxyResult {
...
   const requestInit: RequestInit = {
-    headers: headers,
-    method: method,
+    headers,
+    method: event.httpMethod,
   }
 
   if (event.body) {
-    requestInit.body = event.isBase64Encoded ? atob(event.body) : event.body
+    requestInit.body = event.isBase64Encoded ? Buffer.from(event.body, 'base64') : event.body
```
yusukebe, line 93: Use `atob`/`btoa` instead of `Buffer`. 

We can use `Buffer` on AWS Lambda with Node.js, but the Hono core consistently uses `btoa`/`atob` instead of `Buffer`, so use that.

src/adapter/aws-lambda/handler.ts
```diff
@@ -24,41 +39,94 @@ interface APIGatewayProxyResult {
...
+    .join('&')
+}
+
+function isProxyEventV2(
+  event: APIGatewayProxyEvent | APIGatewayProxyEventV2
+): event is APIGatewayProxyEventV2 {
+  return Object.prototype.hasOwnProperty.call(event, 'rawPath')
+}
+
+async function fromBufferTostring(res: Response) {
```
yusukebe, line 116: You may use `bufferToString` in `utils/buffer.ts`.

https://github.com/honojs/hono/blob/27a966916165f4fbc24d58e2e8554e55c1cab8ed/src/utils/buffer.ts#L43

src/adapter/aws-lambda/handler.ts
```diff
@@ -24,41 +39,94 @@ interface APIGatewayProxyResult {
...
-      result.headers[key] = value
-    })
-
-    return result
+
+    return createResult(res)
+  }
+}
+
+async function createResult(res: Response): Promise<APIGatewayProxyResult> {
```
yusukebe, line 56: Please use arrow functions.

```ts
const createResult = (res: Response) => {} //...
```

src/adapter/aws-lambda/handler.ts
```diff
@@ -24,41 +39,94 @@ interface APIGatewayProxyResult {
...
+
   return new Request(url, requestInit)
 }
+
+function extractQueryString(event: APIGatewayProxyEvent | APIGatewayProxyEventV2) {
+  if (isProxyEventV2(event)) {
+    return event.rawQueryString
+  }
+
+  return Object.entries(event.queryStringParameters)
```
yusukebe, line 104: Include what to do if `queryStringParameters` is falsy. When I tried this adapter, I got the following error.

SS

yusukebe:
Hi @schonert !

Thank you for creating PR! I am glad you wish Lambda adapter for Hono.

I commented on several points of concern. And I also get the following error in the test. How about making `queryStringParameters` optional and checking it in the adapter?

SS

yusukebe:
@schonert 

Sorry! The `package.json` was wrong, and I've fixed it. So, please merge `main` into your branch.

schonert:
@yusukebe 

Sorry about the hold-up. Merged main into PR branch 👍

schonert:
I see the tests did not pass - strange since I recall running before submitting. I'll take a look

src/adapter/aws-lambda/handler.ts
```diff
@@ -24,41 +40,94 @@ interface APIGatewayProxyResult {
...
+  return bufferToString(arrayBuffer)
+}
+
+const fromReadableToString = async (res: Response) => {
+  const chunks = []
+  const stream = res.body || new ReadableStream()
+
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore: asking for asyncIterator
+  for await (const chunk of stream) {
```
yusukebe, line 128: We can use `TextDecoder()`:

```ts
  const decoder = new TextDecoder()
  let string = ''

  // @ts-ignore: asking for asyncIterator
    string += decoder.decode(chunk)
  }
```

src/adapter/aws-lambda/handler.ts
```diff
@@ -24,41 +40,94 @@ interface APIGatewayProxyResult {
...
-    })
-
-    return result
+
+    return createResult(res)
+  }
+}
+
+const createResult = async (res: Response): Promise<APIGatewayProxyResult> => {
+  const isBase64Encoded = res.body instanceof ReadableStream
```
yusukebe, line 58: I'm not sure about this, and should look into it, but `res.body` will always be a `ReadableStream`? What do you think?

yusukebe (commented):
@schonert 

Do you run `yarn test:lambda`?

And I've leave some comments.

schonert:
@yusukebe did you get the chance to check out my comments?

yusukebe:
Hi @schonert !

Which are your comments? I've left the comments on your code, so please check it again!

schonert:
@yusukebe i simply replied to your comments

https://github.com/honojs/hono/pull/1009#discussion_r1160327315
https://github.com/honojs/hono/pull/1009#discussion_r1160328051

yusukebe:
@schonert 

Did you leave comments? I can't see any of your comments.

schonert:
Hmm strange. I've tagged you and linked? Attached a screenshot below 👇 
image

yusukebe:
Hi @schonert !

Yeah, it's strange, but now I understand your comments well, with the help of your screenshots. Thanks a lot!

According to the Web API specs, the `Response` body is always a `ReadableStream`, even if it's plain text. So I think we don't have to check if it's a `ReadableStream` or not. How about writing it like this?:

```ts
const body = await fromReadableToString(res) // always use `fromReadableToString`

const result: APIGatewayProxyResult = {
  //...
  isBase64Encoded: true, // it must be base64
}
```

We'll have to rewrite the tests, but I think this approach might be the best.

yusukebe:
@schonert 

> So we just simply always base64?

Absolutely! So, could you please rewrite the code and tests? We can run the CI again once you've completed the changes.

schonert:
@yusukebe finally got around to implementing the changes above. 
Been using Github codespace - which was an awful slow experience. Should have just cloned the project down from day one 😅

yusukebe:
Hi @schonert !

Great! It's time to merge! Thanks!

Issue #1005, opened 2023-03-20
Feature Request: Get output of JWT payload


Skye-31:
In my application, we need the payload from the validated JWT, however the util currently just returns `true` if it's been validated. This is a feature request to either:

- Allow it to return the payload (this might be breaking though?)
- Put it in the context, using `c.set()` with a key that the user defines as one of the options

This would allow people to use the JWT module more effectively 🙂

yusukebe:
Hi @Skye-31 !

If we will do it, I would use `c.set()`. We will consider supporting it. Thanks!

Skye-31:
Awesome, let me know if there's anything I can do to help 😄

yusukebe:
Hi @Skye-31 !

Since "v3.1.5", we can get the payload with `c.get('jwtPayload')`:

```ts
import { Hono } from 'hono'
import { jwt } from 'hono/jwt'

const app = new Hono()

app.post(
  '/auth/abc',
  jwt({
    secret: 'a-secret',
  }),
  (c) => {
    const payload = c.get('jwtPayload')
    // ...
  }
)
```

Try it!

Skye-31:
Awesome, thanks!

PR #1013, opened 2023-03-26
feat(jwt): support for getting payload from `c.get('jwtPayload')`

### LINKED ISSUE #810: Support for decoding JWT
I wonder if the JWT middleware should have the option to decode the token and push it onto the context after verifying?

yusukebe:
This PR will add support for getting the decoded payload from `c.get('jwtPaload')` for JWT Auth Middleware.

```ts
app.post(
  '/auth/abc',
  jwt({
    secret: 'a-secret',
  }),
  (c) => {
    const payload = c.get('jwtPayload')
    //...
  }
)
```

This will resolve #810 #1005 

### Does it include breaking changes?

The `verify()` API in `utils/jwt/jwt.ts` has been changed. It will return `any` instead of `boolean`. This seems to be  a breaking change, but this function is mainly "utility" used in the Hono internal. We should release the fix, including breaking changes as major version up, but this change will not affect end users. EDIT ~~I'd like to release it as a minor version up "3.2.0".~~

EDIT:

We can release this with a patch release.

* Utilities such as `utils/url.ts` are broken and changed without notice. utilities are intended to be used internally and may be changed in patch releases. Same for `utils/jwt/jwt.ts.
* It is not a major feature addition, and we don't have to include it in the Minor or Major release.

### DIFF
--- a/benchmarks/query-param/package-lock.json
+++ b/benchmarks/query-param/package-lock.json
@@ -1,0 +1,519 @@
+{
+  "name": "query-param",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "license": "MIT",
+      "dependencies": {
+        "fast-querystring": "^1.1.1",
+        "mitata": "^0.1.6"
+      },
+      "devDependencies": {
+        "tsx": "^3.12.2"
+      }
+    },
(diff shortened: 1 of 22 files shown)


JoshZA:
Thanks for this

PR #1023, opened 2023-03-31
fix(utils/url): Fix behavior when parameters with no value specified are mixed.

### LINKED ISSUE #1022: hono 3.1.4 brings new query utils which breaks some query use cases
Let's see the case:

`?pretty&foo=bar` should be resolved with `{ foo: 'bar', pretty: '' }` or `{ foo: 'bar', pretty: undefined }`.
> Which I think the former one is more precise.

screenshot

But with current hono implementations, this resolved as `{ 'pretty$foo': 'bar' }`
screenshot

usualoma:
This PR fixes #1022

There is almost no performance degradation (In particular, there is no degradation for `_getQueryParam(url, key)`.)


yusukebe:
Hi @usualoma !

Looks good to me!

Super thanks for the quick fix.

PR #1036, opened 2023-04-23
feat: introduce PatternRouter and `hono/tiny`

yusukebe:
In this PR, I'll introduce **URLPatternRouter** and `hono/tiny`.

This feature, if merged, is experimental and subject to API changes in the future.

### Motivation

Hono is small. The minimal "Hello World" application using latest "v3.1.6" is **20.6KB** when minified.

SS

About half of that code consists of router implementations. Therefore, I thought we could create a router using `URLPattern` to reduce its size. It would be great if this approach allowed users to choose between a faster version or a smaller yet slower alternative.

### URLPatternRouter

**URLPatternRouter** is implemetend with `URLPattern`.

> The URLPattern interface matches URLs or parts of URLs against a pattern. The pattern can contain capturing groups that extract parts of the matched URL.

https://developer.mozilla.org/en-US/docs/Web/API/URLPattern

`URLPattern` is only available in Cloudflare Workers and Deno. It is already implemented in the runtime, so we don't need to write additional code for matching URL paths with patterns.

We can use URLPatternRouter in the same way as RegExpRouter, TrieRouter, or SmartRouter:

```ts
const app = new Hono({
  router: new URLPatternRouter(),
})
```

### `hono/tiny`

Hono uses a combination of "SmartRouter, RegExpRouter, and TrieRouter" by default.

```ts
readonly router: Router<H> = new SmartRouter({
})
```

This awesome idea to keep it fast and support all routing patterns. However, if you want to use URLPatternRouter, you are forced to import these routers, as you cannot import only URLPatternRouter. This is because these routers are hard-coded in the constructor of `hono.ts`. This design is not ideal. Ideally, if you want to use only URLPatternRouter, it should allow you to import just URLPatternRouter.

Then, I made `hono-base.ts`. It's like an abstract class which has not a really router. So, we have to use it with concrete router implemantation. Belows are new `hono.ts` as default.

```ts
export class Hono extends HonoBase {
  constructor(init: Partial<Pick<Hono, 'router' | 'strict'>> = {}) {
    super(init)
    this.router = new SmartRouter({
    })
  }
}
```

`hono/tiny` uses only URLPatternRouter along with the `hono-base.ts` mechanize. You can use it just like `hono` and It will be very small.

```ts
import { Hono } from 'hono/tiny'

const app = new Hono()

app.get('/', (c) => {
  return c.text('Hello hono/tiny!')
})
```

### Really small

I measured the size, and it is **13.06KB**! It's indeed very small.

SS

### Benchmark

I took the benchmark. `URLPattern` is not available in Node.js and Bun, so I ran it on Deno. As expected, it is very slow.

SS

Nevertheless, this router is featured by its small size, so it is okay if it is slow.

### Implementation

`URLPattern` has several features not found in Hono's other routers. If we allow those features to be used, other routers may have to support them. Therefore, it throws an Error when it tries to use those functions.

---

I like this feature. The implementation may be improved.

---

### Real world benchmarks

I ran the application using both `hono` and `hono/tiny` on Deno, and measured their performance with `bombardier`.

```ts
app.get('/', (c) => {
  return c.text('Hello hono/tiny!')
})

app.get('/user', (c) => {
  return c.text('Hello User')
})

app.get('/user/:name', (c) => {
  return c.text(`Hello ${c.req.param('name')}`)
})
```

With `hono` default:

SS

With `hono/tiny`:

SS


yusukebe:
Hi @usualoma !

What do you think about this feature and implementation?

usualoma:
@yusukebe 
I think hono/tiny's idea is excellent. Please wait a bit, as I'll look at a few more details before commenting.

usualoma:
@yusukebe 
The hono/tiny idea is excellent, and the addition of HonoBase is good.

Is using URLPattern in the implementation the best way to go here? In the PR implementation, I think URLPattern is used to generate regular expressions.

Such a regular expression can be generated in equally short code without URLPattern.
https://github.com/honojs/hono/pull/1039

We can make hono/tiny more hono-like by building a RegExp object ourselves without using URLPattern, which can also be used in node and bun.

yusukebe:
Hi @usualoma !

Thank you. I just wanted to use the API that I've never used before. As you mentioned, there's no need to utilize `URLPattern`, and your implementation is excellent! Moreover, it will result in smaller and efficient solution compared to using `URLPattern`.

SS

Furthermore, this approach proves to be faster than relying on `URLPattern`.

I'm in favor of proceeding with #1039. Please submit a PR to this one, and I'll merge it. After that, we can merge this PR into the "next" branch, with you credited as a co-author.

usualoma:
I added some refactoring and undrafted #1039. The name is still URLPatternRouter, but I'm not sure what to do with it.

32d3d93
3cd4b06

yusukebe:
@usualoma 

Thanks! So, we need to select a new name. Since `URLPattern` is no longer used, "URLPatternRouter" is no longer appropriate. How about make it simple and going with "**PatternRouter**"?

yusukebe:
@usualoma 

Thanks! Great!

yusukebe:
Thanks to @usualoma, we now have a really small router named "**PatternRouter**" that doesn't require `URLPattern`! #1039
Also, creating `hono/tiny` is a good idea. Let's go with it!"

PR #1056, opened 2023-05-01
empty string is a valid header value

AlexErrant:
As mentioned in [this PR](https://github.com/honojs/hono/pull/884#issuecomment-1426925215), empty strings are valid header values.

### DIFF
--- a/deno_dist/request.ts
+++ b/deno_dist/request.ts
@@ -69,14 +69,14 @@ export class HonoRequest<P extends string = '/', I extends Input['out'] = {}> {
   header(name: string): string | undefined
   header(): Record<string, string>
   header(name?: string) {
-    const headerData: Record<string, string> = {}
+    const headerData: Record<string, string | undefined> = {}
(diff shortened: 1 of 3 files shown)


deno_dist/request.ts
```diff
@@ -69,14 +69,14 @@ export class HonoRequest<P extends string = '/', I extends Input['out'] = {}> {
   header(name: string): string | undefined
   header(): Record<string, string>
   header(name?: string) {
-    const headerData: Record<string, string> = {}
+    const headerData: Record<string, string | undefined> = {}
```
AlexErrant, line 72: Without this, I'd have to add `as string | unknown` to line 79, which is ugly. Ideally we'd turn on [`"noUncheckedIndexedAccess": true`](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) in the tsconfig, but that makes _many_ things break so I consider it out of scope for this PR.

yusukebe:
Hi @AlexErrant !

Good fix and test! Thanks!

PR #1064, opened 2023-05-02
feat: Allow getPath to be replaced.

usualoma:


usualoma:
If we are willing to require migration for `strict: false` users, it may be simpler to add the following changes.

```diff
diff --git a/src/hono-base.ts b/src/hono-base.ts
index cac579e..bb28cec 100644
--- a/src/hono-base.ts
+++ b/src/hono-base.ts
@@ -19,7 +19,7 @@ import type {
   MergeSchemaPath,
 } from './types'
 import type { RemoveBlankRecord } from './utils/types'
-import { getPath, getPathNoStrict, mergePath } from './utils/url'
+import { getPath, mergePath } from './utils/url'
 
 type Methods = typeof METHODS[number] | typeof METHOD_NAME_ALL_LOWERCASE
 
@@ -64,13 +64,13 @@ class Hono<E extends Env = Env, S = {}, BasePath extends string = ''> extends de
     To use it, inherit the class and implement router in the constructor.
   */
   router!: Router<H>
-  readonly getPath: (request: Request) => string
+  readonly getPath: (request: Request) => string = getPath
   private _basePath: string = ''
   private path: string = '*'
 
   routes: RouterRoute[] = []
 
-  constructor(init: Partial<Pick<Hono, 'router' | 'getPath'> & { strict: boolean }> = {}) {
+  constructor(init: Partial<Pick<Hono, 'router' | 'getPath'>> = {}) {
     super()
 
     // Implementation of app.get(...handlers[]) or app.get(path, ...handlers[])
@@ -118,10 +118,7 @@ class Hono<E extends Env = Env, S = {}, BasePath extends string = ''> extends de
       return this
     }
 
-    const strict = init.strict ?? true
-    delete init.strict
     Object.assign(this, init)
-    this.getPath ||= strict ? getPath : getPathNoStrict
   }
 
   private clone(): Hono<E, S, BasePath> {
diff --git a/src/hono.test.ts b/src/hono.test.ts
index 4ec928b..2fb23c2 100644
--- a/src/hono.test.ts
+++ b/src/hono.test.ts
@@ -9,6 +9,7 @@ import { RegExpRouter } from './router/reg-exp-router'
 import { TrieRouter } from './router/trie-router'
 import type { Handler, Next } from './types'
 import type { Expect, Equal } from './utils/types'
+import { getPathNoStrict } from './utils/url'
 
 // https://stackoverflow.com/a/65666402
 function throwExpression(errorMessage: string): never {
@@ -111,8 +112,8 @@ describe('strict parameter', () => {
     })
   })
 
-  describe('strict is false', () => {
-    const app = new Hono({ strict: false })
+  describe('getPath: getPathNoStrict', () => {
+    const app = new Hono({ getPath: getPathNoStrict })
 
     app.get('/hello', (c) => {
       return c.text('/hello')
diff --git a/src/hono.ts b/src/hono.ts
index f6c6017..c56387d 100644
--- a/src/hono.ts
+++ b/src/hono.ts
@@ -9,7 +9,7 @@ export class Hono<E extends Env = Env, S = {}, BasePath extends string = ''> ext
   S,
   BasePath
 > {
-  constructor(init: Partial<Pick<Hono, 'router' | 'getPath'> & { strict: boolean }> = {}) {
+  constructor(init: Partial<Pick<Hono, 'router' | 'getPath'>> = {}) {
     super(init)
     this.router = new SmartRouter({
       routers: [new RegExpRouter(), new TrieRouter()],

```

yusukebe:
Thanks! I'll check it later.

jbergstroem:
Very clever!

yusukebe:
Hi @usualoma !

I've checked and measured the performance to see if there is any slowdown with this PR. But, I didn't find any deletions as you mentioned.

> If we are willing to require migration for `strict: false` users, it may be simpler to add the following changes.

This is not a bad idea, as it will reduce the number of options for the Hono constructor. However, it would introduce a breaking change, and using the strict option is simpler for users than importing getPathNoStrict. I think we don't need to implement it.

We can go with it. I'll merge it now. Great work!

By the way, I'm planning to include this feature in the next major version, v4! I'll create an issue to share the idea of v4 later on.

usualoma:
👍

PR #1072, opened 2023-05-06
feat(app): introduce `app.bundleHandlers()`

yusukebe:
This PR introduces a `bundleHandlers()` function in `hono-base.ts`. This function will **bundle a path and handlers** into `[path, ...handlers]`. The bundled handler can be used with `app.get()` and other route methods. The `bundleHandlers()` is useful for writing application code when using RPC-mode.

Without `bundleHandlers()`:

```ts
const routes = app
  .get(
    '/list',
    validator('query', () => {
      return { page: '1' }
    }),
    (c) => c.jsonT(c.req.valid('query'))
  )
  .post(
    '/create',
    validator('form', () => {
      return { title: 'Hono is cool!' }
    }),
    (c) => c.jsonT(c.req.valid('form'))
  )
```

With `bundleHandlers()`, you can write like the following:

```ts
const listHandler = app.bundleHandlers(
  '/list',
  validator('query', () => {
    return { page: '1' }
  }),
  (c) => c.jsonT(c.req.valid('query'))
)

const createHandler = app.bundleHandlers(
  '/create',
  validator('form', () => {
    return { title: 'Hono is cool!' }
  }),
  (c) => c.jsonT(c.req.valid('form'))
)

const routes = app.get(...listHandler).post(...createHandler)

//...

export type AppType = typeof routes
```

It may be more verbose, but it makes a better outlook on the code. Users can choose to use `bundleHandlers()` or chaining without it.

This is implemented as a method in `hono-base.ts`. It slightly increases the code volume, but this implementation is quite simple:

```ts
this.bundleHandlers = (path: string, ...handlers: H[]) => {
  return [path, ...handlers]
}
```

I think this API provides simplicity for users.

This will be released as "experimental" in "v3.2", and we'll gather feedback on its usage. If it proves to be effective, it will be made generally available.

### DIFF
--- a/deno_dist/hono-base.ts
+++ b/deno_dist/hono-base.ts
@@ -17,6 +17,7 @@ import type {
   TypedResponse,
   MergePath,
   MergeSchemaPath,
+  BundleHandlersInterface,
 } from './types.ts'
 import type { RemoveBlankRecord } from './utils/types.ts'
 import { getPath, getPathNoStrict, mergePath } from './utils/url.ts'
@@ -36,6 +37,8 @@ function defineDynamicClass(): {
     on: OnHandlerInterface<E, S, BasePath>
   } & {
     use: MiddlewareHandlerInterface<E, S, BasePath>
+  } & {
+    bundleHandlers: BundleHandlersInterface<E, BasePath>
   }
 } {
   return class {} as never
@@ -116,6 +119,12 @@ class Hono<E extends Env = Env, S = {}, BasePath extends string = ''> extends de
         this.addRoute(METHOD_NAME_ALL, this.path, handler)
       })
       return this
+    }
+
+    // Implementation of app.bundleHandlers(path, ...handler[])
+    this.bundleHandlers = (path: string, ...handlers: H[]) => {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      return [path, ...handlers] as any
     }
 
     const strict = init.strict ?? true
--- a/deno_dist/types.ts
+++ b/deno_dist/types.ts
@@ -302,6 +302,107 @@ type ExtractKey<S> = S extends Record<infer Key, unknown>
 
 ////////////////////////////////////////
 //////                            //////
+//////      Bundle Handlers       //////
+//////                            //////
+////////////////////////////////////////
+
+export interface BundleHandlersInterface<E extends Env, BasePath extends string> {
+  // app.bundleHandlers(path, handler)
+  /**
+   * @experimental
+   */
+  <P extends string = string, O = {}, I extends Input = {}>(
+    path: P,
+    ...handlers: [H<E, P, I, O>]
+  ): [P, H<E, MergePath<BasePath, P>, I, O>]
+
+  // app.bundleHandlers(path, handler, handler)
+  /**
+   * @experimental
+   */
+  <P extends string = string, O = {}, I extends Input = {}>(
+    path: P,
+    ...handlers: [H<E, P, I, O>, H<E, P, I, O>]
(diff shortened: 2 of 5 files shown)


yusukebe:
Hi @usualoma, @yudai-nkt!

What do you think about this feature? Let me know your thoughts. I believe releasing it as "experimental" is a good approach.

P.S.

After the "next" branch is merged into the main, I'd like to release it as "v3.2" instead of "v4". There are no critical breaking changes, only some "deprecations".

usualoma:
Hi @yusukebe !

This is an interesting attempt. However, I am concerned that it may lead to bugs if get and post are specified elsewhere.

```typescript
const listHandler = app.bundleHandlers( // This handler expects a "GET" method.
  '/list',
  validator('query', () => {
    return { page: '1' }
  }),
  (c) => c.jsonT(c.req.valid('query'))
)

const routes = app.post(...listHandler) // Hard to notice if we specify the wrong method
```

### Comparison with other approaches

We can also do the following when separating the handler definition and the AppType export.

```typescript
const app = new Hono()
const listAPI = app.get(
  '/list',
  validator('query', () => {
    return { page: '1' }
  }),
  (c) => c.jsonT(c.req.valid('query'))
)
const createAPI = app.post(
  '/create',
  validator('form', () => {
    return { title: 'Hono is cool!' }
  }),
  (c) => c.jsonT(c.req.valid('form'))
)

export type AppType = typeof listAPI | typeof createAPI
```

or 

```typescript
type GetAppType<T extends unknown[]> = T extends [infer U]
  ? U
  : T extends [Hono<infer E, infer S1, infer BasePath>, Hono<infer E, infer S2, infer BasePath>]
  ? Hono<E, S1 | S2, BasePath>
  : T extends [infer U, ...infer V]
  ? GetAppType<[U, GetAppType<V>]>
  : never
const app = new Hono()
const listAPI = app.get(
  '/list',
  validator('query', () => {
    return { page: '1' }
  }),
  (c) => c.jsonT(c.req.valid('query'))
)
const createAPI = app.post(
  '/create',
  validator('form', () => {
    return { title: 'Hono is cool!' }
  }),
  (c) => c.jsonT(c.req.valid('form'))
)

export type AppType = GetAppType<[typeof listAPI, typeof createAPI]>
```

What are the advantages of using `bundleHandlers` compared to these approaches?

yusukebe:
@usualoma

> // This handler expects a "GET" method.
> // Hard to notice if we specify the wrong method

You're right. As you mentioned, it's not safe. I think your first approach is better!:

```ts
const listAPI = app.get(
  '/list',
  validator('query', () => {
    return { page: '1' }
  }),
  (c) => c.jsonT(c.req.valid('query'))
)
```

So, this feature is not needed.

---

The reason I started thinking about this feature was that some developers might want to write their applications using a "controller" based style, similar to Ruby on Rails:

```ruby
get '/patients/:id', to: 'patients#show'
```

I know that separating routing and controller can be effective, but for Hono, this approach may cause difficulties in inferring types:

SS

SS

If we add generics, it becomes too verbose!

SS

Therefore, in Hono's case, I think it's not suitable for a "controller" style, so we should continue using an Express or Sinatra-like approach:

```ts
const postApp = new Hono()

postApp.get('/', validator('query', () => {
  return {
    page: '1'
  }
}),  (c) => {
  const { page } = c.req.valid('query')
  const posts = getPosts({ page})
  return c.jsonT({
    posts
  })
})

postApp.get(...)
postApp.post(...)
postApp.delete(...)

//

const rootApp = new Hono()
rootApp.route('/posts', postApp)
rootApp.route('/authors', authorApp)
```

usualoma:
Hmmm, it seems difficult to write like Ruby on Rails.

yusukebe:
Yes, it's too difficult or even impossible... And, we might not have to do it.

Issue #1075, opened 2023-05-07
When sending a POST to the Lambda Function URL, it doesn't work.


watany-dev:
If I write the following code in Lambda and try to POST it from FunctionURL, it fails.

```ts
import { Hono } from 'hono'
import { handle } from 'hono/aws-lambda'
import { logger } from 'hono/logger'

const app = new Hono()

interface UserBody {
  user: string;
}

app.use('*', logger())
app.post('/', async (c) => {
  const { user } = await c.req.json<UserBody>();
  if (!user) {
    return c.text('Please Setting "user"', 400)
  } 
  return c.json({"user": user})
})
app.get('/', (c) => { 
    return c.text(`Hono is running on ${c.runtime}\n`)
})
export const handler = handle(app)

```

- error

```json
{
    "errorType": "TypeError",
    "errorMessage": "Request with GET/HEAD method cannot have body.",
    "stack": [
        "TypeError: Request with GET/HEAD method cannot have body.",
        "    at new Request (node:internal/deps/undici/undici:7171:17)",
        "    at createRequest (/var/task/index.js:1570:10)",
        "    at Runtime.handler (/var/task/index.js:1540:17)",
        "    at Runtime.handleOnceNonStreaming (file:///var/runtime/index.mjs:1086:29)"
    ]
```

I suspected it might be an issue with the event, so I investigated it. (I couldn't find anything about it in the official documentation.)

```node
export const handler = async(event, context, callback) => {
    return {
        statusCode: 200,
        body: "EVENT: \n" + JSON.stringify(event, null, 2),
    };
};
```

- GET
```bash
curl -X GET -H 'Content-Type: application/json' -d '{ "usr": "uga" }' https://v6omzqz2b2ck4asqa3f6ktm26m0ogpxq.lambda-url.ap-northeast-1.on.aws/
EVENT: 
{
  "version": "2.0",
  "routeKey": "$default",
  "rawPath": "/",
  "rawQueryString": "",
  "headers": {
    "content-length": "16",
    "x-amzn-tls-cipher-suite": "ECDHE-RSA-AES128-GCM-SHA256",
    "x-amzn-tls-version": "TLSv1.2",
    "x-amzn-trace-id": "Root=1-64577867-46dd53e40ce1f6e5279c68bd",
    "x-forwarded-proto": "https",
    "host": "v6omzqz2b2ck4asqa3f6ktm26m0ogpxq.lambda-url.ap-northeast-1.on.aws",
[...]

PR #1076, opened 2023-05-07
feat(adapter): AWS Lambda Adapter supports requests via Lambda Function URLs.

watany-dev:
Lambda Function URLs are endpoints that allow for simple requests to be passed to Lambda without an API Gateway. 

https://aws.amazon.com/jp/blogs/aws/announcing-aws-lambda-function-urls-built-in-https-endpoints-for-single-function-microservices/

Most of the event is compatible with API Gateway, but the location of the method is different, which is why, as mentioned in https://github.com/honojs/hono/issues/1075, we specified the behavior of GET during a POST request. This patch attempts to fix that issue.

### DIFF
--- a/src/adapter/aws-lambda/handler.ts
+++ b/src/adapter/aws-lambda/handler.ts
@@ -85,20 +85,18 @@ const createResult = async (res: Response): Promise<APIGatewayProxyResult> => {
 
 const createRequest = (event: APIGatewayProxyEvent | APIGatewayProxyEventV2 | LambdaFunctionUrlEvent) => {
   const queryString = extractQueryString(event)
-  const urlPath = isProxyEventV2(event)
-    ? `https://${event.requestContext.domainName}${event.rawPath}`
-    : `https://${event.requestContext.domainName}${event.path}`
-
-  const url = queryString ? `${urlPath}?${queryString}` : urlPath
+  const urlPath = `https://${event.requestContext.domainName}${isProxyEvent(event) ? event.path : event.rawPath}`;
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @watany-dev !

Thank you for the PR. I'd like to try running the application with this PR in my environment. So, please wait until I've done it.

src/adapter/aws-lambda/handler.ts
```diff
@@ -83,7 +98,7 @@ const createRequest = (event: APIGatewayProxyEvent | APIGatewayProxyEventV2) =>
 
   const requestInit: RequestInit = {
     headers,
-    method: event.httpMethod,
+    method: event.httpMethod ?? event.requestContext.http?.httpMethod,
```
yusukebe, line 101: I tried it, but `event.requestContext.http?.httpMethod` is `undefined`. So, `method` must be:

```ts
const method = 'httpMethod' in event ? event.httpMethod : event.requestContext.http.method
```

yusukebe (changes requested):
Hi @watany-dev !

Thank you for the PR.

As you pointed out, POST is not working on Lambda Function URLs, so this change seems to fix it. I've left a comment; please check it!

watany-dev:
I have reviewed the workflow and understood the need to perform runtime tests. I have made the necessary modifications.

watany-dev:
Sorry, I fixed the Format.

yusukebe (approved):

yusukebe:
Thanks! I'll merge it now!

PR #1085, opened 2023-05-09
fix(getQueryParam): Fixed a bug in getQueryParam when a partial matching key was found.

### LINKED ISSUE #1084: c.req.query(  ) cause infinite loop when query has prefix of key
Infinite loop may occur in `c.req.query(key)`.

When you have a parameter (like `&d0=`) with key (like `d`) as prefix

`http://localhost/foo?a=0&b=1`: `c.req.query('d')` -> `undefined`
`http://localhost/foo?c=2&d=3`: `c.req.query('d')` -> `'3'`
`http://localhost/foo?c=2&d0=3`: `c.req.query('d')` -> **infinite loop**

if this parameter on first then return `undefined` correctly.
`http://localhost/foo?d0=3`: `c.req.query('d')` -> `undefined`

usualoma:
Fixes #1084

### DIFF
--- a/deno_dist/utils/url.ts
+++ b/deno_dist/utils/url.ts
@@ -161,7 +161,7 @@ const _getQueryParam = (
       } else if (trailingKeyCode == 38 || isNaN(trailingKeyCode)) {
         return ''
       }
-      keyIndex = url.indexOf(`&${key}`, keyIndex)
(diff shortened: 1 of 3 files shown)


yusukebe:
Thank you for the quick fix! Looks good. I'll merge it.

PR #1088, opened 2023-05-10
ErrorHandler type fix

AliKaanT:
We can use async function in app.onError method and it works correctly. But it has type error.
Return type of callback function isn't including Promise so i changed it to: 
`
(err: Error, c: Context<E>) => Response | Promise<Response>
`


yusukebe:
Hi @AliKaanT !

I would like to merge it, but some fixes and additional tests are needed. Therefore, I've created another PR #1090 based on this one. After review, I plan to merge it and will add you as a "co-author".

PR #1090, opened 2023-05-11
feat(app): `onError()` supports async

yusukebe:
As mentioned by @AliKaanT in #1088, `app.onError()` can be an asynchronous function.

```ts
app.onError(async (e, c) => {
  //...
})
```

This PR enables support for an asynchronous function in `app.onError()`. It is based on #1088, I added tests and a small fix. If this is merged, I will need to add @AliKaanT as a co-author.

### DIFF
--- a/deno_dist/compose.ts
+++ b/deno_dist/compose.ts
@@ -68,10 +68,10 @@ export const compose = <C extends ComposeContext, E extends Env = Env>(
             }
             return context
           })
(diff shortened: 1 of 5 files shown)


usualoma:
@yusukebe 
Yes, looks good to me!

yusukebe:
@usualoma Thanks!

PR #1091, opened 2023-05-11
fix(lambda): don't encode text

yusukebe:
This PR concerns the AWS Lambda adapter.

From PR #1009, it encodes any content as Base64.

This works fine on Function URLs, but it does not perform well on Lambda with API Gateway. We don't need to encode text-based content, only binary. In this PR, it determines whether the content is text or not and decides whether encoding is necessary.


yusukebe:
Hi @schonert !

We had to use a regular expression to determine whether the content is text, and if it's binary, it should be encoded in base 64. This fix should be fine. If you notice any problems, please let me know! Thanks!

PR #1100, opened 2023-05-13
feat: add devcontainer to improve developer experience

C-Dao:
Adding the .devcontainer configuration makes it easier for us to develop projects online.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,12 @@
+FROM mcr.microsoft.com/devcontainers/typescript-node:20
+
+# Install Deno
+ENV DENO_INSTALL=/usr/local
+RUN curl -fsSL https://deno.land/install.sh | sh
+
+# Install Bun
+ENV BUN_INSTALL=/usr/local
+RUN curl -fsSL https://bun.sh/install | bash
+
+# Install Lagon
+RUN yarn global add @lagon/cli esbuild
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -2,7 +2,7 @@ FROM mcr.microsoft.com/devcontainers/typescript-node:20
 
 # Install Deno
 ENV DENO_INSTALL=/usr/local
-RUN curl -fsSL https://deno.land/install.sh | sh
+RUN curl -fsSL https://gist.githubusercontent.com/LukeChannings/09d53f5c364391042186518c8598b85e/raw/ac8cd8c675b985edd4b3e16df63ffef14d1f0e24/deno_install.sh | sh
 
 # Install Bun
 ENV BUN_INSTALL=/usr/local
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,21 @@
+{
+    "build": {
+        "dockerfile": "Dockerfile"
+    },
+    "customizations": {
+        "vscode": {
+            "settings": {
+                "deno.enable": false,
+                "eslint.validate": [
+                    "javascript",
+                    "javascriptreact",
+                    "typescript",
+                    "typescriptreact"
+                ],
+                "editor.codeActionsOnSave": {
+                    "source.fixAll.eslint": true
+                }
+            }
+        }
+    }
+}


yusukebe:
Hi @C-Dao 

I must admit, I'm not very familiar with "devcontainer". Is it something you're planning to use for "GitHub Codespaces"?

Improving developer experience is always a great goal, but I also want to maintain a minimalistic file structure in the project.

If the "devcontainer" is indeed very effective, we would be open to considering it. Could you please elaborate on the advantages that the "devcontainer" provides?

C-Dao:
> Hi @C-Dao
> 
> Thank you for creating this PR.
> 
> I must admit, I'm not very familiar with "devcontainer". Is it something you're planning to use for "GitHub Codespaces"?
> 
> Improving developer experience is always a great goal, but I also want to maintain a minimalistic file structure in the project.
> 
> If the "devcontainer" is indeed very effective, we would be open to considering it.

Hi, @yusukebe, 

Thanks for your review!

The .devcontainer is used for the initial development environment. 

Yes, It can be used in "GitHub Codespaces". 

By the way, We can use it to quickly enable local development or remote development in docker by [vscode](https://code.visualstudio.com/docs/devcontainers/containers#_getting-started). 

I think hono supports so many serverless environments, so it has many steps to initial environment. 

Like installing Deno, Bun or Node.js. 

We can support initial development configuration, So that the developer can easily and quickly get started.

C-Dao:
> @C-Dao 
> 
> 
> 
> Sounds good!
> 
> 
> 
> However, I'm not currently a user of this "devcontainer" and I'm not entirely convinced of its necessity. So, let's leave this PR for now. If other developers express interest in using "devcontainer", we can revisit this discussion and decide whether to merge it.

Yes, I agree

markthree:
> @C-Dao
> 
> Sounds good!
> 
> However, I'm not currently a user of this "devcontainer" and I'm not entirely convinced of its necessity. So, let's leave this PR for now. If other developers express interest in using "devcontainer", we can revisit this discussion and decide whether to merge it.

devcontainer works, it takes care of compatibility issues on everyone's environments, and I've just successfully handled a commit through devcontainer on my windows computer!

CyberFlameGO (changes requested):
I would say most of these installations are unnecessary anyway - you don't need Deno nor Bun and it just adds codespace initialization overhead, but using .devcontainer to ensure yarn is present/up-to-date (and generally to ensure necessary development tools are present) would be an excellent use of it, as it's definitely useful

yusukebe:
Okay! I'll consider to merge this PR.

yusukebe:
Hi @C-Dao 

I would love to use it, but on my M1 Mac I get an error because Linux ARM64 binary for Deno is not available officially. Do you have any ideas?

CyberFlameGO:
> Hi @C-Dao 
> 
> 
> 
> I would love to use it, but on my M1 Mac I get an error because Linux ARM64 binary for Deno is not available officially.
> 
> 
> 
> ```
> 
> 0.466 Error: Official Deno builds for Linux aarch64 are not available. (see: https://github.com/denoland/deno/issues/1846 )
> 
> ```
> 
> 
> 
> Do you have any ideas?

According to https://docs.deno.com/runtime/manual/getting_started/installation there are M1 builds - hope this helps!

yusukebe:
Thanks! But I meant `Linux` on `arm64`. They said:

> On Linux and Windows, only x64 is supported.

I found this project:

https://github.com/LukeChannings/deno-arm64

CyberFlameGO:
Oh I see! I completely misunderstood sorry 😭 
That project seems to work and be maintained yeah - it'll require some Dockerfile adaptation to check the target arch but yeah

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM mcr.microsoft.com/devcontainers/typescript-node:20
+
+# Install Deno
+ENV DENO_INSTALL=/usr/local
+RUN curl -fsSL https://deno.land/install.sh | sh
```
CyberFlameGO, line 5:
```suggestion
RUN curl -fsSL https://gist.githubusercontent.com/LukeChannings/09d53f5c364391042186518c8598b85e/raw/ac8cd8c675b985edd4b3e16df63ffef14d1f0e24/deno_install.sh | sh
```

CyberFlameGO (changes requested):
Amend as per comment

.devcontainer/Dockerfile
yusukebe: Cool! It works on my machine.

Hey @C-Dao . What do you think this way? I'd like to go with this.

C-Dao: > Cool! It works on my machine.
> 
> Hey @C-Dao . What do you think this way? I'd like to go with this.

LGTM, thanks @CyberFlameGO!

PR #1104, opened 2023-05-15
feat: `app.mount()`

yusukebe:
This PR introduces a new feature, `mount()`.

Currently, there are many frameworks that support Web Standard APIs. They provide applications that receive a Request and return a Response:

```
Request => app => Response
```

I've found an interesting idea to utilize this unique point, which is `mount()`.

By using `mount()`, you can integrate applications using other frameworks, such as *itty-router*, with Hono.

```ts
import { Router as IttyRouter } from 'itty-router'
import { Hono } from 'hono'
import { mount } from 'hono/mounter'

// Create itty-router application
const ittyRouter = IttyRouter()

// Handle `GET /itty-router/hello`
ittyRouter.get('/hello', () => new Response('Hello from itty-router'))

// Hono application
const app = new Hono()

// Mount!
app.get(...mount('/itty-router', ittyRouter.handle))

export default app
```

And I also tried the Remix app:

```ts
import { Hono } from 'hono'
import { mount } from 'hono/mounter'

import { serveStatic } from 'hono/cloudflare-workers'

import { createRequestHandler } from '@remix-run/cloudflare'
import * as build from './build'

// Remix application
// @ts-ignore
const handleRemixRequest = createRequestHandler(build, process.env.NODE_ENV)

// Hono application
const app = new Hono()

// Static files for Remix
app.get(
  '/remix/build/*',
  serveStatic({
    root: './',
  })
)

// Mount Remix app
app.get(
  ...mount('/remix', handleRemixRequest, (c) => {
    return { env: env(c) }
  })
)
```

This means we can mount any applications built with any framework like *itty-router*, *Remix*, *Qwik*, *SoildJS* into our Hono app. We don't need to create middleware to mount a specific framework, we can do it simply with `mount()`.

With this implementation, now, we have two concepts: **adapt** and **mount**. In my view, **adapt** means "Hono adapts to any runtime", while **mount** means "Hono can mount any framework". Along with middleware, we can create a large ecosystem like the following:

Image

With these special features, Hono will not just be a web framework, it will be like a "Glue".

In any case, I believe the ability to mount applications built with other frameworks is a good idea.


yusukebe:
Hi @usualoma , @metrue , @yujong-lee , @ThatOneBro , @yudai-nkt and others.

I think this is a good feature. What about you? Although you don't have to review the implementation, if you have any opinions about this feature, its APIs, or concepts, please let me know.

yusukebe:
Hi @aiji42,

If you have any opinions as well, please let us know.

aiji42:
@yusukebe 
Does this mean that while Hono receives requests with the base path, it behaves as if the request was made without the base path portion when viewed from the framework side? In the example you provided with Remix, if Hono receives a request at /remix/foo/bar, does it appear as if access was made at /foo/bar from the Remix side?

yusukebe:
@aiji42 

Yes, you are correct. If the end-user accesses `/remix/foo/bar`, the Remix app will receive the Request as `/foo/bar`.

aiji42:
@yusukebe 
That's very cool🎉 I think it will be useful in the small library I'm currently developing.

However, the actual scenes where it can be used might be limited. In real use cases, you need to be aware of the base path when generating links. Many frameworks support base paths, and I get the feeling that issues may arise due to the path the framework receives not matching the actual path on the browser.

But for me, it's an extremely beneficial feature!

yudai-nkt:
Similar to a feedback above, I also would like to see some motivating use cases. One example I came up with is framework migration (like from Remix to SvelteKit or whatever combinations), but I'm not sure if it is really the case.

Of course, just because I don't come up with real-world use cases doesn't mean this PR isn't appealing. The idea of being runtime/framework agnostic seems interesting anyway!

yusukebe:
Thank you for your comments!

Additionally, there's one advantage to this structure using `mount()`. We can apply **middleware** to any mounted application, regardless of whether it was built with *itty-router*, *Remix*, or *Qwik*:

```ts
app.use('/another-app/admin/*', basicAuth({ username, password }))
```

This is incredibly useful, as we won't have to implement features like BasicAuth for each specific framework like *itty-router* or *Remix*.

metrue:
hey @yusukebe  sorry for the late response, I like your idea, it's amazing, basically it means Hono empowers the other frameworks possibility to run on multiple runtimes (with Hono runtime adapters) while enjoys Hono middlewares out of the box.

yujong-lee:
@yusukebe 

> we can mount any applications built with any framework like itty-router, Remix, Qwik, SoildJS into our Hono app.

I love this idea! Great work!

yusukebe:
@aiji42 @yudai-nkt @metrue @usualoma 

Thank you for your comments! Let's go with this idea.

## About API

As @usualoma mentioned:

> I think there is room for consideration regarding the API.

Indeed.

```ts
app.mount('/remix', handleRemixRequest, (c) => {
  return { env: env(c) }
})
```

I believe this `app.mount()` looks neat! We have to examine the implementation to see how much the code will increase. Nonetheless, it might be worth including in `hono-base.ts`.

yusukebe:
Hi @usualoma!

I've implemented this feature as `app.mount()`:

```ts
const ittyRouter = IttyRouter()

ittyRouter.get('/hello', () => new Response('Hello from itty-router!'))

const app = new Hono()

app.mount('/itty-router', ittyRouter.handle)
```

Code size for `v3.1.8`:

Screenshot 2023-05-17 at 17 49 43

Code size for `mounter/mount()` (`hono-base.ts` does not include `app.mount()`):

Screenshot 2023-05-17 at 18 22 49

Code size for `app.mount()`:

Screenshot 2023-05-17 at 17 48 58

By adding `app.mount()`, the application size will increase, of course. However, the code is smaller than in `v3.1.8`, so I think this increase is acceptable. Also, this API, `app.mount()`, is neat and cool. I would like to proceed with this API!

Could you review my implementation?

usualoma:
@yusukebe Thanks. `app.mount()` is pretty great!

I suggest that the URL passed to the sub-app be generated as follows.

* If we don't consider regular expressions, etc. (I don't think this is usually necessary) : https://github.com/honojs/hono/pull/1112
* If we consider regular expressions : https://github.com/honojs/hono/pull/1111

yusukebe:
Hi @usualoma 

Thanks! I believe we don't need to support regular expressions for `app.mount()`. So, let's go ahead with #1112!

PR #1105, opened 2023-05-15
feat(app): `app.routerName()`

yusukebe:
This PR introduces a `c.routerName()` function in `context.ts`. With this function, we can determine the name of the currently active router. This can be useful for debugging purposes.

```ts
import { Hono } from 'hono'

//...

app.get('/', (c) => {
  return c.text(c.routerName ?? 'N/A') // "SmartRouter + RegExpRouter"
})
```

### DIFF
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -65,6 +65,7 @@ type ContextOptions<E extends Env> = {
   notFoundHandler?: NotFoundHandler<E>
   path?: string
   paramData?: Record<string, string>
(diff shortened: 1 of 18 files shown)


usualoma:
It would be a good feature to get the name of the router!

However, I thought the following
* It is the app, not the context, that should know the router name
* Passing the router's name to the context for each request is a wasteful process in most cases.

I prefer to get it with `app.routerName` as follows.

https://github.com/honojs/hono/pull/1110

PR #1135, opened 2023-05-27
fix: Add missing client types and TypedResponse type

dimik:
Hi @yusukebe we have an issue with `hc` – we'd like to publish typed hono client from our Hono app as a library with zero dependencies, but since hono package.json has `"types": "dist/types/index.d.ts",` and this file doesn't contain client types we can't make it standalone – without `import from 'hono/dist...'` (see screenshots now/after change).

We have such entry point (client.ts) that we build separately from Hono app worker itself:
```
import { hc } from 'hono/client';

import { AppType } from '.';

export type { InferRequestType, InferResponseType } from 'hono/client';

export const client = hc<AppType>;
```

This is what we have now after build:
Screen Shot 2566-05-27 at 20 32 41

This is what we want to have with this changes:
Screen Shot 2566-05-27 at 20 29 39

So i exported few hono client types that we 
need for that. Changed RequestOptions => ClientRequestOptions to avoid possible type name collisions. As it uses only internally now that should not be a breaking change for anybody.
As well I exported the TypedResponse bcs we don't want use jsonT but instead to return Response | TypedResponse from handler and we need this type to be exposed as well.

I'm not sure if I need to change files in deno_dist as well. so please let me know if i need to roll them back

### DIFF
--- a/deno_dist/client/client.ts
+++ b/deno_dist/client/client.ts
@@ -1,7 +1,7 @@
 import type { Hono } from '../hono.ts'
 import type { ValidationTargets } from '../types.ts'
 import type { UnionToIntersection } from '../utils/types.ts'
-import type { Callback, Client, RequestOptions } from './types.ts'
+import type { Callback, Client, ClientRequestOptions } from './types.ts'
 import { replaceUrlParam, mergePath, removeIndexString, deepMerge } from './utils.ts'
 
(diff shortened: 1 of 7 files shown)


yusukebe:
Hi @dimik 

You are right, it should be exported. I'll merge it. Thanks!

PR #1139, opened 2023-05-29
feat: define JSX.ElementChildrenAttribute interface

### LINKED ISSUE #1138: Component children need to be passed as ordinary props because of the absence of `JSX.ElementChildrenAttribute`
The following Hono code produces an error saying `This JSX tag's 'children' prop expects type 'HtmlEscapedString' which requires multiple children, but only a single child was provided. ts(2745)`:

```ts
import { type HtmlEscapedString } from "hono/utils/html";

const Container = (props: { children: HtmlEscapedString }) => (
  <div class="container">{props.children}</div>
);
export const Page = () => (
  <Container>
    <p>Hello world!</p>
  </Container>
);
```

You need to apply the following patch to get the correct type, but this requires redundant props drilling. Alternatively, you can also write `<Container children={<p>Hello world!</p>} />`, but this is not so straightforward.

```diff
- export const Page = () => (
-   <Container>
+ export const Page = (props: { children: HtmlEscapedString }) => (
+   <Container {...props}>
```

On the other hand, the following React counterpart compiles.

```ts
import React from "react";

const Container = (props: { children: React.ReactNode }) => (
  <div className="container">{props.children}</div>
);
export const Page = () => (
  <Container>
    <p>Hello world!</p>
  </Container>
);
```

TypeScript's [`JSX.ElementChildrenAttribute`](https://www.typescriptlang.org/docs/handbook/jsx.html#children-type-checking) makes this possible and I'm preparing a PR for this.

Of course Hono is not React, so the syntax can be different from each other. That said, I suppose the first code snippet would be what users want (at least I do).

yudai-nkt:
Fix #1138.

### DIFF
--- a/src/middleware/jsx/index.ts
+++ b/src/middleware/jsx/index.ts
@@ -8,6 +8,9 @@ declare global {
   // eslint-disable-next-line @typescript-eslint/no-namespace
   namespace JSX {
     type Element = HtmlEscapedString
+    interface ElementChildrenAttribute {
+      children: Child
+    }
     interface IntrinsicElements {
       [tagName: string]: Props
     }


yusukebe:
Hi @yudai-nkt!

Great job! I'd like to merge it.

Just one thing. Could you run the `denoify` command to generate files for Deno and push them?

yudai-nkt:
Oops, I've seen you ask contributors for denoify many times but I forgot 😅

yusukebe:
Yes, `denoify` can be troublesome, but for now, it's safer to do it manually.

Okay, I'll proceed with the merge. Thanks!

Issue #1130, opened 2023-05-24
app.get should handle HEAD requests


alexgleason:
Ideally, HEAD requests would be matched by `app.get()` routes and it would only return the headers (no body).

It's common for other libraries ([like Express](https://expressjs.com/en/4x/api.html#app.METHOD)) to do this as a convenience, so you don't have to specify an `app.head()` for every single route.

yusukebe:
Hi @alexgleason!

Thanks for raising the issue. I'm considering whether we should implement this and, if so, how we should go about it.

If we decide to implement it so that when a HEAD request comes in, `app.get()` always handles it and responds without a body, disabling `app.head()`, it would be easy to implement. Like below:

* `HEAD /foo` will always be replaced with `GET /foo`, `app.get()` will handle it, and return the response without the body.
* Users can't define `app.head()`.

But, it's more challenging (for me) to implement with `app.head()`:

* `HEAD /foo` will be handled by `app.head('/foo')` if it's registered.
* If it was not matched, it falls back to `GET /foo`.

Perhaps the "disable `app.head()`" pattern would be acceptable if we decide that users should not define HEAD responses themselves. However, I'm not sure which approach is better at the moment.

@usualoma 

What do you think? We need to consider whether we should implement this and how we should do it.

yusukebe:
I'm considering this issue.

[According to MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD), the HEAD method is defined as follows:

> The HTTP HEAD method requests the [headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers) that would be returned if the HEAD request's URL was instead requested with the HTTP [GET](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET) method.

So, I think it's acceptable to adopt the following pattern:

* `HEAD /foo` will always be replaced with `GET /foo`, `app.get()` will handle it, and return the response without the body.
* Users can't define `app.head()`.

In addition, according to the specification, the HEAD method should not be defined by users, so this wouldn't be a new feature, but rather a "fix". Therefore, the release including this fix would be a "patch" or "minor" release.

alexgleason:
I agree @yusukebe !

yusukebe:
Created the two PR:

1. Built-in pattern #1142 
2. Middleware pattern #1143 

I think "2. Middleware pattern" is better.

renhiyama:
[...]

PR #1142, opened 2023-05-30
feat: handle HEAD method

yusukebe:
This PR allows `app.get()` to handle the HEAD method, resolving issue #1130.

Currently, we have to define the process in `app.head()` if a HEAD method request comes in. However, the HEAD method is essentially "the response of GET without the body". Therefore, `HEAD /foo` will be handled by `app.get('/foo')` and return the response without the body. Users will not be able to define `app.head()`.

While this is a feature change, on the other hand, it simply corrects the handling of HEAD. Therefore, I believe it's acceptable to release it as a patch.

### DIFF
--- a/deno_dist/hono-base.ts
+++ b/deno_dist/hono-base.ts
@@ -267,19 +267,16 @@ class Hono<E extends Env = Env, S = {}, BasePath extends string = '/'> extends d
   private dispatch(
     request: Request,
     eventOrExecutionCtx?: ExecutionContext | FetchEvent,
-    env?: E['Bindings']
+    env?: E['Bindings'],
+    method?: string,
   ): Response | Promise<Response> {
     const path = this.getPath(request)
-    const method = request.method
+    method ||= request.method
 
     // Handle HEAD method
     if (method === 'HEAD') {
-      request = new Request(request.url, {
-        ...request,
-        method: 'GET',
-      })
       return (async () => {
-        const response = await this.dispatch(request, eventOrExecutionCtx, env)
+        const response = await this.dispatch(request, eventOrExecutionCtx, env, 'GET')
         return new Response(null, response)
       })()
     }
--- a/deno_dist/hono-base.ts
+++ b/deno_dist/hono-base.ts
@@ -274,7 +274,7 @@ class Hono<E extends Env = Env, S = {}, BasePath extends string = '/'> extends d
 
     // Handle HEAD method
     if (method === 'HEAD') {
-      request = new Request(request, {
+      request = new Request(request.url, {
         ...request,
         method: 'GET',
       })
--- a/deno_dist/hono-base.ts
+++ b/deno_dist/hono-base.ts
@@ -29,6 +29,12 @@ interface RouterRoute {
   handler: H
 }
 
+interface DispatchOptions<Env> {
+  eventOrExecutionCtx?: ExecutionContext | FetchEvent
+  env?: Env
+  method?: string
+}
+
 function defineDynamicClass(): {
   new <E extends Env = Env, S = {}, BasePath extends string = '/'>(): {
     [M in Methods]: HandlerInterface<E, M, S, BasePath>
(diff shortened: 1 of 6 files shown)


jackdbd:
Often a HEAD request is made to decide whether making a GET request or not.

As MDN states [in this article](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD):

> [...] if a URL might produce a large download, a HEAD request could read its [Content-Length](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Length) header to check the filesize without actually downloading the file.

If I understood the change introduced by this PR correctly, forcing Hono to make a GET request would kind of  destroy the purpose of making a HEAD request in the first place.

yusukebe:
I believe there are advantages to implementing this as middleware. We wouldn't have to write more code in the core, such as in `hono-base.ts`. I prefer not to complicate the core.

> I would not deprecate app.head. it can be used to override the HEAD request for a particular route by placing it above the app.get.

I somewhat agree, but it's difficult to implement. If we were to do so, the code would become very complex. By using the middleware pattern, we don't have to deprecate `app.head()`, and we can control the areas where the HEAD request handling applies.

I'd like to know others's opinions. cc: @usualoma @metrue @ThatOneBro

ThatOneBro:
I think this makes sense as a concept. Using the middleware would be best probably, because sometimes users may be able to optimize requests in interesting ways by avoiding things like computing the body if the body length is known ahead of time (maybe this is handled in another way too, but this is what comes to mind.)

usualoma:
While I think the middleware idea in #1143 is excellent, I think this PR change is better because I believe in most use cases, the result of a HEAD request will be "the result of a GET request minus the body."

However, I understand that there is a desire to optimize the HEAD request without sending the GET request to the backend. How about adding #1152 to cover such cases? It could be written as follows.

```ts
app.get('/download', (c) => {
  if (c.req.method === 'HEAD') {
    return new Response(null, {
      status: 200,
      headers: {
        'Content-Type': 'application/octet-stream',
        'Content-Length': cachedLength,
      },
    })
  }

  return fetch(backend);
})
```

yusukebe:
@usualoma 

Thank you for your comment and for creating the PR. It's difficult to choose which PR to merge.

Your implementation #1152 is good, but I'm concerned about the increasing number of arguments in `dispatch()`, from three to four with PR #1152. I'd like to keep `hono-base.ts` simple. Therefore, the middleware pattern #1143 it's better, I think.

renhiyama:
> I believe there are advantages to implementing this as middleware. We wouldn't have to write more code in the core, such as in `hono-base.ts`. I prefer not to complicate the core.
> 
> > I would not deprecate app.head. it can be used to override the HEAD request for a particular route by placing it above the app.get.
> 
> I somewhat agree, but it's difficult to implement. If we were to do so, the code would become very complex. By using the middleware pattern, we don't have to deprecate `app.head()`, and we can control the areas where the HEAD request handling applies.
> 
> I'd like to know others's opinions. cc: @usualoma @metrue @ThatOneBro

I know that it might complicate the core code, but most web servers have this feature enabled by default. Most web devs don't know about this feature, and there are larger changes of web devs not enabling them via middleware.

yusukebe:
Okay. Anyway, #1152 should be merge into this PR. I'll do that.

usualoma:
I am not against using middleware, but I still think this PR is better. I think it is standard web server behavior, and hono-base.ts is not that complicated.

Including the #1156 change is also better for the following reasons.
* `dispatch` is a private method, so the caller is in a controllable location, and I think the argument increase here is acceptable
* That there is an overhead if the caller needs to create an object of options at the time of the call.
* That there is a small overhead in `options?.eventOrExecutionCtx`.
* Optional objects are hard to mangle.

yusukebe:
> `dispatch` is a private method, so the caller is in a controllable location, and I think the argument increase here is acceptable

That's right! So, it's private.I've realized your implementation is excellent and this PR is better than the middleware pattern.

Thanks!

yusukebe:
@alexgleason Thanks for raising the issue.
@jackdbd @ThatOneBro @renhiyama Thanks for your comments.
@usualoma Thanks for the great implementation.

Let's go with it. I'll merge it!

alexgleason:
Thank you @yusukebe!

riderx:
guys I dont get it I still see nothing in the doc have handling HEAD, hono is amazing but this bother me for a year now. how to do HEAD correctly without full get? and have a special route for it

PR #1143, opened 2023-05-30
feat: `headMethod` middleware

### LINKED ISSUE #1130: app.get should handle HEAD requests

yusukebe:
This PR introduces the "HEAD Method middleware". This middleware handles `HEAD` method requests.

```ts
import { headMethod } from 'hono/head-method'

// ...

app.use('*', headMethod({ app }))

app.get('/', (c) => {
  return c.text('Hello!')
})
```

Incoming `HEAD` requests will be handled by `app.get()` and it will return a response without the body.

```
curl --head http://localhost:8787/

HTTP/1.1 200 OK
Content-Type: text/plain;charset=UTF-8
```

The advantage of this approach is that it's implemented as middleware. Therefore, users can choose whether to allow HEAD requests or not.

This will resolve #1130 , alternative to #1142

### DIFF
--- a/src/middleware/head-method/index.test.ts
+++ b/src/middleware/head-method/index.test.ts
@@ -1,0 +1,35 @@
+import { Hono } from '../../hono'
+import { headMethod } from './index'
+
+describe('Head Method Middleware', () => {
+  const app = new Hono()
+
+  app.use('/page/*', headMethod({ app }))
+  app.use('*', async (c, next) => {
+    await next()
+    c.header('foo2', 'bar2')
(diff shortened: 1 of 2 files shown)


yusukebe:
This is not necessary.

PR #1145, opened 2023-05-31
fix(trie-router): fixed the named capture issue

### LINKED ISSUE #1141: Should the router handle `params` per handler?
This issue pertains to the router. Currently, the result object is composed of `handlers` and `params`:

```ts
interface Result<T> {
  handlers: T[]
}
```

However, it might be that each `handler` should have its own `params`, rather than all handlers sharing one `params`.

For instance, in the current pattern, the result is not as expected. (This example uses TrieRouter, because RegExpRouter has another issue #1140)

```ts
  await next()
})

})
```

```
GET /webhooks/github => service: undefined
```

I believe it should return `service: github`. To make this work correctly, we might need a structure like the one below, where `params` are handled per handler.

```ts
interface Result<T> {
  routes: Route<T>[]
}

interface Route<T> {
  handler: T
}
```

If we decide to address these issues, it will require significant effort. But, it might be worth doing.

yusukebe:
Will fix #1141 just for TrieRouter.

### DIFF
--- a/deno_dist/router/trie-router/node.ts
+++ b/deno_dist/router/trie-router/node.ts
@@ -33,7 +33,6 @@ export class Node<T> {
   order: number = 0
   name: string
   handlerSetCache: Record<string, HandlerSet<T>[]>
-  shouldCapture: boolean = false
 
   constructor(method?: string, handler?: T, children?: Record<string, Node<T>>) {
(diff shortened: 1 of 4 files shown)


yusukebe:
Move to #1148

PR #1146, opened 2023-06-01
fix(reg-exp-router): Register static paths first for duplicate checking.

### LINKED ISSUE #1140: RegExpRouter returns not found
In this case, RegExpRouter returns a "not found" response, which is not expected.

```ts
  await next()
})

})
```

```
GET /webhooks/github => 404 Not Found
```

If using TrieRouter, it returns a 200 response, though returning `undefined` is another issue.

```
GET /webhook/github => service: undefined
```

usualoma:
Fixes #1140


yusukebe:
@usualoma 

Looks good! Thanks!

PR #1148, opened 2023-06-01
fix(tire-router/pattern-router): fixed the named capture issue

### LINKED ISSUE #1141: Should the router handle `params` per handler?

yusukebe:
This PR will fix the issue of capturing named parameters in TrieRouter and PatternRouter.

If we have the following routes:

```ts
  await next()
})

})
```

Then, when accessing `GET /webhooks/github`, it currently can't handle the named parameter as `github`:

```
GET /webhooks/github => service: undefined
```

With this fix, it can capture `github`:

```
GET /webhooks/github => service: github
```

Resolves #1141

### DIFF
--- a/deno_dist/router/pattern-router/router.ts
+++ b/deno_dist/router/pattern-router/router.ts
@@ -1,7 +1,7 @@
 import type { Result, Router } from '../../router.ts'
 import { METHOD_NAME_ALL } from '../../router.ts'
 
-type Route<T> = [RegExp, string, T] // [pattern, method, handler]
+type Route<T> = [RegExp, string, T, string] // [pattern, method, handler, path]
 
 export class PatternRouter<T> implements Router<T> {
   name: string = 'PatternRouter'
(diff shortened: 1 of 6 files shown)


yusukebe:
Hi @usualoma !

Could you review this? I think the spec is good, but I'd like to you review the implementation.

usualoma:
Hi @yusukebe!

As for TrieRouter, I think that change is fine.

As for PatternRouter, I think it can be written a little shorter, as in #1154.

yusukebe:
@usualoma 

> As for PatternRouter, I think it can be written a little shorter, as in https://github.com/honojs/hono/pull/1154.

Ahhh, only that! Super thank you!

PR #1157, opened 2023-06-02
refactor: Tweaks variable names to reduce code.

usualoma:
It would reduce the number of conditional branches in `result?.handlers` and `result?.params` and shorten the code by a few lines.

Since this is not a particularly big improvement, it is safe to decide not to merge.

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -64,7 +64,7 @@ type ContextOptions<E extends Env> = {
   executionCtx?: FetchEvent | ExecutionContext | undefined
   notFoundHandler?: NotFoundHandler<E>
   path?: string
-  paramData?: Record<string, string>
+  params?: Record<string, string>
 }
 
 export class Context<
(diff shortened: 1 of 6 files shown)


yusukebe:
Hi @usualoma!

It's neat and it's great that `result?` has been removed! I'd like to merge this, but there's a conflict. Could you resolve it?

src/context.ts
```diff
@@ -89,7 +89,7 @@ export class Context<
   private _pH: Record<string, string> | undefined = undefined // _preparedHeaders
   private _res: Response | undefined
   private _path: string = '/'
-  private _pData?: Record<string, string> | null // __paramData
+  private _params?: Record<string, string> | null // __params
```
yusukebe, line 92: The comment `// __params` is not necessary.

usualoma: Fixed!
0e39109

usualoma:
@yusukebe Thanks! Resolved.  6c0b358

yusukebe:
Thanks! Merge now!

PR #1162, opened 2023-06-05
feat: Allow `context.jsonT` to take interface as an argument

### LINKED ISSUE #1158: Allow using interface in `ctx.jsonT()`
I'm using Hono's RPC mode.
I tried to pass the type of interface to the argument of `context.jsonT()`, but I get a type error.

```ts
interface Data {
  data: string;
}
function getData(): Data {
  return { data: "foo" };
}


import { Hono } from "$hono/mod.ts";
const app = new Hono();
const route = app
  .get(
    "/api",
    (c) => {
      const data = getData();
      return c.jsonT(data); // Type 'Data' is not assignable to type 'JSONObject'.
    },
  );
export type AppType = typeof route;
```

```
  Overload 1 of 2, '(object: JSONValue, status?: number | undefined, headers?: HeaderRecord | undefined): TypedResponse<never>', gave the following error.
    Argument of type 'Data' is not assignable to parameter of type 'JSONValue'.
      Type 'Data' is not assignable to type 'JSONObject'.
        Index signature for type 'string' is missing in type 'Data'.  Overload 2 of 2, '(object: JSONValue, init?: ResponseInit | undefined): TypedResponse<never>', gave the following error.
    Argument of type 'Data' is not assignable to parameter of type 'JSONValue'.deno-ts(2769)
```

I know that changing `interface Data {}` to `type Data = {}` will eliminate the type error, but there are types provided by other libraries, so changing everything from `interface` to `type` is not realistic.
How about relaxing the restrictions on the JSONValue type here so that we can return a response of the type defined by the interface?
I think that types (such as `MyClass` / `Function`) that clearly cannot be serialized to JSON can be treated as type errors, and anything else can be considered valid as an argument.

ayame113:
close #1158

### DIFF
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -4,7 +4,7 @@ import type { Env, NotFoundHandler, Input } from './types.ts'
 import type { CookieOptions } from './utils/cookie.ts'
 import { serialize } from './utils/cookie.ts'
 import type { StatusCode } from './utils/http-status.ts'
-import type { JSONValue } from './utils/types.ts'
+import type { JSONValue, InterfaceToType } from './utils/types.ts'
 
 type Runtime = 'node' | 'deno' | 'bun' | 'workerd' | 'fastly' | 'edge-light' | 'lagon' | 'other'
 type HeaderRecord = Record<string, string | string[]>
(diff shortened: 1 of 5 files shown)


deno_dist/context.ts
```diff
@@ -48,9 +48,40 @@ interface JSONTRespond {
...
+  <T>(
+    object: InterfaceToType<T> extends JSONValue ? T : JSONValue,
+    init?: ResponseInit
+  ): TypedResponse<
+    InterfaceToType<T> extends JSONValue
+      ? JSONValue extends InterfaceToType<T>
+        ? never
+        : T
+      : never
   >
```
ayame113, line 85: I had to add overloads for `JSONTRespond` to avoid breaking existing test cases.
(I had to define both `object: T extends JSONValue ? T : JSONValue` and `object: InterfaceToType<T> extends JSONValue ? T : JSONValue` versions.)

yusukebe:
Hi @ayame113 !

Excellent! The implementation looks good, including the changes for the overloads. It's also good that you ran `denoify`. I'll merge it now. Thanks!

ayame113:
Thank you for your review!

PR #1165, opened 2023-06-06
fix: replace to use better base64-encoder

### LINKED ISSUE #1160: [aws-lambda] probrem on implicit conversion binary to base64
Thank you for the nice framework!

I'm trying hono in AWS Lambda and found a small issue in conversion to base64.

https://github.com/honojs/hono/blob/main/src/adapter/aws-lambda/handler.ts

It seems that return value is automatically converted into base64 when content-type means binary type.
The conversion seems to support only `ReadableStream` but in my usecase need to response from `ArrayBuffer` like this (using node-fetch).

```javascript
const data = await fetch(
    'https://pass/to/image.png'
).then((res) => res.arrayBuffer());
c.status(200);
c.header('Content-Type', 'image/png');
return c.body(data);
```

However, this snipet doesn't work, error occurs in the conversion. We need other conversion like:

```javascript
function base64Encode(buf: ArrayBuffer) {
    let string = '';
    new Uint8Array(buf).forEach((byte) => {
        string += String.fromCharCode(byte);
    });
    return btoa(string);
}
```

In my opinion we would have some options.

1. modify current function to support ArrayBuffer: https://github.com/honojs/hono/blob/main/src/adapter/aws-lambda/handler.ts#L143
    a. but not only ArrayBuffer needs support: e.g.) Buffer
2. pass base64 string to res.body and skip conversion to base64 in handle().
    a. may able to skip when content-type is binary and res.body is String
    b. or quit automatic conversion. 

How do you think?

Kanahiro:
discussed in and close #1160 

At first I thought I should add new test but this PR makes no change for a behavior from outside so I decided not to do so.

### DIFF
--- a/src/adapter/aws-lambda/handler.ts
+++ b/src/adapter/aws-lambda/handler.ts
@@ -2,6 +2,8 @@
 import crypto from 'crypto'
 import type { Hono } from '../../hono'
 
+import { encodeBase64 } from '../../utils/encode'
+
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore
 globalThis.crypto = crypto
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @Kanahiro !

Perfect! I'll merge it, thanks!

PR #1168, opened 2023-06-06
fix: avoid range-error

Kanahiro:
In Node.js, spread operator for large-sized array occurs Range Error:

https://stackoverflow.com/questions/51249561/aws-lambda-rangeerror-maximum-call-stack-size-exceeded
https://github.com/nodejs/node/issues/16870

When I use somewhat large image this problem happens. To avoid this I re-wrote some codes. This doesn't make no change on behaiviors.


yusukebe:
Hi @Kanahiro !

Thank you for the PR! I'd like to merge it. So, one thing. Please do `denoify` and add&commit&push the genereted file.

Kanahiro:
@yusukebe 
Thank you for guidance, it's done!

yusukebe:
Okay, I'll merge now!

PR #1177, opened 2023-06-13
Allow setting cookie's maxAge to 0.

Manouchehri:
0 tells the browser to delete the cookie right away, and is a valid value. =)


yusukebe:
Hi @Manouchehri, thank you for creating the PR!

I'd like to merge it, but I need you to do two things first:

1. Adjust the code so that all tests pass.
2. Run `yarn denoify`, then git add, commit, and push the generated files. This is necessary for Deno.

yusukebe:
In addition, I think we need to implement `deleteCookie()` in the cookie middleware, even if this PR is merged. Refer to: #1175. I might create the PR or if you're able, you could create it!

PR #1189, opened 2023-06-21
fix: application/x-www-form-urlencoded decoding

klittlepage:
The current logic for decoding application/x-www-form-urlencoded is overly restrictive and misses `Content-Type` headers with a character encoding set, i.e., `application/x-www-form-urlencoded; charset=UTF-8`.

This fix harmonizes the logic for handling `multipart/form-data` and `application/x-www-form-urlencoded` in the `parseBody` method of request objects.


yusukebe:
Hi @klittlepage!

Great fix! And it's great that you've already do`denoify`. I'll merge it right away. Thanks!

PR #1199, opened 2023-06-23
fix(etag): support multi-value If-None-Match

jamesarosen:
* chore(etag): reduce global state in tests
* perf(etag): don't override ETags from upstream
* perf(etag): 304s include only necessary headers
* fix(etag): Support multi-value If-None-Match
* perf(etag): prefer Header.prototype.get

I recommend reviewing commit-by-commit as each commit has detailed rationale, including references to the relevant specs.

### DIFF
--- a/deno_dist/middleware/etag/index.ts
+++ b/deno_dist/middleware/etag/index.ts
@@ -16,7 +16,7 @@ const RETAINED_304_HEADERS = [
   'cache-control', 'content-location', 'date', 'etag', 'expires', 'vary'
 ]
 
-function etagMatches(etag: string, ifNoneMatch: string | undefined) {
+function etagMatches(etag: string, ifNoneMatch: string | null) {
   return ifNoneMatch != null && ifNoneMatch.split(/,\s*/).indexOf(etag) > -1
 }
 
@@ -25,7 +25,7 @@ export const etag = (options?: ETagOptions): MiddlewareHandler => {
   const weak = options?.weak ?? false
 
   return async (c, next) => {
-    const ifNoneMatch = c.req.header('If-None-Match') || c.req.header('if-none-match')
+    const ifNoneMatch = c.req.headers.get('If-None-Match')
 
     await next()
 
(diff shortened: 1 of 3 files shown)


src/middleware/etag/index.ts
```diff
@@ -5,30 +5,52 @@ type ETagOptions = {
...
 
+/**
+ * List of headers to pass through on 304 responses. From the spec:
+ * > The response must not contain a body and must include the headers that
+ * > would have been sent in an equivalent 200 OK response: Cache-Control,
+ * > Content-Location, Date, ETag, Expires, and Vary.
+ */
+const RETAINED_304_HEADERS = [
+  'cache-control', 'content-location', 'date', 'etag', 'expires', 'vary'
+]
```
jamesarosen, line 16: I can imagine a scenario where an app wants to include additional headers. Should this be configurable via an option alongside `weak`?

yusukebe: Yes, I also think it would be better for these to be options.

yusukebe:
Hi @jamesarosen !

Great work! Looks good.

And, we can ignore the "Bun" CI is faling because it's Bun's matter, after merging, I'll fix the version of Bun.

jamesarosen:
> please run yarn denoify

Done 👍

yusukebe:
Great! I'll merge it now!

deno_dist/middleware/etag/index.ts
```diff
@@ -2,33 +2,59 @@ import type { MiddlewareHandler } from '../../types.ts'
...
       })
-      c.res.headers.delete('Content-Length')
+      c.res.headers.forEach((_, key) => {
+        if (retainedHeaders.indexOf(key.toLowerCase()) === -1) {
+          c.res.headers.delete(key)
+        }
+      })
     } else {
-      c.res = new Response(clone.body, clone)
+      c.res = new Response(undisturbedRes.body, undisturbedRes)
```
jamesarosen, line 57: After some real-world testing, I think this should be

```ts
c.res = undisturbedRes
```

There are some environments where this is causing the non-304 responses to have empty body.

yusukebe: Ah, it appears that copying the Response might be unnecessary. Could you create a PR for this?

jamesarosen: Voilà: https://github.com/honojs/hono/pull/1242

PR #1204, opened 2023-06-26
perf(Request) optimize Request.prototype.header

jamesarosen:
`Request.prototype.header` has two overloads:

 1. with a string name, it returns that header value if present
 2. with no argument, it returns all the headers as an object

Previously, the implementation built the whole object for (2) even when returning a single header. This was wasteful. This commit changes it to return the header by name if specified and only iterate over all the headers when the caller asks for the whole object.

### DIFF
--- a/deno_dist/request.ts
+++ b/deno_dist/request.ts
@@ -69,14 +69,13 @@ export class HonoRequest<P extends string = '/', I extends Input['out'] = {}> {
   header(name: string): string | undefined
   header(): Record<string, string>
   header(name?: string) {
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @jamesarosen !

Good improvement, it looks good. I'd like to merge it, so please run `yarn denoify` as well as in #1199.

jamesarosen:
> please run yarn denoify

Done 👍

yusukebe:
Great! Thanks!

PR #1209, opened 2023-06-29
fix(tests): Lagon env test

QuiiBz:
This PR fixes the env test of Lagon by using the `-e` (or `--env`) flag of `lagon dev` to load the `.env.lagon` file.


yusukebe:
Thanks @QuiiBz !

PR #1214, opened 2023-07-01
Fix replaceUrlParam to ignore regexp in path

### LINKED ISSUE #1213: Inaccurate RPC client requests due to regex patterns in paths
There is a current issue where using regex patterns in paths leads to wrong request path processing by the RPC client. To resolve this, we need to ensure the client ignores regex patterns in paths for proper request handling.

### Steps to Reproduce
Assume we have a router path like `/post/:id{[0-9]+}`. Under normal circumstances, we should be able to make a request with the RPC client using something like `await client.post[':id{[0-9]}'].$get(...)`.

However, the RPC client currently doesn't ignore the regex pattern, leading it to send a request to `/post/123%7B[0-9]+%7D` when the provided path parameter (in this case, id) is 123. This is not the intended behavior and needs to be addressed.

hyp3rflow:
This PR fixes #1213. 

Fix Regex pattern in replaceUrlParam to replace regex path correctly.

### DIFF
--- a/deno_dist/client/utils.ts
+++ b/deno_dist/client/utils.ts
@@ -9,7 +9,7 @@ export const mergePath = (base: string, path: string) => {
 
 export const replaceUrlParam = (urlString: string, params: Record<string, string>) => {
   for (const [k, v] of Object.entries(params)) {
-    const reg = new RegExp('/:' + k)
+    const reg = new RegExp('/:' + k + '({[^}]*})?')
     urlString = urlString.replace(reg, `/${v}`)
   }
   return urlString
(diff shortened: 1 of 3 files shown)


yusukebe:
Hi @hyp3rflow !

Thanks for raising the issue and quick fix.

The implementation and the tests look great! However, there is one more step I'd like you to take. Please run `yarn denoify`, then add, commit, and push the generated file. This step is necessary for Deno.

hyp3rflow:
@yusukebe thanks for the quick review :) I've pushed generated denoify files.

yusukebe:
Great! I'll merge it now! Thanks!

PR #1215, opened 2023-07-01
Add Server-Timing API as middleware

PassiDel:
This adds [Server-Timing](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing) as a middleware.

Please see honojs/middleware#99 for context.

I've replaced `process.hrtime()` with this, since `jest` failed to access performance:
```ts
const getTime = () => {
  try {
    return performance.now()
  } catch {}
  return Date.now()
}
```

This technique may not fully work for cloudflare worker, because they [disallow the use of timers](https://developers.cloudflare.com/workers/learning/security-model/#step-1-disallow-timers-and-multi-threading).


deno_dist/middleware/timing/index.ts
```diff
@@ -0,0 +1,135 @@
...
+    }
+  }
+}
+
+interface Timer {
+  description?: string
+  start: number
+}
+
+interface Config {
```
yusukebe, line 19: It's better to be `TimingOptions` like a`CORSOptions`.

deno_dist/middleware/timing/index.ts
```diff
@@ -0,0 +1,135 @@
...
+}
+
+const getTime = () => {
+  try {
+    return performance.now()
+  } catch {}
+  return Date.now()
+}
+
+export const timing = (config?: Partial<Config>): MiddlewareHandler => {
```
yusukebe, line 34: It also should be `options?: Partial<TimingOptions>`

yusukebe (commented):
Thank you, @PassiDel!

This is great, even though it can't measure in nanoseconds on Cloudflare Workers, it's still useful.

Screenshot 2023-07-02 at 22 20 23

I've left some comments, please check them out.

PassiDel:
Nice, I wasn't sure how CF Workers would handle that, thanks for checking!

yusukebe:
Thanks! I'll merge it and release `v3.3.0` includes this feature soon. Great work!

PR #1234, opened 2023-07-14
perf(deno/middlewares/serveStatic): serve static with stream

markthree:
In deno, we can use streams to handle static services, which makes the response faster and the program memory usage less under large files。


yusukebe:
Hi @markthree !

Thanks for creating this PR. This is a good change and I'd like to merge it, but there are two things I'd like you to do:

### Fix the test

To pass the test, you need to add the following code below these lines:

https://github.com/honojs/hono/blob/5090c3c817fcae0c4de57843c143a0a4b2337e98/runtime_tests/deno/middleware.test.tsx#L94-L96

Here is the code to add:

```ts
await res.body?.cancel()
```

This will close the file opened in the test.

### Write code in the correct place

Files under `deno_dist/*` are auto-generated. Actually, you should write your code in the following file:

```
src/adapter/deno/serve-static.ts
```

Then, run `yarn denoify` in the top directory to generate the files for Deno. Finally, git add, commit and push them.

markthree:
@yusukebe Hi

All steps have been completed 🙌

yusukebe:
@markthree 

Thank you! However, it appears that unnecessary test files have been included in the `deno_dist` directory. We don't need files such as:

```
deno_dist/adapter/aws-lambda/handler.test.ts
deno_dist/adapter/cloudflare-pages/handler.test.ts
....
deno_dist/utils/url.test.ts
deno_dist/validator/validator.test.ts
```

Could you please run `yarn denoify` again? These files should be removed.

markthree:
@yusukebe Hold on, I'll rebuild.

markthree:
@yusukebe Sorry, the `yarn denoify` command is executing denoify and rm in parallel on my windows computer, so I was not able to accurately delete the test files in one go, I'll retry it in the docker container

markthree:
@yusukebe hi
Everything is ok 😁

yusukebe:
Great!! I'll merge it now! Thanks!

PR #1235, opened 2023-07-14
fix(lambda): avoid UTF-8-encoding binary data

denisw:
So far, the AWS Lambda adapter has used `atob` to convert a Base64-encoded request body to a binary. However, `atob` returns a string, and passing that to the `Request` constructor (at least on Node.js, which Lambda runs on) results in the binary string being encoded as UTF-8. This means that if the request body contains bytes which are not in the ASCII range, they are replaced with UTF-8 multi-byte encodings, which leads to `c.req.blob()`, `c.req.formData()`, etc. returning the wrong binary data!

This commit fixes the issue by replacing `atob(body)` with `Buffer.from(body, 'base64')`, which does not have that problem. The `Buffer` can be directly passed as body for `new Request()` because it implements the `Uint8Array` interface. While this makes the AWS Lambda adapter depend directly on the "node:buffer" module, this should be fine because AWS Lambda functions always run on Node.

### DIFF
--- a/runtime_tests/lambda/index.test.ts
+++ b/runtime_tests/lambda/index.test.ts
@@ -20,6 +20,11 @@ describe('AWS Lambda Adapter for Hono', () => {
     return c.text(body.message)
   })
 
+  app.post('/post/binary', async (c) => {
+    const body = await c.req.blob()
+    return c.text(`${body.size} bytes`)
+  })
+
   const username = 'hono-user-a'
   const password = 'hono-password-a'
   app.use('/auth/*', basicAuth({ username, password }))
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @denisw !

Thanks for the PR.I think you are right.

yusukebe:
Hey @watany-dev,

I believe the Lambda@Edge adapter is also using `atob()`. Could you please check this and if it is the case, consider replacing it with `Buffer`?

watany-dev:
@yusukebe 

Yes, that's right, I made it as a reference and it has the same process. I'll remove it for reference.

PR #1237, opened 2023-07-15
fix: Safe modification of Buffer

watany-dev:
This modification replaces the use of `atob()` and `new Buffer()` with `Buffer.from()`. As mentioned in [this article](https://nodejs.org/dist/latest-v6.x/docs/api/buffer.html#buffer_class_buffer), I believe that `new Buffer()` should not be used.

and remove `btoa()`


watany-dev:
Restarting due to lack of support in Lagon.

PR #1238, opened 2023-07-15
fix: Safe modification of Buffer

watany-dev:
This modification replaces the use of atob() and new Buffer() with Buffer.from(). As mentioned in [this article](https://nodejs.org/dist/latest-v6.x/docs/api/buffer.html#buffer_class_buffer), I believe that new Buffer() should not be used.


yusukebe:
Hi @watany-dev,

As you've pointed out, we need to use `Buffer.from()`. Your changes look good. Thanks!

PR #1241, opened 2023-07-16
chore: use `$npm_execpath` when calling package manager

jbergstroem:
This makes `package.json` scripts work when invoked from either npm, yarn or pnpm.

Note: I'm not sure if this works from Windows (`$` vs `%%`).


jbergstroem:
Rationale here is not having to muscle memory which package manager is used in which repo.

yusukebe:
Hi @jbergstroem !

Thank you for creating this PR. I wasn't aware of `$npm_execpath` and it seems good. But, for this Hono project, we use `yarn.lock` to resolve dependencies, which means we should only use **Yarn**. What do you think about this?

jbergstroem:
> which means we should only use Yarn. What do you think about this?

All good of course; the idea is that you can invoke `npm run build` ("muscle memory") and it won't break because you don't have `yarn` installed. This is not a huge deal but I did just that (`pnpm run build`) and thought I might as well put a PR up as a discussion point.

yusukebe:
Hi @jbergstroem,

Thanks for your explanation!

I've pulled this PR branch and tried running `pnpm run build`, but it appears the `publint` - `postbuild` isn't being executed. Could you check on that?

While I think your approach is good, but we need to make sure that we do not break a function that has worked well in the past.

jbergstroem:
> While I think your approach is good, but we need to make sure that we do not break a function that has worked well in the past.

Totally! In the case of `pnpm`, you will have to enable the `{pre,post}-build` via a config flag. You can get more context on why here: https://github.com/pnpm/pnpm/issues/2891#issue-707579979

Let me know if you want me to address this differently by avoiding deprecated pre/post steps.

yusukebe:
Hi @jbergstroem,

I wasn't aware of that - thanks for bringing it to my attention. 

However, after giving this PR some thought, I've decided not to merge it. Although it's not a significant change and is marked as a "`chore`", there are a couple of reasons behind my decision. One reason is that `$npm_execpath` is not compatible with Windows. For Windows, [we would need to use `%npm_execpath%`](https://stackoverflow.com/questions/53883405/npm-execpath-isnt-recognized-as-an-internal-or-external-command-program-or).

Furthermore, although it might not be a concern, I'm worried about potential issues that might arise due to differences between package managers. In this project, we've chosen to use **Yarn**, so I think it's better to keep using `yarn` exclusively, even if it requires contributors to use yarn.

Anyway, I appreciate your insights and the discussion we've had around this. Although it's not directly related to this PR, we might consider changing the package manager from **Yarn v1** in the future.

Does this sound alright to you, @jbergstroem?

jbergstroem:
Yes, of course! As I mentioned early, it was more a discussion point/itch I had. Thanks for the consideration.

PR #1242, opened 2023-07-18
refactor(etag): simplify cloning logic

jamesarosen:
If the upstream response doesn't include an `ETag` header, the ETag middleware needs to clone the response so it can read the body to compute an ETag. This commit simplifies the management of the original and cloned responses.


yusukebe:
Hi @jamesarosen !

Looks good! One thing, could you do `yarn denoify` and git add&commit&push the generated file?

yusukebe:
Hi @jamesarosen !

Thanks!

PR #1246, opened 2023-07-19
fix(trie-router-router): For "Named param with trailing wildcard"

### LINKED ISSUE #1243: Routing `/:id/*`
This is an issue related to routers.

The results when handling routes such as `/:id/*` vary between each router. If you have the following app:

```ts
app.get('/:id/*', (c) => {
  return c.text(`${c.req.param('id')}`)
})
```

What are the results when accessing `/`, `/foo`, and `/foo/bar`? The results are as follows:

```
GET /
TrieRouter: 404
RegExpRouter: 404
LinearRouter: Error
PatternRouter: 404

GET /foo
TrieRouter: 404
RegExpRouter: 200 foo
LinearRouter: Error
PatternRouter: 200 undefined

GET /foo/bar
TrieRouter: 200 foo
RegExpRouter: 200 foo
LinearRouter: Error
PatternRouter: 200 undefined
```

These are not expected. Therefore, we need to decide on the behavior for `/:id/*` and fix the implementation.

I believe we should follow the specification used by *TrieRouter*, which would result in:

```
GET / => 404
GET /foo => 404
GET /foo/bar => 200 foo
```

If we choose this specification, the routes that need fixing are:

*RegExpRouter*
* GET /foo

*PatternRouter*
* GET /foo
* GET /foo/bar

*LinearRouter* throws `UnsupportedPathError` errors for all patterns. If we allow this, we will need to use it with the *SmartRouter* and *TrieRouter* for the preset `hono/quick`.

This is an edge case, but we must resolve it as issues like #1240 have been raised.

### The script

```ts
import { Hono } from 'hono'
import { type Router } from 'hono/dist/types/router'
import { LinearRouter } from 'hono/router/linear-router'
import { PatternRouter } from 'hono/router/pattern-router'
import { TrieRouter } from 'hono/router/trie-router'

const createHono = (router: Router<any>) => {
  const app = new Hono({
    router: router,
  })
  app.get('/:id/*', (c) => {
    return c.text(`${c.req.param('id')}`)
  })
  app.showRoutes()
  return app
}

const routers: Router<any>[] = [
  new TrieRouter(),
  new RegExpRouter(),
  new LinearRouter(),
  new PatternRouter(),
]

const apps: Record<string, Hono> = {}

for (const router of routers) {
  const app = createHono(router)
  apps[router.name] = app
}

const request = async (path: string) => {
  console.log(`GET ${path}`)
  for (const [name, app] of Object.entries(apps)) {
    try {
      const res = await app.request(path)
      if (res.ok) {
        console.log(`${name}: ${res.status} ${await res.text()}`)
      } else {
        console.log(`${name}: ${res.status}`)
      }
    } catch (e) {
      console.log(`${name}: Error`)
    }
  }
}

console.log('')
await request('/')
console.log('')
await request('/foo')
console.log('')
await request('/foo/bar')
```

yusukebe:
Fix #1243

### DIFF
--- a/src/router/trie-router/node.test.ts
+++ b/src/router/trie-router/node.test.ts
@@ -127,6 +127,31 @@ describe('Param prefix', () => {
   })
 })
 
+describe('Named params and a wildcard', () => {
+  const node = new Node()
+
+  node.insert('get', '/:id/*', 'onepart')
+
+  it('get /', () => {
+    const res = node.search('get', '/')
+    expect(res).toBeNull()
+  })
+
+  it('get /foo', () => {
+    const res = node.search('get', '/foo')
+    expect(res).not.toBeNull()
+    expect(res?.handlers).toEqual(['onepart'])
+    expect(res?.params['id']).toEqual('foo')
+  })
+
+  it('get /foo/bar', () => {
+    const res = node.search('get', '/foo/bar')
+    expect(res).not.toBeNull()
+    expect(res?.handlers).toEqual(['onepart'])
+    expect(res?.params['id']).toEqual('foo')
+  })
+})
+
 describe('Wildcard', () => {
   const node = new Node()
   node.insert('get', '/wildcard-abc/*/wildcard-efg', 'wildcard')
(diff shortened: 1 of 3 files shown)


yusukebe:
Hey @usualoma 

Can you review this?

usualoma:
@yusukebe Thanks. LGTM!

PR #1248, opened 2023-07-20
fix(adaptor): Unable to modify headers in lambda@edge

watany-dev:
The callback was not handling cloned responses. This has been fixed, and the operation is now possible.

https://zenn.dev/yusukebe/articles/647aa9ba8c1550#1.%E3%83%AC%E3%82%B9%E3%83%9D%E3%83%B3%E3%82%B9%E3%83%98%E3%83%83%E3%83%80%E3%81%AE%E8%BF%BD%E5%8A%A0

### DIFF
--- a/runtime_tests/lambda-edge/index.test.ts
+++ b/runtime_tests/lambda-edge/index.test.ts
@@ -43,21 +43,13 @@ describe('Lambda@Edge Adapter for Hono', () => {
   app.get('/auth/abc', (c) => c.text('Good Night Lambda!'))
 
   app.get('/header/add', async (c, next) => {
-    c.header("Strict-Transport-Security", "max-age=63072000; includeSubdomains; preload")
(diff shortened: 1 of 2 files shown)


watany-dev:
There are many similar header modifications, so we are standardizing them.

watany-dev:
It is now possible that these simple renovations may not be enough.
1. the following code you were referring to deals with event.Records[0].cf.response instead of event.Records[0].cf.request
https://docs.aws.amazon.com/ja_jp/AmazonCloudFront/latest/DeveloperGuide/lambda-examples.html#lambda-examples-overriding-response- header
2. there is a case like this where both event.Records[0].cf.request and event.Records[0].cf.response are handled. In the case of viewer-response and origin-response
https://docs.aws.amazon.com/ja_jp/AmazonCloudFront/latest/DeveloperGuide/lambda-examples.html#lambda-examples-update-error-status- examples

PR #1251, opened 2023-07-20
feat(adaptor): Lambda@Edge supports response callbacks

watany-dev:
adding

1. Cloudflont response can now be handled.

```ts
type Bindings = {
  callback: Callback
  request: CloudFrontRequest
  response: CloudFrontResponse
}

app.get('/callback/response', async (c, next) => {
  await next()
  c.env.callback(null, c.env.response)
})

```

2. Headers can now be added in OriginResponse. (I have a feeling we can do this in a much cooler way)

```ts
  app.get('/header/add', async (c, next) => {
    c.env.response.headers['Strict-Transport-Security'.toLowerCase()] = [{
      key: 'Strict-Transport-Security',
      value: 'max-age=63072000; includeSubdomains; preload'
    }];
    c.env.response.headers['X-Custom'.toLowerCase()] = [{
      key: 'X-Custom',
      value: 'Foo'
    }];
    await next()
    c.env.callback(null, c.env.response)
  })
```

3. refactoring. I cut out a lot of the internal header manipulation process.

### DIFF
--- a/runtime_tests/lambda-edge/index.test.ts
+++ b/runtime_tests/lambda-edge/index.test.ts
@@ -43,21 +43,13 @@ describe('Lambda@Edge Adapter for Hono', () => {
   app.get('/auth/abc', (c) => c.text('Good Night Lambda!'))
 
   app.get('/header/add', async (c, next) => {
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @watany-dev !

LGTM! Great work!

PR #1260, opened 2023-07-23
Draft: decouple response creation from Context class

Mahamed-Belkheir:
split creating Responses to its own class, and let Context class extend the response creator class, this allows the current header objects init optimizations to remain unchanged

### DIFF
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -1,14 +1,12 @@
 import { HonoRequest } from './request.ts'
+import { HonoResponder } from './response.ts'
 import { FetchEventLike } from './types.ts'
-import type { Env, NotFoundHandler, Input, TypedResponse } from './types.ts'
+import type { Env, NotFoundHandler, Input } from './types.ts'
 import type { CookieOptions } from './utils/cookie.ts'
 import { serialize } from './utils/cookie.ts'
-import type { StatusCode } from './utils/http-status.ts'
-import type { JSONValue, InterfaceToType } from './utils/types.ts'
 
 type Runtime = 'node' | 'deno' | 'bun' | 'workerd' | 'fastly' | 'edge-light' | 'lagon' | 'other'
-type HeaderRecord = Record<string, string | string[]>
-type Data = string | ArrayBuffer | ReadableStream
+
 
 export interface ExecutionContext {
   waitUntil(promise: Promise<unknown>): void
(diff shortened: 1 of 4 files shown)


yusukebe:
Is this primarily for refactoring?

> this allows the current header objects init optimizations to remain unchanged

Could you elaborate on what you mean by this?

If this PR is solely for refactoring, while separating the Responser and Context may be a good idea, I would prefer to keep the current structure. In the Hono core, I would rather not create more classes.

Mahamed-Belkheir:
> Is this primarily for refactoring?
> 
> > this allows the current header objects init optimizations to remain unchanged
> 
> Could you elaborate on what you mean by this?

Originally I wanted to move the functionality to another class without using inheritence, and just using composition to call the methods needed, but that would require either:
- moving the lazy header initialization around
- passing around the headers, including all the different cases where they can be changed
- the json serializing happens before the response is created, that would need to changed too
so instead of going with that route, inheritance was a simpler solution.
> 
> If this PR is solely for refactoring, while separating the Responser and Context may be a good idea, I would prefer to keep the current structure. In the Hono core, I would rather not create more classes.

This refactoring is mainly to allow for code like:
```ts
const notFoundResponse = new HonoResponder().json({ status: "error", message: "item not found" });
// or
const notFoundResponse = (item: string) => new HonoResponder().json({ status: "error", message: `${item} not found` });
```

Mahamed-Belkheir:
The reason this is a draft is because of the following questions:

- Are we okay with the inheritance solution? This leaves Context functionally the same as before, but if in the future we need it to inherit another class, we'd have to either use mixins or have the other class extend HonoRespondor.
  - If we want to go with another solution (e.g Composition), Context would need to be refactored, how would we structure this?
- Do we have a better name than "HonoRespondor"? maybe ResponseBuilder or something.

yudai-nkt:
Just out of curiosity, how `HonoResponder.prototype.json` is different from `Response.json`?

yusukebe:
Hi @Mahamed-Belkheir,

Could you specify which part of the code you're referring to when you mention "moving the lazy header initialization around"?

Mahamed-Belkheir:
> Hi @Mahamed-Belkheir,
> 
> Could you specify which part of the code you're referring to when you mention "moving the lazy header initialization around"?

from what I understood we currently lazily initialize
- ._res (Response)
- ._h (Header)
- ._pH (Object)

none of these are called until needed. when setting headers before a response is created, we lazily initialize either _pH or _h and store the headers there, without creating a Response object.

Now assuming we have a Responder class, it needs to know about _pH and _h, we either:
- pass in _pH, _h  as parameters to all the Responder class **static** methods, **problems with that:**
  - weird API, everyone consuming Responder will be able to pass multiple headers (the normal parameter headers, alongside with _h and _pH) ((e.g json(data, 201, headers, _h, _ph)))
  - Context class still has to handle Response related functionality (headers management), so we did not gain the benefit of a clearer separation of concerns
- store the state in a Responder instance, so now it has _pH and _h state in itself, **problems with that:**
  - one new instance creation on every new Context by default (create Responder instance to store headers)
  - a lot of response related methods would be an empty shell like:
 
```ts
json: JSONRespond = <T = {}>(
        object: T,
        arg?: StatusCode | RequestInit,
        headers?: HeaderRecord
      ) => {
        return this.responder.json(object, arg, headers);
}
```

Last point can be mitigated by exposing this.responder/response, like we do with ctx.req, but then that's a breaking change as ctx would lose those methods.

But yeah, if you think either approach is fine, or know an alternative, I'd be willing to refactor more.

yusukebe:
Hi @Mahamed-Belkheir,

I've carefully considered this PR, but unfortunately, we can't accept it due to a decrease in performance:

Screenshot 2023-07-30 at 7 46 50

Here's the script that was used for benchmarking:

```ts
import { run, bench } from 'mitata'
import { Hono as HonoCurrent } from '../hono-repo/dist'
import { Hono as HonoNext } from './../../../dist'

const appCurrent = new HonoCurrent()
const appNext = new HonoNext()

const request = new Request('http://localhost/json')

const initHono = (app: any) => {
  app.get('/json', async (c) =>
    c.json({
      message: 'hello',
    })
  )
}

initHono(appCurrent)
initHono(appNext)

bench('noop', () => {})

bench('Hono - current', async () => {
  await appCurrent.request(request)
})

bench('Hono - PR 1260', async () => {
  await appNext.request(request)
})

await run()
```

Refactoring should not lead to a decrease in speed, as performance is a priority for us. Thank you for your understanding.

PR #1263, opened 2023-07-24
fix(lambda-edge) Adding Export Type CloudFrontResponse and CloudFrontConfig

watany-dev:
- CloudFrontResponse: 
    - I simply forgot to export it. My apologies!
- CloudFrontConfig: 
    - At first, I thought it wasn't necessary. However, I found a sample online that requires it, so I've included it: https://www.withcoherence.com/post/aws-spa-routing-the-bad-the-ugly-and-the-uglier.


yusukebe:
LGTM! Let's go!

PR #1266, opened 2023-07-25
fix escapeToBuffer vulnerability

Codinak:
Escape single quote (0x27) to limit execution capabilities under certain conditions.

Example, using JSX:
```tsx
const value = "alert('works')"
server.get("/*", c => c.html(<script>{value}</script>))
```
```tsx
const value = "alert('bar!')"
server.get("/*", c => {
  return c.html(<Html><div onmouseover={value}>foo</div></Html>)
})
```

The condition under which it works are limited and should be avoided, but in reality its never as straightforward as in this simple example. Adjusted behavior mimics that of react-dom escaping single quotes.

Here without jsx. More dangerous, you can escape properties and inject script, because single quotes are valid attribute value delimiters and in this example are not enforced like they are in jsx:
```ts
const value = "' onmouseover='alert(`bar!`)' data-x='"
server.get("/*", c => {
  return c.html(html`<html><div class='${value}'>foo</div></html>`)
})
```

### DIFF
--- a/deno_dist/utils/html.ts
+++ b/deno_dist/utils/html.ts
@@ -5,7 +5,7 @@ export type StringBuffer = [string]
 // The `escapeToBuffer` implementation is based on code from the MIT licensed `react-dom` package.
 // https://github.com/facebook/react/blob/main/packages/react-dom/src/server/escapeTextForBrowser.js
 
-const escapeRe = /[&<>"]/
+const escapeRe = /[&<>'"]/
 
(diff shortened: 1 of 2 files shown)


yusukebe:
@Codinak 

This is a good fix. One thing, could you run `yarn denoify` command? This is needed to generate file for Deno. And please git add&commit&push it. Thanks!

yusukebe:
Thanks! I'll merge it now!

PR #1269, opened 2023-07-26
added generic parameters to aws-lambda handler to fix #1265

### LINKED ISSUE #1265: Type issue with aws-lambda handler
When adding variables to the context object as demonstrated https://hono.dev/api/context#set-get and using the lambda-handler from 'hono/aws-lambda' e.g.:

```
import { handle } from 'hono/aws-lambda';
import { Hono } from 'hono';

type Test = string

export const app = new Hono<{  Variables: { test?: Test } }>();

export const lambdaHandler = handler(app);

```

you get a type error when passing the app to the handle:

```
Argument of type 'Hono<{ Variables: { test?: string | undefined; }; }, {}, "/">' is not assignable to parameter of type 'Hono<Env, {}, "/">'.
  The types of 'route(...).basePath(...).onError' are incompatible between these types.
    Type '(handler: ErrorHandler<{ Variables: { test?: string | undefined; }; }>) => Hono<{ Variables: { test?: string | undefined; }; }, Record<string, unknown>, any>' is not assignable to type '(handler: ErrorHandler<Env>) => Hono<Env, Record<string, unknown>, any>'.
      Types of parameters 'handler' and 'handler' are incompatible.
        Types of parameters 'c' and 'c' are incompatible.
          Type 'Context<{ Variables: { test?: string | undefined; }; }, any, {}>' is not assignable to type 'Context<Env, any, {}>'.
            Types of property 'env' are incompatible.
              Type 'unknown' is not assignable to type 'Bindings | undefined'.ts(2345)
```

I think it's fixed by adding the Hono generics to the handle function to pass them through?

DianomiJH:
fixes [#1265](https://github.com/honojs/hono/issues/1265)

### DIFF
--- a/src/adapter/aws-lambda/handler.ts
+++ b/src/adapter/aws-lambda/handler.ts
@@ -1,6 +1,7 @@
 // @denoify-ignore
 import crypto from 'crypto'
 import type { Hono } from '../../hono'
+import type { Env } from '../../types'
 
 import { encodeBase64 } from '../../utils/encode'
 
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @DianomiJH 

Thanks. I'll merge it!

Amen-Musingarimi:
Hello guys. I need your help on importing { handler } from 'hono/aws-lambda". When I run my API gate way linked with this Hono function that has this important I get a handler error. Seem there is no package called hono/aws-lambda. And when i try to install it i get the same error that package doesn't exist. What could I be doing wrong.

PR #1270, opened 2023-07-26
fix(trie-router): don't allow dot segments

yusukebe:
It should not allow don segments in URL path: `/../hello` to avoid a directory traversal attack.


yusukebe:
This issue should fixed by only #1271.

PR #1278, opened 2023-07-31
doc: Adding a Supported Runtime - Lambda@Edge

watany-dev:
I've finished writing the 'GetStarted' section; may I add it?


yusukebe:
Hi @watany-dev !

Yeah, finished! Let's go with it!

Issue #1164, opened 2023-06-06
`setCookie`'s `signed` option is not being used


dnlsandiego:
Hey folks! I noticed that `setCookie` has a `signed` option and I assumed it was for getting the cookie value signed before it's set as a header. But after looking into [setCookie](https://github.com/honojs/hono/blob/main/src/middleware/cookie/index.ts) and the [serialize](https://github.com/honojs/hono/blob/main/src/utils/cookie.ts#L25) util it uses, they are not reading or doing anything for the `signed` option.

I suggest that setCookie should get updated to sign cookies if that option is set as it's a core best practice for protecting cookies from being tampered. [cookie-signature](https://github.com/tj/node-cookie-signature) is a good library to do so. I can submit a PR if this makes sense.

yusukebe:
Hi @dnlsandiego!

You're absolutely right! It's defined but not used 😇

> I suggest that setCookie should get updated to sign cookies if that option is set as it's a core best practice for protecting cookies from being tampered.

That makes sense. It should be implemented.

> I can submit a PR if this makes sense.

Yes, please! But there's one thing to consider. Hono is built using only Web Standard APIs. So we should not use external libraries. "cookie-signature" seems like a good library but we can't use it. So, you'll need to write the code from scratch or port it from that library.

Anyway, thanks for creating the issue.

chrisbrocklesby:
Just notice the signed cookie issue myself (Came here to report the issue)... Would love to use signed cookies with Hono. Thanks.

chrisbrocklesby:
Hey Guys,
Any news or joy with this...

Also, I don't know if this would help but this seems to be a great library to port or base from : https://www.npmjs.com/package/sign

Thanks,
Chris.

yusukebe:
Hi @chrisbrocklesby  !

As of now, we haven't made any progress on this issue. However, you're right that we can use the library you shared as a reference. When I get the chance, I'll tackle this issue. Until then, any contributions are always welcome :)

yusukebe:
Closed!

PR #1279, opened 2023-07-31
feat(middleware): Simple cookie signing functionality

torte:
Relates issue: https://github.com/honojs/hono/issues/1164

Related PR for website/docs: https://github.com/honojs/website/pull/77

Tried to be backwards compatible with the current `setCookie`/`getCookie` API, so decided to introduce new functions for this instead (similar to how express cookie-parser would do that).

Some things I wasn't sure about:
* If a cookie signature is not valid, what should happen with the cookie. At the moment it would return `undefined` in the `getSignedCookie` method. This might be misleading and hiding errors though, but throwing an error here instead might be too invasive. How would you handle this and at notify that the cookie was tampered with?
* Thinking about this a bit more, `getSignedCookie` should probably just work exactly like `getCookie` for multiple cookies (`key` not specified) and return all cookies, but only compare the signature for cookies where a signature is provided (and otherwise ignore).
* Somehow related to previous concern, `getCookie` without a `key` specified does return the signed cookie as well (without checking and removing the signature, so the value of the cookie would be the actual value + signature. Would that be intended behaviour or should signed cookies be removed from a `getCookie` result?

Please let me know if changes are needed.

### DIFF
--- a/deno_dist/middleware/cookie/index.ts
+++ b/deno_dist/middleware/cookie/index.ts
@@ -1,10 +1,15 @@
 import type { Context } from '../../context.ts'
-import { parse, serialize } from '../../utils/cookie.ts'
-import type { CookieOptions, Cookie } from '../../utils/cookie.ts'
+import { parse, parseSigned, serialize, serializeSigned } from '../../utils/cookie.ts'
+import type { CookieOptions, Cookie, SignedCookie } from '../../utils/cookie.ts'
 
 interface GetCookie {
   (c: Context, key: string): string | undefined
   (c: Context): Cookie
+}
+
+interface GetSignedCookie {
+  (c: Context, sercet: string, key: string): Promise<string | undefined | false>
+  (c: Context, secret: string): Promise<SignedCookie>
 }
 
 export const getCookie: GetCookie = (c, key?) => {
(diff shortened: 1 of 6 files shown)


yusukebe:
Hi @torte !

Thank you for creating the PR. Adding `parseSigned` and `serializeSigned` seems like a good idea. I'll review others later. Please wait.

Hi @dnlsandiego, @chrisbrocklesby

Please take a look at this when you have time.

yusukebe:
@torte 

> If a cookie signature is not valid, what should happen with the cookie?

Indeed, it's better to behave differently from the situation where the value does not exist, so `undefined` probably is not preferred. In [Express's cookie-parser](https://expressjs.com/en/resources/middleware/cookie-parser.html), it returns `false` when it's not valid:

Screenshot 2023-08-02 at 9 05 06

I think returning `false` is good, following Express's approach. Alternatively, we can have it show warnings with `console.warn()` for the developer.

> Would that be intended behaviour or should signed cookies be removed from a getCookie result?

I think `getCookie()` should not return signed cookie values. Express's `req.cookies`, not `req.signedCookies`, does not return signed cookies. In my opinion, signed values should not be obtained from the `getCookie()` method.

> but only compare the signature for cookies where a signature is provided (and otherwise ignore).

So, it's better to ignore the value when the signature is not provided, because signed cookies and normal cookies should be distinguished as mentioned above.

This is my opinion.

---

@usualoma @metrue @ThatOneBro  and others:

If you have any opinions, please share them.

torte:
Yep, that makes sense. Will make the changes later if there isn't any disagreement on this.

I won't introduce a `context.signedCookies` property, since `context.cookies` is depcrecated for v4. 

`getSignedCookie` without a key will return all signed cookies and attempt to validate the signature. Might need to swap the arguments `key` <-> `secret` in that function so it makes more sense.

yusukebe:
@torte

Totally, I agree. Could you make the changes when you have time? Thanks!

torte:
@yusukebe Done from my end. Beside the things I mentioned in my last comment, I also changed `parse` and `parseSigned` a bit to accept an optional `name` parameter. Reason for that is that I don't think hono should try to compare signatures for other signed cookie if a cookie name was specified, mostly to preserve CPU time. The same change was made for `parse` to be consistent.

yusukebe:
>  I also changed `parse` and `parseSigned` a bit to accept an optional name parameter.

Make sense. Let's go with it!

Thanks!

torte:
Nice and thank you for reviewing.

Don't forget about the docs PR as well: https://github.com/honojs/website/pull/77

yusukebe:
@torte 

Okay. I'll merge it after releasing new version.

PR #1307, opened 2023-08-09
Add missing wasm mime type

timfish:
Wasm wont load in the browser without a corect Content-Type!


yusukebe:
Hi @timfish 

You are right! One thing, could you do `yarn denoify`, and git add&commit&push the generated file. It is needed for Deno.

timfish:
Done!

PR #1309, opened 2023-08-09
Export Schema type to allow route exports in declaration files

### LINKED ISSUE #1308: TypeScript cannot name route type in declaration files
### Problem

When exporting a Hono route with the compiler option  `declaration: true` inside a module's `tsconfig.json`, the TypeScript compiler complains that the route cannot be named without referencing `node_modules/hono/dist/types/types`:

```
The inferred type of 'route' cannot be named without a reference to './node_modules/hono/dist/types/types'. This is likely not portable. A type annotation is necessary.ts(2742)
```

The infered type of the route references Hono's `Schema` type, which can be found at `node_modules/hono/dist/types/types` but isn't exported anywhere and therefore cannot be named in a TypeScript declaration file.

### Example

**index.ts**
```ts
import { Hono } from "hono";

const app = new Hono();

export const route = app.get("/", (c) => c.json({}));
```

**tsconfig.json:**
```json
{
  "compilerOptions": {
   ...
    "declaration": true
   ...
  }
}
```

In the above example, the infered type of `route` is `Hono<Env, Schema<"get", "/", unknown, {}>, "/">`.
The types `Hono` and `Env` are exported in 'src/index.ts', but `Schema` isn't exported anywhere and therefore cannot be named in TypeScript declaration files.

### Impact

Exporting Hono routes and their type is particularly relevant when using Hono's RPC feature in a monorepo since the route types can be separated into their own module and imported by other packages to be used in conjunction with `hono/client` (see [create-t3-turbo](https://github.com/t3-oss/create-t3-turbo/blob/main/packages/api/index.ts) for an example of such an architecture).

### Solution

Adding `Schema` to the exported types in `src/index.ts` resolves the issue since all referenced types can then be imported in a TypeScript declaration file.

See #1309

rosahaj:
Fixes #1308

### DIFF
--- a/deno_dist/mod.ts
+++ b/deno_dist/mod.ts
@@ -16,6 +16,7 @@ export type {
   NotFoundHandler,
   ValidationTargets,
   Input,
+  Schema,
   TypedResponse,
 } from './types.ts'
 export type { Context, ContextVariableMap } from './context.ts'
--- a/src/index.ts
+++ b/src/index.ts
@@ -10,6 +10,7 @@ export type {
   NotFoundHandler,
   ValidationTargets,
   Input,
+  Schema,
   TypedResponse,
 } from './types'
 export type { Context, ContextVariableMap } from './context'
--- a/src/mod.ts
+++ b/src/mod.ts
@@ -16,6 +16,7 @@ export type {
   NotFoundHandler,
   ValidationTargets,
   Input,
+  Schema,
   TypedResponse,
 } from './types'
 export type { Context, ContextVariableMap } from './context'


yusukebe:
Hi @rosahaj,

I understand the problem, and this is a good solution. I want to merge it, but there's one thing you need to do. Please run `yarn denoify`, then add, commit, and push the generated file. This step is necessary for Deno. Thanks.

rosahaj:
> Hi @rosahaj,
> 
> I understand the problem, and this is a good solution. I want to merge it, but there's one thing you need to do. Please run `yarn denoify`, then add, commit, and push the generated file. This step is necessary for Deno. Thanks.

I have run `yarn denoify`, but it didn't change any files since Deno uses `src/mod.ts`, not `src/index.ts`. I have applied the change to `src/mod.ts` as well and ran `yarn denoify`.

By the way, have you thought about automatically running `denoify` through GitHub actions for every created PR?

yusukebe:
Hi @rosahaj 

> I have run `yarn denoify`, but it didn't change any files since Deno uses `src/mod.ts`, not `src/index.ts`. I have applied the change to `src/mod.ts` as well and ran `yarn denoify`.

Ah, you are right. That's not be emitted. Thanks!

> By the way, have you thought about automatically running `denoify` through GitHub actions for every created PR?

Yeah, sounds good. Do you have any idea? I though it but, I couldn't touch it yet.

Anyway, I'll merge this PR now!

PR #1316, opened 2023-08-10
perf: make parsing a query fast

yusukebe:
For Hono, being "fast" is paramount.

Below is a link to one of the benchmark projects for Bun's web frameworks:

[https://github.com/SaltyAom/bun-http-framework-benchmark](https://github.com/SaltyAom/bun-http-framework-benchmark)

A benchmark is just a benchmark, but it's important as many users refer to it as a guideline to determine which framework is better.

In this benchmark, the fastest frameworks are overly optimized for benchmarking, making them nonsensical to compare with Hono. However, "Elysia" is a good framework for practical use cases. Elysia is also optimized for Bun, making its benchmark scores fast. Moreover, it now supports the "`fetch`" interface, as seen in applications like Cloudflare Workers:

```ts
const app = new Elysia({
  aot: false,
}).get('/', () => 'Hi')

export default {
  fetch: app.fetch,
}
```

With this syntax, the conditions between Hono and Elysia are the same, so I want Hono to outperform Elysia using this approach.

To that end, I've revised the parsing of query parameters since Hono's score for "Params, query & header" was low. Then the result was:

Screenshot 2023-08-11 at 0 17 39

The score for "Get (/)" may be slower, but it will be faster in "Params, query & header," even though it's still slower than Elysia's.

The strategy behind this change is to retrieve the first query parameter `{ name: "bun" }` at the same time as `getPathWithFirstQuery()`. I'm not sure if this is the best or even a better solution, but the score should be improved.

For performance measurement, I'm using this script: 

```ts
import { Elysia } from 'elysia'
import { Hono as CurrentHono } from 'hono'
import { Hono } from '../../../dist'

const appHono = new Hono({ router: new RegExpRouter() })
  .get('/', (c) => c.text('Hi'))
  .get('/id/:id', (c) => {
    const id = c.req.param('id')
    const name = c.req.query('name')
    return c.text(`${id} ${name}`, {
      headers: {
        'x-powered-by': 'benchmark',
      },
    })
  })
  .post('/json', (c) => c.req.json().then(c.json))

const currentHono = new CurrentHono({ router: new RegExpRouter() })
  .get('/', (c) => c.text('Hi'))
  .get('/id/:id', (c) => {
    const id = c.req.param('id')
    const name = c.req.query('name')
    return c.text(`${id} ${name}`, {
      headers: {
        'x-powered-by': 'benchmark',
      },
    })
  })
  .post('/json', (c) => c.req.json().then(c.json))

const appElysia = new Elysia({
  aot: false,
})
  .get('/', () => 'Hi')
  .get('/id/:id', (c) => {
    c.set.headers['x-powered-by'] = 'benchmark'
    return `${c.params.id} ${c.query.name}`
  })
  .post('/json', (c) => c.body)

bench('noop', () => {})

group({ name: 'Get (/)' }, () => {
  const request = new Request('http://localhost/')

  bench('Hono - next', async () => {
    await appHono.fetch(request)
  })

  bench('Hono - current', async () => {
    await currentHono.fetch(request)
  })

  bench('Elysia', async () => {
    await appElysia.fetch(request)
  })
})

group({ name: 'Params, query & header' }, () => {
  const request = new Request('http://localhost/id/1?name=bun')

  bench('Hono - next', async () => {
    await appHono.fetch(request)
  })

  bench('Hono - current', async () => {
    await currentHono.fetch(request)
  })

  bench('Elysia', async () => {
    await appElysia.fetch(request)
  })
})

const createRequest = () => {
  return new Request('http://localhost/json', {
    method: 'POST',
    body: `{
      "hello": "world"
    }`,
    headers: {
    },
  })
}

group({ name: 'Post JSON' }, () => {
  bench('Hono - next', async () => {
    const request = createRequest()
    await appHono.fetch(request)
  })

  bench('Hono - current', async () => {
    const request = createRequest()
    await currentHono.fetch(request)
  })

  bench('Elysia', async () => {
    const request = createRequest()
    await appElysia.fetch(request)
  })
})

await run()
```

### DIFF
--- a/src/compose.test.ts
+++ b/src/compose.test.ts
@@ -1,5 +1,6 @@
 import { compose } from './compose'
 import { Context } from './context'
+import { HonoRequest } from './request'
 
 type C = {
   req: Record<string, string>
@@ -8,6 +9,12 @@ type C = {
 }
 
 class ExpectedError extends Error {}
+
+const req = new HonoRequest(new Request('http://localhost/'), {
+  path: '/',
+  paramData: {},
+  firstQuery: undefined,
+})
 
 describe('compose', () => {
   const middleware: Function[] = []
@@ -79,8 +86,6 @@ describe('compose with returning a promise, non-async function', () => {
 
 describe('Handler and middlewares', () => {
   const middleware: Function[] = []
-
-  const req = new Request('http://localhost/')
   const c: Context = new Context(req)
 
   const mHandlerFoo = async (c: Context, next: Function) => {
@@ -116,7 +121,6 @@ describe('Handler and middlewares', () => {
 describe('compose with Context - 200 success', () => {
   const middleware: Function[] = []
 
-  const req = new Request('http://localhost/')
   const c: Context = new Context(req)
   const handler = (c: Context) => {
     return c.text('Hello')
@@ -140,7 +144,6 @@ describe('compose with Context - 200 success', () => {
 describe('compose with Context - 404 not found', () => {
   const middleware: Function[] = []
 
-  const req = new Request('http://localhost/')
   const c: Context = new Context(req)
   const onNotFound = (c: Context) => {
     return c.text('onNotFound', 404)
@@ -164,7 +167,6 @@ describe('compose with Context - 404 not found', () => {
 describe('compose with Context - 401 not authorized', () => {
   const middleware: Function[] = []
 
-  const req = new Request('http://localhost/')
   const c: Context = new Context(req)
   const handler = (c: Context) => {
     return c.text('Hello')
@@ -190,7 +192,6 @@ describe('compose with Context - 401 not authorized', () => {
 describe('compose with Context - next() below', () => {
   const middleware: Function[] = []
 
-  const req = new Request('http://localhost/')
   const c: Context = new Context(req)
   const handler = (c: Context) => {
     const message = c.req.header('x-custom') || 'blank'
@@ -217,7 +218,6 @@ describe('compose with Context - next() below', () => {
 describe('compose with Context - 500 error', () => {
   const middleware: Function[] = []
 
-  const req = new Request('http://localhost/')
   const c: Context = new Context(req)
 
   it('Error on handler', async () => {
@@ -266,7 +266,6 @@ describe('compose with Context - 500 error', () => {
   })
 })
 describe('compose with Context - not finalized', () => {
-  const req = new Request('http://localhost/')
   const c: Context = new Context(req)
   const onNotFound = (c: Context) => {
     return c.text('onNotFound', 404)
--- a/src/context.test.ts
+++ b/src/context.test.ts
@@ -1,8 +1,13 @@
 import { Context } from './context'
+import { HonoRequest } from './request'
+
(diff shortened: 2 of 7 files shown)


yusukebe:
Hi @usualoma !

How do you think about the benchmark and this PR? Let me know your opinion!

usualoma:
I'll check it out!

usualoma:
Hi @yusukebe !

I'm sorry for not getting back to you sooner.
It isn't something that doesn't make sense in the real world. Still, it is undeniably an optimization biased toward a particular situation. Hmmm, it is troubling.

I've written a lot of code and tried it, but in this situation, you are correct that we can speed up the process by using a regular expression to get the first parameter.

It is challenging to optimize stably because a slight change in assumptions can reverse the order when it comes to this point, but making the following two changes made it faster.
* Cache in `request.url` when possible
* Get the first parameter together with the router's regular expression match

https://github.com/honojs/hono/commit/6ea271c017c1f1fec9f97cdff9b05cdabf4554cf

It would be a certain amount of change (and increase the bundle size somewhat). Still, it is possible to implement it without affecting the other routers by adding the option to RegExpRouter. Would you try this kind of approach?

image

yusukebe:
Hi @usualoma!

Wow, that's fast! I've been considering the idea of capturing the path and the first query in the router as well. This change seems efficient.

> It isn't something that doesn't make sense in the real world. Still, it is undeniably an optimization biased toward a particular situation. Hmmm, it is troubling.

Indeed, it's a concern. We aim to keep the code simple and avoid bloating the file size. However, our top priority is to "make Hono as fast as possible". Given that Hono's core is relatively small, I believe a bit of complexity can be justified, especially since we have extensive tests to ensure correct behavior.

> It would be a certain amount of change (and increase the bundle size somewhat). Still, it is possible to implement it without affecting the other routers by adding the option to RegExpRouter. Would you try this kind of approach?

I'd like to go with it!

On a related thing, have you run the benchmark linked below on your local machine recently?

https://github.com/SaltyAom/bun-http-framework-benchmark

The benchmark I've presented above solely measures the "fetch" method. But, the above benchmark evaluates response times from an actual server. So the results will be different. I'm curious to hear your thoughts on the results.

Thanks!

yusukebe:
Hi @usualoma,

Have you started working on it already? If so, please stop it now.

I'm having second thoughts about whether we should implement this feature. Even though there's a potential performance boost, I'm concerned about making the code more complex. As you're aware, I appreciate the simplicity of our router's API. The `Result` interface, for instance, is very simple:

```ts
export interface Result<T> {
  handlers: T[]
  params: Record<string, string>,
}
```

Adding two more properties just for a slight speed edge might not be justified at this moment.

One of our significant selling points is that we have the fastest router in the JavaScript world. That in itself is a strong enough claim for "ultrafast" branding.

usualoma:
@yusukebe  OK, I haven't started anything yet.

I have looked at https://github.com/SaltyAom/bun-http-framework-benchmark and have yet to come up with any ideas for this.

yusukebe:
@usualoma 

Sorry to put you through all this trouble! Thanks.

yusukebe:
Closing!

PR #1325, opened 2023-08-14
Fix reflecting basePath when no path in handler

sizumita:

- [x] Add tests
- [ ] Run tests

### DIFF
--- a/deno_dist/types.ts
+++ b/deno_dist/types.ts
@@ -72,18 +72,29 @@ export interface HandlerInterface<
   //// app.get(...handlers[])
 
   // app.get(handler, handler)
-  <I extends Input = {}, O = {}>(
-    ...handlers: [H<E, ExtractKey<S>, I, O>, H<E, ExtractKey<S>, I, O>]
+  <
+    P extends string = ExtractKey<S> extends never ? BasePath : ExtractKey<S>,
+    I extends Input = {},
+    O = {}
+  >(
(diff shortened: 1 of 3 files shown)


yusukebe:
Hi @sizumita!

Thanks for your contribution.

Could you please address the error causing the CI to fail? And, I'd appreciate if you could add a test for the following pattern:

```ts
app.post('/books/:id').get((c) => {
  const foo = c.req.param('id');
  return c.text(foo);
});
```

sizumita:
Hello, @yusukebe!

> Hi @sizumita!
> 
> Thanks for your contribution.
> 
> Could you please address the error causing the CI to fail? And, I'd appreciate if you could add a test for the following pattern:
> 
> ```ts
> app.post('/books/:id').get((c) => {
>   const foo = c.req.param('id');
>   return c.text(foo);
> });
> ```

I'm sorry that I missed running test in local.
Since It's caused by missing type and invalid test syntax, I fixed it.

Thank you.

yusukebe:
@sizumita 

Thanks! One thing. Can you add the test code for this:

```ts
app.post('/books/:id').get((c) => {
  const foo = c.req.param('id');
  return c.text(foo);
});
```

sizumita:
@yusukebe 

I'm sorry for missing your order:pray:
I added it and tested in local.

yusukebe:
Thanks! Ship it.

Issue #1179, opened 2023-06-14
Form validator receives empty value on Bun


spiffytech:
I've tested the same code with Node and there's no issue.

Under bun, `req.formData()` returns a promise, which Hono doesn't `await`. I don't know why the promise has a `forEach()` method, that seems pretty weird, but it performs zero iterations.

```ts
import { Hono } from "hono";
import { validator } from "hono/validator";

const app = new Hono();

app.post(
  "/",
  validator("form", (value, ctx) => console.log(value)),
  (c) => c.text("Hello Hono!")
);

export default app
```

```
$ curl 'http://localhost:3000' -X POST  -H 'Content-Type: application/x-www-form-urlencoded' --data-raw 'foo=bar'
```

Result: 
```
{}
```

yusukebe:
Hi @spiffytech !

This is a known issue. It's caused by the cloned request object not working correctly on Bun.

https://github.com/honojs/middleware/issues/81#issuecomment-1509457656

yusukebe:
This was fixed!

Issue #567, opened 2022-09-30
Cache middleware throws warnings on Cloudflare Workers


yusukebe:
It throws the warnings:

```
Your worker called response.clone(), but did not read the body of both clones. This is wasteful, as it forces the system to buffer the entire response body in memory, rather than streaming it through. This may cause your worker to be unexpectedly terminated for going over the memory limit. If you only meant to copy the response headers and metadata (e.g. in order to be able to modify them), use `new Response(response.body, response)` instead.
```

ThatOneBro:
Yes I noticed a similar error before, if you terminate the request before reading the body with either `.json()` or `.body()` after cloning in the middleware (as we are doing), then it gives this error on Cloudflare. We probably need to look to see what are the resource/ performance impacts of this and if we should handle this another way or if there is an optimization to be added here.

implicit-invocation:
`zod-validator` (and `typebox-validator`) still calls to `'hono/validator'` and we are still having `value = await c.req.raw.clone().json()` inside `'hono/validator'`, so the warnings are still thrown when using those middlewares.

micromashor:
When using a validator, I don't see why it's necessary to clone the request. You would get the validated request body with `c.req.valid('json')`, leaving no reason to read the request stream in the route handler itself.

To cover odd cases where one wants to read the request stream directly, perhaps an option could be added to explicitly enable cloning.

yusukebe:
This may be fixed. If not, feel free to reopen. Thanks.

PR #1333, opened 2023-08-15
feat(req): cache body content

### LINKED ISSUE #1317: Cache the request body content

I think it's better to cache the content retrieved by like `c.req.json()` in `HonoRequest`. If we do so, we don't have to `clone()` response object.

yusukebe:
This PR allows `HonoRequest` to cache its body content retrieved from methods like `req.raw.json()`. The following methods will have their results cached:

* `json()`
* `text()`
* `arrayBuffer()`
* `blob()`
* `formData()`

Before this feature, if we needed to call `req.json()` more than once, we would have to use `clone()`. In validators, `clone()` was used to preserve the request for subsequent uses:

```ts
value = await c.req.raw.clone().json()
```

However, this approach triggered warnings in Cloudflare/Wrangler:

```
Your worker created multiple branches of a single stream (for instance, by calling `response.clone()` or `request.clone()`) but did not consume the body of both branches. This behavior is inefficient since it requires the entire stream of data to be buffered in memory rather than being streamed. As a result, your worker might be unexpectedly terminated for exceeding the memory limit. If your intention was only to copy the request or response headers and metadata (e.g., to modify them), consider using the appropriate constructors (e.g., `new Response(response.body, response)`, `new Request(request)`, etc.).
```

With this PR, the content is cached within the `HonoRequest` object, preventing such warnings.

This change does not degrade performance and involves minimal code adjustments. I believe it's a valuable enhancement.

--- 

Fix #1317 #1179 #567 https://github.com/honojs/middleware/issues/62

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -1,3 +1,5 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
 import type {
   Input,
   InputToDataByTarget,
@@ -14,12 +16,16 @@ import { parse } from './utils/cookie'
 import type { UnionToIntersection } from './utils/types'
 import { getQueryParam, getQueryParams, decodeURIComponent_ } from './utils/url'
 
+type BodyType = 'json' | 'text' | 'arrayBuffer' | 'blob' | 'formData'
+type BodyCache = Partial<Record<BodyType, any>>
+
 export class HonoRequest<P extends string = '/', I extends Input['out'] = {}> {
   raw: Request
 
   private paramData: Record<string, string> | undefined
   private vData: { [K in keyof ValidationTargets]?: {} } // Short name of validatedData
   path: string
+  private bodyCache: BodyCache = {}
 
   constructor(
     request: Request,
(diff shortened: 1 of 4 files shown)


yusukebe:
Hi @usualoma @ThatOneBro 

What do you feel about this PR? Let me know your opinion.

src/request.ts
```diff
@@ -115,28 +121,34 @@ export class HonoRequest<P extends string = '/', I extends Input['out'] = {}> {
   }
 
   async parseBody<T extends BodyData = BodyData>(): Promise<T> {
-    return await parseBody(this.raw)
+    return await parseBody(this)
+  }
+
+  private cachedBody = (key: keyof BodyCache) => {
+    const cachedBody = this.bodyCache[key]
```
ThatOneBro, line 128: I like this a little better for perf since it reduces calls to getters. But maybe less clear?
```ts
    const { bodyCache, raw } = this
    const cachedBody = bodyCache[key]
    if (cachedBody) return cachedBody
    return (bodyCache[key] = raw[key]())
```

ThatOneBro (approved):
Looks like a good change to remove the excessive cloning! Glad that's finally gone. Good work

src/request.ts
yusukebe: Ah, you are right. Going with it. Thanks!

yusukebe: Wait. I think these are just accessing to props, so they shouldn't affect performance, right?

yusukebe: @ThatOneBro 

Thanks for explanation the details! yeah, let's go with your code!

yusukebe:
Thanks @ThatOneBro !

We'll go with this.

PR #1340, opened 2023-08-16
Web query undefined check

Tanner-Scadden:
### Notes
- Opened this PR after I was confused why conditional code was being ran in my API when using the RPC client. Realized that I had a query string that contained `&filter=undefined` that was then parsed into `{ filter: 'undefined'}`
- Added check for undefined value in query to prevent API receiving values of `undefined`.


Tanner-Scadden:
> Hi @Tanner-Scadden !
> 
> Thank you for the PR. Looks good! I have one request before merging. Could you run `yarn denoify` and git add&commit&push generated files? They are needed for Deno.

done!

yusukebe:
Thanks! Let's go with this!

Issue #798, opened 2023-01-11
add helmet security for hono


hanumanjiblog:
please add helmet security to secure our api

https://www.npmjs.com/package/helmet

renhiyama:
You cant use helmet with hono. Atleast not out of the box. Helmet provides expressjs like middleware, (ie. `(req, res, next)`) rather than a low level configuration where you can fill in `c.req` and other objects. Therefore, we got two options to choose from:
- find another package that provides low level access and not just like an expressjs compatible API.
- try to polyfill support for expressjs api, I wouldn't recommend since it provides extra overhead during every request.

jackdbd:
I wrote a generic library that can help you generating a `Content-Security-Policy` header (only that particular header, not all other headers supported by Helmet).

https://github.com/jackdbd/content-security-policy

I'm using this library in an Eleventy plugin of mine. Using the library in a Hono middleware shouldn't be different.

The only problem I see is that its dependendencies might not work in the JS runtime where the Hono app is deployed. For example, my library currently depends on [Joi](https://github.com/hapijs/joi), which in turn depends on [@hapi/hoek](https://github.com/hapijs/hoek), which uses `Buffer`. But `Buffer` is not available on Cloudflare Pages Functions (and also Cloudflare Workers I think).

PR #1341, opened 2023-08-16
feat(middleware): Alternative middleware equivalent to Helmet

watany-dev:
I created this middleware as an experimental equivalent to Helmet. Originally made for my own use, I've opened a PR in this repository because there was a request for Helmet in the Github Issues. One difference from the official Helmet is that this middleware only adopts headers that can be set with fixed values, and users can only set each header to true/false.

I'm still undecided about the name of the middleware. It's not a pure Helmet. Names like 'mini-helmet' or 'kabuto' could be interesting, but for now, I've opted for a straightforward naming.

- [x] Add tests
- [x] Run tests

## Doc(sample)

---
### Secure Header Middleware
This middleware simplifies the setup of security headers. Inspired in part by the capabilities of Helmet, it allows you to control the activation and deactivation of specific security headers.

#### Installation
The middleware is built into the Hono framework, so no additional installation is required.

#### Usage
```ts
import { secureHeader } from 'hono/secure-header';

app.use('*', secureHeader({ xFrameOptions: false, xXssProtection: false }));

```

fix: https://github.com/honojs/hono/issues/798


yusukebe:
Cooooool! Really wanted it. I'll check it later!

watany-dev:
Please also refer to the documentation.
https://github.com/honojs/website/pull/83

src/middleware/secure-header/index.ts
```diff
@@ -0,0 +1,63 @@
...
+    .filter(([key]) => options[key as keyof SecureHeaderOptions])
+    .map(([, value]) => value)
+
+  return async (ctx, next) => {
+    headersToSet.forEach(([header, value]) => {
+      ctx.res.headers.set(header, value)
+    })
+
+    ctx.res.headers.delete('X-Powered-By')
+    await next()
```
yusukebe, line 61: The procedures to manipulate headers should be placed on after `await next()`. For example, this test fails:

```ts
it('should remove x-powered-by header', async () => {
  const app = new Hono()
  app.use('*', secureHeader())
  app.use('*', poweredBy())

  expect(res.headers.get('x-powered-by')).toBeFalsy()
})
```

src/middleware/secure-header/index.test.ts
```diff
@@ -0,0 +1,58 @@
...
+describe('Secure Header Middleware', () => {
+  it('all headers enabled', async () => {
+    const app = new Hono()
+    app.use('*', secureHeader())
+    app.get('/test', async (ctx) => {
+      return ctx.text('test')
+    })
+
+    const req = new Request('http://localhost/test', { method: 'GET' })
+    const res = await app.request(req)
```
yusukebe, line 13: You can write:

```ts
const req = new Request('http://localhost/test', { method: 'GET' })
```

to, as one line:

```ts
```

yusukebe:
Hi @watany-dev,

Thank you for the excellent PR! It's evident that this is a much-desired feature, and I'm grateful that you've taken it on.

About the naming: I understand the familiarity of "Helmet" for developers who have experience with Express. However, for those who might not have that background (like myself), it could be a tad confusing.

I concur that `secure-header` is more descriptive and self-explanatory. That being said, considering the plural nature (as we're dealing with multiple headers), `secure-headers` might be more apt. For instance, the `Response` object utilizes the `headers` property, not `header`:

```ts
const res = new Response()
res.headers.get('foo')
```

I've also added some comments in the PR; kindly see them.

Lastly, it's imperative for users to be informed about the positioning of this middleware. As illustrated by the following test, it's evident that order matters:

```ts
it('should remove x-powered-by header', async () => {
  const app = new Hono()
  app.use('*', poweredBy())
  app.use('*', secureHeader())

  expect(res.headers.get('x-powered-by')).toBeFalsy()
})
```

We'll need to highlight this aspect in our documentation on the website.

watany-dev:
@yusukebe 
Thank you for the comment. I've made the corrections; could you please check?

src/middleware/secure-header/index.test.ts
watany-dev: ok! refactored.

src/middleware/secure-header/index.ts
watany-dev: fixed.

zisra:
Some questions/concerns
- Are there plans for including support for the CSP header? (Which is also probably the most important)...
- ... as well as `X-Powered-by` and `Cross-Origin-Embedder-Policy` seem to be missing
- Can we add links to the MDN docs of the respective to the docs?
-`Origin-Agent-Cluster` is included within the docs AND "not supported" list 🤔

I would be very happy to work on it if you don't want to, just let me know

yusukebe:
Hi @watany-dev,

Thank you for the revisions. Also, please take into account the feedback provided by @zisra.

I believe:
> "Can we add links to the MDN docs of the respective to the docs?"

is a great idea.

yusukebe (approved):

watany-dev:
@zisra 
Thank you for the feedback! I believe we've been able to set up the CSP header for general use cases. We've also addressed the other comments. (We'll provide the CSP documentation at a later date.)

watany-dev:
- Address CI errors
- Instructions on using CSP
- Document differences in default values between our implementation and Helmet
- Document the order of middleware usage

src/middleware/secure-headers/index.test.ts
```diff
@@ -0,0 +1,154 @@
+import { Hono } from '../../hono'
```
yusukebe, line 1: You can add the follow line on the top to fix lint errors:

```ts
/* eslint-disable quotes */
```

This allow using double quotes.

watany-dev: thank you !! fix it.

watany-dev:
@yusukebe 
Thank you. I believe I've corrected the areas that needed fixing. There is a part where CI fails with a 429 error, but it seems to be a temporary error unrelated to this matter. I will add "workflow_dispatch" to ci.yaml in a separate PR so that it can be executed manually.

watany-dev:
This is the latest version, including the document.

yusukebe (changes requested):
@watany-dev 

As you said, please do `denoify`! This is a last thing to do!

watany-dev:
Sorry for the wait. It's done now.

yusukebe (approved):
LGTM!

yusukebe:
@watany-dev 

Thank you very much. You've done the awesome work. The next release (minor change) includes this feature will be released later. Thanks again!

Issue #1337, opened 2023-08-15
Cookie helper can't handle with `.`


yusukebe:
### What version of Hono are you using?

3.4.1

### What runtime/platform is your app running on?

Cloudflare and others

### What steps can reproduce the bug?

Using the code below:

```ts
import { getCookie, setCookie } from 'hono/cookie'

const app = new Hono()

app.get('/', (c) => {
  setCookie(c, 'width', '375')
  setCookie(c, 'ratio', '2.5')

  const width = getCookie(c, 'width')
  const ratio = getCookie(c, 'ratio')

  return c.json({
    width,
    ratio,
  })
})
```

And access to `GET /`.

### What is the expected behavior?

```json
{"width":"375","ratio":"2.5"}
```

### What do you see instead?

```json
{}
```

### Additional information

The reason causes this problem is using `.` to separate a signed string:

https://github.com/honojs/hono/blob/954c06ed4bc25680c46d0ef51959aeb7031ca353/src/utils/cookie.ts#L38

torte:
@yusukebe Yeah, I can check. The `.` is used to seperate the value from the signature, since a) it is common practice and b) it can stays the same url encoded.  I was thinking about this problem as well while I added the signature, but didn't have a good idea on how to solve it at that time. Any suggestions?

yusukebe:
@torte 

Thanks for your quick comment. I have no idea now, but I think it's better to use `.` to separate values for signature.

torte:
Ok, I'll think about something.

torte:
Ok, added a fix here: https://github.com/honojs/hono/pull/1342

Fix is backwards compatible. No change on how signatures are created.
I simply used the knowledge that a valid signature (as long it is HMAC SHA256 generated and base64 encoded) is always 44 characters long and ends with an `=`. If this condition is not satisfied for the cookie value the cookie value is considered a regular cookie and not considered signed. This should cover most use-cases with a `.` in the cookie value. 
Of course, there is still a very low chance somebody creates a cookie value seperated by one dot, with the second value being 44 characters long and ending with `=` :grinning:

yusukebe:
@torte 

That's awesome solution! And it's great to have backwards compatible.

> Of course, there is still a very low chance somebody creates a cookie value seperated by one dot, with the second value being 44 characters long and ending with = 😀

Indeed, but, while there's a small chance of that string combination occurring as you said, it should verify correctly, so there's no security concern. It looks good to me.

[...]

PR #1342, opened 2023-08-16
fix(middleware): signed cookie misinterpretation of cookie values with dot

torte:
Related issue: https://github.com/honojs/hono/issues/1337

Done the following:
- [x] Modified tests
- [x] Run tests

### DIFF
--- a/src/utils/cookie.ts
+++ b/src/utils/cookie.ts
@@ -21,6 +21,7 @@ const makeSignature = async (value: string, secret: string): Promise<string> => 
     'verify',
   ])
   const signature = await crypto.subtle.sign(algorithm.name, key, encoder.encode(value))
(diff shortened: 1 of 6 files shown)


src/utils/cookie.ts
```diff
@@ -35,7 +36,16 @@ export const parse = (cookie: string, name?: string): Cookie => {
   const parsedCookie: Cookie = {}
   const unsingedCookies = _parseCookiePairs(cookie, name).filter((pair) => {
```
yusukebe, line 37: It's typo! `unsingedCookies` should be `unsignedCookies`.

torte: Thanks, fixed.

yusukebe:
Thanks! Let's go!

PR #1345, opened 2023-08-18
feat(app): pass `env` to `getPath()`

### LINKED ISSUE #1320: Make `env` values available in custom `getPath` functions

One common use-case for custom `getPath` functions is to allow routing based on the host ([docs](https://hono.dev/api/routing#routing-with-host-header-value)). For example, when using Hono with Cloudflare Workers, it's possible to handle multiple subdomains within a single worker (as long as they belong to the same zone). However, workers are usually deployed to multiple environments (e.g. prod/staging) which could use different subdomain names for the same parts of the app, so it'd be really great if it was possible to specify those subdomain names as bindings and be able to access these bindings in the custom `getPath` function to avoid hardcoding ([here's](https://github.com/honojs/hono/issues/1240#issuecomment-1674100371) an example of such hardcoding).

yusukebe:
Resolve #1320

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/hono-base.ts
+++ b/src/hono-base.ts
@@ -65,7 +65,7 @@ class Hono<E extends Env = Env, S = {}, BasePath extends string = '/'> extends d
     To use it, inherit the class and implement router in the constructor.
   */
   router!: Router<H>
-  readonly getPath: (request: Request) => string
+  readonly getPath: (request: Request, options?: { env?: E['Bindings'] }) => string
   private _basePath: string = '/'
   private path: string = '/'
 
@@ -277,7 +277,7 @@ class Hono<E extends Env = Env, S = {}, BasePath extends string = '/'> extends d
     env: E['Bindings'],
     method: string
   ): Response | Promise<Response> {
-    const path = this.getPath(request)
+    const path = this.getPath(request, { env })
 
     // Handle HEAD method
     if (method === 'HEAD') {
--- a/deno_dist/hono-base.ts
+++ b/deno_dist/hono-base.ts
@@ -65,7 +65,7 @@ class Hono<E extends Env = Env, S = {}, BasePath extends string = '/'> extends d
     To use it, inherit the class and implement router in the constructor.
   */
   router!: Router<H>
-  readonly getPath: (request: Request) => string
+  readonly getPath: (request: Request, options?: { env?: E['Bindings'] }) => string
   private _basePath: string = '/'
   private path: string = '/'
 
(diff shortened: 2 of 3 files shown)


yusukebe:
Hi @baygeldin !

I've implemented the feature to resolve your request. Indeed, this will be useful for some users of Cloudflare Workers.

Could you check this PR?

src/hono-base.ts
```diff
@@ -65,7 +65,7 @@ class Hono<E extends Env = Env, S = {}, BasePath extends string = '/'> extends d
     To use it, inherit the class and implement router in the constructor.
   */
   router!: Router<H>
-  readonly getPath: (request: Request) => string
+  readonly getPath: (request: Request, options?: { env?: E['Bindings'] }) => string
```
baygeldin, line 68: Due to this type declaration I have to write `const env = options?.env ?? ({} as Env)` in `getPath` to get the bindings which feels a bit hacky/unnecessary. 

I tried changing this to:
```ts
readonly getPath: ((request: Request) => string) | ((request: Request, options: { env: E['Bindings'] }) => string)
```

So, basically some kind of overloading instead of an optional param. However, this didn't work out well:
- when declaring `getPath` as `getPath: (req) => { ... }` I get `Parameter 'req' implicitly has an 'any' type.` (i.e. this change would break backward compatibility)
- when declaring `getPath` as `getPath: (req, { env }) => { ... }` the `env` variable still has an optional type for whatever reason (i.e. `Bindings | undefined`)

I'm not super proficient in TypeScript, so I'm not exactly sure what's going on. Do you have any other idea on whether it'd be possible to make type declarations more user-friendly without introducing breaking changes?

baygeldin (approved):
Hey @yusukebe! Thank you so much for such a quick fix! I tested this locally and it works great. The only issue I have is with type declarations, but if you think it's OK feel free to merge! Thanks again for the fix.

src/hono-base.ts
yusukebe: Yeah, sometimes TypeScript can be troublesome.

In this case, it's preferable to keep backward compatibility with my implementation. I believe using `options!.env!` or `options?.env?` isn't too hacky. It simply translates to:

```ts
if (options && options.env) {
  //...
}
```

But thank you for the suggestion!

yusukebe:
@baygeldin 

Thanks for your review. I'll merge this now. Next version (maybe minor upgrade) includes this feature will be released later!

PR #1357, opened 2023-08-20
fix(types): corrected argument types for `Context` methods

asaxeye:
Each method is typed by Interface and doesn't affect the type of function arguments as seen by the caller, but wouldn't it be healthier to modify it?

- [x] Add tests
- [x] Run tests


yusukebe:
Hi @asaxeye !

Thank you!! This is typos/bugs. Looks good, merge now.

PR #1362, opened 2023-08-22
chore: Adding denoify check to CI

watany-dev:

Adding denoify check to CI

- [ ] Add tests
- [ ] Run tests


watany-dev:
Hmmm, ci doesn't work so I don't know how to check... I'll think about it

PR #1379, opened 2023-08-27
fix(context): different types using `jsonT()`

### LINKED ISSUE #1377: TypeScript error when using two different response types (error or success) using c.jsonT

3.5.4

Node

Create a simple route, in one case it should return one type (in my case, an error), and in another case, some other data (which is the actual data in my case). 

These are the types:

```
type Success<T> = {
    success: true;
    data: T;
};

type Error = {
    success: false;
    message: string;
};

type ResultType<T> = Success<T> | Error;
```

This is the simplest possible example:

```
endpoint.get("/all", c => {
    const data = getData();

    if (!data.success) {
        return c.jsonT({
            success: false,
            message: "Could not fetch data"
        });
    }

    return c.jsonT({
        success: true,
        data: data.data,
    });
});
```

In this example, it renders this very long error:

```
    Argument of type '(c: Context<{ Variables: AuthVariables; }, "/all", {}>) => TypedResponse<{ success: boolean; message: string; status: number; }> | TypedResponse<{ success: boolean; data: {}; }>' is not assignable to parameter of type 'H<{ Variables: AuthVariables; }, "/all", {}, { success: boolean; message: string; status: number; }>'.
      Type '(c: Context<{ Variables: AuthVariables; }, "/all", {}>) => TypedResponse<{ success: boolean; message: string; status: number; }> | TypedResponse<{ success: boolean; data: {}; }>' is not assignable to type 'MiddlewareHandler<{ Variables: AuthVariables; }, "/all", {}>'.
        Type 'TypedResponse<{ success: boolean; message: string; status: number; }> | TypedResponse<{ success: boolean; data: {}; }>' is not assignable to type 'Promise<void | Response>'.
          Type 'TypedResponse<{ success: boolean; message: string; status: number; }>' is missing the following properties from type 'Promise<void | Response>': then, catch, finally, [Symbol.toStringTag]ts(2769)
```

I would like to be able to return two different json bodies depending on the circumstances, and still be able to get type safety.

### What do you see instead?

I'm not sure for now, thinking about using normal c.json for the error response, but that would probably mean I would lose some type safety I presume. 

### Additional information

_No response_

yusukebe:
With this PR, `c.jsonT()` can now handle two or more distinct types:

```ts
const app = new Hono()

const routes = app.get('/all', (c) => {
  const data: Data = { success: true, data: {} }

  if (!data.success) {
    return c.jsonT({
      success: false,
      message: 'Could not fetch data',
    })
  }

  return c.jsonT({
    success: true,
    data: data.data,
  })
})
```

The returned types are as follows:

```ts
{
    success: boolean;
    message: string;
} & {
    success: boolean;
    data: any;
}
```

Screenshot 2023-08-27 at 16 19 41

Fix #1377.

### DIFF
--- a/deno_dist/client/types.ts
+++ b/deno_dist/client/types.ts
@@ -19,6 +19,12 @@ type ClientRequest<S extends Schema> = {
     : never
 }
 
+type BlankRecordToNever<T> = T extends Record<infer R, unknown>
+  ? R extends never
+    ? never
+    : T
+  : never
+
 export interface ClientResponse<T> {
   ok: boolean
   status: number
@@ -27,7 +33,7 @@ export interface ClientResponse<T> {
   url: string
   redirect(url: string, status: number): Response
   clone(): Response
-  json(): Promise<T>
+  json(): Promise<BlankRecordToNever<T>>
   text(): Promise<string>
   blob(): Promise<Blob>
   formData(): Promise<FormData>
--- a/deno_dist/types.ts
+++ b/deno_dist/types.ts
@@ -3,7 +3,7 @@
 /* eslint-disable @typescript-eslint/ban-types */
 import type { Context } from './context.ts'
 import type { Hono } from './hono.ts'
-import type { UnionToIntersection, RemoveBlankRecord } from './utils/types.ts'
+import type { UnionToIntersection } from './utils/types.ts'
 
 ////////////////////////////////////////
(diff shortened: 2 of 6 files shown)


banjo:
Wow @yusukebe, that is some first class library maintenance. Impressive (and unbelievably fast) job!

PR #1384, opened 2023-08-29
Add $url method to hono client to extract an endpoint's url

renzor-fist:
Hello,

This change is a proof of concept of a simple feature request. In our project, we are using Hono RPC and need to extract the URL of a Hono router endpoint for mocking in tests.
```
let fetchMock = getMiniflareFetchMock()

const { pathname, hostname } = hono.client.api.bar.$url()

fetchMock
        .get(hostname)
        .intercept({
          method: "POST",
          path: pathname,
        })
        .reply(200, {}, { headers: { "content-type": "application/json" } }),
```

This will mock the request to `http://localhost:8777/api/bar` by using the proposed `$url` method of the Hono client.

The `$url` method returns a URL representing the full url of the configured client request.

Do you think this is good to add to the client? If so, can you guide me in improving the type definitions to include `$url`?

Screenshot 2023-08-28 at 10 48 12 PM

And also to let me know if I should write more tests and if they are in the right file. Thank you for your amazing library!

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/deno_dist/client/client.ts
+++ b/deno_dist/client/client.ts
@@ -128,6 +128,10 @@ export const hc = <T extends Hono<any, any, any>>(
 
     const path = parts.join('/')
     const url = mergePath(baseUrl, path)
+    if (method === 'url') {
+      return new URL(url)
+    }
+
     const req = new ClientRequestImpl(url, method)
     if (method) {
       options ??= {}
--- a/deno_dist/client/types.ts
+++ b/deno_dist/client/types.ts
@@ -17,6 +17,8 @@ type ClientRequest<S extends Schema> = {
           options?: ClientRequestOptions
         ) => Promise<ClientResponse<O>>
     : never
+} & {
+  $url: () => Promise<URL>
 }
 
 export interface ClientResponse<T> {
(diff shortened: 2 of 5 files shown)


yusukebe:
Hi @renzor-fist !

This is a very interesting feature. I'll consider whether we should implement it and how to go about it. Please give me a minute. Thanks!

yusukebe:
Hi @renzor-fist,

After consideration regarding this feature, I've concluded that it would be a valuable addition to the main. Thank you for introducing such a great feature!

> If so, can you guide me in improving the type definitions to include `$url`?

You can simply use intersection types to achieve this. Update the following lines in the mentioned file:

https://github.com/honojs/hono/blob/4d8034ea89b6414f22554df7558b99833ec4cddc/src/client/types.ts#L10-L20

To look like:

```ts
type ClientRequest<S extends Schema> = {
  // existing code...
} & {
  $url: () => URL
}
```

For testing, it should suffice to verify against the actual values you've written. Should there be any type-related errors, they'll be thrown accordingly.

Thank you for your contribution.

yusukebe:
@renzor-fist 

Great work! LGTM.

I'll merge it, but given that this is a new feature rather than a mere fix, I'll merge it into the [`next` branch](https://github.com/honojs/hono/tree/next). This will be released for a minor version release, which should be coming up soon. Thank you for your contribution!

PR #1386, opened 2023-08-29
fix: error to retrieve cookie

uaichat:
TypeError: Cannot read properties of undefined (reading 'split')

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/helper/cookie/index.ts
+++ b/src/helper/cookie/index.ts
@@ -16,11 +16,11 @@ export const getCookie: GetCookie = (c, key?) => {
   const cookie = c.req.raw.headers.get('Cookie')
   if (typeof key === 'string') {
     if (!cookie) return undefined
-    const obj = parse(cookie)
+    const obj = parse(cookie, key)
     return obj[key]
   }
   if (!cookie) return {}
-  const obj = parse(cookie)
+  const obj = parse(cookie, key)
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   return obj as any
 }
--- a/src/helper/cookie/index.ts
+++ b/src/helper/cookie/index.ts
@@ -20,7 +20,7 @@ export const getCookie: GetCookie = (c, key?) => {
     return obj[key]
   }
   if (!cookie) return {}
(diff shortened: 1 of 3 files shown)


src/helper/cookie/index.ts
```diff
@@ -16,11 +16,11 @@ export const getCookie: GetCookie = (c, key?) => {
   const cookie = c.req.raw.headers.get('Cookie')
   if (typeof key === 'string') {
     if (!cookie) return undefined
-    const obj = parse(cookie)
+    const obj = parse(cookie, key)
     return obj[key]
   }
   if (!cookie) return {}
-  const obj = parse(cookie)
+  const obj = parse(cookie, key)
```
yusukebe, line 23: This change is not necessary.

yusukebe (changes requested):

yusukebe:
Hi @uaichat 

Thank you for the PR.

I've left a comment and there are two things what you want to do:

1. Add a test.

Please add the following test in `src/helper/cookie/index.ts`:

```ts
    describe('get null if the value is undefined', () => {
      const app = new Hono()

      app.get('/cookie', (c) => {
        const yummyCookie = getCookie(c, 'yummy_cookie')
        const res = new Response('Good cookie')
        if (yummyCookie) res.headers.set('Yummy-Cookie', yummyCookie)
        return res
      })

      it('Should be null', async () => {
        const req = new Request('http://localhost/cookie')
        const cookieString = 'yummy_cookie='
        req.headers.set('Cookie', cookieString)
        expect(res.headers.get('Yummy-Cookie')).toBe(null)
      })
    })
```

2. `denoify`

Please run `denoify` on the root directory and git add&commit&push it.

CyberFlameGO:
@yusukebe is the option for maintainer edits enabled on their branch? Perhaps you could add said test and merge it instead of waiting if so as the PR seems stale :)

yusukebe:
@CyberFlameGO 

Ah, yes. It's enabled. I'll fix and merge this later!

yusukebe (approved):

Issue #1184, opened 2023-06-17
Is there a bodyParser.urlencoded({ extended: true } equivalent in Hono?


charnould:
Hello,
I'm making my move from `Express`/`node.js` to `Hono`/`Bun`, and it feels refreshing!
Some stuff are really nice and useful: for example: `set()`/`get()`!
But sometimes, I'm struggling...

Consider the simple `c.req.parseBody()` below:

```
{
  id: "123456"
  "timeslots[0][start]": "",
  "timeslots[0][end]": "",
  "timeslots[1][start]": "",
  "timeslots[1][end]": "",
  "timeslots[2][start]": "",
  "timeslots[2][end]": "",
}
```
With `express`, it is automatically parsed using `bodyParser.urlencoded({ extended: true }`, and `timeslots` is returned as an array.

Maybe I'm missing something, but is there something equivalent in `hono`? Thanks.

yusukebe:
Hi @charnould!

Thank you for using Hono.

> Maybe I'm missing something, but is there something equivalent in `hono`?

Hono doesn't automatically parse them as an array like Express. Actually, `c.req.parseBody()` is just a shortcut to manipulate `req.formData()`, which returns a FormData object:

[FormData - Web APIs | MDN](https://developer.mozilla.org/en-US/docs/Web/API/FormData)

You can use the FormData object directly like this:

```ts
const formData = await c.req.formData()
const timeslots= formData.getAll('timeslots'))
```

I'm not sure if this will give you the result you expect, so please give it a try and let me know if it works or not.

charnould:
@yusukebe 

Thanks for your help. Unfortunately, it doesn't :-)
I've posted a question on [Stack Overflow](https://stackoverflow.com/questions/76516804/how-to-transform-formdata-into-an-iterable-array), but current answer looks super complicated for something that was a `bodyParser.urlencoded({ extended: true }`-away with `express`.

Don't you think it could be a `hono` feature?

charnould:
@yusukebe 
I saw your smiley reaction (👀) but don't know how to interpret it 😅...
Is my suggestion a bad idea? Or maybe SO answer looks super-simple to you?
Let me know (if you want, I can close this comment).

yusukebe:
Sorry.

Current my thought is it's not within Hono's scope. If you want to do that, you can make the shortcut method by yourself.

But, I was thinking it's a good idea to make an "Extended Body Parser Middleware". So, it's no problem to keep this issue still opened.

[...]

PR #1389, opened 2023-08-30
feat: body parser helper

krsbx:

- [x] Add tests
- [x] Run tests

## What this PR do?
Adding `BodyParser.parse` helper just like the one we have on `qs` so we could have some sort of (`bodyParser.urlencoded({ extended: true })`)[#1184] in `hono`

### DIFF
--- a/deno_dist/helper.ts
+++ b/deno_dist/helper.ts
@@ -2,3 +2,4 @@
 export * from './helper/cookie/index.ts'
 export * from './helper/html/index.ts'
 export * from './helper/adapter/index.ts'
+export * from './helper/body-parser/index.ts'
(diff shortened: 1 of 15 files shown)


yusukebe:
I can't merge it immediately because it's not needed in the Hono core package, because you can use the `qs` instead.

rafaell-lycan:
@krsbx perhaps this would make sense being a [3rd-party middleware](https://github.com/honojs/middleware).

@yusukebe WDYT?

krsbx:
@rafaell-lycan actually my idea is that the user can just import this helper and parse the user request body

I think it's much better if we just add some sort a params in the `req.json` so the user can parse the request body with the helper or not

PR #1393, opened 2023-08-30
fix(validator): cache `arrayBuffer` to use after validation

### LINKED ISSUE #1387: (bug): Unable to read raw request body after hono 3.5.0

3.5.6

Cloudflare Workers

1. I've made a minimal reproduction repository, please clone it from here - https://github.com/AdiRishi/hono-body-read-bug
2. Run `pnpm install`
3. Run `pnpm dev`. This will start a http server at port `8787`
4. Make a curl POST request to `http://localhost:8787/read-body`. Here is a sample request
```Shell
curl --location 'http://localhost:8787/read-body' \
--header 'Content-Type: application/json' \
--data '{}'
```
For ease, here is the code snippet that the `/read-body` endpoint runs
```typescript
app.post('/read-body', zValidator('json', z.object({})), async (c) => {
  const rawBody = await c.req.text(); // the same happens with c.req.raw.text()
  return c.json({ rawText: rawBody });
})
```
5. On the console, you will see the endpoint return a 500 error response rather than 200, triggering the bug.

The expected behaviour is that this request returns a successful 200 response and succeeds.

**NOTE** This piece of code worked in hono `3.4.3`. From what I can tell this behaviour changed in `3.5.0` and still persists in the latest version. To see a working version of this, I've made a [branch with Hono 3.4.3](https://github.com/AdiRishi/hono-body-read-bug/tree/hono%403.4.3) in the reproduction repository. You can see the exact same code working just fine.

### What do you see instead?

```Shell
TypeError: Body has already been used. It can only be used once. Use tee() first if you need to read it twice.
    at HonoRequest.cachedBody (index.js:596:39)
    at HonoRequest.text (index.js:654:17)
    at index.js:5348:31
    at dispatch (index.js:512:17)
    at index.js:513:33
    at index.js:1684:11
    at async index.js:1591:7
    at async index.js:938:62
    at async jsonError (index.js:5370:12) {
  stack: TypeError: Body has already been used. It can only…:938:62
    at async jsonError (index.js:5370:12),
  message: Body has already been used. It can only be used once. Use tee() first if you need to read it twice.
}
```

### Additional information

Now, this may not be considered a "bug", since I'm sure at some point before we handle the request Hono's zodValidator reads the request body to parse it. However, this leads to it being impossible for us to read the raw request body again in the endpoint definition.

I have a specific use case my code was running before, which broke in `3.5.0` onwards. Specifically, my project had an endpoint that would read the contents of `c.req.raw.text()` and run it through HMAC hashing, checking the result against a signature header. e.g `x-razorpay-signature`. This is a common pattern needed to validate requests by payment providers, which often provide an integrity header to check the body against.

In the latest version of Hono, there seems to be no way of doing this due to the exception I've shown above.

yusukebe:
This PR enhances the handling of request bodies post-validation. After validating a JSON with the validator, you can now invoke `c.req.text()` to access the request body. Additionally, `c.req.text()` is now available `body` validation. We achieve this by caching the `arrayBuffer` in `c.req.bodyCache` and reusing it when calling `c.req.text()`.

Note, if you attempt to use `c.req.raw.text()` after validating the `json` or `body`, you will encounter an error. Instead, always use `c.req.text()` in this context.

Fix #1387

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/request.ts
+++ b/deno_dist/request.ts
@@ -16,8 +16,14 @@ import { parse } from './utils/cookie.ts'
 import type { UnionToIntersection } from './utils/types.ts'
 import { getQueryParam, getQueryParams, decodeURIComponent_ } from './utils/url.ts'
 
-type BodyType = 'json' | 'text' | 'arrayBuffer' | 'blob' | 'formData'
-type BodyCache = Partial<Record<BodyType, any>>
+type Body = {
+  json: any
+  text: string
+  arrayBuffer: ArrayBuffer
+  blob: Blob
+  formData: FormData
+}
+type BodyCache = Partial<Body & { parsedBody: BodyData }>
 
 export class HonoRequest<P extends string = '/', I extends Input['out'] = {}> {
   raw: Request
@@ -25,7 +31,7 @@ export class HonoRequest<P extends string = '/', I extends Input['out'] = {}> {
   private paramData: Record<string, string> | undefined
   private vData: { [K in keyof ValidationTargets]?: {} } // Short name of validatedData
   path: string
-  private bodyCache: BodyCache = {}
+  bodyCache: BodyCache = {}
 
   constructor(
     request: Request,
@@ -121,13 +127,25 @@ export class HonoRequest<P extends string = '/', I extends Input['out'] = {}> {
   }
 
   async parseBody<T extends BodyData = BodyData>(): Promise<T> {
-    return await parseBody(this)
+    if (this.bodyCache.parsedBody) return this.bodyCache.parsedBody as T
+    const parsedBody = await parseBody<T>(this)
+    this.bodyCache.parsedBody = parsedBody
+    return parsedBody
   }
 
-  private cachedBody = (key: keyof BodyCache) => {
+  private cachedBody = (key: keyof Body) => {
     const { bodyCache, raw } = this
     const cachedBody = bodyCache[key]
     if (cachedBody) return cachedBody
+    /**
+     * If an arrayBuffer cache is exist,
+     * use it for creating a text, json, and others.
+     */
+    if (bodyCache.arrayBuffer) {
+      return (async () => {
+        return await new Response(bodyCache.arrayBuffer)[key]()
+      })()
+    }
     return (bodyCache[key] = raw[key]())
   }
 
--- a/deno_dist/validator/validator.ts
+++ b/deno_dist/validator/validator.ts
@@ -1,6 +1,7 @@
 import type { Context } from '../context.ts'
 import { getCookie } from '../helper/cookie/index.ts'
 import type { Env, ValidationTargets, MiddlewareHandler } from '../types.ts'
+import type { BodyData } from '../utils/body.ts'
 
 type ValidationTargetKeysWithBody = 'form' | 'json'
 type ValidationTargetByMethod<M> = M extends 'get' | 'head' // GET and HEAD request must not have a body content.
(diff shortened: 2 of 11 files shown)


deno_dist/request.ts
```diff
@@ -121,13 +127,25 @@ export class HonoRequest<P extends string = '/', I extends Input['out'] = {}> {
...
     const cachedBody = bodyCache[key]
     if (cachedBody) return cachedBody
+    /**
+     * If an arrayBuffer cache is exist,
+     * use it for creating a text, json, and others.
+     */
+    if (bodyCache.arrayBuffer) {
+      return (async () => {
+        return await new Response(bodyCache.arrayBuffer)[key]()
+      })()
```
krsbx, line 147: we can just return `new Response(...` since it return a promise anyway

or we can just change the cachedBody function into an `async function` and whenever a function need to call it, we just need to await it

yusukebe: `Response` is used for generating the value of `req.text()`, `req.json()`, or others. So it should `await`. And avoiding `async` is for perfomance.

deno_dist/validator/validator.ts
```diff
@@ -61,7 +69,20 @@ export const validator = <
...
+            const formData = arrayBufferToFormData(arrayBuffer, contentType)
+            const form: BodyData = {}
+            formData.forEach((value, key) => {
+              form[key] = value
+            })
+            value = form
+            c.req.bodyCache.formData = formData
+            c.req.bodyCache.arrayBuffer = arrayBuffer
+          }
+        })()
```
krsbx, line 85: I think we dont need to create a new `async function` in here, we can just change the code into something like this

```ts
   case 'form': {
      value = await c.req.parseBody();
      const contentType = c.req.header('Content-Type');

      // Just skip if no content type

      if (!contentType) break;
      const arrayBuffer =
        c.req.bodyCache.arrayBuffer ?? (await c.req.raw.arrayBuffer());
      const formData = arrayBufferToFormData(arrayBuffer, contentType);
      const form: BodyData = {};
      formData.forEach((value, key) => {
        form[key] = value;
      });
      value = form;
      c.req.bodyCache.formData = formData;
      c.req.bodyCache.arrayBuffer = arrayBuffer;

      break;
    }
```

src/request.ts
```diff
@@ -121,13 +127,25 @@ export class HonoRequest<P extends string = '/', I extends Input['out'] = {}> {
...
     if (cachedBody) return cachedBody
+    /**
+     * If an arrayBuffer cache is exist,
+     * use it for creating a text, json, and others.
+     */
+    if (bodyCache.arrayBuffer) {
+      return (async () => {
+        return await new Response(bodyCache.arrayBuffer)[key]()
+      })()
+    }
```
krsbx, line 148: we can just return `new Response(...` since it return a promise anyway

or we can just change the cachedBody function into an `async function` and whenever a function need to call it, we just need to await it

src/validator/validator.ts
```diff
@@ -61,7 +69,20 @@ export const validator = <
...
+            const formData = arrayBufferToFormData(arrayBuffer, contentType)
+            const form: BodyData = {}
+            formData.forEach((value, key) => {
+              form[key] = value
+            })
+            value = form
+            c.req.bodyCache.formData = formData
+            c.req.bodyCache.arrayBuffer = arrayBuffer
+          }
+        })()
```
krsbx, line 85: I think we dont need to create a new `async function` in here, we can just change the code into something like this

```ts
   case 'form': {
      value = await c.req.parseBody();
      const contentType = c.req.header('Content-Type');

      // Just skip if no content type

      if (!contentType) break;
      const arrayBuffer =
        c.req.bodyCache.arrayBuffer ?? (await c.req.raw.arrayBuffer());
      const formData = arrayBufferToFormData(arrayBuffer, contentType);
      const form: BodyData = {};
      formData.forEach((value, key) => {
        form[key] = value;
      });
      value = form;
      c.req.bodyCache.formData = formData;
      c.req.bodyCache.arrayBuffer = arrayBuffer;

      break;
    }
```

yusukebe: Okay, fixed it. Thanks!

PR #1406, opened 2023-09-04
feat(context): introduce `c.var`

yusukebe:
This PR introduces two new features: the ability to set variables using `Middleware` generics and accessing them via `c.var`.

Currently, you can globally define variable types using generics for `Hono`.

```ts
type Client = 'foo client'

const app = new Hono<{
  Variables: {
    client: Client
  }
}>()
```

Alternatively, you can specify them for a specific middleware using `ContextVariableMap`:

```ts
// JWT middleare
declare module '../../context' { // Or you can write `declare module 'hono'`
    jwtPayload: any
  }
}
```

The challenge here is that using `ContextVariableMap` makes it globally available. For instance, after importing the JWT middleware, you can access `jwtPayload` in any handler.

With this PR, you can confine middleware-defined variables to specific handlers where the middleware is applied.

```ts
const oneApiMiddleware: MiddlewareHandler<{
  Variables: {
    client: Client
  }
}> = async (c, next) => {
  c.set('client', new Client())
  await next()
}

app.get('/foo', oneApiMiddleware, (c) => {
  const client = c.get('client') // `client` is accessible only in this handler
  // ...
  return c.json({})
})
```

This offers a scoped approach to variable type availability.

The second feature is `c.var`. While the conventional method to access the actual value of a variable is `c.get()`:

```ts
const client = c.get('client')
```

With `c.var`, you can use a more intuitive syntax:

```ts
const result = c.var.client.oneMethod()
```

The combination of these two features will provide a powerful tool for developers:

```ts
const app = new Hono()

const echoMiddleware: MiddlewareHandler<{
  Variables: {
    echo: (str: string) => string
  }
}> = async (c, next) => {
  c.set('echo', (str) => str)
  await next()
}

app.get('/echo', echoMiddleware, (c) => {
  return c.text(c.var.echo('Hello!'))
})
```

Screenshot 2023-09-04 at 22 54 18

This was discussed in issue #414 and is now being realized.

### DIFF
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -195,6 +195,10 @@ export class Context<
     return this._map ? this._map[key] : undefined
   }
 
+  get var(): E['Variables'] {
+    return this._map
+  }
+
   newResponse: NewResponse = (
     data: Data | null,
     arg?: StatusCode | ResponseInit,
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -87,12 +87,12 @@ export class Context<
 > {
   req: HonoRequest<P, I['out']>
   env: E['Bindings'] = {}
+  var: E['Variables'] = {}
   finalized: boolean = false
   error: Error | undefined = undefined
 
   private _status: StatusCode = 200
   private _exCtx: FetchEventLike | ExecutionContext | undefined // _executionCtx
-  private _map: Record<string, unknown> | undefined
   private _h: Headers | undefined = undefined //  _headers
   private _pH: Record<string, string> | undefined = undefined // _preparedHeaders
   private _res: Response | undefined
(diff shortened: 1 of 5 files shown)


yusukebe:
Hi @usualoma , @ThatOneBro , and others.

What do you think about this feature? I think this is good because it will achieve big things with minimal change of the APIs.

ThatOneBro:
I think this makes a lot of sense overall. 👍 LGTM

yusukebe:
With [this commit](https://github.com/honojs/hono/pull/1406/commits/c3358672340197a5e8c2452f1764b5937e514735), I've set the properties to read-only. This is to prevent confusion, emphasizing that users should exclusively use `c.set()` to assign values.

Screenshot 2023-09-06 at 11 15 47

PR #1407, opened 2023-09-04
chore(ci): Avoid denoify API rate limit

### LINKED ISSUE #1398: Avoid `denoify` API rate limit

If call `denoify` many times in CI at once, it will throw ther error:

> API rate limit exceeded for 13.88.99.209. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more **details.**

https://github.com/honojs/hono/actions/runs/6062770394/job/16449345495#step:6:18

We can fix it by setting a GitHub token.

watany-dev:
For now, we'll attempt to mitigate the issue using a GitHub Token.

- [ ] Add tests
- [ ] Run tests

close #1398

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,8 @@ jobs:
   denoify:
     name: "Checking if you've done denoify"
     runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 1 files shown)


yusukebe:
Thanks, hope this works well. Let's go!

PR #1416, opened 2023-09-06
vercel pass RequestContext

lawrencecchen:
waitUntil doesn't work with hono/vercel

https://vercel.com/docs/functions/edge-functions/vercel-edge-package#requestcontext

Not sure about name/types. Do we want to add a dependency to @vercel/edge?


yusukebe:
Moved into #1421

PR #1420, opened 2023-09-07
feat(jsx): export `FC`

### LINKED ISSUE #1419: JSX usage example not working

3.5.8

local

1. Create a project in a file with JSX enabled according to the [documentation](https://hono.dev/guides/jsx).
2. Copy the [example code](https://hono.dev/guides/jsx#usage) from the JSX documentation and paste it in index.tsx. 

Result: the typescript compiler indicates that there is a type error for the `<Layout>` component because Layout only accepts children as a `string`  and not a `string[]`.

For the documented example to have correct typing.

### What do you see instead?

The opposite.

### Additional information

I think the solution isn't to type Layout with `string[]`. Instead, I think the fundamental problem should be solved: JSX Elements should be able to take either one or multiple child nodes, as if they had `<Fragment>` baked in. That way I could create a functional component that can either return a node or returns its children directly. There shouldn't be any distinction.

Consider:

```typescript
const Example = ({ children, value }: { children: HtmlEscapedString; value: boolean }) => {
  if (value) return children;
  return <div>{children}</div>;
};

const Test1 = (
  <Example value={true}>
    <h1>Foo</h1>
  </Example>
);

const Test2 = (
  <Example value={true}>
    <h2>Foo</h2>
    <h3>Bar</h3>
  </Example>
);
```

In this example, Test2 will throw a type error. However, if changing the component children type to `HtmlEscapedString[]`, then both Tests will throw an error.

yusukebe:
This PR makes JSX supports exporting `FC`. You can use it for specifying types to the function component.

```tsx
import { FC } from 'hono/jsx'

const Layout: FC<{ title: string }> = (props) => {
  return (
    <html>
      <head>
        <title>{props.title}</title>
      </head>
      <body>{props.children}</body>
    </html>
  )
}

const Top = (
  <Layout title='Home page'>
    <h1>Hono</h1>
    <p>Hono is great</p>
  </Layout>
)
```

Fix #1419

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/jsx/index.test.tsx
+++ b/deno_dist/jsx/index.test.tsx
@@ -1,6 +1,7 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { html } from '../helper/html/index.ts'
 import { Hono } from '../hono.ts'
+import type { FC } from './index.ts'
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 import { jsx, memo, Fragment } from './index.ts'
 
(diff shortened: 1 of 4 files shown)


huv1k (approved):
Looks good 🙌

huv1k:
Can we maybe export `Child` from `hono/jsx` too? 🤔

msonnemans:
Great improvement!

PR #1428, opened 2023-09-08
cookie parsing improvements

jkeys089:
- [x] Add tests
- [x] Run tests

### Improvements
- validate cookie names and values during parsing per rfc6265 to mitigate various security threats
- protect signed cookies against timing attacks (i.e. utilize `crypto.subtle.verify` instead of comparing strings)
- fix bug where unsigned cookie values that match the shape of signed values are not returned
- remove constraints on signed cookie values (e.g. signed values can contain period now)
- allow use of binary secret for signed cookies
- improved efficiency (e.g. remove duplicated logic / cache secret key when verifying multiple signed values)

### DIFF
--- a/deno_dist/helper/cookie/index.ts
+++ b/deno_dist/helper/cookie/index.ts
@@ -8,7 +8,7 @@ interface GetCookie {
 }
 
 interface GetSignedCookie {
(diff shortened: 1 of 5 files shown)


yusukebe:
Cool! Looks good. Thanks @jkeys089 !

Issue #914, opened 2023-02-21
[Feature Suggestion] `c.stream` Response helper


geelen:
Working on CF Workers and occasionally I find myself wanting to stream chunks of text down to cURL. There are a couple of steps to make it work, but I've wrapped it up in a little util that might make sense for inclusion into Hono?

```ts
app.get("/", async (c) => {
  return c.stream(async (stream) => {
    for (let i = 0; i < 3; i++) {
      stream.writeln(i).flush();
      await new Promise((res) => setTimeout(res, 1000));
    }
  });
});
```

&nbsp;

https://user-images.githubusercontent.com/23264/220233752-98d2abe5-5f9a-4813-80ac-ec45fc23d933.mov

&nbsp;

^ That's an example worker without Hono or Typescript. Source code, if you want to copy/paste it into the CF Workers editor:

<details>

```js
class StreamingApi {
  constructor(writable) {
    this.bytesWritten = 0;
    this.writer = writable.getWriter();
    this.encoder = new TextEncoder();
  }

  write(str) {
    const bytes = this.encoder.encode(str);
    this.bytesWritten += bytes.length;
    this.writer.write(bytes);
    return this;
  }

  writeln(str) {
    return this.write(str + "\n");
  }

  flush() {
    const bytes_needed = 4096 - (this.bytesWritten % 4096);
    const BACKSPACE = 8;
    this.writer.write(new Uint8Array(bytes_needed).fill(BACKSPACE));
    this.bytesWritten = 0;
  }

  close() {
    this.writer.close();
  }
}

const textStream = (cb) => {
  const { readable, writable } = new TransformStream();
  const stream = new StreamingApi(writable);
  cb(stream).finally(() => stream.close());
  return new Response(readable, {
    headers: {
      "content-type": "text/plain",
      "Transfer-Encoding": "chunked",
    },
  });
};

export default {
  async fetch(request, env) {
    return textStream(async (stream) => {
      for (let i = 0; i < 3; i++) {
        console.log(i);
        stream.writeln(i).flush();
        await new Promise((res) => setTimeout(res, 1000));
      }
    });
  }
}
```

</details>

### How does it work?

To do this kind of chunked-streaming from CF Workers, you need a couple of things:

* To create a `ReadableStream` and `WritableStream` using `new TransformStream()`
* Return a `Response` with the `readable` as the body, and `Transfer-Encoding: chunked` header.
* Use `writable.write()` to push content onto the stream
[...]

PR #1437, opened 2023-09-11
feat: Add `c.stream()`

sor4chi:
Hi, @yusukebe, @ geelen

I implemented `c.stream()` according to #914. I'd like to see @geelen added as a co-author as well, since I'm pretty much quoting the code from the issue.

This is the first time I've use a StreamAPI, so feel free to point out any mistakes.

@yusukebe asked me to give some examples of using `c.stream()`.
Below are some examples.

### 1: ChatGPT Proxy
Enable rate limiting to protect or hide external APIs

<details>
<summary>Code</summary>

```ts
import OpenAI from 'openai'
import { Hono } from 'hono'

const app = new Hono<{
  Bindings: {
    OPENAI_API_KEY: string
  }
}>()

const PROMPT = (message: string) => [
  {
    role: 'system' as const,
    content:
      'You are an Web developer. If you receive a question about Web-related technologies, you can answer it.',
  },
  {
    role: 'user' as const,
    content: message,
  },
]

app.post('/', async (c) => {
  const body = await c.req.json()
  const openai = new OpenAI({ apiKey: c.env.OPENAI_API_KEY })

  return c.stream(async (stream) => {
    const chatStream = await openai.chat.completions.create({
      messages: PROMPT(body.message),
      model: 'gpt-3.5-turbo',
      stream: true,
    })

    for await (const message of chatStream) {
      stream.write(message.choices[0]?.delta.content || '').flush()
    }
  })
})

export default app
```

</details>

### 2: Read some large files using stream

In this example, I'm reading a large file from the disk and sending it to the client using stream.
eg: [Read Geojson using stream and gradually draw](
    https://gunmagisgeek.com/blog/javascript/6054
)

<details>
<summary>Code</summary>

```ts
import * as fs from 'fs'
import { Hono } from 'hono'

const app = new Hono()
app.get('/assets/:key', async (c) => {
  const key = c.req.param('key')
  const path = `./assets/${key}`
  const stream = fs.createReadStream(path)
  return c.stream(async (s) => {
      s.write(chunk).flush()
    }
  })
})

serve(app)
```

</detail>

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -336,6 +336,24 @@ export class Context<
     return this.newResponse(null, status)
   }
 
+  textStream = (
+    cb: (stream: StreamingApi) => Promise<void>,
+    arg?: StatusCode | ResponseInit,
+    headers?: HeaderRecord
+  ): Response => {
+    const { readable, writable } = new TransformStream()
+    const stream = new StreamingApi(writable)
+    cb(stream).finally(() => stream.close())
+
+    this._pH ??= {}
+    this._pH['content-type'] = 'text/plain'
+    this._pH['transfer-encoding'] = 'chunked'
+
+    return typeof arg === 'number'
+      ? this.newResponse(readable, arg, headers)
+      : this.newResponse(readable, arg)
+  }
+
   stream = (
     cb: (stream: StreamingApi) => Promise<void>,
     arg?: StatusCode | ResponseInit,
@@ -346,7 +364,6 @@ export class Context<
     cb(stream).finally(() => stream.close())
 
     this._pH ??= {}
-    this._pH['content-type'] = 'text/plain; charset=UTF-8'
     this._pH['transfer-encoding'] = 'chunked'
 
     return typeof arg === 'number'
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -4,6 +4,7 @@ import type { Env, NotFoundHandler, Input, TypedResponse } from './types.ts'
 import type { CookieOptions } from './utils/cookie.ts'
 import { serialize } from './utils/cookie.ts'
 import type { StatusCode } from './utils/http-status.ts'
+import { StreamingApi } from './utils/stream.ts'
 import type { JSONValue, InterfaceToType } from './utils/types.ts'
 
 type Runtime = 'node' | 'deno' | 'bun' | 'workerd' | 'fastly' | 'edge-light' | 'lagon' | 'other'
@@ -333,6 +334,24 @@ export class Context<
     this._h ??= new Headers()
     this._h.set('Location', location)
     return this.newResponse(null, status)
+  }
+
+  stream = (
+    cb: (stream: StreamingApi) => Promise<void>,
+    arg?: StatusCode | ResponseInit,
+    headers?: HeaderRecord
+  ): Response => {
+    const { readable, writable } = new TransformStream()
+    const stream = new StreamingApi(writable)
+    cb(stream).finally(() => stream.close())
+
+    this._pH ??= {}
+    this._pH['content-type'] = 'text/plain; charset=UTF-8'
+    this._pH['transfer-encoding'] = 'chunked'
+
+    return typeof arg === 'number'
+      ? this.newResponse(readable, arg, headers)
+      : this.newResponse(readable, arg)
   }
 
   /** @deprecated
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -336,22 +336,16 @@ export class Context<
     return this.newResponse(null, status)
   }
 
-  textStream = (
+  streamText = (
     cb: (stream: StreamingApi) => Promise<void>,
     arg?: StatusCode | ResponseInit,
     headers?: HeaderRecord
   ): Response => {
-    const { readable, writable } = new TransformStream()
-    const stream = new StreamingApi(writable)
-    cb(stream).finally(() => stream.close())
+    headers ??= {}
+    headers['content-type'] = 'text/plain; charset=UTF-8'
+    headers['x-content-type-options'] = 'nosniff'
 
-    this._pH ??= {}
-    this._pH['content-type'] = 'text/plain'
-    this._pH['transfer-encoding'] = 'chunked'
-
-    return typeof arg === 'number'
-      ? this.newResponse(readable, arg, headers)
-      : this.newResponse(readable, arg)
+    return this.stream(cb, arg, headers)
   }
 
   stream = (
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -344,6 +344,7 @@ export class Context<
     headers ??= {}
     headers['content-type'] = 'text/plain; charset=UTF-8'
     headers['x-content-type-options'] = 'nosniff'
+    headers['transfer-encoding'] = 'chunked'
 
     return this.stream(cb, arg, headers)
   }
@@ -358,7 +359,6 @@ export class Context<
     cb(stream).finally(() => stream.close())
 
     this._pH ??= {}
-    this._pH['transfer-encoding'] = 'chunked'
 
     return typeof arg === 'number'
       ? this.newResponse(readable, arg, headers)
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -358,8 +358,6 @@ export class Context<
     const stream = new StreamingApi(writable)
     cb(stream).finally(() => stream.close())
 
-    this._pH ??= {}
-
     return typeof arg === 'number'
       ? this.newResponse(readable, arg, headers)
       : this.newResponse(readable, arg)
--- a/deno_dist/utils/stream.ts
+++ b/deno_dist/utils/stream.ts
@@ -1,0 +1,35 @@
+export class StreamingApi {
+  private writer: WritableStreamDefaultWriter<Uint8Array>
+  private encoder: TextEncoder
+  buffer: string[] = []
+
+  constructor(writable: WritableStream<Uint8Array>) {
+    this.writer = writable.getWriter()
+    this.encoder = new TextEncoder()
+  }
+
+  write(str: string): StreamingApi {
+    this.buffer.push(str)
+    return this
+  }
+
+  writeln(str: string): StreamingApi {
+    return this.write(str + '\n')
+  }
+
+  clear(): void {
+    this.buffer = []
+  }
+
+  async flush(): Promise<void> {
+    const text = this.buffer.join('')
+    const encoded = this.encoder.encode(text)
+    await this.writer.write(encoded)
+    this.clear()
+  }
+
+  async close(): Promise<void> {
+    await this.writer.close()
+    this.clear()
+  }
+}
--- a/deno_dist/utils/stream.ts
+++ b/deno_dist/utils/stream.ts
@@ -9,7 +9,7 @@ export class StreamingApi {
     this.encoder = new TextEncoder()
   }
 
-  async write(input?: Uint8Array | string) {
+  async write(input: Uint8Array | string) {
     if (typeof input === 'string') {
       input = this.encoder.encode(input)
     }
@@ -17,8 +17,8 @@ export class StreamingApi {
     return this
   }
 
-  async writeln(input?: string) {
-    await this.write((input || '') + '\n')
+  async writeln(input: string) {
+    await this.write(input + '\n')
     return this
   }
 
--- a/deno_dist/utils/stream.ts
+++ b/deno_dist/utils/stream.ts
@@ -27,6 +27,10 @@ export class StreamingApi {
     return this
   }
 
+  wait(ms: number) {
+    return new Promise((res) => setTimeout(res, ms))
+  }
+
   async close() {
     await this.writer.close()
   }
--- a/deno_dist/utils/stream.ts
+++ b/deno_dist/utils/stream.ts
@@ -1,35 +1,39 @@
 export class StreamingApi {
   private writer: WritableStreamDefaultWriter<Uint8Array>
   private encoder: TextEncoder
-  buffer: string[] = []
+  private writable: WritableStream
 
-  constructor(writable: WritableStream<Uint8Array>) {
+  constructor(writable: WritableStream) {
+    this.writable = writable
     this.writer = writable.getWriter()
     this.encoder = new TextEncoder()
   }
 
-  write(str: string): StreamingApi {
-    this.buffer.push(str)
+  async write(input?: Uint8Array | string) {
+    if (typeof input === 'string') {
+      input = this.encoder.encode(input)
+    }
+    await this.writer.write(input)
     return this
   }
 
-  writeln(str: string): StreamingApi {
-    return this.write(str + '\n')
+  async writeln(input?: string) {
+    await this.write((input || '') + '\n')
+    return this
   }
 
-  clear(): void {
-    this.buffer = []
+  async log(arg: any) {
+    await this.writeln(JSON.stringify(arg))
+    return this
   }
 
-  async flush(): Promise<void> {
-    const text = this.buffer.join('')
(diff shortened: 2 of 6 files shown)


geelen:
Hey @sor4chi thanks for taking this up! I'd added a few more convenience methods to my own `StreamingApi` implementation, which you might be interested in including:

```ts
// Lets you easily emit data in JSON lines format
log(obj: any) {
  return this.writeln(JSON.stringify(obj)).flush();
}

// Lets you send a ReadableStream directly, e.g. from a subrequest
  this.writer.releaseLock();
  await body.pipeTo(this.writable, { preventClose: true });
  this.writer = this.writable.getWriter();
}
```

My only thought is whether this should be called `c.textStream` or something, since everything is based around strings so far? But I'll leave that up to @yusukebe to decide. Maybe this implementation can be made generic to allow both strings and binary data? Afaik, it's just the `TextEncoder` and the `"content-type": "text/plain"` header that makes it string-y. Thoughts?

src/utils/stream.ts
```diff
@@ -0,0 +1,35 @@
...
+  private encoder: TextEncoder
+  buffer: string[] = []
+
+  constructor(writable: WritableStream<Uint8Array>) {
+    this.writer = writable.getWriter()
+    this.encoder = new TextEncoder()
+  }
+
+  write(str: string): StreamingApi {
+    this.buffer.push(str)
```
geelen, line 12: Does it make sense to buffer here into memory or just write directly to the stream? If the latter, I don't think you'd even need `.flush()` at all?

sor4chi: Hi @geelen, thanks for your review.
I found a `flush` function in @geelen 's reference implementation, so I implemented this to create a buffer until then so that the stream is written at the timing when the `flush` actually occurs.
Surely, I think it is unnecessary because the user can buffer the stream before writing to it.
Would it be OK to send it when `write` or `writeln` is called?

yusukebe:
Hi @sor4chi!

Awesome! This will be set to be a highlight feature for our next minor update, `v3.7.0`.

I haven't looked closely the implementation yet, but my primary concerns are:

* **Naming**: As @geelen's mentioned, `c.textStream()` might be better if it supports only `"content-type": "text/plain"`.
* **Focus**: If there's potential to make this feature more general – like creating a stream that supports binary data – it would be great. In that case, we could name it `c.stream()`.

I'll be reviewing your implementation more closely later. Either way, this is an important feature, so take our time making it.

sor4chi:
Thank you for the review and comment, @geelen and @yusukebe.

I wrote the sample code to confirm the plan, but does it mean that this is what is supposed to happen?

```ts
textStream = (
    cb: (stream: WritableStreamDefaultWriter<string>) => Promise<void>,
    headers?: HeaderRecord
  ): Response => {
    const encoder = new TextEncoder()
    const { readable, writable } = new TransformStream({
      transform: (chunk, controller) => {
        controller.enqueue(encoder.encode(chunk))
      },
    })
    const writer = writable.getWriter()
    cb(writer).finally(() => writer.close())

    this._pH ??= {}
    this._pH['content-type'] = 'text/plain'
    this._pH['transfer-encoding'] = 'chunked'

    return typeof arg === 'number'
      ? this.newResponse(readable, arg, headers)
      : this.newResponse(readable, arg)
  }

  stream = (
    cb: (stream: WritableStreamDefaultWriter<Uint8Array>) => Promise<void>,
    headers?: HeaderRecord
  ): Response => {
    const { readable, writable } = new TransformStream()
    const writer = writable.getWriter()
    cb(writer).finally(() => writer.close())

    this._pH ??= {}
    this._pH['transfer-encoding'] = 'chunked'

    return typeof arg === 'number'
      ? this.newResponse(readable, arg, headers)
      : this.newResponse(readable, arg)
  }
  ```
  
  ```ts
   it('c.textStream()', async () => {
    const res = c.textStream(async (stream) => {
      for (let i = 0; i < 3; i++) {
        await stream.write(`${i}`)
        await new Promise((res) => setTimeout(res, 1000))
      }
    })
    if (!res.body) {
      throw new Error('Body is null')
    }
    const reader = res.body.getReader()
    for (let i = 0; i < 3; i++) {
      const { value } = await reader.read()
      expect(value).toEqual(new Uint8Array([i + '0'.charCodeAt(0)]))
    }
  })

    it('c.stream()', async () => {
    const res = c.stream(async (stream) => {
      for (let i = 0; i < 3; i++) {
        await stream.write(new Uint8Array([i]))
        await new Promise((res) => setTimeout(res, 1000))
      }
    })
    if (!res.body) {
      throw new Error('Body is null')
    }
    const reader = res.body.getReader()
    for (let i = 0; i < 3; i++) {
      const { value } = await reader.read()
      expect(value).toEqual(new Uint8Array([i]))
    }
  })
  ```

File: src/utils/stream.ts (hunk 1 above)
Comment (geelen) [reply src/utils/stream.ts:12]: Yeah, I was testing something at the time with Cloudflare Workers and found that it was only streaming in 4096 byte chunks, so I made flush send a bunch of bytes of whitespace to force it to send 🤫 

I think I was doing something wrong, though, since when I tried again, chunks were streamed much more rapidly so I think it's fine to send them as soon as `write` is called. A user can always buffer on their own side before sending it...

**Comment by sor4chi** [comment]
Hello @geelen, @yusukebe.
It took me a while, but I changed the interface based on your suggestion. 
How do you like it?

File: deno_dist/utils/stream.ts [hunk 2]
```diff
@@ -0,0 +1,39 @@
...
+    }
+    await this.writer.write(input)
+    return this
+  }
+
+  async writeln(input: string) {
+    await this.write(input + '\n')
+    return this
+  }
+
```
yusukebe, line 24: Should should add the following to remove a lint  error.

```
```

yusukebe:
@sor4chi

Great job! There are two points I'd like us to consider:

### `streamText()`

While `c.stream()` can handle all cases, it's not necessity for `c.textStream()`. But, it could be useful, because when we often use `text/plain` for debugging. I think naming it `c.streamText()` might be more consistent, given the prefix matches `c.stream()`.

And if you create a `c.streamText()`, you can do it without redundant:

```ts
  streamText = (
    headers?: HeaderRecord
  ): Response => {
    headers = headers ?? {}
    headers['content-type'] = 'text/plain; charset=UTF-8'
    headers['x-content-type-options'] = 'nosniff'
    return this.stream(cb, arg ?? 200, headers)
  }
```

Note: Including `x-content-type-options: nosniff` would be beneficial when accessing via a web browser.

### About `stream.wait()`

Though it is also not necessary, `stream.wait()` could be helpful:

```ts
export class StreamingApi {
  //...
  wait(ms: number) {
    return new Promise((resolve) => setTimeout(resolve, ms))
  }
}
```

### Sample

Combining all features, the code looks much cleaner:

```ts
app.get('/', (c) => {
  return c.streamText(async (stream) => {
    while (true) {
      const message = `It is ${new Date().toISOString()}`
      stream.writeln(message)
      await stream.wait(1000)
    }
  })
})
```

Cool!

sor4chi:
@yusukebe 
Thank you for pointing this out.
It certainly seems like a good idea to align stream/streamText and prefix from a completion perspective.
As for the `wait` method, it would certainly be useful.

Is it like leaving it up to the user whether to add `transfer-encoding: chunked` or not?
Also, should I put back `charset=UTF-8`?

yusukebe:
@sor4chi 

> Is it like leaving it up to the user whether to add `transfer-encoding: chunked` or not?

Yes, we have to keep it because I think it should be `chunked` for a streaming response.

> should I put back `charset=UTF-8`?

I've suggested that because of the compatibility for `c.text()`:

https://github.com/honojs/hono/blob/9cb6b37fa3484eda479f9fb5f9c16757009c841a/src/context.ts#L313

I'm planning making `'text/plain; charset=UTF-8'` as a variable to reduce the bundle size, but I'll do that later.

deno_dist/utils/stream.ts
```diff
@@ -0,0 +1,39 @@
...
+  async log(arg: any) {
```
yusukebe, line 25: I've thought it.`log()` is good, but to reduce bundle size, I think we shouldn't add this. It's a so slight change, but we're nervous about increasing bundle sizes.

OK @geelen ?

yusukebe:
Ah, we can refer to the Vercel's AI sdk: https://sdk.vercel.ai/docs

arnaudbzn:
Back-pressure and cancellation management (detailed in Vercel AI SDK [here](https://sdk.vercel.ai/docs/concepts/backpressure-and-cancellation)).

A minimalist code I'm using now with Hono and Server Sent Event message format 

```typescript
const completionStream = await openai.chat.completions.create(....)
const completionStreamIterator = completionStream[Symbol.asyncIterator]();

const eventStream = new ReadableStream({
 // Client asks for more data
 async pull(controller) {
      const { done, value } = await completionStreamIterator.next();
      
      // Nothing more from OpenAI
      if (done) {
       // Close client connection, we could send a "finish" SSE message before closing
        controller.close()
      }
      controller.enqueue(`data: ${value.choices[0]?.delta?.content || ""}\n\n`);
  },
  // Client aborts the connection
   async cancel(reason) {
      // Cancel OpenAI  upstream
      await completionStreamIterator.return?.(reason);
    }
 }).pipeThrough(new TextEncoderStream());
 
// Respond with the stream and SSE required headers
return new Response(eventStream, {
  headers: {
    ...corsHeaders,
    "Content-Type": "text/event-stream; charset=utf-8",
    "Cache-Control": "no-store",
    Connection: "keep-alive",
  },
});
  
```

arnaudbzn:
@yusukebe When testing my openai proxy (not using hono c.stream) with CloudFlare functions I get more than 100ms CPU time according to the completion length form OpenAI (so way above the 50ms CPU time limit).

What CPU time you get with the `c.stream` based solution?

arnaudbzn:
About SSE and cancellation with bun there is this issue:
Bun only does not support http2 right now and the cancellation management is a bit tricky:
https://github.com/oven-sh/bun/issues/2443
https://github.com/oven-sh/bun/pull/3390

watany-dev:
Similarly, it appears that AWS Lambda's response streaming only supports HTTP1.1 as well.

You can find more information at 
- https://docs.aws.amazon.com/lambda/latest/dg/configuration-response-streaming.html 
- https://docs.aws.amazon.com/lambda/latest/dg/runtimes-custom.html#runtimes-custom-response-streaming. 

It's a challenging question to decide how generic to make the system, but I think it would be good to have options in the roadmap.

arnaudbzn:
There is no issues with HTTP1.1, only the limit of 6 parallel SSE connections limit per domain.  The code stays the same.
The only JS runtime that requires a specific request cancellation management is bun for the moment: 
```typescript
async pull(controller) {
   // with bun we have to check if the request is cancelled here
   if (req.signal.aborted) {
          controller.close();
          // Cancel OpenAI upstream
          await completionStreamIterator.return?.(reason);
          return;
    }
    // ...
 }
  ```

**Comment by watany-dev** [comment]
I see...if HTTP 1.1 support is only for AWS lambda. There is no need to support it in this commit. There are other complicated differences, so I will absorb them with adaptor.

**Comment by yusukebe** [comment]
@arnaudbzn 

> When testing my openai proxy (not using hono c.stream) with CloudFlare functions I get more than 100ms CPU time according to the completion length form OpenAI (so way above the 50ms CPU time limit).

> What CPU time you get with the `c.stream` based solution?

I think `c.stream()` doesn't resolve the problem, but when I tried a ChatGPT streaming gateway on Cloudflare Pages, the CPU time is not so long. It works well.

**Comment by yusukebe** [comment]
Hi @arnaudbzn @watany-dev 

Thank you for your comments.

`c.stream()` is **not** dependent of HTTP version. With HTTP/1.1 on Wrangler, Node.js, Deno, and Bun, it will be `chunked` responses. In the case of Cloudflare, With HTTP/2 or H3, the `transfer-encoding` header with `chunked` set will be removed and the response will be kept as streaming. We should not care about protocol versions in Hono implementation.

About SSE. SSE is **not** supported in `c.stream()`. If you want to use SSE, we don't know if we will adopt it, but you can create a helper to make it happen.

So what I'm saying is that `c.stream()` should be a simple function to output HTTP streaming responses. If you want to do something else, you can create your own `ReadableStream`.

File: src/utils/stream.ts [hunk 4]
```diff
@@ -0,0 +1,43 @@
...
+    await this.write(input + '\n')
+    return this
+  }
+
+  async log(arg: any) {
+    await this.writeln(JSON.stringify(arg))
+    return this
+  }
+
+  wait(ms: number) {
```
yusukebe, line 30: On second thought about this, I think the name `sleep()` would be better. What do you think?

deno_dist/utils/stream.ts
yusukebe: @sor4chi 

Let's remove the `log()`!

src/context.ts
```diff
@@ -335,6 +336,35 @@ export class Context<
...
+    cb: (stream: StreamingApi) => Promise<void>,
+    arg?: StatusCode | ResponseInit,
+    headers?: HeaderRecord
+  ): Response => {
+    const { readable, writable } = new TransformStream()
+    const stream = new StreamingApi(writable)
+    cb(stream).finally(() => stream.close())
+
+    this._pH ??= {}
+    this._pH['transfer-encoding'] = 'chunked'
```
yusukebe, line 361: How about specifying `transfer-encoding: chunked` in `c.streamText()` without setting the header in `c.steam()`? This will make `c.stream()` more generic.

src/utils/stream.ts
sor4chi: That is more intuitive. Let's do that.

src/context.ts
sor4chi: I agree. Fixed!

deno_dist/context.ts
```diff
@@ -335,6 +336,35 @@ export class Context<
...
+  stream = (
+    cb: (stream: StreamingApi) => Promise<void>,
+    arg?: StatusCode | ResponseInit,
+    headers?: HeaderRecord
+  ): Response => {
+    const { readable, writable } = new TransformStream()
+    const stream = new StreamingApi(writable)
+    cb(stream).finally(() => stream.close())
+
+    this._pH ??= {}
```
yusukebe, line 361: You can remove it!

src/context.test.ts
```diff
@@ -259,4 +259,39 @@ describe('Pass a ResponseInit to respond methods', () => {
...
+      const { value } = await reader.read()
+      expect(decoder.decode(value)).toEqual(`${i}`)
+    }
+  })
+
+  it('c.stream()', async () => {
+    const res = c.stream(async (stream) => {
+      for (let i = 0; i < 3; i++) {
+        await stream.write(new Uint8Array([i]))
+        await stream.sleep(1000)
```
yusukebe, line 285: Same reason as above.

src/context.test.ts
```diff
@@ -259,4 +259,39 @@ describe('Pass a ResponseInit to respond methods', () => {
     expect(res.headers.get('content-type')).toMatch(/^text\/html/)
     expect(await res.text()).toBe('<h1>foo</h1>')
   })
+
+  it('c.streamText()', async () => {
+    const res = c.streamText(async (stream) => {
+      for (let i = 0; i < 3; i++) {
+        await stream.write(`${i}`)
+        await stream.sleep(1000)
```
yusukebe, line 267: The value here should be as small as possible. For example, `1`.

yusukebe (changes requested):
@sor4chi 

Almost done! See the comments.

sor4chi:
@yusukebe 
I'm done!

yusukebe (approved):

yusukebe:
Hey!

Now, I'll merge it into the `next` branch which will be released as a next minor version `v3.7.0`.

Awesome work, @sor4chi ! Thanks!

PR #1447, opened 2023-09-12
fix(adapter): types for env function

huv1k:
This PR creates a fix for `env` function types from `hono/adapter`. This function did not respect `env` that the `hono` server could accept and it was impossible to use `Bindings` and `Variables`. This could be made even more generic, but this would require making the server even more generic. I didn't add any tests, because there are no tests for this `env` function, so I didn't want to scaffold a new suite.

### DIFF
--- a/deno_dist/helper/adapter/index.ts
+++ b/deno_dist/helper/adapter/index.ts
@@ -1,6 +1,7 @@
 import type { Context } from '../../context.ts'
+import type { Env } from '../../types.ts'
 
-export const env = <T extends Record<string, string>, C extends Context = Context<{}>>(
+export const env = <T extends Env, C extends Context = Context<{}>>(
   c: C
 ): T & C['env'] => {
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @huv1k 

Sorry! I think it already supports Cloudflare Bindings if you set the proper types to `new Hono()` as generics.

Screenshot 2023-09-13 at 22 23 41

This feature was introduced in https://github.com/honojs/hono/pull/1021.

What do you think about it?

huv1k:
> Hi @huv1k
> 
> Sorry! I think it already supports Cloudflare Bindings if you set the proper types to `new Hono()` as generics.
> 
> Screenshot 2023-09-13 at 22 23 41
> This feature was introduced in #1021.
> 
> What do you think about it?

I was trying to set `env` as generic because I wanted to use it in middleware, but I changed `MiddlewareHandler` env and this works as you suggested. We can close it 🤷

yusukebe:
@huv1k 

Thanks!

PR #1451, opened 2023-09-13
feat: add Hono test client (RPC)

hagishi:

- [x] Add tests
- [x] Run tests

### What this PR do?
By using the hc test client, we can easily write tests. This utility enables efficient integration and unit testing by allowing manipulation of the Hono application object rather than making actual HTTP requests.

### Usage Example
```ts
it('test', async() => {
  const app = new Hono().get('/search', (c) => c.jsonT({ hello: 'world' }))
  const res = await hc(app).search.$get()

  expect(await res.json()).toEqual({ hello: 'world' })
})
```

### DIFF
--- a/deno_dist/helper/testing/index.ts
+++ b/deno_dist/helper/testing/index.ts
@@ -1,0 +1,18 @@
+import { hc as HonoClient } from '../../client/index.ts'
+import type { Hono } from '../../hono.ts'
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type ExtractEnv<T> = T extends Hono<infer E, any, any> ? E : never
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const hc = <T extends Hono<any, any, any>>(
+  app: T,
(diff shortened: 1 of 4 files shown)


yusukebe:
Hi @hagishi!

Cool! This is one of the things we want to do with RPC-mode. Introducing the `testing` helper makes sense.

But, there's one thing we should consider. The function name `hc` might be confusing because `hono/client` has a function with the same name. I haven't come up with a perfect name yet, but something like `appToClient()` might be better. Any thoughts?

hagishi:
Hi @yusukebe,

Thank you for the review.

I agree that some parts could indeed be confusing.
Following the example of frameworks like [FastAPI](https://fastapi.tiangolo.com/tutorial/testing), 
I think a clearer naming convention like testClient(app) would be better.
What do you think?"

yusukebe:
Hi @hagishi !

`testClient` is simple and good! Let's go with it.

yusukebe:
Hi @hagishi 

Thanks! Merging it now.

Issue #1452, opened 2023-09-13
Use `getRuntimeKey()` in `env` function


yusukebe:
### What is the feature you are proposing?

`c.runtime` is deprecated , so it should use `getRuntimeKey()`.

https://github.com/honojs/hono/blob/9cb6b37fa3484eda479f9fb5f9c16757009c841a/src/helper/adapter/index.ts#L9-L28

PR #1454, opened 2023-09-13
refactor(helper):  Use getRuntimeKey() in env function

watany-dev:
I've removed the logic that depends on c.runtime. I've also refactored the code to map behavior for each runtime to an object. If this is overkill, I can revert it back to the original code.

- [ ] Add tests
- [x] Run tests

fixed: https://github.com/honojs/hono/issues/1452


yusukebe:
Hi @watany-dev !

Great job with nice refactoring. Thanks!

PR #1458, opened 2023-09-14
fix(types): Use ExpectTypeOf from vitest to test types

vadhe:
before we used custom type to test types but now we choose to use ExpectTypeOf from vitest to test types

### DIFF
--- a/src/hono.test.ts
+++ b/src/hono.test.ts
@@ -1,6 +1,7 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
 /* eslint-disable @typescript-eslint/ban-ts-comment */
 import { vi } from 'vitest'
+import { expectTypeOf } from 'vitest'
 import type { Context } from './context'
 import { Hono } from './hono'
 import { HTTPException } from './http-exception'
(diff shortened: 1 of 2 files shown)


src/hono.test.ts
```diff
@@ -2001,8 +2004,8 @@ describe('Context binding variables', () => {
 
   it('Should get binding variables with correct types', async () => {
     app.get('/', (c) => {
-      type verifyID = Expect<Equal<number, typeof c.env.USER_ID>>
-      type verifyName = Expect<Equal<string, typeof c.env.USER_NAME>>
+      expectTypeOf(c.env.USER_ID).toEqualTypeOf<number>()
```
yusukebe, line 2007: It does not pass, though it seems to be correct. What do you think to write like the following instead of that as a workaround:

```ts
 expectTypeOf(c.env).toEqualTypeOf<Bindings>()
```

yusukebe (commented):
@vadhe 

Thank you! This is what I want to do. Please see the comment I pointed.

src/hono.test.ts
```diff
@@ -2001,8 +2004,7 @@ describe('Context binding variables', () => {
 
   it('Should get binding variables with correct types', async () => {
     app.get('/', (c) => {
-      type verifyID = Expect<Equal<number, typeof c.env.USER_ID>>
-      type verifyName = Expect<Equal<string, typeof c.env.USER_NAME>>
+      expectTypeOf(c.env).toEqualTypeOf<Bindings>()
```
vadhe, line 2007: updated @yusukebe

yusukebe:
Great! Let's go with it!

PR #1462, opened 2023-09-15
feat: Refactor ClientRequest Type for header validation

hagishi:
### Author should do the followings, if applicable

- [x] Add tests
- [x] Run tests
- [x] `yarn denoify` to generate files for Deno

### What this PR do?

The changes are aimed at improving the type safety when dealing with client requests that include headers. 
By enforcing the validation at the type level, we can catch potential issues earlier in the development process, reducing the risk of runtime errors.

PR #1476, opened 2023-09-17
feat(client) add cookie type validate

hagishi:

- [x] Add tests
- [x] Run tests

### What this PR do?
1. Implemented type checking for Cookies using the Hono client.
2. Modified the args arguments to also handle cookie and header.

By incorporating type checking for Cookies through the Hono client, 
we aim to improve security and data integrity. 
Additionally, by allowing args to handle both cookie and header, 
we expand our capacity to address a broader range of validation scenarios.

This branch is branched off from the fixes in #1462

### DIFF
--- a/src/helper/cookie/index.test.ts
+++ b/src/helper/cookie/index.test.ts
@@ -1,5 +1,13 @@
 import { Hono } from '../../hono'
-import { getCookie, getSignedCookie, setCookie, setSignedCookie, deleteCookie } from '.'
+import { serialize, serializeSigned } from '../../utils/cookie'
+import {
+  getCookie,
+  getSignedCookie,
+  setCookie,
+  setSignedCookie,
+  deleteCookie,
+  getCookieValue,
+} from '.'
 
 describe('Cookie Middleware', () => {
   describe('Parse cookie', () => {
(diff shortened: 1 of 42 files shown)


hagishi:
@yusukebe 
Please review the changes and provide your feedback.

src/helper/cookie/index.ts
```diff
@@ -57,3 +57,8 @@ export const setSignedCookie = async (
 export const deleteCookie = (c: Context, name: string, opt?: CookieOptions): void => {
   setCookie(c, name, '', { ...opt, maxAge: 0 })
 }
+
+export const getCookieValue = (cookie: string, key: string): string => {
```
yusukebe, line 61: Since `getCookieValue()` is only used in `src/client/client.test.ts`, you can define it in `src/client/client.test.ts`. You should not include it in the helper to add unnecessary APIs or keep it in size.

src/helper/cookie/index.test.ts
```diff
@@ -1,5 +1,13 @@
 import { Hono } from '../../hono'
-import { getCookie, getSignedCookie, setCookie, setSignedCookie, deleteCookie } from '.'
+import { serialize, serializeSigned } from '../../utils/cookie'
+import {
+  getCookie,
+  getSignedCookie,
+  setCookie,
+  setSignedCookie,
+  deleteCookie,
+  getCookieValue,
```
yusukebe, line 9: For the same reason, this function should be defined in this file.

yusukebe (changes requested):

yusukebe:
Hi @hagishi 

This is great! The Validator had support for `header` and `cookie`, but not yet for the client. Thanks.

I made some comments, please take a look.

src/helper/cookie/index.ts
hagishi: @yusukebe 
Thank you for comment.
I removed it from helper.

yusukebe (approved):
LGTM!

PR #1482, opened 2023-09-19
fix(context): fix `c.stream()` and `c.streamText()` matters

yusukebe:
This PR includes these fixes:

* Make the response headers of `c.streamText()` correct.
* Make `text/plain; charset=UTF-8` as the variable to reduce bundle size.
* Handling errors and ignoring them. If there is an error throw, the process may stop. And also, tracing an error may get in the way. So, I decided to do nothing in the catch.

Errors are as follows. Occurs when accessing while it is streaming.

Screenshot 2023-09-19 at 18 28 02

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -86,6 +86,8 @@ type ContextOptions<E extends Env> = {
   notFoundHandler?: NotFoundHandler<E>
 }
 
+const TEXT_PLAIN = 'text/plain; charset=UTF-8'
+
 export class Context<
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   E extends Env = any,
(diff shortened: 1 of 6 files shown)


yusukebe:
Hi @sor4chi !

Could you review this?

For error handling, nothing is done when an error occurs. This is because it is easier to use. If a user want to trace, the user can create own `ReadableStream`, so I think there should be no problem.

sor4chi (approved):
I agree with the change. 

It is true that `stream.close()` is called internally, so if for some reason it closes first, it will cause an error.

Thank you for fixing and adding some tests!

CyberFlameGO (approved):
Nothing seems breaking here :)

CyberFlameGO (approved):
Nothing seems breaking here - LGTM :)

yusukebe:
Thanks!! Merge now!

PR #1487, opened 2023-09-20
fix(jwt): incorrect error message

NicoPlyley:
### Description

The error message is a duplicate of line 14 `JwtTokenInvalid` and is no descriptive to the actual cause of the issue.

### Updated 

Updated the error message to align with the name of the error

- N/A: Add test
- [x] Run tests


yusukebe:
Hi @NicoPlyley 

Thanks. Looks good. I'll merge it now!

PR #1488, opened 2023-09-21
feat: include multivalue form data as an array

calebkish:
In a `multipart/form-data` or a `application/x-www-form-urlencoded` POST request, if there are multiple values for a given field name (key), `c.req.parseBody()` will not only include one of them.

This makes it so that when using `c.req.parseBody()`, if there are multiple values for a given key, they'll be included in an array.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/utils/body.ts
+++ b/deno_dist/utils/body.ts
@@ -17,7 +17,7 @@ export const parseBody = async <T extends BodyData = BodyData>(
     if (formData) {
       const form: BodyData = {}
       for (const key of formData.keys()) {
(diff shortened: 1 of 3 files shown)


yusukebe:
Hi @calebkish 

Thanks for the PR. I'd like to merge this PR, but this is a breaking change. So, we'll accept when releasing the next major version, "v4".

calebkish:
I'm going to close this in favor of https://github.com/honojs/hono/pull/1461

I think it pretty much covers the concerns of this PR.

Issue #1501, opened 2023-09-24
Support for JWT Payload to allow different types of date inputs


NicoPlyley:
### What is the feature you are proposing?

Currently in the payload you can provide, `iat`, `exp`, and `nbf`. However, the value have to be a unix epoch with no milliseconds. I am proposing support for:

- `Date.now()` - with milliseconds
- `new Date()`
- A string like `30d`

To be able to be converted to the respective time in epoch automatically. 

I will submit a draft PR as a proof on concept and if this is something we can incorporate I will write the test for it as well.

PR #1502, opened 2023-09-24
feat(jwt): add support for different date formats

NicoPlyley:
A draft PR for #1501 

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/utils/jwt/dateFormatter.ts
+++ b/src/utils/jwt/dateFormatter.ts
@@ -1,0 +1,69 @@
+const parseDuration = (input: string) => {
+  const multipliers: {[key: string]: number} = {
+    s: 1000,        // seconds
+    m: 60 * 1000,   // minutes
+    h: 60 * 60 * 1000,   // hours
+    d: 24 * 60 * 60 * 1000,   // days
+    M: 30 * 24 * 60 * 60 * 1000,   // months (approximate)
+    y: 365 * 24 * 60 * 60 * 1000,   // years (approximate)
+  }
+
+  const match = input.match(/^(\d+)([smhdMy])$/)
+
+  if (!match) {
+    throw new Error('Invalid shorthand duration format')
+  }
+
+
+  const value = parseInt(match[1])
+  const unit = match[2]
+
+  const totalMilliseconds = value * multipliers[unit]
+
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @NicoPlyley 

Thanks for this PR. Looks good for me. But I want to know the opinion of @Code-Hex .

cc: @Code-Hex 
What do you think about this? And if it's OK, could you review this?

Code-Hex:
Hi @NicoPlyley and @yusukebe 

According to RFC 7519, it is stated that [exp](https://www.rfc-editor.org/rfc/rfc7519#section-4.1.4), [nbf](https://www.rfc-editor.org/rfc/rfc7519#section-4.1.5), and [iat](https://www.rfc-editor.org/rfc/rfc7519#section-4.1.6) must always be NumericDate values.

NumericDate is the last definition in [Section 2. Terminology](https://datatracker.ietf.org/doc/html/rfc7519#section-2), and is defined as the number of seconds (not milliseconds) since Epoch:
> A JSON numeric value representing the number of seconds from 1970-01-01T00:00:00Z UTC until the specified UTC date/time, ignoring leap seconds.  This is equivalent to the IEEE Std 1003.1, 2013 Edition [[POSIX.1](https://datatracker.ietf.org/doc/html/rfc7519#ref-POSIX.1)] definition "Seconds Since the Epoch", in which each day is accounted for by exactly 86400 seconds, other than that non-integer values can be represented.  See [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339) [[RFC3339](https://datatracker.ietf.org/doc/html/rfc3339)] for details regarding date/times in general and UTC in particular.

I think it is desirable for a widely adopted library that provides a JWT package to be strict in its specifications. This is because users sometimes prioritize strictness over convenience.

yusukebe:
Thanks @Code-Hex 

It is certainly important to follow the Spec and I agree with you.

@NicoPlyley :

What do you think about this?

NicoPlyley:
I am okay with any decision you make. I got the idea from other popular packages like json-web-token and jose which do allow different types of input. But I understand if it's not in the realm.

yusukebe:
Hi @NicoPlyley 

Thank you for your understanding. We will not accept and close this PR at this time. Thank you and contribute again.

PR #1517, opened 2023-09-28
add Server Stop Method

EdamAme-x:
```javascript
const app = new Hono();

app.get("/", async (c) => {

   app.stop(5000) // Server stopped responding for 5 seconds

   return c.text("hey Hono🔥")
})
```

Also started by `app.resume()`. 
`app.resume(1000)` Starts after one second.

- [x] Add tests
- [x] Run tests


yusukebe:
Hi @EdamAme-x 

What is the use case for this feature?

In order to keep `hono-base.ts` as simple and small as possible, it is difficult to include features that have few needs.

EdamAme-x:
It is expected to be able to reduce the damage in the event of a load attack on the server.

Also, this method did not need to be included in HonoBase.

I would like to include only stopped in HonoBase and introduce the stop method as utils.

EdamAme-x:
Stop(app, -1) // Stopped until re-started.
Resume(app, 1000) // resume after 1s

Stop(app, 3600 * 1000) // 1h

yusukebe:
This feature does not seem necessary for the Hono core. So, I will close it this time. Sorry.

EdamAme-x:
Understood.

PR #1524, opened 2023-09-29
Faster (verification required).

EdamAme-x:
Guess the format of the path ahead of time and unused the routers that are not match.

The following benchmark results were obtained in my environment.

```
# Env
codespaces 4 Core / 16GB RAM

Benchmark result

# After
332,013 ops
333,485 ops
338,571 ops

# Before
303,517 ops
324,878 ops
330,072 ops
```

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/router/smart-router/router.ts
+++ b/deno_dist/router/smart-router/router.ts
@@ -24,6 +24,8 @@ export class SmartRouter<T> implements Router<T> {
       throw new Error('Fatal error')
     }
 
+    const pathType: ('RegExpRouter' | 'PatternRouter' | 'LinearRouter')[] = []
+
     const { routers, routes } = this
     const len = routers.length
     let i = 0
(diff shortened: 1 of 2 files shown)


EdamAme-x:
The original code is as follows.

```
/* eslint-disable @typescript-eslint/ban-ts-comment */
import type { Router, Result } from '../../router'
import { UnsupportedPathError } from '../../router'

export class SmartRouter<T> implements Router<T> {
  name: string = 'SmartRouter'
  routers: Router<T>[] = []
  routes?: [string, string, T][] = []

  constructor(init: Pick<SmartRouter<T>, 'routers'>) {
    Object.assign(this, init)
  }

  add(method: string, path: string, handler: T) {
    if (!this.routes) {
      throw new Error('Can not add a route since the matcher is already built.')
    }

    this.routes.push([method, path, handler])
  }

  match(method: string, path: string): Result<T> | null {
    if (!this.routes) {
      throw new Error('Fatal error')
    }

    const { routers, routes } = this
    const len = routers.length
    let i = 0
    let res
    for (; i < len; i++) {
      const router = routers[i]
      try {
        routes.forEach((args) => {
          router.add(...args)
        })
        res = router.match(method, path)
      } catch (e) {
        if (e instanceof UnsupportedPathError) {
          continue
        }
        throw e
      }

      this.match = router.match.bind(router)
      this.routers = [router]
      this.routes = undefined
      break
    }

    if (i === len) {
      // not found
      throw new Error('Fatal error')
    }

    // e.g. "SmartRouter + RegExpRouter"
    this.name = `SmartRouter + ${this.activeRouter.name}`

    return res || null
  }

  get activeRouter() {
    if (this.routes || this.routers.length !== 1) {
      throw new Error('No active router has been determined yet.')
    }

    return this.routers[0]
  }
}
```

EdamAme-x:
It may be necessary to consider the behaviour when explicitly specified.

yusukebe:
@EdamAme-x 

I don't know what benchmarks you have done, but `LinearRouter` and `PatternRouter` are slow and there is no need to use them. Use the following benchmarks:

https://github.com/honojs/hono/tree/main/benchmarks/routers

EdamAme-x:
understood. ty

PR #1525, opened 2023-09-29
feat(jwt): add optional flag for middleware

ivyrze:
I added the option to disable the HTTP 401 response when using the JWT middleware. That way API routes can still make use of the payload context without necessarily requiring users to be authenticated to access the route.

```js
app.use(path, jwt({
  secret: ...,
  optional: true
}));
```

Happy to submit a PR for the website docs as well, or whatever else you may need.

Thanks for all of your work on Hono, it's a joy to work with!

### DIFF
--- a/deno_dist/middleware/jwt/index.ts
+++ b/deno_dist/middleware/jwt/index.ts
@@ -14,7 +14,8 @@ declare module '../../context.ts' {
 export const jwt = (options: {
   secret: string
   cookie?: string
-  alg?: string
+  alg?: string,
+  optional?: boolean
 }): MiddlewareHandler => {
   if (!options) {
     throw new Error('JWT auth middleware requires options for "secret')
@@ -45,6 +46,11 @@ export const jwt = (options: {
     }
 
     if (!token) {
+      if (options.optional) {
+        await next()
+        return
+      }
+
       const res = new Response('Unauthorized', {
         status: 401,
         headers: {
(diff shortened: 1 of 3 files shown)


rafaell-lycan:
@ivyrze in that case, isn't easier to simply avoid using the middleware on certain routes?

What would be the use case for it?

ivyrze:
Our use case has some routes which the general public can view, but being logged in includes additional data on top of that. So we'll need to access `jwtPayload` without requiring it. And on other routes we use the current/standard functionality. Sure, we could write our own middleware to handle this, but this helps simplify things.

yusukebe:
I don't really agree with you on this change. Even though it is optional, this is middleware for authentication, so it should always throw an HTTPException if it fails.

Of course, since the user has enabled it themselves, it is their responsibility. But it is a security risk if, although the authentication has failed, the error is not throwed and the process moves on to the next middleware or handler.

rafaell-lycan:
> Our use case has some routes that the general public can view, but being logged in includes additional data on top of that.

@ivyrze a workaround would be adding a custom middleware that signs a generic JWT a common structure to avoid failing the JWT verification. You can use the same piece of code you have to create tokens as a util.

```tsx
import { sign } from 'hono/jwt'

export const signUnknownToken = (): MiddlewareHandler => {
  return async async (ctx, next) => {
    const hasAuthToken = ctx.req.header('Authorization')
    if (!hasAuthToken) {
      const payload = { user: null, user_type: 'unknown' }
      const token = await JWT.sign(payload, secret, alg as AlgorithmTypes)
      ctx.req.headers.set('Authorization', token)
    }
    next()
  }
}
```

What @yusukebe mentioned regarding being a security risk makes sense, and by signing a new JWT yourself you can have full control of the request flow, and although the middleware could have it built in, it could be too specific to your use case.

Issue #1018, opened 2023-03-30
req.parseBody should allow multiple keys (and converted to array)


Abdillah:
# Currently
Only the last key is recorded into the resulting object after `req.parseBody()`.

# Expected
When parsing request parameters that consists of multiple keys, the keys should be converted to array automatically.

```
tests=1&tests=2&tests=3

{ tests: [ 1,2,3 ] }
```

Also, another use case where using `tests[]` which has been around for some time also need to be supported.

```
tests[]=1

{ tests: [ 1 ] }
```  

Related: https://github.com/honojs/hono/issues/257

bogordesaincom:
please see https://hono.dev/api/request 

```
// Captured params
app.get('/entry/:id', (c) => {
  const id = c.req.param('id')
  ...
})

// Get all params at once
app.get('/entry/:id/comment/:commentId', (c) => {
  const { id, commentId } = c.req.param()
})

```

Abdillah:
@bogordesaincom The one that's more relevant is this https://hono.dev/api/request#queries. Thanks for the hint.

bogordesaincom:
ahh yes. that query section

RussellSanders1:
this doesn't answer the question, as the question is clearly about [`req.parseBody()`](https://hono.dev/api/request#parsebody).

I have run into this issue as well today.  They are right that only the last value is parsed from the body.

yusukebe:
Hi. We may fix it. But it will be a breaking change. So, will be included in the next major release.

RussellSanders1:
I saw a similar issue which suggested manually parsing form data, so it’s not a priority for me. Thanks!

Abdillah:
> this doesn't answer the question, as the question is clearly about [req.parseBody()](https://hono.dev/api/request#parsebody).

Because parseBody and queries has the same format, I hope both enable single key multiple value parsing to array.

yusukebe:
I think this implementation is good, if passing the `all: true` as an option, it return multiple values as an array:

```ts
c.req.parseBody({ all: true })
```

With this way, it will not be a breaking change.

rimai4:
I ran into this as well. 

I think it would be better to parse multiple keys into an array by default for consistency & predictability. `{ all: true }` is inconsistent with the way Hono handles queries. Parsing multiple keys into an array by default also makes it clear for users what to expect: post a key once if you want a single value and post it multiple times if you want an array.

[...]

PR #1528, opened 2023-09-30
feat: `parseBody()` for multi values' field

sor4chi:

Hi @yusukebe 
This is an implemention of the suggesion in #1018
As you mention, this changes have backward compatible.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/request.ts
+++ b/deno_dist/request.ts
@@ -10,7 +10,7 @@ import type {
   ValidationTargets,
 } from './types.ts'
 import { parseBody } from './utils/body.ts'
-import type { BodyData } from './utils/body.ts'
+import type { BodyData, ParseBodyOptions } from './utils/body.ts'
 import type { Cookie } from './utils/cookie.ts'
 import { parse } from './utils/cookie.ts'
 import type { UnionToIntersection } from './utils/types.ts'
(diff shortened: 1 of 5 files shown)


rafaell-lycan:
That's quite specific but it might be useful. 👏

yusukebe:
Hi @sor4chi 

Thanks for a great PR!

One thing. How do we treat an object with a single value? In the following example, it is about the value of `message`.

```ts
const data = new FormData()
data.append('file', 'aaa')
data.append('file', 'bbb')
data.append('message', 'hello')
```

In the current implementation, it returns the value of `message` as an array `['hello']`. But, I think the user would expect a `hello` to be returned in this case.

sor4chi:
Hi, @yusukebe 

Indeed it is.
I completely forgot about the test case so I added it and reimplemented it, thanks.

yusukebe (approved):
Looks good! Let's go with this.

yusukebe:
@sor4chi 

Thanks for fixing. I'd like to merge this, but this is a new feature, so will merge it into "next" branch for a next minor release "v3.8.0" later.

sor4chi:
I look forward to using it at Hono Storage, thank you!

PR #1532, opened 2023-10-03
fix(context): use `FetchEvent` instead of `FetchEventLike`

### LINKED ISSUE #1423: This context has no FetchEvent

3.5.8

Unpublished Worker

When I use `c.event`, it throws `This context has no FetchEvent`.

Get the `FetchEvent`.

### What do you see instead?

An error thrown.

### Additional information

I found that a PR of 3 months ago: https://github.com/honojs/hono/pull/1206

This PR changed the behavior of `c.event`.

`instanceof FetchEvent` -> `instanceof FetchEventLike`. And I found that the JavaScript dist of `FetchEventLike` is:

```js
var FetchEventLike = class {
};
```
> /dist/types.js

yusukebe:
Fixes #1423

- [ ] Add tests
- [ ] Run tests


Cherry:
This change, specifically the addition of `<reference lib="webworker" />`, has unfortunately caused #1546 and #1547

PR #1554, opened 2023-10-08
fix: return status 500 when using validator 'form'

b-marques:
When using `validator('form', ...)` hono is returning a 500 status when receiving a POST request with a JSON in request body, instead of a bad request 400, .

This is happenning due to a unhandled error in an
underlying library (@miniflare).
https://github.com/cloudflare/miniflare/pull/711

The code changes in this PR are responsible to prepare the code to handle possible TypeError that can be thrown in the future, by the lib doing the FormData parsing, as per, https://fetch.spec.whatwg.org/#dom-body-formdata.

This PR should wait for bugfix on @miniflare.

- [x] Add tests
- [x] Run tests (don't have deno to perform `yarn test:all`

### DIFF
--- a/src/validator/validator.ts
+++ b/src/validator/validator.ts
@@ -51,7 +51,7 @@ export const validator = <
       case 'json':
         try {
           const contentType = c.req.header('Content-Type')
-          if (!contentType || !/application\/json/.test(contentType)) {
+          if (!contentType || !contentType.startsWith('application/json')) {
             throw new Error(`Invalid HTTP header: Content-Type=${contentType}`)
           }
           /**
--- a/src/validator/validator.ts
+++ b/src/validator/validator.ts
@@ -70,17 +70,29 @@ export const validator = <
         }
         break
       case 'form': {
-        const contentType = c.req.header('Content-Type')
-        if (contentType) {
-          const arrayBuffer = c.req.bodyCache.arrayBuffer ?? (await c.req.raw.arrayBuffer())
-          const formData = await bufferToFormData(arrayBuffer, contentType)
-          const form: BodyData = {}
-          formData.forEach((value, key) => {
-            form[key] = value
-          })
-          value = form
-          c.req.bodyCache.formData = formData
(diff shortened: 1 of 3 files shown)


b-marques:
I see...

For sure, I will wait for #1558 to be merged to proceed with tests on my side.

yusukebe:
Hi @b-marques 

I've merged #1558 into the main. Try it.

b-marques:
Hi @yusukebe , it stills relevant.

The Node.js native Web APIs can also throw Errors that hono is not handling.

yusukebe:
@b-marques 

> The Node.js native Web APIs can also throw Errors that hono is not handling.

The test on my local and CI passed, is there still a problem?

>  I would like also to mention about the possibility to add a check for the 'appliaction/json' header in the validator.

That's right. If this PR is merged, can you create another one?

b-marques:
@yusukebe  
> The test on my local and CI passed, is there still a problem?
> 
With the changes in this PR, it's not going to be a problem anymore. But without those changes, we still receive a 500 status, when we should receive a 400 Bad request.

> That's right. If this PR is merged, can you create another one?
For sure.

b-marques:
@yusukebe done (:

src/validator/validator.ts
```diff
@@ -50,6 +50,10 @@ export const validator = <
     switch (target) {
       case 'json':
         try {
+          const contentType = c.req.header('Content-Type')
+          if (!contentType || !/application\/json/.test(contentType)) {
```
yusukebe, line 54: `/^application\/json/` is beter.

```ts
if (!contentType || !/^application\/json/.test(contentType)) {
```

yusukebe (commented):
@b-marques 

Thanks! I've left one comment.

yusukebe (approved):

yusukebe:
Hi @b-marques!

Perfect! I'll merge it now. Thanks for the contribution.

Issue #1542, opened 2023-10-04
Duplicate param name


Cherry:
### What version of Hono are you using?

3.7.3

### What runtime/platform is your app running on?

Cloudflare Workers

### What steps can reproduce the bug?

The following code doesn't work as expected:

```javascript
import { Hono } from 'hono'

const app = new Hono()

app.get('/:type/:url', (c) => {
	c.text(`type: ${c.req.param('type')}, url: ${c.req.param('url')}`)
});

app.get('/foo/:type/:url', (c) => {
	c.text(`foo type: ${c.req.param('type')}, url: ${c.req.param('url')}`)
});

export default app
```
It produces the following error:

```
Error: Duplicate param name, use another name instead of 'type' - GET /foo/:type/:url <--- 'type'
```

### What is the expected behavior?

Compared to a similar `express` app in Node.js:

```javascript
const express = require('express');
const app = express();
const port = 3000;

app.get('/:type/:url', (req, res) => {
  res.send(`type: ${req.params.type}, url: ${req.params.url}!`);
});

app.get('/foo/:type/:url', (req, res) => {
	res.send(`foo type: ${req.params.type}, url: ${req.params.url}!`);
});

app.listen(port, () => {
  console.log(`Example app listening on port ${port}`);
});
```

Or an `itty-router` app for Cloudflare Workers:
```javascript
import {Router, error, json} from 'itty-router'

const router = Router()

router.get('/:type/:url', (req) => {
	const params = req.params;
	return new Response(`type: ${req.params.type}, url: ${req.params.url}`)
});

router.get('/foo/:type/:url', (req) => {
	const params = req.params;
	return new Response(`foo type: ${req.params.type}, url: ${req.params.url}`)
});

export default {
	fetch: (request, ...args) => router.handle(request, ...args).then(json).catch(error)
}
```

These work just fine. This should also work fine with Hono.

### Additional information

Looking at https://github.com/honojs/hono/issues/569, it seems like this might be intentional. This isn't ideal and most other routers I've used handle this just fine. It's especially a hit to DX when trying to do things like re-use validation like this:
```javascript
import { Hono } from 'hono'
import { zValidator } from '@hono/zod-validator'
import { z } from 'zod';

const app = new Hono()

const paramsSchema = z.object({
	type: z.enum(['internal', 'external']),
	url: z.string().url().max(2048),
});

app.get('/:type/:url', zValidator('param', paramsSchema), (c) => {
	const params = c.req.valid('param');
[...]

PR #1555, opened 2023-10-08
[POC] feat: one `params` per a handler

yusukebe:
This PR enables the router returns one `params` per handler, so that the value of `c.req.param()` is handled for each handler. We can avoid the `Duplicate param name` mentioned in #1542 by using the appropriate router. This means that the following code will work:

```ts

app.get('/:type/:url', (c) => {
  return c.text(`type: ${c.req.param('type')}, url: ${c.req.param('url')}`)
})

app.get('/foo/:type/:url', (c) => {
  return c.text(`foo type: ${c.req.param('type')}, url: ${c.req.param('url')}`)
})

export default app
```

The key is to handle a `hander` and a `params` as one `Result`.

```ts
export interface Router<T> {
  name: string
  add(method: string, path: string, handler: T): void
  match(method: string, path: string): Result<T>[] | null
}

export interface Result<T> {
  handler: T
}
```

In this PR, I've changed `hono-base.ts`, `compose.ts`, `context.ts` and `TrieRouter`.

The combination of `TrieRouter` and `SmartRouter` can handle all patterns. And we could avoid throwing `Duplicate param name` with another router if it’s implemented.

- [ ] Add tests
- [ ] Run tests


yusukebe:
Hi @usualoma !

Wha do you think this PR?

It will be a big change, but there are almost no breaking changes, and the only performance degradation is a little bit in `TrieRouter`. As a side effect, the size has been reduced. As mentioned in #1542, this will improve DX when done well. So I think it is a nice feature.

usualoma:
@yusukebe 
I think #123 is the way to go for this spec in RegExpRouter.
This changes the correct spec, but the speed reduction seems inevitable.


Benchmarks in "benchmarks/handle-event/" are now as follows.

```
Hono (main) x 367,643 ops/sec ±6.00% (78 runs sampled)
Hono (params-per-handler) x 323,972 ops/sec ±5.51% (76 runs sampled)
itty-router x 244,403 ops/sec ±0.85% (97 runs sampled)
sunder x 303,742 ops/sec ±5.82% (75 runs sampled)
worktop x 187,906 ops/sec ±4.51% (81 runs sampled)
Fastest is Hono
```

usualoma:
@yusukebe 
I have an idea for this feature that would make the router interface more complex but would not reduce the speed (maybe even make it slightly faster).
Please wait a moment so I can propose.
(but possibly with a slight increase in the amount of code)

yusukebe:
Hi @usualoma!

Thanks! I'm looking forward to it.

By the way, there was an issue and I've fixed it with #1565 and released a new version now. So, if you run the benchmark, the result will be different. It will be faster than the old version.

Also, we can run `benchmark/routers`. I've benchmarked it locally. The result is as follows:

```txt
• all together
---------------------------------------------------------------------------- -----------------------------
Hono RegExpRouter                     440.27 ns/iter (407.97 ns … 506.63 ns) 460.97 ns 492.25 ns 506.63 ns
Hono RegExpRouter1555                 519.01 ns/iter    (484.27 ns … 584 ns) 543.85 ns  574.7 ns    584 ns
Hono TrieRouter                         2.16 µs/iter     (2.08 µs … 2.29 µs)   2.17 µs   2.29 µs   2.29 µs
@medley/router                        626.11 ns/iter (589.08 ns … 731.72 ns) 645.49 ns 731.72 ns 731.72 ns
find-my-way                           961.66 ns/iter   (887.45 ns … 1.01 µs) 983.14 ns   1.01 µs   1.01 µs
koa-tree-router                       924.42 ns/iter   (861.33 ns … 1.01 µs) 948.13 ns   1.01 µs   1.01 µs
trek-router                             1.78 µs/iter      (1.7 µs … 1.85 µs)   1.81 µs   1.85 µs   1.85 µs
express (WARNING: includes handling)    4.08 µs/iter     (3.97 µs … 4.38 µs)   4.11 µs   4.38 µs   4.38 µs
koa-router                              1.38 µs/iter     (1.33 µs … 1.51 µs)    1.4 µs   1.51 µs   1.51 µs
radix3                                783.17 ns/iter (750.01 ns … 974.77 ns) 799.75 ns 974.77 ns 974.77 ns
Memoirist                             552.76 ns/iter (510.46 ns … 705.82 ns) 566.35 ns 700.41 ns 705.82 ns

summary for all together
  Hono RegExpRouter
   1.18x faster than Hono RegExpRouter1555
   1.26x faster than Memoirist
   1.42x faster than @medley/router
   1.78x faster than radix3
   2.1x faster than koa-tree-router
   2.18x faster than find-my-way
   3.14x faster than koa-router
   4.03x faster than trek-router
   4.9x faster than Hono TrieRouter
   9.26x faster than express (WARNING: includes handling)
```

`RegExpRouter` is still the fastest, but it will be significantly slower.

usualoma:
@yusukebe How about the #1566 approach?

yusukebe:
@usualoma 

Thanks. I'll check it later!

PR #1556, opened 2023-10-09
feat(reg-exp-router): Improve capture group support.

usualoma:
### What will this PR improve?

* `(a|b)` => Automatically converted to non-capturing group and no longer an error in RegExpRouter. 🆕 
* `(?:a|b)` => If you already explicitly use non-capturing groups, it works without error. (Backward compatible)
* `prefix(a|b)` => Throw `UnsupportedPathError`. 🆕 

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/router/reg-exp-router/node.ts
+++ b/deno_dist/router/reg-exp-router/node.ts
@@ -77,7 +77,14 @@ export class Node {
     let node
     if (pattern) {
       const name = pattern[1]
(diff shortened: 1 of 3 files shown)


yusukebe:
@usualoma 

Awesome work! It's compatible with my other routers and I think it's perfect!

I'll merge this into the `next` branch to release as a next minor version "v3.8". Thanks!

PR #1560, opened 2023-10-10
feat(app): `basePath` option for the constructor, deprecate `app.basePath()`

### LINKED ISSUE #1531: Add `basePath` option for the constructor of Hono

Now, to set the "basePath”, we use the `basePath()` function:

```ts
const app = new Hono().basePath('/api')
```

I propose that we can set the "basePath" by passing the `basePath` option to the constructor of `Hono`:

```ts
const app = new Hono({
  basePath: '/api'
})
```

With this way, we don't have to do `clone()`.

Also, since most use cases, user set `basePath()` when instantiating `Hono`, how about deprecating `basePath()` and making it obsolete in v4? We can also remove `clone()`.

yusukebe:
* Create `basePath` option for the constructor of `Hono`.
* Deprecate `app.basePath()`.

You can specify the base-path like the following:

```ts
const app = new Hono({ basePath: '/api' })
```

Resolves #1531

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/hono.ts
+++ b/src/hono.ts
@@ -1,4 +1,5 @@
 import { HonoBase } from './hono-base'
+import type { HonoOptions } from './hono-base'
 import { RegExpRouter } from './router/reg-exp-router'
 import { SmartRouter } from './router/smart-router'
 import { TrieRouter } from './router/trie-router'
@@ -9,14 +10,10 @@ export class Hono<
   S extends Schema = {},
   BasePath extends string = '/'
 > extends HonoBase<E, S, BasePath> {
-  constructor(
-    init: Exclude<ConstructorParameters<typeof HonoBase>[0], 'basePath'> & {
-      basePath?: BasePath
-    } = {}
-  ) {
-    super(init)
+  constructor(options: HonoOptions<E, BasePath> = {}) {
+    super(options)
     this.router =
-      init.router ??
+      options.router ??
       new SmartRouter({
         routers: [new RegExpRouter(), new TrieRouter()],
       })
--- a/src/hono.ts
+++ b/src/hono.ts
@@ -9,7 +9,11 @@ export class Hono<
   S extends Schema = {},
   BasePath extends string = '/'
 > extends HonoBase<E, S, BasePath> {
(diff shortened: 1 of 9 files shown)


src/hono.ts
```diff
@@ -9,7 +9,11 @@ export class Hono<
   S extends Schema = {},
   BasePath extends string = '/'
 > extends HonoBase<E, S, BasePath> {
-  constructor(init: Partial<Pick<Hono, 'router' | 'getPath'> & { strict: boolean }> = {}) {
+  constructor(
+    init: Exclude<ConstructorParameters<typeof HonoBase>[0], 'basePath'> & {
+      basePath?: BasePath
+    } = {}
+  ) {
```
krsbx, line 16: I've been wondering, why not just make the constructor params into a new type? since it's the same one with the other preset

yusukebe: I also would like do that, but can't set the `BasePath` type (line:10) to `basePath`.

krsbx: why not make the type into some sort of generic instead? so it can be like `type HonoContructor<T extends string>`

yusukebe: Do you mean like this?

```ts
type GetPath<E extends Env> = (request: Request, options?: { env?: E['Bindings'] }) => string

type HonoOptions<E extends Env, BasePath extends string> = {
  strict?: boolean
  basePath?: BasePath
  router?: Router<H>
  getPath?: GetPath<E>
}

export class Hono<
  E extends Env = Env,
  S extends Schema = {},
  BasePath extends string = '/'
> extends HonoBase<E, S, BasePath> {
  constructor(init: HonoOptions<E, BasePath> = {}) {
    super(init)
    this.router =
      init.router ??
      new SmartRouter({
      })
  }
}
```

krsbx: yup, I believe this way anyone who want to contribute also can know that if they change the base preset, they also need to change other preset as well since they have a same type, etc.

yusukebe: I see! I'll consider it and may fix it. Thanks!

PR #1562, opened 2023-10-10
fix: Don't kebab case style property values

bconnorwhite:
Currently, both style property keys and values are kebab cased.

This JSX:
```
fontFamily: "Menlo, Consolas, DejaVu Sans Mono, monospace"
```

Results in the following HTML:
```
font-family:-menlo, -consolas, -deja-vu -sans -mono, monospace;
```

Instead of:
```
font-family:Menlo, Consolas, DejaVu Sans Mono, monospace;
```

This fixes the issue by only applying kebab casing to the key, not the value.


yusukebe:
Hi @bconnorwhite!

Thank you for the PR. This is a good fix!

One thing, I'd like you to write a proper test. Could you do it?

yusukebe:
@bconnorwhite 

And please run the command for fixing styles.

```txt
yarn format:fix
```

yusukebe:
@bconnorwhite 

I've added the test and fixed format. Thanks!

PR #1566, opened 2023-10-12
feat: one `params` per a handler (optimized for RegExpRouter)

usualoma:
After thinking about this issue for a while, I have come to the conclusion that the interface should be almost the same as #1555.

#### Data structure based on array

`Result<T>` is now defined as an Array.
https://github.com/usualoma/hono/blob/bb0ff1fb680ce88d2fce991552e05eb17f283f01/src/router.ts#L14

This is because an Array is faster to access elements than an Object, and the bundle size is smaller when minified.
However, since the contribution to speed by making it an Array is limited and the code is less readable, some may argue that the format using Objects is better in this regard.

#### Optimized for RegExpRouter

The following data structures can now be returned from `match`. This allows the RegExpRouter to work as fast as before.

data-structure

### benchmark

In my environment, the results were about the same as before.

benchmarks/handle-event

```
Hono - main x 484,819 ops/sec ±6.77% (72 runs sampled)
Hono - params-per-handler-2 x 495,303 ops/sec ±6.74% (75 runs sampled)
itty-router x 217,435 ops/sec ±2.53% (92 runs sampled)
sunder x 258,899 ops/sec ±5.35% (75 runs sampled)
worktop x 188,327 ops/sec ±3.43% (85 runs sampled)
Fastest is Hono - params-per-handler-2,Hono - main
```

benchmarks/routers (I perceive the same performance because the ranking often changes when the routers are reordered.)

```
• all together
-------------------------------------------------------------------------------- -----------------------------
Hono TrieRouter                             2.01 µs/iter     (1.99 µs … 2.07 µs)   2.02 µs   2.07 µs   2.07 µs
Hono RegExpRouter - main                  443.81 ns/iter (422.96 ns … 468.42 ns) 446.81 ns 462.18 ns 468.42 ns
Hono RegExpRouter - params-per-handler-2  451.29 ns/iter (431.47 ns … 476.33 ns) 453.57 ns 476.24 ns 476.33 ns
@medley/router                            608.95 ns/iter (591.48 ns … 629.72 ns) 612.27 ns 629.72 ns 629.72 ns
find-my-way                                 1.03 µs/iter     (1.01 µs … 1.11 µs)   1.03 µs   1.11 µs   1.11 µs
koa-tree-router                           659.81 ns/iter (638.36 ns … 724.44 ns) 665.09 ns 724.44 ns 724.44 ns
trek-router                                 1.02 µs/iter   (988.83 ns … 1.05 µs)   1.02 µs   1.05 µs   1.05 µs
express (WARNING: includes handling)        7.83 µs/iter     (7.77 µs … 7.93 µs)   7.86 µs   7.93 µs   7.93 µs
koa-router                                 13.03 µs/iter     (12 µs … 164.54 µs)  12.75 µs  17.46 µs  21.96 µs
radix3                                      1.55 µs/iter      (1.5 µs … 1.59 µs)   1.56 µs   1.59 µs   1.59 µs

summary for all together
  Hono RegExpRouter - main
   1.02x faster than Hono RegExpRouter - params-per-handler-2
   1.37x faster than @medley/router
   1.49x faster than koa-tree-router
   2.29x faster than trek-router
   2.32x faster than find-my-way
   3.5x faster than radix3
   4.53x faster than Hono TrieRouter
   17.64x faster than express (WARNING: includes handling)
   29.35x faster than koa-router
```

- [x] Add tests
- [x] Run tests


yusukebe:
Hi @usualoma,

I have taken the time to think about this and I find it to be an interesting approach!

> However, since the contribution to speed by making it an Array is limited and the code is less readable, some may argue that the format using Objects is better in this regard.

Indeed, but I don't think it's significantly less readable.

So, how about we go this way!? If so, can you complete the PR? It's a hard task to write the code for other routers, but I'd like you to do it.

Thanks!

usualoma:
@yusukebe OK, I will try to implement this specification in other routers.

usualoma:
@yusukebe 
By the way, TrieRouter no longer checks for "Duplicate param name" at all, so it no longer makes an error if multiple cases of the same name exist in a single path.

Before
https://github.com/honojs/hono/blob/main/src/router/trie-router/node.test.ts#L435-L440

After
https://github.com/usualoma/hono/blob/48127094af314dad390907f727d1353ff322c4f1/src/router/trie-router/node.test.ts#L451-L458

What do you think about this point with other routers?

In my opinion, it is better to "not check Duplicate param name at all, even on other routers".
The reason for this is that there is no need for the framework to check it, since it is unlikely that there are multiple instances of the same name in a single path.
Instead, stopping checking and making the code shorter would be more useful.

usualoma:
@yusukebe 
Oh, sorry, I didn't explain myself well enough.

RegExpRouter still makes an error for duplicates like `/:id/:id` in this branch, but I'm talking about whether or not to stop doing that check.
https://github.com/usualoma/hono/blob/48127094af314dad390907f727d1353ff322c4f1/src/router/reg-exp-router/node.ts#L100-L102

yusukebe:
@usualoma

> What do you think about this point with other routers?
> In my opinion, it is better to "not check Duplicate param name at all, even on other routers".

I agree. We don't need to check it. And I've implemented it to make `id` be `123` not `456` in the test of TrieRouter:

```ts
it('self', () => {
  const node = new Node()
  node.insert('get', '/:id/:id', 'foo')
  const [res] = node.search('get', '/123/456')
  expect(res.length).toBe(1)
  expect(res[0][0]).toBe('foo')
  expect(res[0][1]['id']).toBe('123')
})
```

I think it's sufficient to have a rule that if the path has the same param name, the first value will be used.

yusukebe:
Should it throw `UnsupportedPathError`?

> https://github.com/usualoma/hono/blob/48127094af314dad390907f727d1353ff322c4f1/src/router/reg-exp-router/node.ts#L100-L102

yusukebe:
> Should it throw `UnsupportedPathError`?

Ah, if we could throw an error in the insert phase, that would be best.

usualoma:
Done.

@yusukebe Review, please.

### For `/:id/:id`

* LinearRouter, RegExpRouter, TrieRouter : If the path is `/123/456`, `id` be `123` not `456`.
* PatternRouter : `throw new UnsupportedPathError()` in the insert phase

yusukebe:
Hi @usualoma,

Thank you very much for the great work. It looks good!

I've added some tests and merged the `next` branch. I think there's no problem, so I'll merge it tomorrow morning.

usualoma:
@yusukebe Thanks. LGTM!

yusukebe:
Okay!! Let's go!!

PR #1593, opened 2023-10-18
fix: `c.req.params()` in nested app with custom error handler.

### LINKED ISSUE #1590: Path parameter not accessible when using app.onError()

3.8.0

Cloudflare Workers

When using `app.onError()` within a nested route, the path parameters are not accessible.

`c.req.param('assignmentId')` does not work when `app.onError(err, c)` is in the nested route file

index.js
routes/example.js

See this project for example: https://github.com/justinwride/hono-path

I expect to access the path parameter for the route **/assignments/:assignmentId?** using `c.req.param('assignmentId')`

### What do you see instead?

Instead, `c.req.param('assignmentId')` is `undefined`

### Additional information

It works if `app.onError()` is put in **index.js**, but not if it is used in **routes/example.js**.

I have been experimenting with Hono and enjoy the simplicity and speed. I'm the CTO of an ed-tech company and hope to use it more in our tech stack.

usualoma:
Fixes #1590

It may look like a dirty hack at first glance, but if you look closely, you will see it is not a problem.

I have also considered the following approaches.

* https://github.com/usualoma/hono/tree/fix-nested-onerror-params-by-getParams
* https://github.com/usualoma/hono/tree/fix-nested-onerror-params-by-undefined

However, they all increase the code and complexity of types, so this PR approach is better.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/hono-base.ts
+++ b/deno_dist/hono-base.ts
@@ -165,7 +165,7 @@ class Hono<
         app.errorHandler === errorHandler
           ? r.handler
           : async (c: Context, next: Next) =>
-              (await compose<Context>([[r.handler, {}]], app.errorHandler)(c, next)).res
+              (await compose<Context>([], app.errorHandler)(c, () => r.handler(c, next))).res
       subApp.addRoute(r.method, r.path, handler)
     })
     return this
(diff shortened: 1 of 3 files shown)


yusukebe:
Hi @usualoma !

> It may look like a dirty hack at first glance, but if you look closely, you will see it is not a problem.

Indeed, it took a little bit of time to understand what it does. But, as you said, it makes sense and it's good that it can be fixed by modifying only one line.

Thank you very much!

PR #1598, opened 2023-10-19
chore(ci): update node v20

watany-dev:
Starting from October 24, NodeJS V20 will become the LTS version. I believe we should also support this in our CI. For now, I've only added Node, but if there is a need to include other runtimes, I am more than happy to accommodate that as well.

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,13 +7,16 @@ on:
 
 jobs:
   main:
-    name: 'Main'
+    name: 'Main with Node.js v${{ matrix.node }}'
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: ['18.18.2', '20.x']
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @watany-dev,

Thank you for bringing this to our attention. However, I'd like to stick to the "one" `Main` test approach.

All Web-standards APIs are fixed and the differences between the versions of Node.js should be tested in the `runtimes` test. I've addressed the issue with PR #1596 concerning the differences between the versions, but it's not a major problem.

Nevertheless, using v20 in the `Main` test is great. So, could you modify it to a single version - v20 test?

watany-dev:
@yusukebe 
Yes, I've set the Main to test with v20. The Node.js Runtime tests are targeting both 18 and 20.

yusukebe:
@watany-dev 

Thanks! That's what I wanted. Merging now.

PR #1599, opened 2023-10-19
chore(ci): bun v1

watany-dev:

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: oven-sh/setup-bun@v1
         with:
-          bun-version: '1.x.x'
+          bun-version: '1.0.6'
       - run: bun run test:bun
 
   fastly:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: oven-sh/setup-bun@v1
         with:
-          bun-version: '0.6.9'
+          bun-version: '1.x.x'
       - run: bun run test:bun
 
   fastly:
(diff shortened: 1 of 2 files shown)


watany-dev:
After upgrading the version of Bun, the tests related to JSX started to fail. As a temporary fix, I'm importing src/jsx, but I would like to know if there’s a better approach.

.github/workflows/ci.yml
```diff
@@ -55,7 +55,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: oven-sh/setup-bun@v1
         with:
-          bun-version: '0.6.9'
+          bun-version: '1.x.x'
```
yusukebe, line 58: We need to fix the version as Bun's API might change frequently.

runtime_tests/bun/index.test.tsx
```diff
@@ -3,6 +3,8 @@ import { serveStatic } from '../../src/adapter/bun'
 import { Context } from '../../src/context'
 import { env, getRuntimeKey } from '../../src/helper/adapter'
 import { Hono } from '../../src/index'
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import { jsx } from '../../src/jsx'
```
yusukebe, line 7: It's not a big problem. If a user uses JSX on Bun, they can set it up correctly with `tsconfig.json`.

yusukebe (commented):
Hi @watany-dev!

Thank you for the PR. I've left some comments.

runtime_tests/bun/index.test.tsx
watany-dev: I understand that this approach is acceptable, but if there is a need to modify the tsconfig.json, I am more than happy to make the necessary adjustments.

yusukebe: We don't need to modify the `tsconfig.json` for this repository. Users just need to set up the `tsconfig.json` as shown below, which is already written in the starter for Bun.

```json
{
  "compilerOptions": {
    "jsx": "react-jsx",
    "jsxImportSource": "hono/jsx",
  }
}
```

yusukebe (commented):
LGTM!

yusukebe:
@watany-dev 

Thanks! Let's go with this.

PR #1613, opened 2023-10-23
Aws lambda cookies fix

OJ217:
### Issue: Unable to set multiple cookies when using aws-lambda adapter
The `getCookie`  and `setCookie` helpers don't work properly when handling multiple cookies with aws-lambda adapter because of different versions of request and response payload in `API Gateway Proxy Event`.

### Fix implementation: 
I created two functions `getCookies` and `setCookies` in src/adapter/aws-lambda/handler.ts file to get and set cookies payload/header corresponding to the version of the `API Gateway Proxy Event`. Additionally, two tests for this implementation are added in the lambda runtime test file.

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/runtime_tests/lambda/index.test.ts
+++ b/runtime_tests/lambda/index.test.ts
@@ -56,18 +56,32 @@ describe('AWS Lambda Adapter for Hono', () => {
     return c.json(lambdaContext)
   })
 
-  const testCookieKey = 'id'
-  const testCookieValue = crypto.randomUUID()
-  const testCookie = `${testCookieKey}=${testCookieValue}`
+  const testCookie1 = {
+    key: 'id',
+    value: crypto.randomUUID(),
+    get serialized() {
+      return `${this.key}=${this.value}`
+    },
+  }
+  const testCookie2 = {
+    key: 'secret',
+    value: crypto.randomUUID(),
+    get serialized() {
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @OJ217 !

Thank for the PR. Looks good for me.

@watany-dev  Can you review this?

watany-dev:
@OJ217 
It generally seems fine, but it would be good to consider that the Lambda Function URL also has cookies, similar to API Gateway V2.
https://docs.aws.amazon.com/ja_jp/lambda/latest/dg/urls-invocation.html

OJ217:
Hi @watany-dev!
I have checked the documentation for Lambda Function URL. It says *The request and response event formats follow the same schema as the [Amazon API Gateway payload format version 2.0](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html#http-api-develop-integrations-lambda.proxy-format).* Since the way of getting and setting cookies is determined by checking the payload version using the existing **isProxyEventV2**, I think it will also work for Lambda Function URL.

yusukebe:
Thanks both.

@OJ217 Is this ready for merging?

OJ217:
@yusukebe Sure!

yusukebe:
Okay! Merging now.

PR #1623, opened 2023-10-26
perf: The doctype is determined on initialization, not on request.

usualoma:
If the initialized app handles multiple requests, it is better to determine the docType at the time of the createRenderer call.

However, in an environment where only a single request is handled (such as where LinearRouter is active), it is better to determine docType at request time (since the response may not be rendered in JSX). If the focus is on that use case, I don't think this PR is necessary.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/middleware/jsx-renderer/index.ts
+++ b/deno_dist/middleware/jsx-renderer/index.ts
@@ -14,17 +14,19 @@ type RendererOptions = {
   docType?: boolean | string
 }
 
-const createRenderer =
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @usualoma !

It was difficult to decide whether to merge this or not. However, I prefer determining `docType` at the request time, as there's a possibility that JSX may not be called as you mentioned. So I'll close this PR and not merge it. Thanks for the suggestion! (Though, I'd like to use your logic to create `docType`).

usualoma:
@yusukebe Understood, no objection!

Issue #1183, opened 2023-06-16
Support AWS Function URL Streaming


RichiCoder1:
AWS Lambda recently added support for Response Streaming (https://docs.aws.amazon.com/lambda/latest/dg/configuration-response-streaming.html).

This is a much more natural fit for Hono as most environments it supports do response streaming by default. The catches are:

1. It uses a magic global "awslambda.streamifyResponse" to handle the streamification
2. It uses NodeJS writable instead of web streams

yusukebe:
Hi @RichiCoder1 

I'm aware of that and I agree with the idea of supporting streaming for AWS Lambda in Hono. If I find the time, I'll attempt to implement it. However, I must admit that I'm not very familiar with AWS. Hence, I hope that someone else might undertake this task.

yusukebe:
Hi @watany-dev 

Is this issue resolved by recent changes around streaming things?

watany-dev:
@yusukebe 

Yes, here are the next development targets!

yusukebe:
Hi @watany-dev 

Can we close this issue?

watany-dev:
@yusukebe 
Yes, this feature is working successfully in v3.10.0.

yusukebe:
Okay, closing. Thanks a lot!

PR #1625, opened 2023-10-28
feat(adaptor): Support AWS Function URL Streaming

watany-dev:
fixed: https://github.com/honojs/hono/issues/1183

We were able to run the streaming response by operating this module on the actual machine. Unfortunately, awslambda.streamifyResponse provided by AWS Lambda is globally available in the Lambda Nodejs environment, but no method for local testing is provided.

Therefore, we decided to separate it as streamHandle from the existing handle. This way, even if there's a bug, existing users won't be affected. Whether to release this as 'experimental' or to halt integration until a testing method is established is a challenging decision.

- [ ] Add tests
- [ ] Run tests


yusukebe:
Hi @watany-dev,

> Unfortunately, `awslambda.streamifyResponse` provided by AWS Lambda is globally available in the Lambda Node.js environment, but no method for local testing is provided.

Can we create a fake API for mocking in testing only? For example, the `Cache API` does not exist on Node.js, but we mock it in [this file](https://github.com/honojs/hono/blob/main/src/test-utils/setup-vitest.ts) to test Cache Middleware.

```ts
type StoreMap = Map<string | Request, Response>

class MockCache {
  // ...
}

const globalStore: Map<string | Request, Response> = new Map()

const caches = {
  open: (name: string) => {
    return new MockCache(name, globalStore)
  },
}

vi.stubGlobal('caches', caches)
```

I think it's OK to mock that are difficult to reproduce depending on the specific environment.

watany-dev:
@yusukebe 
I've taken into account your helpful advice, and I'm pleased to report that the local tests have passed without issue. Thank you.

yusukebe:
@watany-dev 

Great! Looks good. I'll merge this later.

.pre-commit-config.yaml
yusukebe, line 1: You forgot remove this?

yusukebe (changes requested):

.pre-commit-config.yaml
yusukebe: Ah, sorry. Already removed.

yusukebe (approved):

yusukebe:
It's a time for merging! I'll merge this into the "next" branch and will release a RC version first. Thanks!

PR #1626, opened 2023-10-28
feat(jsx): Support async component.

usualoma:

- [x] Add tests
- [x] Run tests


yusukebe:
Hi @usualoma!

It's awesome to support async components. My concern is adding the additional code to `context.ts`. However, the advantage of supporting async may outweigh the increase in code. I'll consider this matter for a while. Thanks!

usualoma:
@yusukebe 
Thanks for the comment.

I have also added support for nested async components in a15cb21. I also think other modifications would need to be made to merge it into the "main" branch.

yusukebe:
This PR is related to #1552

usualoma:
Close this PR since it is included in #1630 and has already been merged into next.

PR #1630, opened 2023-10-30
feat: Introduce streaming API with `Suspense` and `use`.

usualoma:
This PR is based on #1626.

- [x] Add tests
- [x] Run tests


usualoma:
Hi @yusukebe!

There are no significant changes from https://github.com/usualoma/hono/pull/3, but I have added the following three refactorings and a test of the replacement results using "happy-dom".

* 442538e
* b154592
* d7efbab
* 2d15249

usualoma:
Refactoring is complete.

yusukebe:
Hi @usualoma,

I'll take a look this PR along with #1626 and check how it feels to use it. So please give me a minute. Thanks.

yusukebe:
@usualoma 

Is it OK to assume that this PR includes all of #1626? But, if so, let's keep #1626 open until this PR is merged.

usualoma:
@yusukebe

Thanks for the confirmation!

I have made a separate PR to clarify the scope of the change, but this PR contains #1626 in its entirety.

We can merge #1626, then this PR, and so on. Or we can merge this PR into the "main" branch (or "next" branch) and close #1626.

yusukebe (commented):
Please update `package.json` for `hono/jsx/streaming`:

```js
  "exports": {
    //...
    "./jsx/streaming": {
      "types": "./dist/types/jsx/streaming.d.ts",
      "import": "./dist/jsx/streaming.js",
      "require": "./dist/cjs/jsx/streaming.js"
    },
```

```js
  "typesVersions": {
    "*": {
      //...
      "jsx/streaming": [
        "./dist/types/jsx/streaming.d.ts"
      ],
```

src/jsx/streaming.ts
```diff
@@ -0,0 +1,112 @@
...
+    if (e instanceof Promise) {
+      await e
+      return childrenToString(useContext, children)
+    } else {
+      throw e
+    }
+  }
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
yusukebe, line 26: Add a `@experimental` flag:

```ts
/**
 * @experimental
 * `Suspense` is an experimental feature.
 * The API might be changed.
 */
```

src/jsx/streaming.ts
```diff
@@ -0,0 +1,112 @@
...
+  setUseContext(newUseContext)
+  return newUseContext
+}
+
+const setUseContext = (index: number): void => {
+  useIndex = -1
+  currentUseContext = index
+}
+
+export const use = <T>(promise: Promise<T>): T => {
```
yusukebe, line 74: Add a `@experimental` flag:

```ts
/**
 * @experimental
 * `use()` is an experimental feature.
 * The API might be changed.
 */
```

src/jsx/streaming.ts
```diff
@@ -0,0 +1,112 @@
...
+    return useContexts[currentUseContext][useIndex]
+  }
+
+  promise.then((res) => (useContexts[currentUseContext][useIndex] = res))
+
+  throw promise
+}
+
+const textEncoder = new TextEncoder()
+export const renderToReadableStream = (
```
yusukebe, line 87: Add a `@experimental` flag:

```ts
/**
 * @experimental
 * `renderToReadableStream()` is an experimental feature.
 * The API might be changed.
 */
```

yusukebe:
Hi @usualoma!

Sorry for the late reply. I've tested this feature thoroughly and it's amazing!

The example below demonstrates it working seamlessly:

```tsx
import { Hono } from 'hono'
import { use, Suspense, renderToReadableStream } from 'hono/jsx/streaming'
import type { Shop } from './types'

const app = new Hono()

const fetchData = async (): Promise<{ shop: Shop }> => {
  const res = await fetch('https://ramen-api.dev/shops/yoshimuraya')
  return res.json()
}

const AsyncComponent = () => {
  const data = use(fetchData())
  return <p>I like {data.shop.name} 🍜</p>
}

app.get('/', async (c) => {
    <html>
      <body>
        <h1>SSR Streaming</h1>
        <Suspense fallback={<p>loading...</p>}>
          <AsyncComponent />
        </Suspense>
      </body>
    </html>
  )
  return c.body(stream, {
    headers: {
      'Transfer-Encoding': 'chunked',
      'Content-Type': 'text/html; charset=UTF-8'
    }
  })
})

export default app
```

Additionally, with this PR https://github.com/honojs/vite-plugins/pull/19, the `@hono/vite-dev-server` now supports streaming, allowing us to develop an app utilizing this feature alongside Vite.

I've left some comments on the PR, please check them out.

I believe the streaming methods: `use`, `Suspense`, and `renderToReadableStream` should be tagged as experimental. Although they might not be as frequently used as other features, they indeed highlight our capabilities.

usualoma:
Hi @yusukebe!
Thanks for your comments and advice. I have added a commit to each of your points.

### b91a87d

Basically, exceptions should be caught by the application and not by the framework, but if we don't catch them, streaming will not be closed, so we catch them.

I think the test is OK, but when I add this test, the vitest catches the global unhandledRejection and makes an error, so I now skip it.

https://github.com/vitest-dev/vitest/blob/7f5022994ec645838d9fa1867602b2f30d51673a/packages/browser/src/client/main.ts#L104

https://github.com/honojs/hono/pull/1630/files#diff-73983199d8fca23993d81d61ae2256197fce29b2b2855d6ea10dbf57d7283f15R123

### Demo app

Your demo app is simple and great. It works wll. However, the current code calls fetch() needlessly, so I think it would be better to change it a little. (Sorry if you knew what you were doing...)

As is the case with React, the following code will display "fetchData" multiple times.

```ts
import React, { Suspense, use } from "react";
import { renderToReadableStream } from "react-dom/server";

const fetchData = async () => {
  await new Promise((resolve) => setTimeout(resolve, 10));
  console.log("fetchData");
  return "OK";
};

function Component() {
  const res = use(fetchData());
  return <p>${res}</p>;
}

const elm = (
    <Component />
  </Suspense>
);

const stream = await renderToReadableStream(elm);
const decoder = new TextDecoder("utf-8");
  console.log(decoder.decode(chunk));
}
```

image

For example, if the result of fetch() is cached as a Promise as shown below, unnecessary fetch() will not be called.

```ts
import { Hono } from 'hono'
import { use, Suspense, renderToReadableStream } from 'hono/jsx/streaming'
import type { Shop } from './types'

const app = new Hono()

const fetchDataCache = {}
const fetchData = (name): Promise<{ shop: Shop }> =>
  (fetchDataCache[name] ||= fetch(`https://ramen-api.dev/shops/${name}`).then((res) => res.json()))

const AsyncComponent = () => {
  const data = use(fetchData("yoshimuraya"))
  return <p>I like {data.shop.name} 🍜</p>
}

app.get('/', async (c) => {
    <html>
      <body>
        <h1>SSR Streaming</h1>
        <Suspense fallback={<p>loading...</p>}>
          <AsyncComponent />
        </Suspense>
      </body>
    </html>
  )
  return c.body(stream, {
    headers: {
      'Transfer-Encoding': 'chunked',
      'Content-Type': 'text/html; charset=UTF-8',
    },
  })
})

export default app
```

yusukebe:
Hi @usualoma,

Thanks. I understood both.

Additionally, something awesome happened last night. One of the React members, Dan, mentioned the honojs account on Twitter and gave us an advice. It's an honor to receive an advice from a React committer regarding JSX.

https://x.com/dan_abramov/status/1721179995370914192?s=20
https://x.com/dan_abramov/status/1721230099368841219?s=20

I agree with what he said; the `use()` is not necessary for the API. We can simplify it by removing `use()`:

```tsx
const AsyncComponent = async () => {
  const res = await fetch(`https://ramen-api.dev/shops/yoshimuraya`)
  const data = await res.json()
  return <p>I like {data.shop.name} 🍜</p>
}

app.get('/', async (c) => {
    <html>
      <body>
        <h1>SSR Streaming</h1>
        <Suspense fallback={<p>loading...</p>}>
          <AsyncComponent />
        </Suspense>
      </body>
    </html>
  )
  // ...
})
```

This is enabled if the Async Component inside `Suspense` is treated as a streaming component.

What do you think about this? If we can implement it, I believe we won't need to create `use()`.

usualoma:
Hi @yusukebe!

Oh, that is a great honor and much-appreciated advice!
I didn't know that even in React, if we use the Async Component in Suspense, it is treated as a streaming component without using `use()`. If so, it certainly makes more sense not to use `use()` in hono.

I implemented that in f4589b7.
I removed `use()` in 1573ec1; or if we keep it, users who used `use()` in React may be able to migrate to it with little learning cost.

yusukebe:
@usualoma

Thanks!

> I didn't know that even in React, if we use the Async Component in Suspense, it is treated as a streaming component without using `use()`. If so, it certainly makes more sense not to use `use()` in hono.

He might have been referring only to the hono/jsx spec, and not React, as he knows our hono/jsx doesn't follow the React spec completely. In React, it's still necessary to throw a Promise in Suspense, and maybe it should support client components too (although I don't have took a look the spec well).

Anyway. All things are done. I'll prepare the "next" branch and I'll merge this into it.

usualoma:
@yusukebe 

Uh, I wrote the code to make sure it worked. (I haven't read the code for React's internal implementation though.)

The following code using React's Suspense is handled as streaming content in SSR.

```ts
import { Hono } from "hono";
import React, { Suspense } from "react"; // 18.3.0
import { renderToReadableStream } from "react-dom/server";

const app = new Hono();

async function Component() {
  const data = await (
    await fetch("https://ramen-api.dev/shops/takasagoya")
  ).json();
  await new Promise((resolve) => setTimeout(resolve, 1000));
  return <h1>{data.shop.name}</h1>;
}

app.get("/", async (c) => {
    <html>
      <head>
        <meta charSet="utf-8" />
      </head>
      <body>
          <Component />
        </Suspense>
      </body>
    </html>
  );

  return c.body(await stream, 200, {
    "X-Content-Type-Options": "nosniff",
    "Content-Type": "text/html",
  });
});

export default app;
```

https://github.com/honojs/hono/assets/30598/adf596d0-87d1-49bd-8def-841435b33286

Therefore, the fact that async components are treated as streaming content in Suspense is not a specification unique to hono, but a behavior compatible with the original React.

yusukebe:
@usualoma 

I got it! As you said, this is good thing for React user to migrate to hono/jsx. Great!

yusukebe:
It's a time for merging! I'll merge this into the "next" branch and release a RC version first. Thanks!

yusukebe:
@usualoma 

Wow! https://x.com/dan_abramov/status/1721811659453313480?s=20

shellscape:
@yusukebe I'm testing this out and having some TypeScript trouble with the latest example above. I've got 

```ts
    "jsx": "react",
    "jsxFactory": "jsx",
    "jsxFragmentFactory": "Fragment"
```

in my tsconfig.json file, but JSX components aren't being recognized. Any suggestions?

PR #1634, opened 2023-10-31
feat(req): support generics for query/queries/header

### LINKED ISSUE #1632: `c.req.query` should support generics

Hey

Currently, the type of `query` is `Record<string, string>`. However, when I have a predefined type, such as:

```ts
interface A {
  userId: string
  type: 'all' | 'post'
}
```

I hope the type of `query` can be changed to `A`, so that I can directly use `c.req.query<A>()` to pass parameters to functions that accept type `A`.

Perhaps its type definition is like this:

```ts
query<T = Record<string, string>>(): T
```

yusukebe:
May resolve #1632

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -73,22 +73,24 @@ export class HonoRequest<P extends string = '/', I extends Input['out'] = {}> {
     return null
   }
 
-  query(key: string): string | undefined
-  query(): Record<string, string>
-  query(key?: string) {
-    return getQueryParam(this.url, key)
+  query<T extends Record<string, string | undefined>>(): T
+  query<T extends Record<string, string | undefined>, K = keyof T>(
+    key: K
+  ): K extends string ? T[K] : undefined
+  query<T extends Record<string, string | undefined>>(key?: keyof T) {
+    return getQueryParam(this.url, key as string)
   }
 
-  queries(key: string): string[] | undefined
-  queries(): Record<string, string[]>
-  queries(key?: string) {
-    return getQueryParams(this.url, key)
+  queries<T extends Record<string, string[]>>(): T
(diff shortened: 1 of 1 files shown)


src/request.ts
```diff
@@ -73,22 +73,24 @@ export class HonoRequest<P extends string = '/', I extends Input['out'] = {}> {
...
-  query(): Record<string, string>
-  query(key?: string) {
-    return getQueryParam(this.url, key)
+  query<T extends Record<string, string | undefined>>(): T
+  query<T extends Record<string, string | undefined>, K = keyof T>(
+    key: K
+  ): K extends string ? T[K] : undefined
+  query<T extends Record<string, string | undefined>>(key?: keyof T) {
+    return getQueryParam(this.url, key as string)
   }
```
kahosan, line 82:
```suggestion
  query<T extends Record<string, string | undefined>>(key?: Extract<keyof T, string>) {
```

src/request.ts
```diff
@@ -73,22 +73,24 @@ export class HonoRequest<P extends string = '/', I extends Input['out'] = {}> {
...
 
-  query(key: string): string | undefined
-  query(): Record<string, string>
-  query(key?: string) {
-    return getQueryParam(this.url, key)
+  query<T extends Record<string, string | undefined>>(): T
+  query<T extends Record<string, string | undefined>, K = keyof T>(
+    key: K
+  ): K extends string ? T[K] : undefined
+  query<T extends Record<string, string | undefined>>(key?: keyof T) {
```
kahosan, line 80: The type of key can be used `Extract<keyof T, string>`

PR #1650, opened 2023-11-07
fix(helper): Stream SSE Helper Non-Closure

watany-dev:
I fixed a bug where the Stream SSE Helper would not close, preventing the session from ending. Additionally, I've implemented error handling for the pipe.

Actually, it was known that the AWS Lambda Adaptor did not terminate during testing, so I've added tests to cover that :)

- [x] Add tests
- [x] Run tests

### DIFF
--- a/runtime_tests/lambda/index.test.ts
+++ b/runtime_tests/lambda/index.test.ts
@@ -510,10 +510,9 @@ describe('streamHandle function', () => {
 
     mockReadableStream.push('data: Message\ndata: It is 0\n\n')
     mockReadableStream.push('data: Message\ndata: It is 1\n\n')
-    mockReadableStream.push('event: close\ndata: Stream closed\n\n')
     mockReadableStream.push(null) // EOF
 
-    const res = await handler(event, mockReadableStream)
+    await handler(event, mockReadableStream)
 
     const chunks = []
     for await (const chunk of mockReadableStream) {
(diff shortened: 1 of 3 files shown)


runtime_tests/lambda/index.test.ts
```diff
@@ -490,4 +490,39 @@ describe('streamHandle function', () => {
...
+      // eslint-disable-next-line @typescript-eslint/no-empty-function
+      read() {},
+    })
+
+    mockReadableStream.push('data: Message\ndata: It is 0\n\n')
+    mockReadableStream.push('data: Message\ndata: It is 1\n\n')
+    mockReadableStream.push('event: close\ndata: Stream closed\n\n')
+    mockReadableStream.push(null) // EOF
+
+    const res = await handler(event, mockReadableStream)
```
yusukebe, line 516: Could you remove this if it's not necessary.

yusukebe:
> Actually, it was known that the AWS Lambda Adaptor did not terminate during testing, so I've added tests to cover that :)

Wow, thanks! I've left one comment, please check it out.

runtime_tests/lambda/index.test.ts
watany-dev: OK! fixed.

yusukebe (approved):
LGTM!

PR #1656, opened 2023-11-08
fix(jsx/streaming): Fixed a bug that caused async components to be evaluated multiple times inside Suspense.

### LINKED ISSUE #1654: renderToReadableStream and Suspense runs async component twice

next

Deno

Use the following components

```jsx
const AsyncThing = async () => {
    console.log("Async start");
    await new Promise(resolve => setTimeout(resolve, 1000))
    console.log("Async end")
    return <div>Async thing</div>
}

export const Home = () => {
    console.log("Home")
    return <div>
        <h1>Hey på deg!</h1>
            <AsyncThing />
        </Suspense>
    </div>
}
```

Then use renderToReadableStream on Home component

```typescript
const stream = renderToReadableStream(Home());
```

I expect a console output like this

```terminal
Home
Async start
Async end
```

Indicating that the async component was ran only once.

### What do you see instead?

Instead I see

```terminal
Home
Async start
Async end
Async start
Async end
```

Indicating that the async component is ran twice.

### Additional information

When returning this as a stream the request takes ~2s to finish, instead of the expected ~1s.

usualoma:
Fixes #1654

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/jsx/streaming.ts
+++ b/deno_dist/jsx/streaming.ts
@@ -32,7 +32,7 @@ export const Suspense: FC<{ fallback: any }> = async ({ children, fallback }) =>
     res = children.toString()
   } catch (e) {
     if (e instanceof Promise) {
-      res = e
+      res = e.then(() => childrenToString(children))
     } else {
       throw e
     }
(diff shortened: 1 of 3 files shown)


yusukebe:
LGTM!

PR #1660, opened 2023-11-08
feat(jsx/streaming): Support nested Suspense.

### LINKED ISSUE #1658: Nested Suspense not working

next

Deno

Set up request and components:

```jsx
const AsyncThing = async () => {
  console.log("Async start");
  await new Promise((resolve) => setTimeout(resolve, 1000));
  console.log("Async end");
  return <div class="bg-red-500">Async thing</div>;
};

const Home = async () => {
  console.log("Home");
  await new Promise((resolve) => setTimeout(resolve, 1000));

  return (
    <div>
      <h1>Home</h1>
      <Suspense fallback={<div>Loading AsyncThing...</div>}>
        <AsyncThing />
      </Suspense>
    </div>
  );
};

const TestPage: FC = () => {
  return (
    <div>
      <Suspense fallback={<div>Loading Home...</div>}>
        <Home />
      </Suspense>
    </div>
  );
};

app.get('/', (c) => {
    console.log("------------------")
    console.log("Request start")

    return c.stream(async s => {
        s.write("<!DOCTYPE html>");
        const stream = renderToReadableStream(<TestPage />);
            s.write(chunk);
        }
        s.close();
    }, {
        headers: {
            'Content-Type': 'text/html; charset=utf-8',
        },
        status: 200,
    })
});
```

Then visit "/".

First the outer suspense to resolve and stream down it's result, then the inner suspense should do the same. Taking in total 2s.

### What do you see instead?

Only the outer suspense is resolved and renders it's result. The request ends after 1s with "Loading AsyncThing..." visible.

Server side I can observe the AsyncThing component actually invoked, but the result is not sent to user as the request already ended.

### Additional information

_No response_

usualoma:
Fixes #1658

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/jsx/streaming.ts
+++ b/deno_dist/jsx/streaming.ts
@@ -45,8 +45,8 @@ export const Suspense: FC<{ fallback: any }> = async ({ children, fallback }) =>
       ) as HtmlEscapedString
       res.isEscaped = true
       res.promises = [
-        promise.then((html) => {
-          return `<template>${html}</template><script>
+        promise.then((_html) => {
+          const html = `<template>${_html}</template><script>
 ((d,c,n) => {
 c=d.currentScript.previousSibling
 d=d.getElementById('H:${index}')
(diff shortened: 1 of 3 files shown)


yusukebe:
Cool! Works well.

https://github.com/honojs/hono/assets/10682/bfea1839-54f4-4855-81f8-8b3d029517b2

Thank you for your ultrafast implementation.

PR #1663, opened 2023-11-08
perf(jsx/streaming): In multiple Suspense, go ahead in the order of resolved.

usualoma:
Prior to this PR, if there were multiple nested Suspense, they were processed in stages, but in this PR, they will be processed first in the order of termination.

In the test I added this PR, it was `<p>last</p><p>first</p>` before this PR, but now it is `<p>first</p><p>last</p>` with this PR.

- [x] Add tests
- [x] Run tests


yusukebe:
Looks good. Merge now.

PR #1665, opened 2023-11-08
fix(jsx/streaming): Fixed a problem when multiple children are added directly under Suspense.

usualoma:
@yusukebe I'm adding a few more bug fixes now, so please wait a bit.

- [x] Add tests
- [x] Run tests


usualoma:
@yusukebe 
Work completed.

Although 8af2247 is not directly related to this PR, the refactoring has improved the outlook and slightly shortened the code.

Please review.

yusukebe:
Hi @usualoma 

Using `raw()` is good idea. Looks good for me. Thanks!

Issue #1672, opened 2023-11-09
streamHandle is can not Setting Content Type


watany-dev:
### What version of Hono are you using?

3.10.0-rc.2

### What runtime/platform is your app running on?

AWS Lambda

## SSE

```ts
app.get('/sse', async (c) => {
  return streamSSE(c, async (stream) => {
    for (let i = 0; i < 10; i++) {
      const message = `It is ${new Date().toISOString()}`
      await stream.writeSSE({ data: message, event: 'time-update', id: String(i) })
      await stream.sleep(1000)
    }
  })
})
```

### return

```bash
curl -D - https://xxxxxxxx.lambda-url.ap-northeast-1.on.aws/sse
HTTP/1.1 200 OK
Date: Thu, 09 Nov 2023 15:41:50 GMT
Content-Type: application/octet-stream
Transfer-Encoding: chunked
Connection: keep-alive
x-amzn-RequestId: 0b12164f-f26b-47ed-af94-accfc3fc6fbe
X-Amzn-Trace-Id: root=1-654cfdbc-39cfc19d7817a529600d0bdd;sampled=0;lineage=75a52790:0

event: time-update
data: It is 2023-11-09T15:41:48.779Z
id: 0
...
```

## jsx

```ts
app.use(
  '*',
  jsxRenderer(
    ({ children }) => {
      return (
        <html>
          <body>{children}</body>
        </html>
      )
    },
    {
      docType: true,
      stream: true,
    }
  )
)
const AsyncComponent = async () => {
  const c = useRequestContext()
  return <p>Hello {c.req.query('name')}!</p>
}
app.get('/render', (c) =>
  c.render(
    <Suspense fallback={<p>Loading...</p>}>
      <AsyncComponent />
    </Suspense>,
    { title: 'Title' }
  )
)

```

### return

```bash
curl -D - https://xxxxxxxx.lambda-url.ap-northeast-1.on.aws/render&name=hono
[1] 3109
$ HTTP/1.1 200 OK
Date: Thu, 09 Nov 2023 15:47:41 GMT
Content-Type: application/octet-stream
Transfer-Encoding: chunked
Connection: keep-alive
x-amzn-RequestId: a1d825d7-8ddc-4e8b-8679-98f0105a9dc5
X-Amzn-Trace-Id: root=1-654cff1d-5945d82a58228f351a84d9fd;sampled=0;lineage=75a52790:0

<!DOCTYPE html>[object Promise]
```

### What is the expected behavior?

[...]

PR #1673, opened 2023-11-09
fix(Adaptor): Lambda Response Streaming With Content Type 

watany-dev:
fixed: https://github.com/honojs/hono/issues/1672

I'd like to share that we have completed actual device verification with AWS Lambda for the magic method awslambda.HttpResponseStream.from, which adds metadata (statusCode, Headers) to the responseStream.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/runtime_tests/lambda/index.test.ts
+++ b/runtime_tests/lambda/index.test.ts
@@ -481,13 +481,12 @@ describe('streamHandle function', () => {
     mockReadableStream.push('3\n')
     mockReadableStream.push(null) // EOF
 
-    const res = await handler(event, mockReadableStream)
+    await handler(event, mockReadableStream)
 
     const chunks = []
     for await (const chunk of mockReadableStream) {
       chunks.push(chunk)
     }
-    console.log(res)
     expect(chunks.join('')).toContain('0\n1\n2\n3\n')
   })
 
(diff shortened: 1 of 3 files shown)


yusukebe:
Hi @watany-dev !

I didn't know `application/vnd.awslambda.http-integration-response`!

Like the CI is failing, you can't write the actual logic in a `.d.ts` file. I think you should write `streamifyResponse` and `awslambda.HttpResponseStream.from` in `handler.ts` or create a new file for them.

watany-dev:
@yusukebe 
These specific AWS Lambda headers and methods are already executed within the AWS environment, so they are not necessary in the .d.ts file, right? I have removed them and added a reference to the source library in the comments. (This is the official sample implementation for when users are setting up the AWS Lambda environment.)

usualoma:
FYI, in my environment, this PR code fixed the problem and `text/event-stream` , and `text/html; charset=UTF-8` are now sent.
@watany-dev @yusukebe

yusukebe:
Ah, this PR version works. My configuration was wrong; the content-types are now correct.

One thing not directly related to this PR: HTTP Streaming with `Suspense` and `renderToReadableStream()` does not work.

watany-dev:
@usualoma @yusukebe 
Thanks for confirming this.

I can't find a track record of Suspense or renderToReadableStream working with AWS lambda, even with the original React, so it may be too complicated to handle here.

yusukebe:
Hmm.

This PR is OK, so I'll merge it later.

However, I think we might make `Suspense` and `renderToReadableStream()` work on Lambda function URLs. I saw an example here that suggests it may be possible:
https://github.com/aws-samples/serverless-patterns/tree/main/lambda-streaming-ttfb-write-sam

watany-dev:
Thanks, this is interesting and I'll check the behavior, I hope it's a bug in adaptor, otherwise it's tricky...

yusukebe:
Anyway, I'm merging it now. Thanks!

yusukebe:
Created the issue #1678.

PR #1687, opened 2023-11-12
feat(app): use `console.error()` for default `errorHandler`

yusukebe:
Using `console.error(err)` is better than `console.trace(err)` to understand what is happening.

- [ ] Add tests
- [x] Run tests


yusukebe:
This is a good feature. Merging it immediately.

PR #1694, opened 2023-11-14
Invalid HTTP header error is hidden

RobertSasak:
This error is captured and never shown which may by misleading. This PR suggest to throw this error directly to the reponse.

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/deno_dist/validator/validator.ts
+++ b/deno_dist/validator/validator.ts
@@ -46,18 +46,26 @@ export const validator = <
 ): MiddlewareHandler<E, P, V> => {
   return async (c, next) => {
     let value = {}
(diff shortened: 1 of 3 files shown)


yusukebe:
Hi @RobertSasak!

You're right! It's not handling the error correctly. I would like you to do three things:

1. Write a proper test.
2. Fix any lint errors.
3. Run `yarn denoify` and then `git add/commit/push` the generated file for Deno.

Thanks!

yusukebe:
@RobertSasak

I've resolved some issues myself, and everything looks good now. I'm going to merge it. Thanks!

RobertSasak:
Wow, thank you.

PR #1697, opened 2023-11-14
feat(dev): add `showRoutes()` utility method instead of `app.showRoutes()`

usualoma:

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/dev.ts
+++ b/src/dev.ts
@@ -1,0 +1,30 @@
+import type { Hono } from './hono'
+
+interface showRoutesOptions {
+  includeMiddleware?: boolean
+}
+
+export const showRoutes = (hono: Hono, opts?: showRoutesOptions) => {
+  const length = 8
+  const routeCount: Record<string, number> = {}
+  hono.routes
+    .filter(({ handler }) => opts?.includeMiddleware || handler.length === 1)
+    .map((route) => {
+      const key = `${route.method} ${route.path}`
+      routeCount[key] = routeCount[key] ? routeCount[key] + 1 : 1
+      if (routeCount[key] > 1) {
+        return
+      }
+
+      return {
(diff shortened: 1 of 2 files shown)


usualoma:
Which is a good argument to start with.

* Add 'hono/dev` in this "hono" repository
* Make it an independent repository
  * Since this is a development tool, we don't have to be concerned about zero dependencies; we can add npm modules to dependencies

yusukebe:
I believe integrating this into the "hono" repository is a good idea. It would be beneficial to develop `hono/dev` as a [helper](https://hono.dev/guides/helpers).

I often use the `showRoutes()` method, and I think many other developers do too. While using an external npm package can be effective, it might not be necessary to have it outside the "hono" repo.

Regarding npm packages, are there any specific ones you think would be useful? Honestly, I'm not a old Node.js user, so I'm not entirely sure which package we can use :)

yusukebe:
Or the point is what methods/features will be added in the future. We need to decide whether to include them in the core, depending on that consideration.

usualoma:
@yusukebe  Thanks for your comment.

To be honest, I don't know much about the npm modules either.

I thought there might be an npm that could write these things in one line. That would be cool. But it might be better, just more dependencies and more things to worry about.
https://github.com/honojs/hono/pull/1697/files#diff-34d76a96faa930c4b9909c055beb422e601e2267f72ce12d6f15a1c2cc6ea914R21

It comes down to this point.

> Or the point is what methods/features will be added in the future. We need to decide whether to include them in the core, depending on that consideration.

But as you pointed out, "implement as a helper" seems appropriate for answering the current request.

usualoma:
This PR is closed because it will be reimplemented under helper.

Issue #1699, opened 2023-11-15
Using Context.status() does not work when using Context.text()


sant123:
### What version of Hono are you using?

3.10.0

### What runtime/platform is your app running on?

Deno

### What steps can reproduce the bug?

```ts
import { Hono } from "https://deno.land/x/hono@v3.10.0/mod.ts";

const app = new Hono();

app.get("/", (c) => {
  c.status(403);
  return c.text("Message received!");
});

Deno.serve(app.fetch);

```

### What is the expected behavior?

The status code should be 403.

### What do you see instead?

It is 200.

### Additional information

Other responses work good.

```ts
import { Hono } from "https://deno.land/x/hono@v3.10.0/mod.ts";

const app = new Hono();

app.get("/", (c) => {
  c.status(403);
  return c.html("<h1>Message received!</h1>");
});

Deno.serve(app.fetch);
```

It returns 403 status code.

yusukebe:
Hi @sant123,

Thank you for raising the issue and creating the PR.

Your PR is good, but I've created another PR, #1702, with a different approach to fix this.

My PR offers a slightly faster performance compared to the app with your PR. Therefore, I'll merge #1702. Thanks!

PR #1700, opened 2023-11-15
Calling Context.text() now sets the status set by Context.status()

sant123:

- [x] Run tests


sant123:
Closes https://github.com/honojs/hono/issues/1699

yusukebe:
Hi. We'll take this #1702 . Thanks!

PR #1711, opened 2023-11-16
Export APIGatewayProxyResult from aws-lambda

juho:
Adds an export for `APIGatewayProxyResult` so that it can be referenced from a handler:

```ts
export const handler: APIGatewayProxyResult = handle(app);
```

Without it exposed, TS complains:

```
Exported variable 'handler' has or is using name 'APIGatewayProxyResult' from external module "node_modules/.pnpm/hono@3.10.1/node_modules/hono/dist/types/adapter/aws-lambda/handler" but cannot be named.
```

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/adapter/aws-lambda/index.ts
+++ b/src/adapter/aws-lambda/index.ts
@@ -1,4 +1,5 @@
 // @denoify-ignore
-export { handle, streamHandle, type APIGatewayProxyResult } from './handler'
+export { handle, streamHandle } from './handler'
+export type { APIGatewayProxyResult } from './handler'
 export type { ApiGatewayRequestContext, LambdaFunctionUrlRequestContext } from './custom-context'
 export type { LambdaContext } from './types'
--- a/src/adapter/aws-lambda/index.ts
+++ b/src/adapter/aws-lambda/index.ts
@@ -1,4 +1,4 @@
 // @denoify-ignore
-export { handle, streamHandle } from './handler'
+export { handle, streamHandle, type APIGatewayProxyResult } from './handler'
 export type { ApiGatewayRequestContext, LambdaFunctionUrlRequestContext } from './custom-context'
 export type { LambdaContext } from './types'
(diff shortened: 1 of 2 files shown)


src/adapter/aws-lambda/index.ts
```diff
@@ -1,4 +1,4 @@
 // @denoify-ignore
-export { handle, streamHandle } from './handler'
+export { handle, streamHandle, type APIGatewayProxyResult } from './handler'
```
yusukebe, line 2: Could you avoid writing `type` inside `{}`? We use the syntax.

```ts
export { handle, streamHandle } from './handler'
export type { APIGatewayProxyEvent } from './handler'
```

yusukebe:
Hi @juho 

Thanks for creating the PR. Please fix the error and check the comment!

src/adapter/aws-lambda/index.ts
juho: Sure, done

juho:
@yusukebe Thanks, I've made the change

src/adapter/aws-lambda/index.ts
```diff
@@ -1,4 +1,5 @@
 // @denoify-ignore
 export { handle, streamHandle } from './handler'
+export type { APIGatewayProxyResult } from './handler'
```
yusukebe, line 3: You have to export `APIGatewayProxyResult` from `src/adapter/aws-lambda/handler.ts`.

yusukebe (changes requested):

src/adapter/aws-lambda/index.ts
yusukebe: Hi @juho , last one thing!

juho: Doh, brain fart. Done

yusukebe (approved):
LGTM!

yusukebe:
@juho 

Thanks! I'll merge now and ship it soon.

Issue #1705, opened 2023-11-16
Feature Request: JSX Error Boundary Component (or Try-Catch in JSX)


sjc5:
### What is the feature you are proposing?

Hi there,

It would be great if we could catch errors and render fallbacks in JSX components, either by allowing normal try-catch syntax inside a JSX component, like this (which currently doesn't work):

```tsx
function ThisThrows() {
  throw new Error("Error");
  return <div>Won't be reached</div>;
}

function Fallback() {
  return <div>Caught -- Why won't this render?</div>;
}

function ThisShouldCatch() {
  try {
    return <ThisThrows />;
  } catch {
    return <Fallback />;
  }
}

app.get("/", async (c) => {
  // Results in 500 Internal Server Error, should it render Fallback?
  return c.html(<ThisShouldCatch />);
});
```

...or by introducing a new `ErrorBoundary` component, like this:

```tsx
import { ErrorBoundary } from "hono/jsx";

function ThisShouldCatch() {
  return (
    <ErrorBoundary fallback={<div>Caught</div>}>
      <ThisThrows />
    </ErrorBoundary>
  );
}
```

What do you think?

usualoma:
Hi @sjc5!

Thanks for the suggestion.
I want to keep the spec in line with React and don't want to add too many "Hono-specific extensions."
On the other hand, I see the possibility of an exciting implementation of ErrorBoundary.
Please give me some time to think about it.

yusukebe:
This is fixed with #1714. Now, you can use `ErrorBoundary`!

PR #1714, opened 2023-11-16
Introduce `ErrorBoundary` component

usualoma:
This PR will accomplish the following

### Support `Suspense` in non-streaming mode.

`Suspense` can also be used in a non-streaming configuration. In this case, the output HTML does not contain any JavaScript, but is the text that has been replaced.

However, now that async components are available in Hono, this feature has no particular advantage. It may be useful for users who are used to "throw a Promise in `Suspense`".

```tsx
const app = new Hono()

let shop
function Shop({ id }) {
  if (!shop) {
    throw fetch(`https://ramen-api.dev/shops/${id}`).then((res) => res.json()).then((data) => shop = data.shop)
  }
  return <div>{shop.name}</div>
}

app.get('/suspense-non-streaming', async (c) => {
  return c.html(
    <html>
      <body>
          <Shop id="takasagoya" />
        </Suspense>
      </body>
    </html>
  )
})
```

### Introduce `ErrorBoundary` component.

This is more of a main issue, adding an ErrorBoundary that can be replaced by a fallback when an error occurs in the processing of children.
This is true for synchronous rendering, asynchronous rendering including async, and streaming, all of which produce the same expected results in all environments.

https://github.com/honojs/hono/assets/30598/72d21b6d-01ef-4d92-96c2-2f6b61af3219

```tsx
const app = new Hono()

async function Shop({ id }) {
  const data = await fetch(`https://ramen-api.dev/shops/${id}`).then((res) => res.json())
  return <div>{data.shop.name}</div>
}

function SyncError() {
  throw new Error('Error')
}

async function AsyncError() {
  await new Promise((resolve) => setTimeout(resolve, 2000))
  throw new Error('Error')
}

function Fallback() {
  return <div>Out Of Service</div>
}

app.get('/sync', async (c) => {
  return c.html(
    <html>
      <body>
        <ErrorBoundary fallback={<Fallback />}>
          <SyncError />
        </ErrorBoundary>
      </body>
    </html>
  )
})

app.get('/async', async (c) => {
  return c.html(
    <html>
      <body>
        <ErrorBoundary fallback={<Fallback />}>
          <Shop id='takasagoya' />
        </ErrorBoundary>
        <ErrorBoundary fallback={<Fallback />}>
          <AsyncError />
        </ErrorBoundary>
      </body>
    </html>
  )
})

app.get('/streaming-suspense-all', async (c) => {
  return c.body(
    renderToReadableStream(
      <html>
        <body>
          <ErrorBoundary fallback={<Fallback />}>
              <Shop id='takasagoya' />
            </Suspense>
              <Shop id='sugitaya' />
            </Suspense>
              <AsyncError />
            </Suspense>
          </ErrorBoundary>
        </body>
      </html>
    ),
    {
      headers: {
        'Transfer-Encoding': 'chunked',
        'Content-Type': 'text/html; charset=UTF-8',
      },
    }
  )
})

app.get('/streaming-suspense-each', async (c) => {
  return c.body(
    renderToReadableStream(
      <html>
        <body>
          <ErrorBoundary fallback={<Fallback />}>
              <Shop id='takasagoya' />
            </Suspense>
          </ErrorBoundary>
          <ErrorBoundary fallback={<Fallback />}>
              <Shop id='sugitaya' />
            </Suspense>
          </ErrorBoundary>
          <ErrorBoundary fallback={<Fallback />}>
              <AsyncError />
            </Suspense>
          </ErrorBoundary>
        </body>
      </html>
    ),
    {
      headers: {
        'Transfer-Encoding': 'chunked',
        'Content-Type': 'text/html; charset=UTF-8',
      },
    }
  )
})
```

### Is this a necessary feature for Hono?

A component named ErrorBoundary is introduced in the official React documentation, but it is not provided complete and is to be created by the user as needed. However, there seems to be an OSS that is very widely used.

https://react.dev/reference/react/Component
https://github.com/bvaughn/react-error-boundary

Personally, I would prefer to refrain from original extensions that are not in the original React, but on the other hand, I think that `ErrorBoundary` is useful enough and worth providing officially, as it would be difficult for users to implement it by themselves.

- [ ] Add tests
- [ ] Run tests


usualoma:
@yusukebe What do you think about this PR?

yusukebe:
@usualoma 

Thanks. I'll check this later.

yusukebe:
This PR may close this issue #1705.

yusukebe:
Hi @usualoma 

With this current code, is `Suspense` working correctly?

usualoma:
@yusukebe Thanks!

I have confirmed Suspense behavior without ErrorBoundary with the following code.
Do you have a code that doesn't work?

```tsx
import { jsx } from './src/jsx'
import { Suspense, renderToReadableStream } from './src/jsx/streaming'

const app = new Hono()

const AsyncMoreComponent = async () => {
  await new Promise((resolve) => setTimeout(resolve, 2000))
  const data = await (await fetch('https://ramen-api.dev/shops/takasagoya')).json()
  return <span>{data.shop.name}</span>
}

const AsyncComponent = async () => {
  await new Promise((resolve) => setTimeout(resolve, 2000))
  const data = await (await fetch('https://ramen-api.dev/shops/yoshimuraya')).json()
  return (
    <p>
      I like {data.shop.name} and{' '}
      <Suspense fallback={<span>...</span>}>
        <AsyncMoreComponent />
      </Suspense>{' '}
      🍜
    </p>
  )
}

app.get('/', async (c) => {
    <html>
      <body>
        <h1>SSR Streaming</h1>
        <Suspense fallback={<p>loading...</p>}>
          <AsyncComponent />
        </Suspense>
      </body>
    </html>
  )
  return c.body(stream, {
    headers: {
      'Transfer-Encoding': 'chunked',
      'Content-Type': 'text/html; charset=UTF-8',
    },
  })
})

export default app
```

```
% npx wrangler dev suspense.tsx
```


https://github.com/honojs/hono/assets/30598/a09471c3-ab6e-473f-aed7-0a9bb0f426c4

yusukebe:
@usualoma 

Ah, I made a mistake when importing `jsx`. It works well now!

yusukebe:
@usualoma

I've tried using it and like the feel of it. This is a great feature!

However, I'm considering an alternative syntax. While I appreciate the functionality of `ErrorBoundary`, I don't want write many `<ErrorBoundary>...</ErrorBoundary>`.

```tsx
  <>
    <ErrorBoundary fallback={<Fallback />}>
        <Shop id='takasagoya' />
      </Suspense>
    </ErrorBoundary>
    <ErrorBoundary fallback={<Fallback />}>
        <Shop id='sugitaya' />
      </Suspense>
    </ErrorBoundary>
    <ErrorBoundary fallback={<Fallback />}>
        <Shop id='yoshimuraya' />
      </Suspense>
    </ErrorBoundary>
    <ErrorBoundary>...</ErrorBoundary>
  </>
```

To simplify, what about passing `Fallback` as an attribute, maybe named `onError`?

```tsx
  <>
    <Suspense fallback={<div>Loading...</div>} onError={<Fallback />}>
      <Shop id='takasagoya' />
    </Suspense>
    <Suspense fallback={<div>Loading...</div>} onError={<Fallback />}>
      <Shop id='sugitaya' />
    </Suspense>
    <Suspense fallback={<div>Loading...</div>} onError={<Fallback />}>
      <Shop id='yoshimuraya' />
    </Suspense>
  </>
```

It would also be useful to apply `onError` to `Fragment`:

```tsx
  <Fragment onError={<Fallback />}>
      <Shop id='takasagoya' />
    </Suspense>
      <Shop id='sugitaya' />
    </Suspense>
      <Shop id='yoshimuraya' />
    </Suspense>
  </Fragment>
```

And for cases without `Suspense`, it would be nice to write like this:

```tsx
<Component onError={<Fallback />} />
```

I prefer this syntax as it reduces nesting and complexity though this is a just idea.

usualoma:
@yusukebe 

Ah, yes, I know exactly what you mean. I, too, wondered, "Is it best practice to write an ErrorBoundary every time we use Suspense?" I have been wondering about this question.

Regarding this, in Hono and React (18.3.0 or later?) I think the following code is OK since the expected result is achieved with `async`/`await` in Suspense when using

(One might argue that "it is more declarative to specify with onError.")

```tsx
const Shop = async ({id}) => {
  try {
    const data = await fetch(`https://ramen-api.dev/shops/${id}`).then((res) => res.json())
    return <div>{data.shop.name}</div>
  }
  catch {
    return <Fallback />;
  }
};
   <Shop id='takasagoya' />
</Suspense>
```

Also, regarding the Fragment thing, the same experience can be had by rewriting Fragment to ErrorBoundary (which can be done in this PR branch), but is it different?

```tsx
  <Fragment onError={<Fallback />}>
    ...
  </Fragment>
```

to

```tsx
<ErrorBoundary fallback={<Fallback />}>
    <Shop id='takasagoya' />
  </Suspense>
    <Shop id='sugitaya' />
  </Suspense>
    <Shop id='yoshimuraya' />
  </Suspense>
</ErrorBoundary>
```

My conclusion at this point is that implementing onError is a premature optimization. It would be better only to implement ErrorBoundary in this PR, and if it is still necessary after using it for a while, then add something like onError at that time.

yusukebe:
Hi @usualoma 

> My conclusion at this point is that implementing onError is a premature optimization. It would be better only to implement ErrorBoundary in this PR, and if it is still necessary after using it for a while, then add something like onError at that time.

Fully agree! Let's add the `ErrorBoundary` first. Could you please prepare this PR for review, even though it's almost ready?

usualoma:
@yusukebe
OK, I need to do some refactoring and add some tests, so please wait a bit.

usualoma:
@yusukebe 
Ready for review!

### Exported from top level

ErrorBoundary is a major component, and it is difficult to understand if the import source is separated (e.g., 'hono/jsx/utils'), so I decided to export from the top level. When it is minified, it will be removed by tree shaking if not used, so I do not see a problem with its inclusion in the top level here.

```
import { ErrorBoundary } from 'hono/jsx'
```

### `onError` and `fallbackRender`

The concepts of `onError` and `fallbackRender` are taken from the react-error-boundary.  These can be used to output error logs, display fallbacks according to the error contents, etc.

https://github.com/bvaughn/react-error-boundary#logging-errors-with-onerror
https://github.com/bvaughn/react-error-boundary#errorboundary-with-fallbackrender-prop

### Why support both synchronous and asynchronous rendering?

A lot of additional code is needed to support both. But I still see value in supporting both. Because if there is an "independent component", whether it is handled synchronously or asynchronously may depend on the request handler that renders it.

### Do you really need to use `ErrorBoundary`?

One thing I would like to add.
I am sure this component is beneficial, but when using it, please consider whether it is appropriate to use `ErrorBoundary`. For `jsx` components processed on the server side, the fact that an error occurs in the View is most likely because the View contains business logic. In general, it is better to have the business logic in the process before the View, so when using `ErrorBoundary`, it is better to consider whether the View should supplement the error or not.

usualoma:
### `onError`

You can write the following to record errors that occur in a component.

```tsx
      <html>
        <body>
          <ErrorBoundary fallback={<Fallback />} onError={(e) => console.log(e)}>
              <Shop id='takasagoya' />
            </Suspense>
          </ErrorBoundary>
        </body>
      </html>
```

### `fallbackRender`

You can write the following to output fallback according to the content of the error.

```tsx
      <html>
        <body>
          <ErrorBoundary fallbackRender={(e:Error) => <div>Got an error: {e.message}</div>}>
              <Shop id='takasagoya' />
            </Suspense>
          </ErrorBoundary>
        </body>
      </html>
```

sjc5:
This is fantastic!

yusukebe:
@usualoma 

Please wait a moment for reviewing. I'm busy to check other issues and PRs.

yusukebe (approved):
@usualoma

Super cool! Everything looks great to me. Exporting `ErrorBoundary` from `hono/jsx` is good.

Let's go with this. I'll create a "next" branch for `v3.11.0` and merge this into it. Thanks a lot!

PR #1715, opened 2023-11-16
Fix/request param keys

### LINKED ISSUE #1713: Param value of "keys" resolves to a function, not the value "keys"

3.10.1

Node, Cloudflare Workers

Create a Hono app with multiple nested routes:

```ts
// Adapted from Hono's own hono.test.ts, 'Routing' -> 'Nested route' test:

const book = app.basePath("/book");
book.get("/:id", c => {
	// PROBLEM HERE:
	return c.text(`get /book/${c.req.param("id")}`);
});

const user = app.basePath("/user");
user.get("/login", c => c.text("get /user/login"));

const appForEachUser = user.basePath(":id");
appForEachUser.get("/profile", c => c.text(`get /user/${c.req.param("id")}/profile`));
```

When a caller requests a route where the `:id` param value is `"keys"`, the result of `c.req.param("id")` should be `"keys"`.

### What do you see instead?

`c.req.param("id")` returns a function instead.

### Additional information

The function seems to be `Array.prototype.keys`. I'm not sure yet under exactly which circumstances this happens, but at some point, when constructing the request context, the params get constructed such that Hono tries to access the `"keys"` property of an array, rather than of an object, and finding a value (the `keys` method of the array), returns it instead of the expected string value.

I have a minimal reproduction of this issue at https://github.com/AverageHelper/hono-bug-param-keys

A failing test for this test may be added to Hono at [src/hono.test.ts](https://github.com/honojs/hono/blob/1ed742bdb80518437e6d6f9c6ec7a5fbe300f254/src/hono.test.ts#L296), around line 296, like so:
```ts
// ...

    res = await app.request('http://localhost/book/123', { method: 'GET' })
    expect(res.status).toBe(200)
    expect(await res.text()).toBe('get /book/123') // currently works

    res = await app.request('http://localhost/book/keys', { method: 'GET' })
    expect(res.status).toBe(200)
    expect(await res.text()).toBe('get /book/keys') // currently fails, as described

// ...
```

usualoma:
Fixes #1713

### Reproducible versions

v3.8.0 or later

### Reproducible environment

* A router other than RegExpRouter is being used.
* The value is the name of a property or method of the Array object.

### Details

Routers other than RegExpRouter do not return ParamStash, but Hono was mistakenly referring to it in HonoRequest and returning the method.

The fact that null is no longer returned below appears to be a change in behavior, but in fact there is no change in behavior because the outer `if (this._s)` has never been valid and this is unreachable.
https://github.com/honojs/hono/compare/main...usualoma:hono:fix/request-param-keys?expand=1#diff-a7eab0275ad9a7a7d3fbced508e6fabbea622ceaf3c7e9d38e08824f4687bd87L73

There also appears to be no performance degradation due to this change.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/hono.test.ts
+++ b/src/hono.test.ts
@@ -297,10 +297,6 @@ describe('Routing', () => {
     expect(res.status).toBe(200)
     expect(await res.text()).toBe('get /book/123')
 
-    res = await app.request('http://localhost/book/keys', { method: 'GET' })
-    expect(res.status).toBe(200)
-    expect(await res.text()).toBe('get /book/keys')
-
     res = await app.request('http://localhost/book', { method: 'POST' })
     expect(res.status).toBe(200)
     expect(await res.text()).toBe('post /book')
@@ -661,6 +657,12 @@ describe('param and query', () => {
       const res = await app.request('http://localhost/entry/123')
       expect(res.status).toBe(200)
       expect(await res.text()).toBe('id is 123')
+    })
+
+    it('param of /entry/:id is found, even for Array object method names', async () => {
+      const res = await app.request('http://localhost/entry/key')
+      expect(res.status).toBe(200)
+      expect(await res.text()).toBe('id is key')
     })
 
     it('param of /entry/:id is decoded', async () => {
--- a/src/hono.test.ts
+++ b/src/hono.test.ts
@@ -297,6 +297,10 @@ describe('Routing', () => {
     expect(res.status).toBe(200)
     expect(await res.text()).toBe('get /book/123')
 
+    res = await app.request('http://localhost/book/keys', { method: 'GET' })
+    expect(res.status).toBe(200)
+    expect(await res.text()).toBe('get /book/keys')
+
     res = await app.request('http://localhost/book', { method: 'POST' })
     expect(res.status).toBe(200)
     expect(await res.text()).toBe('post /book')
--- a/src/request.ts
+++ b/src/request.ts
@@ -57,7 +57,6 @@ export class HonoRequest<P extends string = '/', I extends Input['out'] = {}> {
   param(key?: string): unknown {
     if (key) {
       const param = (this._s ? this._s[this._p[key] as any] : this._p[key]) as string | undefined
-      console.log(param)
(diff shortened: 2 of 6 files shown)


usualoma:
Hi @yusukebe!

Would you please review?

src/hono.test.ts
```diff
@@ -297,6 +297,10 @@ describe('Routing', () => {
     expect(res.status).toBe(200)
     expect(await res.text()).toBe('get /book/123')
 
+    res = await app.request('http://localhost/book/keys', { method: 'GET' })
```
yusukebe, line 300: The problem occurs not only with `nested routes` but also in the following case:

```ts
app.get('/foo/:id', (c) => c.text(c.req.param('id')))

// GET /foo/keys
```

So, you might not want to write the test in this section. I think a better place could be under `param and query`.

src/request.ts
```diff
@@ -51,26 +55,24 @@ export class HonoRequest<P extends string = '/', I extends Input['out'] = {}> {
...
-        const keys = Object.keys(this._p)
-        for (let i = 0, len = keys.length; i < len; i++) {
-          const key = keys[i]
-          const value = this._s[this._p[key] as any] ?? this._p[key]
-          if (value && typeof value === 'string') {
-            decoded[key] = /\%/.test(value) ? decodeURIComponent_(value) : value
-          }
+    if (key) {
+      const param = (this._s ? this._s[this._p[key] as any] : this._p[key]) as string | undefined
+      console.log(param)
```
yusukebe, line 60: This is not needed!

yusukebe (commented):
@usualoma 

Thanks! I've left some comments.

src/hono.test.ts
usualoma: Indeed!

src/request.ts
usualoma: Oops!

usualoma:
@yusukebe Thank you for your comments.
Updated!

yusukebe:
LGTM!

Issue #1712, opened 2023-11-16
Do you consider adding a source map feature


Zevino:
### What is the feature you are proposing?

Do you consider adding a source map feature, as it is difficult to locate large files

yusukebe:
Yes. We can consider to support it.

Issue #676, opened 2022-11-24
`app.showRoutes()` has duplicate routes when middleware is used


Skye-31:
Example code
```js
app.get("/a", someMiddleWare, (c) => { 
  new Response("a");
})
app.showRoutes();
```
app.showRoutes() outputs:
```
GET    /a
GET    /a
```

yusukebe:
Hi @Skye-31 !

Thank you for creating the issue.

It is not a problem that `app.showRoutes()` shows duplicate routes. `app.showRoutes()` just displays the registered routes in order. When displaying, it is possible to remove duplicates, but in this way, we can know how many Handlers or Middleware are registered. I think this is a better way.

Skye-31:
It could be neater to display it as the following, or something similar? 🙂 
```
GET    /a (2x)
```

Skye-31:
I just thought I'd raise it since it seemed like a bug - I'm not too bothered if it overcomplicates things. Just found it confusing for my app to display a few hundred routes when I'm actually only running about 70 😅

dario-piotrowicz:
Would it be possible to add a parameter for the `showRoutes` method not to show duplicate routes if that's what the developer wants?

yusukebe:
> Would it be possible to add a parameter for the showRoutes method not to show duplicate routes if that's what the developer wants?

Yes. But I don't want to give the options to developers for the `showRoutes` to make this simple. And it may make the code verbose mentioned above. In this case, the problem is the developer may feel, "Is this bug?". 

For example, there could be a more clear way to show the routes, such as changing the color, numbering, and so on. However, we don't know the best way.

dagnelies:
Hi, here again a little feedback concerning the representation.
Although I understand the output of showRoutes() I was a little bit confused by the routes not representing the endpoints.

I personally think the ideal representation would be something like this: 

```
GET /a   [logger, basicAuth]
GET /b   [logger, cors, bearerAuth]
```

Or alternatively at least:

```
> ALL *  [logger]
> ALL /a  [basicAuth]
GET /a
> GET /b  [cors]
> ALL /b   [bearerAuth]
GET /b
```

Just to distinguish better which routes are middleware and which routes are actual endpoints. Dunno if it's possible to have named middleware show up there, it would be the bonus. :)

yusukebe:
> I personally think the ideal representation would be something like this:
> Just to distinguish better which routes are middleware and which routes are actual endpoints. 

[...]

PR #1716, opened 2023-11-17
Introduce `inspectRoutes()` and `showRoutes()`

usualoma:
Introduce `showRoutes()` and `captureRouteStackTrace()`.

If it were only "simple to externalize showRoutes()," it might be too much functionality, but this is how I implemented it for now.

As for the file name and line number with routing added would be of little use in a small application, but I am trying to add something valuable to the development. (The line numbers shift a bit in Bun. Node.js does not output the line numbers because it minifies them before executing them, while Deno outputs the correct information.)

### How it works?

```ts
import { captureRouteStackTrace, showRoutes } from './src/helper/dev'
import { createMiddleware } from './src/helper/factory'
import { basicAuth } from './src/middleware/basic-auth'
import { logger } from './src/middleware/logger'

const app = new Hono()

if (process.env.HONO_ENV === 'development') {
  captureRouteStackTrace(app)
}

app.use('*', logger())
app.get('/', basicAuth({ username: 'foo', password: 'bar' }), (c) => c.text('hi'))
app.get(
  '/mw',
  createMiddleware((c, next) => next()),
  (c) => c.text('hi')
)

if (process.env.HONO_ENV === 'development') {
  showRoutes(
    app,
    process.argv[2] === '--show-list' ? { includeMiddleware: true, showList: true } : undefined
  )
}

export default app
```

```
% HONO_ENV=development bun run app.ts # default
GET       /
GET       /mw
% HONO_ENV=development bun run app.ts --show-list
ALL       /*
             logger        /Users/taku/src/github.com/honojs/hono/app.ts:13:0
GET       /    
             basicAuth     /Users/taku/src/github.com/honojs/hono/app.ts:14:0
             [handler]     /Users/taku/src/github.com/honojs/hono/app.ts:14:0
GET       /mw  
             [middleware]  /Users/taku/src/github.com/honojs/hono/app.ts:21:2
             [handler]     /Users/taku/src/github.com/honojs/hono/app.ts:21:2
```

or

```ts
import { captureRouteStackTrace, showRoutes } from './src/helper/dev'
import { createMiddleware } from './src/helper/factory'
import { basicAuth } from './src/middleware/basic-auth'
import { logger } from './src/middleware/logger'

const app = new Hono()

app.use('*', logger())
app.get('/', basicAuth({ username: 'foo', password: 'bar' }), (c) => c.text('hi'))
app.get(
  '/mw',
  createMiddleware((c, next) => next()),
  (c) => c.text('hi')
)

if (process.env.HONO_ENV === 'development') {
  app.get('/show-routes', (c) => {
    let html = '<table><tr><th>Method</th><th>Path</th></tr>'
    showRoutes(app, {
      writer: ({ method, path }) => {
        html += `<tr><td>${method}</td><td>${path}</td></tr>`
      },
    })
    html += '</table>'
    return c.html(html)
  })
}

export default app
```

```
% curl http://localhost:3000/show-routes
<table><tr><th>Method</th><th>Path</th></tr><tr><td>GET</td><td>/</td></tr><tr><td>GET</td><td>/mw</td></tr><tr><td>GET</td><td>/show-routes</td></tr></table>
```

- [ ] Add tests
- [ ] Run tests


usualoma:
Hi @yusukebe!
Would you please give your opinion?

yusukebe:
@usualoma 

Woooow, capturing the line from the Error stack trace! Haha cool idea! It's "too much" as you said, but I like it. And using a named function in built-in or 3rd-party middleware to get the middleware name is good.

> Node.js does not output the line numbers because it minifies them before executing them

This matter might be resolved if it supports adding sourcemap #1712. 

And we can refer to this project: https://github.com/azu/express-router-dependency-graph

Overall, it almost looks good. Let's move forward with this.

yusukebe:
I can't imagine the specific use-cases right now, but the JSON output function which that project has is interesting.

usualoma:
Hi @yusukebe!

Thanks for your comment.
I learned about https://github.com/azu/express-router-dependency-graph; that's helpful.

Ah, it would be more convenient if it were provided in CLI. The following code is not beautiful (not terrible).

```
if (process.env.HONO_ENV === 'development') {
  captureRouteStackTrace(app)
}
```

The "app.ts" contains only the production code, and if you run the following in CLI, `captureRouteStackTrace` will run internally, and you will get the following output, which may be what we need.

```
$ hono-inspect-routes --show-list app.ts
ALL       /*
             logger        /Users/taku/src/github.com/honojs/hono/app.ts:13:0
GET       /    
             basicAuth     /Users/taku/src/github.com/honojs/hono/app.ts:14:0
             [handler]     /Users/taku/src/github.com/honojs/hono/app.ts:14:0
GET       /mw  
             [middleware]  /Users/taku/src/github.com/honojs/hono/app.ts:21:2
             [handler]     /Users/taku/src/github.com/honojs/hono/app.ts:21:2
```

It would be better to add `--format=json` so that the output is in JSON format.

usualoma:
Hi @yusukebe!

I thought about it for a while, but I will complete `showRoutes()` as a helper, as a very simple function in this PR.

I think `captureRouteStackTrace()` is an interesting idea, but the call `captureRouteStackTrace(app)` just doesn't seem pretty, so I'd like to think about it more calmly.

I added `inspectRoutes(app)` as an alternative for users who want to get output in JSON format.

How about this as a PR for solving #676?

yusukebe:
Hi @usualoma 

Sorry for the late reply. I am sick in bed. Will comment when I'm better.

usualoma:
@yusukebe 
Take care and don't worry about the PR.

usualoma:
@yusukebe 
> So, we might need to rethink how we display many items.

Hmmm sure, this is hard to understand. But so far no good ideas have come to mind.

I think you are right about the "verbose" option, so I changed it in dde76a8.

We will consider improving the "verbose" display when there are many items later, but how about merging with this content first?

yusukebe:
@yusukebe 

> We will consider improving the "verbose" display when there are many items later, but how about merging with this content first?

Totally agree! I will merge this into "next" and release it first.

PR #1722, opened 2023-11-18
[PoC] feat(helper/factory): introduce `createHandlers()`

yusukebe:
This PR introduces a `createHandlers()` function from `helper/factory`.

```ts
import { createHandlers } from 'helper/factory'

const handlers = createHandlers(
  validator('json', () => {
    return {
      name: 'foo',
    }
  }),
  (c) => {
    const { name } = c.req.valid('json')
    return c.jsonT({
      name,
    })
  }
)

app.get('/api', ...handlers)
```

This is useful for defining handlers outside of `app.get('/api', ...)`. Using this approach, you can structure your application in a Ruby on Rails-like Controller pattern, as discussed in #1072.

A key feature is its ability to correctly infer types.

Screenshot 2023-11-19 at 6 43 54

If this is acceptable, I'll write the tests.

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/helper/factory/index.test.ts
+++ b/src/helper/factory/index.test.ts
@@ -1,6 +1,9 @@
+import { expectTypeOf } from 'vitest'
 import { hc } from '../../client'
 import { Hono } from '../../index'
-import { createMiddleware } from './index'
+import type { ExtractSchema } from '../../types'
+import { validator } from '../../validator'
+import { createMiddleware, createFactory } from './index'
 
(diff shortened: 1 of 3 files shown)


yusukebe:
Hi @usualoma !

What do you think about this feature?

usualoma:
Hi @yusukebe!

👍 I think this is a nice feature!

yusukebe:
@usualoma 

Thanks! I'll work on it later.

PR #1726, opened 2023-11-20
chore: use biome as formatter and linter

watany-dev:
While this undoubtedly leads to efficiency gains, it's worth discussing its adoption, as there may be room for debate. However, it's certain that the integration would speed up the workflow.

- [x] Add tests
- [x] Run tests


yusukebe:
@watany-dev 

I've never tried Biome yet, so I'll try it.

yusukebe:
Biome seems to be good! It's fast!

However, in my opinion, it might be too early to introduce it to our project. Many developers have already installed extensions for `prettier` and `eslint`, but that's not the case for `biome`. While the linter and formatter can work on CI, it's better if they also work in the users' local environments. To achieve this, users would need to have the `biome` extension installed, right?

watany-dev:
Thanks for checking!
You may be right in that sense. We can preset it to Devcontainer, for example, but the local environment of the developer is not limited to that. we'll think about it again when we have enough to change the share of formatter and linter.

yusukebe:
Okay!

CarlosZiegler:
Hello, 
I like this discussion, I aggred what @yusukebe said but user will be add the extensions for prettier and eslint too or ? I am using biome in two projects, and for me is easy to manage both, only install biome, and enable extensions, dont need to care about conflict of my eslint and project eslint or something like that.
Another thing is the rules, of my side I think cover better as es-lint:
Link: https://biomejs.dev/linter/rules
- Accessibility :  Rules focused on preventing accessibility problems.
- Complexity : Rules that focus on inspecting complex code that could be simplified.
- Correctness : Rules that detect code that is guaranteed to be incorrect or useless.
- Performance : Rules catching ways your code could be written to run faster, or generally be more efficient.
- Security : Rules that detect potential security flaws.
- Style : Rules enforcing a consistent and idiomatic way of writing your code.
- Suspicious : Rules that detect code that is likely to be incorrect or useless.
- Nursery : New rules that are still under development.

ajxbit:
@yusukebe any new thoughts on this?

yusukebe:
We are using the common config for ESLint in the projects in the `honojs` org:

https://github.com/honojs/middleware/blob/main/packages/eslint-config/index.js

Each rule was added for historical reasons. If Biome can inherit these rules or use similar rules, then it is acceptable to use Biome, but at this point, I do not see the need to do so.

ajxbit:
Thanks @yusukebe. Seems like Biome is getting more popular with tools like https://www.ultracite.ai/

PR #1728, opened 2023-11-20
Ensure that calls to `c.json()` have a type equivalent to `c.jsonT()`

usualoma:
I am unfamiliar with PRC mode, so I apologize if I am off base.

The idea of the RPC mode is a great one, but having to call jsonT() instead of json() is expensive to learn. Also, hono-base.ts and context.ts have additional processing for this purpose, so if you could eliminate that, that would be great.

It is an evil hack that is declared to return a TypedResponse but only returns a Response, but if this works well, it is better this way.

In my small demo, I had no problems with this change.

image

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/deno_dist/compose.ts
+++ b/deno_dist/compose.ts
@@ -56,9 +56,6 @@ export const compose = <C extends ComposeContext, E extends Env = Env>(
         }
       }
 
-      if (res !== undefined && 'response' in res) {
-        res = res['response']
-      }
       if (res && (context.finalized === false || isError)) {
         context.res = res
       }
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -46,32 +46,26 @@ interface TextRespond {
 }
 
 interface JSONRespond {
-  <T = JSONValue>(object: T, status?: StatusCode, headers?: HeaderRecord): Response
-  <T = JSONValue>(object: T, init?: ResponseInit): Response
-}
-
-interface JSONTRespond {
   <T>(
-    object: InterfaceToType<T> extends JSONValue ? T : JSONValue,
+    object: InterfaceToType<T> extends JSONValue ? T : T extends JSONValue ? JSONValue : T,
     status?: StatusCode,
     headers?: HeaderRecord
-  ): TypedResponse<
-    InterfaceToType<T> extends JSONValue
-      ? JSONValue extends InterfaceToType<T>
-        ? never
-        : T
-      : never
-  >
-  <T>(
-    object: InterfaceToType<T> extends JSONValue ? T : JSONValue,
-    init?: ResponseInit
-  ): TypedResponse<
-    InterfaceToType<T> extends JSONValue
-      ? JSONValue extends InterfaceToType<T>
-        ? never
-        : T
-      : never
-  >
+  ): Response &
+    TypedResponse<
+      InterfaceToType<T> extends JSONValue
+        ? JSONValue extends InterfaceToType<T>
+          ? never
+          : T
+        : never
+    >
+  <T>(object: InterfaceToType<T> extends JSONValue ? T : JSONValue, init?: ResponseInit): Response &
+    TypedResponse<
+      InterfaceToType<T> extends JSONValue
+        ? JSONValue extends InterfaceToType<T>
+          ? never
+          : T
+        : never
+    >
 }
 
 interface HTMLRespond {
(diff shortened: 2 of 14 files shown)


yusukebe:
Hi @usualoma !

Exactly. We can use `c.json()` to keep supporting the RPC! Initially, I made the `c.jsonT()` to allow users use RPC 
expressly. However now PRC is very general function for Hono, merging `c.jsonT()` into `c.json()` makes sense. This PR is great.

Could you finalize this PR? We should also create an alias from `c.jsonT()` to `c.json()` and mark `c.jsonT()` as deprecated.

usualoma:
@yusukebe 
I have been working on it and got to a much better point, but I need some guidance at the end. When an ambiguous type is passed to `json()`, should we be lax or strict?

### Strictly like `c.jsonT()`

Currently, some tests are failing, but if some one has an application that passes the same ambiguous type to `c.json()`, putting in this PR will cause a type error and the need to make the type explicit.

```diff
diff --git a/src/adapter/vercel/handler.test.ts b/src/adapter/vercel/handler.test.ts
index e4f69e5..772be14 100644
--- a/src/adapter/vercel/handler.test.ts
+++ b/src/adapter/vercel/handler.test.ts
@@ -13,7 +13,7 @@ describe('Adapter for Next.js', () => {
          */
         // eslint-disable-next-line @typescript-eslint/ban-ts-comment
         // @ts-ignore
-        waitUntil: c.executionCtx.waitUntil(),
+        waitUntil: c.executionCtx.waitUntil() as string,
       })
     })
     const handler = handle(app)
diff --git a/src/hono.test.ts b/src/hono.test.ts
index a8f5963..6aabb49 100644
--- a/src/hono.test.ts
+++ b/src/hono.test.ts
@@ -1897,10 +1897,10 @@ describe('Parse Body', () => {
   const app = new Hono()
 
   app.post('/json', async (c) => {
-    return c.json(await c.req.parseBody(), 200)
+    return c.json<{}>(await c.req.parseBody(), 200)
   })
   app.post('/form', async (c) => {
-    return c.json(await c.req.parseBody(), 200)
+    return c.json<{}>(await c.req.parseBody(), 200)
   })
 
   it('POST with JSON', async () => {
```

### Loosely like `c.json()`

Alternatively, we could possibly not return a JSONValue in ambiguous cases.

```diff
diff --git a/src/client/client.test.ts b/src/client/client.test.ts
index cdad9d4..2178980 100644
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -502,7 +502,6 @@ describe('Merge path with `app.route()`', () => {
 
   it('Should not allow the incorrect JSON type', async () => {
     const app = new Hono()
-    // @ts-expect-error
     const route = app.get('/api/foo', (c) => c.json({ datetime: new Date() }))
     type AppType = typeof route
     const client = hc<AppType>('http://localhost')
diff --git a/src/context.ts b/src/context.ts
index 983a455..2e4d25d 100644
--- a/src/context.ts
+++ b/src/context.ts
@@ -47,7 +47,7 @@ interface TextRespond {
 
 interface JSONRespond {
   <T>(
-    object: InterfaceToType<T> extends JSONValue ? T : JSONValue,
+    object: InterfaceToType<T> extends JSONValue ? T : T extends JSONValue ? JSONValue : T,
     status?: StatusCode,
     headers?: HeaderRecord
   ): Response &
@@ -317,7 +317,7 @@ export class Context<
   }
 
   json: JSONRespond = <T>(
-    object: InterfaceToType<T> extends JSONValue ? T : JSONValue,
+    object: InterfaceToType<T> extends JSONValue ? T : T extends JSONValue ? JSONValue : T,
     arg?: StatusCode | ResponseInit,
     headers?: HeaderRecord
   ): Response &
@@ -345,7 +345,7 @@ export class Context<
    * `c.json()` now returns data type, so you can just replace `c.jsonT()` to `c.json()`.
    */
   jsonT: JSONRespond = <T>(
-    object: InterfaceToType<T> extends JSONValue ? T : JSONValue,
+    object: InterfaceToType<T> extends JSONValue ? T : T extends JSONValue ? JSONValue : T,
     arg?: StatusCode | ResponseInit,
     headers?: HeaderRecord
   ): Response &
```

In that case, error checking at https://github.com/honojs/hono/blob/main/src/client/client.test.ts#L505 will no longer be in effect. However, since https://github.com/honojs/hono/blob/main/src/client/client.test.ts#L511 will be "never", the client app will be able to notice that something is wrong. (I wonder if it is hard to locate the error)

yusukebe:
Hi @usualoma 

Thank you for your efforts. I would prefer the "Loosely like c.json()" approach. In this case, being too strict isn't the better option.

usualoma:
@yusukebe Fixed!

> Additionally, I think there are some type definitions that might be redundant and could be removed. But, I'm not sure if it's necessary to do. So let's go merge this first.

That's right. I thought I could delete a few more types too, but I couldn't determine which ones could be deleted.

yusukebe (approved):
LGTM!

PR #1735, opened 2023-11-22
feat(types): Support types until 10 handlers

gabrielDonnantuoni:
Handle until 10 handlers on `app.method([path], ...handlers)` by simple adding more overloads.

This is not optimal by far. I'm compromised to find a better and beautiful solution, but until then I think this will cover more use cases, like I'm personally facing now.

- [X] Add tests
- [X] Run tests

### DIFF
--- a/deno_dist/types.ts
+++ b/deno_dist/types.ts
@@ -85,7 +85,7 @@ export interface HandlerInterface<
     handler: H<E2, P, I, R>
   ): Hono<E & E2, S & ToSchema<M, P, I['in'], MergeTypedResponseData<R>>, BasePath>
 
-  // app.get(handler, handler)
+  // app.get(handler x2)
   <
     P extends string = ExtractKey<S> extends never ? BasePath : ExtractKey<S>,
     I extends Input = {},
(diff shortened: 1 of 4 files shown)


yusukebe:
Hi @gabrielDonnantuoni !

Thank you for this PR. As you mentioned, while it may not be the best solution, it seems to be the only way available. It looks good, and I'll review the details later.

yusukebe (commented):
@gabrielDonnantuoni

I would like to extend the `OnHandlerInterface` to also support 10 handlers:

https://github.com/honojs/hono/blob/c3275eebf051b7f3ba9ab11264334ed31f7a98b0/src/types.ts#L337

Could you add the necessary code for this?

gabrielDonnantuoni:
`OnHandlerInterface` updated, @yusukebe 🤝

yusukebe:
@gabrielDonnantuoni

Great work!

Since this is a new feature, I'll merge it into the "[next](https://github.com/honojs/hono/tree/next)" branch for the upcoming minor release `v3.11.0`. It will be released later.

Thank you very much.

PR #1739, opened 2023-11-27
fix(jsx): Write the raw value if the value is a HtmlEscaped string

usualoma:
Fix a double escaping problem when an escaped string is passed as an HTML attribute value.

- [x] Add tests
- [x] Run tests


usualoma:
Hi @yusukebe!
Would you please review?

yusukebe (approved):
LGTM!

PR #1742, opened 2023-11-28
fix(helper/cookie): make default `Path=/` for `getCookie()`/`setSignedCookie()`

### LINKED ISSUE #1741: setCookie does not set Path=/ in response header

3.9.1

Bun

```js
app.get('/a/b/c/d/e/f', async(c) => {
    setCookie(c, 'hello', 'hello2');
    return c.json({ message: 'hello!' }, 200);
})
```

Expected behaviour is Path=/ is set by default in response header.

If Path=/ is not set by default, it will use the current API path (in test case /a/b/c/d/e/f.)

Spent hours debugging of cookie not working, just noticed that Hono does not have Path=/ automatically.

### What do you see instead?

```
< HTTP/1.1 200 OK
< set-cookie: hello=hello2
```

### Additional information

_No response_

yusukebe:
Close #1741 

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/helper/cookie/index.test.ts
+++ b/src/helper/cookie/index.test.ts
@@ -148,7 +148,19 @@ describe('Cookie Middleware', () => {
       const res = await app.request('http://localhost/set-cookie')
       expect(res.status).toBe(200)
       const header = res.headers.get('Set-Cookie')
-      expect(header).toBe('delicious_cookie=macha')
+      expect(header).toBe('delicious_cookie=macha; Path=/')
+    })
+
+    app.get('/a/set-cookie-path', (c) => {
+      setCookie(c, 'delicious_cookie', 'macha', { path: '/a' })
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @usualoma 

Can you review this?

yusukebe:
In this PR, I've rewrite some tests but, the behavior of `setCookie()`/`setSignedCookie()` is not changed for end-users.

usualoma:
@yusukebe 

Yes, I believe this is the expected result of `setCookie()` / `setSignedCookie()` !

You should probably apply `s/getCookie/setCookie/` to the title of this PR.

yusukebe:
> You should probably apply `s/getCookie/setCookie/` to the title of this PR.

Ah!

Thanks!

PR #1743, opened 2023-11-28
feat: introduce `createFactory()` and `createHandlers()`

yusukebe:
I made #1722 a draft and create this new PR instead it.

This PR introduce `Factory` class, `createFactory()`, and `createHandlers()` in `Factory` exported from `hono/factory`.

```ts
import { createFactory } from 'hono/factory'

const factory = createFactory<Env>()

const middleware = factory.createMiddleware(async (c, next) => {
  c.set('foo', 'bar')
  await next()
})

const handlers = factory.createHandlers(logger(), middleware, (c) => {
  return c.json(c.var.foo)
})

app.get('/api', ...handlers)
```

This is useful for defining handlers outside of `app.get('/api', ...)`. Using this approach, you can structure your application in a Ruby on Rails-like Controller pattern, as discussed in #1072.

A key feature is its ability to correctly infer types.

Screenshot 2023-11-29 at 5 56 32

### Why we can't call `createHandlers()`directly

Ideally, you might write shortly as follows:

```ts
import { createHandlers } from 'helper/factory'
```

However, there is a problem that passing `Env` as follows will break other types:

```ts
createHandlers<Env>()
```

### `createMiddleware()`

You can use `createMiddleware` as before:

```ts
import { createMiddleware } from 'helper/factory'
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/helper/factory/index.ts
+++ b/deno_dist/helper/factory/index.ts
@@ -1,6 +1,296 @@
-import type { Env, Input, MiddlewareHandler } from '../../types.ts'
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import type { Env, Input, MiddlewareHandler, H, HandlerResponse } from '../../types.ts'
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
+/**
+ * @experimental
+ * `Factory` class is an experimental feature.
+ * The API might be changed.
+ */
+export class Factory<E extends Env = any, P extends string = any> {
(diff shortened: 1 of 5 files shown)


yusukebe:
Hi @usualoma 

Please review this one!

usualoma (approved):
I think it's good 👍 
It looked like a pain to have to `createFactory()` when I first saw it, but now I understand that it's a necessary process.

yusukebe:
@usualoma 

Thanks! I'll merge this into the "next" later.

PR #1749, opened 2023-11-29
feat/allow `bearerAuth` to accept a list of token strings

mstibbard:
This PR enables the `bearerAuth` middleware to accept a list of strings. It retains the current-state ability to accept a single string (I.e., **not** a breaking change).

This enables simple usage of multiple tokens of varying privileges. A trivial example is below.

```ts
const rToken = 'read'
const pToken = 'read+write'
const validTokens = [rToken, pToken]

api.use('/posts/:id', async (c, next) => {
  // PATCH and DELETE require the privileged token
  if (c.event.request.method === 'PATCH' || c.event.request.method === 'DELETE') {
    const bearer = bearerAuth({ token: pToken });
    return bearer(c, next);
  }
  // GET works with any valid token
  const bearer = bearerAuth({ token: validTokens });
  return bearer(c, next);
});
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/bearer-auth/index.test.ts
+++ b/src/middleware/bearer-auth/index.test.ts
@@ -5,11 +5,13 @@ describe('Bearer Auth by Middleware', () => {
   let app: Hono
   let handlerExecuted: boolean
   let token: string
+  let tokens: string[]
 
   beforeEach(async () => {
     app = new Hono()
     handlerExecuted = false
     token = 'abcdefg12345-._~+/='
+    tokens = ['abcdefg12345-._~+/=', 'alternative']
 
     app.use('/auth/*', bearerAuth({ token }))
     app.use('/auth/*', async (c, next) => {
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @mstibbard,

Awesome! I really like this feature. The implementation is good, and I appreciate that you included tests.

There's one thing. Could you please run `yarn denoify` and then git add, commit, and push the generated file?

@usualoma,

usualoma:
Thanks @mstibbard!
I think it is a necessary feature and a good implementation!

mstibbard:
> There's one thing. Could you please run `yarn denoify` and then git add, commit, and push the generated file?

Done :) I did run it but saw it added a lot of files so I didn't commit them!

yusukebe:
@mstibbard 

Ah, the tests files are not needed in the `deno_dist`.

mstibbard:
@yusukebe third time is the charm? 😆

yusukebe:
@mstibbard 

Yessss! Great!

First, I'll merge this into "next" branch for the next minor version `v3.11.0` while this is new feature. It will be released in a while.

PR #1753, opened 2023-12-01
feat(base): Set Default Path to '*' for app.use()

watany-dev:
fixed: #1754 

In the previous implementation, if a path was not specified in the app.use method, the route would default to the current path. This was inconvenient when wanting to apply middleware to all requests without specifying a specific path.

The main change is as follows:
We modified the use method in the Hono class so that if the first argument is a middleware function, the default path is set to '*'. This could be a breaking change when the first argument is not specified, and therefore, it might be appropriate to manage this under the v4 branch

- [x] Add tests
- [x] Run tests


yusukebe:
Hi @watany-dev,

Totally agree. And as you've pointed out, since this will be a breaking change, let's plan to introduce it in `v4`. Thanks!

jahands:
I like this! This was the behavior I expected when calling .use() without a path, and every time I get confused when it doesn't work

yusukebe:
Hi @watany-dev !

I made the `v4` branch. I'll merge this into `v4` now!

Issue #1688, opened 2023-11-12
Built-in CSRF helper


htunnicliff:
### What is the feature you are proposing?

It would be useful to have some sort of CSRF protection mechanism built in that users could opt in to if desired. 

OSWAP’s [cheat sheet for CSRF Prevention](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html) has a few strategies that could be implemented in a helper. 

If maintainers are open to this, I would be glad to put forth a PR, as this is definitely a feature I would take advantage of.

yusukebe:
@htunnicliff

If you were to create a "CSRF tokens feature", how would you implement it? Hono doesn't have a built-in session mechanism.

htunnicliff:
@yusukebe based on the philosophies behind Hono, I think two CSRF protection techniques would be a good fit: 

- [Signed Double Submit Cookie](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#signed-double-submit-cookie)
- [Custom Request Headers](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#custom-request-headers)

I think that the signed double submit cookie could be implemented simply, but I would omit the reliance on user or session information since—as you mentioned—Hono does not have built-in sessions.

What do you think?

yusukebe:
I believe we should not implement Signed Double Submit Cookie for several reasons:

1. This method does not require storing values server-side, which is advantageous for Hono.
2. However, if an attacker can alter the cookie value, it introduces a vulnerability.

Moreover, users can implement Custom Request Headers by themselves.

These features are not necessary.

htunnicliff:
> However, if an attacker can alter the cookie value, it introduces a vulnerability.

With signed cookies, I don't see how that would be possible unless an attacker gained access to server-side secrets.

> Moreover, users can implement Custom Request Headers by themselves. These features are not necessary.

While this is true (users can write all this code themselves), this middleware can be implemented without third-party dependencies. 

As an example, Hono provides helpers for decoding, signing, and verifying JWTs. It seems like signing and verifying CSRF tokens would be a natural candidate for a helper, especially in light of the JWT helper.

yusukebe:
BTW, The "Signed Cookie" feature is already implemented in the cookie helper:

[...]

PR #1760, opened 2023-12-01
feat(middleware/csrf): add csrf protection middleware

htunnicliff:
Closes honojs/hono#1688

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/deno_dist/middleware/csrf/index.ts
+++ b/deno_dist/middleware/csrf/index.ts
@@ -1,0 +1,212 @@
+import type { Context } from '../../context.ts'
+import { createMiddleware, getCookie, html, setCookie } from '../../helper.ts'
+import type { HtmlEscapedString } from '../../utils/html.ts'
+import { sign, verify } from '../jwt/index.ts'
+
+const sessionSymbol = Symbol('csrfSession')
+
+declare module '../../context.ts' {
+  interface ContextVariableMap {
+    [sessionSymbol]?: string
+    csrfInput: () => Promise<HtmlEscapedString>
+    csrfToken: () => Promise<string>
+  }
+}
+
+type CsrfConfig = {
+  /**
+   * Strategy for checking CSRF tokens
+   * - `both`: Check both header and form field
+   * - `header`: Check header only (for Fetch requests)
+   * - `form`: Check form field only (for HTML forms)
+   */
+  strategy: 'both' | 'header' | 'form'
+  /**
+   * Name of the form field that contains the CSRF token
+   */
+  tokenFormField: string
+  /**
+   * Header name that contains CSRF token for API requests
+   */
+  tokenHeader: string
+  /**
+   * Name of a specific cookie that contains CSRF token for use in JS requests
+   */
+  tokenCookie: string
+  /**
+   * Name of a specific session cookie that is used to compare CSRF tokens
+   */
+  sessionCookie: string
+  /**
+   * Algorithm used for signing and verifying tokens
+   */
+  algorithm: 'HS256' | 'HS384' | 'HS512'
+  /**
+   * Request methods that should be checked for CSRF tokens
+   */
+  protectRequestMethods: Set<string>
+  /**
+   * Paths that should be ignored for CSRF token checks
+   */
+  ignoredPaths?: ReadonlyArray<string | RegExp>
+}
+
+export function csrf(secret: string, options: Partial<CsrfConfig> = {}) {
+  const config: CsrfConfig = {
+    strategy: 'both',
+    tokenFormField: '_csrf',
+    tokenHeader: 'X-CSRF-Token',
+    tokenCookie: '__Host-csrf_token',
+    sessionCookie: '__Host-csrf_session',
+    algorithm: 'HS256' as const,
+    protectRequestMethods: new Set(['POST', 'PUT', 'PATCH', 'DELETE']),
+    ...options,
+  }
+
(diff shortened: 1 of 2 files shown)


usualoma:
Thank you, @htunnicliff.
I read the code.
It works well enough as a "Signed Double Submit Cookie".
To me, the current implementation seems to have some redundancies and would need to be rewritten in case of a merge, but anyway, I am sure the current code works correctly. It is excellent.

As for us, I think we (hono community) should first consider whether "Verifying Origin With Standard Headers" is enough as Sveltekit is doing or whether it is better to do up to "Signed Double Submit Cookie" as this PR does.

As I mentioned in https://github.com/honojs/hono/issues/1688#issuecomment-1836019924, I think "Verifying Origin With Standard Headers" is sufficient in hono's core, but if there seems to be more opinions that "Signed Double Submit Cookie" is needed, I think we can proceed with this PR.

htunnicliff:
Thanks @usualoma! Feel free to modify this PR however you like!

usualoma:
Hi @yusukebe.

My opinion is what I wrote in my comment above.

> As for us, I think we (hono community) should first consider whether "Verifying Origin With Standard Headers" is enough as Sveltekit is doing or whether it is better to do up to "Signed Double Submit Cookie" as this PR does.

What do you think? There is no need to rush to conclusions, and I don't see anything wrong with taking some time to think about it.

htunnicliff:
Hi @yusukebe,

I think that makes sense. 

> The reasons are that this PR's approach is more complex compared to the "Origin Header" approach and is dependent on cookies

However, this might be important to clarify. If an application is not using cookies, my understanding is that there would not be a CSRF vulnerability. CSRF attacks are only possible if an app is using cookies for authentication.

yusukebe:
@htunnicliff 

> CSRF attacks are only possible if an app is using cookies for authentication.

Indeed, you are right! But, strictly, there might be situations where an application, like a mobile app, stores a token or something similar without using cookies. In such cases, we can say that there is a CSRF vulnerability. But the "Origin Header" method might not support these scenarios too, and I think we don't need to focus on them.

yusukebe:
I think if we go with "Original Header" method, basically @usualoma 's implementation as mentioned in the comment is great:

https://github.com/honojs/hono/issues/1688#issuecomment-1836019924

usualoma:
Once, here to get a common understanding within the project about CSRF.
I understand the following.

### When does a CSRF problem occur?

CSRF occurs with submissions from form elements. A typical simple example is as follows

```html
<form action="https://target-site.example.com/admin/update-password" method="POST">
  <input name="new-password" value="P@ssw0rd"/>
</form>
<script>
document.querySelector("form").submit();
</script>
```

When we access this page, even if it is an external site, your browser will send a POST request to https://target-site.example.com with the cookie value. (although it depends on the SameSite attribute).

```html
<script>
fetch('https://target-site.example.com/admin/update-password', {
    method: 'POST',
    headers: {
        'Content-Type': 'application/x-www-form-urlencoded'
    },
    body: 'password=p@ssw0rd',
    credentials: 'include'
})
</script>
```

This problem also occurs with "fetch() with exactly the same content as submitted from a form element" specifying x-www-form-urlencoded or application/multipart-form-data.

### What is not CSRF?

Suppose we have a request that cannot be generated by a form element submit, such as the following, which authenticates with its an application specific request header. This will generate a preflight request, and if the attack succeeds with this, it is a CORS misconfiguration. This is not a CSRF vulnerability.

```html
<script>
fetch('https://target-site.example.com/admin/update-password', {
    method: 'POST',
    headers: {
        'Content-Type': 'application/x-www-form-urlencoded',
        'X-session-id': 'a9f0dcb0-ef1f-11e9-81b4-2a2ae2dbcce4',
    },
    body: 'password=p@ssw0rd',
})
</script>
```

Similarly, even if an attack from a mobile application is successful, it is a matter of session management for that application. This is not a CSRF vulnerability.


### Don't we need to think about applications that are not using sessions?

As seen in the example above, a serious problem occurs in applications that use cookies for session management. However, even simple inquiry forms without session management should be CSRF-protected if possible.

```html
<form action="https://target-site.example.com/inquiry" method="POST">
  <input name="content" value="Text of the criminal threat"/>
</form>
<script>
document.querySelector("form").submit();
</script>
```

Even for applications that do not use session management, there are cases where it is a good idea to take measures against CSRF vulnerabilities, if possible, as a countermeasure against "DDoS" and "spoofing".

https://stackoverflow.com/questions/65236474/does-public-contact-form-require-a-csrf-token

### Is this PR middleware effective against these?

I think this PR middleware works fine, including applications that do not use session management.

However, I still think that the middleware provided by hono’s core should be a simpler "Original Header" validation method. The "Original Header" validation method is also effective for applications that do not use session management, and the procedure for applying it is simple.

rafaell-lycan:
@htunnicliff you can simply use the [csrf](https://www.npmjs.com/package/csrf) instead.

yusukebe:
@usualoma

Thanks for the clear summary.

Therefore, I'd like to proceed with the 'Origin Header' approach for the 'CSRF Protection Middleware'.

@usualoma would you be willing to create the PR? There's no rush for the implementation. And, even though you've already written some code, someone else, including @htunnicliff, could write the code and credit @usualoma  as a co-author.

usualoma:
Sorry for the delay.
Please take a moment as I would like to proceed with the following PR.
https://github.com/honojs/hono/pull/1823

PR #1762, opened 2023-12-01
refactor: use `#` for private properties

### LINKED ISSUE #1738: [Refactor] using `#`
I used to compare using `#` with using only `private` and `#` was slower for accessing properties in a `class`, but I think I was wrong. I might be wrong, wrong?

If it's wrong, we can refactor the code with replacing `private _foo` with `#foo`.

yusukebe:
Closes #1738

* Use `#foo` instead of `private _foo` for private properties.
* Use proper descriptive property names.

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -22,6 +22,7 @@ export interface ContextRenderer {}
 interface DefaultRenderer {
   (content: string | Promise<string>): Response | Promise<Response>
 }
+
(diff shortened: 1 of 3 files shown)


yusukebe:
Hi @usualoma !

Could you review this one?

PR #1778, opened 2023-12-04
fix(types): Added `E` env type argument to `showRoutes` & `inspectRoutes` `hono` parameter

NuroDev:
### What's been changed?

 - Updated `showRoutes` deprecated doc comments to provide the correct import path (`/dev` instead of `/helper`)
 - Added `TEnv` type argument to both `showRoutes` & `inspectRoutes` helpers & passed to the first (`hono: Hono`) parameter. 
	 - This fixes a type error where if you pass in a Hono instance with a custom environment type these 2 functions will complain as it does not match the default / base `Hono` type(s).

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/helper/dev/index.ts
+++ b/deno_dist/helper/dev/index.ts
@@ -17,7 +17,7 @@ const handlerName = (handler: Function) => {
   return handler.name || (isMiddleware(handler) ? '[middleware]' : '[handler]')
 }
 
-export const inspectRoutes = <TEnv extends Env>(hono: Hono<TEnv>): RouteData[] => {
+export const inspectRoutes = <E extends Env>(hono: Hono<E>): RouteData[] => {
   return hono.routes.map(({ path, method, handler }: RouterRoute) => ({
     path,
     method,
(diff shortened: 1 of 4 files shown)


deno_dist/helper/dev/index.ts
```diff
@@ -17,7 +17,7 @@ const handlerName = (handler: Function) => {
   return handler.name || (isMiddleware(handler) ? '[middleware]' : '[handler]')
 }
 
-export const inspectRoutes = (hono: Hono): RouteData[] => {
+export const inspectRoutes = <TEnv extends Env>(hono: Hono<TEnv>): RouteData[] => {
```
yusukebe, line 20: We use `E` instead of `TEnv` in the same cases. So it would be better to use `E`.

https://github.com/honojs/hono/blob/5e0ef3048b5add5dadbbb4e936254b71058351b2/src/hono-base.ts#L53

yusukebe:
Hi @NuroDev !

Good fix! I've left one comment. Check it!

deno_dist/helper/dev/index.ts
NuroDev: Ah, my mistake. Replaced all occurrences of `TEnv` with `E` & re-run the `denoify` script.

yusukebe (approved):
LGTM!

yusukebe:
@NuroDev 

I'll merge this and will release the latest patch version soon. I appreciate your contribution!

PR #1788, opened 2023-12-06
fix: add missing tail to svg regex

### LINKED ISSUE #1787: Why encode SVG as binary?

3.11.1

AWS Lambda

Return an SVG in a route with

```
c.header('Content-Type', 'image/svg+xml;charset=utf-8');
return c.body(svgString)
```

This forces a user to send the `Accept: image/svg+xml` header, which browsers do not do by default. So, if you open the route in a browser, you only get the base64 encoded content.

SVG is just text, so it shouldn't need to be encoded before returning it to API Gateway.

### What do you see instead?

_No response_

### Additional information

The relevant code is here: https://github.com/honojs/hono/blame/main/src/adapter/aws-lambda/handler.ts#L264 that was added as part of https://github.com/honojs/hono/pull/1091

Example on sending an SVG in a response in a "raw" Lambda handler: https://stackoverflow.com/a/66083753

eliasbrange:

- [x] Add tests
- [x] Run tests

closes #1787

### DIFF
--- a/src/adapter/aws-lambda/handler.test.ts
+++ b/src/adapter/aws-lambda/handler.test.ts
@@ -4,6 +4,8 @@ describe('isContentTypeBinary', () => {
   it('Should determine whether it is binary', () => {
     expect(isContentTypeBinary('image/png')).toBe(true)
     expect(isContentTypeBinary('font/woff2')).toBe(true)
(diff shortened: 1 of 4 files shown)


yusukebe:
@eliasbrange 

Looks good! Thanks!

Issue #1770, opened 2023-12-03
Event Listener on Client Disconnect


HeyITGuyFixIt:
### What is the feature you are proposing?

As mentioned in Discord, could we have the ability to create event listeners for client disconnect or abort events? I would expect to be able to use them like `c.addListener('abort', () => { ...}` and optionally `c.on('abort', () => { ...}`. Or course, we need the removeListener and off functions as well for completeness.

yusukebe:
@sor4chi 

Do you know how to handle user's cancellation such as closing the browsers tab. If you run the `c.stream()` or `c.streamText()` on Bun like the following, it will not be stopped.

```ts
app.get('/stream', (c) => {
  return c.streamText(async (stream) => {
    for (;;) {
      await stream.writeln('Hello')
      // Can I know the user's cancellation?
      await stream.sleep(1000)
    }
  })
})
```

yusukebe:
@HeyITGuyFixIt 

What do you mean `client disconnect or abort events?` ? For example, it might not be enabled to handle closing the browser's tab. Or you use some JavaScript on a client-side?

yusukebe:
@HeyITGuyFixIt 

How can I reproduce "disconnect"?

yusukebe:
@HeyITGuyFixIt 

I'm not sure if the `abort event` can handle browser closing.

I might try implementing a Node.js application that uses the `abort` event to see how it handles a "disconnect." However, if you have a good example, please share it.

yusukebe:
I've created an example of a Node.js HTTP server that handles user disconnections. If you close the browser tab, the server stops. We should implement a similar feature in Hono.

```ts
import * as http from 'http'
import { Readable } from 'stream'

const server = http.createServer((req, res) => {
  res.writeHead(200, {
    'Content-Type': 'text/plain; charset=UTF-8',
    'Transfer-Encoding': 'chunked',
    'X-Content-Type-Options': 'nosniff'
  })

  const readable = new Readable({
    read() {}
  })

  const interval = setInterval(() => {
    const data = `${new Date().toISOString()}`
    console.log(data)
    readable.push(`${data}\n`)
  }, 1000)

  readable.pipe(res)

  req.on('close', () => {
    console.log('Connection is closed')
    clearInterval(interval)
    readable.destroy()
  })
})

const PORT = 3000
server.listen(PORT, () => {
  console.log(PORT)
})
```

sor4chi:
By the way, can we observe a response cancellation event?
[...]

PR #1795, opened 2023-12-09
feat: implement `stream.onAbort`

sor4chi:
ref: #1770 

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -398,13 +398,14 @@ export class Context<
     arg?: StatusCode | ResponseInit,
     headers?: HeaderRecord
   ): Response => {
-    const { readable, writable } = new TransformStream()
-    const stream = new StreamingApi(writable)
+    const transformer = new TransformStream()
+
+    const stream = new StreamingApi(transformer.writable, transformer.readable)
     cb(stream).finally(() => stream.close())
 
     return typeof arg === 'number'
-      ? this.newResponse(readable, arg, headers)
-      : this.newResponse(readable, arg)
+      ? this.newResponse(stream.readable, arg, headers)
+      : this.newResponse(stream.readable, arg)
   }
 
   /** @deprecated
--- a/deno_dist/helper/streaming/sse.ts
+++ b/deno_dist/helper/streaming/sse.ts
@@ -9,8 +9,8 @@ export interface SSEMessage {
 }
 
 export class SSEStreamingApi extends StreamingApi {
-  constructor(writable: WritableStream) {
-    super(writable)
+  constructor(writable: WritableStream, readable: ReadableStream) {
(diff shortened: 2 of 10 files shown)


sor4chi:
Hi, @yusukebe
Here is the POC as of now for now.
I tried to implement it with as few destructive changes as possible, but it looks like I need to change the StreamAPI a bit anyway.
The fact that the test is failing shows that there is a lot of difference in runtime.

yusukebe:
@sor4chi 

Thanks! I'll check this later!

yusukebe:
Hi @sor4chi 

This is a great work! It works fine on Node.js and Deno.

One thing you need to consider is that this behavior is changed from the current release version.

The code:

```ts
app.get('/stream', (c) => {
  return c.streamText(async (stream) => {
    for (;;) {
      await stream.writeln('Hello')
      console.log('Hello')
      await stream.sleep(1000)
    }
  })
})
```

The current version:
https://github.com/honojs/hono/assets/10682/60fc7e86-13c3-438a-a41a-587a076ad2e5

This PR:
https://github.com/honojs/hono/assets/10682/a9269705-03ab-4891-91c0-a4779adf41d9

Could you take a look?

sor4chi:
@yusukebe 
Wow, there was such a bug. I was not aware of that, thank you.

I will certainly have to fix this, but it will take some time.

HeyITGuyFixIt:
@sor4chi @yusukebe any updates?

sor4chi:
@HeyITGuyFixIt 
No, I'm Sorry.
I'm too busy and don't have time to work.

sor4chi:
waiting for #1846

sor4chi:
Hi, @yusukebe
I have investigated this and I believe that the behavior up to now was a bug in the first place.
Do you have any concerns about this change to a specification that interrupts the stream in the middle?

yusukebe:
@sor4chi 

> Do you have any concerns about this change to a specification that interrupts the stream in the middle?

I do not know. I can look into it myself, but I would prefer to leave it to you.

sor4chi:
As for the change that writing to the readable stream is terminated in the middle and no further processing is performed, I would like to proceed as is because the original implementation was poor and I think that changing this behavior in the first place will have little effect on users.

yusukebe:
@sor4chi 

Okay! Let's go with this spec, which is implemented in this PR.

yusukebe:
@sor4chi 

Since this referenced the old code, can you create a new PR based on the "[next](https://github.com/honojs/hono/tree/next)" branch?

sor4chi:
@yusukebe 
oh, sorry. Okay!

sor4chi:
@yusukebe 
this is a new pull request. #1871

Issue #1747, opened 2023-11-29
Support Azure Functions


gunta:
### What is the feature you are proposing?

It would be amazing it Hono supported [Azure Functions](https://github.com/Azure/Azure-Functions).
Being able to change back-and-forth between providers is something needed. 
For example, from Azure Functions → Cloudflare Workers.

gunta:
Thank you!

PR #1797, opened 2023-12-09
feat: implement azure-functions adapter

ytnobody:

- [ ] Add tests
- [ ] Run tests

---

## usage:

```typescript
import { Hono } from 'hono'
import { AzureFunction, Context, HttpRequest, handle } from 'hono/adapter/azure-functions'

const app = new Hono()
app.use('*', logger())

app.all('/api/:funcname', async (c) => {
  const message = 'Hello, Azure Functions and Hono!'
  const query = c.req.queries()
  const json_body = c.req.method === "POST" ? await c.req.json() : {};
  const headers = c.req.header()
  return c.json({ message, query, json_body, headers })
})

const httpTrigger: AzureFunction = async function (context: Context, req: HttpRequest): Promise<void> {
  context.log('HTTP trigger function processed a request.')
  context = await handle(app)({ context, req })
}

export default httpTrigger
```

## ~requirement~

- ~@azure/functions `^3.0.0`~


watany-dev:
I am a developer maintaining an AWS lambda handler

If AzureFunctions user values customizability, this interface looks good.

const httpTrigger: AzureFunction = async function (context: Context, req: HttpRequest): Promise<void> {
  context.log('HTTP trigger function processed a request.');
  context = await handle(app)({ context, req });
};

export default httpTrigger;.

But otherwise, it may be useful to include the async function() part in the handler as well

src/adapter/azure-functions/handler.ts
```diff
@@ -0,0 +1,59 @@
+import type {
+  Context as AzureFunctionsContext,
+  HttpRequest as AzureFunctionsHttpRequest,
+} from '@azure/functions'
```
ytnobody, line 4: I am wondering about this dependency.
Is it ok to write @azure/functions in "dependency" in package.json?

watany-dev: I don't understand this part either, when @yusukebe built the AWS Adaptor, the Forked Event Type was built in, so the Lambda@edge type was also forked.

yusukebe: Yeah, we should avoid making it dependent on `@azure/functions`, even if it only provides types.

yusukebe:
Hi @ytnobody,

Thank you for the PR. Supporting Azure Functions is a great feature for Hono. We also have issue #1747 .

I will try it out and assess how it feels to use. So, please wait for a moment.

@watany-dev, I might ask you to review this as well.

yusukebe:
@ytnobody

I believe the feature in this PR is for 'version 3', but 'version 4' is now available:

https://learn.microsoft.com/en-us/azure/azure-functions/functions-node-upgrade-v4?tabs=v4

Since 'version 3' might no longer be supported, we should update it for 'version 4'.

ytnobody:
@yusukebe 

I misunderstood about a version that @azure/functions.

This PR will be making changes to support Azure Functions v4.

yusukebe:
@ytnobody 

Thanks! But, there's no need to hurry!

mhsrio:
Hello awesome peeps, any update on this PR? Waiting for this feat for a long time. Would be a lifesaver!

yusukebe:
Closing.

You can use https://github.com/Marplex/hono-azurefunc-adapter. Ref: https://github.com/honojs/website/issues/346

Issue #1127, opened 2023-05-23
__STATIC_CONTENT_MANIFEST error when bundling with Cloudflare's wrangler.


JeffBue:
When using Hono@3.2.1 for Cloudflare's Workers, when bundling with Wrangler or ESbuild I get this error when importing `serveStatic` from `hono/cloudflare-workers`
```
✘ [ERROR] Could not resolve "__STATIC_CONTENT_MANIFEST"

    node_modules/hono/dist/adapter/cloudflare-workers/server-static-module.js:2:21:
      2 │ import manifest from "__STATIC_CONTENT_MANIFEST";
        ╵                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~

  You can mark the path "__STATIC_CONTENT_MANIFEST" as external to exclude it
  from the bundle, which will remove this error.

1 error
```
Even when I mark it as `external` and `exclude it form the bundle` When running the worker I get an error stating `Uncaught SyntaxError: Cannot use import statement outside a module`

yusukebe:
Hi @JeffBue,

Try using esbuild with this command:

```
esbuild --outdir=dist --external:__STATIC_CONTENT_MANIFEST --bundle ./src/index.ts
```

But it shouldn't throw an error when using Wrangler.

JeffBue:
> Even when I mark it as `external` and `exclude it form the bundle` When running the worker I get an error stating `Uncaught SyntaxError: Cannot use import statement outside a module`

Thank you @yusukebe for responding!
As noted above
> Even when I mark it as external and exclude it form the bundle When running the worker I get an error stating Uncaught SyntaxError: Cannot use import statement outside a module

Here's the command logs I get:

`With no --external flag` `error`
```
 ~/TCG_SITES/simple-renderer-base   main ±  npx esbuild app.js --bundle --outfile=dist/worker.js --loader:.html=text --loader:.md=text --loader:.csv=text --loader:.css=text --platform=node --watch --format=esm
✘ [ERROR] Could not resolve "__STATIC_CONTENT_MANIFEST"

    node_modules/hono/dist/adapter/cloudflare-workers/server-static-module.js:2:21:
      2 │ import manifest from "__STATIC_CONTENT_MANIFEST";
        ╵                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~

  You can mark the path "__STATIC_CONTENT_MANIFEST" as external to exclude it from the bundle, which
  will remove this error.

1 error
[...]

PR #1804, opened 2023-12-11
feat(cloudflare-workers): modified manifest option to be required

Code-Hex:

- [x] Add tests
- [x] Run tests

This is breaking change but fix: https://github.com/honojs/hono/issues/1127#issuecomment-1849001490

To resolve the issue, it may be better not to modify serveStatic (module) directly but to provide the wrapped version of serveStatic. However, I don't have any ideas on how to provide it in that case.

### DIFF
--- a/deno_dist/adapter/cloudflare-workers/server-static-module.ts
+++ b/deno_dist/adapter/cloudflare-workers/server-static-module.ts
@@ -1,5 +1,5 @@
-import type { ServeStaticOptions } from './serve-static'
-import { serveStatic } from './serve-static'
+import type { ServeStaticOptions } from './serve-static.ts'
+import { serveStatic } from './serve-static.ts'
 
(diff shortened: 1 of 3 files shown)


Code-Hex:
The denoify runs successfully in the local environment, but fails in CI...

yusukebe:
"Hi @Code-Hex,

I think this change is acceptable, but as I commented in https://github.com/honojs/hono/issues/1127#issuecomment-1849253502, it introduces a breaking change. So, we can include it in the next major release, `v4.0.0`.

yusukebe:
Hi @Code-Hex !

I've made `v4` branch for the X day. I want to merge this into it, so, please resolve the conflicts. Then, I'll merge!

yusukebe:
Duplicated #1984

PR #1806, opened 2023-12-11
fix(context): `c.json()` allows object and returns JSONParsed<T>

### LINKED ISSUE #1800: JSON response can't contain Dates

3.11.4

Cloudflare Pages Function

Return an object containing a Date with `context.json`. For example:

```
  const router = new Hono();

  router.get('/test', (context) => {
    const response = {
      dateValue: new Date()
    };

    return context.json(response);
  });
```

Because Hono uses JSON.stringify() under the hood, I would expect the code above to be valid.

### What do you see instead?

A type error is shown because `Date` is not considered a `JSONObject` or `JSONPrimitive`.

```
  Overload 1 of 2, '(object: JSONValue, status?: number | undefined, headers?: HeaderRecord | undefined): Response & TypedResponse<never>', gave the following error.
    Argument of type '{ dateValue: Date; }' is not assignable to parameter of type 'JSONValue'.
      Type '{ dateValue: Date; }' is not assignable to type 'JSONObject'.
        Property 'dateValue' is incompatible with index signature.
          Type 'Date' is not assignable to type 'JSONObject | JSONArray | JSONPrimitive'.
            Type 'Date' is not assignable to type 'JSONObject'.
              Index signature for type 'string' is missing in type 'Date'.
  Overload 2 of 2, '(object: JSONValue, init?: ResponseInit | undefined): Response & TypedResponse<never>', gave the following error.
    Argument of type '{ dateValue: Date; }' is not assignable to parameter of type 'JSONValue'.ts(2769)
```

### Additional information

The code executes fine, with the Date being stringified in the response as expected. For this reason especially, I consider the typing to be inaccurate. It disallows a valid type.

yusukebe:
Fixes #1800

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -6,7 +6,7 @@ import { serialize } from './utils/cookie.ts'
 import { resolveStream } from './utils/html.ts'
 import type { StatusCode } from './utils/http-status.ts'
 import { StreamingApi } from './utils/stream.ts'
-import type { JSONValue, InterfaceToType } from './utils/types.ts'
+import type { JSONValue, InterfaceToType, JSONParsed } from './utils/types.ts'
 
 type HeaderRecord = Record<string, string | string[]>
 type Data = string | ArrayBuffer | ReadableStream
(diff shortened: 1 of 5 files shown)


DeanMauro (approved):
Looks good to me.

I tested with
- Date
- Arrays of Date
- Nested objects with Date and varied types

Also tested BigInt and it correctly showed a type error.

yusukebe:
@DeanMauro Thank you for reviewing.

I'll merge it now.

PR #1809, opened 2023-12-12
docs(readme): update `hono/tiny` size

ryuapp:
Hono have grown.
I checked the size of `hono/tiny` with following the command.
```sh
npx esbuild --outdir=dist --bundle --minify ./src/preset/tiny.ts
```
### Result
```sh
npx esbuild --outdir=dist --bundle --minify ./src/preset/tiny.ts

  dist\tiny.js  13.9kb

⚡ Done in 12ms
```
Based on the above result, I update the size in readme.
Minimal app may exceed 14kb.

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ npm create hono@latest my-app
 ## Features
 
 - **Ultrafast** 🚀 - The router `RegExpRouter` is really fast. Not using linear loops. Fast.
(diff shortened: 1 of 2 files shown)


yusukebe:
@ryuapp 

You are right! Thanks, I'll merge it now.

PR #1817, opened 2023-12-13
chore: upgrading vitest to v1.0

watany-dev:
I've made some modifications to address an issue with nested tests, but could you please confirm if the changes are appropriate and aligned with the original intent of the tests?

```bash
FAIL  src/router/trie-router/node.test.ts > Duplicate param name > parent > get /posts/456
Error: Nested tests are not allowed
 ❯ src/router/trie-router/node.test.ts:481:7
    479|       expect(res[1][1]).toEqual({ id: '456' })
    480| 
    481|       it('get /posts/456', () => {
       |       ^
    482|         const [res] = node.search('get', '/posts/abc/comments/edf')
    483|         expect(res.length).toBe(1)
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -466,7 +466,6 @@
     "rimraf": "^3.0.2",
     "start-server-and-test": "^1.15.2",
     "supertest": "^6.3.3",
-    "ts-jest": "^29.1.1",
     "tsx": "^3.11.0",
     "typescript": "^5.3.3",
     "vite-plugin-fastly-js-compute": "^0.4.2",


tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "target": "ES2020",
     "declaration": true,
-    "moduleResolution": "Node",
+    "moduleResolution": "bundler",
```
yusukebe, line 5: `Bundler` is better.

yusukebe:
Hi @watany-dev 

Great PR!

> I've made some modifications to address an issue with nested tests, but could you please confirm if the changes are appropriate and aligned with the original intent of the tests?

Yeah. We can go with your changes.

I've left one comment and please resolve the conflict. Thanks!

tsconfig.json
watany-dev: fixed

PR #1823, opened 2023-12-14
feat(middleware/csrf): Introduce CSRF middleware

usualoma:
This middleware prevents CSRF attacks by checking request headers.

### All you have to do

When you need to protect against CSRF attacks (i.e. submitting with a form element) all you have to do is add the following one line

```ts
app.use('*', csrf())
```

This works well enough in most cases.
For "old browsers that do not send Origin headers" or "environments that use reverse proxies and drop Origin headers," it may not be possible to submit with false positives. If you want to support such environments, this middleware is not sufficient, and it is better to use CSRF tokens.

### Options

By default, the Origin header is compared to the requested URL. Generally, this works well enough, but sometimes an explicit specification is necessary for a variety of reasons. It is advisable to refer to the OWASP cheat sheet for the circumstances in which the specification is required and for notes on the contents of the specification.
https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#identifying-the-target-origin

In csrf middleware, as in cors middleware, "origin" can be specified as string, string[], or Function.

```ts
// string
app.use('*', csrf({origin: "myapp.example.com"}))

// string[]
app.use('*', csrf({origin: ["myapp.example.com", "development.myapp.example.com"]}))

// Function
// It is strongly recommended that the protocol be verified to ensure a match to `$`. You should *never* do a forward match.
app.use('*', csrf({origin: (origin) => /https:\/\/(\w+\.)?myapp\.example\.com$/.test(origin)})
```

### Thanks

This PR could not have been put together without the help of all those involved in the #1688 discussions. The documentation pointed to by @htunnicliff  and the Sveltekit implementation shared by @hjaber were especially helpful.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/csrf/index.ts
+++ b/src/middleware/csrf/index.ts
@@ -1,0 +1,48 @@
+import type { Context } from '../../context'
+import { HTTPException } from '../../http-exception'
+import type { MiddlewareHandler } from '../../types'
+
+type IsAllowedOriginHandler = (origin: string, context: Context) => boolean
+interface CSRFOptions {
+  origin?: string | string[] | IsAllowedOriginHandler
+}
+
+const isSafeMethodRe = /^(GET|HEAD)$/
+const isRequestedByFormElementRe =
+  /^\b(application\/x-www-form-urlencoded|multipart\/form-data|text\/plain)\b/
+
+export const csrf = (options?: CSRFOptions): MiddlewareHandler => {
+  const handler: IsAllowedOriginHandler = ((optsOrigin) => {
+    if (!optsOrigin) {
+      return (origin, c) => origin === new URL(c.req.url).origin
+    } else if (typeof optsOrigin === 'string') {
+      return (origin) => origin === optsOrigin
+    } else if (typeof optsOrigin === 'function') {
+      return optsOrigin
+    } else {
+      return (origin) => optsOrigin.includes(origin)
+    }
+  })(options?.origin)
+  const isAllowedOrigin = (origin: string | undefined, c: Context) => {
+    if (origin === undefined) {
+      // denied always when origin header is not present
+      return false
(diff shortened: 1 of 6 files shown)


usualoma:
@yusukebe 
Would you please review?

src/middleware/csrf/index.ts
```diff
@@ -0,0 +1,48 @@
...
+  return async function cors(c, next) {
+    if (
+      !isSafeMethodRe.test(c.req.method) &&
+      isRequestedByFormElementRe.test(c.req.header('content-type') || '') &&
+      !isAllowedOrigin(c.req.header('origin'), c)
+    ) {
+      const res = new Response('Forbidden', {
+        status: 403,
+      })
+      throw new HTTPException(401, { res })
```
yusukebe, line 43: I think that `401` should be `403`.

P.S.

However, if it remains as `401`, the tests do not fail. This is because if `{ res }` is set, it can be used and the status number in the first argument of `HTTPException` is ignored. Therefore, there is no problem for real use cases with this code. But, for making it the same as other middleware, I think it should be changed to `403`.

yusukebe:
Hi @usualoma!

Everything looks almost perfect! The tests cover all cases, great! I've left one comment. Please take a look.

usualoma:
@yusukebe Thank you!
Fixed typo in abb72bd.

yusukebe (approved):

yusukebe:
@usualoma 

Okay! It's a new feature, so let's release it in `v3.12.0` which is the next minor release!

yusukebe:
Sorry, I closed it by mistake." Merge into the "next" branch.

PR #1828, opened 2023-12-16
Use vitest for fastly compute testing

ysugimoto:

- [x] Add tests
- [x] Run tests

This PR improves Fastly Compute runtime testing. I've created vite plugin for [fastly-js-compute runtime](https://github.com/ysugimoto/vite-plugin-fastly-js-compute) and use it in this PR.

No longer need `jest` so removes from dependencies and now all hono testings will run only `vitest`.

Screen Shot 2023-12-17 at 8 24 07

Hope this PR accelerates hono testing.


yusukebe:
Hi @ysugimoto

Super great! This is exactly what we wanted to do.

There are two issues we need to address.

First, please format the code to pass CI:

```txt
yarn format:fix
```

Second, this test fails on Node.js 20.x:

SS

This failure is due to an issue with [vite-plugin-fastly-js-compute](https://github.com/ysugimoto/vite-plugin-fastly-js-compute). I've already created an issue for it:

https://github.com/ysugimoto/vite-plugin-fastly-js-compute/issues/1

Even after this issue is resolved and we use the fixed plugin, the following test will fail on Node.js 20.x because it doesn’t throw an error. Node.js 20.x includes `crypto` and `crypto.subtle.importKey`.

```ts
describe('JWT Auth Middleware does not work', () => {
  const app = new Hono()

  it('Should throw error', () => {
    expect(() => {
      app.use('/jwt/*', jwt({ secret: 'secret' }))
    }).toThrow(/`crypto.subtle.importKey` is undefined/)
  })
})
```

We may need to update the test to support this condition and add a CI that runs on Node.js 20.x, not just 18.x. I’ll work on it once this PR is merged.

ysugimoto:
@yusukebe Thanks for the quick review!

1. I have done to run `yarn format:fix`
2. Skip `JWT Auth Middleware does not work` tests in nodejs 20 environment.

Just to explain about 2, nodejs 20 has a native WebCrypto module as global and has many features.
it would be fine but not fine in Fastly Compute runtime.

Fastly Runtime does not have WebCrypto features completely so the above test result is different in nodejs and Fastly Compute runtime (due to nodejs crypto having `crypto.subtle.importKey` function)

Therefore, I mark whether this test should skip or not by checking `__fastlyComputeNodeDefaultCrypto` field, which comes from `vite-plugin-fastly-js-compute` - whether crypto is polyfill-ed or not.

If you have any unclear points or any other suggestions, let me know.

yusukebe:
"@ysugimoto

> Therefore, I mark whether this test should skip or not by checking the `__fastlyComputeNodeDefaultCrypto` field, which comes from `vite-plugin-fastly-js-compute` - to determine whether crypto is polyfilled or not.

This makes sense. You are right, we have to consider that Fastly Compute's `crypto` does not have all the features of Node's `crypto`, and using `__fastlyComputeNodeDefaultCrypto` is a smart idea.

Just one thing. Please run `yarn lint:fix` to fix the lint error. After that, I'll merge it.

ysugimoto:
Done to run `yarn lint:fix` 👍

yusukebe:
Thanks! Merging now.

PR #1831, opened 2023-12-18
chore(utils): Move cf module

watany-dev:
I've been wondering about this for a while, so I checked `utils/cloudflare` and realized that it would be better to put it as ` adapter/cloudflare-workers`. 

If sharing with pages etc. is in the roadmap, this Refactor is not needed. 

- [ ] Add tests
- [x] Run tests


yusukebe:
Hi @watany-dev 

This refactoring is good! It is used only for Cloudflare Workers; Pages does not use it. So we can go with it. Can you fix the error? Then, I'll merge it.

watany-dev:
@yusukebe 
Appreciate the review. Yes, I have fixed it, including the lint check.

yusukebe (commented):
@watany-dev 

Sorry! One more thing.

I think the file name `utils.ts` is better than `cloudflare.ts`. So how about renaming `cloudflare.ts` to `utils.ts` and `cloudflare.test.ts`?

watany-dev:
@yusukebe 
I tried Rename. Is this OK?

yusukebe (approved):

yusukebe:
Okay! Let's go with it!

PR #1841, opened 2023-12-22
feat(dev): add `getRouterName()`

yusukebe:
This PR introduces `getRouterName()` function in `hono/dev` helper and deprecates `app.routerName()`.

I think `app.routerName()` is not used by many users, and if users use it, the purpose is debugging in most cases. So, we can place it in `hono/dev` is better.

Usage:

```ts
import { Hono } from 'hono'
import { getRouterName } from 'hono/dev'

const app = new Hono()

//...

console.log(getRouterName(app))
```

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/deno_dist/helper/dev/index.ts
+++ b/deno_dist/helper/dev/index.ts
@@ -60,3 +60,8 @@ export const showRoutes = <E extends Env>(hono: Hono<E>, opts?: ShowRoutesOption
       })
     })
 }
+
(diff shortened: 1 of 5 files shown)


yusukebe:
Hi @usualoma !

How do you think this PR?

usualoma:
I think you are right 👍

yusukebe:
@usualoma 

Thanks! I'll include this feature in the next minor version.

Issue #1843, opened 2023-12-22
Move `app.request()`, `c.stream()` and `c.streamText()`


yusukebe:
### What is the feature you are proposing?

For Hono, being small is as important as being fast.

In Cloudflare Workers, some benchmarks show that the smaller the size, the faster it is under certain conditions.

https://github.com/TigersWay/cloudflare-playground

This is an extreme example, but smaller is better.

Currently, the `hono/tiny` preset Hello World application is 13.04 KB after minify; the goal is to get it down to 10 KB in v4.

```txt
yusuke $ wrangler deploy --minify --dry-run src/tiny.ts
 ⛅️ wrangler 3.22.1
-------------------
--dry-run: exiting now.
Total Upload: 13.04 KiB / gzip: 5.05 KiB
```

I will do some minor refactoring, such as making common strings variables. However, to make it smaller, I would like to move `app.request()` to `hono/dev` and `c.stream()` and `c.streamText()` to `hono/streaming`.

### `app.request()`

There is not a lot of code in `app.request()`, but it does not have to be in `hono-base.ts`. This is because this function is mainly used for testing. So I propose to move it to `hono/dev`.

```ts
import { request } from 'hono/dev'

//... 

const res = await request(app, '/')
```

### `c.stream()` and `c.streamText()`

The `c.stream()` and `c.streamText()` functions are recently created and very attractive. However, they are used less often than `c.text()`, `c.json()`, and `c.html()`, and have slightly more code. If deleted, the size will be reduced by less than 1KB.

```txt
yusuke $ wrangler deploy --minify --dry-run src/tiny.ts
 ⛅️ wrangler 3.22.1
-------------------
--dry-run: exiting now.
Total Upload: 12.20 KiB / gzip: 4.76 KiB
```

Therefore, I propose to move them to `hono/streaming`.

```ts
import { stream } from 'hono/streaming'

app.get('/stream', (c) => {
  return stream(c, async (stream) => {
    // ...
  })
})
```

---

Since these are major breaking changes, I plan to deprecate the previous functions in the next minor update and remove them in v4. Breaking changes are painful for us, but one of the most important things to do is to keep it small, and if there are no major problems, we have to do it.

yusukebe:
@sor4chi If you agree, can you create a PR to move `c.stream()` and `c.streamText()`?

@hagishi If you look at it and agree, can you create a PR to move the `app.request()`?

watany-dev:
[...]

PR #1846, opened 2023-12-23
fix: move `c.stream*` to helper

sor4chi:
ref: #1843 

- [x] Add tests
- [x] Run tests


yusukebe:
@sor4chi 

Is this ready to review?

sor4chi:
@yusukebe Yes, please.

src/context.ts
```diff
@@ -375,6 +375,9 @@ export class Context<
     return this.newResponse(null, status)
   }
 
+  /** @deprecated
+   * Use `streamText()` in `hono/helper/streaming` instead of `c.streamText()`. The `c.streamText()` will be removed in v4.
```
yusukebe, line 379: You can import it from `hono/streaming`. So, `hono/helper/streaming` should be `hono/streaming`.

src/context.ts
```diff
@@ -387,6 +390,9 @@ export class Context<
     return this.stream(cb, arg, headers)
   }
 
+  /** @deprecated
+   * Use `stream()` in `hono/helper/streaming` instead of `c.stream()`. The `c.stream()` will be removed in v4.
```
yusukebe, line 394: Same as above.

src/helper/streaming/index.ts
```diff
@@ -1,59 +1,21 @@
-import type { Context } from '../../context'
+import type { Context, HeaderRecord } from '../../context'
+import type { StatusCode } from '../../utils/http-status'
 import { StreamingApi } from '../../utils/stream'
 
-interface SSEMessage {
-  data: string
-  event?: string
-  id?: string
+export const stream = (
```
yusukebe, line 5: We can make it simpler:

```ts
export const stream = (
  c: Context,
): Response => {
  const { readable, writable } = new TransformStream()
  const stream = new StreamingApi(writable)
  cb(stream).finally(() => stream.close())
  return c.newResponse(readable)
}
```

Users can set the status and the headers with `c.status()` and `c.header(),` so I think it does not need `arg?` and `headers?`. `steamSSE()` also follows this approach.

src/helper/streaming/text.ts
```diff
@@ -0,0 +1,18 @@
+import type { Context, HeaderRecord } from '../../context'
+import { TEXT_PLAIN } from '../../context'
+import type { StatusCode } from '../../utils/http-status'
+import type { StreamingApi } from '../../utils/stream'
+import { stream } from '.'
+
+export const streamText = (
```
yusukebe, line 7: Same as above:

```ts
export const streamText = (
  c: Context,
): Response => {
  c.header('content-type', TEXT_PLAIN)
  c.header('x-content-type-options', 'nosniff')
  c.header('transfer-encoding', 'chunked')
  return stream(c, cb)
}
```

yusukebe (commented):
@sor4chi 

Thank you for the PR. Take a look the comments!

sor4chi:
Hi, @yusukebe 
Thanks, I fixed.

yusukebe (approved):
LGTM

PR #1850, opened 2023-12-24
feat: introduce css Helper

yusukebe:
This PR introduces middleware to enable super simple CSS in JS features.

Usage:

```tsx
import { Hono } from 'hono'
import { css, renderStyles } from 'hono/css'
import { jsxRenderer } from 'hono/jsx-renderer'

const app = new Hono()

app.get(
  '*',
    return (
      <html>
        <head>
          <style>{renderStyles()}</style>
        </head>
        <body>{children}</body>
      </html>
    )
  })
)

app.get('/', (c) => {
  const HeaderClass = css`
    background-color: blue;
    color: white;
    padding: 1rem;
  `
  return c.render(
    <html>
      <body>
        <h1 class={HeaderClass}>Hello!</h1>
      </body>
    </html>
  )
})

export default app
```

Tips: If you use VS Code, you can use [vscode-styled-components](https://marketplace.visualstudio.com/items?itemName=styled-components.vscode-styled-components) for Syntax highlighting and IntelliSense for `css` tagged literals.

It's nice that the code is short, but there might be some things that need to be considered.

- [ ] Add tests
- [ ] Run tests


yusukebe:
Hi @usualoma ! What do you think about this feature?

toinbis:
Don't overthink it! Please ship it, let us try it and then we'll see where to go from there! :)

usualoma:
### Cons 

I did not mention the cons of this router, so I will add them here.

#### The regex generated is large

The RegExpRouter compiled the regexes to be as small as possible, but the FilePatternRouter does not make that effort, so they become large for many routing registrations.

For example

* `app.get('/users/:id', handlerA)`
* `app.get('/users/:id/posts', handlerB)`
* `app.get('/users/:id/posts/:post_id', handlerC)`

RegExpRouter: `/users/([^/]+)(? :()|(? :/posts(? :()|/([^/]+)()))`
FilePatternRouter: `/users/([^/]+)/posts/([^/]+)()||/users/([^/]+)/posts()||/users/([^/]+)())`

The current implementation is a PoC, so no effort has been made to make it smaller, but I think it could be made smaller by adding a few lines and allowing for a little overhead. Additional implementation would be necessary depending on the number of routings envisioned.

usualoma:
@yusukebe 

Very cool idea! We should merge the PRs as it can provide very good DX when used with vscode-styled-components. I think we need to make some modifications and I'm putting it together. Please wait a bit.

usualoma:
@yusukebe 

It appears to me that two major modifications are needed. I have no idea how to fix them yet.

#### Sometimes `renderStyles()` is called first

Sometimes `renderStyles()` is called first when the child component `css` is used. This is even more likely to occur in asynchronous components.

#### Context

I think the current implementation would mix the styles of multiple requests. This may be, though, that the current behavior is the correct specification.

#### Test Examples

With current HEAD, I think the following tests will fail

```ts
describe('child component', () => {
  const app = new Hono()
  const Component = () => {
    const HeaderClass = css`
      background-color: blue;
      color: white;
      padding: 1rem;
    `
    return `<h1 class=${HeaderClass}>Hello!</h1>`
  }
  app.get('/', (c) => {
    return c.text(`<style>${renderStyles()}</style>${Component()}`)
  })

  it('Should render CSS styles', async () => {
    expect(res.status).toBe(200)
    const html = await res.text()
    expect(html).toBe(
      '<style>.css-4167396862 { background-color: blue; color: white; padding: 1rem; }</style><h1 class=css-4167396862>Hello!</h1>'
    )
  })
})

describe('context', () => {
  const app = new Hono()
  app.get('/white', (c) => {
    const HeaderClass = css`
      background-color: blue;
      color: white;
      padding: 1rem;
    `
    return c.text(`<style>${renderStyles()}</style><h1 class=${HeaderClass}>Hello!</h1>`)
  })
  app.get('/black', (c) => {
    const HeaderClass = css`
      background-color: blue;
      color: black;
      padding: 1rem;
    `
    return c.text(`<style>${renderStyles()}</style><h1 class=${HeaderClass}>Hello!</h1>`)
  })

  it('Should render CSS styles only for /black', async () => {
    await app.request('/white')
    const res = await app.request('/black')
    expect(res.status).toBe(200)
    const html = await res.text()
    expect(html).toBe(
      '<style>.css-1147732022 { background-color: blue; color: black; padding: 1rem; }</style><h1 class=css-1147732022>Hello!</h1>'
    )
  })
})
```

yusukebe:
Just idea.

Perhaps, we may make it if treat as a feature of JSX components.

yusukebe:
@usualoma 

How about this, though? We have to consider whether it should be included in the JSX Renderer.

#1855

usualoma:
@yusukebe 

Thanks for reminding me about #1780. I'm glad to hear it.

However, #1780 does not support asynchronous, considering the features we need to implement this time, I think it may not be the best option.

I would like to be able to get the expected results with all of the following patterns

```ts
const Header = async () => {
  const HeaderClass = css`
    background-color: blue;
    color: white;
    padding: 1rem;
  `
  return <h1 class={HeaderClass}>Hello!</h1>
}

const AsyncHeader = async () => {
  await new Promise((r) => setTimeout(r, 1000))
  const HeaderClass = css`
    background-color: red;
    color: white;
    padding: 1rem;
  `
  return <h1 class={HeaderClass}>Hello!</h1>
}

const SuspencedHeader = async () => {
  await new Promise((r) => setTimeout(r, 1000))
  const HeaderClass = css`
    background-color: green;
    color: white;
    padding: 1rem;
  `
  return <h1 class={HeaderClass}>Hello!</h1>
}

app.get('/', (c) => {
  return c.render(
    <html>
      <body>
        <Header />
        <AsyncHeader />
        <Suspence fallback={<p>Loading...</p>}>
          <SuspencedHeader />
        </Suspence>
      </body>
    </html>
  )
})
```

I have an idea for implementation, let me suggest it later.

yusukebe:
@usualoma 

> I have an idea for implementation, let me suggest it later.

Okay! I'll be waiting.

usualoma:
@yusukebe 
I have implemented #1858 based on #1850. What do you think of this approach?

usualoma:
How far should we go in implementing this feature?
I would suggest that pseudo-elements, pseudo-selectors, and `@keyframes` should be able to be written the same way as styled-components.

https://github.com/usualoma/hono/pull/4

yusukebe:
@usualoma 

I've left the comment on #1858!

yusukebe:
Let's land it!

PR #1851, opened 2023-12-24
Refactor(test): update vitest settings

watany-dev:
I have made these settings. 

~~1. Handling Vitest's type checking. This makes tsc --no-emit unnecessary.
https://vitest.dev/config/#typecheck~~
2. I don't think the Vitest configuration file is needed in tsconfig.build. I have listed similar items that I think are unnecessary, but please point out if I am mistaken.
3.  Update from Vitest 1.04 to 1.10.

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -450,7 +450,7 @@
     "@types/node": "^20.8.2",
     "@types/node-fetch": "^2.6.2",
     "@types/supertest": "^2.0.12",
-    "@vitest/coverage-v8": "^1.0.4",
+    "@vitest/coverage-v8": "^1.1.0",
     "arg": "^5.0.2",
     "crypto-js": "^4.1.1",
     "denoify": "^1.6.6",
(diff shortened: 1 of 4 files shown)


yusukebe:
Hi @watany-dev 

Thanks!

I've tried Vite's type checking before, but it did not work well, and also, with this PR, it doesn't catch the type errors.

Screenshot 2023-12-24 at 15 52 19

I don't know if this setting is right or not. Anyway, if it can not check types, we have to keep using `tsc`.

watany-dev:
@yusukebe 
Thank you for the review. Upon rechecking the typecheck tests for Vitest, it appears that '*.test-d.ts' is the scope, and it's not a complete substitute for tsc. I have reverted the changes.

yusukebe:
> 2. I don't think the Vitest configuration file is needed in tsconfig.build. I have listed similar items that I think are unnecessary, but please point out if I am mistaken.

We don't have to modify `tsconfig.build.json` because it is just for the build phase used in `build.ts`. Vitest uses `tsconfig.json`.

yusukebe:
@watany-dev 

Thanks. We can merge this!

Issue #1812, opened 2023-12-13
@typescript-eslint/promise-function-async breaks html


Code-Hex:
### What version of Hono are you using?

3.9.2

### What runtime/platform is your app running on?

Cloudflare Workers

### What steps can reproduce the bug?

update to v3.10.0 and use @typescript-eslint/promise-function-async linter.

For example here is original code.
```ts
return c.html(html`${Component()}`)

const Component: FC = () => { ... }
```

then update to v3.10.0 from v3.9.2, @typescript-eslint/promise-function-async changes to
```ts
return c.html(html`${Component()}`)

const Component: FC = async () => { ... }
```

The template in the string is not updated to await at this time. The result embedded thereby becomes an object of the Promise and the following screen is displayed.

<img width="774" alt="スクリーンショット 2023-12-13 10 33 31" src="https://github.com/honojs/hono/assets/6500104/fcacd9c2-5532-4163-9b86-a1838bcfcae3">

There are many such codes in many places, and it would be difficult to correct each one.

### What is the expected behavior?

I want to be able to handle Promise objects in a nice way, even if I don't await them in the embedding of template strings.

### Additional information

My twitter
https://x.com/codehex/status/1734727289580208301?s=20

yusukebe:
Hi @Code-Hex 

Is this fixed?

Code-Hex:
@yusukebe No
I expect to make it possible to accept a Promise object in TypeScript’s template literal tag, handling it within the html function to assemble HTML

Can we do that?

usualoma:
@Code-Hex @yusukebe 

I think this is better handled by `html`, which is a tagFunction, in my opinion.
Not only for this issue, but the following are generally expected results for `html`. I will create a PR later, so please review.

```diff
diff --git a/src/helper/html/index.test.ts b/src/helper/html/index.test.ts
index 0d18c44f..31615c18 100644
--- a/src/helper/html/index.test.ts
+++ b/src/helper/html/index.test.ts
@@ -28,6 +28,24 @@ describe('Tagged Template Literals', () => {
       '<p>Name:John &quot;Johnny&quot; Smith Contact:<a href="http://example.com/">My Website</a></p>'
     )
   })
+
+  describe('Promise', () => {
+    it('Should return Promise<string> when some variables contains Promise<string> in variables', async () => {
+      const name = Promise.resolve(`John "Johnny" Smith`)
+      const res = html`<p>I'm ${name}.</p>`
[...]

PR #1858, opened 2023-12-26
feat(css): Implement css helper with callback

usualoma:
### What kind of implementation is this?

The callback mechanism used in Suspense and ErrorBoundary is used to insert the class name.
It is primarily targeted to be used in JSX components, but can also be used in `html` tag functions.
It does not depend on jsxRenderer.

### Pros

* It does not depend on jsxRenderer, so you can use any pattern you like.
* In environments where initialization does not occur with every request, It works very fast because CSS is not recalculated when class names are predefined globally.

### Cons

* The mechanism is complicated because the string is always a `Promise<string>` and callback must be properly called to obtain the final result. (However, if you use c.html(), you do not need to be aware of this complexity.)
* Allowing Promise<string> in the `class` attribute may confuse some linters (e.g. https://github.com/honojs/hono/issues/1812 ).
    * https://github.com/honojs/hono/compare/feat/css-helper...usualoma:hono:feat/css-helper?expand=1#diff-eea0d4f8a89b330f26eea854ef087c7d92cfca29268164f8ee315f770e4be7e8L26-R26

### Demo

```ts
import { Hono } from './src'
import { jsx, Fragment } from './src/jsx'
import { Suspense } from './src/jsx/streaming'
import { jsxRenderer } from './src/middleware/jsx-renderer'
import { css, Style } from './src/helper/css'

const app = new Hono()

app.get(
  '*',
  jsxRenderer(
    ({ children }) => {
      return (
        <html>
          <head>
            <Style />
          </head>
          <body>{children}</body>
        </html>
      )
    },
    { stream: true }
  )
)

const Header = () => {
  const HeaderClass = css`
    background-color: blue;
    color: white;
    padding: 1rem;
  `
  return <h1 class={HeaderClass}>Hello!</h1>
}

const AsyncHeader = async () => {
  await new Promise((r) => setTimeout(r, 100))
  const HeaderClass = css`
    background-color: red;
    color: white;
    padding: 1rem;
  `
  return <h1 class={HeaderClass}>Hello!</h1>
}

const Loading = () => {
  const HeaderClass = css`
    background-color: gray;
    color: white;
    padding: 1rem;
  `
  return <h1 class={HeaderClass}>Loading...</h1>
}

const SuspensedHeader = async () => {
  await new Promise((r) => setTimeout(r, 2000))
  const HeaderClass = css`
    background-color: green;
    color: white;
    padding: 1rem;
  `
  return <h1 class={HeaderClass}>Hello!</h1>
}

const GlobalClass = css`
  background-color: pink;
  color: black;
  padding: 1rem;
`
const GlobalHeader = async () => {
  return <h1 class={GlobalClass}>Hello!</h1>
}

app.get('/', (c) => {
  return c.render(
    <>
      <Header />
      <AsyncHeader />
      <GlobalHeader />
      <Suspense fallback={<Loading />}>
        <SuspensedHeader />
      </Suspense>
    </>
  )
})

export default app
```

output
```html
<html><head><style id="hono-css">.css-4167396862{background-color: blue; color: white; padding: 1rem;}.css-1922077261{background-color: red; color: white; padding: 1rem;}.css-3997042484{background-color: pink; color: black; padding: 1rem;}.css-3971143725{background-color: gray; color: white; padding: 1rem;}</style></head><body><h1 class="css-4167396862">Hello!</h1><h1 class="css-1922077261">Hello!</h1><h1 class="css-3997042484">Hello!</h1><template id="H:1"></template><h1 class="css-3971143725">Loading...</h1><!--/$--></body></html><script>document.querySelector('#hono-css').textContent+=".css-1873676827{background-color: green; color: white; padding: 1rem;}"</script><template><h1 class="css-1873676827">Hello!</h1></template><script>
((d,c,n) => {
c=d.currentScript.previousSibling
d=d.getElementById('H:1')
if(!d)return
do{n=d.nextSibling;n.remove()}while(n.nodeType!=8||n.nodeValue!='/$')
d.replaceWith(c.content)
})(document)
</script>
```

https://github.com/honojs/hono/assets/30598/bf08ddb6-9dd3-40a2-9c22-2bba255001ff

- [x] Add tests
- [x] Run tests


yusukebe:
Hey @usualoma !

This is cool! This implementation can cover all use cases! 

> The mechanism is complicated because the string is always a `Promise<string>` and callback must be properly called to obtain the final result.
> Allowing Promise<string> in the `class` attribute may confuse some linters.

I think these are acceptable. Using `Promise` and callback is the only way I can think of to do it, so that's fine. My one concern is the complexity of the implementation, but I think the `css` feature is worth it.

As you mentioned this commend: https://github.com/honojs/hono/pull/1850#issuecomment-1869259251, I think we have to implement pseudo-elements, pseudo-selectors, and `@keyframes`.

I tried it and found one problem: with `stream` turned off, the style tag is empty when accessed a second time.

https://github.com/honojs/hono/assets/10682/902b9aa4-65f8-49fd-a214-748d25d3b358

usualoma:
@yusukebe Thank you for your comment!

> I tried it and found one problem: with stream turned off, the style tag is empty when accessed a second time.

Ah, there was one parameter missing. Fixed in dd92903

https://github.com/usualoma/hono/pull/4 has been merged with this PR.

So, can we proceed to work on this PR, writing tests and tidying up the code so that we can merge it into #1850?

yusukebe:
@usualoma 

> So, can we proceed to work on this PR, writing tests and tidying up the code so that we can merge it into #1850?

Yes, please!

usualoma:
@yusukebe Improved CSS helper by fd2ea73 and 5f0ba8b
How about this?

## Changes


### Pseudo selectors

There's both good news and bad news.

* Bad News: The selector for pseudo-elements using '&' has been removed due to its overly complex implementation.
* Good News: Pseudo-elements using '&' are now almost universally supported by browsers.
    * [caniuse.com/css-nesting](https://caniuse.com/css-nesting)

I believe that we should start with a small release as an MVP (Minimum Viable Product) to evaluate its value. Therefore, if it works in the latest browsers, it would be a wise choice to consider it not supported here.


### `createCssContext()`

Default `css` and `Style` are also provided, but you can create your own using `createCssContext()` if you want to use them in a context separate from this one.

The test code will show you how to use it.
https://github.com/honojs/hono/pull/1858/files#diff-b13e9a99ad47b9364486431c602062116158762edc7a647fa557b8c80a76774cR181-R203

### escape variable

Variables embedded in `${}` are now escaped. This is the same as html`<h1>${'(><;) crying...</h1>` being escaped, which is a satisfactory specification in the hono. However, I feel a little uneasy about the difference from there, since it seems that styled-components does not escape.

```ts
css`
    animation: 1s ${animation} ease-in-out;
    background-color: green;
    color: white;
    padding: 1rem;
    transition: opacity 0.1s ease-in-out;

    &::after {
      margin-left: 1rem;
      content: "${`I'm a hono!`}"; // will be escaped
    }

    &:hover {
      opacity: 0.5;
    }
  `
```


### Embed sub `css`

This was added in the last e4ceb1e. This feature allows a very large possibility to generate CSS programmatically.
Nevertheless, it may be overkill. I think we can revert this commit if it is not needed.

```ts
const FizzBuzz = () => {
  const list = [...Array(20)].map((_, i) => i + 1)
  const className = css`
    ${list.map((n) => {
      const fizz = n % 3 === 0
      const buzz = n % 5 === 0
      return css`
        :nth-child(${n}):after {
          margin-left: 1rem;
          content: '${fizz && buzz ? 'FizzBuzz' : fizz ? 'Fizz' : buzz ? 'Buzz' : ''}';
        }
      `
    })}
  `

  return (
    <ul class={className}>
      {list.map((n) => (
        <li>{n}</li>
      ))}
    </ul>
  )
}

app.get('/', (c) => {
  return c.render(<FizzBuzz />)
})
```

image

image

yusukebe:
@usualoma 

I'm trying to use it now. Please wait a moment!

usualoma:
An immediate, responsive LIKE button without JavaScript (delayed, maybe debouncing and then sending the pressed count via JS) I think it's interesting stuff, but I think there are only a limited number of situations where it has value.

https://github.com/honojs/hono/assets/30598/3594d4e1-6dc3-49ee-be7b-b6ec812dada2

```ts
const LikeButton = () => {
  const list = [...Array(100)].map((_, i) => i + 1)
  const animations: Promise<String>[] = [
    keyframes`
    // even
    0% {
      transform: scale(1);
    }
    50% {
      transform: scale(1.2);
    }
    100% {
      transform: scale(1);
    }
  `,
    keyframes`
      // odd
      0% {
        transform: scale(1);
      }
      50% {
        transform: scale(1.21);
      }
      100% {
        transform: scale(1);
      }
    `,
  ]

  const className = css`
    font-size: 30px;
    position: relative;
    display: inline-block;
    label {
      position: absolute;
      width: 100%;
      height: 100%;
    }
    input {
      display: none;
    }
    span {
      pointer-events: none;
      position: relative;
    }
    ${list.map(
      (n) => css`
        input#id-${n}:checked ~ label[for='id-${n}'] {
          pointer-events: none;
        }
        ${[...Array(n)].map(() => 'input:checked ~').join(' ')} span {
          &:after {
            animation: 0.3s ${animations[n % 2]} ease-in-out;
            border-radius: 50%;
            background: red;
            color: white;
            position: absolute;
            font-size: 10px;
            display: inline-block;
            min-width: 13px;
            text-align: center;
            top: -10px;
            right: -10px;
            padding: 3px;
            content: '${n}';
          }
        }
      `
    )}
  `

  return (
    <div class={className}>
      {list.map((n) => (
        <input id={`id-${n}`} type='checkbox' />
      ))}
      {list.map((n) => (
        <label for={`id-${n}`} />
      ))}
      <span>👍</span>
    </div>
  )
}

app.get('/', (c) => {
  return c.render(<LikeButton />)
})
```

usualoma:
The following patch is required to execute the above code.

```diff
commit 33d158734c8e0ea412703e2e7bb9bf0067c9e5fd
Author: Taku Amano <taku@taaas.jp>
Date:   Wed Dec 27 18:00:39 2023 +0900

    fix(css): propagate selector to parent class

diff --git a/src/helper/css/index.ts b/src/helper/css/index.ts
index 00ede444..c0cdcf17 100644
--- a/src/helper/css/index.ts
+++ b/src/helper/css/index.ts
@@ -4,6 +4,8 @@ import type { HtmlEscapedCallback, HtmlEscapedString } from '../../utils/html'
 type CssClassName = HtmlEscapedString & {
   isCssClassName: true
   styleString: string
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  selectors: String[]
 }
 
 type CssEscaped = {
@@ -108,6 +110,7 @@ const buildStyleString = async (
         styleString += ` ${value.substring(11)} `
       } else {
         if ((value as CssClassName).isCssClassName) {
+          selectors.push(...(value as CssClassName).selectors)
           value = (value as CssClassName).styleString
         } else if (!(value as CssEscapedString).isCssEscaped && /([\\"'\/])/.test(value)) {
           value = value.replace(/([\\"'\/])/g, '\\$1')
@@ -138,7 +141,6 @@ export const createCssContext = ({ id }: { id: Readonly<string> }) => {
     const selectors: String[] = []
     const styleString = await buildStyleString(strings, values, selectors)
     const selector = new String(toHash(styleString))
-    selectors.push(selector)
 
     const appendStyle: HtmlEscapedCallback = ({ buffer, context }): Promise<string> | undefined => {
       const [toAdd, added] = contextMap.get(context) as usedClassNameData
@@ -179,7 +181,7 @@ export const createCssContext = ({ id }: { id: Readonly<string> }) => {
       }
       const [toAdd, added] = contextMap.get(context) as usedClassNameData
       let allAdded = true
-      selectors.forEach((className) => {
+      ;[selector, ...selectors].forEach((className) => {
         if (!added.has(className)) {
           allAdded = false
           toAdd.set(className, true)
@@ -196,6 +198,7 @@ export const createCssContext = ({ id }: { id: Readonly<string> }) => {
       isEscaped: true,
       isCssClassName: true,
       styleString,
+      selectors,
       callbacks: [addClassNameToContext],
     })
 
@@ -213,6 +216,7 @@ export const createCssContext = ({ id }: { id: Readonly<string> }) => {
       isEscaped: true,
       isCssClassName: true,
       styleString,
+      selectors: [],
     })
     return className
   }
```

usualoma:
Oh, I just remembered that I need to write one thing.

I'll leave `Embed sub css` for once and talk about the rest of the implementation.

The current implementation could be better, so I'm sure there are more efficient ways to use memory and minify. However, these improvements can only be made after we decide this feature is worthwhile. We believe the current implementation is of sufficient quality to be tested as an MVP.

usualoma:
### For whom is this a feature?

I think this is a feature not only for CSS guru but also for design tool guru who are not familiar with CSS. This feature can be used when exporting components from design tools.

https://github.com/honojs/hono/assets/30598/8fc459ca-ac24-48f3-a4d0-cb01dea19871

usualoma:
Nevertheless, I am not sure if this is the scope of the issue that hono should cover. I think some would argue that we should concentrate on smaller areas only.

yusukebe:
Hi @usualoma 

Super cool ramen example! I love it!

Please give me some time to continue to try it out.

I have tried to rewrite pages created with Tailwind CSS using hono/css. It is very user-friendly. Greate. For example, I wrote the following code:

```tsx
import dayjs from 'dayjs/esm'
import relativeTime from 'dayjs/esm/plugin/relativeTime'
import { css } from 'hono/css'
import { createRoute } from 'honox/factory'
import Time from '../../components/time'
import { findArticleById } from '../../db'
import { parseMarkdown } from '../../utils'

dayjs.extend(relativeTime)

const articleClass = css`
  background-color: #fff;
  padding: 1.5rem;
  margin-bottom: 1rem;
  border-radius: 0.5rem;
  box-shadow: 0px 4px 8px rgba(0, 0, 0, 0.1);
`

const titleClass = css`
  color: rgb(31 41 55);
  font-weight: 800;
  font-size: 2rem;
`

export default createRoute(async (c) => {
  const { id } = c.req.param()
  const article = await findArticleById(c.env.DB, id)

  if (!article) {
    return c.notFound()
  }

  const parsedContent = parseMarkdown(article.content)

  return c.render(
    <div
      class={css`
        margin-top: 2rem;
      `}
    >
      <article class={articleClass}>
        <header>
          <h2 class={titleClass}>{article.title}</h2>
          <Time created_at={article.created_at}>{dayjs(article.created_at).format('YYYY-MM-DD HH:mm:ss')}</Time>
        </header>
        <section
          class={css`
            margin-top: 1.5rem;
          `}
        >
          <div id="contents" dangerouslySetInnerHTML={{ __html: parsedContent }} />
        </section>
      </article>
    </div>
  )
})
```

The interesting thing is we can make it a variable and make it a component, or we can make it a utility class like Tailwind CSS. We can also write `css` directly in the class attribute. This would be very useful when creating web pages in Hono.

usualoma:
> Please give me some time to continue to try it out.
👍 

Several bug fixes have been pushed.

yusukebe:
@usualoma 

How can we mixin class names, like this in HTML?

```html
<h1 class="title h1">Hello!</h1>
```

For example, allow an attribute to have `Promise<string>[]` as its value?

```tsx
const classH1 = css`
  font-size: 2rem;
`

const classTitle = css`
  color: blue;
`

app.get('/', (c) => {
  return c.html(
    <html>
      <Style />
      <body>
        <h1 class={[classTitle, classH1]}>Hello!</h1>
      </body>
    </html>
  )
})
```

usualoma:
@yusukebe

I think that is an idea that should be considered as well! But I didn't include it in this PR because there are a few issues.

### Different from CSS semantics.

I don't see the problem with the example you showed, but I am concerned about the fact that it differs from CSS semantics when overriding a property.

In CSS, `class="btn btn-primary"` and `class="btn-primary btn"` are the same, but here `class={[btn, btnPrimary]}` and `class={[ btnPrimary, btn]}` will give different results.

It's not a bad thing that "the order makes sense", in fact it's a good thing, but...

```ts
const button = css`
  border-radius: 10px;
`
const myPrimaryButton = css`
  ${btn}
  background: blue;
`
<button class={myPrimaryButton} />
```

would be more explicit and clear about inheritance.

Hmmm, but in the case of your example, it would certainly be better to be able to specify it in an array rather than going to the trouble of generating a new class. 

```
css={css`${classTitle} ${classH1}`}
```

This would do the same thing, but I think it would be a much better experience to be able to specify it as an array.

I think the most "CSS in JS" way to do this in existing CSS in JS products is to generate new class names. I remember this also works for "emotion".

```
const myPrimaryButton = css`.
  ${btn}
  background: blue;.
`
```

However, if we are of the belief that "hono should be able to specify in an array", we are willing to implement it.

What do you think?

usualoma:
@yusukebe 
Sorry, I left out the premise of the discussion, the expected results. I would expect the following results.

```ts
const class1 = css`props1` // => .hono-class-name-1
const class2 = css`props2` // => .hono-class-name-2
renderToString(<div class={[class1, class2]} />) // <div class="hono-new-generated-class-name" />
```

We could consider an implementation that would result in the following, but we don't think this would accomplish one of the things I wanted to solve with JSS in CSS: freedom from the hell of CSS cascading and selctor.

```ts
const class1 = css`props1` // => .hono-class-name-1
const class2 = css`props2` // => .hono-class-name-2
renderToString(<div class={[class1, class2]} />) // <div class="hono-class-name-1 hono-class-name-2" />
```

yusukebe:
> I think that is an idea that should be considered as well! But I didn't include it in this PR because there are a few issues.

Yes, I also think it should not be included in this PR if there are issues! However, I would like to discuss it.

I found a function called `cx` in Emotion. This is almost the same as what I want to do.

https://emotion.sh/docs/@emotion/css#cx

For example:

```tsx
<button class={cx(btn, btnPrimary)}>{children}</button>
```

If we create a function `cx` in hono/css, users will be explicitly aware of the order by using it.

usualoma:
@yusukebe 

Oh sorry, I didn't find `cx()`. This way seems less confusing and better!
Implemented `cx()` in 7bce413. The following features of @emotion/css have not been implemented, but I believe they meet the current requirements.

* Conditional class names
* Using class names from other sources


Also, d2ffabf now ignores boolean and undefined, just like JSX. This allows us to write

```ts
function Button({ primary, children }) {
  const className = css`
    background-color: blue;
    ${primary && primaryColorClassName}
    `
  return (
    <button class={className}>{children}</button>
  )
}
```

or

```ts
<button class={cx(btn, primaryFlag && btnPrimary)}>button</button>
```

How about this?

yusukebe:
@usualoma 

Thanks! Yeah, `cx()` is good!

> How about this?

Great. Very cool that we can write `primaryFlag && btnPrimary`.

> The following features of @emotion/css have not been implemented, but I believe they meet the current requirements.

I personally would like to see a "Using class names from other sources" feature. This is because, in the case of Hono, we sometimes use it with external CSS.

```tsx
app.get('/', (c) => {
  const myPrimary = css`
    color: blue;
  `
  return c.html(
    <html>
      <head>
        <link href='https://cdn.jsdelivr.net/npm/bootstrap/dist/css/bootstrap.min.css' rel='stylesheet' />
        <Style />
      </head>
      <body>
        <h1 class={cx('display-1', myPrimary)}>Hello!</h1>
      </body>
    </html>
  )
})
```

What do you think?

usualoma:
@yusukebe OK!
82ff6a1 to allow mixing of external class names!

yusukebe:
@usualoma 

Awesome! Have you done everything about this PR?

usualoma:
> Have you done everything about this PR?

Yes, I am ready for the merge!

yusukebe:
Okay! I'll merge this into to #1850

PR #1856, opened 2023-12-25
feat(helper): export `SSEStreamingApi` and `SSEMessage`.

watany-dev:
fixes: https://github.com/honojs/hono/issues/1848

### Author should do the followings, if applicable

- [ ] Add tests
- [x] Run tests
- [x] `yarn denoify` to generate files for Deno

PR #1863, opened 2023-12-27
feat(helper): export SSEStreamingApi and SSEMessage.

watany-dev:
retake this.
https://github.com/honojs/hono/pull/1856

- [ ] Add tests
- [x] Run tests


yusukebe:
@watany-dev 

Thanks. I think it is a good idea to create `stream.ts`!

Merge now.

Issue #1810, opened 2023-12-12
Allow turning off `console.warn` when 404 on Bun `serveStatic` adapter


Th1nkK1D:
### What is the feature you are proposing?

First of all, I'm willing to open the PR for this since I already modified it on my project but I need some thought.

The current Bun `serveStatic` adapter will `console.warn` when the requested file is not found.

https://github.com/honojs/hono/blob/0f33cf8d457d246df1324fc6f0e9f8661f7fe4fd/src/adapter/bun/serve-static.ts#L52

However, this is not a behavior I want in the production. I want to propose an option to turn off the warning which might be done in 3 ways:
1. Add a boolean, eg. `showNotFoundWarning`  to the `ServeStaticOptions` to toggle it on /off.
2. Add `onNotFound` to the `ServeStaticOptions` as a handler when the 404 error is occurred. The default value can be the `console.warn` to keep the original behavior.
```ts
onNotFound: (path: string) => void | Promise<void> =
    (path) => console.warn(`Static file: ${path} is not found`) 
```
3. Just remove it, since Hono already provides [app.notFound](https://hono.dev/api/hono#not-found) which anyone can use it handle 404 in the way that they prefer, including `console.warn`.

Please let me know what you think :)

yusukebe:
Hi @Th1nkK1D,

Indeed, you might be right. The purpose of these warnings is to aid in debugging, letting users know which path `serveStatic` is searching. 

```
Static file: ${path} is not found
```

I think option `2` is the best choice because it's flexible and maintains the ability for debugging. If we go with this, the other `serveStatic` implementations, such as for Cloudflare and Deno, should also be updated accordingly.

Th1nkK1D:
Hi! Thank you for the fast response.

> I think option `2` is the best choice because it's flexible and maintains the ability for debugging.

Then do you think the default value of `onNotFound` should be the `console.warn`, or just the `undefined` value that does nothing?  I just feel like putting `onNotFound: () => {}` to disable the warning is a bit unintuitive. In contrast, defaulting to undefined is like, if we are not specific about what to do when an error occurs, then do nothing. I prefer this way but it breaks the old behavior.

> If we go with this, the other `serveStatic` implementations, such as for Cloudflare and Deno, should also be updated accordingly.

I can update those too.

yusukebe:
[...]

PR #1865, opened 2023-12-27
feat: pass context to onNotFound callback in serveStatic

sor4chi:
#1810 

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/adapter/deno/serve-static.ts
+++ b/deno_dist/adapter/deno/serve-static.ts
@@ -1,5 +1,5 @@
 import type { Context } from '../../context.ts'
-import type { MiddlewareHandler } from '../../types.ts'
+import type { Env, MiddlewareHandler } from '../../types.ts'
 import { getFilePath } from '../../utils/filepath.ts'
 import { getMimeType } from '../../utils/mime.ts'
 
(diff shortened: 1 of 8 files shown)


sor4chi:
I have no idea to infer Middleware Handler's Context and pass it to onNotFound argument's context.

yusukebe:
@sor4chi 

> I had no idea how to infer Middleware Handler's Context and pass it to onNotFound argument's context.

Please tell us along with specific examples.

sor4chi:
```ts
export type ServeStaticOptions<E extends Env> = {
  root?: string
  path?: string
  rewriteRequestPath?: (path: string) => string
  onNotFound?: (path: string, c: Context<E>) => void | Promise<void>
}

const DEFAULT_DOCUMENT = 'index.html'

export const serveStatic = <E extends Env = Env>(options: ServeStaticOptions<E> = { root: '' }): MiddlewareHandler<E> => {
```

If we could do something like this, we should be able to access envs type-safely from the callback context, but is this impossible with the current design?

yusukebe:
@sor4chi 

I struggled for hours to implement it, but I can't. Maybe it's impossible. I'll make more time to try.

sor4chi:
Okay, Thank you so much!
Even if this is not possible, it might help users just to be able to put Env Generics in `serveStatic`. If this is not possible, I will do so.

yusukebe:
@sor4chi 

Are you imagining the following code? If so, please implement that to make it true!

```ts
const app = new Hono<Env>()

app.get(
  '/static/*',
  serveStatic<Env>({
    onNotFound: (path, c) => {},
  })
)
```

sor4chi:
Hi, @yusukebe.
I finished supporting Env support for context.

yusukebe:
@sor4chi 

Thanks! But, also please change the `adapter/cloudflare-workers/server-static-module.ts`.

sor4chi:
@yusukebe fixed! thanks for finding it!

yusukebe:
@sor4chi 

Thanks! I'll merge this into "next" now.

PR #1869, opened 2023-12-28
feat: add sveltekit adapter

cstrnt:

- [x] Add tests
- [x] Run tests


yusukebe:
Hi @cstrnt 

Thank you for the PR. However, we are unable to accept this PR. Because the adapter is for the runtime, not for the web framework. If you want to add it, please make it as [3rd party middleware](https://github.com/honojs/middleware). But, we do not recommend this because the code is extremely short.

PR #1871, opened 2023-12-29
feat: implement stream.onAbort

sor4chi:

- [x] Add tests
- [x] Run tests

related: #1795 #1770

### DIFF
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -399,12 +399,12 @@ export class Context<
     headers?: HeaderRecord
   ): Response => {
     const { readable, writable } = new TransformStream()
-    const stream = new StreamingApi(writable)
+    const stream = new StreamingApi(writable, readable)
     cb(stream).finally(() => stream.close())
 
     return typeof arg === 'number'
-      ? this.newResponse(readable, arg, headers)
-      : this.newResponse(readable, arg)
+      ? this.newResponse(stream.responseReadable, arg, headers)
+      : this.newResponse(stream.responseReadable, arg)
   }
 
   /** @deprecated
--- a/deno_dist/helper/streaming/sse.ts
+++ b/deno_dist/helper/streaming/sse.ts
@@ -9,8 +9,8 @@ export interface SSEMessage {
 }
 
 export class SSEStreamingApi extends StreamingApi {
-  constructor(writable: WritableStream) {
-    super(writable)
+  constructor(writable: WritableStream, readable: ReadableStream) {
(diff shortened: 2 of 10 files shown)


sor4chi:
Happy new year @yusukebe.
Yes, please.

yusukebe:
@sor4chi 

Happy New Year!

I've checked it and tried to use it in some environment. But is this expected behavior?

https://github.com/honojs/hono/assets/10682/63ce4840-8c8e-45c9-bb81-28a6eb8a4985

The code:

```ts
import { streamText } from '../../src/helper/streaming'

export const app = new Hono()

app.get('/', async (c) => {
  return streamText(c, async (stream) => {
    stream.onAbort(async () => {
      console.log('Aborted!')
      await stream.close()
    })
    for (;;) {
      stream.writeln('Hello World!')
      console.log('Hello World!')
      await new Promise((r) => setTimeout(r, 1000))
    }
  })
})

serve(app)
```

HeyITGuyFixIt:
@yusukebe in your example, the loop still goes on, but there is no logic to stop the loop after the stream is aborted. You could make it a while loop instead of a for loop, and check for a variable (e.g. `while (!aborted)`) that is set by the event handler.

yusukebe:
Hi @HeyITGuyFixIt 

So you are saying that it should be code like this? I understand!

```ts
app.get('/', async (c) => {
  let aborted = false
  return streamText(c, async (stream) => {
    stream.onAbort(async () => {
      console.log('Aborted!')
      aborted = true
    })
    while (!aborted) {
      stream.writeln('Hello World!')
      console.log('Hello World!')
      await new Promise((r) => setTimeout(r, 1000))
    }
  })
})
```

HeyITGuyFixIt:
@yusukebe yeah. Give it a try to make sure that works

yusukebe:
@HeyITGuyFixIt 

It works well! What do you think of this PR? Looks good for you?

If @sor4chi can check again, I'd like to merge this into the next branch.

sor4chi:
@yusukebe 
I have noticed this behavior, but since the purpose is to take an abort hook in the first place and not to close the stream.

So I'd like to leave it as it is since I think I can handle it by implementing it myself if I want to do it as @HeyITGuyFixIt has presented it.

HeyITGuyFixIt:
@yusukebe it looks good to me

yusukebe:
@sor4chi @HeyITGuyFixIt Thanks! Merge now.

PR #1873, opened 2023-12-31
fix(dev): preserve original handler in [composedHandler].

### LINKED ISSUE #1868: `showRoutes` shows handler as `[middleware]` if subApp has a custom error handler

3.11.11

any

```ts
// src/dev-subapp.ts
import { Hono } from 'hono'
import { showRoutes } from 'hono/dev'

const subApp = new Hono()

subApp.get('/', (c) => c.json(0))
subApp.onError((_, c) => c.json(0))

const app = new Hono()
app.route('/', subApp)

showRoutes(app, {
  verbose: true
})
```

```txt
tsx src/dev-subapp.ts
```

It prints as below:

```txt
GET  /
       [handler]
```

### What do you see instead?

```txt
GET  /
       [middleware]
```

### Additional information

If comment out the following line, it works correctly.

```ts
const subApp = new Hono()

subApp.get('/', (c) => c.json(0))
// subApp.onError((_, c) => c.json(0)) // comment out

const app = new Hono()
app.route('/', subApp)
```

usualoma:
Fixes #1868

How about this solution?

### If we do more

I suppose we could add the information "added from sub pp" in inspectRoutes. But I didn't go that far at this point, because I think it's enough to fix the problem in #1868.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/hono-base.ts
+++ b/src/hono-base.ts
@@ -23,7 +23,7 @@ import type {
 } from './types'
 import { getPath, getPathNoStrict, getQueryStrings, mergePath } from './utils/url'
 
-export const composedHandler = Symbol('composedHandler')
+export const COMPOSED_HANDLER = Symbol('composedHandler')
 
 type Methods = typeof METHODS[number] | typeof METHOD_NAME_ALL_LOWERCASE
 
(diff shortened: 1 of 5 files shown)


src/hono-base.ts
```diff
@@ -23,6 +23,8 @@ import type {
 } from './types'
 import { getPath, getPathNoStrict, getQueryStrings, mergePath } from './utils/url'
 
+export const composedHandler = Symbol('composedHandler')
```
yusukebe, line 26: Since `composedHandler` is a constant like `METHODS`, maybe `COMPOSED_HANDLER` would be better. What do you think?

```ts
export const COMPOSED_HANDLER = Symbol('composedHandler')
```

yusukebe:
@usualoma Happy New Year!

I think this way is good. And I also think we don't need `added from sub app` now. I'v left one comment, please check it.

usualoma:
@yusukebe Happy New Year 🎍 
You're right, I did change it! 
239db3d

yusukebe:
@usualoma 

Thanks! I'll merge this now.

PR #1883, opened 2024-01-02
feat(dev): add `callback` option to `showRoutes()`

### LINKED ISSUE #1881: showRoutes() emits escape codes when running in deployment

3.11.11

Deno/CloudRun

Use `showRoutes()` function when the code runs in a cloud deployment.

It should not use escape codes because they only make sense when running locally in the console.

### What do you see instead?

image

### Additional information

_No response_

yusukebe:
This PR introduces `callback` option for `showRoutes()` in `hono/dev`. 

This allows users to define their own custom outputs.

```ts
import { Hono } from 'hono'
import { showRoutes } from 'hono/dev'

const app = new Hono()
app.get('/', (c) => c.text('/'))
app.get('/foo', (c) => c.text('foo'))

showRoutes(app, {
  callback: ({ method, path }) => {
    console.log(`${method} ${path}`)
  },
})
```

Resolves #1881

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/helper/dev/index.ts
+++ b/deno_dist/helper/dev/index.ts
@@ -2,15 +2,16 @@ import type { Hono } from '../../hono.ts'
 import { COMPOSED_HANDLER } from '../../hono-base.ts'
 import type { Env, RouterRoute } from '../../types.ts'
 
-interface ShowRoutesOptions {
-  verbose?: boolean
-}
-
 interface RouteData {
   path: string
   method: string
   name: string
   isMiddleware: boolean
+}
+
+interface ShowRoutesOptions {
+  verbose?: boolean
+  callback?: (data: { method: string; path: string; routes: RouteData[] }) => void
 }
 
 const isMiddleware = (handler: Function) => handler.length > 1
@@ -58,6 +59,12 @@ export const showRoutes = <E extends Env>(hono: Hono<E>, opts?: ShowRoutesOption
       if (!data) {
         return
       }
+
+      if (opts?.callback) {
+        opts.callback(data)
+        return
+      }
+
       const { method, path, routes } = data
 
       console.log(`\x1b[32m${method}\x1b[0m ${' '.repeat(maxMethodLength - method.length)} ${path}`)
--- a/src/helper/dev/index.test.ts
+++ b/src/helper/dev/index.test.ts
@@ -110,6 +110,24 @@ describe('showRoutes()', () => {
       '           [handler]',
     ])
   })
+
+  it('should render output specified by the `callback` option', async () => {
+    showRoutes(app, {
+      callback: ({ method, path }) => {
+        console.log(`${method} ${path}`)
+      },
+    })
+    expect(logs).toEqual([
+      'GET /',
+      'GET /named',
+      'POST /',
+      'PUT /',
+      'PATCH /',
+      'DELETE /',
+      'OPTIONS /',
+      'GET /static',
+    ])
+  })
 })
 
 describe('geRouterName()', () => {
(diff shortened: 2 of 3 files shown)


yusukebe:
@usualoma 

Could review this? I don't know if `callback` is the best naming, but I think it's not bad.

begoon:
Looks very useful. Thanks.

usualoma:
@yusukebe 
I think it would be better to add a simpler `colorize: boolean` option for #1881, since there are a few users who don't want to output escape sequences. This is like colorize_logging in rails.

https://guides.rubyonrails.org/configuring.html#config-colorize-logging

At this point, I think `colorize: boolean` is better, unless we have some other motivation than "I don't want to output escape sequences" and you want a callback.

If we want to do something more difficult than that, I think you should use `inspectRoutes`.

```
inspectRoutes(app).map(...).filter(...).forEach((args) => console.log(...))
```


`colorize: boolean` means that the following changes are assumed.


```diff
diff --git a/src/helper/dev/index.ts b/src/helper/dev/index.ts
index 5c8b993c..896c10e1 100644
--- a/src/helper/dev/index.ts
+++ b/src/helper/dev/index.ts
@@ -3,6 +3,7 @@ import type { Env, RouterRoute } from '../../types'
 
 interface ShowRoutesOptions {
   verbose?: boolean
+  colorize?: boolean
 }
 
 interface RouteData {
@@ -49,7 +50,10 @@ export const showRoutes = <E extends Env>(hono: Hono<E>, opts?: ShowRoutesOption
       }
       const { method, path, routes } = data
 
-      console.log(`\x1b[32m${method}\x1b[0m ${' '.repeat(maxMethodLength - method.length)} ${path}`)
+      const methodStr = opts?.colorize ?? true ? `\x1b[33m${method}\x1b[0m` : method
+      console.log(
+        `\x1b[32m${methodStr}\x1b[0m ${' '.repeat(maxMethodLength - method.length)} ${path}`
+      )
 
       if (!opts?.verbose) {
         return
```

or

```diff
diff --git a/src/helper/dev/index.ts b/src/helper/dev/index.ts
index 5c8b993c..db2809b7 100644
--- a/src/helper/dev/index.ts
+++ b/src/helper/dev/index.ts
@@ -3,6 +3,7 @@ import type { Env, RouterRoute } from '../../types'
 
 interface ShowRoutesOptions {
   verbose?: boolean
+  colorize?: boolean
 }
 
 interface RouteData {
@@ -30,6 +31,11 @@ export const showRoutes = <E extends Env>(hono: Hono<E>, opts?: ShowRoutesOption
   const routeData: Record<string, RouteData[]> = {}
   let maxMethodLength = 0
   let maxPathLength = 0
+  const colorize =
+    opts?.colorize ?? 'Deno' in globalThis
+      ? // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        (globalThis as any).Deno.isatty(1)
+      : process.stdout.isTTY
 
   inspectRoutes(hono)
     .filter(({ isMiddleware }) => opts?.verbose || !isMiddleware)
@@ -49,7 +55,10 @@ export const showRoutes = <E extends Env>(hono: Hono<E>, opts?: ShowRoutesOption
       }
       const { method, path, routes } = data
 
-      console.log(`\x1b[32m${method}\x1b[0m ${' '.repeat(maxMethodLength - method.length)} ${path}`)
+      const methodStr = colorize ? `\x1b[33m${method}\x1b[0m` : method
+      console.log(
+        `\x1b[32m${methodStr}\x1b[0m ${' '.repeat(maxMethodLength - method.length)} ${path}`
+      )
 
       if (!opts?.verbose) {
         return
```

yusukebe:
@usualoma 

Thanks! `colorize` option seems to be good for me. As you mentioned, we can do some other things with `inspectRoutes`.

PR #1885, opened 2024-01-02
feat(client): add `param` option to `$url()`

### LINKED ISSUE #1870: Allow URL params in `$url()` RPC

I love the [`$url()`](https://hono.dev/guides/rpc#url) feature.
It just that I don't find the returned `mergePath` very useful.
Can it use the `replaceUrlParam` too if the url has input params ?

I'm thinking about some like this :

```ts
const router = new Hono()
  .get(
    "/foo/:id/bar",
    zValidator("param", z.object({ id: z.string() })),
    function ({ text, req }) {
	  const { id } = req.valid("params");
	  // [...]
      return text("OK");
    },
  );

// [...]

app_hc.foo[":id"].bar.$url({ param: { id: "1" } });
// >> http://localhost:3000/foo/1/bar
```

yusukebe:
This PR introduces an option to the `$url()` function in Hono's client.

You can write the param values as an option of `$url()` like the following:

```ts
const route = app.get('/api/:id/bar', (c) => c.json({ foo: 'bar' }))

const client = hc<typeof route>('http://localhost:8787/')
const url = client.api[':id'].bar.$url({
  param: {
    id: '123',
  },
})
```

Inferring types are available:

https://github.com/honojs/hono/assets/10682/eaff7c97-6429-41ae-b2ca-8813387a85af

Resolves #1870

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/client/client.ts
+++ b/deno_dist/client/client.ts
@@ -142,6 +142,9 @@ export const hc = <T extends Hono<any, any, any>>(
     const path = parts.join('/')
     const url = mergePath(baseUrl, path)
     if (method === 'url') {
+      if (opts.args[0] && opts.args[0].param) {
+        return new URL(replaceUrlParam(url, opts.args[0].param))
+      }
       return new URL(url)
     }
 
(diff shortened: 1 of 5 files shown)


douglasduteil (approved):
<p align="center">
    
</p>

yusukebe:
@douglasduteil Thanks!

PR #1886, opened 2024-01-02
feat(dev): add `colorize` option for `showRoutes()`

### LINKED ISSUE #1881: showRoutes() emits escape codes when running in deployment

yusukebe:
This PR introduces the `colorize` option for `showRoutes()` in `hono/dev`.

You can disable colorizing the output if you set the option as `false`:

```ts
showRoutes(app, {
  colorize: false,
})
```

Resolves #1881, related to #1883.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/helper/dev/index.ts
+++ b/deno_dist/helper/dev/index.ts
@@ -4,6 +4,7 @@ import type { Env, RouterRoute } from '../../types.ts'
 
 interface ShowRoutesOptions {
   verbose?: boolean
+  colorize?: boolean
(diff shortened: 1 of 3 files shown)


usualoma:
@yusukebe LGTM!

PR #1887, opened 2024-01-02
refactor: cloudflare workers' serve-static module

sor4chi:

- [x] Add tests
- [x] Run tests

fixes of #1865


sor4chi:
Hi, @yusukebe

Could you please review this one too?
I forgot `onNotFound` in the serve-static module and refactored around it.

yusukebe:
How about this?

Since the `options.namespace` is not `deleted`, the actual value could come in, but since it has a type, I think we can go with this.

```ts
const module = <E extends Env = Env>(
  options: Omit<ServeStaticOptions<E>, 'namespace'> = { root: '' }
) => {
  options.manifest ??= manifest
  return serveStatic<E>(options)
}
```

sor4chi:
@yusukebe exactly, thank you!

yusukebe:
@sor4chi 

Okay! Land it.

PR #1900, opened 2024-01-05
feat: exports router types

nakasyou:
If you merge this PR, Hono users be able to import router types.

I want to custom router, so I have to import router type.
But router type (`src/router.ts`) isn't exported types.
So I must write this code.
```ts
import type { Router } from '....../node_modules/hono/dist/types/router'
```
It isn't smart. So I made this PR.

People who want to make custom router can writes this code if you marge this PR:
```ts
import type { Router } from 'hono/router'
```

memo: I think better name than `hono/router` is probably exists.

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -179,6 +179,11 @@
       "import": "./dist/validator/index.js",
       "require": "./dist/cjs/validator/index.js"
     },
+    "./router": {
+      "types": "./dist/types/router.d.ts",
+      "import": "./dist/router.js",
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @nakasyou !

Looks good to me! `hono/router` is fine! I'll merge this right now and include it in the next patch release because it is a minor feature. Thanks.

nakasyou:
Thank you for merging my Pull Request!

PR #1904, opened 2024-01-06
feat: Introducing a New SSG Adaptor/Helper 

watany-dev:
To celebrate the release of Hono v4, I propose a new feature. This is somewhat niche, but it's a Helper (or Adaptor) designed to convert Hono code into static HTML. Utilizing this, users can easily host SSG on platforms like S3 or R2. Currently, I'm grappling with three main concerns.

1. is this within the scope of responsibilities for the Hono framework? If it's deemed unnecessary, I might consider moving it to hono/middleware.
2. I'm wondering if it's more fitting to provide this as an Adaptor rather than a Helper, given its nature.
3. is the functionality interface appropriate? I've found myself pondering over this and have started studying SSG in other frameworks as well :)

- [x] Add tests
- [x] Run tests


sor4chi:
Hi @watany, excuse me from outside.

In terms of Hono's built-in middleware approach, I think we have to avoid dependence on runtime as much as possible.

For example, I thought it would be cleaner to export only the function to generate `Map<path, renderedHtml>` from `hono/ssg` (maybe `hono/renderer` would be a better name) and provide the function to save the output to disk from the adaptor of each runtime. I thought it would be cleaner to provide a function from the runtime adaptor to save the output to the disk.
This would allow for a more scalable approach, such as caching rendered string output in an environment without fs.

nakasyou:
I wrote PR: https://github.com/watany-dev/hono/pull/3 for this.

You can use `generateStaticPaths` like API if you merge it PR.

watany-dev:
@sor4chi 
Thank you. I hope the fix is not misrecognized in this image.
https://github.com/watany-dev/hono/pull/5

watany-dev:
@nakasyou 
It's a cool feature, but it's middleware that is not in the plan I was trying to do, so please give me a little time to think about it.

sor4chi:
Thanks, @watany-dev.
I mostly agree with the policy, but I personally would like to kick `path` and `Buffer` out of `hono/ssg`.
These are supposed to be nodejs modules...(This can be seen by doing denoify)

yusukebe:
Hey @watany-dev !

Can you check my comments?

watany-dev:
Yes, I am fixing it now. I followed @sor4chi advice and am taking the time to port the toSSG to Bun and Deno!

deno_dist/helper/ssg/index.ts
```diff
@@ -0,0 +1,88 @@
...
+export const toSSG = async (
+  app: Hono,
+  fsModule: FileSystemModule,
+  options: { dir: string }
+): Promise<ToSsgResult> => {
+  try {
+    const maps = await generateHtmlMap(app)
+    const files = await saveHtmlToLocal(maps, fsModule, options.dir)
+
+    console.log('Static site generation completed.')
```
yusukebe, line 82: It would be helpful to have this message, but it would be better to allow users to change it at will. So, we can remove it. Maybe the user should just specify the message based on the result:

```ts
const result = await toSSG(app, fs)
if (result.success) {
  console.log('Static site generation completed!')
} else {
  console.log('Failed!', result.error.message())
}
```

@watany-dev 
How do you think about this?

deno_dist/helper/ssg/index.ts
```diff
@@ -0,0 +1,88 @@
...
+import { inspectRoutes } from '../../helper/dev/index.ts'
+import type { Hono } from '../../hono.ts'
+import type { Env, Schema } from '../../types.ts'
+
+export interface FileSystemModule {
+  writeFile(path: string, data: string | Buffer): Promise<void>
+  mkdir(path: string, options: { recursive: boolean }): Promise<void | string>
+}
+
+export interface ToSsgResult {
```
yusukebe, line 12: If you made `ToSsg` into `ToSSG`, you might as well make `ToSsgResult` into `ToSSGResult` too. Like `URL` and `URLSearchParams`.

watany-dev:
@yusukebe 
I believe I've now addressed all the comments that have come in.

 Notes:
- I've partially added deno.d.ts to the Adaptor to use Deno's file module.
- Bun now supports Vitest, so I'm taking this opportunity to unify the test cases under Vitest. 
- The existing tsconfig.json stopped working, so I've added some settings to it.

To Do:
- Add toSSG to the Node.js Adaptor. I will issue a Pull Request for this after the interfaces for Deno and Bun have been accepted.
- Documentation needs to be created.
- Deno's c.render doesn't work with the same code as Bun for some reason. I have simplified it for now.

deno_dist/helper/ssg/index.ts
watany-dev: fixed.

src/helper/ssg/index.ts
```diff
@@ -0,0 +1,74 @@
...
+    const dirPath = path.dirname(filePath)
+
+    await fsModule.mkdir(dirPath, { recursive: true })
+    await fsModule.writeFile(filePath, html)
+    files.push(filePath)
+  }
+
+  return files
+}
+
```
yusukebe, line 63: I think `helper/ssg` should implement `toSSG` not only exporting `ToSSGInterface`.

```ts
export const toSSG: ToSSGInterface = async (app, fs, options) => {
  try {
    const maps = await generateHtmlMap(app)
    const files = await saveHtmlToLocal(maps, fs, options.dir)
    return { success: true, files }
  } catch (error) {
    const errorObj = error instanceof Error ? error : new Error(String(error))
    return { success: false, error: errorObj }
  }
}
```

Then, for Node.js, you can do the following without an adapter. Also, for platforms other than Node.js, Bun, and Deno, a user can just import `toSSG()` from `hono/ssg`.

```ts
import fs from 'node:fs/promises'

toSSG(app, fs, { dir: './static '})
```

Also, for example, Bun's adapter can be written as follows:

```ts
import { toSSG as baseToSSG } from '../../helper/ssg'

// ...

export const toSSG: ToSSGInterface = async (app, bunFileSystemModule, options: { dir: string }) => {
  return baseToSSG(app, bunFileSystemModule, options)
}
```

yusukebe:
Hi @watany-dev !

It does not seem to be the end yet. Check my comments!

tsconfig.json
```diff
@@ -4,6 +4,8 @@
     "declaration": true,
     "moduleResolution": "Bundler",
     "outDir": "./dist",
+    "rootDir": "./",
```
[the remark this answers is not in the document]
watany-dev: This is a hack because I could not run the test locally, but this was resolved some time ago. Fixed.

tsconfig.json
```diff
@@ -4,6 +4,8 @@
...
+    "allowImportingTsExtensions": true,
```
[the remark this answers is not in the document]
watany-dev: fixed

src/helper/ssg/index.ts
watany-dev: The toSSG in deno and bun's Adaptor is now just a wrapper, so we made an interface that omits the fs, at least for convenience. If you don't need this realization, you probably don't need to put toSSG in Adaptor in the first place.

src/helper/ssg/index.ts
```diff
@@ -0,0 +1,74 @@
...
+export interface ToSSGInterface<
+  E extends Env = Env,
+  S extends Schema = {},
+  BasePath extends string = '/'
+> {
+  (app: Hono<E, S, BasePath>, fsModule: FileSystemModule, options: { dir: string }): Promise<{
```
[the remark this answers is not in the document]
watany-dev: fixed

watany-dev:
@yusukebe 
Thanks for the great review. I am sure we could have handled it. I also set a Default value for path since option was a required argument. Hope you like it.

sor4chi:
By the way, do you have any idea how to deal with dynamic param regarding SSG?

yusukebe (approved):
Looks great!

yusukebe:
@watany-dev 

Thanks! You did it! I'll merge this into the "v4" branch.

@sor4chi 

> By the way, do you have any idea how to deal with dynamic param regarding SSG?

I prefer the approach suggested in @nakasyou's PR. For example:

```ts
app.get(
  '/posts/:id',
  ssgParams(() => {
    return Array.from({ length: 10 }, (_, i) => ({ id: i.toString() }))
  }),
  (c) => {
    return c.html(`<h1>${c.req.param('id')}</h1>`)
  }
)
```

@nakasyou Please create your own PR based on "v4", not necessarily the same API as avebe my one.

yusukebe:
@nakasyou @watany-dev @sor4chi 

Or, instead of specifying it in the handler, you can create hooks in the `toSSG()` options. This way, you don't have to modify the application.

```ts
toSSG(app, fs, {
  afterRequest: (route, app) => {
    if (route.path === '/posts/:id') {
      Array.from({ length: 10 }, (_, i) => {
        app.request(`/posts/${i}`)
      })
    }
  },
  beforeResponse: (res) => {
    if (!res.headers.get('content-type').match(/^text\/html/)) {
      return null
    }
    return res
  }
})
```

PR #1911, opened 2024-01-07
feat: Add headers option to serveStatic()

glinford:
### Add headers option to serveStatic()

I'm encountering a situation were I need to serve static files (with Cloudflare Workers) and each of these files need to have some specific headers when served.

Currently I can do that with the help of a middleware using:
```js
app.use('something that matches my specific file', async (c, next) => {
    c.res.headers.set('foo', `bar`);
    await next()
});
```
or
```js
app.use('something more global', async (c, next) => {
    if(c.req.url.endsWith('myfilename')) {
       c.res.headers.set('foo', `bar`);
    }
    await next()
});
```

This is a proposal to add headers to `serveStatic()` via an optional `Headers` object, this way I can set the headers when serving my static files directly and I don't need to grow my current middlewares.

Example:
```js
cost headers = new Headers();
headers.set('foo', 'bar');
app.get('/mypaththatneedtoservefileswithaspecificheaders', serveStatic({ path: './script.js', headers }));
```

It's non-blocking and simply a suggestion which I think would make the `serveStatic` method more versatile to the Hono users.

It may not be a big of a problem enough to be considered, let me know.

- [x] Add tests
- [x] Run tests
- [x] Files are formatted with the project prettier config
- [ ] The relevant documentation will need to be updated if the proposal is approved on the [hono/website](https://github.com/honojs/website/tree/main/getting-started) repo

### DIFF
--- a/deno_dist/adapter/deno/serve-static.ts
+++ b/deno_dist/adapter/deno/serve-static.ts
@@ -12,6 +12,7 @@ export type ServeStaticOptions<E extends Env = Env> = {
   path?: string
   rewriteRequestPath?: (path: string) => string
   onNotFound?: (path: string, c: Context<E>) => void | Promise<void>
+  headers?: Headers
 }
 
 const DEFAULT_DOCUMENT = 'index.html'
@@ -50,6 +51,11 @@ export const serveStatic = <E extends Env = Env>(
       const mimeType = getMimeType(path)
       if (mimeType) {
         c.header('Content-Type', mimeType)
+      }
+      if (options.headers) {
+        for (const entry of options.headers.entries()) {
+          c.header(...entry)
+        }
       }
       // Return Response object with stream
       return c.body(file.readable)
--- a/runtime_tests/bun/index.test.tsx
+++ b/runtime_tests/bun/index.test.tsx
@@ -101,6 +101,13 @@ describe('Serve Static Middleware', () => {
     })
   )
 
+  const headers = new Headers()
+  headers.set('foo', 'bar')
+  app.all(
+    '/favicon-with-headers.ico',
+    serveStatic({ path: './runtime_tests/bun/favicon.ico', headers })
+  )
+
   beforeEach(() => onNotFound.mockClear())
 
   it('Should return static file correctly', async () => {
@@ -108,6 +115,15 @@ describe('Serve Static Middleware', () => {
     await res.arrayBuffer()
     expect(res.status).toBe(200)
     expect(res.headers.get('Content-Type')).toBe('image/x-icon')
+    expect(res.headers.get('foo')).toBe(null)
+  })
+
+  it('Should return static file correctly with headers', async () => {
+    const res = await app.request(new Request('http://localhost/favicon-with-headers.ico'))
+    await res.arrayBuffer()
+    expect(res.status).toBe(200)
+    expect(res.headers.get('Content-Type')).toBe('image/x-icon')
+    expect(res.headers.get('foo')).toBe('bar')
   })
 
   it('Should return 404 response', async () => {
--- a/runtime_tests/deno/middleware.test.tsx
+++ b/runtime_tests/deno/middleware.test.tsx
@@ -92,10 +92,23 @@ Deno.test('Serve Static middleware', async () => {
       rewriteRequestPath: (path) => path.replace(/^\/dot-static/, './.static'),
     })
   )
+  const headers = new Headers()
+  headers.set('foo', 'bar')
+  app.all(
+    '/favicon-with-headers.ico',
+    serveStatic({ path: './runtime_tests/deno/favicon.ico', headers })
+  )
 
   let res = await app.request('http://localhost/favicon.ico')
   assertEquals(res.status, 200)
   assertEquals(res.headers.get('Content-Type'), 'image/x-icon')
+  assertEquals(res.headers.get('foo'), null)
+  await res.body?.cancel()
+
+  res = await app.request('http://localhost/favicon-with-headers.ico')
+  assertEquals(res.status, 200)
+  assertEquals(res.headers.get('Content-Type'), 'image/x-icon')
+  assertEquals(res.headers.get('foo'), 'bar')
   await res.body?.cancel()
 
   res = await app.request('http://localhost/favicon-notfound.ico')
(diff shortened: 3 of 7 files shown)


yusukebe:
Hi @glinford 

I would like you to use middleware to add the headers. That is why we have middleware. I know your PR makes it `non-blocking`, but as a Hono policy, please use middleware.

glinford:
@yusukebe I understand

As a feedback, I do find it convoluted to have to specify 2 "things" for each of my static files (`serveStatic()` + the equivalent in my middleware) and I'm currently having a use-case with a lot of different path / different ```Headers``` to set based on these paths so it's becoming complex to maintain.

I would think the optional `Headers` would be a widespread enough use case to be part of the `serveStatic()` feature specs. Perhaps I'm wrong.

If so, I'm fine with this proposal being closed 😄 go ahead.

Thanks for looking at this!

yusukebe:
@glinford 

Thank you for understanding!

This may not be what you expect, but you might be able to write it more easily by creating an `addHeader` custom middleware like the following.

```ts
const addHeader = (headers: Headers) => {
  return createMiddleware(async (c, next) => {
    await next()
    for (const entry of headers.entries()) {
      c.header(...entry)
    }
  })
}

const headers = new Headers()
headers.set('foo', 'bar')

app.get('/static/*', serveStatic({ path: './script.js' }), addHeader(headers))
```

glinford:
@yusukebe  Nice one! yes that would clean things up 😁

Closing

cyantree:
I'm under the impression that successfully serving a file through `serveStatic()` doesn't trigger the next middleware as in this case `serveStatic()` isn't executing `next()`:
https://github.com/honojs/hono/blob/main/src/middleware/serve-static/index.ts#L80

So it's only possible to define the middleware before `serveStatic()`. However this might lead to problems in case of multiple overlapping routes and the first one falling through because of a non-existing file as the middleware would run despite `serveStatic()` not serving anything.
```
app.get('/*', cacheControl(...), serveStatic({root: 'assets'}));
app.get('/*', cacheControl(...), serveStatic({path: 'index.html'}));
```

Is this a bug? Should I file a separate issue for that? Or am I missing something?

yusukebe:
Hi @cyantree 

It's not a bug. It should return a Response similar to a normal handler.

yusukebe:
Hi @cyantree 

As mentioned above, It's the same as the normal handler's behavior:

```ts
const mw = (n: number) =>
    console.log('before', n)
    await next()
    console.log('after', n)
  })

app.get(
  '*',
  mw(1),
  // Below handler behaves as same as serveStatic()
  (c) => {
    return c.text('Handler')
  },
  mw(2) // not executed
)
```

cyantree:
Hi @yusukebe ,
Thanks for putting up with me.

Let me provide a more detailed example to illustrate my point:
```ts
function cacheControl(value: string): MiddlewareHandler {
    return async (c, n) => {
        console.log('pre', value);
        await n();
        console.log('post', value);

        if (c.res.ok) {
            c.res.headers.set('cache-control', value);
        }
    };
}

app.get('/*', cacheControl(CC_ASSET), serveStatic({
    root: 'assets',
}));
app.get('/*', cacheControl(CC_NO_CACHE), serveStatic({root: 'assets', path: 'index.html'}));
```

Now let's take a look at two requests:

## 1. Accessing an existing asset like `/styles.css`

- pre CC_ASSET will be logged
- serveStatic() will be called and returns the asset
- post CC_ASSET will be logged
- header will be set: 'cache-control': CC_ASSET

This behaviour is what I expect.

## 2. Accessing a page like `/contact`

- pre CC_ASSET will be logged
- serveStatic() will be called and passes through because it's no asset
- pre CC_NO_CACHE will be logged
- serveStatic('index.html') will be called and returns the `index.html`
- post CC_NO_CACHE will be logged
- header will be set: 'cache-control': CC_NO_CACHE
- post CC_ASSET will be logged
- header will be set: 'cache-control': CC_ASSET

So in this case both `cacheControl()` middlewares will be called and the earlier one then replaces the correct value.

## Workaround

Currently my workaround is to check whether `cache-control` already has been set on the response:
```
        if (c.res.ok && !c.res.headers.has('cache-control')) {
            c.res.headers.set('cache-control', value);
        }
```

However I think that's not a clean solution as it requires post-next()-actions of middlewares used before `serveStatic()` to have this abort mechanism.

## Possible solutions

One option would be to provide a callback that may be used for refining the response. Like:
```ts
get('/assets/*', serveStatic({onFound: (c) => {
    c.res.headers.set('cache-control', 'something');
}}));
```

This also matches ways found in other frameworks:
https://github.com/fastify/fastify-static?tab=readme-ov-file#setheaders
https://github.com/koajs/send#setheaders

Another option might be some way to prevent middlewares to "leak" from one route to another in case the chain doesn't return a response. However unfortunately I don't have a code sample for this atm.

---

I hope that makes it a bit clearer.

yusukebe:
@cyantree 

Hmm. I think the way you mentioned is good:

```ts
get(
  '/assets/*',
  serveStatic({
    onFound: (c) => {
      // ...
    },
  })
)
```

cyantree:
Should I give a PR a try or are you planning on doing it?

yusukebe:
Hi @cyantree 

> Should I give a PR a try or are you planning on doing it?

Ah, we already have the API!

https://hono.dev/getting-started/deno#onnotfound

cyantree:
But I need it the other way around. `onFound()` instead of `onNotFound()`.

yusukebe:
@cyantree 

Honestly, I don't want to add the `onFound()` feature right now.

We do not expect to have multiple serve static handlers that specify `'/*'` in the path, as in your example.

```ts
app.get(
  '/*',
  cacheControl(CC_ASSET),
  serveStatic({
    root: 'assets'
  })
)
app.get('/*', cacheControl(CC_NO_CACHE), serveStatic({ root: 'assets', path: 'index.html' }))
```

If you want to serve only `index.html`, you can just write `app.get('index.html', ...)`. Or you can use a RegExp routing:

```ts
app.get('/:filename{.+.html}', cacheControl(CC_NO_CACHE), serveStatic({ root: 'assets' }))
```

cyantree:
Ok, thanks for your response.

PR #1916, opened 2024-01-08
docs: Add JSDoc

### LINKED ISSUE #1338: JSDoc

We haven't written JSDoc for the Hono code. I used to think that maintaining JSDoc would be troublesome, especially since we also have a website. However, now that the API is mostly stable, it's a good time to start. We don't need to write extensively in JSDoc, but including the basics would be beneficial.

nabeken5:
close #1338

- [ ] Add tests
- [ ] Run tests


src/hono-base.ts
```diff
@@ -55,7 +55,19 @@ const errorHandler = (err: Error, c: Context) => {
 type GetPath<E extends Env> = (request: Request, options?: { env?: E['Bindings'] }) => string
 
 export type HonoOptions<E extends Env> = {
+  /**
+   * `strict` option specifies whether to distinguish whether the last path is a directory or not.
+   * @default true
+   * @see https://hono.dev/api/hono#strict-mode
+   */
```
nabeken5, line 62: This was changed from docs.

Original:
> Strict mode defaults to `true` and distinguishes the following routes.

src/hono-base.ts
```diff
@@ -187,6 +187,14 @@ class Hono<
...
   }
 
+  /**
+   * `.basePath()` allows base paths to be specified.
+   * @example
+   * ```ts
+   * const api = new Hono().basePath('/api')
+   * ```
+   * @see https://hono.dev/api/routing#base-path
+   */
```
nabeken5, line 197: This was changed from docs too.
Original:
> You can specify the base path.

src/request.ts
```diff
@@ -46,6 +46,16 @@ export class HonoRequest<P extends string = '/', I extends Input['out'] = {}> {
...
+  /**
+   * `.req.param()` gets the path parameters.
+   * @example
+   * ```ts
+   * const name = c.req.param('name')
+   * // or all parameters at once
+   * const { id, comment_id } = c.req.param()
+   * ```
+   * @see https://hono.dev/api/routing#path-parameter
+   */
```
nabeken5, line 58: This was added from docs.
https://hono.dev/api/routing#path-parameter

nabeken5: When I only look at the code, it's better to use `.param()` instead

EdamAme-x:
good
I want to help with this PR

nabeken5:
Thanks @EdamAme-x, but this work is almost done and I need your help to make the API docs better!

EdamAme-x:
I've added a few things i think need

nabeken5:
@EdamAme-x please make PR  (to nabeken5/hono).

It looks fine to me, so I'll open it.
@yusukebe  Please check.

EdamAme-x:
hi @nabeken5 
I'm actually using codespaces to do the work because I'm using **Windows** and **storage issues**.
I'll try to add  new PullRequest next (next month) time since codespace's no longer available this month.

nabeken5:
hi @EdamAme-x 
I want to move this PR as fast as we can.
I don't know much about it, but maybe using other 3rd-party apps like [Gitpod](https://www.gitpod.io/) or [CodeSandbox](https://codesandbox.io/) would help?
(I've used CodeSandbox and recommend it because you can SSH from a local VSCode)

EdamAme-x:
thanks.
ideally, they should be able to move ahead since there is other work to be done and it will likely take some time.

EdamAme-x:
hi @nabeken5 
i created PullRequest to your branch.
https://github.com/nabeken5/hono/pull/1

nabeken5:
ready to go🥳

yusukebe:
@nabeken5 

Super thanks! I'll check this later.

yusukebe:
@nabeken5 @EdamAme-x Thanks both! I'd like to merge this PR.

@nabeken5 

Could you do the three things for merging?

1. Resolve the conflicts.
2. `yarn format:fix` to format the code.
3. `yarn denoify` to generate the code for Deno.

nabeken5:
We are ready!

yusukebe:
@nabeken5 @EdamAme-x Thanks both!

Let's land it. Merging now!

PR #1917, opened 2024-01-08
feat: Introduce jsx/dom

usualoma:

- [x] Add tests
- [x] Run tests


usualoma:
There are probably still a lot of bugs, but since we have implemented a minimal number of hooks and added a minimal number of tests, I would like to end this PR work here.
I think we are at a level where we can consider merging.

yusukebe:
Hi @usualoma !

Awesome work! I have made a minimal example and used it, but the feel is good.

I've made the `v4` branch for the X day. I'll merge this into `v4` and release the RC version. Then shall we can improve it!

Thanks.

Issue #1915, opened 2024-01-08
Cache middleware ignores handler Cache-Control settings


eduardvercaemer:
### What version of Hono are you using?

3.12.0

### What runtime/platform is your app running on?

Cloudflare Workers

### What steps can reproduce the bug?

1. Create new hono app
2. Add cache middleware (i.e. max-age=3600)
3. Inside of some handler that returns user data, do `c.header("Cache-Control", "private", { append: true })`

### What is the expected behavior?

Individual handlers should be able to append extra directives to cache-control header.

### What do you see instead?

The cache middleware always overrides the response cache-control to a static header.

### Additional information

The fix is real easy: In https://github.com/honojs/hono/blob/main/src/middleware/cache/index.ts#L13 change to
```
  const addHeader = (c: Context) => {
    if (options.cacheControl)
      c.header("Cache-Control", options.cacheControl, { append: true });
  };
```

I can make a PR if needed.

yusukebe:
Hi @eduardvercaemer !

Thank you for reporting, it's a bug!

> I can make a PR if needed.

Please!

PR #1922, opened 2024-01-08
fix(cache): Allow composing cache control directives

eduardvercaemer:
# Changes

Fix https://github.com/honojs/hono/issues/1915

1. Allow nested cache middlewares to compose a final Cache-Control header
2. Allow handlers to also add cache control directives

This slightly changes the implementation from raw string handling to acknowledge the existance of individual directives, this way we can avoid duplicating directives that were overriden in inner middleware/handlers.

- [X] Add tests
- [X] Run tests

### DIFF
--- a/src/middleware/cache/index.test.ts
+++ b/src/middleware/cache/index.test.ts
@@ -31,8 +31,23 @@ describe('Cache Middleware', () => {
   })
 
   app.use('/wait2/*', cache({ cacheName: 'my-app-v1', wait: true, cacheControl: 'max-age=10' }))
-  app.use('/wait2/*', cache({ cacheName: 'my-app-v1', wait: true, cacheControl: 'max-age=10' }))
+  app.use('/wait2/*', cache({ cacheName: 'my-app-v1', wait: true, cacheControl: 'Max-Age=20' }))
   app.get('/wait2/', (c) => {
+    return c.text('cached')
+  })
+
+  app.use('/wait3/*', cache({ cacheName: 'my-app-v1', wait: true, cacheControl: 'max-age=10' }))
+  app.use(
+    '/wait3/private/*',
+    cache({ cacheName: 'my-app-v1', wait: true, cacheControl: 'private' })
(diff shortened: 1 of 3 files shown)


eduardvercaemer:
Perhaps it makes sense to add a directive-based APi instead of relying on plain strings in the middleware constructor and maybe also in the handlers ?

```js
cache({
  maxAge: 0,
})
// ...
c.cache({ private: true })
```

src/middleware/cache/index.test.ts
```diff
@@ -31,11 +31,23 @@ describe('Cache Middleware', () => {
   })
 
   app.use('/wait2/*', cache({ cacheName: 'my-app-v1', wait: true, cacheControl: 'max-age=10' }))
-  app.use('/wait2/*', cache({ cacheName: 'my-app-v1', wait: true, cacheControl: 'max-age=10' }))
+  app.use('/wait2/*', cache({ cacheName: 'my-app-v1', wait: true, cacheControl: 'max-age=20' }))
```
yusukebe, line 34: If `max-age=20` is `Max-age=20`, the test will fail. If it compares the names of the directives, I think we have to make all uppercase letters lowercase first.

If there is a cost to lowercase, you might prefer to lowercase everything first, even if only `directives`.

yusukebe:
Hi @eduardvercaemer 

Thanks for the PR.

First, please run `yarn format:fix` to fix the format for passing CI. And, added a comment, please check it.

> Perhaps it makes sense to add a directive-based APi instead of relying on plain strings in the middleware constructor and maybe also in the handlers ?

That is certainly a good idea and would improve DX. If you are willing to make a PR, you are welcome to do so.

yusukebe:
@eduardvercaemer 

Awesome!

One more thing. Please run `yarn lint --fix` to fix the lint error.

yusukebe:
@eduardvercaemer 

You've fixed it! Thanks. Merging now!

PR #1924, opened 2024-01-08
fix(client): Fix issue awaiting RPC client

codekrafter:
Right now if you await the RPC client, it produces undefined behavior and can hang the application. This also occurs if the client is returned from an async function that is then awaited. This adds a quick fix to detect that edge case, and handle it properly.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/client/client.ts
+++ b/deno_dist/client/client.ts
@@ -12,6 +12,10 @@ const createProxy = (callback: Callback, path: string[]) => {
       return createProxy(callback, [...path, key])
     },
     apply(_1, _2, args) {
+      if (path.length === 1 && path[0] === "then") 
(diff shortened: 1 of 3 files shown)


yusukebe:
Hi @codekrafter 

Thanks for the PR. Please check the two things:

1. Run `yarn format:fix` to fix the format for passing CI.
2. Write a proper test.

codekrafter:
I've added a test and tweaked the fix to avoid the runtime from detecting the client as a thenable, instead of adding a passthrough promise implementation.

yusukebe (approved):
LGTM!

yusukebe:
@codekrafter I'll merge this now. Thanks a lot!

PR #1928, opened 2024-01-09
feat(css): Introduce pseudo global selector and class name based extend syntax

usualoma:
### Global selector

If you wrap the whole thing (not just part of it) in a pseudo-selector called `:-hono-global`. The output of the  `css` is not wrapped in the generated class name, but is interpreted as a top-level specification.

```ts
css`
:-hono-global {
  body {
    font-size: 10px;
  }
}
`
```

### Class name based extend syntax

You can extend it by embedding the class name by writing it in the following syntax `${baseClass} {`

```ts
const baseClass = css`display: flex`
const extendedByContent = css`
  ${baseClass}
  flex-direction: column;
` // This is already possible
const extendedByClassName = css`
  ${baseClass} {
    flex-direction: column;
    p {
      color: red;
    }
  }
` // This is made possible by this PR
```

### Use both "Global selector" and "Class name based extend syntax"

The following can be written to place the media query at the top level so that it can be interpreted by older browsers that do not support nesting.

```ts
const baseClass = css`display: flex`
const globalClass = css`
:-hono-global {
  @media (min-width: 768px) {
    ${baseClass} {
      flex-direction: column;
    }
  }
}
`
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/helper/css/index.ts
+++ b/deno_dist/helper/css/index.ts
@@ -29,6 +29,8 @@ export const rawCssString = (value: string): CssEscapedString => {
   return escapedString
 }
 
+const PSEUDO_GLOBAL_SELECTOR = ':-hono-global'
+const isPseudoGlobalSelectorRe = new RegExp(`^${PSEUDO_GLOBAL_SELECTOR}{(.*)}$`)
 const DEFAULT_STYLE_ID = 'hono-css'
 
 /**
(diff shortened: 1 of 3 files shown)


usualoma:
### Generated deferent class name for deferent first line comment even if the content is the same

"Class name based extend syntax" can be used to generate unique class names, since there may be times when you want to have the same content but different base classes. If a comment with unique content is added to the first line, its content is included in the calculation of the hash value.

```ts
const headerClassA = css`
  /* class A */
  display: flex;
`
const headerClassB = css`
  /* class B */
  display: flex;
`

const extendedClassA = css`
  ${headerClassA} {
    p { font-size: 10px }
  }
`

const extendedClassB = css`
  ${headerClassB} {
    p { font-size: 11px }
  }
`

// If headerClassA and headerClassB do not have different class names, problems will occur.
```

yusukebe:
@usualoma 

Is this the correct way to use the Global selector?

```tsx
app.get('/', (c) => {
  const globalClass = css`
    :-hono-global {
      body {
        color: red;
      }
    }
  `
  return c.html(
    <html>
      <head>
        <Style />
      </head>
      <body class={globalClass}>
        <h1>Hello!</h1>
      </body>
    </html>
  )
})
```

usualoma:
Hi, @yusukebe 

Please do it that way. css`` does not trigger side-effects, so even if you define global CSS, you need to explicitly embed it somewhere in the template.

Or you can specify it in the child component.

```ts
const baseClass = css`display: flex`
const baseClassWithGlobalClass = css`
:-hono-global {
  @media (min-width: 768px) {
    ${baseClass} {
      flex-direction: column;
    }
  }
}
`
app.get('/', (c) => {
  c.render(<div class={baseClassWithGlobalClass}>Content</div>);
})

// => <style id="hono-css">@media (min-width:768px){.css-3980466870{flex-direction:column}}.css-3980466870{display:flex}</style> ... <div class="css-3980466870">Content</div>
```

### If you want to use straightforward global CSS

By e6cf20f, when "css" is passed to the children of a Style component, it is now interpreted as global CSS. This is a seemingly crazy obvious behavior, but it is useful if you want to define global CSS by embedding variables. " -hono-global" is useful, but if we could define global css without it that would be more beautiful.

image

usualoma:
I've added all the changes I want to make, so you can have it merged into main or v4 whichever is better.

yusukebe:
@usualoma Thanks! makes sense.

> when "css" is passed to the children of a Style component, it is now interpreted as global CSS

This is so cool and super easy to understand.

I'll merge this into `v4` now.

PR #1929, opened 2024-01-09
denied return void from async handler and return context from next

zohayrslileh:

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -19,7 +19,7 @@ export type Env = {
   Variables?: Variables
 }
 
-export type Next = () => Promise<void>
+export type Next = () => Promise<Context>
 
 export type Input = {
   in?: Partial<ValidationTargets>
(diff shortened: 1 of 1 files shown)


yusukebe:
@zohayrslileh 

I don't know what your intention was with this, but it is not the spec, so I'm closing it.

zohayrslileh:
As you can see here for typescript everything is correct, but it will not work because return is void
image

When calling next, the return is void according to the types, but in reality it is a context object
image

thanks

yusukebe:
Hi @zohayrslileh 

> As you can see here for typescript everything is correct, but it will not work because return is void

You are correct; I want to make it show an error if the Response is not returned in a **handler**. However, since it is the `MiddlewareHandler,` it should allow `void.`

> When calling next, the return is void according to the types, but in reality it is a context object

Indeed. However, I don't expect to use the returned context object, so I'm making it `void`.

zohayrslileh:
Okay, I understand that, but if the next function return context object. The middleware will not return a void under any circumstances

Because middleware accepts void only for the next function

yusukebe:
@zohayrslileh 

Thanks for understanding and your advice. Everyone has different ideas about the types. I won't change it now, but I may change the return type next, as you suggest.

PR #1931, opened 2024-01-09
fix(netlify): change path in `deno_dist/adapter/netlify`

charkour:
### Purpose

There is an issue when using the Netlify Adapter via `deno.land/x`. https://github.com/honojs/hono/blob/8de60599aafca908b6214f047571415c6e17f8ea/deno_dist/adapter/netlify/handler.ts#L4 This line references `../../index.ts` but that doesn't exist in the root of `deno_dist/` so we much use `../../hono.ts` or `../../mod.ts`. Please let me know what you prefer. Thank you!

- [ ] Add tests
- [x] Run tests

Photo of error:
image


yusukebe:
Hi @charkour 

Looks good! Thanks!

charkour:
Thank you!

PR #1936, opened 2024-01-10
fix(jsx/dom): fix cleanup for deps

usualoma:
There was a problem that cleanup was not called when changing "deps", so this has been fixed.

- [x] Add tests
- [x] Run tests


usualoma:
@yusukebe, Would you please review?
Looking at the unit tests, you will understand what has been fixed.

yusukebe (approved):
LGTM!

PR #1940, opened 2024-01-10
feat: support multiple optinal parameters

### LINKED ISSUE #1925: Multiple optional parameters - are they supposed to work?

3.12.0

Cloudflare Workers

Create a minimal Hono app (this is running on Cloudflare Workers)
```ts
import { Hono } from "hono";

const app = new Hono();

app.get("/base/:firstoptional?/:secondoptional?", c => c.json(c.req.param()));
// /base ->                     404 (why?)
// /base/hello ->               { "firstoptional?": "hello" }
// /base/hello/world ->         { "secondoptional": "world", "firstoptional?": "hello" }

export default app;

```

- I would expect `/base` to match
- I would expect it to not return `?` in the param key names

### What do you see instead?

I see what I have commented in the code. The `?` is added to the param key name.

### Additional information

My real life use-case is:
```ts
app.get("/v1/leaderboard/:version?/:platform?", handler);
```
This acts in the same as the above examples.
I would like to be able to match `/v1/leaderboard` and then handle if the params are undefined.

If this is intended (I can understand the reasons for being intentional), then this might be seen as a feature request.

yusukebe:
Resolves #1925

This PR is a `feat,` but it's not so major then; I'll include this in the next patch release. 

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/utils/url.test.ts
+++ b/src/utils/url.test.ts
@@ -154,6 +154,11 @@ describe('url', () => {
       expect(checkOptionalParameter('/api/:animals?/type')).toBeNull()
       expect(checkOptionalParameter('/api/animals/:type?/')).toBeNull()
       expect(checkOptionalParameter('/:optional?')).toEqual(['/', '/:optional'])
+      expect(checkOptionalParameter('/v1/leaderboard/:version?/:platform?')).toEqual([
+        '/v1/leaderboard',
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @usualoma !

Could you review this?

Issue #1899, opened 2024-01-04
Guidance on nested layouts and renderers


bruceharrison1984:
### What is the feature you are proposing?

It'd be helpful if some documentation around nested layouts was available.

Currently doing something like the following doesn't work:

```ts
const app = new Hono();
app.use('*', renderer); 
app.get('/signin', signinRoute); // <- use root renderer

const secureRoute = new Hono();
secureRoute.use('*', secureRenderer);
secureRoute.get('/test', (ctx) => ctx.render(<>SOME TEST</>)); // <- use secure renderer, the root renderer doesn't get used

app.route('/secure', secureRoute );
```

Currently, which ever renderer is located closest to the route is the one that will be used. Simply using JSX components to wrap a layout around pages works well enough, but a best practice would be helpful.

SonikJS supports nested layouts in this fashion (at least via __layout) so it would seem that Hono could also support this.

yusukebe:
Hi @bruceharrison1984 !

Actually, I didn't take `c.render()` into consideration for nested layouts. However, as you may know, what `c.render()` does is simple as follows.


```ts
private renderer: Renderer = (content: string | Promise<string>) => this.html(content)

// ...

render: Renderer = (...args: any[]) => this.renderer(...args)
```

I'm thinking about it, but if you come up with a good way to nest layouts, please share!

By the way, Sonik's `__layout` mechanism will be deprecated and replaced "renderer" based. So it is great to be able to use `c.render()` for nested layouts in a smart way.

bruceharrison1984:
> I'm thinking about it, but if you come up with a good way to nest layouts, please share!
> 
> By the way, Sonik's `__layout` mechanism will be deprecated and replaced "renderer" based. So it is great to be able to use `c.render()` for nested layouts in a smart way.

Yeah, that's a tough one. I presume Sonik works via Vite scanning for layouts higher in the tree and merging them, while Hono doesn't have that same mechanism.

You'd need some way to reference a parent Hono instance from the child to check if a Renderer has been registered.

bruceharrison1984:
After thinking about this a bit, it seems like at a minimum you'd need to make the `renderer` visible outside of the the `Context` object to facilitate messing around with it.

https://github.com/honojs/hono/blob/ff34b1ba9adab6e1c9050c2d6a5ef705de31682e/src/context.ts#L107

[...]

PR #1945, opened 2024-01-11
Nested Layouts via Renderer

bruceharrison1984:
The current implementation does not allow for nested Hono instances to inherit from the parent layout. This causes the child instances to have to duplicate whatever the parent context has for a renderer, and then add code as necessary. This is very inconvenient, causes lots of code duplication, and is extremely error prone if ancestor layouts are changed.

This purpose of this feature is to make nested child layouts possible, as well as easy to work with. Ideally, child instances would automatically inherit the parent Renderer(if it exists) and their content will be passed into it. In essence, recursion-like behavior could be used to crawl from the local context, to the furthest ancestor, and  merge them together.

Created based on the following conversation #1899 

### Notes
- This only works with one level of nesting
  - Nested Hono instances via `app.route('/nested', app2)` only seem to support one level of nesting anyway, so maybe this doesn't matter
    - If this changes, this will need to be corrected so it continues to work. Right now since only one layer of nesting is possible, there isn't a reliable way to test full recusion.
- Should you be able to opt-out of this behavior?

- [ ] Add tests
- [ ] Run tests


bruceharrison1984:
@yusukebe Here's an example of what I was talking about. Overall a pretty small changes, and it works as expected with all tests passing unchanged.

yusukebe:
Thanks! I'll check this later.

yusukebe:
Hello @bruceharrison1984 !

Your ideas and implementation are great. I have added my great idea to it.

The introduction of `<Layout>`. If you do nothing, the nested layout will not be enabled, but with `<Layout>`, you can adapt the upper Layout and enable the nested layout.

```tsx
app.use(
    return (
      <html>
        <body>
          <h1>Hello</h1>
          {children}
        </body>
      </html>
    )
  })
)

const app2 = new Hono()
app2.use(
  jsxRenderer(({ children, Layout }) => {
    return (
      <Layout>
        <div>{children}</div>
      </Layout>
    )
  })
)
```

I already implemented it and pushed it to this PR. I think you will like it, what do you think?

bruceharrison1984:
I really like the adding of `Layout` to the `jsxRenderer` function. It provides both explicit intentional usage, and a natural way to opt-out of the parent layout. I had considered an approach similar to this, but wasn't sure how you'd feel about altering the method signature for `jsxRenderer`, but seems you stumbled upon the same approach.

yusukebe:
I like this too!

Can I merge them?

bruceharrison1984:
I'm fine with it if you are! The implementation is simple enough that if it needs to change later, it should be a light lift.

yusukebe:
@bruceharrison1984 

Okay, I'll merge this now! Thanks.

> It's probably worth updating the docs to also include that nesting of Hono instances only goes one layer deep. It's not a big deal but I was surprised when I discovered it while working on this.

Could you please explain this in a little more detail? You can nest several layers of app instance routes by using the `route` as shown below.

```ts
const app = new Hono()
app.get('/', (c) => c.json(0))

const baz = new Hono()
baz.get('/', (c) => c.json(0))

const bar = new Hono()
bar.get('/', (c) => c.json(0))
bar.route('/baz', baz)

const foo = new Hono()
foo.get('/', (c) => c.json(0))
foo.route('/bar', bar)

app.route('/foo', foo)

showRoutes(app)
```

Screenshot 2024-01-14 at 16 53 04

bruceharrison1984:
Hmm, when I tried what you have as an example, nothing below the second layer would register. Maybe I need to give it another shot.

PR #1949, opened 2024-01-12
Fix/response for lambda@edge adapter

mosle:
In the current implementation, when lambda@Edge is launched with a CloudFront origin request, CloudFront returns an error when processing binary data.

Probably because the body is base64 but there is no bodyEncoding property.

When returning a base64 body, the bodyEncoding property should now include base64.

- [x] Add tests
- [x] Run tests


yusukebe:
Hi @mosle !

Thanks for the PR. Seem to be good.

@watany-dev Could you review this?

watany-dev:
@mosle
Thank you. This is correct. If you run `npm run format:fix` `npm run lint:fix`, the CI error will go away.

yusukebe:
Thanks both. I'll merge this now!

PR #1956, opened 2024-01-12
WIP: feat: filesystem wrapper

nakasyou:
I added filesystem wrapper API.
I think v4 will add features require access to the file system, such as file-based routing and SSGs.
API to access files varies depending on the runtime. This PR will solve this problem.

For example, you can probably use `serveStatic` API:
```ts
import { serveStatic } from 'hono/serve-static'
import { fsWrapper } from 'hono/deno' // or custom adapter

app.use('*', serveStatic(fsWrapper()))
```

Moreover, If apply this API, I can apply SSG API this API.

### Changes
- Add `fsWrapper` API. This api is included in adapters.
fsWrapper provide normalized filesystem APIs.
```ts
const file = await fsWrapper().open('file.txt')
file.readable // ReadableStream
```

- [ ] Add tests
- [ ] Run tests


yusukebe:
Hi @nakasyou 

Thanks for the proposal.

Please give me a moment to imagine real world use cases. While basically it's good, we would like to consider the following points:

* There is no file system on the edge. If we expect only an edge environment, we don't need a file system wrapper.
* File-based routing in v4 is Vite reads the files; Hono itself does not need to handle the file system.
* So if we implement it in the core of hono, the Node.js implementation should be in the core. Move `@hono/node-server` serveStatic to here, so that `@hono/node-server` should be just a server.

nakasyou:
> There is no file system on the edges. If we expect only the edge environments, we don't need a file system wrapper.

I think people don't use Hono on only edges.
Also, this PR isn't for just edges. Initially, this PR was created to be able to run on different runtimes.

> File-based routing in v4 is Vite reads the files; Hono itself does not need to handle the file system.

Thank you for telling about File-based routing!  I understood.

> So if we implement it in the core of hono, the Node.js implementation should be in the core. Move @hono/node-server serveStatic to here, so that @hono/node-server should be just a server.

Yes, I think too.
I think it's smart.

nakasyou:
Sorry, there is SSG adapter in Hono.
So this PR isn't needed.

PR #1957, opened 2024-01-12
chore(package.json): use escape double quotes

ryuapp:
Some commands don't work on Windows because of single quotes.
I replaced them with double quotes.

- [ ] Add tests
- [ ] Run tests


yusukebe:
@ryuapp 

Thanks! Seems to be good. Merge now.

PR #1965, opened 2024-01-13
refactor(ssg): SSG works without `node:path`

### LINKED ISSUE #1964: SSG: Avoid using `path`

Currently, SSG helper is using Node.js `path` module:

https://github.com/honojs/hono/blob/04b686ca39421d28d3aa3f8491c13e1f7d1c8de1/src/helper/ssg/index.ts#L1

But because of this, if it imports `hono/ssg`, the application will not start up with wrangler without the `node_compat` flag:

Screenshot 2024-01-13 at 20 30 40

If you only need to merge paths, you can make it yourself, or it may already be implemented as a utility.

nakasyou:
You can close #1964 If you merge this PR.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/helper/ssg/index.ts
+++ b/deno_dist/helper/ssg/index.ts
@@ -2,38 +2,7 @@ import { Buffer } from "node:buffer";
 import { inspectRoutes } from '../../helper/dev/index.ts'
 import type { Hono } from '../../hono.ts'
 import type { Env, Schema } from '../../types.ts'
-
-export const dirname = (path: string) => {
-  const splitedPath = path.split(/[\/\\]/)
-
-  return splitedPath.slice(0, -1).join('/') // Windows supports slash path
-}
-export const joinPaths = (...paths: string[]) => {
-  const nomalizedPaths: string[] = []
-  for (const path of paths) {
-    const nomalizedPath = path.replace(/(\\)/g, '/').replace(/\/$/g, '')
-    nomalizedPaths.push(nomalizedPath)
-  }
-  const resultPaths: string[] = []
-  for (let path of nomalizedPaths.join('/').split('/')) {
-    // Handle `..` or `../`
-    if (path === '..') {
-      if (resultPaths.length === 0) {
-        resultPaths.push('..')
-      } else {
-        resultPaths.pop()
-      }
-      continue
-    } else {
-      // Handle `.` or `./`
-      path = path.replace(/^\./g, '')
-    }
-    if (path !== ''){
-      resultPaths.push(path)
-    }
-  }
-  return resultPaths.join('/')
-}
+import { joinPaths, dirname } from './utils.ts'
 
 export interface FileSystemModule {
   writeFile(path: string, data: string | Buffer): Promise<void>
--- a/deno_dist/helper/ssg/index.ts
+++ b/deno_dist/helper/ssg/index.ts
@@ -1,8 +1,39 @@
 import { Buffer } from "node:buffer";
-import * as path from 'node:path'
 import { inspectRoutes } from '../../helper/dev/index.ts'
 import type { Hono } from '../../hono.ts'
 import type { Env, Schema } from '../../types.ts'
+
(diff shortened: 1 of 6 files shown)


deno_dist/helper/ssg/index.ts
```diff
@@ -1,9 +1,40 @@
 import { Buffer } from "node:buffer";
-import * as path from 'node:path'
 import { inspectRoutes } from '../../helper/dev/index.ts'
 import type { Hono } from '../../hono.ts'
 import type { Env, Schema } from '../../types.ts'
 
+export const dirname = (path: string) => {
```
yusukebe, line 6: How about moving the `dirname` and `joinPaths` to `utilts.ts`?

nakasyou: I think it's nice idea.
In this case, Should I put files outside `src/helper/ssg`? If I put this code outside `src/helper/ssg`, other codes can easily imports `dirname` and `joinPaths` functions. If you don't need this feature, I put this functions to `src/helper/ssg/utils.ts`.

yusukebe: I think `src/helper/ssg/utils.ts` is better.

If you don't plan to use it in other places right now, it is better to make it exclusive to this feature. You can make changes without worrying about the impact on others.

nakasyou: I understand. I moved that functions.

src/helper/ssg/utils.test.ts
```diff
@@ -0,0 +1,19 @@
+import { describe, it, expect } from 'vitest'
+import { joinPaths, dirname } from './utils'
+
+describe('joinPath', () => {
+  it('Should joined path is valid.', () => {
+    expect(joinPaths('test', 'test2')).toBe('test/test2')
+    expect(joinPaths('test', 'test2', '../test3')).toBe('test/test3')
+    expect(joinPaths('.', '../')).toBe('..')
```
yusukebe, line 8: Could you add these cases to cover all patterns?

```ts
expect(joinPaths('test/', 'test2/')).toBe('test/test2') // trailing slashes
expect(joinPaths('./test', './test2')).toBe('test/test2') // dot and slash
expect(joinPaths('', 'test')).toBe('test') // empty path
expect(joinPaths('/test', '/test2')).toBe('test/test2') // root path
```

EdamAme-x: I further believe that the following cases are also necessary.
```typescript
expect(joinPaths('../', 'test')).toBe('../test') // parent and single
```

nakasyou: @yusukebe @EdamAme-x I added those test code, and passed test.

@yusukebe
```ts
expect(joinPaths('/test', '/test2')).toBe('test/test2') // root path
```
Is this code valid?
It test is successful, but I think `joinPaths('/test', '/test2')` should be `/test2`

nakasyou: > Is this code valid?
> It test is successful, but I think joinPaths('/test', '/test2') should be /test2

I tried this code using `node:path`, the result is `/test/test2`. So I will fix code.

nakasyou: I fixed.

EdamAme-x:
It appears that the spec does not allow me to do what I have in mind, so here is my suggestion.

This code could be made as concise as the following
https://github.com/EdamAme-x/hono/blob/refactor/ssg-without-node-path/src/helper/ssg/utils.ts

EdamAme-x:
Of course, all test cases are passed.

yusukebe (approved):
LGTM!

yusukebe:
@nakasyou 

Thanks! I'll merge this now.

@EdamAme-x 

Thank you for the suggestion. I like your code, it's very clear to understand. In this case, I will accept this PR code since we only need to be able to keep the specs by testing. Please contribute again!

EdamAme-x:
thanks.
i do it after this

Issue #1944, opened 2024-01-11
`dangerouslySetInnerHTML` is not working in `hono/jsx/dom`


yusukebe:
### What version of Hono are you using?

4.0.0-rc.0

### What runtime/platform is your app running on?

any

### What steps can reproduce the bug?

```tsx
import { render } from 'hono/jsx/dom'

const root = document.getElementById('root') as HTMLElement

const Div = () => {
  const inner = { __html: '<h1>Hello</h1>' }
  return <div dangerouslySetInnerHTML={inner} />
}

render(<Div />, root)
```

### What is the expected behavior?

Render the content:

```html
<div><h1>Hello</h1></div>
```

### What do you see instead?

```html
<div dangerouslysetinnerhtml="[object Object]"></div>
```
<img width="829" alt="Screenshot 2024-01-11 at 23 27 06" src="https://github.com/honojs/hono/assets/10682/03e38379-066b-4576-b066-fe5f3fa455fc">

EdamAme-x:
https://github.com/EdamAme-x/hono/blob/5ba31090dabe4dd275babde22cfb427167ee4add/src/jsx/dom/index.ts#L348

It looks like it could be implemented here.

yusukebe:
@EdamAme-x 

Thanks. But this will be fixed by #1955.

yusukebe:
Fixed, thanks!

Issue #1943, opened 2024-01-11
boolean attribute as `false` problem in `hono/jsx/dom`


yusukebe:
### What version of Hono are you using?

4.0.0-rc.0

### What runtime/platform is your app running on?

any

### What steps can reproduce the bug?

With `hono/jsx/dom`:

```tsx
import { render } from 'hono/jsx/dom'

// ...

const CheckBox = () => {
  return (
    <div>
      <label>
        Check
        <input type="checkbox" checked={false} />
      </label>
    </div>
  )
}

render(<CheckBox />, root)
```

### What is the expected behavior?

<img width="321" alt="Screenshot 2024-01-11 at 23 09 48" src="https://github.com/honojs/hono/assets/10682/8a5788d6-7601-457a-b367-fd872c30fcb7">

### What do you see instead?

<img width="668" alt="Screenshot 2024-01-11 at 23 06 25" src="https://github.com/honojs/hono/assets/10682/842c2af0-22f8-4968-9928-2a09dadc2d66">

yusukebe:
Fixed!

Issue #1942, opened 2024-01-11
`onChange` behavior


yusukebe:
### What version of Hono are you using?

4.0.0-rc.0

### What runtime/platform is your app running on?

any

### What steps can reproduce the bug?

This is not a bug, but a request.

With the following code, the behavior of the `onChange` event is different from React.

```tsx
import { render } from 'hono/jsx/dom'
import { useState } from 'hono/jsx/hooks'

const root = document.getElementById('root') as HTMLElement

const TextArea = () => {
  const [value, setValue] = useState('')

  const handleChange = (e: Event) => {
    const target = e.target as HTMLTextAreaElement
    setValue(target.value)
  }

  return (
    <div>
      <textarea onChange={handleChange} />
      <div>{value}</div>
    </div>
  )
}

render(<TextArea />, root)
```

### What is the expected behavior?

This is the React code and the demo.

https://github.com/honojs/hono/assets/10682/bfb370fa-2c76-4444-9196-d49208ee7809

```tsx
import React, { useState } from 'react'
import { hydrateRoot } from 'react-dom/client'

const root = document.getElementById('root') as HTMLElement

const TextArea = () => {
  const [value, setValue] = useState('')

  const handleChange = (e: Event) => {
    const target = e.target as HTMLTextAreaElement
    setValue(target.value)
  }

  return (
    <div>
      <textarea onChange={handleChange} />
      <div>{value}</div>
    </div>
  )
}

hydrateRoot(root, <TextArea />)
```

### What do you see instead?

This is the `hono/jsx/demo` demo.  I need to click outside of the `textarea` to trigger the `onChange` event:

https://github.com/honojs/hono/assets/10682/b24323be-77ef-4f3d-82b6-1423dc595561

### Additional information

This may be a specification issue, but since I want to get a value for each input, I think it should match React's behavior.

usualoma:
👀

yusukebe:
@usualoma 

What do you think of this?

yusukebe:
Fixed! But, another issue came up: #2016

Issue #1941, opened 2024-01-11
`render` should handle a fragment node


yusukebe:
### What version of Hono are you using?

4.0.0-rc.0

### What runtime/platform is your app running on?

any

### What steps can reproduce the bug?

```tsx
import { render } from 'hono/jsx/dom'
import { Fragment } from 'hono/jsx'

const root = document.getElementById('root') as HTMLElement
function FC() {
  return (
    <Fragment>
      <b>bar</b>
    </Fragment>
  )
}
render(<FC />, root)
```

### What is the expected behavior?

Don't throw an error.

### What do you see instead?

The error in the console:

```txt
Uncaught DOMException: Failed to execute 'createElement' on 'Document': The tag name provided ('') is not a valid name.
```

yusukebe:
Hello @usualoma 

Could you see this?

usualoma:
@yusukebe 
OK, I'm working on being able to Suspense and ErrorBoundary now, so we'll deal with it together there, just give me a minute.

```ts
const Component = () => {
  return <div>0</div>
}

const AsyncComponent = async () => {
  const num = await new Promise((resolve) => setTimeout(() => resolve(1), 1000))
  return <div>{num}</div>
}

const promise = new Promise((resolve) => setTimeout(() => resolve(2), 2000))
const UseComponent = () => {
  const num = use(promise)
  return <div>{num}</div>
}

const SyncErrorComponent = () => {
  throw new Error('error')
}

const AsyncErrorComponent = async () => {
  await new Promise((resolve) => setTimeout(() => resolve(1), 3000))
  throw new Error('async error')
}

export const App = () => {
  const [count, setCount] = useState(0)
  return (
    <div>
      <div>
        <h1>Count: {count}</h1>

        <button onClick={() => setCount(count + 1)}>+1</button>

        <div>
          <div>
            Component
            <Component />
          </div>
        </div>

        <div>
          <Suspense fallback={<div>Loading AsyncComponent ...</div>}>
            <div>
              AsyncComponent
              <AsyncComponent />
            </div>
          </Suspense>
        </div>

        <div>
          <Suspense fallback={<div>Loading UseComponent ...</div>}>
            <div>
              UseComponent
              <UseComponent />
            </div>
          </Suspense>
        </div>

        {/* <div>
          <ErrorBoundary fallback={<div>error</div>}>
            <div>no error</div>
            <div>3</div>
          </ErrorBoundary>
        </div> */}

        <div>
[...]

PR #1981, opened 2024-01-15
feat(jsx/dom): rewrite renderer to use virtual tree

usualoma:
By #1917 and #1955 we found that we could achieve "render to dom" with our own implementation and it made sense, but it was, to put it mildly, a rather dirty implementation and an impossible one to maintain.

As mentioned on https://github.com/honojs/hono/pull/1909#issuecomment-1880043938, I think the cost of implementation is high, but still, without a virtual tree approach, the code for this feature is I don't think it would be maintainable.

So, I would like to replace it with this implementation that creates a virtual tree.

#### Async component is not supported

Unfortunately, as with the original React, it is not possible to use async components when rendering to dom. I was very worried about this, but I still think it would be better not to support this because it would not be possible to match the policy for rendering JSX to DOM.

```ts
const Component = async () => {
  const num = await promise
  return <div>{num}</div>;
};
```

`use` hook is available and should be used when rendering to dom.

```ts
const Component = () => {
  const num = use(promise)
  return <div>{num}</div>;
};
```

### Small

"src/jsx/dom/render.ts"  is very small, about 3 Kb.

```
$ npx esbuild --minify src/jsx/dom/render.ts | wc
       1      66    3132
```

- [x] Add tests
- [x] Run tests


usualoma:
This PR also fixes these issues.

Fixes https://github.com/honojs/hono/issues/1944
Fixes https://github.com/honojs/hono/issues/1941
Fixes https://github.com/honojs/hono/issues/1942
Fixes https://github.com/honojs/hono/issues/1943

yusukebe:
@usualoma 

It's amazingly small! It is good that the implementation is small and easy to maintain. It's good if working AsyncComponent with `use`.

What I am concerned about is the case where the server and the client see the same component. This is an important issue for Islands Architecture, which will be introduced in v4.

I tried this PR code, but, of course, the `use` did not work on the server. The code is below.

```tsx
import { useState, use } from 'hono/jsx/hooks'

export default function Counter() {
  const promise = new Promise((resolve) => setTimeout(resolve, 2000))
  use(promise)
  const [count, setCount] = useState(0)
  return (
    <div>
      <p>Count: {count}</p>
      <button onClick={() => setCount(count + 1)}>Increment</button>
    </div>
  )
}
```

Repo: https://github.com/yusukebe/hono-jsx-front-poc/tree/dom

What we need to consider is the `use` specification. If we use `use` on the server side as well, should it wait until Promise resolves it or not? If we follow the client's spec, it should wait. But, for example, it might be a good UI if the server does **not** wait for `use` to resolve, but returns HTML and waits at the client.

What do you think about the server and client referring to the same component regardless of `use`?

usualoma:
### Behavior of `use()` in React

#### render() to DOM

This code does not work

```ts
export default function Counter() {
  const promise = new Promise((resolve) => setTimeout(resolve, 2000))
  use(promise)
  const [count, setCount] = useState(0)
  return (
    <div>
      <p>Count: {count}</p>
      <button onClick={() => setCount(count + 1)}>Increment</button>
    </div>
  )
}
```

We need to pass the same same object to `use()`.

```ts
const promise = new Promise((resolve) => setTimeout(resolve, 2000)) // Define outside the component or cache it somewhere

export default function Counter() {
  use(promise)
  const [count, setCount] = useState(0)
  return (
    <div>
      <p>Count: {count}</p>
      <button onClick={() => setCount(count + 1)}>Increment</button>
    </div>
  )
}
```

#### renderToString()

`renderToString()` does not support `Suspense`

#### renderToReadableStream()

This code works.

```ts
export default function Counter() {
  const promise = new Promise((resolve) => setTimeout(resolve, 2000))
  use(promise)
  const [count, setCount] = useState(0)
  return (
    <div>
      <p>Count: {count}</p>
      <button onClick={() => setCount(count + 1)}>Increment</button>
    </div>
  )
}
```

### Behavior of `use()` in hono/jsx

#### render() to DOM

As with React, the former code will not work. The latter code works.

#### toString()

at caba00e, the former code will not work. The latter code works. I believe you saw this and commented.

at f7070ae, also the former code works like React's `renderToReadableStream()`


### Conclusion

If the result should be the same whether rendering to DOM or toString(), then the code in caba00e is sufficient and f7070ae is unnecessary.
However, I think that f7070ae is also useful, so I think it is OK to include it.

usualoma:
@yusukebe 
If you do not see a problem, you may have this PR merged.
This PR is still not that small because bundling always introduces unnecessary code for `toString()`, but #1986 will allow bundling of only the dom render, which is expected to make it very small.

yusukebe:
@usualoma 

Thanks! 

I am wondering if the counter sample in the following repository will work or not. From what I have tried, it does not work with the latest code from this PR. I think this should work, am I wrong? I would be glad if you could check.

https://github.com/yusukebe/hono-jsx-front-poc/tree/dom

usualoma:
@yusukebe 
OK, `use()` should be used in `Suspense`.

```ts
import { Suspense } from 'hono/jsx'
import { useState, use } from 'hono/jsx/hooks'

const promise = new Promise((resolve) => setTimeout(resolve, 2000))

function Counter() {
  use(promise)
  const [count, setCount] = useState(0)
  return (
    <div>
      <p>Count: {count}</p>
      <button onClick={() => setCount(count + 1)}>Increment</button>
    </div>
  )
}

export default function App() {
  return (
      <Counter />
    </Suspense>
  )
}
```

yusukebe:
@usualoma 

Thanks! I would like to do the following.

* Server and client use the same AsyncComponent.
* The server returns AsyncComopent without waiting. In this case, it returns the contents of the `Suspense` fallback.
* Client displays the contents of the fallback until the Promise is resolved, then displays it resolved.

In the following code, the application waits 2 seconds on the server side for the first access and then waits 2 seconds on the client. On the second and after, the app waits 2 seconds on the client only.

What I'm expecting is this "behavior after the second time". This means that the client only waits 2 seconds, the server not wait  - In actuality, the server returns the contents of the fallback, and the client also initially renders the fallback and then shows the contents after 2 seconds.

```tsx
// counter.tsx
import { use } from 'hono/jsx/hooks'

const promise = new Promise((resolve) => setTimeout(resolve, 2000))

export default function Component() {
  use(promise)
  return <p>Done!</p>
}
```

```tsx
// server.tsx
app.get('/', (c) => {
  return c.render(
    <div id="target">
      <Suspense fallback={'on server...'}>
        <Component />
      </Suspense>
    </div>
  )
})
```

```tsx
// client.tsx
document.addEventListener('DOMContentLoaded', function () {
  const target = document.getElementById('target')
  render(
    <Suspense fallback={'on client...'}>
      <Component />
    </Suspense>,
    target
  )
})
```

https://github.com/honojs/hono/assets/10682/d26f4327-de32-4d07-b94a-eec005c11d06

This behavior is expected according to the specification. Therefore, it may not be necessary to change it. However, I cannot realize what I want to do. What do you think about this matter? Or do I expect the wrong things? I'm sorry to bother you.

yusukebe:
Or it may be better to separate the roles of the server and client components as follows:

* Handling AsyncComponent with Suspense - e.g. fetch external resources => Server Components
* Build a UI with interactions => Client Components

A quick search shows that Next.js has a similar approach.

yusukebe:
@usualoma 

I'm sorry. Please ignore what I have written above!

It is a very rare case that this server and client reference the same AsynComponent and SSR that hydrates it. I think it is not used very often (especially in the file-based routing framework we are trying to create) and should not be recommended.

I will merge if you are OK!

usualoma:
@yusukebe 
Thanks for checking!
Let's follow up on any improvements that need to be made. As for this PR, I am fine to have this merged.

yusukebe:
Yeah! I'll merge this now.

PR #1989, opened 2024-01-15
refactor: faster for loop

EdamAme-x:
For loop use the length of the array cached first.

- [ ] Add tests
- [x] Run tests


yusukebe:
@EdamAme-x 

Thanks! I think this is a good improvement. Merge now.

Issue #1966, opened 2024-01-13
SSG: Handling wildcards, path params, regexp routing


yusukebe:
### What is the feature you are proposing?

If the following routing is used, an unexpected file will be generated.

```ts
app.get('/shops/:id', (c) => c.html(''))
app.get('/shops/:id/:comments{[0-9]+}', (c) => c.html(''))
app.get('/foo/*', (c) => c.html(''))
```

<img width="547" alt="Screenshot 2024-01-13 at 22 51 22" src="https://github.com/honojs/hono/assets/10682/ed9ae44a-129e-4d3b-92c1-4cdf40ae369e">

If we don't specify the URL to access with a feature like `ssgParams`, we can ignore the generation of these files.

yusukebe:
Fixed by #1990

PR #1960, opened 2024-01-13
feat: Added `ssgParams` middleware

nakasyou:
I added `ssgParams` middleware.
This is implemented https://github.com/honojs/hono/pull/1904#issuecomment-188809839.

There is 2 changes.

#### 1. API like `generateStaticPaths`.
You can use API like `generateStaticPaths`.
Example:
```tsx
import { Hono } from 'hono'
import { toSSG, ssgParams } from 'hono/ssg'
import * as fs from 'fs/promises'

app.get('/', c => c.html(<h1>Top page</h1>))
app.get('/post/:id', ssgParams([{ id: '1' }]), c => c.html(<h1>{c.req.params('id')}</h1>)

await toSSG(app, fs)
```
#### 2. Switching SSR/SSG
You can add No-SSG routes such as:
```tsx
import { Hono } from 'hono'
import { toSSG, ssgParams } from 'hono/ssg'
import * as fs from 'fs/promises'

app.get('/', c => c.html(<h1>Top page</h1>)) // SSG
app.get(
  '/api',
  ssgParams(false), // Don't SSG
  c => c.json({ status: 'ok' })
) 

await toSSG(app, fs)
```
This code is SSG-based mode like [Astro hybrid mode](https://docs.astro.build/en/guides/server-side-rendering/#enable-on-demand-server-rendering).

Not only this code, you can use SSR-Based mode.
```tsx
import { Hono } from 'hono'
import { toSSG, ssgParams } from 'hono/ssg'
import * as fs from 'fs/promises'

app.get('/',
  ssgParams(true), // SSG
  c => c.html(<h1>Top page</h1>)
) // SSG
// Or, this code also works.
app.get('/post/:id', ssgParams([{ id: '1' }]), c => c.html(<h1>{c.req.params('id')}</h1>))

app.get(
  '/api',
  c => c.json({ status: 'ok' })
) // This route will not generate SSG files. `toSSG` don't generate route if not explicitly specified.

await toSSG(app, fs, { default: 'ssr' })
```

Hereby, you can write code to combine both SSG and SSR such as:
```tsx
// dev.tsx
import { Hono } from 'hono'
import { ssgParams } from 'hono/ssg'

const app = new Hono()

app.get('/', ssgParams(true), c => c.html(<h1>Hello world!</h1>)) // Top page is static.
app.get('/api/random', c => c.json({ result: Math.random() })) // API is dynamic.

export default app
```
```tsx
// prediction.ts
import { Hono } from 'hono'
import app from './dev'

[...]

PR #1990, opened 2024-01-16
feat(ssg): Ignore Dynamic Route

watany-dev:
Resolves https://github.com/honojs/hono/issues/1966

We have added an implementation that excludes certain types of routes - specifically, wildcard, path parameters, and regex-based routing - from the SSG (Static Site Generation) process.

It's important to note that this does not take into account the ssgParams middleware. Depending on the timing of the merge, we can include support for this middleware in this PR if necessary.

- [x] Add tests
- [x] Run tests


deno_dist/helper/ssg/index.ts
```diff
@@ -78,17 +78,41 @@ export const fetchRoutesContent = async <
...
+    return { content, mimeType }
+  } catch (error) {
+    console.error(`Error fetching content for path: ${path}`, error)
+    return null
   }
+}
 
-  return htmlMap
+const isDynamicRoute = (path: string): boolean => {
+  return /:\w+|\*/.test(path)
```
yusukebe, line 115: The determination of the named path parameters here is incorrect.

`/foo:bar` is not a dynamic route. You can see the spec here:

https://github.com/honojs/hono/blob/a00fc566e93374acd4c8999443f4b48f79da05b9/src/utils/url.ts#L55

So you have to split the path with `/` and determine if there is a `:` at the beginning of it or not.

deno_dist/helper/ssg/index.ts
```diff
@@ -78,17 +78,41 @@ export const fetchRoutesContent = async <
...
+
+  return htmlMap
+}
+
+interface RouteContent {
+  content: string | ArrayBuffer
+  mimeType: string
+}
+
+const fetchRouteContent = async <E extends Env = Env, S extends Schema = {}>(
```
yusukebe, line 97: Should you include these `fetchRouteContent` things in this PR?

yusukebe:
Let's complete this PR first, and merge it into v4. I left a comment; please take a look.

deno_dist/helper/ssg/index.ts
watany-dev: I don't need it, so I reverted it.

deno_dist/helper/ssg/index.ts
watany-dev: fixed.

yusukebe (approved):
LGTM!

yusukebe:
Hi @watany-dev 

I've merged #1960 into v4 now, but the CI fails:

https://github.com/honojs/hono/actions/runs/7553089110/job/20563202188

I think `isDynamicRoute` causes the mismatch. Could you see it?

watany-dev:
@yusukebe 

In the short term, revert my commits and it will be cured.
If I fix this, I will need to maintain hono/dev and identify the route where ssgparam is set and do some additional processing. This will take a bit of time.

PR #1991, opened 2024-01-16
feat: Add Type to request for Lambda@Edge adapter to support S3 origin

mosle:
The origin field in the request object in the Lambda@Edge adapter is assumed to be only a custom origin.

Referring to 
https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-event-structure.html ,
I have defined a TypeScript type that is also compatible with S3 origins.

- [ ] Add tests
- [x] Run tests


yusukebe:
Thanks. Merge now.

PR #1995, opened 2024-01-16
fix(types): correct `c.get()` inference

yusukebe:

- [x] Add tests
- [x] Run tests


yusukebe:
We can go with this.

PR #1997, opened 2024-01-16
refactor(jsx): export components and hook function from top level 'hono/jsx'

usualoma:
+ 15204380da15ef4ffea341475da3f22d7a7c56e1

Up to this point, hook functions have been exported from "hono/jsx/hooks", but today, when tree shaking is commonplace, there is little motivation to separate export source modules.
It would be better if all JSX-related components and hook functions could be imported from "hono/jsx".

```ts
import {
  Suspense,
  ErrorBoundary,
  useState,
  useEffect,
  useRef,
  useCallback,
  use,
  startTransition,
  useTransition,
} from 'hono/jsx'
import type { RefObject } from 'hono/jsx'
```

As for dom-related or dom-specific items, they can continue to be imported as follows.

```ts
export { render, Suspense, ErrorBoundary } from 'hono/jsx/dom'
```


yusukebe:
Hi @usualoma 

I think this is a good improvement. With our Hono, it's important to make it "easy".

When ready, I'll merge!

usualoma:
Reapply "refactor(jsx): export components and hook function from top level 'hono/jsx'" at 0fb71c6
And adds some minor refactoring.

This completes the work on this PR.

yusukebe:
@usualoma 

Thanks! Merging now.

PR #1999, opened 2024-01-17
feat(jsx/dom): support createContext and useContext in jsx/dom

usualoma:
```ts
import { createContext, useContext } from 'hono/jsx'
// or dom specific version
// import { createContext, useContext } from 'hono/jsx/dom'

const Context = createContext(0)
const Content = () => {
  const num = useContext(Context)
  return <p>{num}</p>
}
const Component = () => {
  return (
    <>
      <Context.Provider value={1}>
        <Content />
        <Context.Provider value={3}>
          <Content />
        </Context.Provider>
        <Content />
      </Context.Provider>
    </>
  )
}
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/jsx/context.ts
+++ b/deno_dist/jsx/context.ts
@@ -1,0 +1,47 @@
+import { raw } from '../helper/html/index.ts'
+import type { HtmlEscapedString } from '../utils/html.ts'
+import { createContextProviderFunction } from './dom/context.ts'
+import { RENDER_TO_DOM } from './dom/render.ts'
+import { JSXFragmentNode } from './index.ts'
+import type { FC } from './index.ts'
+
+export interface Context<T> {
+  values: T[]
+  Provider: FC<{ value: T }>
+}
+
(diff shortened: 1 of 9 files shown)


yusukebe:
Hi @usualoma !

Does this mean that this `useContext` can be used from both the server and DOM? It works fine, but I am a little reluctant to use `jsx/dom/context.ts` for DOM in the `jsx/index.ts`.  Or do I understand this wrong?

usualoma:
@yusukebe Thanks for pointing that out. I think that is indeed true.

> Does this mean that this useContext can be used from both the server and DOM?

Yes, it is.

In 05e0b5a, I have re-organized the placement of components in files.
How about this?

yusukebe:
@usualoma 

> In 05e0b5a, I have re-organized the placement of components in files.
> How about this?

Thanks! I like this! Let's go with it.

usualoma:
@yusukebe 
Thanks for confirming.
The test failing is a HEAD issue on the v4 branch, so I think it is safe to have this PR merged.

PR #2003, opened 2024-01-17
refactor(jsx/dom): make `useTransition()` handling more simple

usualoma:
After relearning about TaskQueue and Microtask in JavaScript, I understood that execution by `setTimeout()` does not overtake the first `update()`. There is no need to use `promise` here.
Also, I don't think a 10ms wait is necessary.

- [x] Add tests
- [x] Run tests


yusukebe (approved):
LGTM!

PR #2005, opened 2024-01-17
Refactor middleware, tests

ariskemper:

- [x] Add tests
- [x] Run tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,8 @@
     "dist"
   ],
   "scripts": {
-    "test": "tsc --noEmit && vitest --run",
+    "test": "vitest --run",
+    "test:watch": "vitest --watch",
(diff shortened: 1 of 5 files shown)


yusukebe:
Hi @ariskemper 

Thanks. I want to merge this. So, please run the `yarn format:fix`  to format the code.

yusukebe:
@ariskemper 

Thanks, but please `denoify` again!

package.json
```diff
@@ -10,7 +10,8 @@
     "dist"
   ],
   "scripts": {
-    "test": "tsc --noEmit && vitest --run",
```
yusukebe, line 13: @ariskemper 

Sorry, I didn't notice the change here.

`tsc --noEmit` is necessary to check the types in the test phase.

package.json
```diff
@@ -10,7 +10,8 @@
...
+    "test": "vitest --run",
+    "test:watch": "vitest --watch",
```
yusukebe, line 14: This change is not bad, but we want to be careful about changing `package.json`, so we should not introduce it in this PR.

ariskemper: @yusukebe will create a separate PR for this change, but i know vitest, that's why i made this change :)

yusukebe:
@ariskemper 

Thanks! I'll merge this now.

PR #2012, opened 2024-01-18
refactor(ssg): Removal of Libraries Dependent on Node.js

watany-dev:
These modifications were made to eliminate the dependency on Node.js. 

- Changes in the fs for the Bun Adaptor.
- Revision of the Buffer type.
- Changed to 'toDenoSSG', 'toBunSSG' for each adapter to address duplicate naming of toSSG for deno.

- [x] Add tests
- [x] Run tests


yusukebe:
Hi @watany-dev !

Thanks for the PR.

I think you can use `@types/bun` instead of `bun.d.ts` for only types matter. In the core of Hono, you are not allowed to use specific platform types. However, in this case, it is an adapter for Bun, so you may use it.

Actually, I think you can install `@types/bun` as `devDependencies` and write triple slash directives in `ssg.ts`.

```ts
/// <reference types="bun" />
```

src/adapter/bun/ssg.ts
```diff
@@ -1,28 +1,28 @@
 // @denoify-ignore
 /* eslint-disable @typescript-eslint/ban-ts-comment */
-import { mkdir, writeFile } from 'fs/promises'
 import { toSSG as baseToSSG } from '../../helper/ssg'
 import type { FileSystemModule, ToSSGAdaptorInterface } from '../../helper/ssg'
 
+// @ts-ignore
+const { write } = Bun
```
watany-dev, line 7: The import is the same as for the Bun.file because it took a long time to break other parts of the Bun type when it was introduced.

watany-dev:
@yusukebe 
I am having trouble writing RuntimeTest because I am not familiar with Deno. Is there any chance you could confirm how to write it for me since I am not in a hurry?

deno_dist/adapter/deno/index.ts
```diff
@@ -1,2 +1,2 @@
 export { serveStatic } from './serve-static.ts'
-export { toSSG, denoFileSystemModule } from './ssg.ts'
+export { toDenoSSG, denoFileSystemModule } from './ssg.ts'
```
yusukebe, line 2: I would prefer to keep `toSSG`. If there is any confusion, the user can add `as`. `serveStatic` is also on a different platform but has the same naming convention.

```ts
import { toSSG as toBunSSG } from 'hono/bun'
```

yusukebe:
Hi @watany-dev !

I made CI pass!

watany-dev:
Thanks for the great job!!! We will deal with the rest!

deno_dist/adapter/deno/index.ts
watany-dev: OK, the name has been restored.

deno_dist/adapter/deno/ssg.ts
```diff
@@ -18,7 +19,7 @@ export const denoFileSystemModule: FileSystemModule = {
 
 /**
  * @experimental
- * `toSSG` is an experimental feature.
+ * `toDenoSSG` is an experimental feature.
```
yusukebe, line 22: This should be `toSSG`.

src/adapter/deno/ssg.ts
```diff
@@ -18,7 +19,7 @@ export const denoFileSystemModule: FileSystemModule = {
 
 /**
  * @experimental
- * `toSSG` is an experimental feature.
+ * `toDenoSSG` is an experimental feature.
```
yusukebe, line 22: This should be `toSSG`.

src/adapter/bun/ssg.ts
```diff
@@ -1,26 +1,26 @@
...
+  writeFile: async (path, data) => {
+    await write(path, data)
   },
+  mkdir: async () => {},
 }
 
 /**
  * @experimental
- * `toSSG` is an experimental feature.
+ * `toBunSSG` is an experimental feature.
```
yusukebe, line 23: This should be `toSSG`.

yusukebe (commented):
Hi @watany-dev 

I've left comments. Check them!

watany-dev:
@yusukebe 
Sorry for being so careless. I have reverted the name of the comment.

yusukebe (approved):
Looks great!

yusukebe:
Thanks. Merge now!

PR #2013, opened 2024-01-18
perf(ssg): make route requests in parallel in default and add interval option for origin's ratelimit

sor4chi:

To increase the speed of ssg, all requests were made in parallel and used as the default.
On the other hand, this may result in multiplexing of requests and overloading the external API. Therefore, we made it possible to choose whether to parallelize requests or not, and also to allow for intervals between queries.

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/deno_dist/helper/ssg/index.ts
+++ b/deno_dist/helper/ssg/index.ts
@@ -83,7 +83,19 @@ export const ssgParams: SSGParamsMiddleware = (generateParams) => async (c, next
 }
 
 export interface ToSSGOptions {
+  /** The directory to save the generated files. */
   dir?: string
+  /** Whether to fetch each route in parallel. */
+  parallel?: boolean
+  /** The interval between requests. (enabled only when `parallel` is `false`) */
+  interval?: number
+}
+
+export interface FetchRoutesContentOptions {
+  /** Whether to fetch each route in parallel. */
+  parallel: boolean
+  /** The interval between requests. (enabled only when `parallel` is `false`) */
+  interval: number
 }
 
 /**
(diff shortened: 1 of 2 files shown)


yusukebe:
@sor4chi 

Thanks. Very interesting idea.

@watany-dev @nakasyou What do you think about this feature?

watany-dev:
It is a wonderful speed-up. I think we should definitely adopt it.
One thing, this may be overkill, but it would be nice to have an upper limit of 10 for when a `paralel` comes in too large, or to split it up into chunks.

```ts
const batchSize = 10

if (options.parallel) {
  for (let i = 0; i < forGetInfoURLRequest.ssgParams.length; i += batchSize) {
    const batch = forGetInfoURLRequest.ssgParams.slice(i, i + batchSize)
    await Promise.all(batch.map(execDynamicRequest))
  }
} else {
…
```

sor4chi:
Surely that could be a good idea.

There is another idea.
A route with dynamic parameters is likely to throw a request to the same external resource, so the idea is to specify request control in ssgParam.

```ts
app.get('/', (c) => c.html('<h1>Hello</h1>')) // render this route in parallel
app.get(
  '/:id',
  ssgParam([1, 2, 3], {
    parallel: false,
  }),
  (c) => c.html(`<h1>Hello ${c.req.param('id')}</h1>`) // render this route in series
)
```
I would like to hear @yusukebe and @nakasyou's opinion if they prefer that interface.

yusukebe:
That is useful, but I don't want to put too many build settings in `ssgParams`. But, I just can't think of any other good solution.

yusukebe:
@sor4chi 

Do you know how other SSG frameworks are doing these rate-limit things?

sor4chi:
@yusukebe 
I don't know much about it and am looking into it, let me know if I need to rush for hono@v4.

yusukebe:
@sor4chi 

I am not sure if we will implement it or not, but I would like to know as soon as possible. However, I will leave that up to you.

yusukebe:
@sor4chi 

I think this PR idea is great. However, it will not be ready in time for the v4 release, so merging should be done after that.

PR #2018, opened 2024-01-18
fix(types): use middlewares with different env types

agatan:
Hello!
First off, thank you for creating Hono. I'm using it in my personal projects and it's been a great experience!

This pull request aims to improve type inference when extending context variables in middlewares.

Currently, When using handlers like `app.get` to register middlewares that extends context variables, the context type is correctly inferred.

```typescript
const app = new Hono()
const mw = createMiddleware<{ Variables: { foo: string } }>(async () => {});

app.get(mw, async (c) => {
    // this is correctly typed.
    const foo: string = c.var.foo
})
```

However, when using `app.use`, the types aren't as well defined.

```typescript
app.use(mw).get(async (c) => {
    // this isn't correctly typed before merging this PR.
    const foo: string = c.var.foo
})
```

While `app.use` can accept path parametrers like `app.use('/foo', ...)`, but the type inference in these cases is quite complex and isn't covered in this PR.

I believe this improvement will make Hono even more robust for TypeScript users. Looking forward to your feedback!

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/hono-base.ts
+++ b/deno_dist/hono-base.ts
@@ -123,7 +123,8 @@ class Hono<
       handlers.map((handler) => {
         this.addRoute(METHOD_NAME_ALL, this.#path, handler)
       })
-      return this
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      return this as any
     }
 
     const strict = options.strict ?? true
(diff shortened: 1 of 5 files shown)


agatan:
Sorry, I forgot run `yarn denoify` after `yarn format:fix`. I fixed it and force-pushed.

yusukebe (approved):
LGTM!

agatan:
Thank you for the quick response! I'm looking forward to the release of v4!

PR #2020, opened 2024-01-18
feat(jsx/dom): Introduce startViewTransition()

usualoma:
### Included in this PR

* Introduce `startViewTransition()`
* Partial support for `<Style />`

### `startViewTransition()`

This is an API not found in the original React.
The usage of this API is the same as the existing `startTransition()`, with state updates wrapped and passed as callback

#### Is this API something that is needed?

If you simply want to use `document.startViewTransition()`, you can do so by writing the following, although it is a bit long.

```ts
<button onClick={() => {
  if (!document.startViewTransition) {
    setCount((c) => c + 1)
  }
  else {
    document.startViewTransition(() => {
      setCount((c) => c + 1)
    });
  }
}}>
  {count}
</button>
```

Replacing this with the `startViewTransition()` provided by hono here has the following two advantages.

* Can be written in short
* Transition can be applied to fallback by `Suspense` and subsequent delayed rendering.

```ts
<button onClick={() => {
  startViewTransition(() => {
    setCount((c) => c + 1)
  });
}}>
  {count}
</button>
```

#### Restrictions

Currently, it cannot be used with `useTransition()` at the same time. However, I am considering making it possible to use it at the same time in subsequent PRs.

### Partial support for `<Style />`

`<Style />` is not likely to be rendered dom in a production environment, but it is useful in a demo environment, so I partially supported it. API for passing global CSS to children is not yet supported.

### Demo

https://github.com/honojs/hono/assets/30598/7ffecf98-3c5b-4433-9655-00b27538fc8d

```ts
import { Style, css } from './src/helper/css'
import { render, Suspense } from './src/jsx/dom'
import { useState, startViewTransition, use } from './src/jsx/hooks'

const imagePromise = new Map<string, Promise<string>>()
const loadImage = (src: string) => {
  if (!imagePromise.has(src)) {
    imagePromise.set(
      src,
      new Promise((resolve) => {
        ;(async () => {
          await new Promise((resolve) => setTimeout(resolve, 1000))
          const image = new window.Image()
          image.onload = () => resolve(src)
          image.src = src
        })()
      })
    )
  }
  return imagePromise.get(src) as Promise<string>
}

const ThumbnailImage = () => <img src='https://avatars.githubusercontent.com/u/98495527' />
const OriginalImage = () => {
  const image = use(
    loadImage('https://github.com/honojs/hono/blob/main/docs/images/hono-logo.png?raw=true')
  )
  return (
    <div
      class={css`
        background: url('${image}');
        background-size: contain;
        background-repeat: no-repeat;
        background-position: center;
        width: 600px;
        height: 600px;
      `}
    />
  )
}

const Image = () => {
  const [showOriginalImage, setShowOriginalImage] = useState(false)
  return (
    <div
      onClick={() => startViewTransition(() => setShowOriginalImage(!showOriginalImage))}
      class={css`
        img,
        div {
          view-transition-name: image;
        }
      `}
    >
      {showOriginalImage ? (
        <Suspense
          fallback={
            <div
              class={css`
                width: 600px;
                height: 600px;
                display: flex;
                justify-content: center;
                align-items: center;
                background-image: url('https://avatars.githubusercontent.com/u/98495527');
                background-size: contain;
                background-repeat: no-repeat;
                background-position: center;
                background-color: rgba(255, 255, 255, 0.5);
                background-blend-mode: lighten;
              `}
            >
              Loading...
            </div>
          }
        >
          <OriginalImage />
        </Suspense>
      ) : (
        <ThumbnailImage />
      )}
    </div>
  )
}

export const App = () => {
  return (
    <>
      <Style />
      <div>
        <div>
          <div>
            <Image />
          </div>
        </div>
      </div>
    </>
  )
}

render(<App />, document.getElementById('root'))
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/helper/css/index.ts
+++ b/deno_dist/helper/css/index.ts
@@ -304,11 +304,14 @@ export const createCssContext = ({ id }: { id: Readonly<string> }) => {
     return className
   }
 
-  const Style = ({ children }: { children?: Promise<string> } = {}) =>
+  const Style = ({ children }: { children?: Promise<string> } = {}): Promise<HtmlEscapedString> =>
     children
-      ? children.then((cssData) =>
-          raw(`<style id="${id}">${(cssData as CssClassName)[STYLE_STRING]}</style>`)
-        )
+      ? children instanceof Promise
+        ? children.then((cssData) =>
+            raw(`<style id="${id}">${(cssData as CssClassName)[STYLE_STRING]}</style>`)
+          )
+        : // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          ({ tag: 'style', children, props: { id } } as any) // TBD: <Style>{css``}</Style> is currently not working. only support <Style/>
       : raw(`<style id="${id}"></style>`)
 
   return {
--- a/deno_dist/jsx/dom/render.ts
+++ b/deno_dist/jsx/dom/render.ts
@@ -51,8 +51,10 @@ export type PendingType =
 export type Context =
   | [
       PendingType, // PendingType
-      boolean // got an error
+      boolean, // got an error
+      boolean // with startViewTransition
     ]
+  | [PendingType, boolean]
   | [PendingType]
   | []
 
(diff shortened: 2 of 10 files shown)


yusukebe:
Hi @usualoma !

This is a super cool feature!

This is not directly related to this, but I was playing with this feature and noticed that `keyframes` works on the DOM?

yusukebe:
Anyway. This PR is great and completed. I will merge this now.

PR #2027, opened 2024-01-19
feat!(deno): move middleware to helper

watany-dev:
This may also need to be noted in the Migration Guide.

- [ ] Add tests
- [ ] Run tests


yusukebe:
@watany-dev 

Thaks! I've renamed the title. I'll merge it now.

Issue #2031, opened 2024-01-19
[SSG] `FileSystemModule`'s `writeFile` data must not have `ArrayBuffer`?


yusukebe:
### What version of Hono are you using?

Latest: https://github.com/honojs/hono/tree/24723ffb91e34ce42e1b9d01f818e636f6a2fb65

### What runtime/platform is your app running on?

any

### What steps can reproduce the bug?

```ts
import fs from 'node:fs/promises'
import { Hono } from '../../src'
import { toSSG } from '../../src/helper/ssg'

const app = new Hono()

toSSG(app, fs)
```

### What is the expected behavior?

There is no type error.

### What do you see instead?

A type error.

<img width="259" alt="Screenshot 2024-01-19 at 18 58 12" src="https://github.com/honojs/hono/assets/10682/3f773cc9-b99e-4ed5-b75a-92cba3a05bce">

### Additional information

If we just want to be compatible with Node.js `fs`, we can remove the `ArrayBuffer`.

```diff
diff --git a/src/helper/ssg/index.ts b/src/helper/ssg/index.ts
index 3834b6e..ef078f7 100644
--- a/src/helper/ssg/index.ts
+++ b/src/helper/ssg/index.ts
@@ -13,7 +13,7 @@ import { joinPaths, dirname } from './utils'
  * The API might be changed.
  */
 export interface FileSystemModule {
-  writeFile(path: string, data: string | ArrayBuffer | Uint8Array): Promise<void>
+  writeFile(path: string, data: string | Uint8Array): Promise<void>
   mkdir(path: string, options: { recursive: boolean }): Promise<void | string>
 }
```

yusukebe:
@watany-dev 

Could you take a look?

watany-dev:
@yusukebe 

I certainly don't need it. Fixed.

yusukebe:
Thanks. Fixed by #2032

PR #2032, opened 2024-01-19
fix(ssg): Remove ArrayBuffer from FileSystemModule's writeFile

watany-dev:
fixes: https://github.com/honojs/hono/issues/2031
 
- [ ] Add tests
- [ ] Run tests


yusukebe:
Thanks!

PR #2044, opened 2024-01-20
fix(ssg): Skip the creation of meaningless static files.

watany-dev:
Currently, the toSSG function can generate static files, but there are routes where unexpected content, such as 404 Not Found, is retrieved. This can be identified in advance through a fetch process and prevented from being created.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/helper/ssg/index.ts
+++ b/deno_dist/helper/ssg/index.ts
@@ -121,7 +121,7 @@ export const fetchRoutesContent = async <
 
     for (const param of forGetInfoURLRequest.ssgParams) {
       const replacedUrlParam = replaceUrlParam(route.path, param)
-      const response = await app.request(replaceUrlParam(route.path, param))
+      const response = await app.request(replacedUrlParam)
       const mimeType = response.headers.get('Content-Type')?.split(';')[0] || 'text/plain'
       const content = await parseResponseContent(response)
       htmlMap.set(replacedUrlParam, {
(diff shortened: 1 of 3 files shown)


watany-dev:
As you can see from the tests, routes using app.mount are also ignored, but whether this is appropriate behavior may be subject to debate.

yusukebe:
Hi @watany-dev !

This feature needs to be discussed.

When I used to create a static site using `toSSG()` and upload it to Cloudflare Pages, there was no need to ignore `404`.

In Cloudflare Pages, the file `/404.html` will be a custom Not Found page.

https://developers.cloudflare.com/pages/configuration/serving-pages/

If you want to customize the page yourself, you need to set the path as follows and write it out with `toSSG()`.

```tsx
app.get('/404', (c) => {
  c.status(404)
  return c.render(<h1>Not found!!</h1>)
})
```

So I would not want a page with a 404 status code to be ignored.

watany-dev:
I see. You have a pattern for the intended custom error page. It looks like we should not adopt it as is.

By the way, should we include a "ignore all but app.all, app.get" process? We will see if this is technically possible.

watany-dev:
I have looked up people using Hono and sometimes I see examples of using c.render for the return value of post. Maybe we can close this PR once.

yusukebe:
@watany-dev 

How about creating hook options like this?


```ts
toSSG(app, fs, {
  beforeRequest: (req) => {
    if (req.method === 'GET') {
      return req
    }
    return false
  },
  afterResponse: (res) => {
    if (res.status === 200 || res.status === 500) {
      return res
    }
    return false
  },
})
```

watany-dev:
@yusukebe 
Thank you. That implementation seems more versatile. I will close this PR. I have opened this one as a successor.
https://github.com/honojs/hono/pull/2054

Issue #2019, opened 2024-01-18
When chaining routes with middlewares, the context type might be incorrect.


agatan:
### What version of Hono are you using?

3.12.5

### What runtime/platform is your app running on?

Node

### What steps can reproduce the bug?

Run and type check the code below:

```typescript
import { Hono, MiddlewareHandler } from 'hono'

const mw1: MiddlewareHandler<{ Variables: { foo1: string } }> = async (c, next) => { 
  c.set('foo1', 'foo1')
  await next()
}
const mw2: MiddlewareHandler<{ Variables: { foo2: string } }> = async (c, next) => {
  c.set('foo2', 'foo2')
  await next()
}

const app = new Hono()
  .get('/1', mw1, async (c) => {
    // this is correctly typed.
    return c.json({ foo1: c.var.foo1 })
  })
  .get('/1-then-2', mw1, mw2, async (c) => {
    // this is also correctly typed.
    return c.json({ foo1: c.var.foo1, foo2: c.var.foo2 })
  })
  .get('/nothing', async (c) => {
    // In this scope, c.var should be empty because no middlewares are applied.
    // However, c.var is typed as { foo1: string; foo2: string; }.
    // In runtime, v is {}.
    const v: { foo1: string, foo2: string } = c.var;
    return c.json(v);
  });


console.log(await (await app.request('/1')).json())
// => { foo1: 'foo1' }

console.log(await (await app.request('/1-then-2')).json())
// => { foo1: 'foo1', foo2: 'foo2' }

console.log(await (await app.request('/nothing')).json())
// => {}
```

### What is the expected behavior?

In the handler for `/nothing`, no middlewares are applied.
So, the type of `c.var` should be inferred as an empty object.

### What do you see instead?

In the handler for `/nothing`, the type of `c.var` is  inferred as `{ foo1: string; foo2: string }`.

### Additional information

I believe the issue with Hono lies in the overloaded types of [HandlerInterface](https://github.com/honojs/hono/blob/a5d9ca88c025b2e604593d77abff5ece38eb55d3/src/types.ts#L81).

The HandlerInterface returns Hono<E, S, P>, where E is extended by the Env definitions of handlers. However, this extension is limited to the same handler scope.

I attempted a fix ([see commit](https://github.com/agatan/hono/commit/e71c4b6fb7a1ae60b67fd857ca8a4b4f4cdc9d09)), but then realized that this might conflict with issue [#1666](https://github.com/honojs/hono/issues/1666).

[...]

PR #2046, opened 2024-01-20
fix!(types): Fix context type when chaining routes with middlewares

agatan:
Fix https://github.com/honojs/hono/issues/2019.
As

- [x] Add tests
- [x] Run tests


agatan:
Sorry, I forgot to run `yarn format:fix`.

src/types.test.ts
```diff
@@ -1003,6 +851,30 @@ describe('c.var with chaining - test only types', () => {
...
+      })
+      .get('/', (c) => {
+        expectTypeOf(c.get('init')).toEqualTypeOf<number>()
+        expectTypeOf(c.var.init).toEqualTypeOf<number>()
+        // @ts-expect-error foo1 is not typed
+        c.get('foo1')
+        // @ts-expect-error foo1 is not typed
+        c.var.foo1
+        return c.json(0)
+      })
```
yusukebe, line 877: If you can, I'd like you to write the rest of the test too.

```ts
new Hono<Env>()
.get('/', mw1, mw2, (c) => {
  expectTypeOf(c.get('init')).toEqualTypeOf<number>()
  expectTypeOf(c.var.init).toEqualTypeOf<number>()
  expectTypeOf(c.get('foo1')).toEqualTypeOf<string>()
  expectTypeOf(c.var.foo1).toEqualTypeOf<string>()
  expectTypeOf(c.get('foo2')).toEqualTypeOf<string>()
  expectTypeOf(c.var.foo2).toEqualTypeOf<string>()
  return c.json(0)
})
.get('/', (c) => {
  expectTypeOf(c.get('init')).toEqualTypeOf<number>()
  expectTypeOf(c.var.init).toEqualTypeOf<number>()
  // @ts-expect-error foo1 is not typed
  c.get('foo1')
  // @ts-expect-error foo1 is not typed
  c.var.foo1
  // @ts-expect-error foo1 is not typed
  c.get('foo2')
  // @ts-expect-error foo1 is not typed
  c.var.foo2
  return c.json(0)
})

// ...
```

src/types.test.ts
```diff
@@ -836,158 +836,6 @@ describe('c.var with chaining - test only types', () => {
       return c.json(0)
     })
 
-    // app.get('/', handler...)
-
-    new Hono().get('/', mw1).get('/', (c) => {
```
yusukebe, line 841: As you are doing, I think these tests should be removed.

[As I mentioned](https://github.com/honojs/hono/issues/2019#issuecomment-1900254498), the correct behavior is for these tests to fail, since the chained return value app should not inherit the env defined in `mw1`.

yusukebe (commented):
@agatan 

Thank you for the PR! I think this is good. I've left comments.

src/types.test.ts
agatan: Sorry for late reply. I'll try it.

agatan: @yusukebe I added tests for `app.get(path, ...handlers x 3)` ~ `app.get(path, ...handlers x 10)`!

yusukebe (approved):
LGTM!

yusukebe:
@agatan 

Awesome! Thanks for the hard work!

PR #2056, opened 2024-01-22
fix(ssg): Uniformly Convert Paths Ending with Slash to 'index.ext' Format

watany-dev:
- before:  `/api/` -> `./static/api/.html`
- after:  `/api/` -> `./static/api/index.html`

@yusukebe 
Currently, the implementation uniformly converts paths like `/css/` to `/css/index.css`. Considering this, there are two possible approaches to the implementation, and I have chosen the first one:

1. Treat all paths ending with a slash uniformly as `index.ext`, even for non-HTML files. Those who dislike this approach can use paths like `/css` without the trailing slash.
2. For non-HTML files ending with a slash, complete them as `/css.css`. I did not adopt this approach as it seems counterintuitive

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/helper/ssg/index.ts
+++ b/deno_dist/helper/ssg/index.ts
@@ -30,8 +30,13 @@ export interface ToSSGResult {
 
 const generateFilePath = (routePath: string, outDir: string, mimeType: string) => {
   const extension = determineExtension(mimeType)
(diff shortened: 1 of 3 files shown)


yusukebe (approved):
Seems to be good!

yusukebe:
@watany-dev 

Thanks! It is good. We may have a problem with these path issues, but let's handle it then.

PR #2069, opened 2024-01-23
test(jsx): add test for Context with Suspense

usualoma:

- [ ] Add tests
- [ ] Run tests


usualoma:
Close this PR as we will add the test in another PR.

Issue #2070, opened 2024-01-23
make `files` in `ToSSGResult` not optional


yusukebe:
### What is the feature you are proposing?

I think it's convenient if it's not optional. If there are no files, it can be `[]`.

```ts
export interface ToSSGResult {
  success: boolean
  files: string[]
  error?: Error
}
```

yusukebe:
Done in #2071

PR #2071, opened 2024-01-24
fix:(ssg): make 'files' mandatory in ToSSGResult

watany-dev:
fixes https://github.com/honojs/hono/issues/2070

- [x] Add tests
- [x] Run tests


yusukebe:
Thanks @watany-dev !

PR #2072, opened 2024-01-24
fix: all statusCode (for type)

EdamAme-x:

- [ ] Add tests
- [ ] Run tests


EdamAme-x:
miss

PR #2073, opened 2024-01-24
feat(types): explicitly specify statusCode

EdamAme-x:
explicitly specify statusCode
I explicitly specified all status codes defined in the specification.

```ts
app.get("/", (c) => {
   // redirect status code 
    return c.redirect("https://hono.dev", 301);
}).get("/2", (c) => {
    // general
    // ...
    return c.text("success!", 201);
})
```

```ts
app.post("/3", (c) => {
    // type error
    return c.text("bad status code", 666)
})
```

but, if statusCode is unoffical statusCode
```ts
import type { UnOfficalStatusCode } from "hono/utils/http-status";

app.post("/4", (c) => {
    return c.text'("unoffical status code", 666 as UnOfficalStatusCode)
})
```

- [ ] Add tests
- [x] Run tests


EdamAme-x:
What do you think?
I believe the UX will go up

EdamAme-x:
I'm about to start my afternoon classes.
I won't be able to check in for a little while...

yusukebe:
Hi @EdamAme-x 

Seems to be good!

As for the Unofficial status code, passing `ResponseInit` as the second argument will not cause a type error.

```ts
app.post('/666', (c) => {
  return c.text('unofficial status code', {
    status: 666,
  })
})
```

Either way, I would like to merge. Is that okay?

EdamAme-x:
Of course. Thanks!

yusukebe:
Let's go.

PR #2054, opened 2024-01-22
feat(ssg): support Improve Hook Handling

watany-dev:
Following @yusukebe  advice, I have created the beforeRequestHook and AfterRequestHook. Users can now prevent unexpected outputs with these hooks.

### Author should do the followings, if applicable

- [x] Add tests
- [x] Run tests
- [x] `yarn denoify` to generate files for Deno

PR #2074, opened 2024-01-24
fix: await generate hook

watany-dev:
ref https://github.com/honojs/hono/pull/2054

The test appears to be failing accidentally. This is probably due to something not awaiting generatehook

- [ ] Add tests
- [ ] Run tests


yusukebe:
@watany-dev 

Thanks. Merging.

PR #2079, opened 2024-01-24
test: compress middleware custom error message

ariskemper:

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/compress/index.test.ts
+++ b/src/middleware/compress/index.test.ts
@@ -62,14 +62,22 @@ describe('Parse Compress Middleware', () => {
     expect(res.headers.get('Content-Length')).toBe('5')
   })
 
-  it('custom error message 400 Not Found', async () => {
+  it('error 400 Not Found with headers accept-encoding', async () => {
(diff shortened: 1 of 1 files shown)


ariskemper:
@usualoma @yusukebe extra tests for not found error with compress middleware

usualoma:
@ariskemper 

Thanks for creating the PR.

I like the idea of more testing, but on the other hand I think it is also true that more test code is also more maintenance cost. I believe we need a test code for our needs.

I think the file src/middleware/compress/index.test.ts should contain tests related to the middleware of compress. The “matched route that includes only one middleware breaks applying order “ (#2080) is a bug of hono's own, not of the middleware. 
Also, the middleware of compress does not contain any code related to 404.

Given that, I see no need for this PR test code in src/middleware/compress/index.test.ts.

Do you think that "in the future, regressions that cannot be detected by the test code of hono itself can be detected by this test code?"

ariskemper:
@usualoma i agree with you, so lets keep it as you mentioned, but please check https://github.com/honojs/node-server/pull/130

PR #2080, opened 2024-01-25
fix(hono-base): fix not found handler calling timing when one middleware

### LINKED ISSUE #2061: NotFound Handler + Compress does not return body

3.12.6

Node

```typescript
import { Hono } from 'hono'
import { compress } from 'hono/compress'

const app = new Hono()
app.use('*', compress())
app.notFound((c) => c.json({ message: 'error message' }, 400))

serve({
  fetch: app.fetch,
  port: 3000,
})
```

Returns error body:
```json
{ "message": "error message" }
```

### What do you see instead?

No body content - browser default error page

### Additional information

Might be an issue in hono + compress or maybe in `@hono/node-server` package, as there were similar issues in regular handlers in older versions

usualoma:
Fixes #2061

Fix problem where notFound was not handled correctly when the result returned from the router was a single middleware.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/hono-base.ts
+++ b/deno_dist/hono-base.ts
@@ -308,28 +308,21 @@ class Hono<
     if (matchResult[0].length === 1) {
       let res: ReturnType<H>
       try {
-        res = matchResult[0][0][0][0](c, async () => {})
-        if (!res) {
-          return this.notFoundHandler(c)
-        }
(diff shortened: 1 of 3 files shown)


yusukebe:
That was a bug! Thank you for the good fix.

Merging now.

PR #2085, opened 2024-01-25
fix: enable to return Response like object 

usualoma:
This PR fixes regression in the https://github.com/honojs/hono/pull/2080.

At first glance, it appears to be a significant change from the following code, but the conditions are the same.

```ts
      if (res instanceof Response) return res

      return (async () => {
        let awaited: Response | void
        try {
          awaited = await res
          if (!awaited) {
            return this.notFoundHandler(c)
          }
        } catch (err) {
          return this.handleError(err, c)
        }
        return awaited
      })()
```

If `res instanceof Response` is `false`, whether it is a `Promise<Response>` or a Response-like object, it is resolved in the subsequent `await res`. Therefore, it is sufficient to check the type of res with `res instanceof Promise` and switch its behavior.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/hono.test.ts
+++ b/src/hono.test.ts
@@ -20,73 +20,222 @@ function throwExpression(errorMessage: string): never {
 }
 
 describe('GET Request', () => {
-  const app = new Hono()
+  describe('without middleware', () => {
+    // In other words, this is a test for cases that do not use `compose()`
 
-  app.get('/hello', async () => {
-    return new Response('hello', {
-      status: 200,
-      statusText: 'Hono is OK',
+    const app = new Hono()
+
+    app.get('/hello', async () => {
+      return new Response('hello', {
+        status: 200,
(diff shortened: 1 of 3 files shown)


src/hono.test.ts
```diff
@@ -20,73 +20,222 @@ function throwExpression(errorMessage: string): never {
...
 
-  it('GET http://localhost/hello is ok', async () => {
-    const res = await app.request('http://localhost/hello')
-    expect(res).not.toBeNull()
-    expect(res.status).toBe(200)
-    expect(res.statusText).toBe('Hono is OK')
-    expect(await res.text()).toBe('hello')
-  })
+    app.get(
+      '/proxy-object',
```
yusukebe, line 46: This - adding the tests to handle the `Proxy` object is good!

yusukebe (approved):
LGTM!

PR #2094, opened 2024-01-26
feat(serve-static): `mimes` option for serve-static

### LINKED ISSUE #2067: `mimes` option for serve-static

This feature supports rarely used content-types and great amount of types.
Also, I think to remove unnecessary default MIME to make it lighter in the future.
```ts
// HLS
const mimes = {
  m3u8: 'application/x-mpegURL',
  ts: 'video/mp2t'
}
app.use('/video/*', serveStatic({mimes}))
```

ryuapp:
close #2067 

This option enables you to set your own mime to serveStatic().
This `mimes` overwrites default mime in Hono.
```ts
// e.g. HLS
const mimes = {
  m3u8: 'application/x-mpegURL',
  ts: 'video/mp2t'
}
app.use('/video/*', serveStatic({mimes}))
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/utils/mime.ts
+++ b/deno_dist/utils/mime.ts
@@ -1,8 +1,8 @@
-export const getMimeType = (filename: string, mime = baseMime): string | undefined => {
+export const getMimeType = (filename: string, mimes = baseMimes): string | undefined => {
   const regexp = /\.([a-zA-Z0-9]+?)$/
   const match = filename.match(regexp)
   if (!match) return
-  let mimeType = mime[match[1]]
+  let mimeType = mimes[match[1]]
   if ((mimeType && mimeType.startsWith('text')) || mimeType === 'application/json') {
     mimeType += '; charset=utf-8'
   }
@@ -10,10 +10,10 @@ export const getMimeType = (filename: string, mime = baseMime): string | undefin
 }
 
 export const getExtension = (mimeType: string): string | undefined => {
-  return Object.keys(baseMime).find((ext) => baseMime[ext] === mimeType)
+  return Object.keys(baseMimes).find((ext) => baseMimes[ext] === mimeType)
 }
 
-const baseMime: Record<string, string> = {
+const baseMimes: Record<string, string> = {
   aac: 'audio/aac',
   abw: 'application/x-abiword',
   arc: 'application/x-freearc',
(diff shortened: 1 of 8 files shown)


src/adapter/cloudflare-workers/serve-static.ts
```diff
@@ -46,7 +47,7 @@ export const serveStatic = <E extends Env = Env>(
     })
 
     if (content) {
-      const mimeType = getMimeType(path)
+      const mimeType = getMimeType(path, options.mime)
```
yusukebe, line 50: I would suggest using the default mime types if there were no mime types specified.

```ts
let mimeType: string | undefined = undefined
if (options.mime) {
  mimeType = getMimeType(path, options.mime) ?? getMimeType(path)
} else {
  mimeType = getMimeType(path)
}
```

deno_dist/utils/mime.ts
```diff
@@ -1,19 +1,19 @@
-export const getMimeType = (filename: string): string | undefined => {
+export const getMimeType = (filename: string, mime = baseMime): string | undefined => {
```
yusukebe, line 1: I think `mimes` is better since it handles multiple Mime Types.

```ts
mimes = baseMimes
```

deno_dist/utils/mime.ts
```diff
@@ -1,19 +1,19 @@
...
   return mimeType
 }
 
 export const getExtension = (mimeType: string): string | undefined => {
-  return Object.keys(mimes).find((ext) => mimes[ext] === mimeType)
+  return Object.keys(baseMime).find((ext) => baseMime[ext] === mimeType)
 }
 
-const mimes: Record<string, string> = {
+const baseMime: Record<string, string> = {
```
yusukebe, line 16: I would prefer `baseMimes` for the same reason as above.

src/adapter/cloudflare-workers/serve-static.ts
```diff
@@ -9,6 +9,7 @@ import { getContentFromKVAsset } from './utils'
 export type ServeStaticOptions<E extends Env = Env> = {
   root?: string
   path?: string
+  mime?: Record<string, string>
```
yusukebe, line 12: I think `mimes` is better, for the same reason as above.

yusukebe:
Hi @ryuapp !

Thanks. I've commented. Please check them.

ryuapp:
@yusukebe now ready

yusukebe (approved):
Looks great to me!

yusukebe:
@ryuapp Thanks! I'll merge this now.

PR #1405, opened 2023-09-04
chore(pnpm): move from `yarn` to `pnpm`

ThatOneBro:
### Author should do the followings, if applicable

- [x] Run tests
- [x] Run `denoify`

## What this PR does
* Replaces `yarn.lock` with `pnpm-lock.yaml`
* Replaces references to `yarn` with `pnpm` in the `scripts` section in `package.json`.
* Replaces `yarn` with `pnpm` and adds installation step to CI for `pnpm`
* Fixes missing `glob` devDependency, probably was just in `yarn.lock` so it was always installed 😅 

Closes #1392

PR #2105, opened 2024-01-28
chore: use Bun as a package manager

yusukebe:
This PR proposes to use Bun as a package manager instead of Yarn.

## Motivation

We have been using Yarn v1. However, we are having problems with the slow installation of dependencies. Also, the development of Yarn v1 has stopped.

I am not that much of a package manager, but I have an admiration for other modern tools such as Yarn v4 (Berry) and pnpm.

However, we cannot use [np](https://github.com/sindresorhus/np) in yarn v4 or pnpm. We are currently using np for this honojs/hono release, which must be able to use it.

## Bun

Hono supports Bun and loves Bun. Bun is also an excellent package manager. So this time we will use Bun as a package manager instead of Yarn v4 or pnpm.

Note: Bun can be a bundler and a test environment, but in this case, we will use it as a package manager and task runner.

## Bun vs Yarn v4/pnpm

Compare Bun with package managers such as Yarn v4 and pnpm.

Bun:

* Pros - Ultra-fast.
* Pros - Bun can emit `yarn.lock` compatible with Yarn v1. - So, this project will include `bun.lockb` and `yarn.lock`.
* Pros - We may use np.
* Pros - Does not need `yarn`.
* Cons - Need to install Bun.
* Cons - Does not support Windows. - but will support it soon.

Yarn v4/pnpm:

* Pros - Does not need to install Bun.
* Pros - Many projects use it.
* Cons - np does not support it.  Though supporting yarn v4, it can not upload to `registry.npmjs.org`.
* Cons - Fast, but slower than Bun.

## How fast?

https://github.com/honojs/hono/assets/10682/0a53a4a7-455c-440d-a6a4-ff375e85f884

## On which runtime will the tests be run?

Node.js.

We're using Vitest, which means that tests run on Node.js. The result of `getRuntimeKey()` is `node`.

## Changes

There are no major changes, but contributors will type the following commands.

* `yarn test` => `bun run test`
* `yarn denoify` => `bun run denoify` or `bun denoify`
* `yarn build` => `bun run build`

However, there is not only `bun.lockb` but also `yarn.lock`, so you can use the yarn commands.

## Are you sure about this?

Yes. If there is a major problem, we can go back to yarn. Plus, Hono likes Bun.

---

Related #2051 #1405


yusukebe:
Hi @ThatOneBro !

First, I want you to see this PR. I think this is a great challenge. What do you think?

ThatOneBro:
Hey @yusukebe !

I think Bun at this point is a good choice for the Hono project in terms of package manager and task runner since these both have stabilized a lot over the last half of a year. 

I would probably choose Bun over pnpm at this point, especially since Hono is currently using Yarn lockfile which is supported by Bun but not pnpm (aside from one-way import).

LGTM!

rafaell-lycan:
I'd say that `pnpm` over `yarn` is already a good enhancement (which is what my company did last year).

I have nothing against Bun I considered using it as runtime at work over Node, but only when it gets more mature. Hopefully, I'll switch by Summer 😎

yusukebe:
@rafaell-lycan 

> @yusukebe have you considered dead simple `npm` instead?

Yeah. But `npm` has no special advantages, and in that case, I'd rather keep using yarn v1. And, Bun is more exciting.

> I have nothing against Bun, and I even considered using it as runtime at work over Node, but only when it gets more mature/stable. Hopefully, I'll switch by Summer 😎

Nice! I bet it will be even better by then!

yusukebe:
Merge time! Let's go.

PR #2107, opened 2024-01-29
refactor: Improve readability of dispatch() in compose

the-pesar:
I was reading the base code when I came across this file. I tried to remove the extra curly braces and improve readability with a few changes.

- [x] Run tests


yusukebe:
Hi @the-pesar 

The linter rule is added. https://github.com/honojs/hono/pull/2112

This will not remove the curly brace. I reject this PR. Thanks.

PR #2109, opened 2024-01-29
feat(body-limit): Replace c.req.raw with body limit middleware proxy

usualoma:

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/middleware/body-limit/index.ts
+++ b/deno_dist/middleware/body-limit/index.ts
@@ -1,5 +1,8 @@
 import type { Context } from '../../context.ts'
+import { HTTPException } from '../../http-exception.ts'
 import type { MiddlewareHandler } from '../../types.ts'
+
+const ERROR_MESSAGE = 'Payload Too Large'
 
 type OnError = (c: Context) => Response | Promise<Response>
 type BodyLimitOptions = {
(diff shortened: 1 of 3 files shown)


rafaell-lycan:
That's a very clever move.

Btw, Express [supports up to 100kb](https://expressjs.com/en/resources/middleware/body-parser.html#limit) by default, and it's quite easy to configure on its middleware.

usualoma:
@yusukebe ready for review!

deno_dist/middleware/body-limit/index.ts
```diff
@@ -25,42 +33,57 @@ type BodyLimitOptions = {
  * )
  * ```
  */
-export const bodyLimit = (options: BodyLimitOptions): MiddlewareHandler =>
-  async function bodyLimit(c, next) {
-    if (!c.req.raw.body) return next()
+export const bodyLimit = (options: BodyLimitOptions): MiddlewareHandler => {
+  const onError: OnError = options.onError || ((c) => c.text('413 Request Entity Too Large', 413))
```
yusukebe, line 37: It is better to make `'413 Request Entity Too Large'` as a constant to reuse in line 66.

yusukebe:
@usualoma Thanks!

I've left a comment.

deno_dist/middleware/body-limit/index.ts
usualoma: @yusukebe Thank you!

refactored in 2a326ca

The message has been changed to "Payload Too Large" in accordance with the RFC.
https://datatracker.ietf.org/doc/html/rfc7231#section-6.5.11

yusukebe: Great 👍

yusukebe (approved):
LGTM! Merge now.

PR #2115, opened 2024-01-29
fix: replace fs.existsSync() with Bun.file().exists()

the-pesar:
- [x] Run tests


yusukebe (approved):
Looks good to me!

yusukebe:
@the-pesar Thanks. Merge now.

PR #2119, opened 2024-01-30
feat!(mime): reduce default mimes

ryuapp:
As the number of MIME increases in the future, it will be impossible to support them all due to file size problems.
We think that it should primarily support files used by browsers.
Instead, you can maintain the previous state by using mimes option #2094 .

## Delete extension List

**azw(application/vnd.amazon.ebook)**
If Amazon is using Hono, it should not be deleted.

**abw, csh, doc, docs, xls, xlsx, odp, ods, odt, ppt, pptx, vsd**
Files used by specific software.

**swf, xul**
Files for software whose support has ended.

**mpkg, sh**
Installer and scripts.

**jar, php**
Other language files.

**arc, bz, bz2, tar, 7z**
Archive and compressed files.

- [x] Add tests
- [x] Run tests


yusukebe:
> Instead, you can maintain the previous state by using mimes option #2094 .

Exactly, now we have a `mimes` option! Thanks. Merge now.

yusukebe (approved):
LGTM

PR #2121, opened 2024-01-30
feat(types): better `c.var` type

Kyiro:
I noticed `ContextVariableMap` is pretty much useless for `c.var` if `Env` in `Context` is `any` (the default), so I decided to make it respect it without breaking much code.

## Before

```ts
declare module 'hono' {
  export interface ContextVariableMap {
    x: string;
  }
}

const c = new Context(...);

c.var.x; // any
c.var.y; // any
```

## Now

```ts
declare module 'hono' {
  export interface ContextVariableMap {
    x: string;
  }
}

const c = new Context(...);

c.var.x; // string
c.var.y; // still any, as to not break any code
```

### DIFF
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -4,7 +4,7 @@ import type { HonoRequest } from './request.ts'
 import type { Env, FetchEventLike, NotFoundHandler, Input, TypedResponse } from './types.ts'
 import { resolveCallback, HtmlEscapedCallbackPhase } from './utils/html.ts'
 import type { RedirectStatusCode, StatusCode } from './utils/http-status.ts'
-import type { JSONValue, InterfaceToType, JSONParsed } from './utils/types.ts'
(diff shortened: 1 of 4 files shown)


yusukebe:
Hi @Kyiro 

I think this PR looks good, but it can infer type correctly in my environment.

Screenshot 2024-01-31 at 5 02 38

Or, my understanding is wrong?

Kyiro:
It's only an issue with middleware handlers and such
```ts
const middleware: MiddlewareHandler = (c, next) => {
    c.var.x; // it's different here because Context is any by default
};
```

yusukebe:
@Kyiro 

Ah, understood well. Please re-run `denoify`.

yusukebe:
@Kyiro 

Thanks! Merge now.

yusukebe:
This is a `feat` but it's a minor change. So, I'll include it in the next patch-release.

PR #2124, opened 2024-01-31
fix(jsx): fix context provider with async component

### LINKED ISSUE #2118: `useRequestContext` returns incorrect URL

3.12.3

CF Workers

For a given component that is part of a larger layout:
```tsx
export const NavBar: FC = async () => {
  const ctx = useRequestContext();
  console.log(ctx.req.url);

  return (
    <div
      id="topNav"
      class="navbar sticky top-0 z-50 border-b border-base-300 bg-base-100 shadow-sm"
    >CONTENT</div>
  )
};
```

The URL that is returned is always the same as the first request that was made. For example, if your first request is to `http://localhost:5173/test`, all subsequent requests will have `ctx.req.url` provided from `useRequestContext` as that same value, regardless of if you change routes.

It should be noted that this behavior was observed during local development, and I have not tested it against a deployed CF Pages.

I would expect `ctx.req.url` to always reflect the current URL that is being displayed.

### What do you see instead?

The value is always the URL of the first page that was accessed.

### Additional information

_No response_

usualoma:
### What will be fixed?

In JSX `toString()`, `useContext()` did not work correctly for nested async component. This PR fixes that problem.
It is possible that this PR will fix #2118, but it may be another cause, which will need to be confirmed after merged.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/jsx/index.ts
+++ b/deno_dist/jsx/index.ts
@@ -88,12 +88,14 @@ const childrenToStringToBuffer = (children: Child[], buffer: StringBuffer): void
   }
 }
 
+type LocalContexts = [Context<unknown>, unknown][]
 export type Child = string | Promise<string> | number | JSXNode | Child[]
 export class JSXNode implements HtmlEscaped {
   tag: string | Function
   props: Props
   children: Child[]
   isEscaped: true = true as const
+  localContexts?: LocalContexts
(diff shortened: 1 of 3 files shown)


yusukebe:
Thanks! @usualoma 

Merge now.

Issue #2131, opened 2024-02-02
Improved compatibility with react


usualoma:
### Specific Purpose

Write the following in vite.config.ts to make the existing module for 'react' work

```ts
import { defineConfig } from 'vite'
import path from 'node:path'

export default defineConfig({
  resolve: {
    alias: {
      'react': path.join(__dirname, 'node_modules', 'hono', 'dist', 'jsx', 'dom'),
    },
  },
})
```

### Example

The following projects should be made to work

https://github.com/usualoma/hono-jsx-dnd

usualoma:
#2132 works as follows.
This is powered by honor/jsx/dom and react-dnd, with no react used at all.

https://github.com/honojs/hono/assets/30598/2ad873ac-29ec-4c19-87af-d70018eed07d

yusukebe:
@usualoma 

Cooool! 

Maybe the React utilities libraries other than React DnD will work without React.

flipvh:
This is amazing. You have another sponsor @yusukebe ;D. Could this eventually help me SSR parts of my SPA, such as the about page and the contact page, with the same React components I already have in my frontend?

usualoma:
Hi @flipvh 
Thank you for your sponsorship! I am delighted.

Thank you for sponsoring me! I am glad.

The "jsx/dom" of hono is basically for small apps now, and I would like to follow the KISS principle: it is not impossible to create a SPA, but I think it is more suitable to serve HTML with SSR and then hydrate it as a partial client component.

React is a really, really great product, and I respect it, but it can be too big to create small apps. hono doesn't need to worry about compatibility with older apps, so it keeps the size small by not supporting class components.

Also, React is implemented to maintain high responsiveness in "large SPA applications," but hono does not consider this. However, it is not "unsupported by SPA", so if it works well with SPA, we would like to consider using it that way.

usualoma:
### Update

In the first example, the dependency was specified by relative path, but we found that it did not work, so we decided to specify it by tar archive. I put the contents of #2132 in a tar archive and confirmed that it works.

### Bundle size

In my example, bundling hono/jsx/dom and react-dnd came to 56.65 kB. react, react-dom and react-dnd were 192.72 kB.

#### hono/jsx/dom and react-dnd

<img width="518" alt="image" src="https://github.com/honojs/hono/assets/30598/fb685ffe-ee14-4a60-ad8b-7e29ffbe517b">

#### react, react-dom and react-dnd

[...]

PR #2132, opened 2024-02-02
feat(jsx/dom): more react staff

usualoma:
Merging this PR will accomplish the purpose of #2131.

### New staff members

`useMemo`, `useLeyoutEffect`, `isValidElement`, `cloneElement`

#### `useContext`

Fixed the same problem with dom version as #2124

#### `useEffect`

As in the original React, useEffect is now executed asynchronously after the DOM is rendered, rather than synchronously.

### Enable to use 'hono/jsx/dom' for replacement of 'react'. 

You can use dom-specific version of hooks and utils by importing from 'hono/jsx/dom' by 09da043

- [x] Add tests
- [x] Run tests


usualoma:

yusukebe (approved):
I've tried it, and it works well. LGTM!

PR #2139, opened 2024-02-02
refactor(jsx): Remove unused HONO_COMPONENT feature

### LINKED ISSUE #2133: Drop HONO_COMPONENT feature

HONO_COMPONENT was an idea to make hydration easier, but is no longer used. It would make more sense to write this process on the build system side. Therefore, I think this code should be removed before v4 is released.

https://github.com/usualoma/hono/blob/61202eda1453da8265c076ae80665ccc3e2e6d21/src/jsx/index.ts#L29-L30

usualoma:
Fixes #2133

- [x] Add tests
- [x] Run tests


yusukebe (approved):
LGTM!

PR #2143, opened 2024-02-03
fix(html): Remove circular dependencies in `hono/html`

javascripter:
# Why
There is a circular dependency between `src/helper/html/index.ts` and `src/utils/html.ts`. This PR removes this circular dependency by removing imports to helper files in utility files.

I found this circular dependency while running Hono in [Expo API Routes](https://blog.expo.dev/rfc-api-routes-cce5a3b9f25d). Displaying warnings for node_modules dependencies is [a bug](https://github.com/expo/expo/issues/26613) in Expo but I believe it's good to remove circular dependencies in the first place. 

Screenshot 2024-02-03 at 18 31 40

# How
Move the `raw` function to `src/utils/html.ts` and re-export `raw` in `src/helper/html/index.ts`

# Test Plan
- Run the following command to find circular dependencies warnings disappear after the change.
```
npx dpdm --transform --no-tree ./src/helper/html/index.ts
```
Screenshot 2024-02-03 at 18 14 13

- Tests for the `raw` remain in the `src/helper/html/index.ts` file so that we continue to test all public API surfaces there. 

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/helper/html/index.ts
+++ b/deno_dist/helper/html/index.ts
@@ -1,18 +1,11 @@
-import { escapeToBuffer, stringBufferToString } from '../../utils/html.ts'
+import { escapeToBuffer, stringBufferToString, raw } from '../../utils/html.ts'
 import type {
   StringBuffer,
   HtmlEscaped,
   HtmlEscapedString,
-  HtmlEscapedCallback,
 } from '../../utils/html.ts'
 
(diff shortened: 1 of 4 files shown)


yusukebe:
Hi @javascripter 

Thanks for the PR. This is good!

One thing, please run `yarn lint:fix && yarn format:fix` to format the codes.

javascripter:
@yusukebe Thanks for taking a look! I've reformatted the code and rerun `yarn denoify`.

PR #2144, opened 2024-02-03
refactor: run `prettier` and `linter`

yusukebe:

- [ ] Add tests
- [ ] Run tests


yusukebe:
We don't need this.

PR #2146, opened 2024-02-03
feat(jsx): "className" is now an alias for "class"

### LINKED ISSUE #2141: `className` as `class`

We have been using `class` instead of `className`, but still want to use `className` as well for React compatibility.

So why not make `className` an alias to `class`? Specifically, `className="light"` is converted to `class="light"`.

usualoma:
Fixes #2141

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/jsx/utils.ts
+++ b/src/jsx/utils.ts
@@ -1,0 +1,6 @@
+export const normalizeIntrinsicElementProps = (props: Record<string, unknown>): void => {
+  if (props && 'className' in props) {
+    props['class'] = props['className']
+    delete props['className']
+  }
+}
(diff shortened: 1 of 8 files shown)


usualoma:
@yusukebe  Would you please review?

src/jsx/utils.ts
```diff
@@ -0,0 +1,6 @@
+export const normalizeIntrinsicElementProps = (props: Record<string, unknown>): void => {
```
yusukebe, line 1: Creating the function is good!

yusukebe (commented):
LGTM!

Issue #2142, opened 2024-02-02
Remove "children" from FC


usualoma:
### What is the feature you are proposing?

(Due to my lack of insight) FC no longer includes children since React 18, so I would like to exclude children from hono accordingly.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L853

```diff
diff --git a/src/jsx/index.ts b/src/jsx/index.ts
index 60b1eb1a..a1c052e7 100644
--- a/src/jsx/index.ts
+++ b/src/jsx/index.ts
@@ -273,9 +273,7 @@ export const jsxFn = (
   }
 }
 
-export type FC<T = Props> = (
-  props: T & { children?: Child }
-) => HtmlEscapedString | Promise<HtmlEscapedString>
+export type FC<T = Props> = (props: T) => HtmlEscapedString | Promise<HtmlEscapedString>
 
 const shallowEqual = (a: Props, b: Props): boolean => {
   if (a === b) {

```

yusukebe:
Hi @usualoma 

It's good and how about providing `PropsWithChildren` [like the `@types/react`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/917e731d78e21a8c4886f48c0fead0e64553d484/types/react/index.d.ts#L1433)?

```tsx
type PropsWithChildren<P = unknown> = P & { children?: Child | undefined }

type Post = {
  id: number
  title: string
}

function Parent({ title, children }: PropsWithChildren<Post>) {
  return (
    <div>
      <h1>{title}</h1>
      {children}
    </div>
  )
}
```

yusukebe:
This was fixed by #2151. Thanks!

PR #2151, opened 2024-02-04
refactor(jsx): Tidyup the types to be exported

usualoma:
Tidyup and organize the type definitions exported from hono/jsx in src/jsx/types.ts.

I don't think this will affect most users, but it could be a breaking change for a small percentage of users, so I think this should be included in a major version upgrade (v4).

### New exported type

* PropsWithChildren #2142

### Changed types

* Remove "children" from FC #2142
* JSXNode
    * The previous "class as a value" import is no longer available, and only "type" import is now available.
        * `import { JSXNode } from 'hono/jsx'` is no longer available, only `import type { JSXNode } from 'hono/jsx'` is available.

### No longer be exported

Until now, the types used internally were over-exported. With this PR change, types needed only internally should not be exported.

* Props
* JSXFragmentNode
* jsxFn

- [x] Add tests
- [x] Run tests


usualoma:
I am trying to figure out how to solve the following problem
https://github.com/honojs/hono/actions/runs/7776848588/job/21204467952?pr=2151

yusukebe:
@usualoma 

Maybe this issue: https://github.com/garronej/denoify/issues/124

usualoma:
@yusukebe Oh, thank you. I understand.
I think `export *` would work here, so I changed it temporarily with 874ca99.
Please review.

yusukebe (approved):
LGTM!

yusukebe:
@usualoma 

I think it is perfect except for the `denofy` matter. We can update that once `denoify` is fixed.

Thank you for your hard work.

PR #2161, opened 2024-02-07
Fix/function for attribute

usualoma:
Like React's `renderToString()`, the value should be ignored if the key starts with "on" and an error should occur if the key does not.

https://github.com/honojs/hono/compare/v4...usualoma:hono:fix/function-for-attribute?expand=1#diff-bbb085cf2da64bb28cf34775197b5fb966c1f425b9d12087a4f6e892a7324557R296-R308

Depending on the configuration of the application, users may want to pass functions as strings. However, it is not a good behavior for hono to automatically convert functions to strings, since the stringified function may unintentionally contain server-side code. If an application needs to stringify functions, it should explicitly stringify them using `toString()`.

```tsx
<button data-handler={((ev) => { clientSideCode(ev) }).toString()}>Click</button>
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/jsx/base.ts
+++ b/deno_dist/jsx/base.ts
@@ -163,6 +163,11 @@ export class JSXNode implements HtmlEscaped {
       } else if (v instanceof Promise) {
         buffer[0] += ` ${key}="`
         buffer.unshift('"', v)
+      } else if (typeof v === 'function') {
+        if (!key.startsWith('on')) {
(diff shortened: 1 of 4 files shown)


yusukebe:
@usualoma 

Thanks! Looks good!

Serializing something like a function (maybe not a real function) and passing it from the server side to the client is an interesting challenge, but I don't think we need to do it now.

PR #2162, opened 2024-02-07
fix(jsx): The third argument of jsx(), key, is optional

usualoma:
As shown in the following example, if key is unspecified, the third argument is not passed, so it is correct that it is optional.

image
https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html


yusukebe:
Looks good 👍

PR #2165, opened 2024-02-07
Feat/more jsx event

### LINKED ISSUE #2140: Types for Events

We want to have React-like TypeScript types for HTML events like a `React.ChangeEvent`.

For example, the simple one could be implemented as the following:

```tsx
interface ChangeEvent<T = Element> {
  target: EventTarget & T
}

const Input = (
  <input
    type="checkbox"
    onChange={(e: ChangeEvent<HTMLInputElement>) => setFoo(e.target.checked)}
  />
)
```

There are some event types other than `React.ChangeEvent`:

```ts
type Props = {
  onClick: (event: React.MouseEvent<HTMLInputElement>) => void
  onChange: (event: React.ChangeEvent<HTMLInputElement>) => void
  onKeypress: (event: React.KeyboardEvent<HTMLInputElement>) => void
  onBlur: (event: React.FocusEvent<HTMLInputElement>) => void
  onFocus: (event: React.FocusEvent<HTMLInputElement>) => void
  onSubmit: (event: React.FormEvent<HTMLFormElement>) => void
  onClickDiv: (event: React.MouseEvent<HTMLDivElement, MouseEvent>) => void
}
```

usualoma:
fixes #2140

### Event types

In hono, handlers are registered directly with addEventListener, so we defined the raw Event as it is.
The list of events is referenced below.
https://developer.mozilla.org/en-US/docs/Web/API/Element
https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement#events
https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement#events
https://developer.mozilla.org/en-US/docs/Web/API/Element/input_event

Some events may be missing, in which case you can add your own events.

```ts
declare global {
  namespace Hono {
    interface EventAttributes {
      onMyCustomEvent?: (event: Event) => void
    }
  }
}
```

### Support event capture phase

As with React, if the event name ends with `Capture`, it is considered a capture phase handler.

### Names are now strictly handled

React requires the on to be followed by an uppercase letter, as in `onClick`. This PR will add a type, and I don't think any user will write it like `onclick`, so I think it would be less confusing to adapt hono to React as well.

- [x] Add tests
- [x] Run tests


usualoma:
I found a bug, but it has been fixed in b536fa2.
The work is complete.

yusukebe:
@usualoma 

Super cool! Now we have all the features to v4. Thanks!

PR #2169, opened 2024-02-08
PoC: feat(ssg): generate files concurrently

usualoma:
This PR refers to v4, but I believe the merge to v4 can be done at a later date.

### Overview

#2013 is great. However, in SSG, as well as requests to external APIs, writing to the file system is also a heavy process, so I think it would be better to parallelize everything, including writing to the file system. Also in v4, the SSG feature reads all pages into memory before writing them out, but it would be preferable to be able to parallelize the generation and writing so that not all pages are read.

In this PR, the following number of requests are processed in parallel at the same time.
https://github.com/honojs/hono/compare/v4...usualoma:hono:feat/ssg-generator?expand=1#diff-5bc74b06798a0aee17368e4fb5a84c845e66631cab1e30a35342a8b5a4f7d545R1

### Usage

```ts
toSSG(app, fs) // Default number of parallelism
```

```ts
toSSG(app, fs, {
  concurrency: 2,
}) // Change the number of parallels to 2
```

### Throttling of external API calls

When it comes to external API calls, I don't think limiting the number of concurrent executions is inherently the job of the SSG. I think we should use something like `createPool()` added in this PR to limit it when calling the fetch API. 

```ts
const fetchPoolMap: Record<string, Pool> = {}
// maximum 2 fetches per host at the same time
const throttledFetch = async (url: string, init?: RequestInit) => {
  const host = new URL(url).host
  const pool = fetchPoolMap[host] ||= createPool({ concurrency: 2 })
  return pool.run(() => fetch(url, init))
}
```

Nevertheless, well, it may be useful to be able to limit the SSG when it is a hassle to set up the `fetch` yourself.

### Benchmark - slow external API

<details>
<summary>code</summary>

```ts
/** @jsxRuntime automatic @jsxImportSource ./src/jsx */
import fs from 'fs/promises'
import { Hono } from './src'
import { toSSG, ssgParams } from './src/helper/ssg'
import { jsxRenderer } from './src/middleware/jsx-renderer'

const app = new Hono()

app.get(
  '*',
  jsxRenderer(
    ({ children }) => {
      return (
        <html>
          <body>{children}</body>
        </html>
      )
    },
    {
      docType: true,
    }
  )
)

const getShops = async () => {
  await new Promise((resolve) => setTimeout(resolve, 1000))
  return [
    { id: '1', name: 'Shop 1' },
    { id: '2', name: 'Shop 2' },
    { id: '3', name: 'Shop 3' },
    { id: '4', name: 'Shop 4' },
    { id: '5', name: 'Shop 5' },
    { id: '6', name: 'Shop 6' },
    { id: '7', name: 'Shop 7' },
    { id: '8', name: 'Shop 8' },
    { id: '9', name: 'Shop 9' },
    { id: '10', name: 'Shop 10' },
  ]
}

const getShop = async (id: string) => {
  await new Promise((resolve) => setTimeout(resolve, 1000))
  return (await getShops()).find((shop) => shop.id === id)
}

app.get(
  '/shops/:id',
  ssgParams(async () => {
    const shops = await getShops()
    return shops.map((shop) => ({ id: shop.id }))
  }),
  async (c) => {
    const shop = await getShop(c.req.param('id'))
    if (!shop) {
      return c.notFound()
    }
    return c.render(
      <div>
        <h1>{shop.name}</h1>
      </div>
    )
  }
)

toSSG(app, fs)
```
</details>

```
$ npx esbuild --platform=node --bundle ssg-app.tsx | time node # with this PR
node  0.06s user 0.04s system 2% cpu 3.765 total
$ npx esbuild --platform=node --bundle ssg-app.tsx | time node # v4
node  0.12s user 0.05s system 0% cpu 33.748 total
```

### Benchmark - large site

<details>
<summary>code</summary>

```ts
/** @jsxRuntime automatic @jsxImportSource ./src/jsx */
import fs from 'fs/promises'
import { Hono } from './src'
import { toSSG, ssgParams } from './src/helper/ssg'
import { jsxRenderer } from './src/middleware/jsx-renderer'

const app = new Hono()

app.get(
  '*',
  jsxRenderer(
    ({ children }) => {
      return (
        <html>
          <body>{children}</body>
        </html>
      )
    },
    {
      docType: true,
    }
  )
)

const shops = new Array(10000).fill(0).map((_, i) => ({ id: String(i + 1), name: `Shop ${i + 1}` }))
const getShops = async () => {
  return shops
}

const getShop = async (id: string) => {
  return (await getShops()).find((shop) => shop.id === id)
}

app.get(
  '/shops/:id',
  ssgParams(async () => {
    const shops = await getShops()
    return shops.map((shop) => ({ id: shop.id }))
  }),
  async (c) => {
    const shop = await getShop(c.req.param('id'))
    if (!shop) {
      return c.notFound()
    }
    return c.render(
      <div>
        <h1>{shop.name}</h1>
      </div>
    )
  }
)

toSSG(app, fs)
```
</details>

```
$ npx esbuild --platform=node --bundle ssg-app.tsx | time node # with this PR
node  1.44s user 2.37s system 139% cpu 2.733 total
$ npx esbuild --platform=node --bundle ssg-app.tsx | time node # v4
node  19.70s user 5.25s system 218% cpu 11.410 total
```

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/helper/ssg/index.ts
+++ b/src/helper/ssg/index.ts
@@ -177,7 +177,10 @@ export const fetchRoutesContent = function* <
                 try {
                   const replacedUrlParam = replaceUrlParam(route.path, param)
                   let response = await pool.run(() =>
-                    app.request(replacedUrlParam, forGetInfoURLRequest)
+                    app.request(replacedUrlParam, {
+                      method: forGetInfoURLRequest.method,
+                      headers: forGetInfoURLRequest.headers,
+                    })
                   )
                   if (response.headers.get(X_HONO_DISABLE_SSG_HEADER_KEY)) {
                     resolveReq(undefined)
--- a/src/helper/ssg/index.ts
+++ b/src/helper/ssg/index.ts
@@ -91,7 +91,10 @@ type AddedSSGDataRequest = Request & {
  * Define SSG Route
  */
 export const ssgParams: SSGParamsMiddleware = (params) => async (c, next) => {
-  ;(c.req.raw as AddedSSGDataRequest).ssgParams = Array.isArray(params) ? params : await params(c)
+  if (c.req.raw.headers.get(X_HONO_SSG_HEADER_KEY)) {
+    ;(c.req.raw as AddedSSGDataRequest).ssgParams = Array.isArray(params) ? params : await params(c)
+    return c.text('')
+  }
   await next()
 }
 
@@ -162,6 +165,9 @@ export const fetchRoutesContent = function* <
           }
           forGetInfoURLRequest.ssgParams = [{}]
         }
+
+        // FIXME: I want to skip ssgParams somehow.
+        forGetInfoURLRequest.headers.delete(X_HONO_SSG_HEADER_KEY)
 
         resolveGetInfo(
           (function* () {
--- a/src/helper/ssg/index.ts
+++ b/src/helper/ssg/index.ts
@@ -4,6 +4,7 @@ import { inspectRoutes } from '../../helper/dev'
 import type { Hono } from '../../hono'
 import type { Env, MiddlewareHandler, Schema } from '../../types'
 import { bufferToString } from '../../utils/buffer'
+import { createPool } from '../../utils/concurrent'
 import { getExtension } from '../../utils/mime'
 import { joinPaths, dirname } from './utils'
 
@@ -103,6 +104,7 @@ export interface ToSSGOptions {
   beforeRequestHook?: BeforeRequestHook
   afterResponseHook?: AfterResponseHook
   afterGenerateHook?: AfterGenerateHook
+  concurrency?: number
 }
 
 /**
@@ -110,17 +112,25 @@ export interface ToSSGOptions {
  * `fetchRoutesContent` is an experimental feature.
  * The API might be changed.
  */
-export const fetchRoutesContent = async <
+export const fetchRoutesContent = function* <
   E extends Env = Env,
   S extends Schema = {},
   BasePath extends string = '/'
 >(
   app: Hono<E, S, BasePath>,
   beforeRequestHook?: BeforeRequestHook,
-  afterResponseHook?: AfterResponseHook
-): Promise<Map<string, { content: string | ArrayBuffer; mimeType: string }>> => {
-  const htmlMap = new Map<string, { content: string | ArrayBuffer; mimeType: string }>()
+  afterResponseHook?: AfterResponseHook,
+  concurrency?: number
+): Generator<
+  Promise<
+    | Generator<
+        Promise<{ routePath: string; mimeType: string; content: string | ArrayBuffer } | undefined>
+      >
+    | undefined
+  >
+> {
   const baseURL = 'http://localhost'
+  const pool = createPool({ concurrency })
 
   for (const route of inspectRoutes(app)) {
     if (route.isMiddleware) {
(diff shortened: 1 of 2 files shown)


usualoma:
I would like to get comments from anyone here who has an opinion on this PR.

nakasyou:
I understood a little, but I have idea for fetching external API.
```ts
app.get('/posts/:id', ssgCache(async () => {
   c.set('posts', await fetch('https://example.com/api/posts').then(res => res.json()))
}, c => {
  const posts = c.get('posts')
  ...
})
```
This idea is adding cache middle.
When `toSSG` is executed, it is only executed once in one `ssgCache`.
You can ignore rate-limit if you use this idea.

If this comment is something out of place, ignore this comment.

usualoma:
@nakasyo Thanks a lot. I don't think your comment is out of place.

I believe that the behavior of `ssgParams()` is worth considering. I would like to draft a separate issue on that.

This PR also mentions throttling of external API calls, but more importantly, this PR showed that significant performance improvement can be expected by "running in parallel without waiting for await".

The last two ( 32d4c8d, 21ddf56 ) may have been superfluous.

Assuming only dcf6d45 and 0ecf747 are used, even if no external API is called, the speed of SSG is still improved by this PR, even with small concurrency values.

watany-dev:
@usualoma 
Thank you very much. Waiting to create the htmlMap seems to be a heavy process, so I think it should be proactively addressed! One question, though. I believe this might be due to my lack of understanding, but in this case, would using a promise-returning function be faster than using an async function*?

usualoma:
@watany-dev Thanks for your comment.

There is potentially heavy asynchronous processing nested in two steps in `fetchRoutesContent`. If implemented naively with `async function*`, if the first stage (SSGParams) is heavy, it would not be possible to proceed immediately to the next loop in parallel. That is why this PR currently returning Promise.
It may be possible to write the function in such a way that it can be processed by `async function*`, for example by separating the functions.

* https://github.com/honojs/hono/blob/v4.0.0-rc.4/src/helper/ssg/index.ts#L142
* https://github.com/honojs/hono/blob/v4.0.0-rc.4/src/helper/ssg/index.ts#L153

watany-dev:
@usualoma 
Thank you. I understand now. I feel that splitting the large `fetchRoutesContent` will eventually be necessary, but at least for now, I believe your implementation is the best.

yusukebe:
Ooops. I've merged the v4 into the main and the v4 has been deleted automatically.

@usualoma Could you make the PR to the main or change the base branch?

usualoma:
@yusukebe Thank you. Currently I can't change the branch (v4 -> main) with my authority, can I do it with yours?

usualoma:
@yusukebe I know you are busy, so it will be fine later.

yusukebe:
@usualoma 

> Currently I can't change the branch (v4 -> main) with my authority, can I do it with yours?

Hmm. I've investigated how to do it with ChatGPT, but maybe we can't change the branch in the Web UI, so you should make another PR.

usualoma:
@yusukebe Thanks for the research. Understood!

sor4chi:
I’m sorry for the delay in replying.
Thank you for your understanding of my intentions.
I would like you to adopt usuaroma's code as it is without any particular objection.
It's completely upward compatible with my implementation...

PR #2186, opened 2024-02-10
refactor(ssg): create request from saved requestInit in order to avoid memory leak warnings

usualoma:
Continued from #2177

- [x] Add tests
- [x] Run tests


yusukebe (approved):
LGTM!

PR #2187, opened 2024-02-10
feat(ssg): generate files concurrently

usualoma:
Continued from #2169.

### Deletion of Commit

I've divided detailed topics into separate PRs and issues.

* #2186
* https://github.com/honojs/hono/issues/2175
    * https://github.com/honojs/hono/pull/2174
    * https://github.com/honojs/hono/pull/2178

### The default number of concurrency in SSG is 2.

The default value is debatable, but for this PR it was set to 2. The reasons are as follows.
* Even at 2 (or 1), this PR will parallelize writes to the file system, which is expected to be a sufficient speed improvement over the current main
* When calling external APIs, it is not good if the default is too large

Users can easily specify their own values with `concurrency`, so a small default is fine.

- [x] Add tests
- [x] Run tests


usualoma:
My work is done.

If there is enough discussion in #2169 to form a consensus on the implementation policy, I think we can merge in this PR content.

yusukebe (approved):
LGTM!

watany-dev:
It seems to have been CLOSED by some accident.

yusukebe:
@watany-dev 

Thanks for the notice.

@usualoma 

I've checked this again and run this real situation; it works well. I'd like to merge this, so please merge the "next".

usualoma:
@yusukebe merged!

yusukebe:
@usualoma 

Thanks! Merging now.

PR #2191, opened 2024-02-10
fix(context): remove duplicate definition of render method

usualoma:
The `render` method was defined twice in the `Context` class. This duplicate definition was causing a TypeScript error below if `tsc` is used to compile the project:

```
node_modules/hono/dist/types/context.d.ts(134,5): error TS2300: Duplicate identifier 'render'.
```

- [x] Run tests


yusukebe:
@usualoma 

Ah! I don't know why the code was like that, but it should be fixed. Thanks!

PR #2192, opened 2024-02-11
fix: `createHandler` Response Types (handler x1)

nakasyou:
I tried to resolve https://github.com/honojs/honox/issues/24.
But I can't resolve it because HonoX's `createRoute` is missing response type.
`createRoute` is using Hono's`createFactory`
So I search it, I found cause.

`createHandler` is missing response type when there is **one** handler.
We can solve this bug if this PR merged.

- [ ] Add tests
- [x] Run tests


nakasyou:
@yusukebe Could you review it?

yusukebe (approved):
LGTM!

yusukebe:
@nakasyou 

Thanks. That was a bug!

nakasyou:
Thank you!

PR #2193, opened 2024-02-11
Fix regex pattern with length limit at replaceUrlParam

the-fukui:
### Summary

There seems to be an issue with the function `replaceUrlParam` that replaces regular expressions in the path within the Client. Specifically, the problem arises when there's a length specification (`{}`) within the regular expression, and it appears at the end of the regex section.

```
/year/:year{[0-9]{4}}/month/:month{[0-9]{1,2}}
```

In this case, when the Client is given the params:

```
 param: {
   year: "2024",
   month: "2",
 }
```

it parses them as:

```
/year/2024}/month/2}
```

### Proposed Changes

The regular expression that "detects the regex section" within the path was only matching a single `}`. I propose to modify this to match one or more times (`+`), ensuring correct detection even when the length specification (`{}`) is at the end of the section.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/client/utils.ts
+++ b/deno_dist/client/utils.ts
@@ -9,7 +9,7 @@ export const mergePath = (base: string, path: string) => {
 
 export const replaceUrlParam = (urlString: string, params: Record<string, string>) => {
   for (const [k, v] of Object.entries(params)) {
-    const reg = new RegExp('/:' + k + '(\\{([^{}]*?(\\{[^{}]*?\\})?)+?\\})?')
+    const reg = new RegExp('/:' + k + '(?:{[^/]+})?')
     urlString = urlString.replace(reg, `/${v}`)
   }
   return urlString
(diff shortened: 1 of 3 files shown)


deno_dist/client/utils.ts
```diff
@@ -9,7 +9,7 @@ export const mergePath = (base: string, path: string) => {
 
 export const replaceUrlParam = (urlString: string, params: Record<string, string>) => {
   for (const [k, v] of Object.entries(params)) {
-    const reg = new RegExp('/:' + k + '({[^}]*})?')
+    const reg = new RegExp('/:' + k + '(\\{([^{}]*?(\\{[^{}]*?\\})?)+?\\})?')
```
yusukebe, line 12: If we just pass that test pattern, the following simple code will work as expected.

```ts
const reg = new RegExp('/:' + k + '({[^/]+})?')
```

yusukebe:
Hi @the-fukui !

Thanks. Indeed, it's a bug. I've commented on one. Please check it.

deno_dist/client/utils.ts
the-fukui: I committed it including non capturing group. Please review it.

yusukebe (approved):
LGTM!

yusukebe:
@the-fukui 

Looks good! Mergin now. Thanks for your contribution!

Issue #2016, opened 2024-01-18
Focus is lost on the `textarea`


yusukebe:
### What version of Hono are you using?

4.0.0-rc.1

### What runtime/platform is your app running on?

client

### What steps can reproduce the bug?

```tsx
import { render } from 'hono/jsx/dom'
import { useState } from 'hono/jsx'

const TextArea = () => {
  const [value, setValue] = useState('')

  const handleChange = (e: Event) => {
    const target = e.target as HTMLTextAreaElement
    setValue(target.value)
  }

  return (
    <div>
      <textarea onChange={handleChange} />
      <div>{value}</div>
    </div>
  )
}

const root = document.getElementById('root') as HTMLElement

render(<TextArea />, root)
```

### What is the expected behavior?

As the same as React:

https://github.com/honojs/hono/assets/10682/674e1c32-3386-4544-8aef-a7e1d3ac5fce

### What do you see instead?

After typing a single character in the text area, the focus is lost and I cannot type continually.

https://github.com/honojs/hono/assets/10682/313bdef0-ff22-432b-8b85-d975f07fa54a

Demo: https://hono-client-components.pages.dev/,

yusukebe:
Fixed!

PR #2196, opened 2024-02-12
fix(jsx/dom): Do not call insertBefore if the element position does not change

usualoma:
Fixed the same problem as #2016 when one child is updated in an element having multiple children.

- [x] Add tests
- [x] Run tests


yusukebe (approved):
LGTM!

yusukebe:
@usualoma 

Thanks! I'll merge it now.

PR #2197, opened 2024-02-12
feat(jsx): more react staff

usualoma:
Once #2196  and this PR are merged, I believe the following example will work with 'hono/jsx/dom' (without react).

https://tutorial.jotai.org/examples/textLenght

### New staff members

`useReducer`, `useDebugValue`, `createElement`

`useDebugValue` is just a dummy and useless, but it was added to avoid import errors.

### export default

Although rarely used in this way today, the following library for React, which is written as follows, can be used as is without modification.

```ts
import React from 'react' // reaplced with 'hono/jsx/dom'

const App = () => {
  const [value, updateValue] = React.useState(0)
  // ...
}
```

- [x] Add tests
- [x] Run tests


deno_dist/jsx/dom/index.ts
```diff
@@ -36,3 +53,55 @@ export const cloneElement = <T extends JSXNode | JSX.Element>(
...
+  ErrorBoundary,
+  createContext,
+  useContext,
+  memo,
+  isValidElement,
+  createElement,
+  cloneElement,
+}
+
+export default {
```
yusukebe, line 82: Does not it export `jsx`? I know it's not necessary only to make it compatible with React. So, it might not be the problem, but I'm just curious.

usualoma: @yusukebe Thank you!
I've been struggling with this.

I think `jsx` will be used in the old React import style, as shown below. I didn't add it because I didn't think it would be used in "hono/jsx/dom" usage.  I think "hono/jsx/dom" is only used with "jsxImportSource".

```ts
/* @jsx jsx */
import { jsx } from "hono/jsx/dom";
```

That said, `jsx` has been exported from 'hono/jsx' and I think you are right that it would be easier to understand if `jsx` was exported from “hono/jsx/dom" as well, as you pointed out.

What should we do, export it?

yusukebe: @usualoma 

I was wondering, but I think it should export `jsx`. Not many people may use `jsx`, but I may sometimes use `jsx` directly.

For example, in the case of HonoX, `jsx` is imported [here](https://github.com/honojs/honox/blob/a186c42bfd35c05954df614078e89fcb19c1263c/src/client/client.ts#L2C1-L2C56) from `hono/jsx/dom/jsx-runtime`, but `hono/jsx/dom` is better.

yusukebe: @usualoma 

Ah, perhaps, if that HonoX case, can we use `createElement`?

usualoma: @yusukebe Oh, I see.

It is customary for a function named `jsx` to be exported from '*/jsx-runtime’. I think that [this](https://github.com/honojs/honox/blob/a186c42bfd35c05954df614078e89fcb19c1263c/src/client/client.ts#L2C1-L2C56) is acceptable to use. 

```ts
import { jsx } from "react/jsx-runtime";
console.log(jsx('div', {}, 'Hello, World!'))
```

```ts
import { jsx } from "preact/jsx-runtime";
console.log(jsx('div', {}, 'Hello, World!'))
```


Also, since `jsx` is an optimized version of `createElement`, I think it is not a bad idea to use `import { jsx as jsxFn } from '*/jsx-runtime'` instead of `createElement`.


### About HonoX

I don't think the current code that I am importing from hono/jsx/dom/jsx-runtime will result in minimal output and I don't think it is bad usage. But on the other hand,  you are right, `import { createElement } from 'hono/jsx/dom'` also makes sense since it is intended to be used as a `createElement`.

I see no problem either way.

yusukebe: @usualoma 

Thanks for the explanation! Then, how about go with "exporting `jsx`"? One more reason is that `hono/jsx` is exporting `jsx`, so I want to make it the same as that.

If you agree, could you add that?

usualoma: @yusukebe OK
Exported by b188977cdfa5ac0421e03a4fd9f7103978f6d85d!

yusukebe: @usualoma 

Thanks! Looks good! I'll merge this later.

watany-dev:
もちろんapp.get(`/tags/${encodeURI(t)}` とすればdev時には見れるんだけど、ファイルとして作るときは /tags/%E3%82%BF%E3%82%B0/index.html ではなくて /tags/タグ/index.html って名前で作ってほしい……。

watany-dev:
It seems to have been CLOSED by some accident.

yusukebe:
Merging!

PR #2211, opened 2024-02-14
fix(jsx-renderer): support async component

### LINKED ISSUE #2207: Get response [object Promise] in async await jsxRenderer

4.0.1

Cloudflare Workers

This code will render `[object Promise]`

```ts
app
    .use(
      '/test',
      jsxRenderer(async ({ children }) => { // w/o async, await, will be successful
        return await (<div>{children}</div>);
      }),
    )
    .get('/test', async (c) => {
      return await c.render(<span>Hello</span>);
    });
```

Get response like below:
```
<div><span>Hello</span></div>
```

### What do you see instead?

[object Promise] 

### Additional information

_No response_

yusukebe:
Fixes #2207

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/middleware/jsx-renderer/index.ts
+++ b/deno_dist/middleware/jsx-renderer/index.ts
@@ -24,7 +24,7 @@ const createRenderer =
     component?: FC<PropsForRenderer & { Layout: FC }>,
     options?: RendererOptions
   ) =>
-  async (children: JSXNode, props: PropsForRenderer) => {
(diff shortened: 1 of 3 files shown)


yusukebe:
Hey @usualoma !

Could you review this one?

yusukebe:
@usualoma

Thanks! That's an elegant way. I've applied your patch. Can you review this again?

PR #2216, opened 2024-02-15
refactor: jsx streaming

usualoma:
There is no change in behavior, but it is not a useful refactoring for existing applications, so it is a question of whether to include it in a patch version. A minor version might be fine.

* 867f85722b33038606a8f3d7b826742d490a0134 : Depending on the content, it may look like the following, and in this case a JS error occurs.
    * `<template id="E:0"></template><! -E:0->`
* 71c8ec52bbcedf1149420a81bb9fd0b53ce791ec : When handling https://github.com/honojs/honox/issues/47, it is useful to be able to retrieve the resolved value later, so the data attribute was added. The data will be slightly larger, but I don't think it will be a problem.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/jsx/components.ts
+++ b/deno_dist/jsx/components.ts
@@ -79,7 +79,7 @@ export const ErrorBoundary: FC<
       }
       return buffer
         ? ''
-        : `<template>${fallbackResString}</template><script>
+        : `<template data-hono-target="E:${index}">${fallbackResString}</template><script>
 ((d,c,n) => {
 c=d.currentScript.previousSibling
 d=d.getElementById('E:${index}')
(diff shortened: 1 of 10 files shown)


yusukebe:
Hi @usualoma 

Looks great! I think there is no problem with this code.

I'm trying to decide whether to put this in the patch release or the minor version. So, I'll merge this later.

Maybe 4.1 will be released earlier without many features.

yusukebe:
@usualoma 

I'm merging and releasing this in a patch version, as there are no clear feature additions in this PR and I don't want this to become a blocker.

Thanks!

PR #2218, opened 2024-02-15
fix(context): Inherit current status if not specified

usualoma:
There is a problem that the status code is not inherited when using `jsxRenderer` with `stream:true`.

```ts
      app.use(
        '*',
        jsxRenderer(
          ({ children }) => {
            return (
              <html>
                <body>{children}</body>
              </html>
            )
          },
          { stream: true }
        )
      )
```

Here we call `c.body()` with two arguments, but since status is not inherited in this call, it is always 200.

https://github.com/honojs/hono/blob/48c6ce9b6f630e215bd1142af2b52307c3f1ac52/src/middleware/jsx-renderer/index.ts#L54-L62

This issue has caused problems with 404s turning into 200s in uses such as honox.

https://github.com/honojs/honox/blob/main/src/server/server.ts#L179-L182

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -359,7 +359,7 @@ export class Context<
       const headers = setHeaders(new Headers(arg.headers), this.#preparedHeaders)
       return new Response(data, {
         headers,
-        status: arg.status,
+        status: arg.status ?? this.#status,
       })
     }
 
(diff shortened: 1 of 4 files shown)


yusukebe (approved):
Ah, this was a bug!

LGTM.

yusukebe:
@usualoma 

Thanks!

PR #2219, opened 2024-02-15
feat(helper): new helper toPunyCode

EdamAme-x:
before
```ts
c.redirect("https://🔥.com") // throw Error
```

after
```ts
import { toPunyCode } from "hono/punycode"
...
c.redirect(toPunyCode("https://🔥.com")) // redirect to https://xn--4v8h.com
```

## Options
```ts
toPunyCode("🔥.com", {
    strict: true
}) // 🔥.com => 🔥.com (not URL)

toPunyCode("🔥.com", {
    strict: false
}) // 🔥.com => xn--4v8h.com (URL)

// default strict is true
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/helper/punycode/index.ts
+++ b/deno_dist/helper/punycode/index.ts
@@ -1,0 +1,42 @@
+/**
+ * Converts string to puny code
+ * @param string - string to be converted
+ * @param options - options
+ * @returns converted string (puny code)
+ * @example
+ * ```typescript
+ * const url = toPunyCode('https://🔥.com'); // https://xn--4v8h.com
+ * // usage
+ * c.redirect(url);
+ * ```
+ */
+
+const prefix = 'http://'
+
+export const toPunyCode = (
+  string: string,
+  options: {
+    strict?: boolean
+  } = {
+    strict: true,
+  }
+): string => {
+  const { strict } = options
+
+  let punyCode = ''
+  try {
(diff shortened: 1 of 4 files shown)


EdamAme-x:
hi @yusukebe 
I ran into this problem when creating a Shortened link site (as x.gd) on Hono and decided to implement it.

yusukebe:
Hi @EdamAme-x 

This is interesting, but I think there are few users who use it. So it is difficult to apply it as a built-in middleware. Adding more features means higher maintenance costs, so we don't want to include features that are rarely used.

Perhaps it would be interesting to create a repository of middleware by the user community, separate from the honojs org, that would not be included in the build-in, and that you could manage.

EdamAme-x:
hi @yusukebe 
I can agree with that.
But I'm sad that users who want to use Unicode for url will give up. 

I do not believe that if I were to manage this I would be able to reach users with such requests.

So I would like to be listed as a guide on hono.dev. What do you think?

yusukebe:
> So I would like to be written as a guide on hono.dev. What do you think?

This is a great idea!

We've created the `snippets` section:

https://hono.dev/snippets/top

How about creating a category "Recipes" in it? So how about writing how to build an application for each use case in the category?

EdamAme-x:
hi @yusukebe 
i ready
https://github.com/honojs/website/pull/245

PR #2220, opened 2024-02-15
fix(hono-base): custom not found with middleware like compress

yusukebe:
Fixed Custom NotFound not working with middleware that creates a new Response object after `next()` such as Compress Middleware. And added proper tests.

This was fixed by #2080, but it may not be merged to the main branch well. And there is no test to handle it, so we can't notice it has the bug.

- [x] Add tests
- [x] Run tests


yusukebe:
Hey @usualoma !

Could you see this?

usualoma:
@yusukebe LGTM!

PR #2224, opened 2024-02-16
feat: Add a "cause" option to HTTPException

### LINKED ISSUE #2170: Add a "cause" option to HTTPException

The JavaScript Error class can be passed the [cause option](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause).
This is a property that indicates the cause of the error and is used to catch the error and construct the error including the original error when it is thrown again.

The [HTTPException class](https://github.com/honojs/hono/blob/main/src/http-exception.ts) defined by Hono does not have a cause property.
This makes it impossible for monitoring tools such as Sentry to determine the detailed cause of the error when throwing an HTTPException error.
It would be easier to debug if the cause property could be passed to HTTPException as well.

Karibash:
Add a "cause" option to HTTPException for easier debugging.
close #2170 

### How to use

```ts
app.get('/users', (c) => {
  try {
    const result = await db.select().from(users);
    return c.json(result);
  } catch (error) {
    throw new HTTPException(500, {
      message: 'Internal Server Error',
      cause: error,
    })
  }
})
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/http-exception.ts
+++ b/deno_dist/http-exception.ts
@@ -3,6 +3,7 @@ import type { StatusCode } from './utils/http-status.ts'
 type HTTPExceptionOptions = {
   res?: Response
   message?: string
+  cause?: unknown
 }
 
 /**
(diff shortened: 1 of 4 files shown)


yusukebe:
Hi @Karibash 

Thanks for the PR. This is a new feature. So, it will be introduced in the next "minor" version. It can't be merged to "main" immediately. I'll prepare "next" branch later. Then, I'll review the details and after that, merge it later. Please wait a little.

yusukebe (approved):
Looks good to me!

yusukebe:
Hey @Karibash !

I'll merge this into the "next" branch for `v4.1.0`. Thanks! It will be released maybe soon.

PR #2229, opened 2024-02-16
docs(contribution): add install flag `--frozen-lockfile`

ryuapp:
Match the install command with Installing dependencies.

- [ ] Add tests
- [ ] Run tests


yusukebe:
Hi @ryuapp 

Good! Merging now. Thanks!

PR #2230, opened 2024-02-17
fix(jsx/dom): fix finding element to insert before

usualoma:
fixes https://github.com/honojs/honox/issues/68

- [x] Add tests
- [x] Run tests


yusukebe (approved):
I was able to confirm this works well. LGTM!

PR #2231, opened 2024-02-17
refactor(jsx): shorten use hook a bit

usualoma:
This is a very small refactoring and changes nothing in the specifications.

When processing the result of an "if/else" type of promise, it is shorter and clearer to pass two functions to `.then()` than to use `.catch()`.

- [x] Run tests


yusukebe (approved):
Ah, I see. We can pass two functions to `then`!
Looks good!

PR #2234, opened 2024-02-17
fix(validator): use the cached content

### LINKED ISSUE #2163: Inconsistent Request Body Caching Between c.req.json() and Validator Middleware

3.12.11

Cloudflare Workers

I've encountered an issue with how Hono caches the request body in different formats, leading to unexpected behavior when using middlewares in a specific order. 

When one middleware calls `c.req.json()`, it updates `bodyCache[json]` without also updating `bodyCache[arrayBuffer]`. 
If we later use the `validator` middleware, which expects the `arrayBuffer` to be cached, it results in an error for a malformed request.

```ts
app.post(
  '/test',
  async (c, next) => {
    const body = await c.req.json();
    await next();
  },
  validator('json', (value, c) => {
    return value;
  }),
  async (c) => {
    return c.json({ message: 'OK' }, 200);
  },
);
```

The request passes through both middlewares without issues, with the validator correctly validating the parsed JSON.

### What do you see instead?

Receive a 400 Bad Request error.
```
[  <-- POST /TEST]: undefined
✘ [ERROR] Error: Malformed JSON in request body
```

### Additional information

The validator middleware throws an error indicating a malformed request body, despite the body being valid JSON. This seems to occur because c.req.json() does not update bodyCache[arrayBuffer], leading to a mismatch in the cached body data.

```ts
// request.ts
  private cachedBody = (key: keyof Body) => {
// ...
    if (bodyCache.arrayBuffer) {
      return (async () => {
        return await new Response(bodyCache.arrayBuffer)[key]()
      })()
    }
    return (bodyCache[key] = raw[key]()) // doesn't update arrayBuffer
  }
```

```ts
// validator/validator.ts
        /**
         * Get the arrayBuffer first, create JSON object via Response,
         * and cache the arrayBuffer in the c.req.bodyCache.
         */
        try {
          const arrayBuffer = c.req.bodyCache.arrayBuffer ?? (await c.req.raw.arrayBuffer()) // Uses arrayBuffer
          value = await new Response(arrayBuffer).json()
          c.req.bodyCache.json = value
          c.req.bodyCache.arrayBuffer = arrayBuffer
        } catch {
          console.error('Error: Malformed JSON in request body')
          return c.json(
            {
              success: false,
              message: 'Malformed JSON in request body',
            },
            400
          )
        }
```

yusukebe:
Fixes #2163

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/validator/validator.ts
+++ b/deno_dist/validator/validator.ts
@@ -67,9 +67,12 @@ export const validator = <
           const message = `Invalid HTTP header: Content-Type=${contentType}`
           throw new HTTPException(400, { message })
         }
+        if (c.req.bodyCache.json) {
+          value = await c.req.bodyCache.json
+          break
+        }
         /**
-         * Get the arrayBuffer first, create JSON object via Response,
-         * and cache the arrayBuffer in the c.req.bodyCache.
+         * Get the arrayBuffer, create JSON object via Response, and cache the arrayBuffer in the c.req.bodyCache if there is no cached content in c.req.bodyCache.json.
          */
         try {
           const arrayBuffer = c.req.bodyCache.arrayBuffer ?? (await c.req.raw.arrayBuffer())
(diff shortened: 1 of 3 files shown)


yusukebe:
Hey @usualoma !

Could you review this one?

usualoma (approved):
Good 👍

yusukebe:
Thanks!

PR #2241, opened 2024-02-18
feat(jsx/dom): support namespace for svg and mathml elements

usualoma:
fixes https://github.com/honojs/honox/issues/71

In the case of SVG and MathML elements, it is necessary to use createElementNS instead of createElement to support them.

### Differences from React

In React, it is necessary to specify in lower camel case, and I think the output will be converted to kebab case. This PR does not do that conversion, but I think it is OK to have hono specify kebab case from the beginning. If there are many requests, we will consider adding a conversion process.

```tsx
   // in React 
   <circle
     cx="40"
     cy="40"
     r="35"
     stroke="black"
     strokeWidth="4"
     fill="lightgrey"
   /> // => <circle cx="40" cy="40" r="35" stroke="black" stroke-width="4" fill="lightgrey" />
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/jsx/dom/render.ts
+++ b/deno_dist/jsx/dom/render.ts
@@ -5,11 +5,17 @@ import type { Context as JSXContext } from '../context.ts'
 import { globalContexts as globalJSXContexts } from '../context.ts'
 import type { EffectData } from '../hooks/index.ts'
 import { STASH_EFFECT } from '../hooks/index.ts'
+import { useContext, createContext } from './index.ts'
 
 const eventAliasMap: Record<string, string> = {
   Change: 'Input',
   DoubleClick: 'DblClick',
-}
+} as const
+
+const nameSpaceMap: Record<string, string> = {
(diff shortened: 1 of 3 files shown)


yusukebe (approved):
Ohh, I see, using `useContext`! 

LGTM!

yusukebe:
@usualoma 

Thanks!

> but I think it is OK to have hono specify kebab case from the beginning. If there are many requests, we will consider adding a conversion process.

I think that's fine with me, too.

I would like to merge this PR, but I am considering whether to include it in a patch release or a minor release, so please wait a bit.

zxpzdtom:
> @usualoma
> 
> Thanks!
> 
> > but I think it is OK to have hono specify kebab case from the beginning. If there are many requests, we will consider adding a conversion process.
> 
> I think that's fine with me, too.
> 
> I would like to merge this PR, but I am considering whether to include it in a patch release or a minor release, so please wait a bit.

@yusukebe 
Sounds good! I'm excited about the release.

PR #2236, opened 2024-02-17
fix(ssg): path of already extention

watany-dev:
- before
    - /index.html -> static/index.html.html 
- after
    - /index.html -> static/index.html

### Author should do the followings, if applicable

- [x] Add tests
- [x] Run tests
- [x] `yarn denoify` to generate files for Deno

Issue #2239, opened 2024-02-17
Improvement of extension determination


usualoma:
### What is the feature you are proposing?

With #2236, the presence or absence of extensions is now automatically determined, but I think this behavior should be improved.

### Test

Currently, adding the following test succeeds, but is this the expected result?

```diff
diff --git a/src/helper/ssg/index.test.tsx b/src/helper/ssg/index.test.tsx
index dde90e5a..07726d04 100644
--- a/src/helper/ssg/index.test.tsx
+++ b/src/helper/ssg/index.test.tsx
@@ -297,6 +297,10 @@ describe('saveContentToFiles function', () => {
       ['/about/', { content: 'About Page', mimeType: 'text/html' }],
       ['/bravo/index.html', { content: 'About Page', mimeType: 'text/html' }],
       ['/bravo/index.tar.gz', { content: 'About Page', mimeType: 'application/gzip' }],
+      ['/bravo/release-4.0.0', { content: 'Release 4.0.0', mimeType: 'text/html' }],
+      ['/bravo/2024.02.18-sweet-memories', { content: 'Sweet Memories', mimeType: 'text/html' }],
+      ['/bravo/deep.dive.to.html', { content: 'Deep Dive To HTML', mimeType: 'text/html' }],
+      ['/bravo/alert.js', { content: 'alert("evil content")', mimeType: 'text/html' }],
       ['/bravo.text/index.html', { content: 'About Page', mimeType: 'text/html' }],
       ['/bravo.text/', { content: 'Bravo Page', mimeType: 'text/html' }],
     ])
@@ -312,6 +316,10 @@ describe('saveContentToFiles function', () => {
     expect(fsMock.writeFile).toHaveBeenCalledWith('static/bravo/index.tar.gz', 'About Page')
     expect(fsMock.writeFile).toHaveBeenCalledWith('static/bravo.text/index.html', 'About Page')
     expect(fsMock.writeFile).toHaveBeenCalledWith('static/bravo.text/index.html', 'Bravo Page')
+    expect(fsMock.writeFile).toHaveBeenCalledWith('static/bravo/release-4.0.0', 'Release 4.0.0')
+    expect(fsMock.writeFile).toHaveBeenCalledWith('static/bravo/deep.dive.to.html', 'Deep Dive To HTML')
+    expect(fsMock.writeFile).toHaveBeenCalledWith('static/bravo/2024.02.18-sweet-memories', 'Sweet Memories')
[...]

PR #2243, opened 2024-02-19
fix: Dot-containing paths do not correctly receive extension completion.

watany-dev:
fixes: https://github.com/honojs/hono/pull/2236

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/helper/ssg/index.ts
+++ b/deno_dist/helper/ssg/index.ts
@@ -30,13 +30,27 @@ export interface ToSSGResult {
   error?: Error
 }
 
+const getPathWithExpectedExtension = (
+  routePath: string,
+  expectedExtension: string
+): string | null => {
+  const match = routePath.match(/\.(?<extension>[^\.\/]+)$/)
+  const existingExtension = match ? match[0] : ''
+
+  if (existingExtension === `.${expectedExtension}`) {
+    return routePath
+  }
+  return null
+}
+
 const generateFilePath = (routePath: string, outDir: string, mimeType: string) => {
-  const hasExtension = /\.[^\/]+$/.test(routePath)
-  if (hasExtension) {
-    return joinPaths(outDir, routePath)
+  const extension = determineExtension(mimeType)
+
+  const adjustedPath = getPathWithExpectedExtension(routePath, extension)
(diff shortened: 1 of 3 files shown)


watany-dev:
This partially resolves this issue, and allows the "." except for the extension completion accuracy.
https://github.com/honojs/hono/issues/2239

watany-dev:
@usualoma 
Can I have it reviewed?

usualoma:
@watany-dev Thank you!

First, in response to your comment in #2239, I think your point below is spot on.

> For those who want to use extensions like .html.html, they can simply specify it explicitly in the Route :)

We can write as follows

```ts
app.get(
  '/shops/:id.html',.
  async (c) => { ... }
)
```

Thus, I think the question I raised in #2239 will be answered by this PR.

I have no objection to this PR test case, but I think `.endsWith()` would be sufficient for `getPathWithExpectedExtension()`. No?

```ts
const generateFilePath = (routePath: string, outDir: string, mimeType: string) => {
  const extension = determineExtension(mimeType)

  if (routePath.endsWith(`.${extension}`)) {
    return joinPaths(outDir, routePath)
  }

  if (routePath === '/') {
    return joinPaths(outDir, `index.${extension}`)
  }
  if (routePath.endsWith('/')) {
    return joinPaths(outDir, routePath, `index.${extension}`)
  }
  return joinPaths(outDir, `${routePath}.${extension}`)
}
```

watany-dev:
Thanks for reviewing this!
I didn't know there was an endWith in the standard API. This one is great.
https://developer.mozilla.org/ja/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith

usualoma:
Oh, sorry, not quite right, the following doesn't work good. It would be a variable `id.html`

```ts
app.get(
  '/shops/:id.html',.
  async (c) => { ... }
)
```

Well, okay, that's a separate issue, so let's consider it in other issue 🙇

watany-dev:
> I have no objection to this PR test case, but I think .endsWith() would be sufficient for getPathWithExpectedExtension(). No?
This has been fixed.

yusukebe (approved):
Looks good to me! I'll merge.

Issue #2244, opened 2024-02-19
SSG: Should the default Content-Type be text/plain?


usualoma:
### What is the feature you are proposing?

Currently, if Content-Type is not found, it is processed as text/plain, resulting in a .txt file name.
https://github.com/usualoma/hono/blob/dcb740176f8e4e26783bac05ba5bd6ea7ef4e09d/src/helper/ssg/index.ts#L164

I wonder if there is a case where this fallback works effectively. 

hono's app and external APIs almost never return a response without Content-Type, and if Content-Type is not found, it is most likely that "some error has occurred". I think it is better to make it an error as the default behavior.

3w36zj6:
This issue was probably just left open by mistake and should have been closed with https://github.com/honojs/hono/pull/2666. @yusukebe

yusukebe:
@3w36zj6 Thanks!

This issue is closed by #2244

Issue #2180, opened 2024-02-09
SSG: If the status code is anything other than 200, the output is not output by default


usualoma:
### What is the feature you are proposing?

Even now, afterResponseHook can be used to check status codes and process errors, but by default, processing continues with any status code other than 200. In general, if a request for SSG returns anything other than 200, an error should be reported. (Ignoring it and continuing is also not a good idea.)

usualoma:
When it comes to POST requests, POST is generally a request to change state, and even if the POST results is generated by `render()`, there is often POST data in it. I do not think it is correct behavior for POST to be written out by default.

usualoma:
Or... perhaps, as far as CloudFlare Pages are concerned, what we need is this.

```ts
app.get('/404', onlySSG(), (c) => {
  return c.render(<h1>Not found! </h1>)
})
```

usualoma:
Also, if 404s due to accidents are not ignored, it will be quite painful to generate a sitemap, since it will be necessary to select the target files when generating the sitemap.

#2125

usualoma:
Oh, or maybe it's a bit redundant, but maybe 404.html should be generated from the plugin.

As this comment states,

https://github.com/honojs/hono/issues/2125#issuecomment-1920167497
 
Write the following.

```ts
import { notFound } from 'hono/ssg/plugins'

toSSG(app, fs, {
  plugins: [notFound()], // write /404.html by default
})
export default app
```

or

```ts
import { notFound } from 'hono/ssg/plugins'

toSSG(app, fs, {
  plugins: [notFound({routes => ['/404.html', '/blog/404.html']})], // write /404.html and /blog/404.html
})
export default app
```

usualoma:
I'll write simply one more time

For example, when we call an external API to generate content, we sometimes get a "temporary error that results in a non-200 result". Currently, in that case, I think the result (contains an error message) is saved without issuing a warning. (Am I wrong?)
When you publish it to the public server, the error page will serve with status code 200.

I think this behavior should be fixed.

3w36zj6:
As of v4.10.0 (https://github.com/honojs/hono/pull/4394), a default plugin has been added. By default, responses other than `200 OK` are ignored, and users can disable this behavior if needed. Based on this, I think we can re-evaluate the direction going forward.

Regarding 404 pages for static site hosting platforms, the following approach should work fine:

```tsx
app.get('/404', onlySSG(), (c) => {
  return c.render(<h1>Not found! </h1>)
})
```

[...]

PR #2247, opened 2024-02-20
fix(ssg):  Error when ContentType is missing

watany-dev:
fixes: https://github.com/honojs/hono/issues/2244

This sets the specification for the defaultContentType of SSG, which was implicitly set to `text/plain` until now.

However, as per the issue mentioned above, this may not be appropriate. Therefore, we have added the option defaultContentType. In normal settings, routes like c.redirect, which include no content-type, will result in an error.

 If set, it will fallback to the specified value. One issue is that setting the DefaultContent-type to an arbitrary value like 'aaaa' results in an empty .html file being output. This is because the default value in determineExtension is .html, but there is room for discussion on what the specification should be regarding this issue.

- [x] Add tests
- [x] Run tests


watany-dev:
I made it, but if we are going to solve the .html completion problem, it might not be much different from not providing a defaultContentType option and automatically making it an error.

usualoma:
Hi @watany-dev. Thank you!

I can't think of any situation where `defaultContentType` would be needed. I think it would be very rare, if any, so instead of providing a `defaultContentType` option, I think we would be better to encourage to use `afterResponseHook()` for prepareing fallback.

yusukebe:
Hey @watany-dev @usualoma !

I agree with @usualoma. I don't think `defaultContentType` is necessary. That's because he's right, the Content-Type is set in the Response in many cases.

> However, as per the issue mentioned above, this may not be appropriate. Therefore, we have added the option defaultContentType. In normal settings, routes like c.redirect, which include no content-type, will result in an error.

Certainly the case for redirect responses must be considered, but it is a question of how to handle 301 and 302 status codes. We can discuss that in #2180.

watany-dev:
close.

Ref: https://github.com/honojs/hono/issues/2284

PR #1614, opened 2023-10-23
feat: types for JSX elements

yusukebe:
This PR introduces type definitions for JSX elements to enhance the developer experience.

### Approach

There are several ways to add types to JSX elements. With this PR, the approach taken is as follows:

* Define types manually, without referring to the DOM lib.
* Main tags are ported from React's definitions: https://www.npmjs.com/package/@types/react?activeTab=code
* Removed some elements and attributes such as ARIA's to keep it minimum.
* Renamed some properties for HTML compatibility, e.g., `readOnly` to `readonly`.
* For compatibility with unknown elements and attributes, it allows any elements and attributes.

### Screenshots

Screenshot 2023-10-23 at 17 47 03

Screenshot 2023-10-23 at 17 49 03

### Allowing any elements and attributes

In this PR, any elements and any attributes are allowed. This is defined as follows:

Attributes:

```ts
type AnyAttributes = { [attributeName: string]: any }

//...

interface HTMLAttributes extends JSXAttributes, AnyAttributes {
```

Elements:

```ts
type IntrinsicElements = IntrinsicElementsDefined & { [tagName: string]: Props }
```

This means that the feature doesn't support type safety completely, but it offers flexibility. You can use user-defined tags and maintain backward compatibility.

This doesn't throw an error.

Screenshot 2023-10-23 at 18 06 09

I believe it shouldn't be too strict in this case for JSX, as flexibility is important and this approach can sufficiently improve DX.

---

Resolves #1409

PR #2248, opened 2024-02-20
refactor(context): skip jsx type import

### LINKED ISSUE #2190: `hono/client` is overriding React types in a Next.js project

4.0.0

Node.js 20

If we're on a Next.js 14.1 project and we're importing `hono` as a dependency to use the `hono/client`, we can't use the following code : 
```typescript
import type { ComponentProps, ElementType, PropsWithChildren, ReactNode } from "react";

type Props = Omit<ComponentProps<"div">, "className"> & {
  as?: ElementType;
  children: ReactNode;
};

export function Container({ className, children, as }: Props) {
  const Tag = as || "div";
  return (
    <Tag>
      {children}
    </Tag>
  );
}
```

Because the following lines are overriding the JSX.ElementType type
https://github.com/honojs/hono/blob/ca6f36511094522b84b6575d99bde4728bcea87c/src/jsx/base.ts#L13C1-L24C2

Here is a reproduction: https://codesandbox.io/p/devbox/elegant-bogdan-ktsdc4

We shouldn't have the following errors : 
> 1. JSX element type 'Tag' does not have any construct or call signatures. [2604]
> 2. 'Tag' cannot be used as a JSX component.
     Its type 'React.ElementType' is not a valid JSX element type.
       Type 'number' is not assignable to type 'ElementType'. [2786]

### What do you see instead?

_No response_

### Additional information

_No response_

usualoma:
Maybe fixes #2190

### #2190 What is at stake here?

Importing "hono/client" is causing problems because hono's JSX types are being read.

```ts
```

This is because "context.ts" imports "hono/jsx" and "hono/middleware/jsx-renderer". This PR eliminates this.

### Where are the conflicts?

However, even after resolving this, if I try to use `React.ElementType` and "hono/jsx" at the same time (I am not aware if there is such a case or not), we still have problems. 

The following part cannot avoid conflict with the `React.ElementType` definition in any way.
https://github.com/usualoma/hono/blob/33f2d484f651ea55913ebe7022ed028e8ac7743f/src/jsx/base.ts#L20-L22

If a workaround is needed, I think the only way is to remove `[tagName: string]: Props`.

### DIFF
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -1,5 +1,3 @@
-import type { FC } from './jsx/index.ts'
-import type { PropsForRenderer } from './middleware/jsx-renderer/index.ts'
 import type { HonoRequest } from './request.ts'
 import type { Env, FetchEventLike, NotFoundHandler, Input, TypedResponse } from './types.ts'
 import { resolveCallback, HtmlEscapedCallbackPhase } from './utils/html.ts'
@@ -22,6 +20,12 @@ interface DefaultRenderer {
 }
 
 export type Renderer = ContextRenderer extends Function ? ContextRenderer : DefaultRenderer
+export type PropsForRenderer = [...Required<Parameters<Renderer>>] extends [unknown, infer Props]
+  ? Props
+  : unknown
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type Layout<T = Record<string, any>> = (props: T) => any
 
 interface Get<E extends Env> {
   <Key extends keyof ContextVariableMap>(key: Key): ContextVariableMap[Key]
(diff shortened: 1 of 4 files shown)


yusukebe:
Thanks @usualoma !

As you mentioned, this PR can fix the issue.

> If a workaround is needed, I think the only way is to remove `[tagName: string]: Props`.

Why do we need `[tagName: string]: Props`? To allow types other than the elements listed the followng, right?

https://github.com/usualoma/hono/blob/33f2d484f651ea55913ebe7022ed028e8ac7743f/src/jsx/intrinsic-elements.ts#L609

usualoma:
@yusukebe 
Yes, I didn't actually add the commits involved in that, but it was added here first,

https://github.com/honojs/hono/pull/306

And I am aware that each HTML element was defined here, so it is no longer needed, but it is left in place, probably for compatibility. If the app is using custom elements, it would be useful to not have to set anything to avoid errors.

https://github.com/honojs/hono/pull/1614

Although, well, I think there is an option to remove `[tagName: string]: Props` to be the same as React.

yusukebe:
@usualoma 

Yeah, I wrote that :)

> If the app is using custom elements, it would be useful to not have to set anything to avoid errors.

I know people use custom elements, so I am in favor of keeping the type loose.

Okay. Currently, I'll merge this PR without removing `[tagName: string]: Props`. Thanks!

yusukebe (approved):

PR #2258, opened 2024-02-23
test: add router/common.case.test.ts

### LINKED ISSUE #1953: Refactor router tests
Currently, we are writing tests for each router, but we want to use one common set of tests to test all routers.

usualoma:
fixes #1953

### Tidy up

Suppress errors from `yarn lint:fix`

* 2c521b2b771bacbba6b90072b52c7cc0849714fe
* 57cef10b0dee45a40b75e7fde8a0bd097bd23976

### Tweaks vitest.config.ts

Exclude "*.case.test.ts" from vitest

These special files are used to share test cases between different modules or runtime environments.
* shared among multiple routers
* shared by renderToString and renderToDom

Since these files are test files, they should be excluded from the build, in which case it is convenient to name them "*.test.ts". However, the file itself does not contain the test execution, so it should be excluded from `vitest watch`. With that motivation, we added "*.case.test.ts" to the exclude of vitest.

### src/router/common.case.test.ts

All routers run tests using `runTest` exported from this file. For routers other than "TireRouter", there are a few test cases that do not pass (this was intended at the time of implementation), so we need to explicitly specify to skip them and instead add router-specific tests.

- [x] Add tests
- [x] Run tests


yusukebe:
Hi @usualoma !

Awesome work! You did what I have wanted to do for a long time.

I checked and it is probably correct. Merging. I really appreciate you!

PR #2260, opened 2024-02-23
fix(serve-static): support extensionless files and refactor

### LINKED ISSUE #2251: serveStatic doesn't work with extensionless files

4.0.5

Deno

I'm using [Pagefind](https://pagefind.app/) which generates the following file structure:
imaxe

When the browser request the file `/pagefind/pagefind.unknown_278027f53872923.pf_meta` it should be served correctly.

### What do you see instead?

The browser gets a 404 error, and Hono logs

```
Error: Not a directory (os error 20): open './_site/pagefind/pagefind.unknown_278027f53872923.pf_meta/index.html'
```

### Additional information

This is because Hono automatically add `/index.html` to any file without an extension (See the code here: https://github.com/honojs/hono/blob/main/src/utils/filepath.ts#L19-L21)

It should check first if the file exist before adding the `defaultDocument` sufix.

yusukebe:
Fixes #2251 and refactored.

Refactoring methods:

* Created `middleware/serve-static`, which is not used by users directly.
* Wrap the middleware by the environment, such as Deno, Bun, and Cloudflare Workers.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/adapter/deno/serve-static.ts
+++ b/deno_dist/adapter/deno/serve-static.ts
@@ -1,65 +1,32 @@
-import type { Context } from '../../context.ts'
+import type { ServeStaticOptions } from '../../middleware/serve-static/index.ts'
+import { serveStatic as baseServeStatic } from '../../middleware/serve-static/index.ts'
 import type { Env, MiddlewareHandler } from '../../types.ts'
-import { getFilePath } from '../../utils/filepath.ts'
-import { getMimeType } from '../../utils/mime.ts'
 
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore
 const { open } = Deno
 
-export type ServeStaticOptions<E extends Env = Env> = {
-  root?: string
-  path?: string
-  mimes?: Record<string, string>
-  rewriteRequestPath?: (path: string) => string
-  onNotFound?: (path: string, c: Context<E>) => void | Promise<void>
-}
-
-const DEFAULT_DOCUMENT = 'index.html'
-
 export const serveStatic = <E extends Env = Env>(
-  options: ServeStaticOptions<E> = { root: '' }
+  options: ServeStaticOptions<E>
 ): MiddlewareHandler => {
-  return async (c, next) => {
-    // Do nothing if Response is already set
-    if (c.finalized) {
-      await next()
-      return
-    }
-
-    const url = new URL(c.req.url)
-    const filename = options.path ?? decodeURI(url.pathname)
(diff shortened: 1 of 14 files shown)


yusukebe:
@watany-dev @sor4chi 

Could you review this one?

sor4chi:
Nice refactoring. LGTM
Are you planning to make the same changes to `@hono/node-server` after this?

watany-dev:
@yusukebe 

This standardization is wonderful. Rather than a feature, I'm curious about its positioning as middleware. Unlike the current example of 'powered-by', this middleware seems like something users might mistakenly try to configure in the application. Therefore, placing it under 'helper/dev' or a similar naming convention to prevent direct configuration mistakes might be necessary.

yusukebe:
Thanks both!

@sor4chi 
> Are you planning to make the same changes to `@hono/node-server` after this?

`@hono/node-server` uses Stream, no need to use this middleware. But, if it is not handling extensionless files, we will need to fix it.

@watany-dev 
> Rather than a feature, I'm curious about its positioning as middleware.

I think the current implementation, i.e. putting it in `src/middleware/serve-static`, is fine.

For example, [`powered-by` middleware is exported in `package.json`](https://github.com/honojs/hono/blob/e6913190b745f673b7d7fba37bdc1ccbeaf1ef87/package.json#L182-L186), but `serve-staic` is not. So users cannot use `hono/server-static`.

sor4chi:
Okay, thank you!
I too was curious about the positioning of middleware. 
How about making it explicit that it is internal by adding a prefix such as `_`, `internal` or `base`?

yusukebe:
> How about making it explicit that it is internal by adding a prefix such as `_`, `internal` or `base`?

Hmm. I don't think we have to do it.

This middleware is basically a wrapper, but if a user wants to use it, the user can specify `getContent`. So it does not have to be *perfectly* internal. So, we just need to deal with it by not exporting it and not writing it in the document.

yusukebe:
One way is to create `utils/serve-static`, but I don't want to put anything that handles Context in `utils/*`. Besides, the `server-static` implementation is really middleware. So, if we don't want users to use it, the solution is not to export it, as mentioned above. In the future, it may be possible to export and have users specify `getContent()`. Like `toSSG()`.

sor4chi:
I see. Thank you for answering!

yusukebe:
Hey everyone.

I'd like to merge this as soon as possible. So, merge now. If you have ideas to refactor or do something, please create the issue later. Thanks!

PR #2264, opened 2024-02-24
perf(jsx/dom): update textContent only when content changes

usualoma:
When there is no update, assing to textContent is not performed to improve performance.

- [x] Add tests
- [x] Run tests


usualoma:
I added a little refactoring in f625648, but it is done.

yusukebe (approved):
LGTM!

yusukebe:
Indeed, we can improve it with your implementation.

Thanks!

Issue #1153, opened 2023-06-02
WebSockets


hanumanjiblog:
Please implement WebSockets feature in honojs.

https://developers.cloudflare.com/workers/runtime-apis/websockets/

zisra:
Not just cloudflare, but also Node.js and deno

yusukebe:
You can use WebSockets with the current version of Hono on Deno. 

Refer: https://github.com/orgs/honojs/discussions/755 

It can also be supported on Cloudflare Workers. 

More info: https://developers.cloudflare.com/workers/learning/using-websockets/ 

For Node.js, you can't use WebSockets with the current Node.js Adapter and there are no plans to implement it. But, if you want this feature, you might be able to write the PR yourself.

yusukebe:
Hi @JustJoostNL !

I've written an example. Try this!

```ts
import type { WebSocket as CFWebSocket } from '@cloudflare/workers-types'
import { Hono } from 'hono'
import { html } from 'hono/html'

const app = new Hono()

app.get('/', (c) => {
  return c.html(
    html`<html>
      <body>
        <script>
          const hostname = window.location.host
          const protocol = document.location.protocol === 'http:' ? 'ws://' : 'wss://'
          const ws = new WebSocket(protocol + hostname + '/websocket')
          ws.addEventListener('message', ({ data }) => {
            console.log(data)
          })
        </script>
        <h1>WebSocket</h1>
      </body>
    </html>`
  )
})

app.all('/websocket', async (c) => {
  const upgradeHeader = c.req.header('Upgrade')
  if (upgradeHeader !== 'websocket') {
    return c.text('Expected websocket', 400)
  }

  const webSocketPair = new WebSocketPair()
  const client = webSocketPair[0]
  const server = webSocketPair[1] as unknown as CFWebSocket
  server.accept()
  server.send('Hello')

  return new Response(null, {
    status: 101,
    webSocket: client,
  })
})

export default app
```

JustJoostNL:
Thanks!

Kyiro:
It should be fairly trivial to make universal middleware for Workers, Deno and possibly Node. If I have the time, I could look into it as 3rd party middleware but no promise. Bun seems to be out of the question because of how their API works.
Useful links for reference:
[Deno.upgradeWebSocket](https://deno.land/api?s=Deno.upgradeWebSocket) 
[Cloudflare](https://developers.cloudflare.com/workers/runtime-apis/websockets/) 
[...]

PR #2265, opened 2024-02-24
feat: Introduce WebSocket Helper / Adapter

nakasyou:
Reference: https://github.com/honojs/hono/issues/1153

If we marge this PR, we can use WebSocket API through Hono easier.

Example (Bun):
```tsx
// app.tsx

/** @jsx jsx */
/** @jsxImportSource ./hono/src/jsx */
import { Hono } from 'hono'
import { createBunWebSocket } from 'hono/ws/bun'

const { upgradeWebSocket, websocket } = createBunWebSocket()

const app = new Hono()

app.get('/', c => {
  return c.html(
  <html>
    <head>
      <meta charset="UTF-8" />
    </head>
    <body>
      <h1>Hono WebSocket Example</h1>
      <div id="now-time"></div>
      <script dangerouslySetInnerHTML={{
        __html: `
        const ws = new WebSocket('/ws')
        const $nowTime = document.getElementById('now-time')
        ws.onmessage = evt => {
          $nowTime.textContent = evt.data
        }
        `
      }}></script>
    </body>
  </html>)
})

const ws = app.get('/ws', upgradeWebSocket(c => {
  let intervalId: number
  return {
    onOpen(evt, ws) {
      intervalId = setInterval(() => {
        ws.send(new Date().toString())
      }, 200) as unknown as number
    },
    onClose () {
      clearInterval(intervalId)
    }
  }
}))

export type AppType = typeof ws

Bun.serve({
  fetch: app.fetch,
  port: 3030,
  websocket
})
```
Then, it is working!

https://github.com/honojs/hono/assets/79000684/100d4bba-36fd-4c56-a74a-f779d5cb55c9

In Cloudflare workers, it probably work (I didn't test it):
```ts
// Replace
import { createBunWebSocket } from 'hono/ws/bun'
const { upgradeWebSocket, websocket } = createBunWebSocket()
// To
import { upgradeWebSocket } from 'hono/ws/cloudflare-workers'
```

And, Hono client can work on it!
```ts
import type { AppType } from './app'

const client = hc<AppType>('/')
const socket = client.ws.$ws() // socket is just `WebSocket` instance

socket.onmessage = ...
```
Cons for client:
 - This implementing code is not smart, I think good solution it is exists.
 - Not type safe

### Implemented adapters:
- [x] Bun
- [x] Cloudflare Workers
- [ ] Node.js
- [ ] AWS lambda
- [x] Cloudflare Pages
- [x] Deno
- [ ] lambda-edge
- [ ] netlify
- [ ] ~~Vercel~~ (Vercel doesn't support WebSockets)

### ~~New `devDependencies`~~
- ~~[`bun-types`](https://www.npmjs.com/package/bun-types)~~
  - ~~The reason is to make Bun Adapter~~
  - ~~It is only included Types.~~
- [x] Add tests
- [x] Run tests


nakasyou:
@yusukebe How do you think this idea?

nakasyou:
In Cloudflare Workers (Server side), the `open` event is not handled.
When I didn't use Hono, the same problem occurred.
I don't know how to resolve it. I took a wrangler and did a test, so It may work in the production environment.

And, I discovered that Workers Adapter is working in Pages, so we don't need Pages adapter for WebSocket.

nakasyou:
If we need, we can marge now!

yusukebe:
Hi @nakasyou !

This is a super cool feature. I would like to take a moment to review it, so please give me a few minutes!

yusukebe:
Now I'm trying to use it. But we have to hear the opinion from @Kyiro ! What do you think?

yusukebe:
I'm playing it now. Wow, awesome.

https://github.com/honojs/hono/assets/10682/e801ead4-1a4c-452d-997e-d4e95f2de966

yusukebe:
> In Cloudflare Workers (Server side), the `open` event is not handled.

That's right, `open` doesn't work on Cloudflare. It didn't work in the production environment either.

> Not type safe

I don't think it is "not type safe" because the `client` will suggest the paths and `$ws`.

nakasyou:
Hi @yusukebe, thank you for reviewing!

> I'm playing it now. Wow, awesome.
Thank you!

> That's right, open doesn't work on Cloudflare. It didn't work in the production environment either.

I see. I was relieved.

> I don't think it is "not type safe" because the client will suggest the paths and $ws.

Yes, I can also think so. My words were not enough.
I think the message data is not type safe.

If we can make it, I want message validator and sending type.
I've tried to make them for few times, but they didn't go well.

pi0:
Hi there!

It seems good progress almost at the same time we are developing [CrossWS](https://crossws.unjs.io/) for being used across ecosystem. I think it would be a really good opportunity to collaborate and not reinvent same adapters but increase coverage from one place.

CrossWS can be embedded into hono so you can build any API/Client on top. And I am more than open to discuss how we can make it flexbible enough for Hono to be using it. ❤️

src/adapter/bun/websocket.ts
```diff
@@ -0,0 +1,115 @@
...
+    readyState: ws.readyState,
+    url: ws.data.url,
+    protocol: ws.data.protocol,
+    close(code, reason) {
+      ws.close(code, reason)
+    },
+  }
+}
+
+export const createBunWebSocket: CreateBunWebSocket = () => {
```
yusukebe, line 59: I think it does not need `Bun` in the name. So it can be `createWebSocket`.

The user will know if it is for Bun because it is imported from `'hono/ws/bun'`.

yusukebe:
Hi @nakasyou 

> If we can make it, I want message validator and sending type.
> I've tried to make them for few times, but they didn't go well.

I don't think there is any need to Type because the data can be sent by WebSocket since it is not structured data; what do you think?

Screenshot 2024-02-28 at 2 50 33

yusukebe:
Hi @pi0 

Thank you for the comment.

> CrossWS can be embedded into hono so you can build any API/Client on top. And I am more than open to discuss how we can make it flexbible enough for Hono to be using it. ❤️ 

That is certainly true. But the feature is already in place and ready to ship as soon as this PR review is done. For us, it is important to "release quickly. Instead, a long discussion is needed on the use of CrossWS. Because our policy is "not to depend on external libraries".

We cannot collaborate immediately, but we may be able to change the logic of WebSocket to CrossWS in the future. Thanks.

pi0:
I am mainly caring because if the user API is is consistent (not impl but even interface) we can have much better Hono support within UnJS ecosystem in the future and also possibly for Nuxt integration. If we diverge yet another layer (WS), it would be trickier.

Mind you, also the current API design in this PR that makes new callbacks inside `upgradeWebSocket` is not most performant choice. Bun docs for reference: 

image

---

I am not how much you are in hurry to ship this. I think this PR was exactly made the day I publicly announced CrossWS so not sure why is the rush and CrossWS API was hammer and thought for a while. 

Defenitly we can also think of another adapter between Hono<>CrossWS but it would make things one more level slower.

Regardless I trust your final choice and always happy to help building better JS ecosystem together.

yusukebe:
@pi0 

Thanks for your comment.

> we can have much better Hono support within UnJS ecosystem in the future and also possibly for Nuxt integration

One thing I did not write about before is this. This is very interesting, and one of the most important things for Hono is to be able to integrate with other layers of the framework.

>  I think this PR was exactly made the day I publicly announced CrossWS so not sure why is the rush and CrossWS API was hammer and thought for a while.

It is true that this PR was created recently, but the idea of WebSocket support and its implementation has been around for a long time.

> Regardless I trust your final choice and always happy to help building better JS ecosystem together.

Thanks!

src/client/types.ts
```diff
@@ -30,6 +30,12 @@ export type ClientRequest<S extends Schema> = {
         : {}
       : {}
   ) => URL
+} & {
+  $ws: S['$get'] extends { input: { json: '_websocket' | undefined } }
```
yusukebe, line 34: It's better to use the `_websocket` as a constant because it's also used in line 17 in `src/helper/websocket/index.ts`.

And, I like `__websocket` rather than `_websocket` to avoid duplicating user input values as much as possible.

nakasyou:
Hi @yusukebe, thank you for your opinion!
> I don't think there is any need to Type because the data can be sent by WebSocket since it is not structured data; what do you think?

When we use JSON. I feel sending JSON through WebSocket is popular.
If we implement it, I think will be a breakthrough.
But, we can implement it in future, not in this PR.

Hi @pi0, thank you for comment!
> I publicly announced CrossWS so not sure why is the rush and CrossWS API was hammer and thought for a while.

I'll make a supplement about it.

I didn't think about CrossWS for sending this PR, and I created this PR because I want it.
So it's a coincidence, I'm very surprised.

src/adapter/bun/websocket.ts
```diff
@@ -0,0 +1,115 @@
...
+            reason,
+          }),
+          createWSContext(ws)
+        )
+      }
+    },
+    message(ws, message) {
+      const websocketListeners = websocketConns[ws.data.connId]
+      if (websocketListeners.onMessage) {
+        const nomalizedReceiveData =
```
yusukebe, line 101: This is a typo. `nomalizedReceiveData` should be `normalizedReceiveData`.

yusukebe (commented):
Have you thought about exporting functions from `hono/cloudflare-workers` instead of `hono/ws/cloudflare-workers`?

```ts
import { upgradeWebSocket } from 'hono/cloudflare-workers'
```
And can you write a runtime test for Cloudflare and Deno?

yusukebe:
@nakasyou 

> When we use JSON. I feel sending JSON through WebSocket is popular.
> If we implement it, I think will be a breakthrough.
> But, we can implement it in future, not in this PR.
> What do you think?

Ah, I see. But maybe it's super difficult! Both server and client `send` and `onmessage` must be supported. I don't want to make the type definitions too complicated, so I think it is unnecessary to implement this in this PR.

src/adapter/bun/websocket.ts
yusukebe: I'm reconsidering. `BunWebSocket` might be a good name for this since it contains Bun specific objects.

src/adapter/bun/websocket.ts
nakasyou: I fixed it.

nakasyou:
Hi @yusukebe 

> Have you thought about exporting functions from `hono/cloudflare-workers` instead of `hono/ws/cloudflare-workers`? This is the same as the `toSSG` approach.
> 
> ```ts
> import { upgradeWebSocket } from 'hono/cloudflare-workers'
> ```
> 
> And can you write a runtime test for Cloudflare and Deno?

I think that this is good idea, so I changed it.

src/client/types.ts
nakasyou: I implemented it.

yusukebe:
@nakasyou 

> I think that this is good idea, so I changed it.

Okay! When you finish your work, please ping me.

nakasyou:
Hi @yusukebe, finished my work.

And, I want to discuss, I think it may be good API:
```ts
import { upgradeWebSocket } from 'hono/ws'
import { cfWebSocketContext } from 'hono/cloudflare-workers'

app.get('/ws', upgradeWebSocket(cfWebSocketContext, () => ({...}))
```
It is similar to the toSSG adapter.
It maybe more extensible than now API. 
How about this API?

src/adapter/cloudflare-workers/websocket.ts
```diff
@@ -0,0 +1,69 @@
...
+      (evt) => events.onOpen && events.onOpen(evt as Event, wsContext)
+    )
+  }
+  if (events.onClose) {
+    server.addEventListener(
+      'close',
+      (evt) => events.onClose && events.onClose(evt as CloseEvent, wsContext)
+    )
+  }
+  console.log(events.onMessage)
```
yusukebe, line 50: Is this necessary?

yusukebe:
Thanks @nakasyou !

> And, I want to discuss, I think it may be good API...

I think your API is good. But, I want the user to import just one line:

```ts
import { upgradeWebSocket } from 'hono/cloudflare-workers'
```

In the case of your proposed API, it makes sense if the `upgradeWebSocket` exported from `'hono/ws'` has a lot of common processes. However, its implementation varies widely from runtime to runtime, so I think the current implementation where `hono/ws` exports only type `UpgradeWebSocket` is fine.

Of course, you can make that adapter like this:

```ts
// src/adapter/cloudflare-workers/websocket.ts
import { upgradeWebSocket as commonUpgradeWebSocket } from 'hono/ws'
// ...
export const upgradeWebSocket = (arg) => commonUpgradeWebSocket(cfWebSocketContext, arg)
```

That way is not a bad idea.

Eather way, what defines a good API is open to debate, but in the case of an adapter, I think a good API is one that allows you to simply change the import source when the runtime is changed.

```ts
// Cloudflare
import { upgradeWebSocket } from 'hono/cloudflare-workers'
// Another runtime
import { upgradeWebSocket } from 'hono/another'
```

src/adapter/cloudflare-workers/websocket.ts
nakasyou: Oh, sorry. This isn't necessary, this is for debugging. I deleted this.
I'll be careful of things like this.

nakasyou:
Hi @yusukebe, thank you for your comment! 
I was able to rethink my thoughts.
> In the case of your proposed API, it makes sense if the upgradeWebSocket exported from 'hono/ws' has a lot of common processes. However, its implementation varies widely from runtime to runtime, so I think the current implementation where hono/ws exports only type UpgradeWebSocket is fine.

I see. I read it, and I thought so too. As you said, them implementation is different for each runtime. Also we can easily switch the runtime by switching import.

So I think that the current API is better.

yusukebe:
Hi @nakasyou 

Thanks for understanding.

The remaining thing. What do you think about this? https://github.com/honojs/hono/pull/2265#pullrequestreview-1904981060

nakasyou:
Hi @yusukebe 
> The remaining thing. What do you think about this?
> https://github.com/honojs/hono/pull/2265#pullrequestreview-1904981060

When we use `toSSG` on adapter, the adapter is for only accessing file system.
One side, the WebSocket adapter is exported `upgradeWebSocket` function. It's all about the function.

If we implement like https://github.com/honojs/hono/pull/2265#issuecomment-1974190405, `import { upgradeWebSocket } from 'hono/cloudflare-workers` is good because adapter exports only core features.

But we will not implement like this.
As mentioned above, the WebSocket adapter is exported `upgradeWebSocket` function. It's all about the function. So `hono/cloudflare-workers/ws` or `hono/ws/cloudflare-workers` is good.

This approach is similar of `serveStatic` adapter.

nakasyou:
Hi @yusukebe 
As I pointed out, I had various misunderstandings.
So I think `import { upgradeWebSocket } from 'hono/cloudflare-workers'` is good.

> So, I think the following . /ws/cloudflare-workers etc. are not necessary?
Yes, I'll remove it.

nakasyou:
Hi @yusukebe, I removed how you can import from ws adapter such as `hono/ws/cloudflare-workers`.

yusukebe (approved):
LGTM!

yusukebe:
Hi @nakasyou !

Thanks! Finished! Let's land it.

lilnasy:
IMO the shipped API is way nicer than what crossws offers!

It is trivial to implement multiple websocket endpoints that behave differently, whereas crossws (and bun for that matter) would have you doing workarounds or require another library to multiplex it.

I love it, thanks for implementing this! @nakasyou :yellow_heart:

PR #2268, opened 2024-02-24
fix(jsx/dom): import from correct file for avoiding circular dependency

usualoma:
Fixes https://github.com/honojs/honox/issues/92

- [x] Run tests


yusukebe (approved):
Works well! Thanks!

Issue #1203, opened 2023-06-26
Add SetCookie prefixing feature for `__Secure-` and `__Host-`


Manouchehri:
https://www.sjoerdlangkemper.nl/2017/02/09/cookie-prefixes/

So for `__Secure-`, the usage would look like:

```typescript
setCookie(c, 'example', 'secret_example', {
  path: '/api/demo',
  httpOnly: true,
  sameSite: 'Strict',
  securePrefix: true,
  secure: true // if omitted, set to true. If set to false, throw an error
})
```

This would result in the cookie name actually being set to `__Secure-example`.

```typescript
const incoming_cookie = getCookie(c, 'example', {
  securePrefix: true
})
```

And for `__Host-`, the usage would look like:

```typescript
setCookie(c, 'example', 'secret_example', {
  path: '/', // if omitted, set to /. If set to anything other than /, then throw an error.
  httpOnly: true,
  sameSite: 'Strict',
  hostPrefix: true,
  secure: true // if omitted, set to true. If set to false, throw an error
  // If domain is set, throw an error
})
```

```typescript
const incoming_cookie = getCookie(c, 'example', {
  hostPrefix: true
})
```

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#cookie_prefixes

yusukebe:
Hi @Manouchehri 

Ah, this is a good feature! Will you implement this?

Manouchehri:
Not in the immediate future, wasn't 100% sure if you would agree with the error handling/throws. 😄

yusukebe:
This was closed by #2269. Thank you.

PR #2269, opened 2024-02-25
feat(cookie): add secure and host prefix support

Datron:
Hello @yusukebe!

I have added support for secure and host prefixes. This PR resolves https://github.com/honojs/hono/issues/1203

- [x] Add tests
- [x] Run tests

cc: @Code-Hex

### DIFF
--- a/src/helper/cookie/index.test.ts
+++ b/src/helper/cookie/index.test.ts
@@ -195,6 +195,123 @@ describe('Cookie Middleware', () => {
       )
     })
 
+    app.get('/get-secure-prefix-cookie', async (c) => {
+      const cookie = getCookie(c, 'delicious_cookie', 'secure')
+      if (cookie) {
+        return c.text(cookie)
+      } else {
+        return c.notFound()
+      }
+    })
+
+    app.get('/get-host-prefix-cookie', async (c) => {
+      const cookie = getCookie(c, 'delicious_cookie', 'host')
+      if (cookie) {
+        return c.text(cookie)
+      } else {
+        return c.notFound()
+      }
+    })
+
+    app.get('/set-secure-prefix-cookie', (c) => {
+      setCookie(c, 'delicious_cookie', 'macha', {
+        prefix: 'secure',
+        secure: false, // this will be ignore
+      })
+      return c.text('Set secure prefix cookie')
+    })
+
+    it('Set cookie with secure prefix', async () => {
+      const res = await app.request('http://localhost/set-secure-prefix-cookie')
+      expect(res.status).toBe(200)
+      const header = res.headers.get('Set-Cookie')
+      expect(header).toBe('__Secure-delicious_cookie=macha; Path=/; Secure')
+    })
+
+    it('Get cookie with secure prefix', async () => {
+      const setCookie = await app.request('http://localhost/set-secure-prefix-cookie')
+      const header = setCookie.headers.get('Set-Cookie')
+      if (!header) {
+        assert.fail('invalid header')
+      }
+      const res = await app.request('http://localhost/get-secure-prefix-cookie', {
+        headers: {
+          Cookie: header,
+        },
+      })
+      const response = await res.text()
+      expect(res.status).toBe(200)
+      expect(response).toBe('macha')
+    })
+
+    app.get('/set-host-prefix-cookie', (c) => {
+      setCookie(c, 'delicious_cookie', 'macha', {
+        prefix: 'host',
+        path: '/foo', // this will be ignored
+        domain: 'example.com', // this will be ignored
+        secure: false, // this will be ignored
+      })
+      return c.text('Set host prefix cookie')
+    })
+
+    it('Set cookie with host prefix', async () => {
+      const res = await app.request('http://localhost/set-host-prefix-cookie')
+      expect(res.status).toBe(200)
+      const header = res.headers.get('Set-Cookie')
+      expect(header).toBe('__Host-delicious_cookie=macha; Path=/; Secure')
+    })
+
+    it('Get cookie with host prefix', async () => {
+      const setCookie = await app.request('http://localhost/set-host-prefix-cookie')
+      const header = setCookie.headers.get('Set-Cookie')
+      if (!header) {
+        assert.fail('invalid header')
+      }
+      const res = await app.request('http://localhost/get-host-prefix-cookie', {
+        headers: {
+          Cookie: header,
+        },
+      })
+      const response = await res.text()
+      expect(res.status).toBe(200)
+      expect(response).toBe('macha')
+    })
+
+    app.get('/set-signed-secure-prefix-cookie', async (c) => {
+      await setSignedCookie(c, 'delicious_cookie', 'macha', 'secret choco chips', {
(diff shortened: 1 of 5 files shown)


yusukebe (approved):
Looks good to me!

yusukebe:
Hey @Code-Hex !

Could you also review this one?

src/helper/cookie/index.ts
```diff
@@ -47,8 +71,27 @@ export const getSignedCookie: GetSignedCookie = async (c, secret, key?) => {
...
 export const setCookie = (c: Context, name: string, value: string, opt?: CookieOptions): void => {
-  const cookie = serialize(name, value, { path: '/', ...opt })
-  c.header('set-cookie', cookie, { append: true })
+  // Cookie names prefixed with __Secure- can be used only if they are set with the secure attribute.
+  // Cookie names prefixed with __Host- can be used only if they are set with the secure attribute, must have a path of / (meaning any path at the host)
+  // and must not have a Domain attribute.
+  // Read more at https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#cookie_prefixes'
+  const cookies = []
+  if (opt?.securePrefix) {
+    cookies.push(serialize('__Secure-' + name, value, { path: '/', secure: true, ...opt }))
```
Code-Hex, line 80: Do not change the secure attribute.

```suggestion
    cookies.push(serialize('__Secure-' + name, value, { path: '/', ...opt, secure: true  }))
```

src/helper/cookie/index.ts
```diff
@@ -47,8 +71,27 @@ export const getSignedCookie: GetSignedCookie = async (c, secret, key?) => {
...
+  }
+
+  if (opt?.hostPrefix) {
+    cookies.push(
+      serialize('__Host-' + name, value, { secure: true, path: '/', domain: undefined, ...opt })
```
Code-Hex, line 85: Do not change the secure, path, and domain attributes.
```suggestion
      serialize('__Host-' + name, value, { ...opt, secure: true, path: '/', domain: undefined })
```

src/utils/cookie.ts
```diff
@@ -12,6 +12,12 @@ export type CookieOptions = {
   signingSecret?: string
   sameSite?: 'Strict' | 'Lax' | 'None'
   partitioned?: boolean
+  securePrefix?: boolean
+  hostPrefix?: boolean
+}
+export type CookiePrefixOptions = {
+  secure?: boolean
+  host?: boolean
```
Code-Hex, line 20: I don't think there are cases where you want to use two prefixes, so I think it's better to design it so that you choose one or the other. That way, it's easier to understand that only one prefix is being used.

```suggestion
  prefix?: 'host' | 'secure'
}
export type CookiePrefixOptions = {
  prefix: 'host' | 'secure'
```

src/helper/cookie/index.test.ts
```diff
@@ -195,6 +195,118 @@ describe('Cookie Middleware', () => {
...
+      })
+      const response = await res.text()
+      expect(res.status).toBe(200)
+      expect(response).toBe('macha')
+    })
+
+    app.get('/set-host-prefix-cookie', (c) => {
+      setCookie(c, 'delicious_cookie', 'macha', {
+        hostPrefix: true,
+        domain: 'example.com', // this will be ignored
```
Code-Hex, line 249:
```suggestion
        domain: 'example.com', // this will be ignored
        path: '/foo', // this will be ignored
        secure: false, // this will be ignored
```

Code-Hex:
@Datron I added a few comments but almost looks great to me!

src/helper/cookie/index.ts
Datron: @Code-Hex what if a developer gives false as a value for `secure` with `prefix` value set to 'Secure'? This would be invalid. I think it would be better to override it in this case.

src/helper/cookie/index.test.ts
Datron: Sure!

Datron:
@Code-Hex I've resolved your comments.

src/helper/cookie/index.test.ts
```diff
@@ -195,6 +195,121 @@ describe('Cookie Middleware', () => {
...
+      expect(header).toBe(
+        '__Secure-delicious_cookie=macha.i225faTyCrJUY8TvpTuJHI20HBWbQ89B4GV7lT4E%2FB0%3D; Path=/; Secure'
+      )
+    })
+
+    app.get('/set-signed-host-prefix-cookie', async (c) => {
+      await setSignedCookie(c, 'delicious_cookie', 'macha', 'secret choco chips', {
+        prefix: 'host',
+        domain: 'example.com', // this will be ignored
+      })
```
Code-Hex, line 300: I'm sorry this is the last thing!
```suggestion
      await setSignedCookie(c, 'delicious_cookie', 'macha', 'secret choco chips', {
        prefix: 'host',
        path: '/foo', // this will be ignored
        domain: 'example.com', // this will be ignored
        secure: false, // this will be ignored
      })
```

Code-Hex (approved):
LGTM

Code-Hex:
@yusukebe Could you check one more time?

src/helper/cookie/index.ts
Code-Hex: @Datron Considering the specifications, when using a prefix, it's always necessary to set `secure` to true. Therefore, I believe the `secure: false` option should be ignored.

Code-Hex: This is implicit, so how about providing two option types to accept either one? What do you think, @yusukebe ?

like below:
```ts
type CookieOptions = {
  domain?: string
  expires?: Date
  httpOnly?: boolean
  maxAge?: number
  path?: string
  secure?: boolean
  signingSecret?: string
  sameSite?: 'Strict' | 'Lax' | 'None'
  partitioned?: boolean
} | {
  prefix: 'host',
  expires?: Date
  httpOnly?: boolean
  maxAge?: number
  signingSecret?: string
  sameSite?: 'Strict' | 'Lax' | 'None'
} | {
  prefix: 'secure',
  expires?: Date
  httpOnly?: boolean
  maxAge?: number
  path?: string
  signingSecret?: string
  sameSite?: 'Strict' | 'Lax' | 'None'
  domain?: string
}
```

Datron: @Code-Hex perhaps we can keep the implicit behaviour and document this on hono.dev?

src/utils/cookie.ts
yusukebe: Ah, this is a good suggestion. Thanks @Code-Hex !

src/helper/cookie/index.ts
yusukebe: I think we don't need to go so far and it is possible that the IDE will not do a good job of Types suggestion. Let's leave it as is.

yusukebe:
Looks good again!

Thanks @Datron and @Code-Hex !

I'll merge this into the "next" branch for `v4.1.0`. I'll release it maybe soon (when the WebSocket feature is merged!).

Issue #2068, opened 2024-01-23
Handler stuck when awaiting aborted SSE stream


jonasb:
### What version of Hono are you using?

3.12.6

### What runtime/platform is your app running on?

Node

### What steps can reproduce the bug?

Start the following server and run this command `curl http://localhost:8000/sse`. After a few seconds stop the curl command.

```js
import { serve } from "@hono/node-server";
import { Hono } from "hono";
import { streamSSE } from "hono/streaming";

const app = new Hono();

let testId = 0;

app.get("/sse", (c) =>
  streamSSE(c, async (stream) => {
    try {
      const currentTestId = ++testId;
      let aborted = false;
      stream.onAbort(() => {
        aborted = true;
      });

      for (let i = 0; i < 10; i++) {
        console.log(currentTestId, "Iteration", i, aborted ? "(aborted)" : "");
        // If I await the writeSSE like this, it sometimes never returns when the stream is aborted
        await stream.writeSSE({ data: JSON.stringify({ iteration: i }) });
        await stream.sleep(1000);
      }
      console.log("Done");
    } catch (error) {
      console.error("Error", error);
    }
  })
);

const port = 8000;
console.log(`Server is running on port ${port}`);

serve({
  fetch: app.fetch,
  port,
});

```

### What is the expected behavior?

I expect either the `await stream.writeSSE()` to "succeed" or to throw an exception. Now the call never completes after a for writes after the stream is aborted.

### What do you see instead?

The server logging stops after a few writes and the handler never finishes.

Example output:
```
1 Iteration 0 
1 Iteration 1 
1 Iteration 2 
1 Iteration 3 
1 Iteration 4 (aborted)
1 Iteration 5 (aborted)
```
(it doesn't complete all the iterations and Done is not logged)

### Additional information

For my use case I want to continue with all iterations even if the stream is closed. I think I have two workarounds:

1) Skip awaiting `stream.writeSSE()`
2) Skip calling `stream.writeSSE()` when the stream is aborted

Not sure if there are any drawbacks of not awaiting the writes (1). And I'm a bit worried that handler can be stuck if there are issues with writing.

yusukebe:
@watany-dev or @sor4chi Can you see this?

watany-dev:
[...]

PR #2273, opened 2024-02-25
Handler stuck when awaiting aborted SSE stream

abhi12299:
Fixes: #2068
In the current version of Hono, the `stream.writeSSE` method was getting blocked indefinitely when the user disconnects from the SSE stream prematurely. Consider this example (taken from #2068):

```typescript
import { serve } from "@hono/node-server";
import { Hono } from "hono";

const app = new Hono();

let testId = 0;

app.get("/sse", (c) =>
  streamSSE(c, async (stream) => {
    try {
      const currentTestId = ++testId;
      let aborted = false;
      stream.onAbort(() => {
        aborted = true;
      });

        console.log(currentTestId, "Iteration", i, aborted ? "(aborted)" : "");
        // If I await the writeSSE like this, it sometimes never returns when the stream is aborted
        await stream.writeSSE({ data: JSON.stringify({ iteration: i }) });
        await stream.sleep(1000);
      }
      console.log("Done");
    } catch (error) {
      console.error("Error", error);
    }
  })
);

const port = 8000;
console.log(`Server is running on port ${port}`);

serve({
  fetch: app.fetch,
  port,
});
```

When the client disconnects from the stream while the loop is still going on, the following things happen:
In the file `src/utils/stream.ts`:
```typescript
this.responseReadable = new ReadableStream({
  async pull(controller) {
    const { done, value } = await reader.read()
    done ? controller.close() : controller.enqueue(value)
  },
  cancel: () => {
    this.abortSubscribers.forEach((subscriber) => subscriber())
  },
})
```

`cancel` function runs, thereby running all the abort subscribers, but due to the loop in the test code provided above, the `pull` function still receives `done` as `false` and some value from the `reader`, and gets blocked indefinitely because the `controller` enqueued it.

It is difficult to reproduce this behaviour in the test suites, because aborting the request from code works fine. This is not the case when using a client such as `curl` or postman.

### DIFF
--- a/deno_dist/utils/stream.ts
+++ b/deno_dist/utils/stream.ts
@@ -12,6 +12,13 @@ export class StreamingApi {
 
     const reader = _readable.getReader()
 
+    // in case the user disconnects, let the reader know to cancel
+    // this in-turn results in responseReadable being closed
+    // and writeSSE method no longer blocks indefinitely
+    this.abortSubscribers.push(async () => {
+      await reader.cancel()
+    })
+
     this.responseReadable = new ReadableStream({
       async pull(controller) {
         const { done, value } = await reader.read()
--- a/src/utils/stream.ts
+++ b/src/utils/stream.ts
@@ -12,6 +12,13 @@ export class StreamingApi {
 
     const reader = _readable.getReader()
 
+    // in case the user disconnects, let the reader know to cancel
+    // this in-turn results in responseReadable being closed
+    // and writeSSE method no longer blocks indefinitely
+    this.abortSubscribers.push(async () => {
+      await reader.cancel()
+    })
+
     this.responseReadable = new ReadableStream({
       async pull(controller) {
         const { done, value } = await reader.read()


yusukebe:
Hi @abhi12299 !

Thanks. Seems to be good. However, I want to hear the options from @watany-dev and @sor4chi .

abhi12299:
Adding a suggestion here: we should also add a `connected` property on the `stream` object to let the user know if the client is still connected. As of now, there's no way to do that (as per my understanding).

watany-dev:
Thanks. This is the event I saw when I reproduced some issue. Glad we were able to resolve it.

I think it is better to discuss stream.connected in another PR. As far as the use case is concerned, connected may be sufficient to be implemented only in streamSSE.

sor4chi:
Thank you for this feature, I never thought of blocking readable. I agree.

yusukebe:
@abhi12299 @watany-dev @sor4chi Thanks!

We will discuss `connected` in other PRs, but this PR will be merged!

PR #2275, opened 2024-02-25
fix(ssg): preserve binary files as-is

### LINKED ISSUE #2274: SSG: Binary files corrupted when using SSG helper

4.0.7

Node, Bun

Just `return c.body(/* some binary, like a png */)` and `toSSG`

see. https://github.com/berlysia/repro-hono-ssg-binary-broken

The generated files are identical at the binary level.

### What do you see instead?

Files work as expected on development server and pages, but corrupt in SSG output.

### Additional information

I think this is runtime agnostic.

On the routes which respond binary, `bufferToString` is always applied. This means all binaries are decoded in UTF-8.
see. https://github.com/honojs/hono/blob/a32482ddb0d8c9b75310c28750389a1819be6618/src/helper/ssg/index.ts#L204-L205

berlysia:
resolve #2274

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/helper/ssg/index.test.tsx
+++ b/src/helper/ssg/index.test.tsx
@@ -1,4 +1,4 @@
-import { beforeEach, describe, expect, it } from 'vitest'
+import { Mocked, beforeEach, describe, expect, it } from 'vitest'
 import { Hono } from '../../hono'
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 import { jsx } from '../../jsx'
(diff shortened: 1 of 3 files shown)


src/helper/ssg/index.test.tsx
```diff
@@ -336,6 +366,30 @@ describe('saveContentToFiles function', () => {
     expect(fsMock.writeFile).toHaveBeenCalledWith('static/bravo.text/index.html', 'Bravo Page')
   })
 
+  describe('should correctly create binary files with the right content', () => {
+    it('for application/gzip', async () => {
+      await saveContentToFiles(htmlMap, fsMock, './static')
+
+      expect(fsMock.writeFile).toHaveBeenCalledWith('static/bravo/index.tar.gz', expect.anything())
```
berlysia, line 373: Perhaps this line is useless

yusukebe: Does that mean that there is a following test?

https://github.com/honojs/hono/blob/f6957ee1b2b83588a9ba472a91df1f48ebcf1480/src/helper/ssg/index.test.tsx#L374-L378

I think we can leave this line in, since the other files are tested with `toHaveBeenCalledWith` in the same way.

yusukebe:
Hi @berlysia !

Nice fix! I've commented. Check it and run `bun run lint:fix` to fix the lint error.

src/helper/ssg/index.test.tsx
berlysia: OK, I've found that the followings is more unnecessary. This is the remains of my investigation

Removed in 7b17acbbc0be407621af4aaa340f3fdd25fd5dfd

berlysia:
I've made the updates and re-ran the three checks + lint. I think it's become simpler.

src/helper/ssg/index.test.tsx
berlysia: hmm... we can merge again into `"should correctly create files with the right content and paths"` test

yusukebe (approved):
Looks good! It's neat!

PR #2277, opened 2024-02-26
fix(devDependencies): Fix the problem of missing glob package

xx1124961758:
Resolved the issue of the glob package being unlocated due to its phantom dependency status during dependency installation via pnpm.

- [ ] Add tests
- [ ] Run tests


yusukebe:
Hi @xx1124961758 

Sorry, I don't understand what exactly is the problem. We recommend using bun or yarn instead of pnpm for this project. So, the pnpm issue is out of scope.

xx1124961758:
> Hi @xx1124961758
> 
> Sorry, I don't understand what exactly is the problem. We recommend using bun or yarn instead of pnpm for the development. So, the pnpm issue is out of scope.

In the project, the glob package was used but was not explicitly declared in the package.json file.

yusukebe:
@xx1124961758 

Thanks, I see. Could you run `bun install`?

xx1124961758:
> @xx1124961758
> 
> Thanks, I see. Could you run `bun install`?

yes, it's ok with 'bun install'

yusukebe:
Hi @xx1124961758 

It wasn't working, so I fixed it myself. Thanks!

PR #2278, opened 2024-02-26
fix: fix comment

nakasyou:
`c.jsonT` is removed now, but there is a reference to it in the comment.
I found it when I was reading the code, it may cause confusion for contributors, so I fixed it.

- [ ] Add tests
- [x] Run tests


yusukebe:
You are right! Thanks @nakasyou !

PR #2283, opened 2024-02-28
fix(deno/jsx-precompile): use html() tag function if value is a promise

usualoma:

- [x] Add tests
- [x] Run tests


yusukebe (approved):
Good fix! LGTM!

PR #2285, opened 2024-02-28
Optimize RegExpRouter and maybe more

aquapi:
I'm not done yet.

Gonna do more optimizations to the build step as well.

### DIFF
--- a/src/router/reg-exp-router/node.ts
+++ b/src/router/reg-exp-router/node.ts
@@ -53,7 +53,7 @@ export class Node {
     pathErrorCheckOnly: boolean
   ): void {
     if (tokens.length === 0) {
-      if (this.index !== undefined) {
+      if (typeof this.index !== 'undefined') {
         throw PATH_ERROR
       }
       if (pathErrorCheckOnly) {
@@ -87,17 +87,17 @@ export class Node {
       }
 
       node = this.children[regexpStr]
-      if (!node) {
-        if (
-          Object.keys(this.children).some(
-            (k) => k !== ONLY_WILDCARD_REG_EXP_STR && k !== TAIL_WILDCARD_REG_EXP_STR
-          )
-        ) {
-          throw PATH_ERROR
+      if (typeof node === 'undefined') {
+        for (const k in this.children) {
+          if (k !== ONLY_WILDCARD_REG_EXP_STR && k !== TAIL_WILDCARD_REG_EXP_STR) {
+            throw PATH_ERROR
+          }
         }
+        
         if (pathErrorCheckOnly) {
           return
         }
+        
         node = this.children[regexpStr] = new Node()
         if (name !== '') {
           node.varIndex = context.varIndex++
@@ -108,15 +108,13 @@ export class Node {
       }
     } else {
       node = this.children[token]
-      if (!node) {
-        if (
-          Object.keys(this.children).some(
-            (k) =>
-              k.length > 1 && k !== ONLY_WILDCARD_REG_EXP_STR && k !== TAIL_WILDCARD_REG_EXP_STR
-          )
-        ) {
-          throw PATH_ERROR
+      if (typeof node === 'undefined') {
+        for (const k in this.children) {
+          if (k.length > 1 && k !== ONLY_WILDCARD_REG_EXP_STR && k !== TAIL_WILDCARD_REG_EXP_STR) {
+            throw PATH_ERROR
+          }
         }
+        
         if (pathErrorCheckOnly) {
           return
         }
@@ -146,6 +144,6 @@ export class Node {
       return strList[0]
     }
 
-    return '(?:' + strList.join('|') + ')'
+    return `(?:${strList.join('|')})`
   }
 }
--- a/src/router/reg-exp-router/router.ts
+++ b/src/router/reg-exp-router/router.ts
@@ -200,7 +200,7 @@ export class RegExpRouter<T> implements Router<T> {
               findMiddleware(middleware[METHOD_NAME_ALL], path) ??
               []),
           ]
-          routes[currentMethod][path].push([handler, paramCount - len + i + 1])
+          routes[currentMethod][path].push([handler, paramCount - length + i + 1])
         }
       }
     }
--- a/src/router/reg-exp-router/router.ts
+++ b/src/router/reg-exp-router/router.ts
@@ -104,17 +104,12 @@ function findMiddleware<T>(
   middleware: Record<string, T[]> | undefined,
   path: string
 ): T[] | undefined {
-  if (!middleware) {
-    return undefined
-  }
-
-  for (const k of Object.keys(middleware).sort((a, b) => b.length - a.length)) {
-    if (buildWildcardRegExp(k).test(path)) {
-      return [...middleware[k]]
+  if (typeof middleware !== 'undefined')
+    for (const k of Object.keys(middleware).sort((a, b) => b.length - a.length)) {
+      if (buildWildcardRegExp(k).test(path)) {
+        return [...middleware[k]]
+      }
     }
-  }
-
-  return undefined
 }
 
 export class RegExpRouter<T> implements Router<T> {
@@ -209,12 +204,12 @@ export class RegExpRouter<T> implements Router<T> {
       const matcher = (matchers[method] || matchers[METHOD_NAME_ALL]) as Matcher<T>
 
       const staticMatch = matcher[2][path]
-      if (staticMatch) {
+      if (typeof staticMatch !== 'undefined') {
         return staticMatch
       }
 
       const match = path.match(matcher[0])
-      if (!match) {
+      if (match === null) {
         return [[], emptyParam]
       }
 
@@ -222,7 +217,7 @@ export class RegExpRouter<T> implements Router<T> {
       return [matcher[1][index], match]
     }
 
-    return this.match(method, path)
+    return this.match(method, path);
   }
 
   private buildAllMatchers(): Record<string, Matcher<T> | null> {
--- a/src/router/reg-exp-router/router.ts
+++ b/src/router/reg-exp-router/router.ts
@@ -47,7 +47,7 @@ function buildMatcherFromPreprocessedRoutes<T>(
     )
 
   const staticMap: StaticMap<T> = {}
-  for (let i = 0, j = -1, len = routesWithStaticPathFlag.length; i < len; i++) {
+  for (let i = 0, j = -1, { length } = routesWithStaticPathFlag; i < length; ++i) {
     const [pathErrorCheckOnly, path, handlers] = routesWithStaticPathFlag[i]
     if (pathErrorCheckOnly) {
       staticMap[path] = [handlers.map(([h]) => [h, {}]), emptyParam]
@@ -78,14 +78,15 @@ function buildMatcherFromPreprocessedRoutes<T>(
   }
 
   const [regexp, indexReplacementMap, paramReplacementMap] = trie.buildRegExp()
-  for (let i = 0, len = handlerData.length; i < len; i++) {
-    for (let j = 0, len = handlerData[i].length; j < len; j++) {
+  for (let i = 0, { length } = handlerData; i < length; ++i) {
+    for (let j = 0, { length } = handlerData[i]; j < length; ++j) {
       const map = handlerData[i][j]?.[1]
       if (!map) {
         continue
       }
-      const keys = Object.keys(map)
-      for (let k = 0, len = keys.length; k < len; k++) {
+      
+      const keys = Object.keys(map);
+      for (let k = 0, { length } = keys; k < length; ++k) {
         map[keys[k]] = paramReplacementMap[map[keys[k]]]
       }
     }
@@ -104,12 +105,13 @@ function findMiddleware<T>(
   middleware: Record<string, T[]> | undefined,
   path: string
 ): T[] | undefined {
-  if (typeof middleware !== 'undefined')
+  if (typeof middleware !== 'undefined') {
     for (const k of Object.keys(middleware).sort((a, b) => b.length - a.length)) {
       if (buildWildcardRegExp(k).test(path)) {
         return [...middleware[k]]
       }
     }
+  }
 }
 
 export class RegExpRouter<T> implements Router<T> {
@@ -201,7 +203,7 @@ export class RegExpRouter<T> implements Router<T> {
     const matchers = this.buildAllMatchers()
 
     this.match = (method, path) => {
-      const matcher = (matchers[method] || matchers[METHOD_NAME_ALL]) as Matcher<T>
+      const matcher = (matchers[method] ?? matchers[METHOD_NAME_ALL]) as Matcher<T>
 
       const staticMatch = matcher[2][path]
       if (typeof staticMatch !== 'undefined') {
@@ -212,9 +214,8 @@ export class RegExpRouter<T> implements Router<T> {
       if (match === null) {
         return [[], emptyParam]
       }
-
-      const index = match.indexOf('', 1)
-      return [matcher[1][index], match]
+      
+      return [matcher[1][match.indexOf('', 1)], match]
     }
 
     return this.match(method, path);
--- a/src/router/reg-exp-router/router.ts
+++ b/src/router/reg-exp-router/router.ts
@@ -149,34 +149,42 @@ export class RegExpRouter<T> implements Router<T> {
     if (/\*$/.test(path)) {
       const re = buildWildcardRegExp(path)
       if (method === METHOD_NAME_ALL) {
-        Object.keys(middleware).forEach((m) => {
+        for (const m in middleware) {
           middleware[m][path] ??=
             findMiddleware(middleware[m], path) ??
             findMiddleware(middleware[METHOD_NAME_ALL], path) ??
             []
-        })
+        }
       } else {
         middleware[method][path] ??=
           findMiddleware(middleware[method], path) ??
           findMiddleware(middleware[METHOD_NAME_ALL], path) ??
           []
       }
-      
-      Object.keys(middleware).forEach((m) => {
-        if (method === METHOD_NAME_ALL || method === m) {
-          Object.keys(middleware[m]).forEach((p) => {
-            re.test(p) && middleware[m][p].push([handler, paramCount])
-          })
-        }
-      })
 
-      Object.keys(routes).forEach((m) => {
+      for (const m in middleware) {
         if (method === METHOD_NAME_ALL || method === m) {
-          Object.keys(routes[m]).forEach(
-            (p) => re.test(p) && routes[m][p].push([handler, paramCount])
-          )
+          const currentMiddleware = middleware[m];
+
+          for (const p in currentMiddleware) {
+            if (re.test(p)) {
+              currentMiddleware[p].push([handler, paramCount]);
+            }
+          }
         }
-      })
+      }
+
+      for (const m in routes) {
+        if (method === METHOD_NAME_ALL || method === m) {
+          const currentRoutes = routes[m];
+
+          for (const p in currentRoutes) {
+            if (re.test(p)) {
+              currentRoutes[p].push([handler, paramCount]);
+            }
+          }
+        }
+      }
 
       return
     }
--- a/src/router/reg-exp-router/router.ts
+++ b/src/router/reg-exp-router/router.ts
@@ -81,7 +81,7 @@ function buildMatcherFromPreprocessedRoutes<T>(
   for (let i = 0, { length } = handlerData; i < length; ++i) {
     for (let j = 0, { length } = handlerData[i]; j < length; ++j) {
       const map = handlerData[i][j]?.[1]
-      if (!map) {
+      if (typeof map === 'undefined') {
         continue
       }
       
@@ -127,11 +127,11 @@ export class RegExpRouter<T> implements Router<T> {
   add(method: string, path: string, handler: T) {
     const { middleware, routes } = this
 
-    if (!middleware || !routes) {
+    if (typeof middleware === 'undefined' || typeof routes === 'undefined') {
       throw new Error(MESSAGE_MATCHER_IS_ALREADY_BUILT)
     }
 
-    if (!middleware[method]) {
+    if (typeof middleware[method] === 'undefined') {
       ;[middleware, routes].forEach((handlerMap) => {
         handlerMap[method] = {}
         Object.keys(handlerMap[METHOD_NAME_ALL]).forEach((p) => {
@@ -144,23 +144,24 @@ export class RegExpRouter<T> implements Router<T> {
       path = '*'
     }
 
-    const paramCount = (path.match(/\/:/g) || []).length
+    const paramCount = (path.match(/\/:/g) ?? []).length
 
     if (/\*$/.test(path)) {
       const re = buildWildcardRegExp(path)
       if (method === METHOD_NAME_ALL) {
         Object.keys(middleware).forEach((m) => {
-          middleware[m][path] ||=
-            findMiddleware(middleware[m], path) ||
-            findMiddleware(middleware[METHOD_NAME_ALL], path) ||
+          middleware[m][path] ??=
+            findMiddleware(middleware[m], path) ??
+            findMiddleware(middleware[METHOD_NAME_ALL], path) ??
             []
         })
       } else {
-        middleware[method][path] ||=
-          findMiddleware(middleware[method], path) ||
-          findMiddleware(middleware[METHOD_NAME_ALL], path) ||
+        middleware[method][path] ??=
+          findMiddleware(middleware[method], path) ??
+          findMiddleware(middleware[METHOD_NAME_ALL], path) ??
           []
       }
+      
       Object.keys(middleware).forEach((m) => {
         if (method === METHOD_NAME_ALL || method === m) {
           Object.keys(middleware[m]).forEach((p) => {
@@ -180,27 +181,27 @@ export class RegExpRouter<T> implements Router<T> {
       return
     }
 
-    const paths = checkOptionalParameter(path) || [path]
-    for (let i = 0, len = paths.length; i < len; i++) {
+    const paths = checkOptionalParameter(path) ?? [path]
+    for (let i = 0, { length } = paths; i < length; i++) {
       const path = paths[i]
 
-      Object.keys(routes).forEach((m) => {
-        if (method === METHOD_NAME_ALL || method === m) {
-          routes[m][path] ||= [
-            ...(findMiddleware(middleware[m], path) ||
-              findMiddleware(middleware[METHOD_NAME_ALL], path) ||
+      for (const currentMethod in routes) {
+        if (method === METHOD_NAME_ALL || method === currentMethod) {
+          routes[currentMethod][path] ??= [
+            ...(findMiddleware(middleware[currentMethod], path) ??
+              findMiddleware(middleware[METHOD_NAME_ALL], path) ??
               []),
           ]
-          routes[m][path].push([handler, paramCount - len + i + 1])
+          routes[currentMethod][path].push([handler, paramCount - len + i + 1])
         }
-      })
+      }
     }
   }
 
   match(method: string, path: string): Result<T> {
     clearWildcardRegExpCache() // no longer used.
 
-    const matchers = this.buildAllMatchers()
+    const matchers = this.buildAllMatchers();
 
     this.match = (method, path) => {
       const matcher = (matchers[method] ?? matchers[METHOD_NAME_ALL]) as Matcher<T>
@@ -224,9 +225,12 @@ export class RegExpRouter<T> implements Router<T> {
   private buildAllMatchers(): Record<string, Matcher<T> | null> {
     const matchers: Record<string, Matcher<T> | null> = {}
 
-    ;[...Object.keys(this.routes!), ...Object.keys(this.middleware!)].forEach((method) => {
-      matchers[method] ||= this.buildMatcher(method)
-    })
+    for (const method in this.routes!) {
+      matchers[method] ??= this.buildMatcher(method)
+    }
+    for (const method in this.middleware!) {
+      matchers[method] ??= this.buildMatcher(method)
+    }
 
     // Release cache
     this.middleware = this.routes = undefined
--- a/src/router/reg-exp-router/trie.ts
+++ b/src/router/reg-exp-router/trie.ts
@@ -30,7 +30,7 @@ export class Trie {
      *  - /* wildcard
      *  - character
      */
-    const tokens = path.match(/(?::[^\/]+)|(?:\/\*$)|./g) || []
+    const tokens = path.match(/(?::[^\/]+)|(?:\/\*$)|./g) ?? []
     for (let i = groups.length - 1; i >= 0; i--) {
       const [mark] = groups[i]
       for (let j = tokens.length - 1; j >= 0; j--) {
@@ -58,11 +58,11 @@ export class Trie {
 
     regexp = regexp.replace(/#(\d+)|@(\d+)|\.\*\$/g, (_, handlerIndex, paramIndex) => {
       if (typeof handlerIndex !== 'undefined') {
-        indexReplacementMap[++captureIndex] = Number(handlerIndex)
+        indexReplacementMap[++captureIndex] = +handlerIndex
         return '$()'
       }
       if (typeof paramIndex !== 'undefined') {
-        paramReplacementMap[Number(paramIndex)] = ++captureIndex
+        paramReplacementMap[+paramIndex] = ++captureIndex
         return ''
       }
 


usualoma:
Hi @aquapi 
Thanks for trying to optimize RegExpRouter.

However, regarding d907da1 and 6dcc1a4, I do not want to merge those changes. I do not see benefit to be gained by merging them.
If you think there are improvements that could be made by merging these PRs, please let us know specifically, e.g. by benchmarking.

But I think that maybe the following changes could have a good outcome.

https://github.com/honojs/hono/pull/2285/commits/d907da1bbb1de166df37ff5c24ec457357fe4cd9#diff-06f981941f95eecd6e249629fd9d5d4 badaf8dfbe96ade6219e67691359df679L217-R212

If there are changes that clearly make a difference in performance, I would like to see them broken down into smaller pieces instead of being grouped into one PR.

Best regards.

usualoma:
@aquapi

Thanks for the explanation!

I don't think it needs to be a significant improvement. Even a small improvement would be fine, and if the benchmark shows a change, I think we can consider merging. If the benchmark doesn't show a change, then I think we can't merge.

aquapi:
@usualoma I noticed that Hono has been slower in 3.11 than in 3.8 when I benchmark with my previous framework and Elysia.

usualoma:
@aquapi Thanks! Your information is very helpful.

The following changes were also observed on the benchmark In particular, `if (typeof staticMatch ! == 'undefined') {` is considered to be an awesome improvement. Could you please make this part a separate PR? I think that could be merged quickly.

CleanShot 2024-02-28 at 22 17 28@2x

usualoma:
Ah, it's difficult, again, I've checked several times and there seems to be little difference. Might be very slightly faster, though.

usualoma:
I'll post the results of comparing the contents of this PR with `benchmarks/routers`, as measured in my environment. I don't think we'll see any obvious "improved" difference.

```
% npm run bench:node

> bench:node
> tsx ./src/bench.mts

cpu: Apple M2 Pro
runtime: node v20.0.0 (arm64-darwin)

benchmark                                 time (avg)             (min … max)       p75       p99      p995
---------------------------------------------------------------------------- -----------------------------
• short static - GET /user
---------------------------------------------------------------------------- -----------------------------
Hono TrieRouter                       200.24 ns/iter  (188.75 ns … 251.8 ns) 202.55 ns 244.61 ns 245.11 ns
@medley/router                         90.09 ns/iter  (83.31 ns … 157.88 ns)  93.35 ns 114.32 ns 147.54 ns
find-my-way                            81.89 ns/iter     (74 ns … 288.66 ns)  84.11 ns 129.25 ns 137.24 ns
Hono RegExpRouter                      62.36 ns/iter   (57.42 ns … 81.39 ns)  62.73 ns   75.9 ns  78.48 ns
Hono RegExpRouter#2285                 63.64 ns/iter   (58.45 ns … 93.98 ns)  64.34 ns  76.62 ns  78.25 ns
koa-tree-router                        77.98 ns/iter   (71.67 ns … 98.77 ns)  78.92 ns  90.14 ns  92.25 ns
trek-router                           100.83 ns/iter  (93.38 ns … 127.11 ns) 102.15 ns 113.52 ns 117.52 ns
express (WARNING: includes handling)    1.02 µs/iter   (993.73 ns … 1.04 µs)   1.03 µs   1.04 µs   1.04 µs
koa-router                              1.86 µs/iter     (1.83 µs … 1.92 µs)   1.86 µs   1.92 µs   1.92 µs
radix3                                 78.55 ns/iter   (72.69 ns … 125.3 ns)  78.64 ns  91.29 ns  95.83 ns

summary for short static - GET /user
  Hono RegExpRouter
   1.02x faster than Hono RegExpRouter#2285
   1.25x faster than koa-tree-router
   1.26x faster than radix3
   1.31x faster than find-my-way
   1.44x faster than @medley/router
   1.62x faster than trek-router
   3.21x faster than Hono TrieRouter
   16.36x faster than express (WARNING: includes handling)
   29.75x faster than koa-router

• static with same radix - GET /user/comments
---------------------------------------------------------------------------- -----------------------------
Hono TrieRouter                       254.78 ns/iter  (236.2 ns … 284.16 ns) 256.94 ns 273.08 ns 274.19 ns
@medley/router                        133.68 ns/iter (125.25 ns … 156.16 ns) 135.47 ns 149.01 ns 150.74 ns
find-my-way                           146.24 ns/iter (136.72 ns … 164.38 ns) 150.86 ns 163.87 ns 164.27 ns
Hono RegExpRouter                      70.24 ns/iter  (64.86 ns … 224.49 ns)  70.02 ns     84 ns  91.42 ns
Hono RegExpRouter#2285                 68.87 ns/iter  (63.05 ns … 148.28 ns)  68.89 ns  86.98 ns 100.79 ns
koa-tree-router                       111.47 ns/iter  (102.3 ns … 187.54 ns) 116.38 ns 135.73 ns 142.25 ns
trek-router                           161.92 ns/iter  (149.58 ns … 234.5 ns) 166.57 ns 194.85 ns 206.42 ns
express (WARNING: includes handling)    1.08 µs/iter     (1.04 µs … 1.36 µs)   1.08 µs   1.36 µs   1.36 µs
koa-router                              1.88 µs/iter     (1.83 µs … 1.99 µs)   1.89 µs   1.99 µs   1.99 µs
radix3                                 85.25 ns/iter  (78.69 ns … 115.03 ns)  85.76 ns 105.43 ns  111.1 ns

summary for static with same radix - GET /user/comments
  Hono RegExpRouter#2285
   1.02x faster than Hono RegExpRouter
   1.24x faster than radix3
   1.62x faster than koa-tree-router
   1.94x faster than @medley/router
   2.12x faster than find-my-way
   2.35x faster than trek-router
   3.7x faster than Hono TrieRouter
   15.67x faster than express (WARNING: includes handling)
   27.25x faster than koa-router

• dynamic route - GET /user/lookup/username/hey
---------------------------------------------------------------------------- -----------------------------
Hono TrieRouter                       406.94 ns/iter  (387.08 ns … 442.2 ns) 412.72 ns 440.16 ns  442.2 ns
@medley/router                        193.02 ns/iter (179.98 ns … 262.43 ns) 197.28 ns 223.98 ns 258.86 ns
find-my-way                           226.82 ns/iter (210.57 ns … 318.23 ns) 232.25 ns  258.6 ns 282.37 ns
Hono RegExpRouter                      140.7 ns/iter (129.01 ns … 189.75 ns) 145.83 ns 178.47 ns 185.73 ns
Hono RegExpRouter#2285                138.44 ns/iter (126.77 ns … 209.34 ns)  143.7 ns 173.94 ns 176.15 ns
koa-tree-router                       168.47 ns/iter  (154.13 ns … 340.2 ns) 171.35 ns 223.29 ns 296.15 ns
trek-router                           262.39 ns/iter (244.13 ns … 304.09 ns) 268.13 ns 293.23 ns 293.69 ns
express (WARNING: includes handling)    1.74 µs/iter     (1.68 µs … 1.89 µs)   1.75 µs   1.89 µs   1.89 µs
koa-router                              1.89 µs/iter     (1.84 µs … 2.11 µs)   1.91 µs   2.11 µs   2.11 µs
radix3                                 418.4 ns/iter (387.91 ns … 727.88 ns)  420.8 ns 569.46 ns 727.88 ns

summary for dynamic route - GET /user/lookup/username/hey
  Hono RegExpRouter#2285
   1.02x faster than Hono RegExpRouter
   1.22x faster than koa-tree-router
   1.39x faster than @medley/router
   1.64x faster than find-my-way
   1.9x faster than trek-router
   2.94x faster than Hono TrieRouter
   3.02x faster than radix3
   12.6x faster than express (WARNING: includes handling)
   13.67x faster than koa-router

• mixed static dynamic - GET /event/abcd1234/comments
---------------------------------------------------------------------------- -----------------------------
Hono TrieRouter                       432.77 ns/iter (409.43 ns … 814.11 ns) 434.37 ns 708.63 ns 814.11 ns
@medley/router                        155.01 ns/iter (144.18 ns … 175.14 ns) 159.88 ns 173.83 ns 174.76 ns
find-my-way                            196.1 ns/iter (180.36 ns … 352.44 ns) 200.85 ns 247.74 ns  254.9 ns
Hono RegExpRouter                     138.03 ns/iter (127.41 ns … 175.88 ns) 143.03 ns 167.53 ns 170.13 ns
Hono RegExpRouter#2285                 138.3 ns/iter (124.95 ns … 415.39 ns) 142.59 ns 179.96 ns 328.19 ns
koa-tree-router                       142.11 ns/iter (130.26 ns … 173.81 ns) 147.19 ns 167.04 ns 168.52 ns
trek-router                           230.74 ns/iter (212.86 ns … 611.05 ns) 232.97 ns  277.3 ns 416.57 ns
express (WARNING: includes handling)    1.91 µs/iter     (1.81 µs … 2.45 µs)   1.93 µs   2.45 µs   2.45 µs
koa-router                              1.95 µs/iter     (1.87 µs … 2.16 µs)   1.96 µs   2.16 µs   2.16 µs
radix3                                407.98 ns/iter (372.38 ns … 873.07 ns) 408.31 ns 757.85 ns 873.07 ns

summary for mixed static dynamic - GET /event/abcd1234/comments
  Hono RegExpRouter
   1x faster than Hono RegExpRouter#2285
   1.03x faster than koa-tree-router
   1.12x faster than @medley/router
   1.42x faster than find-my-way
   1.67x faster than trek-router
   2.96x faster than radix3
   3.14x faster than Hono TrieRouter
   13.81x faster than express (WARNING: includes handling)
   14.11x faster than koa-router

• post - POST /event/abcd1234/comment
---------------------------------------------------------------------------- -----------------------------
Hono TrieRouter                        447.1 ns/iter  (415.69 ns … 843.1 ns) 442.07 ns 781.73 ns  843.1 ns
@medley/router                        154.92 ns/iter  (140.07 ns … 741.8 ns) 156.95 ns 201.45 ns  404.4 ns
find-my-way                           196.99 ns/iter (182.73 ns … 346.23 ns) 201.01 ns 219.08 ns 275.64 ns
Hono RegExpRouter                     125.18 ns/iter  (114.34 ns … 208.3 ns) 128.68 ns 156.86 ns  159.9 ns
Hono RegExpRouter#2285                121.42 ns/iter  (111.3 ns … 173.65 ns) 125.89 ns 151.38 ns 157.38 ns
koa-tree-router                       132.07 ns/iter (119.13 ns … 383.92 ns) 135.25 ns 190.56 ns 285.68 ns
trek-router                           195.78 ns/iter (178.18 ns … 720.18 ns) 198.48 ns 240.87 ns 384.69 ns
express (WARNING: includes handling)    1.94 µs/iter     (1.85 µs … 2.76 µs)   1.94 µs   2.76 µs   2.76 µs
koa-router                              1.95 µs/iter     (1.87 µs … 2.81 µs)   1.92 µs   2.81 µs   2.81 µs
radix3                                402.49 ns/iter   (365.15 ns … 1.18 µs) 399.18 ns  936.3 ns   1.18 µs

summary for post - POST /event/abcd1234/comment
  Hono RegExpRouter#2285
   1.03x faster than Hono RegExpRouter
   1.09x faster than koa-tree-router
   1.28x faster than @medley/router
   1.61x faster than trek-router
   1.62x faster than find-my-way
   3.31x faster than radix3
   3.68x faster than Hono TrieRouter
   15.98x faster than express (WARNING: includes handling)
   16.08x faster than koa-router

• long static - GET /very/deeply/nested/route/hello/there
---------------------------------------------------------------------------- -----------------------------
Hono TrieRouter                       374.03 ns/iter   (340.91 ns … 1.06 µs) 366.38 ns   1.05 µs   1.06 µs
@medley/router                        116.51 ns/iter (105.59 ns … 743.05 ns) 119.17 ns 148.58 ns  153.7 ns
find-my-way                           198.75 ns/iter  (183.9 ns … 247.71 ns) 202.96 ns 230.45 ns 247.64 ns
Hono RegExpRouter                      76.53 ns/iter  (67.16 ns … 971.64 ns)  74.97 ns  97.45 ns 115.39 ns
Hono RegExpRouter#2285                 73.18 ns/iter  (66.76 ns … 114.22 ns)  73.94 ns  92.64 ns  96.16 ns
koa-tree-router                       116.87 ns/iter (105.44 ns … 631.86 ns) 119.46 ns 146.16 ns 157.74 ns
trek-router                           134.99 ns/iter (123.71 ns … 803.73 ns) 136.61 ns 159.68 ns 181.27 ns
express (WARNING: includes handling)    1.39 µs/iter     (1.32 µs … 1.97 µs)   1.38 µs   1.97 µs   1.97 µs
koa-router                              1.87 µs/iter     (1.83 µs … 1.96 µs)   1.89 µs   1.96 µs   1.96 µs
radix3                                  88.2 ns/iter    (77.64 ns … 1.02 µs)  86.78 ns 114.21 ns 139.58 ns

summary for long static - GET /very/deeply/nested/route/hello/there
  Hono RegExpRouter#2285
   1.05x faster than Hono RegExpRouter
   1.21x faster than radix3
   1.59x faster than @medley/router
   1.6x faster than koa-tree-router
   1.84x faster than trek-router
   2.72x faster than find-my-way
   5.11x faster than Hono TrieRouter
   18.93x faster than express (WARNING: includes handling)
   25.6x faster than koa-router

• wildcard - GET /static/index.html
---------------------------------------------------------------------------- -----------------------------
Hono TrieRouter                       339.98 ns/iter   (314.06 ns … 1.16 µs) 335.93 ns 556.85 ns   1.16 µs
@medley/router                        113.63 ns/iter   (102.24 ns … 1.36 µs)  113.7 ns  128.6 ns 134.74 ns
find-my-way                           170.89 ns/iter (157.09 ns … 221.62 ns) 174.91 ns 209.06 ns 212.34 ns
Hono RegExpRouter                     136.59 ns/iter (125.53 ns … 182.78 ns) 141.18 ns  163.6 ns 168.46 ns
Hono RegExpRouter#2285                142.06 ns/iter   (126.14 ns … 1.05 µs) 141.12 ns 175.07 ns 966.14 ns
koa-tree-router                       150.44 ns/iter   (136.25 ns … 1.08 µs) 151.68 ns 192.38 ns    237 ns
trek-router                           181.29 ns/iter    (158.03 ns … 1.4 µs)  179.3 ns 258.46 ns 713.84 ns
express (WARNING: includes handling)     2.1 µs/iter     (2.02 µs … 2.92 µs)   2.08 µs   2.92 µs   2.92 µs
koa-router                              1.93 µs/iter     (1.84 µs … 3.08 µs)   1.88 µs   3.08 µs   3.08 µs
radix3                                384.35 ns/iter   (357.76 ns … 1.04 µs)  383.9 ns 429.39 ns   1.04 µs

summary for wildcard - GET /static/index.html
  @medley/router
   1.2x faster than Hono RegExpRouter
   1.25x faster than Hono RegExpRouter#2285
   1.32x faster than koa-tree-router
   1.5x faster than find-my-way
   1.6x faster than trek-router
   2.99x faster than Hono TrieRouter
   3.38x faster than radix3
   16.99x faster than koa-router
   18.51x faster than express (WARNING: includes handling)

• all together
---------------------------------------------------------------------------- -----------------------------
Hono TrieRouter                         2.26 µs/iter     (2.08 µs … 3.45 µs)   2.18 µs   3.45 µs   3.45 µs
@medley/router                        673.41 ns/iter   (605.69 ns … 1.86 µs) 652.89 ns   1.86 µs   1.86 µs
find-my-way                             1.06 µs/iter     (1.01 µs … 1.79 µs)   1.06 µs   1.79 µs   1.79 µs
Hono RegExpRouter                     459.55 ns/iter (437.51 ns … 553.96 ns) 464.16 ns 543.01 ns 553.96 ns
Hono RegExpRouter#2285                478.36 ns/iter   (433.17 ns … 1.35 µs) 470.97 ns   1.01 µs   1.35 µs
koa-tree-router                       659.92 ns/iter   (586.73 ns … 1.92 µs) 622.73 ns   1.92 µs   1.92 µs
trek-router                           990.82 ns/iter      (940 ns … 2.02 µs) 982.63 ns   2.02 µs   2.02 µs
express (WARNING: includes handling)   12.07 µs/iter    (9.96 µs … 14.05 ms)  10.83 µs  14.54 µs  16.42 µs
koa-router                             13.66 µs/iter    (11.83 µs … 9.85 ms)  12.75 µs     17 µs  30.17 µs
radix3                                  1.66 µs/iter     (1.51 µs … 3.12 µs)   1.59 µs   3.12 µs   3.12 µs

summary for all together
  Hono RegExpRouter
   1.04x faster than Hono RegExpRouter#2285
   1.44x faster than koa-tree-router
   1.47x faster than @medley/router
   2.16x faster than trek-router
   2.31x faster than find-my-way
   3.6x faster than radix3
   4.91x faster than Hono TrieRouter
   26.27x faster than express (WARNING: includes handling)
   29.74x faster than koa-router
```

aquapi:
@usualoma You need some options to explicitly enable JIT
and also test with JITLess

I normally do testing in Bun so here is what I use
```ts
"scripts": {
    "jit": "BUN_JSC_jitPolicyScale=0.0 BUN_JSC_thresholdForOptimizeSoon=0.0 BUN_JSC_thresholdForJITSoon=0.0 bun run",
    "jitless": "BUN_JSC_UseJit=0 bun run"
}
```

usualoma:
@aquapi
Thanks!
Anyway, as I mentioned in the beginning I have the following opinions.

* If the benchmark shows even a small improvement in performance, I want to merge, if no improvement, I don't want to merge. If you think you can see an improvement, then I would like you to provide the benchmarks.
* If the change brings improvement, I want to merge only the part that brings improvement, and I don't want to merge changes (commits) that don't show improvement. For this reason, we would like the `match()` improvement and the build step improvement to be separate PRs.

Since modern JavaScript runtime often has no difference between for loop and forEach, and even if there is, it is very slight, I see no need to rewrite forEach differently in build step. Unless there is a clear performance improvement by benchmarking.

aquapi:
Here are reasons why you should use for loop instead of for each

- for each callback often use outer dependencies (objects, arrays, etc...) which is in many cases impossible to inline for the JIT compiler
- each time you use for each it creates a closure (the reason why middleware is slow as well)
- For in exists for a reason and it is way more straightforward than Object.keys().forEach and it is faster
- For loop is obviously faster in jitless mode (you know it jitless is important for hono right?)
I can post a benchmark tomorrow to prove my point

usualoma:
@aquapi
Thanks.

Is there, in fact, a production environment that is typically run jitless with bun, nodejs, or other edge runtimes, etc.? If so, it would be useful to benchmark with the same setup as actual production.

I am aware that there are runtimes like LLRT that do not use JIT. If this is an optimization for such an environment, I think it would be better to benchmark in that environment rather than in bun's jitless mode.

usualoma:
@aquapi 

Thanks a lot!

### `Object.keys(obj).forEach` to for-in loop

Ah well, thanks, I was mistaken. I thought you were talking about "changing all forEach to for loop", but you mean "Object.keys(obj).forEach to for-in loop". That certainly makes sense.

The build step of RegExpRouter is heavy in other parts, so I don't really see the need to optimize forEach for a few keys and improve it by a few tens of ns. I also think the following changes are good changes, even without performance, as I think they simply improve readability.

```diff
diff --git a/src/router/reg-exp-router/router.ts b/src/router/reg-exp-router/router.ts
index a7bd3262..8ef9a8a6 100644
--- a/src/router/reg-exp-router/router.ts
+++ b/src/router/reg-exp-router/router.ts
@@ -187,7 +187,7 @@ export class RegExpRouter<T> implements Router<T> {
     for (let i = 0, len = paths.length; i < len; i++) {
       const path = paths[i]
 
-      Object.keys(routes).forEach((m) => {
+      for (const m in routes) {
         if (method === METHOD_NAME_ALL || method === m) {
           routes[m][path] ||= [
             ...(findMiddleware(middleware[m], path) ||
@@ -196,7 +196,7 @@ export class RegExpRouter<T> implements Router<T> {
           ]
           routes[m][path].push([handler, paramCount - len + i + 1])
         }
-      })
+      }
     }
   }
 
```

However, as for the following code, I am not sure if it is a good idea to increase the number of lines and change it to for-in, but I think it is okay to do so.

```diff
@@ -228,9 +228,12 @@ export class RegExpRouter<T> implements Router<T> {
   private buildAllMatchers(): Record<string, Matcher<T> | null> {
     const matchers: Record<string, Matcher<T> | null> = {}
 
-    ;[...Object.keys(this.routes!), ...Object.keys(this.middleware!)].forEach((method) => {
+    for (const method in this.routes!) {
       matchers[method] ||= this.buildMatcher(method)
-    })
+    }
+    for (const method in this.middleware!) {
+      matchers[method] ||= this.buildMatcher(method)
+    }
 
     // Release cache
     this.middleware = this.routes = undefined
```


### `for (let i = 0, { length } = keys; i < length; i++) {`

The benchmarks I took showed no improvement, can you show an example of improvement?

```js

const obj = {}
new Array(100).fill(0).forEach((_, i) => {
    obj[i] = Math.random()
})

group('for loop', () => {
    function desctructuring() {
        const keys = Object.keys(obj)
        const arr = []
        for (let i = 0, { length } = keys; i < length; i++) {
            arr.push(keys[i])
        }
    }

    function assignment() {
        const keys = Object.keys(obj)
        const arr = []
        for (let i = 0, length = keys.length; i < length; i++) {
            arr.push(keys[i])
        }
    }

    bench('desctructuring', desctructuring)
    bench('assignment', assignment)
})

run()
```

```
% node src/for-len.mjs
cpu: Apple M2 Pro
runtime: node v20.0.0 (arm64-darwin)

benchmark           time (avg)             (min … max)       p75       p99      p995
------------------------------------------------------ -----------------------------
• for loop
------------------------------------------------------ -----------------------------
desctructuring    1.14 µs/iter      (1.1 µs … 1.16 µs)   1.14 µs   1.16 µs   1.16 µs
assignment        1.09 µs/iter     (1.05 µs … 1.14 µs)   1.09 µs   1.14 µs   1.14 µs

summary for for loop
  assignment
   1.04x faster than desctructuring
% node --jitless src/for-len.mjs
Warning: disabling flag --expose_wasm due to conflicting flags
cpu: Apple M2 Pro
runtime: node v20.0.0 (arm64-darwin)

benchmark           time (avg)             (min … max)       p75       p99      p995
------------------------------------------------------ -----------------------------
• for loop
------------------------------------------------------ -----------------------------
desctructuring    3.85 µs/iter     (3.79 µs … 3.97 µs)   3.87 µs   3.97 µs   3.97 µs
assignment        3.87 µs/iter     (3.78 µs … 4.22 µs)   3.87 µs   4.22 µs   4.22 µs

summary for for loop
  desctructuring
   1x faster than assignment
% BUN_JSC_jitPolicyScale=0.0 BUN_JSC_thresholdForOptimizeSoon=0.0 BUN_JSC_thresholdForJITSoon=0.0 bun run src/for-len.mjs
cpu: Apple M2 Pro
runtime: bun 1.0.26 (arm64-darwin)

benchmark           time (avg)             (min … max)       p75       p99      p995
------------------------------------------------------ -----------------------------
• for loop
------------------------------------------------------ -----------------------------
desctructuring    2.86 µs/iter     (2.71 µs … 3.59 µs)   2.88 µs   3.59 µs   3.59 µs
assignment        2.84 µs/iter     (2.71 µs … 2.97 µs)   2.88 µs   2.97 µs   2.97 µs

summary for for loop
  assignment
   1.01x faster than desctructuring
% BUN_JSC_UseJit=0 bun run src/for-len.mjs
cpu: Apple M2 Pro
runtime: bun 1.0.26 (arm64-darwin)

benchmark           time (avg)             (min … max)       p75       p99      p995
------------------------------------------------------ -----------------------------
• for loop
------------------------------------------------------ -----------------------------
desctructuring    5.18 µs/iter     (4.96 µs … 5.47 µs)   5.24 µs   5.47 µs   5.47 µs
assignment        5.05 µs/iter      (4.8 µs … 5.31 µs)   5.17 µs   5.31 µs   5.31 µs

summary for for loop
  assignment
   1.02x faster than desctructuring
```

### `||` to `??`

The benchmarks I took showed no improvement, can you show an example of improvement?

```ts

const path = '/:a/:b'

group('|| or ??', () => {
    function a() {
        let c = 0
        c += (path.match(/\/:/g) || []).length
    }

    function b() {
        let c = 0
        c += (path.match(/\/:/g) ?? []).length
    }

    bench('||', a)
    bench('??', b)
})

run()
```

```
% node src/null.mjs
cpu: Apple M2 Pro
runtime: node v20.0.0 (arm64-darwin)

benchmark      time (avg)             (min … max)       p75       p99      p995
• || or ??
||          66.87 ns/iter   (60.84 ns … 85.08 ns)   68.7 ns  77.58 ns   80.4 ns
??          68.16 ns/iter   (62.38 ns … 88.46 ns)  70.06 ns  75.63 ns  78.72 ns

summary for || or ??
  ||
   1.02x faster than ??
% node --jitless src/null.mjs
Warning: disabling flag --expose_wasm due to conflicting flags
cpu: Apple M2 Pro
runtime: node v20.0.0 (arm64-darwin)

benchmark      time (avg)             (min … max)       p75       p99      p995
• || or ??
||         121.53 ns/iter  (114.6 ns … 142.52 ns) 123.55 ns 129.86 ns 130.56 ns
??         122.71 ns/iter (115.03 ns … 144.98 ns) 124.69 ns 134.41 ns 137.47 ns

summary for || or ??
  ||
   1.01x faster than ??
% BUN_JSC_UseJit=0 bun run src/null.mjs
cpu: Apple M2 Pro
runtime: bun 1.0.26 (arm64-darwin)

benchmark      time (avg)             (min … max)       p75       p99      p995
• || or ??
||          358.2 ns/iter     (329.5 ns … 525 ns) 359.37 ns 513.72 ns    525 ns
??         364.04 ns/iter (332.35 ns … 672.52 ns)  361.8 ns 570.53 ns 672.52 ns

summary for || or ??
  ||
   1.02x faster than ??
% BUN_JSC_jitPolicyScale=0.0 BUN_JSC_thresholdForOptimizeSoon=0.0 BUN_JSC_thresholdForJITSoon=0.0 bun run src/null.mjs  
cpu: Apple M2 Pro
runtime: bun 1.0.26 (arm64-darwin)

benchmark      time (avg)             (min … max)       p75       p99      p995
• || or ??
||          53.91 ns/iter  (49.65 ns … 149.01 ns)  52.73 ns  94.96 ns 100.22 ns
??          54.16 ns/iter  (49.98 ns … 114.03 ns)  54.27 ns  94.96 ns    101 ns

summary for || or ??
  ||
   1x faster than ??
```


### `match()`

The router benchmarks I took showed no improvement. I even tried applying each one individually, but none of them worked.

Incidentally, in LLRT, the `typeof staticMatch ! == 'undefined'` slowed it down so much that I think at least this line change needs to be reconsidered.

Can you show in your benchmarks that there is an improvement?

```diff
diff --git a/src/router/reg-exp-router/router.ts b/src/router/reg-exp-router/router.ts
index a7bd3262..114ef8db 100644
--- a/src/router/reg-exp-router/router.ts
+++ b/src/router/reg-exp-router/router.ts
@@ -206,20 +206,19 @@ export class RegExpRouter<T> implements Router<T> {
     const matchers = this.buildAllMatchers()
 
     this.match = (method, path) => {
-      const matcher = (matchers[method] || matchers[METHOD_NAME_ALL]) as Matcher<T>
+      const matcher = (matchers[method] ?? matchers[METHOD_NAME_ALL]) as Matcher<T>
 
       const staticMatch = matcher[2][path]
-      if (staticMatch) {
+      if (typeof staticMatch !== 'undefined') {
         return staticMatch
       }
 
       const match = path.match(matcher[0])
-      if (!match) {
+      if (match === null) {
         return [[], emptyParam]
       }
-
-      const index = match.indexOf('', 1)
-      return [matcher[1][index], match]
+      
+      return [matcher[1][match.indexOf('', 1)], match]
     }
 
     return this.match(method, path)
```

usualoma:
At the moment, I think there is some value only in changing `Object.keys(obj).forEach` to for-in loop.

aquapi:
@usualoma 
The `??` only checks if the value is `null` or `undefined` instead of checking for falsy values like `||`
Type coercion is a bit faster than typeof === 'undefined' in JITLess but not in JIT so I'm making a trade-off here (you can remove it).
Idk if `match.indexOf('', 1)` is the fastest way to check if a string is empty or not because I didn't test it yet but that doesn't seem like the best way to check.

aquapi:
Does the router checks if in case the method matcher selected does not match and it fallbacks to the all method matcher?

usualoma:
> @usualoma The `??` only checks if the value is `null` or `undefined` instead of checking for falsy values like `||` Type coercion is a bit faster than typeof === 'undefined' in JITLess but not in JIT so I'm making a trade-off here (you can remove it). Idk if `match.indexOf('', 1)` is the fastest way to check if a string is empty or not because I didn't test it yet but that doesn't seem like the best way to check.
> 
> I'm not really sure about node behavior but if you move the bench() call of '||' to after bench() call of '??' it might change the result (you can't really microbenchmark some cases)

Thanks for the comment.

While I agree that `??` may have a performance advantage with fewer comparisons, but if there is no performance improvement in microbenchmark and no improvement in the benchmark as a router, then there is no basis for merging.

usualoma:
> Does the router checks if in case the method matcher selected does not match and it fallbacks to the all method matcher?

Are you talking about this line?

https://github.com/honojs/hono/blob/81264a9cf5837462c9af6b5653ea1675759fa8db/src/router/reg-exp-router/router.ts#L209

This line is needed to apply `app.use('*')` to unregistered methods.

usualoma:
Hi @yusukebe 
I would like to hear from other maintainers.

My opinion is this, as I stated in my comment above.
> At the moment, I think there is some value only in changing `Object.keys(obj).forEach` to for-in loop.

@aquapi's explanation regarding `??` is not wrong in general terms, and I think there are situations where it would improve performance. However, as long as benchmarks cannot show improvement here, I don't think there is any need to change the current code.

As for other things in general, I don't think it is necessary to change the current code if benchmarks (either microbenchmark or hono's app benchmarks) can't show improvement.

aquapi:
@usualoma
For example I have 2 routes
GET '/user/:id': (ctx) => ctx.text("0")
ALL '/user/*': (ctx) => ctx.text("1")

What will the router return if I call `match('GET', '/user/90/info')`.

It should be returning `ctx.text("1")` but as I see in the matching code what it will do in this case is:
- The `GET` matcher is selected because it exists
- Static path does not match
- Dynamic path does not match -> Return nothing

aquapi:
Nvm I figured it out

usualoma:
@aquapi Thank you for your comment.

It is the specification that `ctx.text("1")` is returned, and it should be working that way now. If it is not working that way, it is a bug, so it would be helpful if you could submit a report with reproducible code.

If you want to verify the behavior, you can use the following test code to do so.
https://github.com/honojs/hono/blob/81264a9cf5837462c9af6b5653ea1675759fa8db/src/router/common.case.test.ts#L97-L107

aquapi:
Maybe that's enough for `RegExpRouter`. Let see what I can do with other thing

aquapi:
I'm curious why Hono TrieRouter is way slower compared to `@medley/router`.

usualoma:
@aquapi Thank you!
I appreciate your efforts and the very informative perspective you have given us on this PR.
But I still think it is unnecessary except for the `Object.keys(obj).forEach` to for-in loop change.

aquapi:
@usualoma The type coercion changes really depend on the environment you are running the code in (JIT or JITLess).

usualoma:
@aquapi If you can show me one "situation where the change would make hono faster," I'll consider it.

aquapi:
I don't know if any edge environment uses `node` but what you can do to test this in Node is:
image
Explicit checks are faster in this case.

usualoma:
@aquapi I think Cloudflare used v8, so I think we can assume the same trend as nodejs.

Nevertheless, if https://github.com/honojs/hono/pull/2285#issuecomment-1974153629 is the basis for the claim that "this PR will improve performance of hono" it seems inadequate to me.

I would like to see a benchmark that shows evidence that this PR (except for the Object.key thing) will improve performace, even if it is only one thing. Without that, I don't think we can start considering it.

aquapi:
@usualoma Can you test my router with the Hono router benchmark suite?

usualoma:
@aquapi Thanks for the mentions. I don't plan to write a benchmark using your router, but I look forward to seeing the results of the benchmarks you took.

However, since the greatness of your router is not directly related to the greatness of the results of this PR
I would like you to benchmark and explain the greatness of the PR results separately.

aquapi:
Bun with JIT:
image

Bun with JITLess:
image

Node with JIT:
image

Node with JITLess:
image

aquapi:
```ts
import { RegExpRouter as OldRegExpRouter } from './hono_/src/router/reg-exp-router/router';
import { RegExpRouter as NewRegExpRouter } from './hono_/src/router/reg-exp-router/router';
import { group, run, bench } from 'mitata';

for (let i = 0; i < 15; ++i) bench('noop', () => { });

function createOld() {
    const router = new OldRegExpRouter();
    registerRoutes(router);
    return router.match;
}

function createNew() {
    const router = new NewRegExpRouter();
    registerRoutes(router);
    return router.match;
}

const routes = [
    ['PUT', '/user/:id/comments'],
    ['GET', '/user/:id'],
    ['GET', '/user'],
    ['GET', '/'],
    ['GET', '/comments/*']
]

function registerRoutes(router) {
    for (let i = 0, { length } = routes; i < length; ++i)
        router.add(...routes[i], i);

    // Initialize the match function
    router.match('GET', '/');
}

group('Startup time', () => {
    bench('Old', createOld);
    bench('New', createNew);
});

group('Performance', () => {
    const matchOld = createOld();
    const fn1 = (route) => matchOld(...route);

    const matchNew = createNew();
    const fn2 = (route) => matchNew(...route);

    bench('Old', () => routes.map(fn1));
    bench('New', () => routes.map(fn2));
});

run();
```
This is the benchmark code
I bundle it with `Bun.build` to use in node.

aquapi:
Can you add my router to Hono benchmark suite in a different PR?

usualoma:
@aquapi  Thanks.
I get an error at the following, and this PR code doesn't work in my environment in the first place.

https://github.com/honojs/hono/blob/1b313419407a894b0a7da61ca4e72f015a736158/src/router/reg-exp-router/router.ts#L203

usualoma:
> Can you add my router to Hono benchmark suite in a different PR?

It is not a decision I can make on my own, but if it would be good competition, we can consider it.

usualoma:
Postscript.

I think it would be better to add to the benchmark not only "just a fast router" but also "a router in a framework that is used in many production environments".

usualoma:
By the way, does @aquapi  have an answer for this? If so, please let me know.

> Is there, in fact, a production environment that is typically run jitless with bun, nodejs, or other edge runtimes, etc.?

I don't know of any production environments that use nodejs or bun with jitless, so I don't know the usefulness of benchmarking with jitless specified.

aquapi:
> Postscript.
> 
> I think it would be better to add to the benchmark not only "just a fast router" but also "a router in a framework that is used in many production environments".

My router is a 'proof-of-concept' that a compiled router is possible as nobody has ever done it before.

aquapi:
> @aquapi Thanks. I get an error at the following, and this PR code doesn't work in my environment in the first place.
> 
> https://github.com/honojs/hono/blob/1b313419407a894b0a7da61ca4e72f015a736158/src/router/reg-exp-router/router.ts#L203

I cloned and it works fine on both Bun and Node

usualoma:
> My router is a 'proof-of-concept' that a compiled router is possible as nobody has ever done it before.

I see, it seems difficult to add.

usualoma:
> I cloned and it works fine on both Bun and Node

There's no `len` in this context.  Get `ReferenceError: len is not defined`  error?

aquapi:
> > I cloned and it works fine on both Bun and Node
> 
> There's no `len` in this context. Get `ReferenceError: len is not defined` error?

Ah I see
I need to change that to length

usualoma:
@aquapi 

Here are the results I got from AWS t3.medium
https://github.com/usualoma/honojs-2285-bench

On the jitless side, this PR code often wins slightly, but on the jit side, the main branch often wins. I also took measurements on my MacBook just to be sure, and the trend is generally the same.

I think the jit side is more important for hono, so given this result, the current main should be adopted.

Is this my measurement method wrong?

usualoma:
Oh, excuse me, when I pasted the `node:jit` result into honojs-2285-bench above, this PR beat it by "1.01x". But it's a very small difference here, and since they switch every run, I think it's safe to assume there's no difference in performance.

The following is the result of another run immediately afterwards.

```
[ec2-user@ip-10-1-0-53 honojs-2285-bench]$ npm run node:jit

> honojs-2285-bench@1.0.0 node:jit
> esbuild --bundle --platform=node bench.mts | node

cpu: Intel(R) Xeon(R) Platinum 8259CL CPU @ 2.50GHz
runtime: node v21.6.2 (x64-linux)

benchmark      time (avg)             (min … max)       p75       p99      p999
noop          180 ps/iter     (121 ps … 2'426 ns)    131 ps    241 ps   1.57 ns !
noop          160 ps/iter     (121 ps … 3'048 ns)    131 ps    231 ps    463 ps !
noop          144 ps/iter     (121 ps … 1'131 ns)    129 ps    169 ps    302 ps !
noop          172 ps/iter     (121 ps … 2'857 ns)    129 ps    188 ps    674 ps !
noop          182 ps/iter     (121 ps … 3'232 ns)    131 ps    217 ps    673 ps !
noop          183 ps/iter     (121 ps … 1'856 ns)    131 ps    243 ps   3.35 ns !
noop          213 ps/iter     (121 ps … 2'402 ns)    131 ps    215 ps   3.68 ns !
noop          155 ps/iter     (121 ps … 2'102 ns)    131 ps    217 ps    331 ps !
noop          173 ps/iter     (121 ps … 1'931 ns)    131 ps    247 ps    877 ps !
noop          184 ps/iter     (121 ps … 2'713 ns)    130 ps    208 ps    404 ps !
noop          177 ps/iter     (121 ps … 2'480 ns)    130 ps    243 ps    667 ps !
noop          165 ps/iter     (121 ps … 2'449 ns)    131 ps    232 ps    334 ps !
noop          183 ps/iter     (121 ps … 3'128 ns)    130 ps    194 ps   1.17 ns !
noop          156 ps/iter     (121 ps … 2'009 ns)    131 ps    220 ps    322 ps !
noop          165 ps/iter     (121 ps … 1'764 ns)    130 ps    192 ps    545 ps !

• Startup time
Old        75'710 ns/iter (30'040 ns … 16'463 µs) 64'628 ns    292 µs  6'045 µs
New        62'039 ns/iter (28'687 ns … 15'356 µs) 60'357 ns    228 µs  2'876 µs

summary for Startup time
  New
   1.22x faster than Old

• Performance
Old           751 ns/iter     (525 ns … 3'241 ns)    749 ns  2'600 ns  3'241 ns
New           760 ns/iter     (532 ns … 3'685 ns)    748 ns  2'695 ns  3'685 ns

summary for Performance
  Old
   1.01x faster than New
```

usualoma:
I think this PR wins on "Startup time" because of the change in `Object.keys`. I am looking at the "Performance" one.

From these results, I still have no change from the following opinion.

> At the moment, I think there is some value only in changing `Object.keys(obj).forEach` to for-in loop.

aquapi:
For me I prefer explicit checks as they are more clear what you are checking than type coercion.
And you cannot really tell what is the performance difference of these two because the difference is in nanoseconds.
The best way to tell if there is a difference is to look at the JS spec.

usualoma:
@aquapi Thank you!
I think we have exhausted the factors necessary to make a decision, so this is the last comment from me. I will not add any further comments. I would consider the following.

This PR change is not necessary for hono and should be declined. "Object.keys(obj).forEach to for-in loop" change would be useful, so if anyone thinks it is necessary, it would be good to create another PR and merge it in.

aquapi:
@usualoma 
Do you really think that any outside contributor will actually spend time optimizing Hono in the first place?
The thing is that, your codebase is really confusing and the code style is just making the interpreter (or the compiler) work as hard as possible.

Many frameworks right now just try to be fast in the first like 2 major versions or something then everything will be slow as usual.
Hono keeps adding abstractions on top of abstractions and it just gonna be worse over time.

There are many problems with Hono, for examples:
- JSX (They shouldn't be on the server as they are pure overhead).
- Performance is gonna be worse as long as you keep adding features (3.8 -> 3.11 has a huge performance decrease).
- Middleware (Middlewares chain is at least 10x slower than without middleware)

About JSX:
The only thing you can change in JSX compilation is `jsxFactory`, even tools like `@kitajs/html` is still at least 2x slower than just a simple template string.

aquapi:
I love how easy it is to surpass the fastest JS web frameworks right now.

usualoma:
@aquapi 

Thanks for creating the PR for this issue.
You seem to want comments so I will comment, but this is really the last time.

Yes, I don't particularly disagree with your opinion on hono. I think there is room for many improvements. I thought that I should decline regarding this PR, but @yusukebe's opinion might be good to hear as well.

Just well, while I don't deny your frustration with hono, I see the following problems with your approach in this PR.
* Trying to rewrite your own style by mixing necessary and unnecessary changes in your first contribution to a particular OSS.
* Not answering some questions to the end.

It may be none of my business, but I think you can write great code and I think you should take things a little more step-by-step and carefully.

I wish you all the best in your future endeavors.

aquapi:
@usualoma @yusukebe
image
This shows clearly how bad Hono is, compared to a `compiled` router, which does way more work than a `linear` one.
https://github.com/bit-js/byte/tree/main/bench

PR #2293, opened 2024-03-01
fix(stream): remove async from onabort

### LINKED ISSUE #2289: Why is onAbort returning a promise?
Perhaps I'm uninformed, but my understanding is that the onAbort is unnecessarily marked as async which can confuse consumers of the `StreamingApi`

https://github.com/honojs/hono/blob/81264a9cf5837462c9af6b5653ea1675759fa8db/src/utils/stream.ts#L68

sor4chi:

- [x] Add tests
- [x] Run tests

resolve #2289

### DIFF
--- a/deno_dist/utils/stream.ts
+++ b/deno_dist/utils/stream.ts
@@ -58,7 +58,7 @@ export class StreamingApi {
     this.writer = this.writable.getWriter()
   }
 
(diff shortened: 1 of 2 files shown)


yusukebe (approved):
Looks good to me!

PR #2307, opened 2024-03-03
fix (jsx/dom): keep ref.current value during lifecycle.

### LINKED ISSUE #2303: Client components: ref values are not reflected in the UI

4.0.8

dev

# Preface

I have just recently tried Hono Client Components, and it's definitely great to operate the same primitives (like JSX) om both sides of the wire.

I have some fundamental issue, though. The documentation is very brief and doesn't give much insight into how they work. Are they using VDOM? Reconciliation? Surgical updates? Do I need to provide keys when rendering an array? How do I type my component functions so that the TS language server understands that I can pass the special props, like `key`?

For the lack of docs, I base my assumptions on how React operates, given Hono CCs are clearly modeled after it.

# The Issue

The specific issue I have encountered just now is: updating the state doesn't seem to cause the full component rerender. In particular, the earlier updated ref values are not reflected in the UI.

**Example**

[This React app](https://reactplayground.vercel.app/#N4IgLgziBcBmCGAbCBTANCAbrK1QEsA7AExQA8A6AK1xHwFsAHAewCcwACAQUcY9lbN6HAOQUA9D0bUIZEQB1CDFuw4AlFPADGnAUNGtNOhUqZtOwDlsPwwKNc2acAvv0HCRNnQFpiQ8VqI+CiEYCaKWsyEEJyRofBEKKwcALwcfloArvQhYBQA5ihgAKKIKDmhAEIAngCSxAAUno5hAJQRUTEcgk6pVjZ2Dk4NcWAJhEnthD15hiRJDYoc6kZ51pp2peW5i4TLyxraeQDKYKz4OgCyzKRoS-uEmYiId3v7hzoU67YoWxVgDSkU2WU1aIAwUhkZBgdDMqg+YDQHEsmVQGlgSNRKFOPw4rj0Hi8YUUynMHBi1TKxDc+hEFKp3kiZgmoQg4QmZBUukyhB0+Ci3F4DVayPucS68D6WPRDQQyBQrQA3GLOpwANoAIyRqDAlQAulLUDi7LKkKglYp7oYwJlWHtdvsOAAeAB890dTsYLq40GRp3OhHyDXgX1tczArWcTvEXvd+09LsqvuA-qIQY1kejsbeHo1mTAYCiccdUQAwkEtABrFLAYWpF2inOO-YhrKscN9M6ZFDKpvN8lFSoNLsK3v95bOZzFjhuvv7Y5FDiVDiFlesbvT6N5gtRWf94CtsO5DgAMhPHA1p-PCYA8gBpLMuqdN6N7jgWwjPjrRTiVfOFvY0npFBiAobcAIAA3uWAojAbwIHwAAvFBfQAJkMeheygwhFHILl0hQBAnk4KQQEnIA) is using a ref and a state variable. When the state is updated, the UI rerenders, and the most recent ref value is displayed.

https://github.com/emirotin/hono-vite-repro represents the equivalent Hono app. Unfortunately, the new ref value is never reflected in the UI.

# Background

This contrived example is in fact based on the real issue. In my app, I'm processing the streaming data, and under the certain circumstances I set a flag (must survive re-renders) that we have reached the certain point. Finally, after the stream is drained, I set the state variable to cause re-render. Finally, I display the action element (button) iff the processing is done, and we have seen the point of interest in the stream.

_No response_

### What do you see instead?

_No response_

### Additional information

_No response_

usualoma:
fixes #2303

We have had simple examples that worked, but the actual implementation was almost what we would call a mock😢  This PR will make sure that the values are properly stored within the lifecycle.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/jsx/hooks/index.ts
+++ b/deno_dist/jsx/hooks/index.ts
@@ -9,6 +9,7 @@ export const STASH_EFFECT = 1
 const STASH_CALLBACK = 2
 const STASH_USE = 3
 const STASH_MEMO = 4
+const STASH_REF = 5
 
 export type EffectData = [
   readonly unknown[] | undefined, // deps
@@ -263,7 +264,16 @@ export const useCallback = <T extends (...args: unknown[]) => unknown>(
 
 export type RefObject<T> = { current: T | null }
 export const useRef = <T>(initialValue: T | null): RefObject<T> => {
-  return { current: initialValue }
+  const buildData = buildDataStack.at(-1) as [unknown, NodeObject]
+  if (!buildData) {
+    return { current: initialValue }
+  }
+  const [, node] = buildData
+
+  const refArray = (node[DOM_STASH][1][STASH_REF] ||= [])
+  const hookIndex = node[DOM_STASH][0]++
+
+  return (refArray[hookIndex] ||= { current: initialValue })
 }
 
 export const use = <T>(promise: Promise<T>): T => {
--- a/src/jsx/dom/index.test.tsx
+++ b/src/jsx/dom/index.test.tsx
@@ -919,52 +919,80 @@ describe('DOM', () => {
     expect(renderCount).toBe(2)
   })
 
-  it('useRef', async () => {
-    const Input = ({ label, ref }: { label: string; ref: RefObject<HTMLInputElement> }) => {
-      return (
-        <div>
-          <label>{label}</label>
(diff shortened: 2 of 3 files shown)


yusukebe (approved):
Ah, indeed, even updating the state did not change the rendering of the ref. I've confirmed this PR can fix it!

LGTM!

PR #2314, opened 2024-03-05
feat(cookie): Add Cookie bis condition check

Jxck:
New Cookie RFC (a.k.a cookie-bis) and CHIPS includes some best practices for Cookie settings which developers should follow.

This PR adding some condition check for Cookie handling especially.

- [RFC6265bis-13](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-13)
  - `Max-Age`/`Expires` limitation
  - `--Host-`/`--Secure_` prefix limitation 
- [CHIPS-01](https://www.ietf.org/archive/id/draft-cutler-httpbis-partitioned-cookies-01.html)
  - `Partitioned` limitation

Note: I'm adding link to latest draft in each comment, but should be updated to RFC link.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/utils/cookie.ts
+++ b/deno_dist/utils/cookie.ts
@@ -112,7 +112,34 @@ export const parseSigned = async (
 const _serialize = (name: string, value: string, opt: CookieOptions = {}): string => {
   let cookie = `${name}=${value}`
 
+  if (name.startsWith('__Secure-') && !opt.secure) {
+    // FIXME: replace link to RFC
+    // https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-13#section-4.1.3.1
+    throw new Error('__Secure- Cookie must have Secure attributes')
+  }
+
+  if (name.startsWith('__Host-')) {
+    // FIXME: replace link to RFC
+    // https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-13#section-4.1.3.2
+    if (!opt.secure) {
+      throw new Error('__Host- Cookie must have Secure attributes')
+    }
+
+    if (opt.path !== '/') {
+      throw new Error('__Host- Cookie must have Path attributes with "/"')
+    }
+
+    if (opt.domain) {
+      throw new Error('__Host- Cookie must not have Domain attributes')
+    }
+  }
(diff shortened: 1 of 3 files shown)


yusukebe:
Hi @Jxck !

Looks good to me!

For an application that does not follow that best practice, this change would be a breaking change. However, since it is a bad behavior that should be fixed, I think this change can be accepted as a minor release. If so, this PR will be merged to "next" for the next minor release "v4.1".

Hi @usualoma !

One point, it throws errors with `new Error()`, is that the best? I think this is the way to go, what do you think?

Jxck:
Hono is a new framework, so it's a good chance to guide developers to best practice, and also be secure by default. So I'm happy to help here.

Some note for this PR

# Max-Age & Expires

if both Max-Age and Expires exist, Max-Age will be respected.
And Max-Age is recommended since DateTime parsing causes interoperable issues all the time.
So adding a warning (console.warn) to the Expires usage can be considered.
But I didn't do it here since CookieBis doesn't clearly mention it in the draft now.

# Secure for SameSite=None

Chrome now requires Secure for SameSite=None cookies. But it's not based on a spec (IIUC), and other browsers don't require it.
Chrome is the browser with the largest market share so it's reasonable to add a warning (console.warn) to set SameSite=None without Secure.
But I didn't do it here since it's not spec related.

# Secure by default

HTTPSEveryWhere is default for the web today.
So it's better to guide users by adding Secure to every cookie.
I think there is an option to provide a default setting for every cookie setting. But I didn't do it here. It requires other agreement for design by team.

yusukebe:
Thanks @Jxck ! Great clear explanation, makes a lot of sense.

For the three above, we would like to deal with cases where they are defined as general specifications or best practices, such as with cookie-bis.

I think "Type Safe guard" is a very interesting approach. I have no idea how to implement it right now, but it would be very Hono-like.

Jxck:
TypeSafe Cookie options is on my branch https://github.com/Jxck/hono/commit/d6e3c0c2756057e16e417917a77d1c7dc2163a86

I'll send other PR later.

yusukebe:
@Jxck 

> I'll send other PR later.

Please!

PR #2339, opened 2024-03-12
feat(jwt): supported RS256, RS384, RS512 algorithm for JWT

Code-Hex:

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/utils/jwt/types.ts
+++ b/src/utils/jwt/types.ts
@@ -33,6 +33,13 @@ export class JwtTokenIssuedAt extends Error {
   }
 }
 
+export class JwtHeaderInvalid extends Error {
+  constructor(header: object) {
+    super(`jwt header is invalid: ${JSON.stringify(header)}`)
+    this.name = 'JwtHeaderInvalid'
+  }
+}
+
 export class JwtTokenSignatureMismatched extends Error {
   constructor(token: string) {
     super(`token(${token}) signature mismatched`)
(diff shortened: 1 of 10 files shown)


src/utils/jwt/types.ts
```diff
@@ -44,4 +51,45 @@ export enum AlgorithmTypes {
...
+  RS384 = 'RS384',
+  RS512 = 'RS512',
+}
+
+export enum CryptoKeyUsage {
+  Ecrypt = 'encrypt',
+  Decrypt = 'decrypt',
+  Sign = 'sign',
+  Verify = 'verify',
+  Deriverkey = 'deriveKey',
```
yusukebe, line 64: `Deriverkey` should be `Derivekey`.

yusukebe:
Hi @Code-Hex thanks!

For me, it looks good.

@metrue If you have time, could you review this?

metrue:
Sure, I will finish the review by end of today.

yusukebe:
@metrue Thanks!

metrue:
It's great to have this, LGTM, thanks @Code-Hex

Code-Hex:
@metrue Thanks for your review!

yusukebe:
Great job @Code-Hex !

We can ship this change right now, but this is a new feature. I wonder if we should release this either as a patch release or as a minor release.

I believe this is an important feature for JWT users, so I would prefer to ship it in a minor release after all or some of the remaining algorithms are implemented.

Code-Hex:
@yusukebe It was completely unexpected.

Wouldn't it be better to provide the parts of the system in the order in which they are available to us as users? This is my opinion, but I thought it would be better if we could release them in smaller units.

Anyway, There are several points to consider when releasing a batch.

1. the goal is to implement and release all algorithms only
2. only validate the various critical parameters (headers and payloads) that remain vulnerable
3. implement all of 1 and 2

Each of these is what we are considering.

1. this does not require any destructive changes and can be released on a fine-grained, algorithm-by-algorithm basis, regardless of my schedule.
2. it will almost certainly involve disruptive changes. I don't know what kind of change it will be, and I honestly don't know if it will be a big change or just a small one.
3. disruptive changes can happen because of 2.

The one thing they all have in common is that I won't know what the changes will be until I actually write the code. That and I need the code base for this PR.

@yusukebe How far do you consider working with 1, 2, or 3?

---

I am not a repository manager, so it is difficult to develop on a per-feature basis.

Can you provide a base branch for JWT on honojs/hono?

If you can do that, I can merge into it and then send a PR to merge that branch into the main branch before release.

yusukebe:
Hi @Code-Hex .

Sorry for deciding without your opinion!

> 1. the goal is to implement and release all algorithms only

I was thinking of **1**. This means "just the algorithms".

Since "header and payload validation" would result in a breaking change, I think it would be better to make that change at the time of the major releasing `v5`.

> Wouldn't it be better to provide the parts of the system in the order in which they are available to us as users? 

Yes. I agree. You do not have to rush to implement them! 

There is no definite release plan, but minor releases have been made at least once a month. I think the next one will be in mid-April and it's better to ship **only** those algorithms that are complete at that time.

Or do you want to release it as a patch-release (We must not necessarily follow semver unless breacking changes!)? This is a PR you created, and we would like to honor it.

yusukebe:
Again, I do not force or rush contributors into anything! If I am missing that consideration, I am sorry!

Code-Hex:
@yusukebe Thanks! No problem at all, let's go with plan 1.

> 1. the goal is to implement and release all algorithms only

I would like to use the code in this PR as a base, so I would like to either have you _create a new branch_ or _make this PR branch the base branch_.

> Can you provide a base branch for JWT on honojs/hono?

Which would you prefer?

If the latter, I thought it would be better to change the title of this PR to look like an algorithm implementation 🙏

yusukebe:
@Code-Hex 

How about creating a `jwt-algorithms` branch and merging this PR into it? We'll also merge other PRs which implement other algorithms into it.

Code-Hex:
@yusukebe Sounds great!

yusukebe:
Thanks! Created it. Merging now.

PR #2350, opened 2024-03-14
feat(utils/cookie): Typesafe cookie options

Jxck:
add Types for settings cookie options and make below operations typesafe

- `--Secure_` prefix
- `--Host_` prefix
- `Partitoned` option

as discussed in https://github.com/honojs/hono/pull/2314#issuecomment-1980809011

- [N/A] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/utils/cookie.ts
+++ b/deno_dist/utils/cookie.ts
@@ -2,6 +2,13 @@ import { decodeURIComponent_ } from './url.ts'
 
 export type Cookie = Record<string, string>
 export type SignedCookie = Record<string, string | false>
+
+type PartitionCookieConstraint =
+  | { partition: true; secure: true }
+  | { partition?: boolean; secure?: boolean } // reset to default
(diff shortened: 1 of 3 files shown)


Jxck:
Note:

in #2314, I added some tests for make sure invalid operation throws error.
But adding type guard makes them compile error, so I removed them in this PR. https://github.com/honojs/hono/pull/2350/commits/6a05c93a376af52ef2bfbbeaab1ac536ea98e63e

It works fine on TypeScript but if hono used on node.js with vanilla js, it should covered by runtime error handling.
for example, running error handling test only on .js or so could cover here.

yusukebe (approved):
Looks good to me!

yusukebe:
Hi @Jxck !

Awesome idea to validate them with Type-safe. I think removing the tests is good because of what you mentioned.

This is a new feature, but it's a minor change, and changing behavior only util functions. So, let's merge this into "main" and release it as a patch release later.

Thanks!

PR #2351, opened 2024-03-14
Fix(adaptor): Lambda@Edge GET/HEAD body error.

trkbt10:
This PR addresses the issue where the exception Request with GET/HEAD method cannot have body. is raised when using the GET or HEAD method with Lambda@Edge.

```typescript
import { Hono } from "hono";
import { handle } from "hono/lambda-edge";
app.get("/", (c, next) => {
    return c.json({}, 200);
});
app.post("/", (c, next) => {
    return c.json({}, 200);
});

export const handler = handle(app);
```

While POST requests work without any issues, sending a GET request results in an error.

### Changes
Made sure that the request does not include a body when the method is GET or HEAD.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/adapter/lambda-edge/handler.ts
+++ b/src/adapter/lambda-edge/handler.ts
@@ -155,16 +155,27 @@ const createRequest = (event: CloudFrontEdgeEvent) => {
   })
 
   const requestBody = event.Records[0].cf.request.body
-  const body =
-    requestBody?.encoding === 'base64' && requestBody?.data
-      ? Buffer.from(requestBody.data, 'base64')
-      : requestBody?.data
+  const method = event.Records[0].cf.request.method
+  const body = createBody(method, requestBody)
 
   return new Request(url, {
     headers,
-    method: event.Records[0].cf.request.method,
+    method,
     body,
   })
+}
+
+export function createBody(method: string, requestBody: CloudFrontRequest['body']) {
+  if (!requestBody || !requestBody.data) {
+    return undefined
+  }
+  if (method === 'GET' || method === 'HEAD') {
+    return undefined
+  }
+  if (requestBody.encoding === 'base64' && requestBody.data) {
+    return Buffer.from(requestBody.data, 'base64')
+  }
+  return requestBody.data
 }
 
 export const isContentTypeBinary = (contentType: string) => {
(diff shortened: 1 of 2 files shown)


src/adapter/lambda-edge/handler.ts
```diff
@@ -155,18 +155,29 @@ const createRequest = (event: CloudFrontEdgeEvent) => {
...
 }
 
+export function createBody(method: string, requestBody: CloudFrontRequest['body']) {
+  if (!requestBody || !requestBody.data) {
+    return undefined
+  }
+  if (method === 'GET' || method === 'HEAD') {
+    return undefined
+  }
+  if (requestBody.encoding === 'base64' && requestBody.data) {
```
watany-dev, line 175: I don't think checking requestBody.data is necessary since it is done above.

src/adapter/lambda-edge/handler.ts
```diff
@@ -155,18 +155,29 @@ const createRequest = (event: CloudFrontEdgeEvent) => {
...
 
   return new Request(url, {
     headers,
-    method: event.Records[0].cf.request.method,
+    method,
     body,
   })
 }
 
+export function createBody(method: string, requestBody: CloudFrontRequest['body']) {
```
watany-dev, line 168: Hono project is recommended to be written in Allow function, please modify it!

watany-dev:
@trkbt10 
Thanks for the correction. A `bun run denoify` after review will also resolve the CI

yusukebe (approved):
LGTM!

yusukebe:
Hi @trkbt10 

Thank you for the PR. The code is good! But, CI fails with a "denoify" matter.

The weird thing is that we wrote `// @denoify-ignore` in `src/adapter/lambda-edge/handler.ts`, so the code is not generated by the `denoify` command, so this CI should go through, but it fails.

@trkbt10 Could you run `bun run denoify` or `yarn denoify` on your side?

trkbt10:
I ran `bun run denoify` in my local environment, but no errors occurred.

Here is the result:

```bash
> bun run denoify
$ rimraf deno_dist && denoify && rimraf "deno_dist/**/*.test.{ts,tsx}"
Denoify is reading sources files from src
Deno distribution will be generated at deno_dist

```
Additionally, This did not generate any new files or change any files.

If there is anything else I can do, I would be grateful if you could let me know.

yusukebe:
@trkbt10 

I found that the CI was caused by another issue, fixed it,  and the CI on main is passed.

So, everything is good. I'll merge now. Thanks!

PR #2365, opened 2024-03-16
fix(types): `MergePath` merge blank paths correctly

yusukebe:

- [ ] Add tests
- [ ] Run tests


poteboy (approved):
Thanks! If you can tolerate that this PR slightly breaks the current behavior, I think your solution is 100x times better!

yusukebe:
Merging now!

PR #2382, opened 2024-03-19
feat(ssg): Support `extentionMap`

watany-dev:
ref: https://github.com/honojs/middleware/pull/412

- [x] Add tests
- [x] Run tests


yusukebe:
Hi @watany-dev 

Is it not enough to add YAML mime types to `src/utils/mime.ts` instead of this PR?

watany-dev:
@yusukebe 

That is simpler and better! Fixed.

src/utils/mime.ts
```diff
@@ -70,6 +70,7 @@ const baseMimes: Record<string, string> = {
   woff2: 'font/woff2',
   xhtml: 'application/xhtml+xml',
   xml: 'application/xml',
+  yaml: 'application/yaml',
```
yusukebe, line 73: Sorry, we have to reconsider it.

if it has another key for YAML mime like `yml`:

```ts
yml: 'application/yaml'
```

If more than one key exists for `application/yaml`, it is not known which extension is used. So, it may be difficult to specify the extension used in SSG by mime type alone.

yusukebe:
@watany-dev 

How about adding an `extensionMap` option like the following? It exports `defaultExtensionMap` so the user can easily add additional values to the default values.

```diff
diff --git a/src/helper/ssg/ssg.ts b/src/helper/ssg/ssg.ts
index 17040e5..b1c3608 100644
--- a/src/helper/ssg/ssg.ts
+++ b/src/helper/ssg/ssg.ts
@@ -30,8 +30,13 @@ export interface ToSSGResult {
   error?: Error
 }
 
-const generateFilePath = (routePath: string, outDir: string, mimeType: string) => {
-  const extension = determineExtension(mimeType)
+const generateFilePath = (
+  routePath: string,
+  outDir: string,
+  mimeType: string,
+  extensionMap?: Record<string, string>
+) => {
+  const extension = determineExtension(mimeType, extensionMap)
 
   if (routePath.endsWith(`.${extension}`)) {
     return joinPaths(outDir, routePath)
@@ -62,17 +67,19 @@ const parseResponseContent = async (response: Response): Promise<string | ArrayB
   }
 }
 
-const determineExtension = (mimeType: string): string => {
-  switch (mimeType) {
-    case 'text/html':
-      return 'html'
-    case 'text/xml':
-    case 'application/xml':
-      return 'xml'
-    default: {
-      return getExtension(mimeType) || 'html'
-    }
+export const defaultExtensionMap: Record<string, string> = {
+  'text/html': 'html',
+  'text/xml': 'xml',
+  'application/xml': 'xml',
+  'application/yaml': 'yaml',
+}
+
+const determineExtension = (mimeType: string, extensionMap?: Record<string, string>): string => {
+  extensionMap ??= defaultExtensionMap
+  if (mimeType in defaultExtensionMap) {
+    return defaultExtensionMap[mimeType]
   }
+  return getExtension(mimeType) || 'html'
 }
 
 export type BeforeRequestHook = (req: Request) => Request | false | Promise<Request | false>
@@ -85,6 +92,7 @@ export interface ToSSGOptions {
   afterResponseHook?: AfterResponseHook
   afterGenerateHook?: AfterGenerateHook
   concurrency?: number
+  extensionMap?: Record<string, string>
 }
 
 /**
@@ -204,14 +212,15 @@ const createdDirs: Set<string> = new Set()
 export const saveContentToFile = async (
   data: Promise<{ routePath: string; content: string | ArrayBuffer; mimeType: string } | undefined>,
   fsModule: FileSystemModule,
-  outDir: string
+  outDir: string,
+  extensionMap?: Record<string, string>
 ): Promise<string | undefined> => {
   const awaitedData = await data
   if (!awaitedData) {
     return
   }
   const { routePath, content, mimeType } = awaitedData
-  const filePath = generateFilePath(routePath, outDir, mimeType)
+  const filePath = generateFilePath(routePath, outDir, mimeType, extensionMap)
   const dirPath = dirname(filePath)
 
   if (!createdDirs.has(dirPath)) {
```

src/utils/mime.ts
watany-dev: Can I remove this place because it is represented in the DefaultExtentionMap?

watany-dev:
I've made modifications to allow specifying an extensionMap. Differing from the original idea, it seemed more straightforward for users to specify only the differences they want. Therefore, the system now adds to or overrides the defaultExtensionMap with the user-provided map. This way, users can customize only the necessary parts without having to be aware of the default values.

src/utils/mime.ts
yusukebe: Yes! You can remove it.

src/helper/ssg/ssg.ts
```diff
@@ -62,17 +67,22 @@ const parseResponseContent = async (response: Response): Promise<string | ArrayB
...
+  'text/xml': 'xml',
+  'application/xml': 'xml',
+  'application/yaml': 'yaml',
+}
+
+const determineExtension = (
+  mimeType: string,
+  userExtensionMap?: Record<string, string>
+): string => {
+  const extensionMap = { ...defaultExtensionMap, ...userExtensionMap }
```
yusukebe, line 81: I think that we can export the `defaultExtensionMap` and the user can use it and add what they want to add to it. Or the user could not use the `defaultExtensionMap`.

```ts
import { toSSG, defaultExtensionMap } from 'hono/ssg'
import { Hono } from 'hono'

//...

toSSG(app, fs, {
  extensionMap: { ...defaultExtensionMap, ...{ 'application/foo': 'foo' } }
})
```

This idea from Vitest. We can add the `exclude` paths with `defaultExclude`:

```ts
import { defineConfig, defaultExclude } from 'vitest/config'

export default defineConfig({
  test: {
    // defaultExclude is defined at:
    // https://github.com/vitest-dev/vitest/blob/fee7d8be9d6e6f710270600ae91fa35d861b7075/packages/vitest/src/defaults.ts#L6
    exclude: [...defaultExclude, 'abc', 'def']
  }
})
```

How do you think about this?

yusukebe:
Hi @watany-dev !

Thanks! I've left the comments. Chech them!

src/helper/ssg/ssg.ts
yusukebe: The nice thing about this is that users can add values and rewrite the default values.

watany-dev: Yes, let's make the default value a user-accessible method. Fixed.

yusukebe (approved):
LGTM!

yusukebe:
Hi @watany-dev 

I'll merge this into the "next" branch now.

Issue #2383, opened 2024-03-19
Unable to add content to the link element of JSX


DIYgod:
### What version of Hono are you using?

4.1.2

### What runtime/platform is your app running on?

Node.js

### What steps can reproduce the bug?

I understand that it may seem unusual, but I do need to add content to the link element.

I am using Hono to render RSS in [RSSHub](https://github.com/DIYgod/RSSHub). There is a crucial field called "link" in the RSS feed that should contain a link within the link element instead of being placed as an attribute, which is the usual practice.

You can find specific code details at: https://github.com/DIYgod/RSSHub/pull/14863/files#diff-af123bfb44d396186ec16b495df4fa9bf75a998975fe41b5ec71ebf84669744fR12

You can view the rendering result here: https://github.com/DIYgod/RSSHub/pull/14863#issuecomment-2008053080

The issue lies in the rendering result where the link appears as `<link />` instead of the expected `<link>https://example.com</link>`

Furthermore, I checked Hono's code, I discovered that it forcefully designates link as `emptyTags` at: https://github.com/honojs/hono/blob/95be96c2848d56e3c920b84b15ee5ec2448b5d12/src/jsx/base.ts#L36

(but I am not sure if it is the only limitation)

While this setting makes sense in most scenarios, I propose disabling these emptyTags settings when `jsxRenderer` is set to a non-html `docType`(just like https://github.com/DIYgod/RSSHub/pull/14863/files#diff-c1bf8fd1428bd389c2d6600077ad5b013d95300ea7ad861ca0b7f93eabe9418dR34) or allowing for customization of emptyTags.

usualoma:
Hi @DIYgod
Thanks for the report!

I think it is an obvious bug that the output is output with a self closing tag despite the presence of the CHILDREN, so this will be fixed in #2384.

PR #2384, opened 2024-03-19
fix(jsx): use self closing tag only if element has no children

usualoma:
Fixes https://github.com/honojs/hono/issues/2383

- [x] Add tests
- [x] Run tests


yusukebe (approved):
Looks good to me!

Issue #2387, opened 2024-03-20
Add useId() hook


yajihum:
### What is the feature you are proposing?

Currently, [Hono does not provide the useId hook among React Hooks](https://hono.dev/guides/jsx-dom#hooks-compatible-with-react), and we would like to add it.

Official Documentation on React's useId: 
https://react.dev/reference/react/useId

Implementation location of useId in React: 
https://github.com/facebook/react/blob/main/packages/react-reconciler/src/ReactFiberHooks.js#L3097-L3132
 
## Example usage
This is primarily used for accessibility attributes.
 By using useId to provide a unique ID, rather than hardcoding the ID, we can ensure that the generated IDs do not overlap even when the same component is rendered multiple times.

For example:
```tsx
import { useId } from 'react';

function PasswordField() {
  // In the first PasswordField, it outputs id=':r1:'
  // In the second PasswordField, it outputs id=':r2:'
  const passwordHintId = useId();

  return (
    <>
      <label>
        Password:
        <input
          type="password"
          aria-describedby={passwordHintId}
        />
      </label>
      <p id={passwordHintId}>
        The password should contain at least 18 characters
      </p>
    </>
  );
}

export default function App() {
  return (
    <>
      <h2>Choose password</h2>
      <PasswordField />
      <h2>Confirm password</h2>
      <PasswordField />
    </>
  );
}
```

## Suggestion
If you are open to this issue, I would like to submit a PR. 
However, this would be my first major contribution to an open-source project, so there might be areas where I lack experience. I would appreciate your support in such cases!

usualoma:
Hi @yajihum
Thanks for the suggestion!

Patches are welcome, but "hono/jsx" would like to keep the implementation as small as possible and would like to get it done in #2389. Is this implementation likely to achieve the results you are looking for?

usualoma:
### Why is it so simple in "hono/jsx"?

React states that it is recommended to use `useId()` to ensure the same id when used with `hydration`.
https://react.dev/reference/react/useId#why-is-useid-better-than-an-incrementing-counter

Since hono does not have a clever implementation like React's `hydration`, it is preferable to keep the implementation as simple as possible, since there is no motivation to generate the same id at this time.

yusukebe:
Hi @yajihum 

Thank you for creating the issue. I've approved #2389 , but this change includes a new feature. We are following a semver in almost case, so, this will be shipped in the next minor version `4.2`. May be it will not be so far. Please wait a moment.

[...]

PR #2389, opened 2024-03-20
feat(jsx): add useId hook

usualoma:
### Why is it so simple in "hono/jsx"?

https://github.com/honojs/hono/issues/2387#issuecomment-2009251402

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/jsx/dom/index.ts
+++ b/deno_dist/jsx/dom/index.ts
@@ -15,6 +15,7 @@ import {
   useMemo,
   useLayoutEffect,
   useReducer,
+  useId,
   useDebugValue,
 } from '../hooks/index.ts'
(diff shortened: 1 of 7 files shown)


usualoma:
Hi @yusukebe 
Would you please review?

It is not fully compatible with React's `useId()`, but I think it is compatible enough for use within hono/jsx. I would like to add this to help users migrating from React to hono/jsx.

It would be appropriate for them to be merged in a minor version.

yusukebe (approved):
LGTM!

yusukebe:
Hey @usualoma !

Thanks for the PR. I've approved it, but I'll merge it later. As you mentioned, it's better to ship this feature in a minor release ( I think it will not be so far ).

PR #2394, opened 2024-03-20
fix(jsx): allow null, undefined, and boolean as children

### LINKED ISSUE #2391: Incorrect JSX type error when one child element is undefined

4.0.6

Node

Try to pass children from one JSX component typed with `PropsWithChildren` to another:

```tsx
import type { PropsWithChildren } from "hono/jsx";

function Main({ children }: PropsWithChildren) {
  return (
    <main>{children}</main>
  )
}

function Layout({ children }: PropsWithChildren) {
  return (
    <Main>
      <nav><a href="/">Home</a></nav>
      {children}
    </Main>
  )
}
```

No type error, since children should be valid children of another component. A single undefined child is fine, so an array with an undefined child should also be fine.

### What do you see instead?

A type error:

```
Type 'Child | undefined' is not assignable to type 'Child'.
  Type 'undefined' is not assignable to type 'Child'.ts(2322)
```

### Additional information

This only occurs when passing multiple children. I assume this is because TS narrows the `Child | undefined` type to `Child[]`. Since `Child` is not allowed to be undefined this means an array of children cannot contain any undefined elements.

I think the easiest solution would be to add `undefined` to the `Child` union.

My current work around is to wrap the children in a useless fragment, to avoid creating an array of children.

Edit: I think #1193 is the same underlying problem.

usualoma:
fixes #2391

### Insufficient type definitions

`null` and `undefined` should also be allowed for JSX child elements, and even hono was implemented with this assumption in the following places, but it was omitted from the type definition. This will be fixed in c16c6c0e839188742f24eec6f86266ead02ec707

https://github.com/usualoma/hono/blob/b5ad5f4418a6ed50ac61bdb07b8f016693bfd533/src/jsx/base.ts#L76-L77

### Insufficient null/undefined/boolean handling

There were some child elements of `Suspense` and `ErrorBoundary` that had insufficient handling of `null` and `undefined`, which were fixed with e151c792cf40751cacd988fc2745ef5e08ae56e9.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/jsx/base.ts
+++ b/deno_dist/jsx/base.ts
@@ -92,7 +92,15 @@ const childrenToStringToBuffer = (children: Child[], buffer: StringBuffer): void
 }
 
 type LocalContexts = [Context<unknown>, unknown][]
-export type Child = string | Promise<string> | number | JSXNode | Child[]
+export type Child =
+  | string
+  | Promise<string>
+  | number
+  | JSXNode
+  | null
+  | undefined
+  | boolean
+  | Child[]
 export class JSXNode implements HtmlEscaped {
   tag: string | Function
   props: Props
--- a/deno_dist/jsx/components.ts
+++ b/deno_dist/jsx/components.ts
@@ -10,7 +10,9 @@ let errorBoundaryCounter = 0
 
 export const childrenToString = async (children: Child[]): Promise<HtmlEscapedString[]> => {
   try {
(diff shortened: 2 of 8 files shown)


yusukebe (approved):
LGTM! Good fix!

yusukebe:
@usualoma 

Thanks. Merge it into main now.

PR #2408, opened 2024-03-23
feat: add middlewares resolve trailing slashes on GET request

rnmeow:

- [x] Add tests
- [x] Run tests

As titled. Tests provided and denoified.


yusukebe:
Hi @rnmeow 

Thanks for the PR. This is nice! But I'm now considering whether to merge this into the core. This is so simple that users may define it themselves. And we could put how to make it in the document.

But we may accept this later.

rnmeow:
RE: @yusukebe

Understand. Dealing with trailing slashes, in my point of view, is perhaps an essential need for people building web applications, especially the larger ones.

As a framework (and also plays an important role of router), it would be convenient if Hono offers the solution out-of-the-box.  
Using middlewares might not be the best way to achieve that. Instead, for a better architecture, I may suggest to implement that in:

a) the `Hono` interface

```js
const app = new Hono({ trailingSlash: 'append' }) // 'append' | 'trim' | 'as-is'
```

b) the routing path

```js
// reqs without slash will be redirected
app.get('/path/to/route/^', (c) => {
  // ...
})
// reqs with slash will be redirected
app.get('/path/to/route^', (c) => {
  // ...
})
```

Though that is able to be a breaking, I believe that it must be an exciting change made.

PS. a new commit is awaiting approval for tests.

yusukebe:
@rnmeow 

>: Understand. Dealing with trailing slashes, in my point of view, is perhaps an essential need for people building web applications, especially the larger ones.
> As a framework (and also plays an important role of router), it would be convenient if Hono offers the solution out-of-the-box.

Thanks for sharing your thoughts. I understand well the purpose of this middleware.

> Using middlewares might not be the best way to achieve that. Instead, for a better architecture, I may suggest to implement that in:

I think using middleware is **best**. It's a really Hono-like way. We don't want to modify or add code for the core of Hono. So, using middleware is the best solution.

Then, let's make this Trailing Slash Middlere as built-in middleware!

src/middleware/trailing-slash/index.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    let resp: Response
+
+    resp = await app.request('/the/example/endpoint/with/trailing/slash')
+    expect(resp).not.toBeNull()
+    expect(resp.status).toBe(301)
+
+    resp = await app.request('/the/example/simulate/a.file')
+    expect(resp).not.toBeNull()
+    expect(resp.status).toBe(200)
+  })
```
yusukebe, line 42: Can you add a test when the request path is `/the/example/endpoint/with/trailing/slash/`?

src/middleware/trailing-slash/index.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    app.use('*', trimTrailingSlash())
+
+    app.get('/the/example/endpoint/without/trailing/slash', async (c) => {
+      return c.text('301 moved permanently')
+    })
+
+    const resp = await app.request('/the/example/endpoint/without/trailing/slash/')
+    expect(resp).not.toBeNull()
+    expect(resp.status).toBe(301)
+  })
```
yusukebe, line 19: Can you add a test when the request path is `/the/example/endpoint/without/trailing/slash` and `/`?

yusukebe (commented):
`yarn.lock` should not be changed. Can you remove the change for it?

yusukebe:
@rnmeow I've left some comments. Check them!

rnmeow:
RE: @yusukebe 

> I think using middleware is **best**. It's a really Hono-like way. We don't want to modify or add code for the core of Hono. So, using middleware is the best solution.

Yup. I didn't notice that modifying the core code would add complexity and weaken the simplicity then.  
It's better to reduce the base features; currently, there are only middlewares available.

> `yarn.lock` should not be changed. Can you remove the change for it?

That may be a mistake. Let me check.

src/middleware/trailing-slash/index.test.ts
rnmeow: I'd consider that as unneeded. Middlewares are designed for specific usages.

Unifying paths for their application is the developers' job, and adding a not matched test is probably not a good idea since redirecting all (or specified) paths is the expected behavior, but automatically detect the routes isn't.

src/middleware/trailing-slash/index.test.ts
rnmeow: Per the reason provided, currently, I'm adding a test for `/` and awaiting for further discussions.

src/middleware/trailing-slash/index.test.ts
```diff
@@ -0,0 +1,57 @@
...
+    expect(resp).not.toBeNull()
+    expect(resp.status).toBe(200)
+
+    resp = await app.request('/the/example/endpoint/without/trailing/slash/')
+    expect(resp).not.toBeNull()
+    expect(resp.status).toBe(301)
+
+    resp = await app.request('/the/example/endpoint/without/trailing/slash')
+    expect(resp).not.toBeNull()
+    expect(resp.status).toBe(301)
```
yusukebe, line 28: I think the expected value is `200`. In my thoughts, if the request URL does not have a tailing slash and enabled `trimTrailingSlash()`, it should not redirect. Wha do you think?

src/middleware/trailing-slash/index.test.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    let resp: Response
+
+    resp = await app.request('/the/example/endpoint/with/trailing/slash')
+    expect(resp).not.toBeNull()
+    expect(resp.status).toBe(301)
+
+    resp = await app.request('/the/example/endpoint/with/trailing/slash/')
+    expect(resp).not.toBeNull()
+    expect(resp.status).toBe(301)
```
yusukebe, line 51: As with the above, I think this should be `200`. If the `appendTralingSlash()` is applied and the URL has a trailing slash, it should not redirect.

yarn.lock
```diff
@@ -6495,4 +6495,4 @@ youch@^3.2.2:
 zod@^3.20.2, zod@^3.20.6:
   version "3.22.4"
   resolved "https://registry.npmjs.org/zod/-/zod-3.22.4.tgz"
-  integrity sha512-iC+8Io04lddc+mVqQ9AZ7OQ2MrUKGN+oIQyq1vemgt46jwCwLfhq7/pwnBnNXXXZb8VTVLKwp9EDkx+ryxIWmg==
```
yusukebe, line 6498: Can you remove the change of `yarn.lock` completely?

yusukebe:
Hi @rnmeow 

Thanks! I've left some comments.

And could you add the export settings to `package.json`? You can add them like other middleware under these lines:

https://github.com/honojs/hono/blob/d3403942c12c16847b0808cee1dcb0a2d6c81d1b/package.json#L201
https://github.com/honojs/hono/blob/d3403942c12c16847b0808cee1dcb0a2d6c81d1b/package.json#L425

usualoma:
Hi @rnmeow 
Thanks for the middleware suggestion!

I have the following idea.

* I think it would be better to apply this only to GET requests, and simply return a 404 for POST and other requests. (although in some cases it could be redirected in HEAD as well)
* I suspect that the current PR does not work correctly when a query string is given.
* I think it would be better to decide whether or not to redirect based on whether or not the request was a 404, rather than whether or not `.` is present.

For example, the following would be a good example

```ts
export const appendTrailingSlash = (): MiddlewareHandler => {
  return async function appendTrailingSlash(c, next) {
    await next()

    if (c.res.status === 404 && c.req.method === 'GET' && c.req.path.at(-1) !== '/') {
      const url = new URL(c.req.url)
      url.pathname += '/'
      c.res = new Response(null, {
        status: 301,
        headers: { Location: url.toString() },
      })
    }
  }
}
```

rnmeow:
RE: @usualoma 

> I think it would be better to apply this only to GET requests, and simply return a 404 for POST and other requests. (although in some cases it could be redirected in HEAD as well)
> ...
> I suspect that the current PR does not work correctly when a query string is given.
> ...
> I think it would be better to decide whether or not to redirect based on whether or not the request was a 404, rather than whether or not `.` is present.

Thank you for your feedback.

By the way, if a nonexistent file is requested, the above code will likely throws 301, and not 404.

I'm trying to make the codes consider more.

usualoma:
> By the way, if a nonexistent file is requested, the above code will likely return status 301, and not 404.

That's certainly true. We might want to think a little more carefully about this. I guess it depends on the use case, but in some cases it might be better to judge by the `.` in some cases.

If you are routing to a static file, it is possible that an image.jpg that did not exist an hour ago may exist now. Therefore, returning a 301 for a 404 is undesirable.

But, well, it is normal for `.` in URLs is normal, so for core middleware, it is not a good idea to consider it as a "URL that includes the extension if there is `.` " Therefore, I still think 404 is more appropriate.

Or we could simplify it to the extreme and make the behavior "slash on all unslashed URLs". (`. ` or status code, neither of which is referenced).

usualoma:
> 5bfac38

If you are judging by `c.res.status === 404`, you must execute `await next()` first. Before executing `await next()`, `c.res.status === 404` will always be true.

https://github.com/honojs/hono/blob/5bfac38d61238ca4d9d2e7ca0fffff3295988584/src/context.ts#L183

rnmeow:
> But, well, it is normal for `.` in URLs is normal, so for core middleware, it is not a good idea to consider it as a "URL that includes the extension if there is `.` " Therefore, I still think 404 is more appropriate.
>
> Or we could simplify it to the extreme and make the behavior "slash on all unslashed URLs". (`. ` or status code, neither of which is referenced).

Detecting status 404 might be the most acceptable solution for now.

According to the [performance test](https://perf.link/#eyJpZCI6IjBtcHBubXhreHAyIiwidGl0bGUiOiJGaW5kaW5nIG51bWJlcnMgaW4gYW4gYXJyYXkgb2YgMTAwMCIsImJlZm9yZSI6ImNvbnN0IHN0YXR1cyA9IDQwNCwgbWV0aG9kID0gJ0dFVCcsIHBhdGggPSAnL3Rlc3QvYV9maWxlLmV4dCdcblxuIiwidGVzdHMiOlt7Im5hbWUiOiJGaW5kIGl0ZW0gMTAwIiwiY29kZSI6InN0YXR1cyA9PT0gNDA0ICYmIG1ldGhvZCA9PT0gJ0dFVCcgJiYgcGF0aFtwYXRoLmxlbmd0aCAtIDFdICE9PSAnLyciLCJydW5zIjpbNTAzNDAwMCwyNjU1MDAwLDE2NDYwMDAsMTQxODYwMDAsNjY2MzAwMCw4MDQzMDAwLDE0MDUwMDAsMTI4ODAwMCwyNTQ2MDAwLDEwNDY0MDAwLDYzMzYwMDAsMTU3NjAwMCwxMDAwMCw0NDg3MDAwLDEyOTQwMDAsNDA5MDAwLDI3NjEwMDAsMjM2ODAwMCwxNDA4MjAwMCw2MTkwMDAwLDcwODcwMDAsMjU0NTAwMCwxMzk5MzAwMCw3MjAwMCw2OTI0MDAwLDE1MDUwMDAsMTE3ODUwMDAsMTU3OTAwMCw0Mzk4MDAwLDEwNTU5MDAwLDE3NDIwMDAsNTE3ODAwMCwyNDAwMDAwLDEyMDAwLDI2MzEwMDAsMjYwNjAwMCwxNTI2MDAwLDYwOTEwMDAsMzkyMDAwLDMxNjUwMDAsMzE2NjAwMCw4MDA2MDAwLDIxMjUxMDAwLDUxOTAwMDAsNjM1OTAwMCwxMTE4ODAwMCwzNjYyMDAwLDUzMTMwMDAsNDgxODAwMCw5MDAwLDcxNDAwMDAsMjU4MDAwLDk1NjMwMDAsNjIyNzAwMCwyMDcyMDAwLDM1MDkwMDAsMTY0NjAwMCw1MzM3MDAwLDIyMTkwMDAsOTUwMDAwLDQ4NzAwMCwzMTAwMCwzNzQxMDAwLDc5MTYwMDAsMTkwMDAwMCw2MDcwMDAwLDU5ODQwMDAsOTAwMCwzNzMyMDAwLDE1MjgwMDAsODQzOTAwMCwyODcxMDAwLDg5NzkwMDAsMTY1NTAwMCwxNjQ2MDAwLDYxOTMwMDAsMTIwNDkwMDAsMzUyMDAwLDYyMjIwMDAsMzI1NTAwMCwyNTU3MDAwLDMyOTgwMDAsMTAzODcwMDAsMTM4MjAwMCw0OTcwMDAsNjA2MDAwLDQ5NTMwMDAsNDk3NDAwMCw0OTg5MDAwLDgwNDgwMDAsNTY2NTAwMCw1NDY5MDAwLDMwOTgwMDAsMjQ1NzAwMCw5MDAwLDM0ODkwMDAsNDkyODAwMCw4MjY0MDAwLDEwMTk4MDAwLDgzOTAwMF0sIm9wcyI6NDU2NjgyMH0seyJuYW1lIjoiRmluZCBpdGVtIDgwMCIsImNvZGUiOiJwYXRoLm1hdGNoKC9eKFxcLy4qKStcXC4uKiQvKSAmJiBtZXRob2QgPT09ICdHRVQnICYmIHBhdGhbcGF0aC5sZW5ndGggLSAxXSAhPT0gJy8nIiwicnVucyI6WzkyNzAwMCwxMTAwMCw0NTUwMDAsMjg2NTAwMCw3NjcwMDAsMTI4OTAwMCwyODUwMDAsMTAwMCw5MTUwMDAsMTQzNzAwMCw1NDIwMDAsOTAwMCwxMTAwMCw3OTgwMDAsMTAwMCw5MDAwLDcxNTAwMCwyMDcwMDAwLDI4NDAwMDAsMTQ3OTAwMCwxNTMwMDAwLDM4OTAwMCw0ODI3MDAwLDc4MTAwMCw5MjcwMDAsMTQ1NjAwMCwyMjg1MDAwLDExMDAwLDQzMDAwMCwxNDgyMDAwLDEwMDAsMTgwMDAsMTEwMDAsMTAwMCwxMTAwMCw0NjkwMDAsMTAwMDAwLDkyNzAwMCw1NjQwMDAwLDU5ODAwMCw5ODAwMCw3ODEwMDAsNDY4MjAwMCwxMTQzMDAwLDgzNjAwMCwxNTg0MDAwLDEwMTAwMCw4NDcwMDAsNTgyMDAwLDEwMDAsOTI3MDAwLDEwMDAsMTM4MzAwMCw5MjcwMDAsMTAwMDAwLDMxMzAwMCw1OTgwMDAsNDMzMDAwLDE0MTcwMDAsNDMwMDAsMTAwMCw4MzQwMDAsMzk4MDAwLDkwOTAwMCwxMTAwMCw4NDUwMDAsOTI3MDAwLDgzNjAwMCwzNzQwMDAsMTAwMCw4OTAwMDAsMTMwMDAsNjk5MDAwLDEyMzAwMCwxMTAwMCw5MjcwMDAsMTU3NTAwMCwxMDAwLDk5MDAwMCwxMTAwMCw5MDAwLDExMDAwLDE2MzEwMDAsMTAwMCw4NzAwMCw0MzgwMDAsMTI0MTAwMCw0OTAwMDAsNzkzMDAwLDE1MTQwMDAsODk2MDAwLDE2OTAwMCwyNDYwMDAsMzQwMDAsNzk5MDAwLDM0MDAwMCw1NTIwMDAsMTY5NzAwMCwxNDczMDAwLDEyNDAwMDBdLCJvcHMiOjgwMTU0MH1dLCJ1cGRhdGVkIjoiMjAyNC0wMy0yOVQxMjozNToyMS41NzdaIn0%3D), just a simple RegEx pattern will cause a significantly reduction of the speed and simplicity.

usualoma:
@rnmeow Thank you!

In hono, if you want to change the response with middleware, you need to assign it to c.res. If you have the middleware added in the test code, you will see that in some cases the expected result will not be achieved without the assignment.

```diff
diff --git a/src/middleware/trailing-slash/index.test.ts b/src/middleware/trailing-slash/index.test.ts
index 7530ef59..5db7b65d 100644
--- a/src/middleware/trailing-slash/index.test.ts
+++ b/src/middleware/trailing-slash/index.test.ts
@@ -7,6 +7,7 @@ describe('Resolve trailing slash', () => {
   it('Trim', async () => {
     app = new Hono({ strict: true })
 
+    app.use('*', async (c, next) => await next())
     app.use('*', trimTrailingSlash())
 
     app.get('/', async (c) => {
@@ -38,6 +39,7 @@ describe('Resolve trailing slash', () => {
   it('Append', async () => {
     app = new Hono({ strict: true })
 
+    app.use('*', async (c, next) => await next())
     app.use('*', appendTrailingSlash())
 
     app.get('/', async (c) => {
diff --git a/src/middleware/trailing-slash/index.ts b/src/middleware/trailing-slash/index.ts
index 667f4bba..ead3975b 100644
--- a/src/middleware/trailing-slash/index.ts
+++ b/src/middleware/trailing-slash/index.ts
@@ -18,7 +18,7 @@ export const trimTrailingSlash = (): MiddlewareHandler => {
       const url = new URL(c.req.url)
       url.pathname = url.pathname.substring(0, url.pathname.length - 1)
 
-      return c.redirect(url.toString(), 301)
+      c.res = c.redirect(url.toString(), 301)
     }
   }
 }
@@ -40,7 +40,7 @@ export const appendTrailingSlash = (): MiddlewareHandler => {
       const url = new URL(c.req.url)
       url.pathname += '/'
 
-      return c.redirect(url.toString(), 301)
+      c.res = c.redirect(url.toString(), 301)
     }
   }
 }
```

src/middleware/trailing-slash/index.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    let resp: Response
+
+    resp = await app.request('/')
+    expect(resp).not.toBeNull()
+    expect(resp.status).toBe(200)
+
+    resp = await app.request('/the/example/endpoint/without/trailing/slash/')
+    expect(resp).not.toBeNull()
+    expect(resp.status).toBe(301)
```
yusukebe, line 27: Can you add the test to check the `location` header value?

And, you can write the same tests for all `301` responses.

rnmeow: Why not?

rnmeow:
Ready for review and awaiting for tests! @yusukebe @usualoma

yusukebe (approved):
Great! LGTM.

usualoma:
@rnmeow Thank you. LGTM!

yusukebe:
@rnmeow 

I'll merge this into [the next branch](https://github.com/honojs/hono/tree/next) for the next minor version "v4.2". We can release it later. Thank you for your contribution!

rnmeow:
RE: @yusukebe 

Got it. Thanks for your work!

PR #2411, opened 2024-03-25
fix(reg-exp-router): escape meta characters for find middleware

### LINKED ISSUE #2345: Middleware not applied correctly when path contains $

4.1.0

Bun

I understand this might be an edge case, but it caused an actual security risk in my app. When I create a nested router for the route `/$$$` the middleware for `*` are not running. This seems to be specific to the route name containing dollar signs.

Repro:

```
import { Hono } from "hono";
const app = new Hono();
app.get("/", (c) => c.text("Hono!"));
const userApp = new Hono();
userApp.use("*", (c, next) => {
  console.log("User middleware");
  return next();
});
userApp.get("/:id", (c) => c.text("User " + c.req.param().id));
userApp.get("/", (c) => c.text("User"));
app.route("/$$$", userApp);
export default app;
```

The middleware logging "User middleware" is not executed when accessing `/$$$` or `/$$$/123`

### What do you see instead?

The middleware should log `"User middleware"` on all routes of `/$$$`

### Additional information

_No response_

usualoma:
fixes #2345

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/router/reg-exp-router/router.ts
+++ b/deno_dist/router/reg-exp-router/router.ts
@@ -21,7 +21,11 @@ const nullMatcher: Matcher<any> = [/^$/, [], Object.create(null)]
 let wildcardRegExpCache: Record<string, RegExp> = Object.create(null)
 function buildWildcardRegExp(path: string): RegExp {
   return (wildcardRegExpCache[path] ??= new RegExp(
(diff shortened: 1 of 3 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@usualoma Thanks!

Issue #2399, opened 2024-03-22
[Perf] Type instantiation is excessively deep and possibly infinite.


yusukebe:
### Discussed in https://github.com/orgs/honojs/discussions/2380

<div type='discussions-op-text'>

<sup>Originally posted by **RomanNabukhotnyi** March 19, 2024</sup>
Hey! I have a large application using hono and I'm running into a problem with types.

[Repo with a reproduction](https://github.com/cellajs/hono-case)

In the `main` branch I moved all my routes and added one additional test route and then the error appeared.

<img width="527" alt="image" src="https://github.com/honojs/hono/assets/97584054/6e377d31-c9c6-4308-85ed-16cfd33f5919">
<img width="752" alt="image" src="https://github.com/honojs/hono/assets/97584054/379eb41b-76bc-48cf-ad55-7602a5f7f831">

In the `test` branch I moved all the routes from the modules into one file and was able to add 35 additional routes without this error.

I also managed to compare the difference between Hono and Elysia
With Elysia I was able to manage 1410 routes (470 routes / 3 instances) and with Hono - 120 (40 / 3).
[Repo](https://github.com/cellajs/hono-elysia-test)
I might miss or forget something here, in that case please let me know.</div>

flipvh:
Thanks for putting this on the roadmap. I have talked about this problem with @SaltyAom, who is willing to try to help out. Not sure in what capacity, but that could be very interesting and cool?! ElysiaJS just recently invested quite some time improving type performance, so perhaps we can reach out to @SaltyAom? I leave that up to you. And thanks for the suggestion you mentioned in the discussion: https://github.com/orgs/honojs/discussions/2380#discussioncomment-8874722

yusukebe:
Yes. I see that @SaltyAom is working hard on TypeScript, and Elysia is performing very well. It's awesome.

I'd love his advice if he don't mind, but I know he is busy, so I'll try it on my end; I can't use all of Elysia's code:) but it's very helpful!

bcheung:
Ran into this issue recently.

This comment helped but this makes it seem like you have to split your hono app up and deploy separate apps: https://github.com/orgs/honojs/discussions/2380#discussioncomment-8874722

You do not need to do that. I managed to get past this issue and still deploy 1 single hono application. 

I just needed to split up the clients, not the actual hono app. The actual hono app can still be merged together because the failure only occurs when you create the client and add the route types:
```javascript
const client = hc<Routes>("");
```

[...]

PR #2412, opened 2024-03-25
refactor(types): optimize inferring handler types

yusukebe:
I've refactored the code for some types of matters.

* Fixed the handlers' overload order. This improves performance when inferring handler return types.
* Fixed some bugs for types.
* Apply `Prettify`.
* Fixed `tsc` errors.

Related to #2399

### Performance improvement

Getting faster inferring handler return types. Below are the results if it's a simple case that it has one handler.

Before:

Screenshot 2024-03-25 at 16 16 31

This PR:

Screenshot 2024-03-25 at 16 15 59

This improves RPC-mode performance but does not change much because the implementation of the `hc` function is not so good.

#### The scripts for measuring 

```ts
// create.ts
import { writeFileSync } from 'fs'

let code = `
import { Hono } from 'hono'

const app = new Hono()

const routes = app
`
for (let i = 1; i < 300; i++) {
  code += `.get('/route${i}', (c) => {
    return c.json({
      ok: true
    })
  })`
}

code += `

export type AppType = typeof routes
export default routes
`

writeFileSync(import.meta.dir + '/index.ts', code)
```

```ts
// client.ts
import { AppType } from '.'

const client = hc<AppType>('/')
```

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/deno_dist/helper/testing/index.ts
+++ b/deno_dist/helper/testing/index.ts
@@ -1,4 +1,5 @@
 import { hc } from '../../client/index.ts'
+import type { ExecutionContext } from '../../context.ts'
 import type { Hono } from '../../hono.ts'
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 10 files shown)


KaelWD:
@yusukebe this breaks #2019 again

yusukebe:
@KaelWD 

Oops. Fixed with #3051. Thank you!

Issue #2401, opened 2024-03-22
AWS Lambda Streaming only works with response bodies


KnisterPeter:
### What version of Hono are you using?

4.1.3

### What runtime/platform is your app running on?

Node.js + AWS Lambda

### What steps can reproduce the bug?

- Create an app which just sends a redirect
- Deploy that to lambda with the streaming API
- It will not return the redirect

### What is the expected behavior?

The Lambda streaming should respect responses without body.

### What do you see instead?

The lambda only setup response metadata when a body is given.

### Additional information

These are the lines in the code which are limiting and leading to the described behavior.

https://github.com/honojs/hono/blob/main/src/adapter/aws-lambda/handler.ts#L129-L134

If it's welcome I'll try to create a fix for it.

yusukebe:
Hi @KnisterPeter 

Thank you for creating the issue.

> The Lambda streaming should respect responses without body.

I see!

@watany-dev Could you see this issue?

watany-dev:
@KnisterPeter
Yes, you are correct, information on streamingResponse was scarce and only the bare minimum was implemented. I will gladly review your PR.

yusukebe:
This issue can be closed. Thanks!

PR #2413, opened 2024-03-25
fix(aws-lambda): handle response without body (#2401)

KnisterPeter:
In case of a streaming lambda, the response should still contain metadata
like status and content type, even if no body is given.
This could be for example a redirect response.

### DIFF
--- a/src/adapter/aws-lambda/handler.ts
+++ b/src/adapter/aws-lambda/handler.ts
@@ -126,11 +126,13 @@ export const streamHandle = <
           headers: Object.fromEntries(res.headers.entries()),
         }
 
+        // Update response stream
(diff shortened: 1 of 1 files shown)


KnisterPeter:
@yusukebe @watany-dev I've fixed the aws-lambda issue I've reported. Please have a look and let me know if there is anything required on this PR.

yusukebe (approved):
LGTM!

yusukebe:
Thanks @KnisterPeter !

@watany-dev Could you please review this?

KnisterPeter:
@yusukebe @watany-dev Any update on this?

yusukebe:
@KnisterPeter Sorry!

Let's go with it! Thank you for contributing!

PR #2420, opened 2024-03-26
feat: introduce Method Override Middleware

yusukebe:
This PR introduces new middleware, "Method Override Middleware."

This middleware executes the handler of the specified method, which is different from the actual method of the request, depending on the value of the form, header, or query, and returns its response.

For example, since HTML forms cannot send a DELETE method, you can put the value `DELETE` in the property named `_method` and send it. And the handler for `app.delete()` will be executed.

The HTML form:

```html
<form action='/posts' method='POST'>
  <input type='hidden' name='_method' value='DELETE' />
  <input type='text' name='id' />
</form>
```

The application:

```ts
import { methodOverride } from 'hono/method-override'

const app = new Hono()
app.use('/posts', methodOverride({ app }))

app.delete('/posts', () => {
  // ...
})
```

You can change the default values or use the header value and query value:

```ts
app.use('/posts', methodOverride({ app, form: '_custom_name' }))
app.use('/posts', methodOverride({ app, header: 'X-METHOD-OVERRIDE' }))
app.use('/posts', methodOverride({ app, query: '_method' }))
```

### TODO

- [x] Ignore a GET request
- [x] Write tests for Header
- [x] Write tests for Query

- [x] Add tests
- [x] Run tests


yusukebe:
Hi @usualoma !

I think this is good. But what do you think? Maybe there's a point I'm missing.

usualoma:
Wow, I think it is great!
I think it is working fine.

I thought there was room for refactoring with type definitions and minor conditional branching, so I made the following PR.
#2429

deno_dist/middleware/method-override/index.ts
```diff
@@ -0,0 +1,123 @@
...
+          const request = new Request(newRequest, {
+            body: newParams,
+            method: method as string,
+          })
+          return app.fetch(request, c.env, getExecutionCtx(c))
+        }
+      }
+    }
+    // Method override by header
+    // There is no test now.
```
usualoma, line 82: The test appears to have already been added, no?

yusukebe: Yeah, you are right!

usualoma (approved):
Thank you!

yusukebe:
Let's go!

Issue #2395, opened 2024-03-21
Support for `Vary` Header in Cache Middleware


naporin0624:
### What is the feature you are proposing?

I suggest adding support for the `Vary` header in the cache middleware to improve content negotiation, particularly for image optimization use cases. This enhancement would allow serving different image formats based on the client's capabilities, enhancing compatibility across various browsers.

Here's a proposed change to include the `Vary` header:

```diff
export const cache = (options: {
  cacheName: string
  wait?: boolean
  cacheControl?: string,
+ vary?: string
}): MiddlewareHandler => {
  ```

```diff
await next()
if (!c.res.ok) {
  return
}
addHeader(c)
+ if (options.vary !== undefined) {
+   c.header('Vary', options.vary)
+ }
const res = c.res.clone()
```

Incorporating this option would allow for more flexible and efficient caching strategies, particularly for serving optimized images in formats like AVIF, which may not be supported by all browsers. This could be a simple yet impactful improvement to the middleware.

Would love to hear your thoughts on this addition.

yusukebe:
Hi @naporin0624 !

This is not bad. But I think we can do the same thing by adding the following middleware manually:

```ts
app.get(
  '/cache/*',
  async (c, next) => {
    await next()
    c.header('Vary', 'Accept')
  },
  cache({
    cacheName: 'my-cache',
  })
)
```

If it doesn't work, I'm sorry, but could you try?

naporin0624:
Hi, @yusukebe 

Yes, I have created the Vary middleware and use it in my projects.
Since Vary is a cache related header, I suggested adding it as an option in the cache middleware.

If this is redundant, I will close this issue.

thank you.

yusukebe:
> Since Vary is a cache related header, I suggested adding it as an option in the cache middleware.

Yeah, I think so; the Vary is "cache related". But we may not have to include it in Cache Middleware.

> I will close this issue.

Can you give me some time to think about it? I'll give it some thought.

yusukebe:
Hi @naporin0624 !

I've considered this. I think we should implement a supporting `Vary` header for Cache Middleware!

Can you create the PR? It may be a new feature in the next minor release. So please make it based on the [`next` branch](https://github.com/honojs/hono/tree/next). Thanks.

naporin0624:
hi, @yusukebe

Thanks, I will start working on the PR!

I would like to confirm this in the creation of the PR!

[...]

PR #2426, opened 2024-03-27
feat: support for `vary` header in cache middleware

naporin0624:
This PR adds support for the `Vary` header in cache middleware, allowing developers to specify headers that should trigger separate cache entries. This feature is essential for applications serving content that varies based on client headers, such as Accept or Accept-Language, enhancing content negotiation and caching efficiency.

closed: https://github.com/honojs/hono/issues/2395

### Example Usage

```ts
app.use('/example/*', cache({ cacheName: 'my-app-cache', vary: 'Accept, Accept-Encoding' }));
app.get('/example/', (c) => {
  return c.text('This content is cached with Vary header.');
});
```

This simple addition enables the server to cache different versions of a resource based on the specified header(s), ensuring that clients receive content that's tailored to their needs, such as different image formats or languages.

**Author should do the following, if applicable:**
- [X] Add tests.
- [X] Run tests.
- [X] `yarn denoify` to generate files for Deno.

### DIFF
--- a/deno_dist/middleware/cache/index.ts
+++ b/deno_dist/middleware/cache/index.ts
@@ -5,7 +5,7 @@ export const cache = (options: {
   cacheName: string
   wait?: boolean
   cacheControl?: string
-  vary?: string[]
+  vary?: string | string[]
 }): MiddlewareHandler => {
   if (options.wait === undefined) {
     options.wait = false
@@ -14,6 +14,9 @@ export const cache = (options: {
   const cacheControlDirectives = options.cacheControl
     ?.split(',')
     .map((directive) => directive.toLowerCase())
+  const varyDirectives = Array.isArray(options.vary)
+    ? options.vary
+    : options.vary?.split(',').map((directive) => directive.trim())
   // RFC 7231 Section 7.1.4 specifies that "*" is not allowed in Vary header.
   // See: https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.4
   if (options.vary?.includes('*')) {
@@ -38,7 +41,7 @@ export const cache = (options: {
       }
     }
 
-    if (options.vary) {
+    if (varyDirectives) {
       const existingDirectives =
         c.res.headers
           .get('Vary')
@@ -47,7 +50,7 @@ export const cache = (options: {
 
       const vary = Array.from(
         new Set(
-          [...existingDirectives, ...options.vary].map((directive) => directive.toLowerCase())
+          [...existingDirectives, ...varyDirectives].map((directive) => directive.toLowerCase())
         )
       ).sort()
 
--- a/deno_dist/middleware/cache/index.ts
+++ b/deno_dist/middleware/cache/index.ts
@@ -5,7 +5,7 @@ export const cache = (options: {
   cacheName: string
   wait?: boolean
   cacheControl?: string
-  vary?: string
+  vary?: string[]
 }): MiddlewareHandler => {
   if (options.wait === undefined) {
     options.wait = false
@@ -14,10 +14,9 @@ export const cache = (options: {
   const cacheControlDirectives = options.cacheControl
     ?.split(',')
     .map((directive) => directive.toLowerCase())
-  const varyDirectives = options.vary?.split(',').map((directive) => directive.trim())
   // RFC 7231 Section 7.1.4 specifies that "*" is not allowed in Vary header.
   // See: https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.4
-  if (varyDirectives?.includes('*')) {
+  if (options.vary?.includes('*')) {
     throw new Error(
       'Middleware vary configuration cannot include "*", as it disallows effective caching.'
     )
@@ -39,19 +38,18 @@ export const cache = (options: {
       }
     }
 
-    if (varyDirectives) {
+    if (options.vary) {
       const existingDirectives =
         c.res.headers
           .get('Vary')
           ?.split(',')
           .map((d) => d.trim()) ?? []
+
       const vary = Array.from(
         new Set(
-          [...existingDirectives, ...varyDirectives].map((directive) =>
-            directive.toLocaleLowerCase()
-          )
+          [...existingDirectives, ...options.vary].map((directive) => directive.toLowerCase())
         )
-      ).sort((a, b) => a.localeCompare(b))
+      ).sort()
 
       if (vary.includes('*')) {
         c.header('Vary', '*')
--- a/deno_dist/middleware/cache/index.ts
+++ b/deno_dist/middleware/cache/index.ts
@@ -5,26 +5,58 @@ export const cache = (options: {
   cacheName: string
   wait?: boolean
   cacheControl?: string
+  vary?: string
 }): MiddlewareHandler => {
   if (options.wait === undefined) {
     options.wait = false
   }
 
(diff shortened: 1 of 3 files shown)


deno_dist/middleware/cache/index.ts
```diff
@@ -5,28 +5,60 @@ export const cache = (options: {
...
+  const cacheControlDirectives = options.cacheControl
+    ?.split(',')
+    .map((directive) => directive.toLowerCase())
+  const varyDirectives = options.vary?.split(',').map((directive) => directive.trim())
+  // RFC 7231 Section 7.1.4 specifies that "*" is not allowed in Vary header.
+  // See: https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.4
+  if (varyDirectives?.includes('*')) {
+    throw new Error(
+      'Middleware vary configuration cannot include "*", as it disallows effective caching.'
+    )
```
naporin0624, line 23: > A proxy MUST NOT generate a Vary field with a "*" value.

cf: https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.4

Seeing this in the RFC, we now throw when `Vary` is set to `*`.

Please check this.

src/middleware/cache/index.ts
```diff
@@ -5,28 +5,60 @@ export const cache = (options: {
...
+  const cacheControlDirectives = options.cacheControl
+    ?.split(',')
+    .map((directive) => directive.toLowerCase())
+  const varyDirectives = options.vary?.split(',').map((directive) => directive.trim())
+  // RFC 7231 Section 7.1.4 specifies that "*" is not allowed in Vary header.
+  // See: https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.4
+  if (varyDirectives?.includes('*')) {
+    throw new Error(
+      'Middleware vary configuration cannot include "*", as it disallows effective caching.'
+    )
```
naporin0624, line 23: > A proxy MUST NOT generate a Vary field with a "*" value.

cf: https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.4

Seeing this in the RFC, we now throw when `Vary` is set to `*`.

Please check this.

src/middleware/cache/index.ts
```diff
@@ -5,28 +5,60 @@ export const cache = (options: {
...
+    if (varyDirectives) {
+      const existingDirectives =
+        c.res.headers
+          .get('Vary')
+          ?.split(',')
+          .map((d) => d.trim()) ?? []
+      const vary = Array.from(
+        new Set(
+          [...existingDirectives, ...varyDirectives].map((directive) =>
+            directive.toLocaleLowerCase()
```
naporin0624, line 51: > The value consists of either a single asterisk ("*") or a list of header field names (case-insensitive).

cf: https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.4

src/middleware/cache/index.ts
```diff
@@ -5,28 +5,60 @@ export const cache = (options: {
...
+          )
+        )
+      ).sort((a, b) => a.localeCompare(b))
```
naporin0624, line 54: I sorted the `Vary` to be given to the header, but I am willing to remove it if it affects performance.

src/middleware/cache/index.ts
yusukebe: I think this is a good behavior.

src/middleware/cache/index.ts
yusukebe: This is a matter for discussion, but I think sorting is good. Because there are cases where the cache keys are different if the header values are in a different order.

deno_dist/middleware/cache/index.ts
```diff
@@ -5,28 +5,60 @@ export const cache = (options: {
...
+    if (varyDirectives) {
+      const existingDirectives =
+        c.res.headers
+          .get('Vary')
+          ?.split(',')
+          .map((d) => d.trim()) ?? []
+      const vary = Array.from(
+        new Set(
+          [...existingDirectives, ...varyDirectives].map((directive) =>
+            directive.toLocaleLowerCase()
```
yusukebe, line 51: Is it not good good to use `toLowerCase()` instead of `toLocaleLowerCase()`?

deno_dist/middleware/cache/index.ts
```diff
@@ -5,28 +5,60 @@ export const cache = (options: {
...
+          )
+        )
+      ).sort((a, b) => a.localeCompare(b))
```
yusukebe, line 54: How about write just `.sort()`?

src/middleware/cache/index.ts
```diff
@@ -5,28 +5,60 @@ export const cache = (options: {
   cacheName: string
   wait?: boolean
   cacheControl?: string
+  vary?: string
```
yusukebe, line 8: How about accepting `string[]`?

```ts
vary: ['Vary', 'Accept-Encoding']
```

I think it makes better DX.

yusukebe:
Hi @naporin0624 !

Thanks! I've left some comments. Please check them.

@usualoma If you have time, please review this; you too? I think this feature is good!

deno_dist/middleware/cache/index.ts
naporin0624: Yes, it is!
I used LocaleLowerCase out of habit, so I'll change it to LowerCase.

[029ea29](https://github.com/honojs/hono/pull/2426/commits/029ea2992aba00365eb7b2432f160c85d60ab1a0)

deno_dist/middleware/cache/index.ts
naporin0624: It's simple! I will!

https://github.com/honojs/hono/pull/2426/commits/bc50e38b2d50043dfc594534b970ee24e68e7c7f

src/middleware/cache/index.ts
naporin0624: Nice idea. Thank you very much.

[23e339e](https://github.com/honojs/hono/pull/2426/commits/23e339e0e8ec9dab7446a3ad3806b4c29ffc05a9)

naporin0624:
> こんにちは@naporin0624！
> 
> ありがとう！いくつかコメントを残しました。ぜひチェックしてみてください。
> 
> @usualoma時間があれば、これを確認してください。あなたも？この機能は良いと思います！

@yusukebe The changes have been made. Please review them! Thank you.

src/middleware/cache/index.ts
yusukebe: @naporin0624 

Sorry for not fully explaining! I think `string | string[]` is best!

yusukebe:
@naporin0624 

There is one thing we have to consider: https://github.com/honojs/hono/pull/2426#discussion_r1544892480

src/middleware/cache/index.ts
naporin0624: You meant to accept string[] as well! It would be even better to accept the original string, but also an array.

Thanks for pointing that out.

naporin0624:
> @naporin0624
> 
> There is one thing we have to consider: [#2426 (comment)](https://github.com/honojs/hono/pull/2426#discussion_r1544892480)

@yusukebe 

string is now also accepted.
I think this is better for DX.

yusukebe (approved):
Looks good to me!

yusukebe:
@naporin0624 Thanks! Let's land it.

naporin0624:
Thanks for the PR review and looking forward to 4.2 :)

PR #2436, opened 2024-03-29
feat(middleware/cors): pass context to options.origin function

okmr-d:
Hello!
I opened this PR because I wanted to use a context in the options.origin function as follows:
```
app.use(
  '/api/*',
  cors({
    origin: (origin, c) => {
      return c.env.MY_ALLOW_ORIGIN
    },
  })
)
```

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/middleware/cors/index.ts
+++ b/deno_dist/middleware/cors/index.ts
@@ -1,7 +1,8 @@
+import type { Context } from '../../context.ts'
 import type { MiddlewareHandler } from '../../types.ts'
 
 type CORSOptions = {
-  origin: string | string[] | ((origin: string) => string | undefined | null)
+  origin: string | string[] | ((origin: string, c: Context) => string | undefined | null)
   allowMethods?: string[]
   allowHeaders?: string[]
   maxAge?: number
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @okmr-d !

Thank you for creating the PR.

In this case, we recommend wrapping the `cors()` with another middleware:

```ts
const app = new Hono<{
  Bindings: {
    MY_ALLOW_ORIGIN: string
  }
}>()

app.use('/api/*', async (c, next) => {
  const middleware = cors({
    origin: () => {
      return c.env.MY_ALLOW_ORIGIN
    },
  })
  return middleware(c, next)
})
```

This way is often [used in Basic Auth Middleware](https://hono.dev/getting-started/cloudflare-workers#using-variables-in-middleware).

okmr-d:
Hi @yusukebe !
Thank you for check the PR. My problem is solved.

In the case of the `csrf` middleware, I could use the context. 
https://github.com/honojs/hono/blob/13039a9526d032f854901888473567a4bd8a6c08/src/middleware/csrf/index.ts#L5-L7

It is easier to understand if the arguments of the two functions are the same. How's that?
- `csrf` middleware origin option
- `cors` middleware origin option

yusukebe:
> It is easier to understand if the arguments of the two functions are the same. How's that?

You are right! Making the two functions the same is good. Let's go with it.

yusukebe (approved):
LGTM!

yusukebe:
@okmr-d 

I'll merge this into the "next" branch for releasing the next minor version `4.2.0`. It will be released later. Thanks!

PR #2448, opened 2024-03-31
feat(jwt): add `JwtTokenInvalid` object as `cause` when JWT is invalid

### LINKED ISSUE #2415: Less readable status for JWT mismatch and invalid JWT token

We receive the Status Code ```401 JwtTokenSignatureMismatched: token() signature mismatched``` whenever a token is mismatched during verification.
This long string compromises readability due to long token string that it appends in the status
May be we can strip it down to ```401 JwtTokenSignatureMismatched```

similar issue with ```401 JwtTokenInvalid```

image

I am using the jwt middleware 

```
const app = new Hono()
app.route("/api/v1",router)

```
```
router.use("/*", async (c, next) => {
    const middleware = jwt({secret: c.env.JWT_SECRET})
    return middleware(c, next)
})
```

yusukebe:
Fixes #2415

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/middleware/jwt/index.ts
+++ b/deno_dist/middleware/jwt/index.ts
@@ -61,21 +61,22 @@ export const jwt = (options: {
     }
 
     let payload
-    let msg = ''
+    let cause
     try {
       payload = await Jwt.verify(token, options.secret, options.alg)
     } catch (e) {
-      msg = `${e}`
+      cause = e
     }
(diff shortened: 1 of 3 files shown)


Code-Hex (approved):
Thank you so much!

metrue (approved):
LGTM

Issue #2410, opened 2024-03-24
Expand `basicAuth` and `bearerAuth` middleware


callumthomson:
### What is the feature you are proposing?

I'm interested in bearer token authentication for APIs with Hono. I noticed that Hono has a [built-in middleware](https://hono.dev/middleware/builtin/bearer-auth) for this. Upon reading the docs, I found myself with two unanswered questions:

1. Is a pre-fetched / hardcoded whitelist of valid tokens the only way to use this middleware?
1. If so, what is the actual use case for the built-in bearer token middleware beyond _extremely_ simple applications?

I thought perhaps it was just me, but then I've just been reading through #2316 and I think @tcurdt is talking a lot of sense, and I'd like to see if others have input on this matter.

When I authenticate using bearer tokens, I expect to check the token (somehow) and decide if it's valid or not. Maybe it's encrypted/encoded, or I need to verify a signature, or look up in a DB. Simply having a pre-determined list of tokens that are acceptable is not going to cut it for 99% of applications, and I personally feel that in Hono's official bearer token middleware, such a limitation is bad DX and also does not look good to potential evaluators. Something does not feel right about having such basic functionality missing from an official middleware.

I believe this would be a reasonably simple change to make (which I can contribute), but I should gather opinions first.

Also - all of the above applies to the basic auth middleware too!

Thanks.

bingtsingw:
In the case of `bearerAuth` and `basicAuth`, I found an easy way to extend the middleware:
```
import { decrypt } from 'xxx';
import type { MiddlewareHandler } from 'hono';
import { basicAuth } from 'hono/basic-auth';
import { HTTPException } from 'hono/http-exception';
import { ENV } from '../env';

const decryptWithException = (password: string) => {
  try {
    return decrypt(password, ENV.AUTH_SECRET_BASIC);
  } catch {
    throw new HTTPException(401);
  }
};

export const authBasic = (
  options: { username: string; password: string; realm?: string; hashFunction?: Function },
  ...users: { username: string; password: string }[]
): MiddlewareHandler => {
  return async function authBasic(ctx, next) {
    const auth = basicAuth(
      {
        ...options,
        password: decryptWithException(options.password),
      },
      ...users.map((o) => ({ ...o, password: decryptWithException(options.password) })),
    );

    return auth(ctx, next).catch(() => {
      throw new HTTPException(401);
    });
  };
};
```
The extended middleware accept a encrypted password. 
So I can write code like this. and the actual password is `123456`.
```
[...]

PR #2449, opened 2024-03-31
feat(bearer-auth): add `verifyToken` option

### LINKED ISSUE #2316: `basicAuth` and `bearerAuth` should support dynamic users

Having a fix set of credentials that never changes during the course of the process lifetime is not particular realistic.
With the current implementation I don't see a good way to connect the auth implementations to a database.
They both would need some form of a callback function instead of just passing in all the date in as config.

https://github.com/honojs/hono/blob/4ca5f60def726731b3afbf3f6137f0a4b92d8b27/src/middleware/bearer-auth/index.ts#L8

yusukebe:
This PR will fix #2316 and #2410.

With this PR, it enables passing the function to verify the token in Bearer Middleware.

```ts
app.use(
  '/auth-verify-token/*',
  bearerAuth({
    verifyToken: async (token) => {
      return token === 'dynamic-token'
    },
  })
)
```

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/deno_dist/middleware/bearer-auth/index.ts
+++ b/deno_dist/middleware/bearer-auth/index.ts
@@ -16,7 +16,7 @@ type BearerAuthOptions =
   | {
       realm?: string
       prefix?: string
-      verifyToken: (token: string, c: Context) => Promise<boolean>
+      verifyToken: (token: string, c: Context) => boolean | Promise<boolean>
       hashFunction?: Function
     }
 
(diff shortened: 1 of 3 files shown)


metrue (approved):
LGTM

callumthomson:
LGTM thanks for the work @yusukebe! 👍 

If I had to nitpick I'd say `c` should be the first argument passed to `verifyToken` instead of `token` but it's great as it is!

callumthomson (approved):

PR #2458, opened 2024-04-02
fix(jws): Only import necessary helper (not all helpers)

nicksrandall:
This change prevents bundlers from pulling in all helpers, including the "ssg" helper which is not compatible with Cloudflare workers as it creates a "Request" in the global scope, and only imports the required (adapter) helper.

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/deno_dist/utils/jwt/jws.ts
+++ b/deno_dist/utils/jwt/jws.ts
@@ -1,4 +1,4 @@
-import { getRuntimeKey } from '../../helper.ts'
+import { getRuntimeKey } from '../../helper/adapter/index.ts'
 import { decodeBase64 } from '../encode.ts'
 import type { SignatureAlgorithm } from './jwa.ts'
 import { JwtAlgorithmNotImplemented } from './types.ts'
--- a/src/utils/jwt/jws.ts
+++ b/src/utils/jwt/jws.ts
@@ -1,4 +1,4 @@
-import { getRuntimeKey } from '../../helper'
+import { getRuntimeKey } from '../../helper/adapter'
 import { decodeBase64 } from '../encode'
 import type { SignatureAlgorithm } from './jwa'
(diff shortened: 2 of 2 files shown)


nicksrandall:
An easy way to see this happening is by looking at the output of `esm.sh`. If you navigate to `https://esm.sh/hono@4.2.0/utils/jwt/jws.js`, you'll see the following output:

```js
/* esm.sh - hono@4.2.0/utils/jwt/jws */
import "/v135/hono@4.2.0/es2022/utils/encode.js";
import "/v135/hono@4.2.0/es2022/utils/jwt/types.js";
import "/v135/hono@4.2.0/es2022/utils/jwt/utf8.js";
import "/v135/hono@4.2.0/es2022/accepts.js";
import "/v135/hono@4.2.0/es2022/adapter.js";
import "/v135/hono@4.2.0/es2022/cookie.js";
import "/v135/hono@4.2.0/es2022/css.js";
import "/v135/hono@4.2.0/es2022/factory.js";
import "/v135/hono@4.2.0/es2022/html.js";
import "/v135/hono@4.2.0/es2022/streaming.js";
import "/v135/hono@4.2.0/es2022/testing.js";
import "/v135/hono@4.2.0/es2022/dev.js";
import "/v135/hono@4.2.0/es2022/jwt.js";
import "/v135/hono@4.2.0/es2022/ssg.js";
export * from "/v135/hono@4.2.0/es2022/utils/jwt/jws.js";
```

Here you'll notice that we're pulling in all helpers even thought the "jws" utility only needs the "adapter" helper.

yusukebe:
Hi @nicksrandall !

Thanks for the PR. This should be fixed and looks good. One thing. Could you run `bun run denoify` or `yarn denoify` to generate a file for Deno?

yusukebe:
@nicksrandall 

I've done it by myself. Merging now. Thanks!

PR #2459, opened 2024-04-03
feat(jsx-renderer): pass the context as 2nd arg

yusukebe:
Fixes https://github.com/honojs/honox/issues/132

With this PR, you can get the context object in the component:

```ts
app.use(
  jsxRenderer(
    ({ children }, c) => {
      return (
        <div>
          {children} at {c.req.path}
        </div>
      )
    },
    { docType: false }
  )
)
```

- [x] Add tests
- [x] Run tests


yusukebe:
Hi @usualoma !

Could you review this one?

usualoma:
LGTM 👍

yusukebe:
@usualoma Thanks!

I'll release this feature as a patch release since this is a minor `feat.`

PR #2461, opened 2024-04-03
feat(client): accept a function that provides dynamic headers to hc

### LINKED ISSUE #2460: hc with dynamic headers

It would be very handy if you could pass a function to `hc` that gets called on every request and returns the headers. My use case would be to get the JWT from localStorage on every request because the JWT is managed by another library and can change at any time.

```ts
const client = hc<App>("https://example.com", {
  headers: () => ({
    Authorization: `Bearer ${localStorage.getItem("jwt")}`,
  }),
});
```

This would eliminate the need to create a new client whenever the JWT changes.

niko-gardenanet:

- [x] Add tests
- [x] Run tests

Fixes #2460

### DIFF
--- a/deno_dist/client/client.ts
+++ b/deno_dist/client/client.ts
@@ -35,7 +35,7 @@ class ClientRequestImpl {
     this.url = url
     this.method = method
   }
-  fetch = (
+  fetch = async (
     args?: ValidationTargets & {
       param?: Record<string, string>
     },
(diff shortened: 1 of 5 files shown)


niko-gardenanet:
Fixes #2460

yusukebe (approved):
LGTM!

yusukebe:
Hi @niko-gardenanet 

Thanks for the PR. Looks good. Let's go with it.

I'll merge this into the main right now and release a patch that includes this change.

PR #2477, opened 2024-04-08
fix(ssg): use response header to mark as disabled routes for SSG

### LINKED ISSUE #2476: onlySSG middleware not support Cloudflare worker

4.2.2

Cloudflare Workers

Demo Code:
```typescript
import { Hono } from 'hono'
import { onlySSG } from 'hono/ssg'

const app = new Hono()

app.get('/', onlySSG(), (c) => {
  return c.html('<h2>Hello Hono!</h2>')
})

export default app
```

Run command: `npm run dev`, will got error.

Miniflare start normal without error.

### What do you see instead?

✘ [ERROR] service core:user:pagereader: Uncaught Error: Disallowed operation called within global scope. Asynchronous I/O (ex: fetch() or connect()), setting a timeout, and generating random values are not allowed within global scope. To fix this error, perform this operation within a handler. https://developers.cloudflare.com/workers/runtime-apis/handlers/

    at null.<anonymous> (index.js:1521:29)

✘ [ERROR] MiniflareCoreError [ERR_RUNTIME_FAILURE]: The Workers runtime failed to start. There is likely additional logging output above.

### Additional information

I has install the vite-ssg plugin with this command: `npm i -D vite @hono/vite-ssg`

usualoma:
Fixes #2476, https://github.com/honojs/honox/issues/58

As for #2179, using `c.env` to make the decision was a good change, but the decision in the response header should have been left as is. The response header is used again in this PR.

Since we use response headers in this PR, we can also remove the `new Response` before the request that causes an error in Cloudflare worker, but since it should not be changed in the patch version, we leave it marked as deprecated.

Also, if we can make it so that the response headers are used, the following issue will be fixed.
https://github.com/honojs/honox/issues/58

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/helper/ssg/index.ts
+++ b/deno_dist/helper/ssg/index.ts
@@ -1,2 +1,8 @@
 export * from './ssg.ts'
-export { SSG_DISABLED_RESPONSE, ssgParams, isSSGContext, disableSSG, onlySSG } from './middleware.ts'
+export {
+  X_HONO_DISABLE_SSG_HEADER_KEY,
+  ssgParams,
+  isSSGContext,
+  disableSSG,
+  onlySSG,
+} from './middleware.ts'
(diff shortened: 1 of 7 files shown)


usualoma:
@yusukebe 
Would you please review?

yusukebe (approved):
Looks good to me!

yusukebe:
Hi @usualoma 

Thanks. I'll merge this now.

PR #2479, opened 2024-04-08
fix(client): fix websocket client protocol

naporin0624:
This PR fixes an issue in the Hono client where WebSocket connections could not be established due to the `hc` function's `baseUrl` containing `http` or `https` schemes. Given that the WebSocket API requires URLs to use the `ws` or `wss` schemes, our existing implementation led to errors when attempting WebSocket connections with `http` or `https` URLs.

To address this, we've introduced a utility function, `replaceUrlProtocol`, which converts `http` URLs to `ws` and `https` URLs to `wss` for WebSocket connections. This ensures compatibility with the WebSocket API and resolves the protocol mismatch issue.

This change allows for seamless WebSocket communication within Hono applications by correctly handling the URL schemes required for WebSocket connections.

- [X] Add tests
- [X] Run tests

### DIFF
--- a/src/client/utils.test.ts
+++ b/src/client/utils.test.ts
@@ -42,6 +42,32 @@ describe('replaceUrlParams', () => {
   })
 })
 
+describe('replaceUrlProtocol', () => {
+  it('Should replace http to ws', () => {
+    const url = 'http://localhost'
+    const newUrl = replaceUrlProtocol(url, 'ws')
+    expect(newUrl).toBe('ws://localhost')
+  })
+
+  it('Should replace ws to http', () => {
+    const url = 'ws://localhost'
+    const newUrl = replaceUrlProtocol(url, 'http')
+    expect(newUrl).toBe('http://localhost')
+  })
+
+  it('Should replace https to wss', () => {
+    const url = 'https://localhost'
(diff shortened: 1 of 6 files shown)


src/client/utils.test.ts
```diff
@@ -42,6 +48,20 @@ describe('replaceUrlParams', () => {
   })
 })
 
+describe('replaceUrlProtocol', () => {
+  it('Should replace http to ws', () => {
+    const url = 'http://localhost'
+    const newUrl = replaceUrlProtocol(url, 'ws')
+    expect(newUrl).toBe('ws://localhost')
+  })
+
```
yusukebe, line 57: Can you write a test for the case if the url is `https://localhost`?

yusukebe:
@naporin0624 

Thanks! This will be a good fix. I've left one comment. Check it.

src/client/utils.test.ts
naporin0624: ok!

naporin0624: Thanks for the review.

In addition to https://localhost, I added a test for wss://localhost! https://github.com/honojs/hono/pull/2479/commits/0bffae5199da73298845472ad2b58992c4364d71

yusukebe (approved):
LGTM!

PR #2488, opened 2024-04-09
fix(jwt): Make JWT Header `typ` Field Optional to Enhance Compatibility

naporin0624:
This PR modifies the JWT header verification to make the `typ` field optional. This change is aligned with the JWT specifications as outlined in RFC 7519 Section 5.1, where the `typ` header parameter is optional. This adjustment was necessary to accommodate tokens, such as those decoded from Cloudflare Access's `CF_Authorization`, which do not include the `typ` field, thus previously preventing successful verification.

### Changes Made
- Made the `typ` field in the `TokenHeader` interface optional.
- Updated the `isTokenHeader` function to validate headers with or without the `typ` field.

Reference to the RFC: [JWT RFC 7519 Section 5.1](https://datatracker.ietf.org/doc/html/rfc7519#section-5.1)

The necessity for this change arose during the verification of JWT tokens from Cloudflare Access, where the absence of the `typ` field led to verification issues. By accommodating tokens without the `typ` field, this PR ensures broader compatibility and adherence to the JWT standard.

- [X] Add tests
- [X] Run tests

### DIFF
--- a/deno_dist/utils/jwt/jwt.ts
+++ b/deno_dist/utils/jwt/jwt.ts
@@ -22,7 +22,7 @@ const decodeJwtPart = (part: string): unknown =>
 
 export interface TokenHeader {
   alg: SignatureAlgorithm
-  typ: 'JWT'
+  typ?: 'JWT'
 }
 
 // eslint-disable-next-line
@@ -32,8 +32,7 @@ export function isTokenHeader(obj: any): obj is TokenHeader {
     obj !== null &&
     'alg' in obj &&
     Object.values(AlgorithmTypes).includes(obj.alg) &&
-    'typ' in obj &&
-    obj.typ === 'JWT'
+    (!('typ' in obj) || obj.typ === 'JWT')
   )
 }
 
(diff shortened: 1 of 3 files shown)


deno_dist/utils/jwt/jwt.ts
```diff
@@ -32,8 +32,7 @@ export function isTokenHeader(obj: any): obj is TokenHeader {
     obj !== null &&
     'alg' in obj &&
     Object.values(AlgorithmTypes).includes(obj.alg) &&
-    'typ' in obj &&
-    obj.typ === 'JWT'
+    (!('typ' in obj) || obj.typ === 'JWT')
```
naporin0624, line 35: typ is case-insensitive, but left uppercase because uppercase is recommended for compatibility.

> While media type names are not case sensitive, it is RECOMMENDED that "JWT" always be spelled using uppercase characters for compatibility with legacy implementations.

Code-Hex (approved):
LGTM!

yusukebe:
Thanks! I'll merge this now.

naporin0624:
thanks!

PR #2499, opened 2024-04-11
feat(rpc): Add status code to response type

kosei28:
This PR adds a status code to the response type and allows type filtering in RPC as below.

server.ts
```ts
const app = new Hono().get(
  "/posts/:id",
  zValidator(
    "param",
    z.object({
      id: z.string(),
    })
  ),
  async (c) => {
    const { id } = c.req.valid("param");

    try {
      const post: Post | undefined = await getPost(id);

      if (post !== undefined) {
        return c.json({ post }, 200);
      }

      return c.json({ error: "not found" }, 404);
    } catch (e) {
      return c.json({ error: "internal server error" }, 500);
    }
  }
);

export type AppType = typeof app;
```

client.ts
```ts
const client = hc<AppType>("/api");

const res = await client.posts[":id"].$get({
  param: { id: "1" },
});

if (res.status === 404) {
  const data: { error: string } = await res.json();
  console.log(data.error);
}

if (res.ok) {
  const data: { post: Post } = await res.json();
  console.log(data.post);
}

// { post: Post } | { error: string }
type ResponseType = InferResponseType<(typeof client.posts)[":id"]["$get"]>;

// { post: Post }
type ResponseType200 = InferResponseType<(typeof client.posts)[":id"]["$get"], 200>;
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/client/types.ts
+++ b/deno_dist/client/types.ts
@@ -1,6 +1,7 @@
 import type { UpgradedWebSocketResponseInputJSONType } from '../helper/websocket/index.ts'
 import type { Hono } from '../hono.ts'
-import type { Schema } from '../types.ts'
+import type { Endpoint, Schema } from '../types.ts'
+import type { StatusCode, SuccessStatusCode } from '../utils/http-status.ts'
 import type { HasRequiredKeys } from '../utils/types.ts'
 
 type HonoRequest = (typeof Hono.prototype)['request']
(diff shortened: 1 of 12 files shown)


yusukebe:
@kosei28 Awesome! This will be a great improvement. I'll see the details later.

yusukebe (commented):
How about specifying `200` as the default value? I think the user expects `200` in most cases, and it's `200`, the default value of `c.json()`.

kosei28:
I thought so at first, but c.status() may have set a different status code than the default, and the type and value may not match. It would be nice if the context had a type for the status code and could override that type with c.status(), but this is not possible in TypeScript.

yusukebe:
@kosei28 

> I thought so at first, but `c.status()` may have set a different status code than the default, and the type and value may not match. It would be nice if the context had a type for the status code and could override that type with `c.status()`, but this is not possible in TypeScript.

Understood well! Indeed, it is not possible. 

I'd like to merge this into the branch for the next minor release later. Can you fix the CI error?

kosei28:
I fixed the CI error.

I'm looking forward to its release!

yusukebe:
Hi @kosei28 

Now I've fixed the conflict again to the main. I'll merge this into the "next" branch for the next minor release, 4.3.0. Thanks!

PR #2500, opened 2024-04-12
fix(validator): Default use to `OutputTypeExcludeResponseType` when `InputType` is unknown

nagasawaryoya:
This PR fixed an issue where the type was not shared properly on the Hono Client side when `json` was specified for the `validator` function.

### Changes

- Default use to `OutputTypeExcludeResponseType` when `InputType` is unknown

### Example Scenario and Code Snippet

```ts
const app = new Hono()

const route = app.post(
  '/foo',
  validator('json', () => {
    return {} as {
      id: number
      title: string
    }
  }),
  (c) => {
    return c.json({})
  }
)

type AppType = typeof route
const client = hc<AppType>('http://localhost')

client.foo.$post({
  // Previously, this would result in `unknown` instead of `{ id: number; title: string; }`
  json: {
    id: 1,
    title: 'title'
  }
})
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/validator/validator.ts
+++ b/deno_dist/validator/validator.ts
@@ -34,14 +34,18 @@ export const validator = <
   V extends {
     in: {
       [K in U]: K extends 'json'
-        ? InputType
+        ? unknown extends InputType
+          ? OutputTypeExcludeResponseType
+          : InputType
         : { [K2 in keyof OutputTypeExcludeResponseType]: ValidationTargets[K][K2] }
     }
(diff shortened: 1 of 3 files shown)


yusukebe (approved):
LGTM!

yusukebe:
Hi @nagasawaryoya !

Looks good. I'll merge this into the main now. Thanks for your contribution.

nagasawaryoya:
thanks!

PR #2503, opened 2024-04-13
refactor(trie-router): parentPatterns is updated but never queried

exoego:
It appears that `parentPatterns` is updated but its content is never referenced.

- [ ] Add tests
- [x] Run tests


yusukebe (approved):
LGTM!

yusukebe:
@exoego Thanks!

PR #2510, opened 2024-04-15
fix(client): Allow calling toString and valueOf on the proxy object

### LINKED ISSUE #2509: TypeError: Cannot determine default value of object

4.2.3

Expo

Documenting this for posterity, but will have a PR to fix on the way.

1. Create a new expo app
2. Instantiate a `hono/client` instance (e.g.`const client = hc<AppType>(API_URL)`
3. console.log the client: `console.log(client)`

Logs the client object

### What do you see instead?

`TypeError: Cannot determine default value of object`

This comes from the hermes javascript engine: https://github.com/facebook/hermes/issues/205

What's happening is that when you try to console.log the client, react-native tries to console.log a function. So it calls `client.name.valueOf()` and `client.name.toString()`. But because the client creates proxy objects for methods, it tries to return a instance of `ClientRequestImpl` with the url of `/name/valueOf` and `/name/toString` respectively.

Fix is to catch these cases (and maybe others?) and return a string instead.

### Additional information

_No response_

ibash:
Fixes #2509

When a hono client is instantiated in expo and then console.log(client) is called, an error is thrown:

`TypeError: Cannot determine default value of object`

This is because the hermes javascript engine expects `new String(val)` to always return a string. But the proxy object would return a ClientRequestImpl object instead.

This commit makes a string / function returned in the cases where .toString() and .valueOf() is called on the function.name or function respectively.

Also see https://github.com/facebook/hermes/issues/205

- [X] Add tests
- [X] Run tests (vitest seems to mess up the bun test runner, so comment out that import)

### DIFF
--- a/deno_dist/client/client.ts
+++ b/deno_dist/client/client.ts
@@ -132,8 +132,27 @@ export const hc = <T extends Hono<any, any, any>>(
   baseUrl: string,
   options?: ClientRequestOptions
 ) =>
-  createProxy((opts) => {
+  createProxy(function proxyCallback(opts) {
     const parts = [...opts.path]
+
+    // allow calling .toString() and .valueOf() on the proxy
+    if (parts[parts.length - 1] === 'toString') {
+      if (parts[parts.length - 2] === 'name') {
+        // e.g. hc().somePath.name.toString() -> "somePath"
+        return parts[parts.length - 3] || ''
+      }
+      // e.g. hc().somePath.toString()
+      return proxyCallback.toString()
+    }
+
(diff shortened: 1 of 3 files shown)


yusukebe:
Hi @ibash 

Thank you for the PR. Almost everything is okay. But can you fix the TypeScript errors? I think it's not a problem to use `// @ts-ignore`.

ibash:
@yusukebe woops, sorry about that, fixed!

yusukebe (approved):
LGTM!

yusukebe:
@ibash 

Thanks! I'll merge this now.

PR #2512, opened 2024-04-15
feat(helper/cookie): delete cookie returns the deleted value

### LINKED ISSUE #2321: Session flash?

I don't have a strong opinion on this and I'm fine with the way Hono currently works.

However, I'm wondering if something for making flash messages easier would make sense to be in the core? For example, to show a message after a successful POST request and a redirect.

Here are three options that I see:

1. Create a `setFlash` and `getFlash`
2. Make `deleteCookie` return the deleted value
3. No changes and keep Hono core small

What do you think?

sor4chi:

- [x] Add tests
- [x] Run tests

resolve #2321

### DIFF
--- a/deno_dist/helper/cookie/index.ts
+++ b/deno_dist/helper/cookie/index.ts
@@ -118,6 +118,8 @@ export const setSignedCookie = async (
   c.header('set-cookie', cookie, { append: true })
 }
 
-export const deleteCookie = (c: Context, name: string, opt?: CookieOptions): void => {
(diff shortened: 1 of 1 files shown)


yusukebe (approved):
LGTM!

yusukebe:
Hi @sor4chi 

Looks good to me! We can merge this right now, but I'm wondering if it should be included in a patch release or a minor release. Please wait a bit.

yusukebe:
Hi @sor4chi !

Now, I'll merge this into the "next" branch. Thanks!

PR #2514, opened 2024-04-16
feat(bearer-auth): add `headerName` option

eliasbrange:
Adds the possibility to use custom headers, such as `X-Auth` or `X-Access-Token` instead of `Authorization` with the `bearer-auth` middleware.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/bearer-auth/index.ts
+++ b/src/middleware/bearer-auth/index.ts
@@ -5,17 +5,20 @@ import { timingSafeEqual } from '../../utils/buffer'
 
 const TOKEN_STRINGS = '[A-Za-z0-9._~+/-]+=*'
 const PREFIX = 'Bearer'
+const HEADER = 'Authorization'
 
 type BearerAuthOptions =
   | {
       token: string | string[]
       realm?: string
       prefix?: string
+      headerName?: string
       hashFunction?: Function
     }
   | {
       realm?: string
       prefix?: string
+      headerName?: string
       verifyToken: (token: string, c: Context) => boolean | Promise<boolean>
       hashFunction?: Function
     }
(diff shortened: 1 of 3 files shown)


yusukebe:
Hi @eliasbrange 

Thanks for the PR. But I think we should consider whether we make it possible to change the header name.

@callumthomson What do you think about this matter?

callumthomson:
@yusukebe I suppose it is somewhat reasonably common to want to use a different header name for providing a bearer token for authentication. I guess it depends on where you draw the line between 'batteries included' and 'extend it yourself'.

From a web standards perspective, according to [RFC 6750](https://datatracker.ietf.org/doc/html/rfc6750#section-2.1), the header name _can_ be changed, but it is clear which one is considered standard.
> Clients SHOULD make authenticated requests with a bearer token using the "Authorization" request header field with the "Bearer" HTTP authorization scheme.  Resource servers MUST support this method.

eliasbrange:
I might agree on pulling back this one. For my use case, I would most likely write my own middleware anyway to get more control over caching JWKS and such. :)

yusukebe:
@callumthomson Thank you for the comment!

>  I guess it depends on where you draw the line between 'batteries included' and 'extend it yourself'.

Indeed, you are right. For this case, I prefer "batteries included." This means we can accept this feature to enable the change of the header name.

@eliasbrange So, shall we go with it? I think there are some points to be reviewed, I'll comment.

src/middleware/bearer-auth/index.ts
```diff
@@ -5,17 +5,20 @@ import { timingSafeEqual } from '../../utils/buffer'
...
 const TOKEN_STRINGS = '[A-Za-z0-9._~+/-]+=*'
 const PREFIX = 'Bearer'
+const HEADER = 'Authorization'
 
 type BearerAuthOptions =
   | {
       token: string | string[]
       realm?: string
       prefix?: string
+      header?: string
```
yusukebe, line 15: I like the option name `header`, but it may be too *general*. So how about using `headerName` or something else?

eliasbrange: I agree. I changed it to `headerName`.

yusukebe (approved):
LGTM!

yusukebe:
Hi @eliasbrange 

Now, I'll merge this into the "next" branch for the next minor release `v4.3.0` which will be shipped soon. Thanks for your contribution.

PR #2515, opened 2024-04-16
feat: expose serve-static builder

cometkim:
This allows accessing `middleware/serve-static` from third-party

### Use case

I run my own handler based on Cloudflare R2 to avoid the limitations of Cloudflare KV / Pages. And I found that most of the code I needed there was already built into hono.

So what I expect is like:

```ts
import { Hono } from 'hono';
import { serveStatic as makeServe } from 'hono/base-serve-static';

const app = new Hono<Env>();

const serveContent = makeServe<Env>({
  getContent: async (path, c) => {
    // retrieve
    let response = await cache.match(c.req.raw);

    // ...get R2 object
    if (!response) {
      let obj = await c.env.CONTENT.get(path);
      if (obj) {
        response = obj.body;
        c.executionCtx.waitUntil(cache.put(c.req.raw, obj.body));
      }
    }

    c.body(response);
  },
});

app.use(serveContent);

export default app;
```

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -279,7 +279,7 @@
       "import": "./dist/helper/factory/index.js",
       "require": "./dist/cjs/helper/factory/index.js"
     },
-    "./base-serve-static": {
+    "./serve-static": {
       "types": "./dist/types/middleware/serve-static/index.d.ts",
       "import": "./dist/middleware/serve-static/index.js",
       "require": "./dist/cjs/middleware/serve-static/index.js"
@@ -480,7 +480,7 @@
       "factory": [
         "./dist/types/helper/factory/index.d.ts"
       ],
-      "base-serve-static": [
+      "serve-static": [
         "./dist/types/middleware/serve-static"
       ],
       "cloudflare-workers": [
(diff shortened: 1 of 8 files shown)


yusukebe:
Hi @cometkim 

This is good! But please wait until I look over the details!

cometkim:
@yusukebe added a type definition rather than `any` as it would be a public interface.

And made some changes to allow a situation where users needed to create the Response object themselves. (e.g. using with Browser/Cloudflare's Cache API)

package.json
```diff
@@ -279,6 +279,11 @@
       "import": "./dist/helper/factory/index.js",
       "require": "./dist/cjs/helper/factory/index.js"
     },
+    "./base-serve-static": {
```
yusukebe, line 282: Simply, I think `./base-serve` is better than `./base-serve-static`.

```ts
import { serveStatic } from 'hono/serve-static 
```

yusukebe (approved):
LGTM!

yusukebe:
@cometkim Thanks!

PR #2516, opened 2024-04-16
feat: expose built-in MIME types

cometkim:
This allows users reuse builtin MIME types while adding some custom.

```js
import { baseMimes, getMimeType } from 'hono/utils/mime';

const mimes = {
  ...baseMimes,
  'my-own-ext': 'text/plain',
};

getMimeType('base.my-own-ext', mimes)
```

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/utils/mime.ts
+++ b/src/utils/mime.ts
@@ -19,7 +19,7 @@ export const getExtension = (mimeType: string): string | undefined => {
   }
 }
 
-const baseMimes: Record<string, string> = {
+export const baseMimes: Record<string, string> = {
   aac: 'audio/aac',
   avi: 'video/x-msvideo',
   avif: 'image/avif',
--- a/src/utils/mime.ts
+++ b/src/utils/mime.ts
@@ -19,7 +19,8 @@ export const getExtension = (mimeType: string): string | undefined => {
   }
 }
 
-export const baseMimes: Record<string, string> = {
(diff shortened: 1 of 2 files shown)


cometkim:
I recognize that it can be accessed indirectly by calling `getMimeType` twice, and it's way more performant than copying all values per each request.

However, copying once in advance is obviously 2x better than calling twice.

src/utils/mime.ts
```diff
@@ -19,7 +19,7 @@ export const getExtension = (mimeType: string): string | undefined => {
   }
 }
 
-const baseMimes: Record<string, string> = {
+export const baseMimes: Record<string, string> = {
```
yusukebe, line 22: I think when it exports it, naming `mimes` is better than `baseMimes`. 

```ts
const mimes: Record<string, string> = {
  aac: 'audio/aac',
  // ...
```

The prefix `base` is not necessary. It's easy to understand for users without the `base`.

yusukebe:
Hi @cometkim !

Sorry to be late.

This sounds good. I've left a comment. Check it!

If this PR is OK, I'll merge this into the "next" branch for the next minor release.

yusukebe (approved):
LGTM!

yusukebe:
@cometkim 

Thanks! Merging now.

PR #2521, opened 2024-04-18
refactor(adapter/aws): Optimize multiple call of same conditions with polymorphism 

exoego:
### Summary

- Conditional logics are capsulated in the subclasses of `EventProcessor<E extends LambdaEvent>`.
- Those classes are stateless, so singleton is used to avoid allocation.
- Redundant calling `isProxyEventV2` is optimized, now only one time per event.
- Not in this PR, but this refactoring may also allow some room for optimization for particular Lambda events.

- [ ] Add tests
- [x] Run tests


yusukebe (approved):
LGTM!

yusukebe:
@exoego 

Merging now. Thanks!

benmhara (approved):

PR #2524, opened 2024-04-18
fix(jsx/dom): remove lookbehind assertion in event regexp

### LINKED ISSUE #2517: Unable to use honox-island before iOS 16.4 due to Safari is not supoorted lookbehind regex.

4.0.5 (latest honox uses this version)

iOS 16.3 or lower

Prepare a iOS 16.3 or lower iOS device.

Access the website which created with honox island component (e.g. honox-samples).

Errors occurred in browser console:
```
Unhandled Promise Rejection: SyntaxError: Invalid regular expression: invalid group specifier name
```

I haven't tested it without honox. Sorry.

No error occurred.

### What do you see instead?

hono/jsx uses lookbehind regex in JavaScipt.

https://github.com/honojs/hono/blob/13c3156c67414fdb43ce30d6ce74f3ab1c349305/src/jsx/dom/render.ts#L93 

It is not supported by iOS 16.3 or lower.

https://caniuse.com/js-regexp-lookbehind

To resolve this issue, we need to rewrite regex without lookbehind pattern as I think.

### Additional information

I don't know about hono's compatibility policy for older devices or OS, if hono decides that this issue will not be resolved, I will follow your decision.👍

Everything else seemed to be working fine on iOS 15.4 I tested.

I think hono will be more compatible with older devices.

usualoma:
fixes #2517

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/jsx/dom/render.ts
+++ b/deno_dist/jsx/dom/render.ts
@@ -90,7 +90,7 @@ let nameSpaceContext: JSXContext<string> | undefined = undefined
 const isNodeString = (node: Node): node is NodeString => Array.isArray(node)
 
 const getEventSpec = (key: string): [string, boolean] | undefined => {
-  const match = key.match(/^on([A-Z][a-zA-Z]+?)((?<!Pointer)Capture)?$/)
+  const match = key.match(/^on([A-Z][a-zA-Z]+?(?:PointerCapture)?)(Capture)?$/)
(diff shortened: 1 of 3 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@usualoma 

Thanks! Merging now.

PR #2532, opened 2024-04-19
fix(client): Don't show `$ws` when not used WebSockets

nakasyou:
In client, IntelliSense always shows `(route).$ws` property.
I think client doesn't have to show `(route).$ws` property when route doesn't use WebSocket helper.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/client/types.ts
+++ b/deno_dist/client/types.ts
@@ -33,14 +33,13 @@ export type ClientRequest<S extends Schema> = {
         : {}
       : {}
   ) => URL
-} & {
-  // WebSocket
-  $ws: S['$get'] extends { input: { json: UpgradedWebSocketResponseInputJSONType } }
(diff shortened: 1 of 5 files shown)


yusukebe:
Hi @nakasyou !

Super cool! I love it.

I want to merge this, but do you have any idea how to write tests? It's better to write a test, though it seems a little bit difficult because it is Types of matter.

nakasyou:
@yusukebe
> I want to merge this, but do you have any idea how to write tests? It's better to write a test, though it seems a little bit difficult because it is Types of matter.

I searched about it, then I found that Vitest has a way of testing types. So I made test using this.

Reference:
- https://www.totaltypescript.com/how-to-test-your-types
- https://vitest.dev/guide/testing-types.html

yusukebe (approved):
LGTM!

yusukebe:
@nakasyou 

Awesome! That's best.

Let's ship it. Thanks!

PR #2543, opened 2024-04-23
[WIP] Update secure-headers align with best current practice

### LINKED ISSUE #2540: default referrer-policy for secure-headers middleware

# Default Referrer-Policy for secure-headers middleware

Currently the default referrer-policy is `no-referrer`.

https://github.com/honojs/hono/blob/main/src/middleware/secure-headers/index.ts#L73

When referrer-policy sets no-referrer, it also hides not only `referer` header but also `origin` header.

https://fetch.spec.whatwg.org/#origin-header

> - "no-referrer"
>    - Set serializedOrigin to `null`.

`Origin` header is very important for mitigating CSRF attacks.
So making it `null` can make the application less secure, since the application server can't make sure it really comes from the same origin.

What makes it secure is not hiding every referrer, but avoiding leaking much information for cross origin servers. There's no reason to avoid sending the referrer value to the same origin server.

## What should be the default

There are many choices on Referrer-Policy but I prefer using `strict-origin-when-cross-origin`.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy#strict-origin-when-cross-origin_2

It minimizes information leaks and keeps sending origin for keep `origin` header based csrf guards works.

`same-origin` seems an alternative for securerer headers, but it hides sending origin to cross origin. In this case, it can cause some problems if your application interacts with cross origin services.

That's the reason why `strict-orign-when-cross-orign` is the default policy for modern browsers now.

but it's one of the practices to explicitly set `strcit-origin-when-cross-orign` from middleware to make sure it is enabled.

Jxck:
closes: #2540

update headers in secure-headers middleware align with best practices.

1. `Referrer-Policy` to `strict-origin-when-cross-origin`
as I filed #2540 , `no-referer` also make `origin` header to `null`.
`origin` header is important for avoid unexpected api call from browser (which includes CSRF).
there are no reason to hide referrer from same-origin, so change it to reasonable setting.
`strict-origin-when-cross-origin` is currently default for modern browsers, but explicitly set it is fine.

see also: https://cheatsheetseries.owasp.org/cheatsheets/HTTP_Headers_Cheat_Sheet.html#referrer-policy

2. don't use x-xss-protection
 
x-xss-protection is not fully standardized, deprecated, not recommended header. xss filter is depreacted on all browser so there are no reason to serve it even if it intends to disable since there is no guarantee `0` behaves as expected. use CSP instead.

see also: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection

3. user QPAC registered value. it compress this header field in 1byte on QUIC

there are no reason to shorten `Strict-Transport-Security` 's max-age.
so what is the appropriate value ? => anything long value is fine basically.
in this case you can see the QPACK static table.
https://www.rfc-editor.org/rfc/rfc9204.html#name-static-table-2
the basic values are defined in this table, and qpack will compress it in 1byte on quic transport.

```
56	strict-transport-security	max-age=31536000
57	strict-transport-security	max-age=31536000; includesubdomains
58	strict-transport-security	max-age=31536000; includesubdomains; preload
```

57 is fine here.

4. X-Frame-Options are replaced by frame-ancestors in CSP.
 
it's widely supported in browsers and allow more appropriate directive to customize.

## And more.

Basically, "**kichen-sink**" headers are kind of bad practice.
From point of view of that, the headers below are also ignorable for me.
since only for old browser, or the Risks are not clear (secure for who ?).

`X-Download-Options: noopen`: it's for deprecated IE
`X-Permitted-Cross-Domain-Policies: none'`: it's for deprecated Flash / Silverlight

I don't know it's really required. but don't have strong opinion to delete thought.

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/middleware/secure-headers/index.ts
+++ b/src/middleware/secure-headers/index.ts
@@ -77,7 +77,6 @@ const HEADERS_MAP: HeadersMap = {
   xDownloadOptions: ['X-Download-Options', 'noopen'],
   xFrameOptions: ['X-Frame-Options', 'SAMEORIGIN'],
   xPermittedCrossDomainPolicies: ['X-Permitted-Cross-Domain-Policies', 'none'],
-  xXssProtection: ['X-XSS-Protection', '0'],
 }
 
 const DEFAULT_OPTIONS: SecureHeadersOptions = {
--- a/src/middleware/secure-headers/index.ts
+++ b/src/middleware/secure-headers/index.ts
@@ -66,16 +66,16 @@ type HeadersMap = {
 }
 
 const HEADERS_MAP: HeadersMap = {
+  contentSecurityPolicy: ['Content-Security-Policy', "frame-ancestors 'none'"],
   crossOriginEmbedderPolicy: ['Cross-Origin-Embedder-Policy', 'require-corp'],
-  crossOriginResourcePolicy: ['Cross-Origin-Resource-Policy', 'same-origin'],
   crossOriginOpenerPolicy: ['Cross-Origin-Opener-Policy', 'same-origin'],
+  crossOriginResourcePolicy: ['Cross-Origin-Resource-Policy', 'same-origin'],
   originAgentCluster: ['Origin-Agent-Cluster', '?1'],
   referrerPolicy: ['Referrer-Policy', 'strict-origin-when-cross-origin'],
   strictTransportSecurity: ['Strict-Transport-Security', 'max-age=31536000; includesubdomains'],
   xContentTypeOptions: ['X-Content-Type-Options', 'nosniff'],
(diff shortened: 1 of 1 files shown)


watany-dev:
This is just an idea, but to avoid breaking changes, it might be good to create and operate options like "useHelmetDefault" or "useHonoDefault".

- v4.x
```ts
// To select a default value that includes this change
app.use(
  '*',
  secureHeaders({
    useHonoDefault: true // deprecate v5.0
  })
)
```

- 5.x
```ts
// To select a current default value
app.use(
  '*',
  secureHeaders({
    useExpressDefault: true
  })
)
```

PR #2545, opened 2024-04-23
fix(cloudflare-workers): do not buffer content

cometkim:

- [ ] Add tests
- [x] Run tests


yusukebe (approved):
LGTM!

yusukebe:
@cometkim 

Thanks. Indeed, it should be a stream. Thanks.

PR #2548, opened 2024-04-23
fix(types): useState() has no type errors without arguments

### LINKED ISSUE #2547: useState() does not require initialState

I'm developing an [embla-carousel](https://github.com/davidjerleke/embla-carousel) plugin for hono/jsx.
And I found that useState() in hono/jsx requires initialState.
```ts
import { useState } from 'hono/jsx'

const [count, setCount] = useState() // <= error
```
But that in react doesn't need an argument.
```ts
import { useState } from 'react'

const [count, setCount] = useState() // <= no error
```
I would like to allows `undefined` to improve compatibility with react.
It's probably just the problem for the type.

ryuapp:
fix #2547

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/deno_dist/jsx/hooks/index.ts
+++ b/deno_dist/jsx/hooks/index.ts
@@ -138,9 +138,15 @@ const setShadow = (node: Node) => {
   ;(node as any).s?.forEach(setShadow)
 }
 
-export const useState = <T>(initialState: T | (() => T)): [T, UpdateStateFunction<T>] => {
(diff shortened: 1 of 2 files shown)


usualoma:
@ryuapp Thank you.
LGTM!

yusukebe (approved):
LGTM!

yusukebe:
@ryuapp 

Thanks. Merging right now.

PR #2549, opened 2024-04-23
chore(package.json): prettier use cache

ryuapp:
It takes more than 5s to format codes with Prettier.
Using the cache, it will be less than 1s from the second time onwards.

https://prettier.io/docs/en/cli.html#--cache

- [ ] Add tests
- [ ] Run tests


yusukebe:
Wow cooool! This is awesome! Thanks.

Issue #2554, opened 2024-04-24
newResponse overwrites existing headers (e.g. set-cookie) on the local response (c.res)


ItsWendell:
### What version of Hono are you using?

4.2.7

### What runtime/platform is your app running on?

Cloudflare Workers

### What steps can reproduce the bug?

```
 // file: context.test.ts
 
  it('Should extend the cookie headers when res already has cookie headers', () => {
    c.res.headers.append('set-cookie', 'foo=bar; Path=/')
    setCookie(c, 'foo2', 'bar2', { path: '/' })

    const res = c.json({ message: 'Hello' })
    
    const cookies = res.headers.getSetCookie()

    expect(cookies.includes('foo=bar; Path=/')).toBe(true)
    expect(cookies.includes('foo2=bar2; Path=/')).toBe(true)
  })
  ```

### What is the expected behavior?

In my opinion, when an existing header with set-cookie exists in the response, this header shouldn't be overwritten, but added too.

file: src/context.ts:384
```
    if (this.#res) {
      this.#res.headers.forEach((v, k) => {
        // NOTE: Maybe only append like in `set res = (_res) ...`
        this.#headers?.append(k, v)
      })
      setHeaders(this.#headers, this.#preparedHeaders)
    }
 ```

### What do you see instead?

I see the set-cookie header being completely overwritten.

### Additional information

I think a discussion is needed whether we actually want to fix this or not, and what the impact might be on existing projects.

In my case, I want to be able to share the `c.res` to a library which might manipulate it's cookies, but these will be overwritten if I later set my own cookies.

ItsWendell:
As mentioned in additional information, a way for me to fix this after learning about some of the code in `context.ts` would be to do the following:

```ts

  const myLibrary = (res?: Response = new Response()) => {
    res.headers.append('set-cookie')
    return res;
  }

  it('Should extend the cookie headers when res already has cookie headers', () => {
    c.res = myLibrary(c.res);
    setCookie(c, 'foo2', 'bar2', { path: '/' })

    const res = c.json({ message: 'Hello' })
    
    const cookies = res.headers.getSetCookie()

    expect(cookies.includes('foo=bar; Path=/')).toBe(true)
    expect(cookies.includes('foo2=bar2; Path=/')).toBe(true)
  })
  
  ```

ItsWendell:
Fixed and merged https://github.com/honojs/hono/pull/2555

PR #2555, opened 2024-04-24
fix(context): keep response cookies in newResponse instead of overwriting

ItsWendell:
Potential fix for https://github.com/honojs/hono/issues/2554

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/context.test.ts
+++ b/src/context.test.ts
@@ -249,14 +249,10 @@ describe('Context header', () => {
   it('Should keep previous cookies in response headers', () => {
     c.res.headers.append('set-cookie', 'foo=bar; Path=/')
     setCookie(c, 'foo2', 'bar2', { path: '/' })
-
     const res = c.json({ message: 'Hello' })
-
     const cookies = res.headers.getSetCookie()
-
-    console.debug('test cookies', cookies)
-
-    expect(cookies.length).toBe(2)
+    expect(cookies.includes('foo=bar; Path=/')).toBe(true)
+    expect(cookies.includes('foo2=bar2; Path=/')).toBe(true)
   })
 })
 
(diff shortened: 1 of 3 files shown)


src/context.test.ts
```diff
@@ -244,6 +245,19 @@ describe('Context header', () => {
...
+
+  it('Should keep previous cookies in response headers', () => {
+    c.res.headers.append('set-cookie', 'foo=bar; Path=/')
+    setCookie(c, 'foo2', 'bar2', { path: '/' })
+
+    const res = c.json({ message: 'Hello' })
+
+    const cookies = res.headers.getSetCookie()
+
+    console.debug('test cookies', cookies)
```
yusukebe, line 257: This line is not needed.

src/context.test.ts
```diff
@@ -244,6 +245,19 @@ describe('Context header', () => {
...
+
+    expect(cookies.length).toBe(2)
```
yusukebe, line 259: It's better to add tests to check the values as you wrote in your issue:

```ts
expect(cookies.includes('foo=bar; Path=/')).toBe(true)
expect(cookies.includes('foo2=bar2; Path=/')).toBe(true)
```

src/context.test.ts
```diff
@@ -244,6 +245,19 @@ describe('Context header', () => {
     c.res = res
     expect(c.res.headers.getSetCookie().length).toBe(2)
   })
+
+  it('Should keep previous cookies in response headers', () => {
+    c.res.headers.append('set-cookie', 'foo=bar; Path=/')
+    setCookie(c, 'foo2', 'bar2', { path: '/' })
+
```
yusukebe, line 252: You have to remove these break likes.

yusukebe:
Hi @ItsWendell !

The issue should be fixed and this PR is good! I've left some comments.

ItsWendell:
@yusukebe appreciate it, I've cleaned up the tests and resolved your comments!

yusukebe (approved):
LGTM!

PR #2562, opened 2024-04-26
feat: add `forMiddleware` types to Input

nakasyou:
I added a `forMiddleware` prop to `Input`.
In WebSocket helper, helper uses `json` property for telling data to client, but I think it isn't smart because websocket helper actually doesn't use JSON.

This PR adds properties for various helpers and middleware to tell data to clients.

In creating this PR, I enabled that `ToSchema` receives `Input` instead of `Input['in']`.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/client/types.ts
+++ b/deno_dist/client/types.ts
@@ -1,4 +1,3 @@
-import type { UpgradedWebSocketResponseInputJSONType } from '../helper/websocket/index.ts'
 import type { Hono } from '../hono.ts'
 import type { Schema } from '../types.ts'
 import type { HasRequiredKeys } from '../utils/types.ts'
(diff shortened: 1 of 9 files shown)


yusukebe:
Hi @nakasyou !

This is an interesting feature. Please let us know another use case, not only the WebSocket helper case.

nakasyou:
@yusukebe

For example, you can create your protocol such as type-safe SSE:
```ts
// server
import { typedSSE, sseClient } from 'hono-typed-sse'
import { Hono } from 'hono'

const app = new Hono().get('/sse', typedSSE<Data>(...))

// client

const sse = sseClient(client.sse)
```

In this case, this package can use `forMiddleware` namespace.

yusukebe:
@nakasyou 

Cool! I'll use it to know the feeling of using it.

yusukebe:
Hi @nakasyou 

Sorry for the delay. This is good. We can merge this.

But one thing. I wonder if the naming `forMiddleware` is not so good because it's not just for middleware; we have to name it with a more general name. My suggestion is **`stash`**. There are other words, but `stash` is proper to mean temporary places to store the types.

Any thoughts?

nakasyou:
Hi @yusukebe, thank you for trying.
> But one thing. I wonder if the naming forMiddleware is not so good because it's not just for middleware; we have to name it with a more general name. My suggestion is stash. There are other words, but stash is proper to mean temporary places to store the types.
>
> Any thoughts?

I agreed. I think so. So I changed.

Also, I changed `stash` from `unknown` to `Record<string, unknown>`

yusukebe:
@nakasyou 

Thank you!

However, we have other very interesting PRs. I'll consider merging this feature with them.

Anyway, can you see #2581 ?

nakasyou:
#2588

PR #2553, opened 2024-04-24
feat(jsx/dom): improve compatibility with React

usualoma:
#2508

### What is this?

Improve compatibility with React and make the Getting started level code work for the following UI components

* [react-toastify](https://github.com/fkhadra/react-toastify)
* [spinners-react](https://github.com/adexin/spinners-react)
* [Radix UI Primitives](https://www.radix-ui.com/primitives)

https://github.com/honojs/hono/assets/30598/c86423d3-90d0-4296-8ae4-ff90a5432188

by https://github.com/usualoma/hono-react-compat-demo

### Simple usage 

tsconfig.json
https://github.com/usualoma/hono-react-compat-demo/blob/main/tsconfig.json#L22-L28

vite.config.ts
https://github.com/usualoma/hono-react-compat-demo/blob/main/vite.config.ts#L5-L8

#### You can use `/hono/jsx`

It is better to use `hono/jsx/dom` in terms of performance, but simply specifying `hono/jsx` works just as well.

tsconfig.json
```json
   "paths": {
      "react": ["./node_modules/hono/dist/jsx"],
      "react-dom": ["./node_modules/hono/dist/jsx/dom"],
    },
```

vite.config.ts
```ts
    alias: {
      'react': 'hono/jsx',
      'react-dom': 'hono/jsx/dom',
    },
```

### New staffs

The following React compatible staffs are now exported from `hono/jsx` and `hono/jsx/dom`

* `createRef`
* `forwardRef`
* `useImperativeHandle`
* `useSyncExternalStore`

These staff are also exported from `hono/jsx/dom`.

* `flushSync`
* `createPortal`

### Children

Implement `Children` API

https://react.dev/reference/react/Children#alternatives

### CSS custom properties in style attribute

CSS custom properties can now be exported.

```tsx
<div style={{"--my-custom-property": "10px"}}>{children}</div>
```

### defaultProps

`defaultProps` is an old API, but there seems to be a library that uses it, so it is supported

https://legacy.reactjs.org/docs/react-component.html#defaultprops

### Bug fixes

Various bugs have been fixed, such as the behaviour when deleting Elements.

### Compatibility with hono@v4.2.7

There are no breaking changes from v4.2.7. However, the internal structure has been significantly refactored and there is a possibility of unintended incompatible behaviour, so please check the behaviour of your application when updating.

### Author should do the followings, if applicable

- [x] Add tests
- [x] Run tests
- [x] `yarn denoify` to generate files for Deno

PR #2563, opened 2024-04-27
fix(jsx): preserve the state of element even if it is repeatedly evaluated by children

usualoma:
There was a bug where the hook state was not maintained when passed as `children`.
This is not a specification change, but a fix to the original expected behaviour, so a patch version release would be acceptable, but it might cause changes in the results in existing applications, so it is better to include it in a minor version update with #2553.

- [x] Add tests
- [x] Run tests


yusukebe:
Thank you @usualoma !

I'll merge this into the "next" branch, which I will create later. The next minor version will be shipped soon!

yusukebe (approved):
LGTM!

PR #2573, opened 2024-04-30
feat(factory): add `createApp()`

yusukebe:
This PR introduces `createApp()` in `hono/factory` as an experimental feature. If you use this method with `createFactory()`, you can avoid redundancy in the definition of `Env` types.

Before:

```ts
const app = new Hono<Env>()

const mw = createMiddleware<Env>(async(c, next) => {
  await next()
})
```

With this PR:

```ts
const factory = createFactory<Env>()

const app = factory.createApp()

const mw = factory.createMiddleware(async (c, next) => {
  await next()
})
```

Using the "Simple Factory" pattern with the Factory helper, there is no need to write type definitions in `global.d.ts` with possible type conflicts. Also, you don't have to have `Env` definitions in multiple places.

```ts
// factory-with-db.ts
type Env = {
  Bindings: {
    MY_DB: D1Database
  }
  Variables: {
    db: DrizzleD1Database
  }
}

export default createFactory<Env>({
  initApp: (app) => {
    app.use(async (c, next) => {
      const db = drizzle(c.env.MY_DB)
      c.set('db', db)
      await next()
    })
  },
})
```

```ts
// crud.ts
import factoryWithDB from './factory-with-db'

const app = factoryWithDB.createApp()

app.post('/posts', (c) => {
  c.var.db.insert
})
```

In the future, I recommend the use of this "Simple Factory" pattern for larger applications, such as those using HonoX.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/helper/factory/index.ts
+++ b/deno_dist/helper/factory/index.ts
@@ -1,7 +1,28 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
+import { Hono } from '../../hono.ts'
 import type { Env, H, HandlerResponse, Input, MiddlewareHandler } from '../../types.ts'
 
+type InitApp<E extends Env = Env> = (app: Hono<E>) => void
+
 export class Factory<E extends Env = any, P extends string = any> {
+  private initApp?: InitApp<E>
+
+  constructor(init?: { initApp?: InitApp<E> }) {
+    this.initApp = init?.initApp
+  }
+
+  /**
+   * @experimental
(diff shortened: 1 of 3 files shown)


yusukebe:
Hi, @usualoma What do you think of this? I'd like to recommend the Factory pattern.

usualoma:
@yusukebe 👍 We will be able to define an app concisely!

yusukebe:
@usualoma 

Thanks!

PR #2582, opened 2024-05-01
chore(pr_template): Use Bun instead of yarn

nakasyou:
I feel strange when I create PR.

This project recommends Bun by default, so I think PR Template should use `bun denoify` instead of `yarn denoify`.

- [ ] Add tests
- [x] Run tests


yusukebe:
@nakasyou 

Exactly. It should be `bun`!

PR #2585, opened 2024-05-01
fix: avoid duplicating already set headers

### LINKED ISSUE #2550: Regression caused by multiHeaderValue handling on Lambda

4.2.5

AWS

https://github.com/honojs/hono/pull/2494 introduced the handling of multiHeaderValues. However, this seemed to have caused header duplication. On AWS, using API Gateway, a header is put both in `headers` and in `multiValueHeaders` when using proxy integration (even single value ones). The handling introduced in #2494 adds both to the headers array, causing duplicates. As a result, header values are concatenated with commas.

Example output from API Gateway test console:
```
Wed Apr 24 00:49:47 UTC 2024 : Endpoint request body after transformations: {"resource":"/{proxy+}","path":"/follows","httpMethod":"GET","headers":{"test":"123"},"multiValueHeaders":{"test":["123"]},"queryStringParameters":null,"multiValueQueryStringParameters":null,"pathParameters":{"proxy":"follows"},"stageVariables":null,"requestContext":
```
As you can see, the same header is in both collections, even though provided only once.

No header value duplication.

### What do you see instead?

Duplicate header values:

```js
c.req.Header("test") // value: '123,123'
```

### Additional information

_No response_

Amorim33:
fix #2550

### Description

This pull request avoids duplicating already set headers; this was happening because API Gateway V1 event has "multiValueHeaders" for headers with multiple values.

Currently, Hono AWS Lambda adapter appends any value from these "multiValueHeaders" in the current Header instance. `This leads to duplicated values when a key is already defined`.

### How
- Added a condition to only append a value from "multiValueHeaders" when the current key is not already defined or the current value is not already included;
- added unit tests for the createRequest method from EventProcessor to make sure that the Request is adapted correctly, headers are not duplicated and to avoid future regressions;
> Tests are following the arrange / act / assert pattern.
- `getProcessor` function was exported for unit testing purposes;
- Added a comment because the conditions became a little complex.

### References
https://docs.aws.amazon.com/lambda/latest/dg/services-apigateway.html
https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html

- [x] Add tests
- [x] Run tests

### Extra
Thanks to @jean-leonco that found the bug in @BemteviSeguros API and @kbrandwijk that opened the issue.

Please let me know if this aligns with the library goals and if the code follows the repository recommendations.

Thanks in advance.

### DIFF
--- a/src/adapter/aws-lambda/handler.test.ts
+++ b/src/adapter/aws-lambda/handler.test.ts
@@ -1,4 +1,5 @@
-import { isContentTypeBinary, isContentEncodingBinary } from './handler'
+import type { LambdaEvent } from './handler'
+import { getProcessor, isContentEncodingBinary, isContentTypeBinary } from './handler'
 
 describe('isContentTypeBinary', () => {
   it('Should determine whether it is binary', () => {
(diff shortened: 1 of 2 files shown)


yusukebe:
Thanks @Amorim33 !

@watany-dev Could you review this?

watany-dev:
I have tried to fix the problem at hand for a short time, but it appears that the header rewrite does not work if I just change the createRequest. In fact, I can see that the test is still failing with this fix.

Amorim33:
Hello @watany-dev , can you please run the tests locally?
Or rerun the action.

I don't know why they are breaking on CI.

image

Amorim33:
Hello @watany-dev!

Sorry for this, I didn't realize that tests were not passing when running `test:all`.

This was happening because of other LambdaEvent related global definitions, the "content-type" header was missing in my test. Just fixed it!

Now, I believe that the tests are running properly and headers are not getting duplicated anymore.

yusukebe:
Hi @Amorim33 

Thanks! Can you run `bun run format:fix && bun run lint:fix` to format the code?

Amorim33:
Sure, thank you @yusukebe !

yusukebe (approved):
LGTM!

PR #2592, opened 2024-05-02
feat(hc): add `init` option

### LINKED ISSUE #2485: hc with Credentials?
### Discussed in https://github.com/orgs/honojs/discussions/2291

<div type='discussions-op-text'>

<sup>Originally posted by **lajbel** March  1, 2024</sup>
Looks pretty obvious to have something like `credentials` in the configuration of `hc`, something like this

```js
const authClient = hc<SomeType>("/api/auth/", {
    headers: {} // You can already set headers!
    credentials: "include",
});
```

I was searching here and there's only one option available to do this approach

```js
const api = hc<AppType>("api/auth", {
  fetch: (req, init) =>
    fetch(req, {
      ...init,
      credentials: "include", // Why not have all fetch options too?
    })
})
```

RPC is pretty awesome, but some things are missing</div>

NamesMT:
Initally I used a function to merge the headers, but then realized it could be great if it could be used as highest priority control, added comment to tell user to be careful instead.

Resolves #2485

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/client/client.ts
+++ b/deno_dist/client/client.ts
@@ -123,6 +123,7 @@ class ClientRequestImpl {
       body: setBody ? this.rBody : undefined,
       method: methodUpperCase,
       headers: headers,
+      ...opt?.init,
     })
   }
 }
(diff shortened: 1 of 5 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@NamesMT 

This is a good feature! Let's ship it.

Issue #379, opened 2022-07-14
how to get connectinfo(remoteAddress)


biluohc:

yusukebe:
Hi @biluohc 

If your script run on Cloudflare Workers, you can get it by `x-real-ip` Request header.

```ts
app.get('/', (c) => {
  console.log(c.req.header('x-real-ip'))
  return c.text('foo')
})
```

On Deno and Bun, we don't have any way to get the client remote address.

biluohc:
Thanks.

The second parameter of the Handler is ConnInfo in the the standard library of Deno, consider supporting it?

https://deno.land/x/std@0.148.0/http/server.ts?code#L35

biluohc:
Solved.

```sh
export function sa(da: Deno.Addr) {
  if (["tcp", "udp"].includes(da.transport)) {
    const a = (da as Deno.NetAddr);
    return `${a.hostname}:${a.port}`;
  } else {
    const a = (da as Deno.UnixAddr);
    return a.path;
  }
}

serve(async (request: Request, ci: ConnInfo) => {
  const env = { sa: sa(ci.remoteAddr) };
  return app.fetch(request, env)
})
```

PR #2595, opened 2024-05-03
feat: Introduce ConnInfo helper/adapter

nakasyou:
I created ConnInfo helper/adapter for Bun, Deno and CF Workers.
You can get client IP Address with Hono easier if we marge PR.

This provides runtime-less ConnInfo object.

For example, you can get IP Address on Bun:
```ts
import { Hono } from 'hono'
import { getConnInfo } from 'hono/bun'

const app = new Hono()

app.get('/', c => c.text(`Your ip address is ${getConnInfo(c).remote.address}!!`))

Bun.serve(app)
```

If you want to use Workers, replace import:
```ts
import { getConnInfo } from 'hono/cloudflare-workers'
```

In Deno:
```ts
import { getConnInfo } from 'https://deno.land/x/hono/helper.ts'
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/adapter/deno/conninfo.ts
+++ b/deno_dist/adapter/deno/conninfo.ts
@@ -1,0 +1,18 @@
+import type { GetConnInfo } from '../../helper/conninfo/index.ts'
+
+/**
+ * Get conninfo with Deno
+ * @param c Context
+ * @returns ConnInfo
+ */
+export const getConnInfo: GetConnInfo = (c) => {
+  const { remoteAddr } = c.env
+  return {
+    remote: {
+      address: remoteAddr.hostname,
+      port: remoteAddr.port,
+      transport: remoteAddr.transport,
+      addressType: 'unknown',
+    },
+  }
+}
--- a/src/adapter/bun/conninfo.test.ts
+++ b/src/adapter/bun/conninfo.test.ts
@@ -41,12 +41,12 @@ describe('getConnInfo', () => {
     expect(info.remote.addressType).toBe('IPv6')
     expect(info.remote.transport).toBeUndefined()
   })
-  it('Should throw error when user didn\'t give server', () => {
+  it('Should throw error when user did not give server', () => {
     const c = new Context(new HonoRequest(new Request('http://localhost/')), { env: {} })
 
     expect(() => getConnInfo(c)).toThrowError(TypeError)
   })
-  it('Should throw error when requestIP isn\'t function', () => {
+  it('Should throw error when requestIP is not function', () => {
     const c = new Context(new HonoRequest(new Request('http://localhost/')), {
       env: {
         requestIP: 0,
--- a/src/adapter/bun/conninfo.test.ts
+++ b/src/adapter/bun/conninfo.test.ts
@@ -41,12 +41,12 @@ describe('getConnInfo', () => {
     expect(info.remote.addressType).toBe('IPv6')
     expect(info.remote.transport).toBeUndefined()
   })
-  it("Should throw error when user didn't give server", () => {
+  it(`Should throw error when user didn't give server`, () => {
     const c = new Context(new HonoRequest(new Request('http://localhost/')), { env: {} })
 
     expect(() => getConnInfo(c)).toThrowError(TypeError)
   })
-  it("Should throw error when requestIP isn't function", () => {
+  it(`Should throw error when requestIP isn't function`, () => {
     const c = new Context(new HonoRequest(new Request('http://localhost/')), {
       env: {
         requestIP: 0,
--- a/src/adapter/bun/conninfo.test.ts
+++ b/src/adapter/bun/conninfo.test.ts
@@ -2,26 +2,56 @@ import { Context } from '../../context'
 import { HonoRequest } from '../../request'
 import { getConnInfo } from './conninfo'
 
+const createRandomBunServer = () => {
+  const address = Math.random().toString()
+  const port = Math.floor(Math.random() * (65535 + 1))
+  return {
+    address,
+    port,
+    server: {
+      requestIP() {
+        return {
+          address,
+          family: 'IPv6',
+          port,
+        }
+      },
+    },
+  }
+}
 describe('getConnInfo', () => {
   it('Should info is valid', () => {
-    const address = Math.random().toString()
-    const port = Math.floor(Math.random() * (65535 + 1))
-    const c = new Context(new HonoRequest(new Request('http://localhost/')), {
-      env: {
-        requestIP() {
-          return {
-            address,
-            family: 'IPv6',
-            port,
-          }
-        },
-      },
-    })
+    const { port, server, address } = createRandomBunServer()
+    const c = new Context(new HonoRequest(new Request('http://localhost/')), { env: server })
     const info = getConnInfo(c)
 
     expect(info.remote.port).toBe(port)
     expect(info.remote.address).toBe(address)
     expect(info.remote.addressType).toBe('IPv6')
     expect(info.remote.transport).toBeUndefined()
+  })
+  it('Should getConnInfo works when env is { server: server }', () => {
+    const { port, server, address } = createRandomBunServer()
+    const c = new Context(new HonoRequest(new Request('http://localhost/')), { env: { server } })
+
+    const info = getConnInfo(c)
+
+    expect(info.remote.port).toBe(port)
+    expect(info.remote.address).toBe(address)
+    expect(info.remote.addressType).toBe('IPv6')
+    expect(info.remote.transport).toBeUndefined()
+  })
+  it("Should throw error when user didn't give server", () => {
+    const c = new Context(new HonoRequest(new Request('http://localhost/')), { env: {} })
+
+    expect(() => getConnInfo(c)).toThrowError(TypeError)
+  })
+  it("Should throw error when requestIP isn't function", () => {
+    const c = new Context(new HonoRequest(new Request('http://localhost/')), {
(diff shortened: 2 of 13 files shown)


nakasyou:
reference:
- https://github.com/honojs/hono/issues/379
- https://github.com/orgs/honojs/discussions/1439
- https://github.com/orgs/honojs/discussions/1600
- https://github.com/orgs/honojs/discussions/1187

deno_dist/adapter/deno/conninfo.ts
```diff
@@ -0,0 +1,18 @@
+import type { GetConnInfo } from '../../helper/conninfo/index.ts'
+
+/**
+ * Get conninfo with Deno
+ * @param c Context
+ * @returns ConnInfo
+ */
+export const getConnInfo: GetConnInfo = (c) => {
+  const { remoteAddr } = c.env
```
ryuapp, line 9: This is funny.
But if you set up your own `env`, this will no longer work.
```ts
Deno.serve((req) => {
  const env = { url: "https://hono.dev"};
  return app.fetch(req, env);
});
```

nakasyou: Yes, it's fault. Bun also so.
Bun's WebSocket helper doesn't work as well.

Should I validate `c.env`?

ryuapp: `c.env` is environment variables that users should use freely, and I don't think it's a good idea to have an API that becomes unusable if you set env.
Although a common API to get an IP address is a great idea, I think we need to consider another way.

I don't have any good ideas at this point :'(

nakasyou: Hmm..., surely.

But I think that c.env has created to get Cloudflare Workers's bindings, so this usage may be correct.

Also, you can make custom envs by the way:
```ts
Deno.serve((req, server) => {
  const env = { ...server, url: "https://hono.dev"};
  return app.fetch(req, env);
});
```
@yusukebe What do you think about that?

yusukebe: @nakasyou 

You may use `env()` in `hono/adapter`!

https://hono.dev/helpers/adapter#env

nakasyou: @yusukebe
This helper needs 2nd argument of fetch, it doesn't need runtime environment values.

yusukebe: @nakasyou 

Sorry! I understood.

As you said, the second arg of `app.fetch` is designed for Cloudflare Workers's Bindings first. But now, as @ryuapp said, users can use it freely. The node-server append the `HttpBindings` to it:

https://github.com/honojs/node-server/blob/main/src/listener.ts#L186

yusukebe: Hmm.

If we don't change the API of the core, this method @nakasyou said is a better way now.

```ts
Deno.serve((req, server) => {
  const env = { ...server, url: 'https://hono.dev' }
  return app.fetch(req, env)
})
```

nakasyou: @yusukebe, thank you.

@ryuapp @yusukebe, Do you think that this problem can resolve if we add details of second argument of fetch to docs when pr merged?

yusukebe: @nakasyou I think yes.

But, this Bun case seems not to work:

```ts
Bun.serve({
  fetch: (req: Request, server) => {
    const env = { ...server, foo: 'bar' }
    return app.fetch(req, env)
  },
})

// bun run src/index.ts
```

Could you try it?

nakasyou: Hi @yusukebe.
Oh, I got error too because `server` is class instance.

What do you think about this code?:
```ts
Bun.serve({
  fetch: (req: Request, server) => {
    const env = { sever, foo: 'bar' }
    return app.fetch(req, env)
  },
})
```
In this code, if `c.env.requestIP` is undefined, helper tries to access `c.env.server.requestIP`.

I'll implement it if it's better.

ryuapp: It's ok if we write that on docs.
I've reconsidered that the complexity of setting up env outweighs the benefits of this feature.
Thank you.

yusukebe: @nakasyou 

> What do you think about this code?

Seems to be good! The naming `server` would be better because the Bun is naming the second argument of `fetch` as `server`:

CleanShot 2024-05-06 at 22 19 50@2x

https://bun.sh/docs/api/http#bun-serve

nakasyou: @ryuapp @yusukebe thank you for reviewing!
I'll implement it!

src/adapter/bun/conninfo.ts
```diff
@@ -0,0 +1,36 @@
...
+        requestIP?: (req: Request) => {
+          address: string
+          family: string
+          port: number
+        }
+      }
+    | undefined
+
+  if (!server) {
+    throw new TypeError('env has to include 2nd argument of fetch.')
```
yusukebe, line 22: This is nitpicking but, `env has to include 2nd argument of fetch.` would be

```
env has to include the 2nd argument of fetch.
```

src/adapter/bun/conninfo.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
+  if (typeof server.requestIP !== 'function') {
+    throw new TypeError('server.requestIP is not function.')
```
yusukebe, line 25: `server.requestIP is not a function.` is better.

deno_dist/adapter/deno/conninfo.ts
yusukebe: @nakasyou 

Gooood!

yusukebe:
@nakasyou 

CI fails. Could you run `bun run lint:fix` (As you said on X, I might have made a mistake in configuring Prettier and ESLint)?

src/adapter/bun/conninfo.ts
nakasyou: I fixed!

src/adapter/bun/conninfo.ts
nakasyou: I fixed it!

nakasyou:
@yusukebe
If strings include single quote, CI fails due to Prettier or ESLint.

Prettier recommends `"it's"`, but ESLint recommends `'it\'s'`

src/adapter/bun/conninfo.test.ts
```diff
@@ -0,0 +1,57 @@
...
+    const c = new Context(new HonoRequest(new Request('http://localhost/')), { env: { server } })
+
+    const info = getConnInfo(c)
+
+    expect(info.remote.port).toBe(port)
+    expect(info.remote.address).toBe(address)
+    expect(info.remote.addressType).toBe('IPv6')
+    expect(info.remote.transport).toBeUndefined()
+  })
+  it('Should throw error when user didn\'t give server', () => {
```
yusukebe, line 44: A simple solution. How about re-writing `didn\'t` to `did not`?

src/adapter/bun/conninfo.test.ts
```diff
@@ -0,0 +1,57 @@
...
+    const c = new Context(new HonoRequest(new Request('http://localhost/')), { env: {} })
+
+    expect(() => getConnInfo(c)).toThrowError(TypeError)
+  })
+  it('Should throw error when requestIP isn\'t function', () => {
```
yusukebe, line 49: `isn\'t` to `is not`.

src/adapter/bun/conninfo.test.ts
nakasyou: Thanks, I do it.
However, I think it's not a fundamental solution, it looks like coping therapy.

This discussion is out-topic, but I think we should open issue to resolve it.

src/adapter/bun/conninfo.test.ts
nakasyou: https://github.com/honojs/hono/pull/2595#discussion_r1593428888

src/adapter/bun/conninfo.test.ts
yusukebe: > This discussion is out-topic, but I think we should open issue to resolve it.

Yeah! Exactly.

nakasyou: I'll mark this problem as resolved.

yusukebe (approved):
LGTM!

yusukebe:
Hey @nakasyou 

Looks great! It's finished. This is a new feature, so I'm planning to release it in the next minor release, `v4.4`. So, I leave this PR not merged yet. Thanks!

curioustolearn:
Hello, 
I tried to use:
```
import { getConnInfo } from 'https://deno.land/x/hono/helper.ts'
```
However, I get the following error. Is this because this is not yet available in `v4.3.11`? Is there a way I can use this with `v4.3.11`?  

```
Module '"https://deno.land/x/hono/helper.ts"' has no exported member 'getConnInfo'.deno-ts(2305)
```

Thank you.

yusukebe:
Hi @curioustolearn 

Now `deno.land/x` is not used to publish the Deno module. Use [the JSR package](https://jsr.io/@hono/hono) instead of it:

```ts
import { Hono } from 'jsr:@hono/hono'
import { getConnInfo } from 'jsr:@hono/hono/deno'
```

PR #2605, opened 2024-05-04
fix(types): `ToSchema` compatibility

NamesMT:
Resolves honojs/middleware#496

This PR make `ToSchema` compatible with older versions.

Some informations about the issue:
> Okay, so on `4.3.0` we have two types breaking on `ToSchema`:
> 
> * Update from `O` is output to `extends TypedResponse` - 8129a3a
> * Update from `I` accepting `Input['in']` to `Input` - ea4e998
> 
> Previously I thought `ToSchema` is only for internal use so I didn't think much about it, but if it is exposed for other libraries I think we should revert `ToSchema` and creates a new `ToSchemaV2` for future uses to revert the breaking changes?

- [ ] Add tests - Need help adding tests
- [x] Run tests

### DIFF
--- a/deno_dist/types.ts
+++ b/deno_dist/types.ts
@@ -6,9 +6,11 @@ import type { Hono } from './hono.ts'
 import type { StatusCode } from './utils/http-status.ts'
 import type {
   IfAnyThenEmptyObject,
+  IsAny,
   JSONValue,
   Prettify,
   RemoveBlankRecord,
+  Simplify,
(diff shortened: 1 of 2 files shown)


yusukebe (approved):
LGTM!

yusukebe:
Hi @NamesMT 

Thanks! I, too, have confirmed it works well with Zod OpenAPI. I'll merge it now.

Issue #2607, opened 2024-05-04
Wrong type with InferResponseType<> and status code when using .route() method


armandsalle:
### What version of Hono are you using?

4.3.1

### What runtime/platform is your app running on?

Bun

### What steps can reproduce the bug?

This code reproduce the bug. You need to create a now router and import it to the current `app` router exposed to the client

```ts
import { Hono } from "hono"
import { hc } from "hono/client"

const myRouter = new Hono()
  .get("/", (c) => {
    const randomBoolean = Math.random() < 0.5
    if (!randomBoolean) {
      return c.json({ error: "not found" }, 404)
    }

    return c.json({ success: true }, 200)
  })

const app = new Hono()
  .basePath("/api")
  .route("/hello", myRouter)

const client = hc<typeof app>("/")

type ResponseType200 = InferResponseType<typeof client.api.hello.$get, 200>
type ResponseType404 = InferResponseType<typeof client.api.hello.$get, 404>
```

### What is the expected behavior?

In the previous code, `ResponseType200` should be of type `{ success: boolean }` and `ResponseType404` should be of type `{ error: string }`

### What do you see instead?

`ResponseType200` and `ResponseType404` are both typed as 
```ts
{ success: boolean } | { error: string }
```

### Additional information

In the [4.3.0 releaseStatus code type]( https://github.com/honojs/hono/releases/tag/v4.3.0) the example show a working example that doesn't use `.route()` method. I think the issue came from the type merging with this method.

## The error
![CleanShot 2024-05-04 at 15 38 34@2x](https://github.com/honojs/hono/assets/28579123/e1eeeec2-57ee-424f-8d5c-e5246db98ce9)

## ✅ Types in the custom router that I want
![CleanShot 2024-05-04 at 15 39 47@2x](https://github.com/honojs/hono/assets/28579123/6232ff17-f634-4839-8c7c-a8c066da6e29)

## ❌ Types of the main router with my custom router imported with the `.route()`
![CleanShot 2024-05-04 at 15 40 16@2x](https://github.com/honojs/hono/assets/28579123/940752f9-2c28-4e2a-8ffe-3579bec6ab0a)

armandsalle:
the type `MergeSchemaPath` seems to flatten the schema

PR #2610, opened 2024-05-04
fix(types): correct inferring when Method has Endponts as Union

yusukebe:
Fixes #2607

### Author should do the followings, if applicable

- [x] Add tests
- [x] Run tests
- [x] `bun denoify` to generate files for Deno

PR #2611, opened 2024-05-04
fix: Status code of response type with route

armandsalle:

- [x] Add tests
- [x] Run tests

Close https://github.com/honojs/hono/issues/2607


armandsalle:
duplicate of https://github.com/honojs/hono/pull/2610

PR #2103, opened 2024-01-27
feat: introduce Body Limit Middleware using stream

yusukebe:
This is an implementation of the feature proposed in #2077 using stream.

### How to use

Simply specify the maximum size.

```ts
app.post(
  '/hello',
  bodyLimit({
    maxSize: 15 * Unit.b, // byte,
    onError: (c) => {
      return c.text('oveflow :(', 413)
    },
  }),
  (c) => {
    return c.text('pass :)')
  }
)
```

### Using stream

The problem discussed in #2077 that a body must be read at a time is solved by using stream.

### `c.req.bodyCache`

It uses `c.req.bodyCache` used by the Validator implementation. By setting the loaded ArrayBuffer to `c.req.bodyCache.arrayBuffer`, such as `c.req.json()` will use that cached body from then on.

### Author should do the followings, if applicable

- [x] Add tests
- [x] Run tests
- [x] `yarn denoify` to generate files for Deno

PR #2615, opened 2024-05-05
feat(middleware): Introduce Timeout Middleware

watany-dev:
I used to use it for personal use, but I see here that there seems to be a demand for it, so I share it with you.
https://github.com/orgs/honojs/discussions/1765

Many cloud environments implement infrastructure-level timeouts, but it is useful if you want to set per-route timeouts on the application side.

---

## Timeout Middleware
This middleware enables you to easily manage request timeouts in your application. It allows you to set a maximum duration for requests and define custom error responses if the specified timeout is exceeded.

## Import

```typescript
import { Hono } from 'hono';
import { timeout } from 'hono/timeout';
```

## Usage

Here's how to use the Timeout Middleware with default and custom settings:

Default Settings:

```typescript
const app = new Hono();

// Applying a 5-second timeout
app.use('/api', timeout(5000));

// Handling a route
app.get('/api/data', async (c) => {
  // Your route handler logic
  return c.json({ data: 'Your data here' });
});
```

## Custom Timeout Settings:

```typescript
app.use('/api/long-process', timeout('1m', {
  message: 'Request timeout. Please try again later.',
  code: 408,
}));

app.get('/api/long-process', async (c) => {
  // Simulate a long process
  await new Promise(resolve => setTimeout(resolve, 61000));
  return c.json({ data: 'This usually takes longer' });
});
```

## Understanding Timeouts
The duration for the timeout can be specified in milliseconds or as a string in a human-readable format, such as '1m' for one minute or '10s' for ten seconds. The parseDuration function translates these string values into milliseconds.

## Middleware Conflicts
Be cautious about middleware order, especially when using error-handling or other timing-related middleware, as it might affect the behavior of this timeout middleware.

---

- [x] Add tests
- [x] Run tests


src/middleware/timeout/index.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+    totalMilliseconds += value * units[unit]
+  }
+
+  return totalMilliseconds
+}
+
+export const timeout = (
+  duration: number | string,
+  options: TimeoutOptions = {}
```
yusukebe, line 38: How about taking `exception` as the second arg:

```ts
export const timeout = (
  duration: number | string,
  exception?: HTTPException | ((c: Context) => HTTPException)
)
```

How to use it (callback pattern):

```ts
app.get(
  '/',
  timeout(1000, (c) => {
    return new HTTPException(408, {
      message: `Timeout occurred at ${c.req.path}`,
    })
  }),
  async (c) => {
    // ...
  }
)
```

yusukebe:
@watany-dev 

Thank you for the PR. I like this! But we have to discuss some points. One is that I've left a comment.

### Is `parseDuration` necessary?

Honestly, I don't think `parseDuration` is not necessary. It's OK to get only a number of `ms` in `timeout()`:

```ts
export const timeout = (
  duration: number,
  //
)
```

We've discussed the Body Limit Middleware in #2103 about a similar issue.  In the Body Limit Middleware case, we decided not to accept introducing "Unit" https://github.com/honojs/hono/pull/2103/commits/a2c35ddee3380e227a0c44c5d5532477dd56c2d4. I also think we don't need the `parseDuration` method for the same reason. If we want to add `11s`, we can simply write `1000 * 11`.

What do you think?

yusukebe:
Hey @usualoma, I want to know your thoughts on this PR. I think it's a good idea!

usualoma:
@watany-dev Thank you!
I agree with the merge as it is a useful feature.

### `parseDuration`

It is close to the alternative one of my following comments on BodyLimit, which I think is one possible way of doing it.
https://github.com/honojs/hono/pull/2103#issuecomment-1967881342

However, well, It seems to me that only number is sufficient, given the following points.

* That hono's other middleware does not convert units.
* JavaScript (or TypeScript) users are used to setTimeout() and setInterval(), and are used to milliseconds when specifying duration

#### If this PR continues to use `parseDuration`

The following strings should be made an error.

`'500milliseconds'`

watany-dev:
I actually think that only ms is needed for Duration. I adopted this method because of users of this method in other middleware.

I feel like I could leave the helper function in between, but if there are no opinions, I'll remove it once.

```ts
  app.get(
    '/',
    timeout(parseDuration("1s50ms")),
    async (c) => {
      // ...
    }
  )
```

src/middleware/timeout/index.ts
```diff
@@ -0,0 +1,61 @@
...
+): MiddlewareHandler => {
+  const errorMessage = options.message ?? DEFAULT_ERROR_MESSAGE
+  const errorCode = options.code ?? DEFAULT_ERROR_CODE
+  const ms = typeof duration === 'string' ? parseDuration(duration) : duration
+
+  return async (context, next) => {
```
usualoma, line 44: For the debugging of users, it would be better to return a named function like other middleware.
https://github.com/honojs/hono/blob/e9e3c8ab65039a8bd2d9f3d961ff8d1fa7a6bdfe/src/middleware/body-limit/index.ts#L50

watany-dev:
@yusukebe @usualoma 
Thanks for your comment. I actually think that the DURATION should only be ms. I adopted this method because of users of this method in other middleware. Allow ms in simple's number format.

watany-dev:
---

## Timeout Middleware
This middleware enables you to easily manage request timeouts in your application. It allows you to set a maximum duration for requests and optionally define custom error responses if the specified timeout is exceeded.

## Import

```typescript
import { Hono } from 'hono';
import { timeout } from 'hono/timeout';
```

## Usage
Here's how to use the Timeout Middleware with both default and custom settings:

Default Settings:

```typescript
const app = new Hono();

// Applying a 5-second timeout
app.use('/api', timeout(5000));

// Handling a route
app.get('/api/data', async (c) => {
  // Your route handler logic
  return c.json({ data: 'Your data here' });
});
```

Custom Timeout Settings:

```typescript

import { HTTPException } from 'hono/http-exception';

// Custom exception factory function
const customTimeoutException = (context) => new HTTPException(408, {
  message: `Request timeout after waiting ${context.req.headers.get('Duration')} seconds. Please try again later.`,
});

// for Static Exception Message
// const customTimeoutException = new HTTPException(408, {
//   message: 'Operation timed out. Please try again later.'
// });

// Applying a 1-minute timeout with a custom exception
app.use('/api/long-process', timeout(60000, customTimeoutException));

app.get('/api/long-process', async (c) => {
  // Simulate a long process
  await new Promise(resolve => setTimeout(resolve, 61000));
  return c.json({ data: 'This usually takes longer' });
});

```

## Note
The duration for the timeout can be specified in milliseconds. The middleware will automatically reject the promise and potentially throw an error if the specified duration is exceeded.


## Middleware Conflicts
Be cautious about the order of middleware, especially when using error-handling or other timing-related middleware, as it might affect the behavior of this timeout middleware. Always ensure that the timeout middleware is placed appropriately within your middleware stack to avoid unintended interactions or overwrites.

src/middleware/timeout/index.ts
watany-dev: thx! I fixed.

src/middleware/timeout/index.ts
watany-dev: Nice. I fixed it, can you see if the type definitions and other perceptions you think match?

deno_dist/middleware/timeout/index.ts
```diff
@@ -0,0 +1,31 @@
+import type { Context } from '../../context.ts'
+import { HTTPException } from '../../http-exception.ts'
+import type { MiddlewareHandler } from '../../types.ts'
+
+export type ExceptionFactory = (context: Context) => HTTPException
```
yusukebe, line 5: This is a matter of taste, but "Factory" is the name for a method or a class that creates multiple other things, so I think it's not a good idea to give it to the method.

How about using `ExecptionFunction` or `HTTPExecptionFunction`?

yusukebe:
@watany-dev 

> Thank you. I think I have mostly rewritten it. Here is the revised spec document.

Awesome! I like it.

But I have a problem with the current code. I tried the following example on Cloudflare Workers with Wrangler and Bun:

```ts
import { HTTPException } from '../../src/http-exception'
import { timeout } from '../../src/middleware/timeout'

const app = new Hono()

app.get(
  '/',
  timeout(1000, (c) => {
    return new HTTPException(408, {
      message: `Timeout occurred at ${c.req.path}`,
    })
  }),
  async (c) => {
    await new Promise((resolve) => setTimeout(resolve, 3000))
  }
)

export default app
```

It behaves weirdly. See the screencast:

https://github.com/honojs/hono/assets/10682/feda8f4b-ec80-4c83-8b75-b2825ebe127a

It works correctly in the first request. But from the second request, it will not stop until 3 seconds have passed, and it is accessed 3 times. This behavior happens to both Wrangler and Bun.

Can you see it?

watany-dev:
I will check. As I was writing the Stream example above, I thought that maybe the clearTimeout is not working correctly.

watany-dev:
Note. 
I did not reproduce it with bun 1.1.6, and Cloudflare workers. I don't have time today, I will try again after tomorrow.

yusukebe:
Thanks @watany-dev 

By the way. If I don't set the second arg like the following, it works well.

```ts
const app = new Hono()

app.get('/', timeout(1000), async (c) => {
  await new Promise((resolve) => setTimeout(resolve, 3000))
})

export default app
```

usualoma:
I believe this is Google Chrome's behaviour for 408

https://groups.google.com/a/chromium.org/g/chromium-dev/c/urswDsm6Pe0/m/JBObkCT0AAAJ

According to the RFC, requests can be repeated in status 408, so I think Google Chrome is resending requests accordingly.

> If the client has an outstanding request in transit, the client MAY repeat that request on a new connection.
https://datatracker.ietf.org/doc/html/rfc7231#section-6.5.7

yusukebe:
@usualoma 

> I believe this is Google Chrome's behaviour for 408

You are right! I changed the status code `408` to `504`, and it works as expected!

@watany-dev 

Sorry for bothering you (If you paid for Cloudflare only for this purpose, super sorry!).

deno_dist/middleware/timeout/index.ts
watany-dev: fixed!

watany-dev:
@yusukebe 
Thank you for your feedback. I've also corrected the comments section. I'm looking forward to experimenting with Cloudflare Workers :)

package.json
```diff
@@ -179,6 +179,11 @@
       "import": "./dist/middleware/jwt/index.js",
       "require": "./dist/cjs/middleware/jwt/index.js"
     },
+    "./timeout": {
+      "types": "./dist/types/middleware/timeout/index.d.ts",
+      "import": "./dist/middleware/timeout/index.js",
+      "require": "./dist/cjs/middleware/timeout/index.js"
+    },
```
yusukebe, line 186: Could you add a definition in `typesVersions` too?

https://github.com/honojs/hono/blob/main/package.json#L343

deno_dist/middleware/timeout/index.ts
yusukebe: Good!

package.json
watany-dev: ok.
https://github.com/honojs/hono/pull/2615/commits/14e516478809dede603b922e58223d08c485ad55

src/middleware/timeout/index.ts
```diff
@@ -0,0 +1,31 @@
+import type { Context } from '../../context'
+import { HTTPException } from '../../http-exception'
+import type { MiddlewareHandler } from '../../types'
+
+export type HTTPExecptionFunction = (context: Context) => HTTPException
```
yusukebe, line 5: It's a typo! Should be:

```
HTTPExceptionFunction
```

watany-dev: sorry!! fixed.

yusukebe (approved):
LGTM!

yusukebe:
Hey @watany-dev !

Looks good! Let's goo with this. We'll release it in the next minor version`v4.4`. Let's leave this open until I add the `next` branch later and merge this into it. Thanks!

watany-dev:
@yusukebe 
I've added a JS Doc.

fzn0x:
I'm thinking we should improve the PR template to tell every PR that adding a new feature should add TSDoc/JSDoc @yusukebe This way we can slowly migrate codebase to use TSDoc/JSDoc

PR #2618, opened 2024-05-05
feat(utils): add named http codes

qoomon:

- [x] Add tests
- [x] Run tests


watany-dev:
In my personal opinion, this does not seem very necessary. The reason-phrase has been ignored since HTTP/2 and would only increase the size.

qoomon:
I agree the phrases do not add a lot of value, however still I think the named status adds a lot of readability.

yusukebe:
Hi @qoomon 

Thank you for the PR.

We had the same feature but removed it in this PR: https://github.com/honojs/hono/pull/1037.

I'm still thinking it's unnecessary for the same reason.

qoomon:
Eventually I agree and understand.

yusukebe:
Thanks!

PR #2619, opened 2024-05-06
feat: custom response in hono/streaming

Kanahiro:
related: https://github.com/orgs/honojs/discussions/2617

## proposed signature

added new argument

```typescript
// typeof stream()
(
  c: Context,
  onError?: (e: Error, stream: StreamingApi) => Promise<void>,
  responseFactory?: (c: Context, stream: StreamingApi) => Response // new!
): Response
```

same fix can be applied to `streamText` and `streamSSE`.

- [x] Add tests
- [x] Run tests


Kanahiro:
Or this may be better?

```typescript
// typeof stream()
(
  c: Context,
  options: {
    onError?: (e: Error, stream: StreamingApi) => Promise<void>,
    responseFactory?: (c: Context, stream: StreamingApi) => Response
  } = {}
): Response
```

PR #2623, opened 2024-05-06
fix(factory): support `createFactory()` destruction

### LINKED ISSUE #2621: Declaration of `Factory.createHandlers` is not appropriate

4.3.2

Cloudflare Workers

1. clone this repo. https://github.com/sushi-chaaaan/hono-factory-bind
2. run `pnpm install`.
3. open `src/factory.ts`.

No ESLint errors from typescript-eslint

### What do you see instead?

typescript-eslint will cause `unbound-method` error.

### Additional information

This is a error because `createHandlers` is not declared in the Arrow function and `this` is not explicitly bound.
Also, the actual function implementation does not access `this`, so no runtime error occurs.

yusukebe:
Fixes #2621

- [x] Add tests
- [x] Run tests


yusukebe:
Hi @sushi-chaaaan!

Thank you for raising the issue. I've fixed it in this PR. Could you review this?

sushi-chaaaan:
Thanks very much @yusukebe ! Looks good to me.
I was trying to fix the problem at hand, and it was almost the same implementation.
I see that specifying any for the body of createHandlers is optimal in this case. I learned a lot.

yusukebe:
> I learned that it is optimal to specify “any” for argument in this case.

If I understand correctly, if we define the proper types like `CreateHandlersInterface`, we can use `any` in the actual implementation like `createHandlers()`.

Anyway, thank you!

Issue #2633, opened 2024-05-07
Support json subtypes like application/merge-patch+json in validator


ztiromoritz:
### What version of Hono are you using?

4.3.2

### What runtime/platform is your app running on?

Bun (but should affect all)

### What steps can reproduce the bug?

Create a route with a validator with target 'json'.
Request with a Content-Type Header like 'application/merge-patch+json' 

### What is the expected behavior?

Validation should not fail with a message like:
Invalid HTTP header: ContentType=application/merge-patch+json.

### What do you see instead?

The request should result in a 200, or at least should not fail on this specific check.

### Additional information

One could additionally argue, that it should not be the responsibility of the json validator to check the request header.
Instead the the validator with target 'header' should be responsible for this. But I see the convenience and brevity of the current solution.

We also have no option to control the error response in this case.

PR #2634, opened 2024-05-07
feat(validator): check for json subtypes in validator

ztiromoritz:
A minimal fix for #2633 using a regex. 
I liked the simplicity of the current `startsWith`-approach but it excluded subtypes.

I did:
- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/validator/validator.ts
+++ b/deno_dist/validator/validator.ts
@@ -67,7 +67,7 @@ export const validator = <
 
     switch (target) {
       case 'json':
-        if (!contentType || !contentType.startsWith('application/json')) {
+        if (!contentType || !contentType.match(/^application\/[a-z-]*\+json(\s*;.*)?$/)) {
           const message = `Invalid HTTP header: Content-Type=${contentType}`
           throw new HTTPException(400, { message })
         }
--- a/deno_dist/validator/validator.ts
+++ b/deno_dist/validator/validator.ts
@@ -67,7 +67,7 @@ export const validator = <
 
     switch (target) {
       case 'json':
-        if (!contentType || !contentType.match(/^application\/[a-z-]*\+json(\s*;.*)?$/)) {
+        if (!contentType || !/^application\/([a-z-]+\+)?json/.test(contentType)) {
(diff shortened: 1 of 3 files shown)


mgrithm:
lgtm

deno_dist/validator/validator.ts
```diff
@@ -67,7 +67,7 @@ export const validator = <
 
     switch (target) {
       case 'json':
-        if (!contentType || !contentType.startsWith('application/json')) {
+        if (!contentType || !contentType.match(/^application\/[a-z-]*\+json(\s*;.*)?$/)) {
```
yusukebe, line 70: * You should use the `test` instead of the `match` if you don't capture strings.
* `^application\/[a-z-]*\+` should be `application\/([a-z-]+\+)*`.
* The last `(\s*;.*)?` is necessary?

ztiromoritz:
Thanks!
 * I use `test` now
 * Landed on this regexp: `/^application\/([a-z-]+\+)?json/`
   - the `^` in front is needed to not fail the `Xapplication/json` test
   - I added `?` to the `([a-z-]+\+)` group instead of `*`, because there can only be one subtype
 * Yes, `(\s*;.*)?` is not needed.

yusukebe:
Hi @ztiromoritz 

Thanks! Looks good. To avoid the CI error, could you run `bun run format:fix && bun run lint:fix` to format the code?

yusukebe (approved):
LGTM!

yusukebe:
Hi @ztiromoritz 

Thank you for your contribution. I'll merge this into the main now and will release it in a `patch` release since this is a minor `feat`.

deno_dist/validator/validator.ts
ForceTower: Yeah... that last part was needed.

Some clients (like android's retrofit) sends the content type as: application/json; charset=UTF-8. Without that last bit the content-type header doesn't pass the regex.

Retrofit's discussion on this: https://github.com/square/retrofit/issues/794

PR #2635, opened 2024-05-07
refactor: functiom from filepath and docs

mgrithm:
- [X] Add tests
- [X] Run tests
- [X] `bun denoify` to generate files for Deno

### DIFF
--- a/deno_dist/utils/filepath.ts
+++ b/deno_dist/utils/filepath.ts
@@ -1,51 +1,59 @@
+const DEFAULT_DOCUMENT = 'index.html'
+const VALID_EXTENSION_REGEX = /\.[a-zA-Z0-9]+$/
+const PARENT_DIRECTORY_REGEX = /(?:^|[\/\\])\.\.(?:$|[\/\\])/
+
 type FilePathOptions = {
   filename: string
   root?: string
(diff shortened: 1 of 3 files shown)


yusukebe:
@mgrithm 

I appreciate the PR, but I don't really welcome the refactoring because there will be a lot of diffs by those who didn't implement it.

mgrithm:
@yusukebe 

I kept the same functionality. I just changed the way we did it for reading/efficiency

yusukebe:
> I kept the same functionality. I just changed the way we did it for reading/efficiency

I know that. But if this PR is accepted, other parts of it will be refactored. Some styles can implement one logic. However, we currently don't have documentation like the "code-style-guide," so it's not good to refactor the code to your style without it. So, even if it seems like a good idea, it is difficult to accept this. This is the opinion of a maintainer.

mgrithm:
> I know that. But if this PR is accepted, other parts of it will be refactored. Some styles can implement one logic. However, we currently don't have documentation like the "code-style-guide," so it's not good to refactor the code to your style without it. So, even if it seems like a good idea, it is difficult to accept this. This is the opinion of a maintainer.

Alright. How can I contribute from other way?

PR #2661, opened 2024-05-11
feat(hono-base): tweaks `fetch` signature for better compatibility with `deno serve`

### LINKED ISSUE #2653: Support for `deno serve` in Den oadapter

Deno now supports a new `serve` command in the command line. It's syntactic sugar for `deno run`, which automatically starts an HTTP server. It works by expecting the executed script to be a module that exports an object with a `fetch` property. This property is an asynchronous function that expects one parameter of type `Request` and returns a `Response`.

The following code theoretically should work:

```typescript
const app = new Hono()

app.get('/', (c) => {
    return c.text('Hello Hono!')
})

export default {fetch: app.fetch};
```

Or even simply:
```typescript
export default app;
```

However, both this code fails when Deno checks if `fetch` is a function expecting one parameter.
Invalid type for fetch: must be a function with a single parameter

It's possible to rewrite the code as follows:

```typescript
const app = new Hono()

app.get('/', (c) => {
    return c.text('Hello Hono!')
})

export default {
    fetch: async (request: Request): Promise<Response> => await app.fetch(request)
};
```

But it would be good if the `fetch` function exported from `Hono` passed the Deno test, allowing the use of the abbreviated notation.

usualoma:
fixes #2653

Since deno checks `.length` as follows, this PR adjust `app.fetch.length` to be 1.
https://github.com/denoland/deno/blob/dac49a116e094be1a3048305dfb6b10bbddcc030/ext/http/00_serve.ts#L796-L800

We will be able to use `export default app` for `deno serve` with this PR.

This definition may seem a bit odd, but it is not overly complicated; I think the advantage of "you can write `dexport defult app` in deno just as well as in bun" is more significant.

- <del>[ ] Add tests</del>
- [x] Run tests

### DIFF
--- a/deno_dist/hono-base.ts
+++ b/deno_dist/hono-base.ts
@@ -371,8 +371,12 @@ class Hono<
    * `.fetch()` will be entry point of your app.
    * @see https://hono.dev/api/hono#fetch
    */
-  fetch = (request: Request, Env?: E['Bindings'] | {}, executionCtx?: ExecutionContext) => {
-    return this.dispatch(request, executionCtx, Env, request.method)
+  fetch: (
+    request: Request,
+    Env?: E['Bindings'] | {},
+    executionCtx?: ExecutionContext
+  ) => Response | Promise<Response> = (request, ...rest) => {
+    return this.dispatch(request, rest[1], rest[0], request.method)
   }
 
   /**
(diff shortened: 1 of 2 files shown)


usualoma:
@yusukebe What do you think about this?

yusukebe (approved):
LGTM!

I didn't know the `length` will be `1` with that implementation!

yusukebe:
Thanks @usualoma !

This is good! I also wanted this feature. I'll merge this into the `main` right now and release it as a patch release later.

PR #2663, opened 2024-05-12
perf(types): add type annotations

yusukebe:
This PR added type annotations to reduce "[Slow Types](https://jsr.io/docs/about-slow-types)".

We are planning to support the [JSR](https://jsr.io/). For this, it's better to remove [Slow Types](https://jsr.io/docs/about-slow-types). Adding type annotations is one big issue in reducing Slow Types. In addition, it will improve TypeScript inferring performance.

It could be included in PRs to support the JSR, but not only that, so make it a separate PR and merge it into the `main` now.

### DIFF
--- a/deno_dist/context.ts
+++ b/deno_dist/context.ts
@@ -1,6 +1,6 @@
 import type { HonoRequest } from './request.ts'
-import type { Env, FetchEventLike, NotFoundHandler, Input, TypedResponse } from './types.ts'
-import { resolveCallback, HtmlEscapedCallbackPhase } from './utils/html.ts'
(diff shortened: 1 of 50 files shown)


yusukebe:
Hey @nakasyou !

What do you think of this PR? I think it would be good to merge it before creating a PR to support JSR.

nakasyou:
Hi @yusukebe 
> What do you think of this PR? I think it would be good to merge it before creating a PR to support JSR.

Yes! I think so.
It's good to reduce Slow Types, even if it's not related to JSR

yusukebe:
@nakasyou 

Thanks. This PR does not have a problem. I'll merge this now.

PR #2667, opened 2024-05-13
feat(jsr): remove helper.ts and middleware.ts

nakasyou:
In `feat/jsr`, denoify has removed.
exports field of npm and jsr don't have reference to `helper.ts` and `middleware.ts`, so these files should be removed.

- [ ] Add tests
- [x] Run tests


yusukebe:
@nakasyou 

You are right. We can remove them. Thanks!

PR #2669, opened 2024-05-13
feat(jsr): delete `mod.ts`

nakasyou:
The reason is the same as #2667.
I forgot remove `mod.ts`.

- [ ] Add tests
- [x] Run tests


yusukebe:
Thanks! Merging.

PR #2675, opened 2024-05-15
feat(utils/body): add dot notation support for `parseBody`

### LINKED ISSUE #2656: context.req.parseBody(): option for support dot notation

For now this method has only option **all** for array support. It is awesome!

In my opinion, it would be very convenient to add an option to support dot notation (**dot: boolean**), which would greatly simplify the processing of nested objects.

Example:

```js
let formData = new FormData();
formData.append('obj.key1', 'value1');
formData.append('obj.key2', 'value2');

// processed without dot notation support, like `await context.req.parseBody()`
{
	"obj.key1": "value1",
    "obj.key2": "value2",
} 

// processed with dot notation support, like `await context.req.parseBody({ dot: true })`
{
	"obj": {
		"key1": "value1",
		"key2": "value2"
	}
}
```

fzn0x:
Closes #2656

cc @yusukebe  @MathurAditya724 

- [x] Add tests
- [x] Run tests


fzn0x:
Tag you guys @yusukebe @MathurAditya724 to review this. 👍

src/utils/body.ts
```diff
@@ -60,23 +75,57 @@ function convertFormDataToBodyData<T extends BodyData = BodyData>(
     const shouldParseAllValues = options.all || key.endsWith('[]')
 
     if (!shouldParseAllValues) {
-      form[key] = value
+      if (options.dot && key.includes('.')) {
+        setNestedValue(form, key.split('.'), value)
```
MathurAditya724, line 79: Maybe don't split the keys here to make the API look similar to `handleParsingAllValues`. You can do it inside the func too.

src/utils/body.ts
```diff
@@ -1,6 +1,6 @@
 import { HonoRequest } from '../request'
 
-export type BodyData = Record<string, string | File | (string | File)[]>
+export type BodyData = Record<string, string | object | File | (string | File)[]>
```
MathurAditya724, line 3: In place of an `object` type, we should now have a nested generic type that would be more helpful

MathurAditya724 (commented):
These are some basic changes you can do to improve this. I haven't looked at the `setNestedValue` func and the test cases (this was all the time I had 😅), but I will try to look at it soon.

src/utils/body.ts
```diff
@@ -18,11 +18,26 @@ export type ParseBodyOptions = {
...
+   * If dot is true:
+   * parseBody should return { obj: { key1: 'value1', key2: 'value2' } }
+   */
+  dot?: boolean
 }
 
 export const parseBody = async <T extends BodyData = BodyData>(
   request: HonoRequest | Request,
-  options: ParseBodyOptions = { all: false }
+  options: ParseBodyOptions = { all: false, dot: false }
```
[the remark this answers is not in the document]
fzn0x: You're right, I will try to improve this

src/utils/body.ts
```diff
@@ -18,11 +18,26 @@ export type ParseBodyOptions = {
...
+   * data.append('obj.key1', 'value1')
+   * data.append('obj.key2', 'value2')
+   *
+   * If dot is false:
+   * parseBody should return { 'obj.key1': 'value1', 'obj.key2': 'value2' }
+   *
+   * If dot is true:
+   * parseBody should return { obj: { key1: 'value1', key2: 'value2' } }
+   */
+  dot?: boolean
```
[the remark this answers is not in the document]
fzn0x: I'm not sure the best name to take except the well-known one `dot`, but +1 on `transformDotNotation`.

src/utils/body.ts
```diff
@@ -60,23 +75,57 @@ function convertFormDataToBodyData<T extends BodyData = BodyData>(
     const shouldParseAllValues = options.all || key.endsWith('[]')
 
     if (!shouldParseAllValues) {
-      form[key] = value
+      if (options.dot && key.includes('.')) {
```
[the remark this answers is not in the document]
fzn0x: Sure

src/utils/body.ts
fzn0x: Sure, will try same pattern

src/utils/body.ts
fzn0x: I'm not sure we really need generic type for an empty object `{}` used on dot notation logic, maybe you consider more on the user usage ?

src/utils/body.ts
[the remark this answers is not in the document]
fzn0x: Commit added

src/utils/body.ts
MathurAditya724: I believe this will provide a better type completion compared to having just an object type

```ts
type BodyDataValue = string, string | File | (string | File)[]
export type BodyData = Record<string, BodyDataValue | Record<string, BodyDataValue>>
```

fzn0x: Looks like can't use `type` circular references

fzn0x: I ended up with this
```js
type BodyDataValue = string | {} | File | (string | File)[] | { [key: string]: BodyDataValue }
export type BodyData = Record<string, BodyDataValue>
```

Looks neat!

MathurAditya724: we can change the linting rules and all. @yusukebe should we change them? or maybe we can use `// @ts-ignore` or something?

fzn0x:
@MathurAditya724 Reviews resolved, to decide whether is it `dot` or `transformDotNotation` naming, we can wait @yusukebe first.

src/utils/body.ts
fzn0x: Reopen it, whether continue use `{ [key: string]: BodyDataValue }` or changing linting rules.

fzn0x:
Dot notation support documentation will be added here: https://hono.dev/api/request#parsebody

src/utils/body.ts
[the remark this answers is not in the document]
vanodevium: @MathurAditya724 @fzn0x 

Let me say it as a topic starter.

The option to support arrays is not called **transformBracketsOrMultiValuesToArray**, but simply **all**.

Therefore, for this case too, you can use a simple name, **dot** or **dots**.

Thanks!

yusukebe:
Woooow. Super fast and quick! I'll review this now.

src/utils/body.ts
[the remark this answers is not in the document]
yusukebe: @MathurAditya724 @fzn0x @vanodevium Thank you for the suggestion.

I like `dot`. It's simple. Let's go with `dot`!

fzn0x:
Thanks @yusukebe  I will run denoify and lint once again

src/utils/body.ts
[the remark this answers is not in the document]
fzn0x: Okay allSettled, we will continue with `dot`.

src/utils/body.ts
```diff
@@ -18,20 +19,37 @@ export type ParseBodyOptions = {
...
+   */
+  dot?: boolean
 }
 
 export const parseBody = async <T extends BodyData = BodyData>(
   request: HonoRequest | Request,
-  options: ParseBodyOptions = { all: false }
+  options: ParseBodyOptions = {}
 ): Promise<T> => {
+  const { all = options?.all ?? false, dot = options?.dot ?? false } = options
```
MathurAditya724, line 43: This could be just `const { all = false, dot = false } = options`

src/utils/body.ts
```diff
@@ -18,20 +19,37 @@ export type ParseBodyOptions = {
...
+
   const headers = request instanceof HonoRequest ? request.raw.headers : request.headers
   const contentType = headers.get('Content-Type')
 
   if (
     (contentType !== null && contentType.startsWith('multipart/form-data')) ||
     (contentType !== null && contentType.startsWith('application/x-www-form-urlencoded'))
   ) {
-    return parseFormData<T>(request, options)
+    return parseFormData<T>(request, { ...options, all, dot })
```
MathurAditya724, line 52: Then this will be `{ all, dot }`

MathurAditya724 (commented):
Just some minor changes for the improvements you did based on my review

src/utils/body.ts
fzn0x: This will override the values given in options to `false`

src/utils/body.ts
fzn0x: Sure

vanodevium:
```js
if (
    contentType !== null && (contentType.startsWith('multipart/form-data')) || contentType.startsWith('application/x-www-form-urlencoded')
  )
```

src/utils/body.ts
MathurAditya724: If the value is not defined then it will override it to false.

fzn0x: But anyway lets add that

fzn0x: Added ✔️

yusukebe:
Hey. The following test does not pass now. But is this expected? What do you think? @fzn0x @MathurAditya724 

```ts
it('should parse data if both `all` and `dot` are set', async () => {
  const data = new FormData()
  data.append('obj.sub.foo', 'value1')
  data.append('obj.sub.foo', 'value2')
  data.append('key', 'value3')

  expect(await parseBody(req, { dot: true, all: true })).toEqual({
    obj: { sub: { foo: ['value1', 'value2'] } }, // Is tis expected??
    key: 'value3',
  })
})
```

Test test result:

```
- Expected
+ Received

  Object {
    "key": "value3",
    "obj": Object {
      "sub": Object {
-       "foo": Array [
-         "value1",
-         "value2",
-       ],
+       "foo": "value2",
      },
    },
  }
```

src/utils/body.ts
fzn0x: Is something wrong with the contentType? @vanodevium 

I believe we need to wrap it like this, right?

Since OR have two conditions, so we need to use parentheses

```ts
if (
  contentType !== null && ((contentType.startsWith('multipart/form-data')) || contentType.startsWith('application/x-www-form-urlencoded'))
)
```

vanodevium:
```js
if (
        contentType !== null && (contentType.startsWith('multipart/form-data') || contentType.startsWith('application/x-www-form-urlencoded'))
    )
```

fzn0x:
This is unexpected, let me fix that @yusukebe

yusukebe:
Anyway, this is a new "feature," so it may not be merged into the main immediately. So you don't have to hurry up:)

vanodevium:
@yusukebe 

I need it for everyday tasks :)

MathurAditya724:
@fzn0x I believe `handleNestedValues` and `handleParsingAllValues` might be clashing. Do let me know if you require any help

MathurAditya724:
Interesting, let me check

MathurAditya724:
Can I make changes to this PR?

fzn0x:
I found the solution but the hard thing is to keep the code clean 😅  @MathurAditya724 @yusukebe 

I guess we can continue with the review 🔥

src/utils/body.ts
fzn0x: Okay, let me remove it

fzn0x:
I believe we can add this in `v4.4`, what do you think? @yusukebe

yusukebe:
> I believe we can add this in `v4.4`, what do you think? @yusukebe 

Yes. Let's include this in `v4.4`.

usualoma:
Hi @fzn0x!
Thank you for the great PR.
I have two comments.

### `all` is always enabled.

Since `all` is always enabled in the current branch, I think the code needs to be modified by adding the following test.

```diff
diff --git a/src/utils/body.test.ts b/src/utils/body.test.ts
index cc154ef4..f9f7e3c7 100644
--- a/src/utils/body.test.ts
+++ b/src/utils/body.test.ts
@@ -50,6 +50,20 @@ describe('Parse Body Util', () => {
     })
   })
 
+  it('should override value if `all` option is false', async () => {
+    const data = new FormData()
+    data.append('file', 'aaa')
+    data.append('file', 'bbb')
+    data.append('message', 'hello')
+
+    const req = createRequest(FORM_URL, 'POST', data)
+
+    expect(await parseBody(req)).toEqual({
+      file: 'bbb',
+      message: 'hello',
+    })
+  })
+
   it('should parse multiple values if `all` option is true', async () => {
     const data = new FormData()
     data.append('file', 'aaa')
```

### Objects containing arbitrary data should be initialized using `Object.create(null)`

This would also require modification of existing code, but for objects that receive a request and put in arbitrary data, they should be initialized using `Object.create(null)`.

```diff
diff --git a/src/utils/body.ts b/src/utils/body.ts
index a3382806..8ef63893 100644
--- a/src/utils/body.ts
+++ b/src/utils/body.ts
@@ -72,7 +72,7 @@ function convertFormDataToBodyData<T extends BodyData = BodyData>(
   formData: FormData,
   options: ParseBodyOptions
 ): T {
-  const form: BodyData = {}
+  const form: BodyData = Object.create(null)
 
   formData.forEach((value, key) => {
     const shouldParseAllValues = options.all || key.endsWith('[]')
@@ -140,7 +140,7 @@ const handleNestedValues = (
           typeof nestedForm[key] !== 'object' ||
           Array.isArray(nestedForm[key])
         ) {
-          nestedForm[key] = {}
+          nestedForm[key] = Object.create(null)
         }
         nestedForm = nestedForm[key] as BodyData
       }
```

fzn0x:
Sorry guys, just got from work, I will check the review

fzn0x:
Fix added! kindly review, thanks @usualoma

usualoma:
@fzn0x Thank you! > 716de84

Sorry to add one more point. From a security standpoint, I think it would be better to prevent the File object's properties from being able to be updated.

```diff
diff --git a/src/utils/body.test.ts b/src/utils/body.test.ts
index 41235292..fb6da697 100644
--- a/src/utils/body.test.ts
+++ b/src/utils/body.test.ts
@@ -191,6 +191,32 @@ describe('Parse Body Util', () => {
     })
   })
 
+  it('should not update file object properties', async () => {
+    const file = new File(['foo'], 'file1', {
+      type: 'application/octet-stream',
+    })
+    const data = new FormData()
+
+    const req = createRequest(FORM_URL, 'POST', data)
+    vi.spyOn(req, 'formData').mockImplementation(
+      async () =>
+        ({
+          forEach: (cb) => {
+            cb(file, 'file', data)
+            cb('hoo', 'file.hoo', data)
+          },
+        } as FormData)
+    )
+
+    const parsedData = await parseBody(req, { dot: true })
+    expect(parsedData.file).not.instanceOf(File)
+    expect(parsedData).toEqual({
+      file: {
+        hoo: 'hoo',
+      },
+    })
+  })
+
   it('should parse multiple values if key ends with `[]`', async () => {
     const data = new FormData()
     data.append('file[]', 'aaa')
diff --git a/src/utils/body.ts b/src/utils/body.ts
index 1c77ab98..f07c6433 100644
--- a/src/utils/body.ts
+++ b/src/utils/body.ts
@@ -138,7 +138,8 @@ const handleNestedValues = (
         if (
           !nestedForm[key] ||
           typeof nestedForm[key] !== 'object' ||
-          Array.isArray(nestedForm[key])
+          Array.isArray(nestedForm[key]) ||
+          nestedForm[key] instanceof File
         ) {
           nestedForm[key] = Object.create(null)
         }
```

src/utils/body.ts
```diff
@@ -18,20 +19,37 @@ export type ParseBodyOptions = {
    * parseBody should return { file: ['aaa', 'bbb'], message: 'hello' }
    */
   all?: boolean
```
MathurAditya724, line 21: We can change these to be required and in the options, we can use `Partila<ParseBodyOptions>` to get them. This will provide a better type completion in the code.

src/utils/body.ts
```diff
@@ -18,20 +19,37 @@ export type ParseBodyOptions = {
...
+   * data.append('obj.key1', 'value1')
+   * data.append('obj.key2', 'value2')
+   *
+   * If dot is false:
+   * parseBody should return { 'obj.key1': 'value1', 'obj.key2': 'value2' }
+   *
+   * If dot is true:
+   * parseBody should return { obj: { key1: 'value1', key2: 'value2' } }
+   */
+  dot?: boolean
```
MathurAditya724, line 36: Same goes for here, update this to be required `dot: boolean`

src/utils/body.ts
```diff
@@ -18,20 +19,37 @@ export type ParseBodyOptions = {
...
 }
 
 export const parseBody = async <T extends BodyData = BodyData>(
   request: HonoRequest | Request,
-  options: ParseBodyOptions = { all: false }
+  options: ParseBodyOptions = Object.create(null)
```
MathurAditya724, line 41: We can change this to be `options: Partial<ParseBodyOptions> = Object.create(null)`

src/utils/body.ts
```diff
@@ -54,29 +72,92 @@ function convertFormDataToBodyData<T extends BodyData = BodyData>(
   formData: FormData,
   options: ParseBodyOptions
 ): T {
-  const form: BodyData = {}
+  const form: BodyData = Object.create(null)
 
   formData.forEach((value, key) => {
```
MathurAditya724, line 77: I have updated this loop

```ts
formData.forEach((value, key) => {
  const shouldParseAllValues = options.all || key.endsWith('[]')
  const shouldParseDotValues = options.dot && key.includes('.')

  if (shouldParseDotValues) {
    handleParsingNestedValues(form, key, value, shouldParseAllValues)
  } else if (shouldParseAllValues) {
    handleParsingAllValues(form, key, value)
  } else {
    form[key] = value
  }
})
```

src/utils/body.ts
```diff
@@ -54,29 +72,92 @@ function convertFormDataToBodyData<T extends BodyData = BodyData>(
...
+      handleParsingAllValues(form, key, value, options.dot)
     }
   })
 
   return form as T
 }
 
-const handleParsingAllValues = (form: BodyData, key: string, value: FormDataEntryValue): void => {
-  const formKey = form[key]
+const handleParsingAllValues = (
```
MathurAditya724, line 90: Removed the dot logic from here. This makes it similar to the original code too

```ts
const handleParsingAllValues = (
  form: BodyData,
  key: string,
  value: FormDataEntryValue,
): void => {
  const formKey = form[key]

  if (formKey && Array.isArray(formKey)) {
    ;(form[key] as (string | File)[]).push(value)
  } else if (formKey) {
    form[key] = [formKey as (string | File), value]
  } else {
    form[key] = value
  }
}
```

src/utils/body.ts
```diff
@@ -54,29 +72,92 @@ function convertFormDataToBodyData<T extends BodyData = BodyData>(
...
+      form[key] = value
+    }
+  }
+}
 
-  if (formKey && Array.isArray(formKey)) {
-    ;(form[key] as (string | File)[]).push(value)
-  } else if (formKey) {
-    form[key] = [formKey, value]
+const handleNestedValues = (
```
MathurAditya724, line 111: And finally here

```ts
const handleParsingNestedValues = (
  form: BodyData,
  key: string,
  value: FormDataEntryValue,
  parseAllValues: boolean
): void => {
  let nestedForm = form
  const keys = key.split('.')

  keys.forEach((key, index) => {
    if (index === keys.length - 1) {
      if (parseAllValues) {
        handleParsingAllValues(nestedForm, key, value)
      } else {
        nestedForm[key] = value
      }
    } else {
      if (
        !nestedForm[key] ||
        typeof nestedForm[key] !== 'object' ||
        Array.isArray(nestedForm[key]) ||
        nestedForm[key] instanceof File
      ) {
        nestedForm[key] = Object.create(null)
      }
      nestedForm = nestedForm[key] as BodyData
    }
  })
}
```

We can improve this more by removing the `parseAllValues` param in this function and somehow abstracting the logic to assign the value back in the `convertFormDataToBodyData` loop or maybe something else. I didn't want to add a dynamic function to manage this, but it's up to you guys, we can go that way also

MathurAditya724 (commented):
Hey @fzn0x, I have gone through the functions and made some improvements. All the tests are passing. Also, I added some pointers where we can improve this more. Let me know what you think

fzn0x:
Thanks for the reviews, 
I will submit a/some PR/s for the review above.

fzn0x:
Adding tsdoc for better user experience

src/utils/body.ts
fzn0x: I think creating a new function to just handle nested on all values is good to avoid a dynamic function, what do you think?

```ts
  formData.forEach((value, key) => {
    const shouldParseAllValues = options.all || key.endsWith('[]')
    const shouldParseDotValues = options.dot && key.includes('.')

    if (shouldParseDotValues) {
      if (shouldParseAllValues) {
        handleParsingNestedAllValues(form, key, value)
      } else {
        handleParsingNestedValues(form, key, value)
      }
    } else {
      assignValue(form, key, value, shouldParseAllValues)
    }
  })
```

MathurAditya724: This will just add bloat rather than help. Honestly, the functionality is working so this is just over-engineering. Let me explain the problem I am seeing in the current code. Down below I have shown the 2 worst case scenarios for this
```js
FormData {
  "obj.some.thing": "1",
  "obj.some.thing": "1",
  "obj.some.thing": "1",
}

// OR

FormData {
  "obj.some.thing": "1",
  "obj.some.one": "1",
  "obj.some.hello": "1",
}
```

Both cases have a complexity of **O(mn)** where _m_ is the number of entries and _n_ is the number of subkeys in an entry. We can improve this to **O(mn)** simply by changing the loop of `convertFormDataToBodyData` and minor changes in the `handleParsingNestedValues` function. Here are the changes respectively -

We are just processing the values separately. This first put the data in arrays the same as the old code and then converts it into a nested structure if the dot prop is `true`, only to the ones who have a `.`.
```ts
const form: BodyData = Object.create(null)

formData.forEach((value, key) => {
  const shouldParseAllValues = options.all || key.endsWith('[]')

  if (shouldParseAllValues) {
    handleParsingAllValues(form, key, value)
  } else {
    form[key] = value
  }
})

if (options.dot) {
  const nestedForm: BodyData = Object.create(null)

  Object.entries(form).forEach(([key, value]) => {
    const shouldParseDotValues = key.includes('.')

    if (shouldParseDotValues) {
      handleParsingNestedValues(nestedForm, key, value)
    } else {
      nestedForm[key] = value
    }
  })

  return nestedForm as T
}

return form as T
```
And remove the `parseAllValues` param as we don't need it now.
```ts
const handleParsingNestedValues = (
  form: BodyData,
  key: string,
  value: BodyDataValue
): void => {
  let nestedForm = form
  const keys = key.split('.')

  keys.forEach((key, index) => {
    if (index === keys.length - 1) {
        nestedForm[key] = value
    } else {
      if (
        !nestedForm[key] ||
        typeof nestedForm[key] !== 'object' ||
        Array.isArray(nestedForm[key]) ||
        nestedForm[key] instanceof File
      ) {
        nestedForm[key] = Object.create(null)
      }
      nestedForm = nestedForm[key] as BodyData
    }
  })
}
```
Overall this will improve the simplicity of the code and also the complexity. We can further improve the complexity of this but as I said before it's just over-engineering. Until @yusukebe wants us to improve this further we are good.

src/utils/body.test.ts
```diff
@@ -64,6 +104,101 @@ describe('Parse Body Util', () => {
...
+  it('should parse data if both `all` and `dot` are set', async () => {
+    const data = new FormData()
+    data.append('obj.sub.foo', 'value1')
+    data.append('obj.sub.foo', 'value2')
+    data.append('key', 'value3')
+
+    const req = createRequest(FORM_URL, 'POST', data)
+
+    expect(await parseBody(req, { dot: true, all: true })).toEqual({
+      obj: { sub: { foo: ['value1', 'value2'] } }, // Is tis expected??
```
MathurAditya724, line 179: You can remove this comment from here

fzn0x: Removed, nice catch

src/utils/body.ts
```diff
@@ -50,33 +84,93 @@ async function parseFormData<T extends BodyData = BodyData>(
...
   }
 }
+
+/**
+ * Handles parsing nested values using dot notation keys.
+ *
+ * @param {BodyData} form - The form data object.
+ * @param {string} key - The dot notation key.
+ * @param {FormDataEntryValue} value - The value to assign.
+ * @param {boolean} parseAllValues - Whether to parse all values as arrays.
```
MathurAditya724, line 155: Sorry for the nit-picking but now you can remove this line from the TSDocs and then do the denofy thing

fzn0x: Thanks, it my mistake actually, focusing on multiple work 😄

src/utils/body.ts
MathurAditya724: @fzn0x, just some last-minute formatting of `convertFormDataToBodyData` function, wrote it without thinking so much. Thanks

```ts
function convertFormDataToBodyData<T extends BodyData = BodyData>(
  formData: FormData,
  options: ParseBodyOptions
): T {
  const form: BodyData = Object.create(null)

  formData.forEach((value, key) => {
    const shouldParseAllValues = options.all || key.endsWith('[]')

    if (!shouldParseAllValues) {
      form[key] = value
    } else {
      handleParsingAllValues(form, key, value)
    }
  })

  if (options.dot) {
    Object.entries(form).forEach(([key, value]) => {
      const shouldParseDotValues = key.includes('.')

      if (shouldParseDotValues) {
        handleParsingNestedValues(form, key, value)
        delete form[key]
      }
    })
  }

  return form as T
}
```

yusukebe (approved):
LGTM!

yusukebe:
Thanks y'all. Now ready to merge to the `next` branch for the `v4.4.0`.

iveshenry18:
Was about to implement this myself when I found this—awesome work!

Curious—has there been any discussion around implementing the inverse (a flattenObject function) for the Hono client? It's relatively simple for the user to implement, but it would be nice to have a Hono-supported implementation that's guaranteed to be kept in sync (and losslessly round-trip) with this implementation. Would be awesome in terms of giving Hono even more of an RPC feel.

yusukebe:
@iveshenry18 

Oops! I just checked and it was not a flattenObject, but a function that parses a dot-notation.

iveshenry18:
Interesting—that's essentially what the code in this PR does.

I'm suggesting a utility that can take a nested object and parse it into a flat object with dot-notation keys. That way you can have a nested object on the client side, parse it into a flat object, send it as multipart/form-data, then use `parseBody({ dot: true })` (as implemented here) on the server side to get the original nested object.

Either way, I totally understand if that's out of scope for Hono! Just thought it might be a nice companion to this functionality. It seems like anyone using `parseBody({ dot: true })` will end up implementing their own version of what I just described.

PR #2679, opened 2024-05-15
fix(method-override): remove un-needed import of URLSearchParams in method override middleware

f5io:
this PR simply removes the `URLSearchParams` import from the node-specific package `url`.

when running hono and using the method override middleware in cloudflare workers, wrangler complains about the use of the `url` package. it prompts you to add the `node_compat` flag.

this seems a bit erroneous to me, since node has had `URLSearchParams` as a global since version 10. maybe i'm missing something specific about other runtimes, but all tests pass.

please nuke this PR if i'm flat-out wrong ;)

- [ ] Add tests
- [x] Run tests


yusukebe:
Hi @f5io 

You are right! This is our mistake. Thank you for the fix. Marging right now.

Issue #2672, opened 2024-05-14
Router does not decode URIs


szmarczak:
### What version of Hono are you using?

4.3.6

### What runtime/platform is your app running on?

Deno

### What steps can reproduce the bug?

Deno/Hono

```ts
import { Hono } from 'npm:hono';

const h = new Hono();

h.get('/|', async (c) => c.text('ayy'));

Deno.serve({
	port: 8080,
}, h.fetch);
```

Node

```js
// ayy
http.get({ host: 'localhost', port: 8080, path: '/|' }, res => res.pipe(process.stdout));'';

// 404 Not Found, expected ayy
http.get({ host: 'localhost', port: 8080, path: '/'+encodeURI('|') }, res => res.pipe(process.stdout));'';
```

### What is the expected behavior?

ayy

### What do you see instead?

404 Not Found

mgrithm:
@yusukebe, do you think that it is a bug?

yusukebe:
Hi @szmarczak 

The router doesn't need to decode the path. Instead, if you want to decode it, you can use the `getPath` option for `new Hono()`.

```ts
import { Hono } from 'hono'

const app = new Hono({
  getPath: (req) => {
    const url = req.url
    const queryIndex = url.indexOf('?', 8)
    const path = url.slice(url.indexOf('/', 8), queryIndex === -1 ? undefined : queryIndex)
    return decodeURI(path)
  },
})
```

szmarczak:
I disagree. The params are decoded out-of-the-box (though in [Request](https://github.com/honojs/hono/blob/f7225eefe6b055c04f041d325cc81db86c19ba4e/src/request.ts#L84)), I don't see why the path can't be:

Deno/Hono
```ts
import { Hono } from 'npm:hono';

const h = new Hono();

h.get('/:foobar', async (c) => c.text(c.req.param('foobar')));

Deno.serve({
	port: 8080,
}, h.fetch);
```

Node
```js
http.get({ host: 'localhost', port: 8080, path: '/'+encodeURIComponent('/') }, res => res.pipe(process.stdout));'';
```

[...]

PR #2688, opened 2024-05-16
feat: decode percent-encoded path in getPath

usualoma:
#2672

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/utils/url.ts
+++ b/src/utils/url.ts
@@ -70,10 +70,23 @@ export const getPattern = (label: string): Pattern | null => {
 }
 
 export const getPath = (request: Request): string => {
-  // Optimized: indexOf() + slice() is faster than RegExp
   const url = request.url
-  const queryIndex = url.indexOf('?', 8)
-  return url.slice(url.indexOf('/', 8), queryIndex === -1 ? undefined : queryIndex)
+  const start = url.indexOf('/', 8)
+  let i = start
+  for (; i < url.length; i++) {
+    const charCode = url.charCodeAt(i)
+    if (charCode === 37) {
+      // '%'
+      // If the path contains percent encoding, use `indexOf()` to find '?' and return the result immediately.
+      // Although this is a performance disadvantage, it is acceptable since we prefer cases that do not include percent encoding.
+      const queryIndex = url.indexOf('?', i)
+      return decodeURIComponent(url.slice(start, queryIndex === -1 ? undefined : queryIndex))
+    } else if (charCode === 63) {
+      // '?'
+      break
+    }
+  }
+  return url.slice(start, i)
 }
 
 export const getQueryStrings = (url: string): string => {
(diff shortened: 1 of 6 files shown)


src/utils/url.ts
```diff
@@ -70,10 +70,23 @@ export const getPattern = (label: string): Pattern | null => {
...
+  const start = url.indexOf('/', 8)
+  let i = start
+  for (; i < url.length; i++) {
+    const charCode = url.charCodeAt(i)
+    if (charCode === 37) {
+      // '%'
+      // If the path contains percent encoding, use `indexOf()` to find '?' and return the result immediately.
+      // Although this is a performance disadvantage, it is acceptable since we prefer cases that do not include percent encoding.
+      const queryIndex = url.indexOf('?', i)
+      return decodeURIComponent(url.slice(start, queryIndex === -1 ? undefined : queryIndex))
```
szmarczak, line 83:
```suggestion
      return decodeURI(url.slice(start, queryIndex === -1 ? undefined : queryIndex))
```

usualoma: Thanks for the suggestion.

With the current code, if we `s/decodeURIComponent/decodeURI/` , the following test fails, are you saying that "id is çawa yî%3F" is the expected result?

```diff
- id is çawa yî?
+ id is çawa yî%3F
````

https://github.com/usualoma/hono/blob/29735fe9cb5038ba5a45245be93344e5c16cfaef/src/hono.test.ts#L742-L748


This can be fixed by reverting 7192497. If we do that, the following test will fail. (double decoded)

```diff
- posts of %25
+ posts of %
````

https://github.com/usualoma/hono/blob/29735fe9cb5038ba5a45245be93344e5c16cfaef/src/hono.test.ts#L768-L774

Comment (szmarczak) [reply src/utils/url.ts:83]: > This can be fixed by reverting https://github.com/honojs/hono/commit/7192497e69a38e2915c84b1a4b7f80551e3e99f9.

Correct.

> If we do that, the following test will fail. (double decoded)

Ah, I see now, thanks for explaining! Well, in that case we need to `.replaceAll('%', '%25')` prior to calling `decodeURIComponent`.

Comment (usualoma) [reply src/utils/url.ts:83]: @szmarczak Thank you!

Unfortunately,

```
decodeURI('%2F') === decodeURI('%252F')
```

And cannot in principle be reverted with `replaceAll()`, since the original information is lost when the `decodeURI()` is applied.

For my part, I think that "using decodeURI() on a requested URL" is impossible.

szmarczak: What about doing `.replaceAll('%25', '%2525')` before calling `decodeURI`? This means that in order to catch `%` people will need to use `%25` in the path string, e.g. `/foo%25bar` (not: `/foo%bar`), but for all other percent-encoded characters it should decode as expected.

usualoma: Ah I see, that is true. That's great! I think it will be the spec you want.

I'm not going to include this change in this PR, since this PR has the same result as Django, but I'll make a separate PR for this.

usualoma: Thank you!

PR #2690, opened 2024-05-17
fix(context): Retain all cookies when passing `ResponseInit` to `c.body`

codeflows:
It seems cookie headers are not correctly appended when passing `ResponseInit` as a second parameter to `c.body(..., res)`.

Support for setting cookies was recently added to `@hono/trpc-server` in
https://github.com/honojs/middleware/pull/517, but this bug prevents multiple cookies from being set.

[Here's a reproduction](https://github.com/codeflows/hono-cookie-test/tree/main)

- [x] Add tests
- [x] Run tests


yusukebe (approved):
LGTM!

yusukebe:
@codeflows 

Thanks! Looks good to me. Merging now.

PR #2662, opened 2024-05-12
feat(jsr): support JSR

yusukebe:
This PR will enable Hono to support the [JSR](https://jsr.io/).

We've done:

* Added type annotations to reduce [Slow Types](https://jsr.io/docs/about-slow-types). https://github.com/honojs/hono/pull/2663 
* I added the `jsr-dry-run` command in CI to check if it's ready to publish JSR. https://github.com/honojs/hono/pull/2662/commits/8b247079ce0620aa487fa1efef15035214444298
* Added `deno.json`. https://github.com/honojs/hono/pull/2662/commits/7ac63060928d04beab51e5a3f94adaf02658f37e And modified.
* Made `JSX` be exported from `hono/jsx/jsx-runtime` not `global`. https://github.com/honojs/hono/pull/2662/commits/5bd2b71709a3e8d28a83c193fb70d7363b65dc85
* Made `hono-base.ts` does not use `dynamicClass`. https://github.com/honojs/hono/pull/2662/commits/b9025907175d31879e5f3648661d60536af1a787
* Made `ExecutionContext` is not declared in `global`. https://github.com/honojs/hono/pull/2662/commits/cf020c64c52df488393343efc06a5a2ce3bb6e5a
* Removed `deno_dist` and `denoify` completely. https://github.com/honojs/hono/pull/2662/commits/0c776fa22696b54cbc31c51c7262a0a056316499
* Made `deno.json` exports `./`, `./jsx/jsx-runtime`, ~~`./middleware`, and `./helper`~~.
* Made `deno.json` exports each helper, middleware, and adapter following @nakasyou's [comment](https://github.com/honojs/hono/pull/2662#issuecomment-2106218638). https://github.com/honojs/hono/pull/2662/commits/90892be2a00ca500d337dfa1fd78ceafa9612676

## Usage

Currently, we are uploading Hono as [`@hono/do-not-use-this` on the JSR](https://jsr.io/@hono/do-not-use-this).

```sh
deno add @hono/do-not-use-this
```

You don't use it, but you can use it for experimental purposes.

## We will support both npm and JSR

Though this PR will merged and Hono supports JSR, we don't stop supporting the npm. I'm thinking we can do it with the following publishing steps:

1. Publish the npm package with the `np` command on my local machine.
2. The new commit and tag will be pushed to this repo.
3. If the tag is created, the GitHub Action will execute and release the new version of the JSR package to the JSR. Can we do it with this? https://github.com/dsherret/jsr-publish-on-tag

[...]

PR #2694, opened 2024-05-17
fix(middleware): Context declaration

### LINKED ISSUE #2693: Type bug introduced in `v4.3.0` in `secureHeaders` middleware

4.3.0

Bun

1. Install version `4.3.0` of Hono.
2. Create a TS file with an app that uses `secureHeaders` middleware and another arbitrary middleware that sets the context (I used `@hono/clerk-auth` as an example). Also in this file you'll mount the routes from another file with `app.route()`.
3. In another file, create another app and try to access the context value with `context.get` from this other middleware, and see it fail, as the context only contains "secureHeadersNonce" variable.

`4.3.0` CodeSandbox with a bug: https://codesandbox.io/p/devbox/hono-4-3-0-not-working-9kg2s7

`4.2.9` CodeSandbox working fine: https://codesandbox.io/p/devbox/hono-4-2-9-working-vlx83h

One can access the context variable in another file.

### What do you see instead?

_No response_

### Additional information

I confirmed that the issue lies in the `secureHeaders` middleware, as by commenting it out everything works fine as before. I suspect [these lines](https://github.com/honojs/hono/blob/d87d9964433a4777b09abff9360ff12643d00440/src/middleware/secure-headers/index.ts#L4) to be the issue from PR #2577:
```ts
declare module '../../context' {
    secureHeadersNonce?: string
  }
}
```

It seems like this is overwriting the context variables set by other middleware, therefore it's impossible to access variables from other middleware without importing it directly (which would be an unused import technically).

MonsterDeveloper:
Fixes #2693 

@yusukebe not sure it's the best way though, as `denoify` automatically adds `node:` to `hono` import... Without this import TS yells that `hono` is an "unknown" module...

- [ ] ~Add tests~ (n/a)
- [x] Run tests

### DIFF
--- a/src/middleware/jwt/index.ts
+++ b/src/middleware/jwt/index.ts
@@ -5,7 +5,7 @@ import type { MiddlewareHandler } from '../../types'
 import { Jwt } from '../../utils/jwt'
 import type { SignatureAlgorithm } from '../../utils/jwt/jwa'
 
-declare module 'hono' {
+declare module '../..' {
   interface ContextVariableMap {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     jwtPayload: any
--- a/src/middleware/jwt/index.ts
+++ b/src/middleware/jwt/index.ts
@@ -3,10 +3,9 @@ import { getCookie } from '../../helper/cookie'
 import { HTTPException } from '../../http-exception'
 import type { MiddlewareHandler } from '../../types'
 import { Jwt } from '../../utils/jwt'
-import '../../context'
 import type { SignatureAlgorithm } from '../../utils/jwt/jwa'
 
-declare module '../../context' {
+declare module 'hono' {
   interface ContextVariableMap {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     jwtPayload: any
--- a/src/middleware/secure-headers/index.ts
+++ b/src/middleware/secure-headers/index.ts
@@ -1,7 +1,8 @@
 import type { Context } from '../../context'
 import type { MiddlewareHandler } from '../../types'
+import {} from 'hono'
 
-declare module '../../context' {
+declare module 'hono' {
   interface ContextVariableMap {
     secureHeadersNonce?: string
   }
--- a/src/middleware/secure-headers/index.ts
+++ b/src/middleware/secure-headers/index.ts
@@ -1,8 +1,7 @@
 import type { Context } from '../../context'
 import type { MiddlewareHandler } from '../../types'
-import {} from 'hono'
 
-declare module 'hono' {
+declare module '../..' {
   interface ContextVariableMap {
     secureHeadersNonce?: string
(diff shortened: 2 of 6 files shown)


yusukebe:
Hi @MonsterDeveloper thank you for the PR!

About the PR, can you wait for me to merge? Perhaps it will be fixed completely if Hono supports [JSR](https://github.com/honojs/hono/pull/2662).

src/middleware/jwt/index.ts
```diff
@@ -3,10 +3,9 @@ import { getCookie } from '../../helper/cookie'
 import { HTTPException } from '../../http-exception'
 import type { MiddlewareHandler } from '../../types'
 import { Jwt } from '../../utils/jwt'
-import '../../context'
 import type { SignatureAlgorithm } from '../../utils/jwt/jwa'
 
-declare module '../../context' {
+declare module 'hono' {
```
yusukebe, line 8: Maybe we should write the following:

```ts
declare module '../..'
```

src/middleware/secure-headers/index.ts
```diff
@@ -1,7 +1,8 @@
 import type { Context } from '../../context'
 import type { MiddlewareHandler } from '../../types'
+import {} from 'hono'
```
yusukebe, line 3: It's not necessary.

src/middleware/secure-headers/index.ts
```diff
@@ -1,7 +1,8 @@
...
 
-declare module '../../context' {
+declare module 'hono' {
```
yusukebe, line 5: The same as above:

```ts
declare module '../..'
```

src/middleware/timing/index.ts
```diff
@@ -1,8 +1,7 @@
 import type { Context } from '../../context'
 import type { MiddlewareHandler } from '../../types'
-import '../../context'
 
-declare module '../../context' {
+declare module 'hono' {
```
yusukebe, line 4: The same as above:

```ts
declare module '../..'
```

yusukebe:
@MonsterDeveloper 

I got it. We should extend `../..` instead of `../../context`. Could you fix them?

MonsterDeveloper:
Oh my lovely TypeScript:)

Corrected to `../..`. Seems to be working too.

MonsterDeveloper:
But now Deno is not happy...
CleanShot 2024-05-18 at 06 02 53@2x

yusukebe:
@MonsterDeveloper 

Oh... Then, wait until [this PR](https://github.com/honojs/hono/pull/2662) is merged. The next minor version may be released soon, and it will be merged at that time!

MonsterDeveloper:
@yusukebe hi! Seems like #2662 is merged now, can we resume the work on this PR?

yusukebe:
The issue was fixed. This PR is not needed. Thanks!

PR #2680, opened 2024-05-15
docs(jsdoc): add JSDoc for Middleware

yusukebe:
In this PR, let's write the JSDoc for all builtin Middleware.  Only Middleware is addressed here; other documentation than Middleware is done in other PRs. Example #2677.

Will fix #1338

---

- [x] [Basic Auth Middleware](https://hono.dev/middleware/builtin/basic-auth) https://github.com/honojs/hono/pull/2680/commits/3ba73862283b2c1e859782f67a9b77326b6659e6
- [x] [Bearer Authentication](https://hono.dev/middleware/builtin/bearer-auth)
- [x] [Body Limit](https://hono.dev/middleware/builtin/body-limit)
- [x] [Cache](https://hono.dev/middleware/builtin/cache)
- [x] [Compress](https://hono.dev/middleware/builtin/compress)
- [x] [CORS](https://hono.dev/middleware/builtin/cors)
- [x] [CSRF Protection](https://hono.dev/middleware/builtin/csrf)
- [x] [ETag](https://hono.dev/middleware/builtin/etag)
- [x] [JSX Renderer](https://hono.dev/middleware/builtin/jsx-renderer)
- [x] [JWT](https://hono.dev/middleware/builtin/jwt)
- [x] [Logger](https://hono.dev/middleware/builtin/logger)
- [x] [Method Override](https://hono.dev/middleware/builtin/method-override)
- [x] [Pretty JSON](https://hono.dev/middleware/builtin/pretty-json)
- [x] [Secure Headers](https://hono.dev/middleware/builtin/secure-headers)
- [x] [Timing](https://hono.dev/middleware/builtin/timing)
- [x] [Trailing Slash](https://hono.dev/middleware/builtin/trailing-slash)

PR #2703, opened 2024-05-18
docs(jsdoc): Server-Timing Middleware

goisaki:
This PR is to add JSDoc for Server-Timing Middleware.
Note that the target of the PR is not `main`.

Related:
- #1338
- #2680

### DIFF
--- a/deno_dist/middleware/timing/index.ts
+++ b/deno_dist/middleware/timing/index.ts
@@ -31,6 +31,45 @@ const getTime = (): number => {
   return Date.now()
 }
 
+/**
+ * Server-Timing Middleware middleware for Hono.
+ *
+ * @see {@link https://hono.dev/middleware/builtin/timing}
+ *
+ * @param {TimingOptions} [config] - The options for the timing middleware.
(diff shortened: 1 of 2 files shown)


deno_dist/middleware/timing/index.ts
```diff
@@ -31,6 +31,45 @@ const getTime = (): number => {
   return Date.now()
 }
 
+/**
+ * Server-Timing Middleware middleware for Hono.
+ *
+ * @see {@link https://hono.dev/middleware/builtin/timing}
+ *
+ * @param {Partial<TimingOptions>} [config] - The options for the timing middleware.
```
yusukebe, line 39: I think `Partial` is not easy to understand for users. So, I've refactored in  #2712. Can you check the latest code in `main` and fix the types?

goisaki:
Hi @yusukebe
Currently, the change to remove Partial is made only in JSDoc. To include these changes in the code, we need to merge honojs:main into honojs:docs/jsdoc-middleware.

(It is possible to do this merge in this PR, but it is not recommended because it will include out-of-scope changes.)

yusukebe (approved):

PR #2714, opened 2024-05-18
feat: decode percent-encoded path in getPath

usualoma:
#2672

This PR contains #2688 + https://github.com/honojs/hono/pull/2688#discussion_r1605780997

### Specification Changes

There are no major release-level changes, but there are some minor changes as follows

#### Symbols and multibyte characters can be included in the static string of the routing definition

When you define a routing as follows, it will now match, whereas before it did not match because the percent encoding for static strings was not taken into account when routing.

```ts
app.get('/🔥', (c) => c.text(`I'm hono`)
```

```
http://example.com/%F0%9F%94%A5
```

#### Requests made with a URL that are encoded more than necessary are now matched with a decoded URL

Suppose we have a routing definition like this,

```ts
app.get('/static', (c) => c.text('static)
```

Even if unreserved characters are extra encoded, they are now decoded and matched.

```
http://example.com/%73tatic // %73 is 's'
```

#### Regular expressions for path parameters are now tested against decoded characters

We do not think that this will break any applications, but applications that use regular expressions in path parameters should check when updating to see if they are affected.

Suppose we have a routing definition like this,

```ts
app.get('/about/:product{🔥|Hono}', (c) => c.text(`Hono is a framework`)
```

The following URLs match

```
http://example.com/about/%F0%9F%94%A5
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/utils/url.ts
+++ b/deno_dist/utils/url.ts
@@ -70,10 +70,23 @@ export const getPattern = (label: string): Pattern | null => {
 }
 
 export const getPath = (request: Request): string => {
-  // Optimized: indexOf() + slice() is faster than RegExp
   const url = request.url
-  const queryIndex = url.indexOf('?', 8)
-  return url.slice(url.indexOf('/', 8), queryIndex === -1 ? undefined : queryIndex)
+  const start = url.indexOf('/', 8)
+  let i = start
+  for (; i < url.length; i++) {
+    const charCode = url.charCodeAt(i)
+    if (charCode === 37) {
+      // '%'
+      // If the path contains percent encoding, use `indexOf()` to find '?' and return the result immediately.
+      // Although this is a performance disadvantage, it is acceptable since we prefer cases that do not include percent encoding.
+      const queryIndex = url.indexOf('?', i)
+      return decodeURIComponent(url.slice(start, queryIndex === -1 ? undefined : queryIndex))
+    } else if (charCode === 63) {
+      // '?'
+      break
+    }
+  }
+  return url.slice(start, i)
 }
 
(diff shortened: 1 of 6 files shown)


szmarczak (approved):

deno_dist/utils/url.ts
```diff
@@ -70,10 +70,25 @@ export const getPattern = (label: string): Pattern | null => {
...
+  let i = start
+  for (; i < url.length; i++) {
+    const charCode = url.charCodeAt(i)
+    if (charCode === 37) {
+      // '%'
+      // If the path contains percent encoding, use `indexOf()` to find '?' and return the result immediately.
+      // Although this is a performance disadvantage, it is acceptable since we prefer cases that do not include percent encoding.
+      const queryIndex = url.indexOf('?', i)
+      return decodeURI(
+        url.slice(start, queryIndex === -1 ? undefined : queryIndex).replace(/%25/g, '%2525')
```
szmarczak, line 84: From my benchmarks it seems like `str.includes('%25') ? str.replaceAll('%25', '%2525') : str` is faster than calling `.replace` with regex. Can you verify?

usualoma: Thank you. You are right!

It's faster to check in `.includes` first.
`s/replace/replaceAll/` didn't change in my benchmark results, so I'll leave `replace(//g)` as it is, as it is generally used elsewhere in Hono.

Applied in 0eeaa7ec

szmarczak (approved):

yusukebe (approved):

yusukebe:
@usualoma 

Thanks! I'll include this in the next minor release and merge it into the `next` branch later.

PR #2726, opened 2024-05-20
RFC(adapter/aws-lambda): generalize `LambdaEvent`, `EventProcessor`

### LINKED ISSUE #2723: hono/aws-lambda + @hono/graphql-server GET not working!

4.3.7

Lambda

Any idea why the GET method doesn't work?

```ts
import rootResolver from "./resolvers/resolvers"
import { GraphQLFileLoader } from "@graphql-tools/graphql-file-loader"
import { loadSchemaSync } from "@graphql-tools/load"
import { addResolversToSchema } from "@graphql-tools/schema"
import { graphqlServer } from "@hono/graphql-server"
import { serve } from "@hono/node-server"
import { Hono } from "hono"
import { handle } from "hono/aws-lambda"
import { compress } from "hono/compress"
import { cors } from "hono/cors"
import { logger } from "hono/logger"
import { prettyJSON } from "hono/pretty-json"
import path from "path"

const app = new Hono()

app.use(logger())
app.use(cors())
app.use(compress())
app.use(prettyJSON())

const schemaPath = path.join(__dirname, "schemas/schema.graphql")

const schema = loadSchemaSync(schemaPath, {
  loaders: [new GraphQLFileLoader()],
})

const schemaWithResolvers = addResolversToSchema({
  schema,
  resolvers: rootResolver,
})

app.use(
  "/graphql",
  graphqlServer({
    schema: schemaWithResolvers,
  }),
)

if (process.env.AWS_LAMBDA_FUNCTION_NAME) {
  console.log("Running in AWS Lambda environment.")
  exports.handler = handle(app)
} else {
  console.log("Server is running on http://localhost:3000")
  serve(app)
}

```

image

image

_No response_

### What do you see instead?

_No response_

### Additional information

_No response_

NamesMT:
Resolves #2723 - Edit: unsure, see https://github.com/honojs/hono/pull/2726#issuecomment-2120316641

Lambda's event could come from any source, e.g: manual invokations, AWS triggers like S3, EventBridge, etc. So it should have the type of `unknown`

This PR generalizes `EventProcessor` to accept an `unknown` event, with the old processor logic for "Request"-based event sources moved to `RequestEventProcessor`

Because the processor's logic was moved to a different file, for easier reviewing what have really changed, the maintainer could check specifically the commit: [fix: generalize LambdaEvent](https://github.com/honojs/hono/commit/9c489bf6a51cee03fe380656e4aa5f151cd5bb61)

---

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/src/adapter/aws-lambda/handler.ts
+++ b/src/adapter/aws-lambda/handler.ts
@@ -3,12 +3,12 @@ import crypto from 'crypto'
 import type { Hono } from '../../hono'
 import type { Env, Schema } from '../../types'
 
-import { encodeBase64 } from '../../utils/encode'
 import type {
   ApiGatewayRequestContext,
   ApiGatewayRequestContextV2,
   ALBRequestContext,
 } from './custom-context'
+import { getProcessor } from './processor'
 import type { LambdaContext } from './types'
 
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
@@ -18,7 +18,7 @@ globalThis.crypto ??= crypto
 export type LambdaEvent = unknown | LambdaRequestEvent
 export type LambdaRequestEvent = APIGatewayProxyEvent | APIGatewayProxyEventV2 | ALBProxyEvent
 
-// When calling HTTP API or Lambda directly through function urls
+// When calling API Gateway's HTTP API or Lambda directly through function urls
 export interface APIGatewayProxyEventV2 {
   version: string
   routeKey: string
@@ -41,7 +41,7 @@ export interface APIGatewayProxyEventV2 {
   }
 }
 
-// When calling Lambda through an API Gateway
+// When calling Lambda through an API Gateway's REST API
 export interface APIGatewayProxyEvent {
   version: string
   httpMethod: string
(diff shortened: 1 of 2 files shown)


NamesMT:
Hello @yusukebe,

Out of the PR's scope, is there an ETA to remove binding of top-level `requestContext`?, It seems to have marked as deprecated since [v3.10.0](https://hono.dev/getting-started/aws-lambda#before-v3-10-0-deprecated) but is still in the code and bloating it.
Github Search also doesn't show any repository using `.env.requestContext`

exoego:
As a contributor to adapter/aws-lambda, it would be great if tests added to demonstrate this change fixes #2723.

It appears that the issue reporter attempted to let Hono handling Lambda's Function URL:
image
We could gain a help from [a type for Function URL in `@types/aws-lambda`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/aws-lambda/trigger/lambda-function-url.d.ts).

In my opinion, adding a new type for Function URL is a straightforward solution for #2723.
I am not sure making Hono handle `unknown` (arbitrary events), including non-HTTP events on S3 or EventBridge, is reasonable from maintenance PoV.
I am new to Hono, so such use case might make sense, but I have no idea yet.

watany-dev:
Basically, I think that Hono's standard repository should handle HTTP requests. I personally disagree with making the role of the adaptor too fat.
As for non-http adaptor, I think it is good to manage it as non-http adaptor from @hono/aws/lambda for example.

watany-dev:
> Hello @yusukebe,
> 
> 
> 
> Out of the PR's scope, is there an ETA to remove binding of top-level `requestContext`?, It seems to have marked as deprecated since [v3.10.0](https://hono.dev/getting-started/aws-lambda#before-v3-10-0-deprecated) but is still in the code and bloating it.
> 
> Github Search also doesn't show any repository using `.env.requestContext`

Sorry, I failed to delete this in v4.

NamesMT:
@exoego 
You're right, this PR is more about adding support for "unknown" event processing than to fix #2723, I changed the PR's title to "Request For Comment" for now.

I'm also unsure about adding `unknown` event handling to the official Hono adapter, that's why I have previously created a personal fork of the adapter, though IMO Hono as a "framework", the official adapter supporting the platform's feature does make sense,
And now, because it didn't work with the featured graphql middleware, I created the PR.

However, coming back to the issue with the GraphQL middleware (#2723), I'm not familiar with GraphQL and previously I thought graphql connects using it's own protocol, hence the issue of "requestContext" undefined, but reading more about it, it's using POST HTTP method, Function URL is using "APIGateway V2 Payload" so it should be already supported and requestContext should exists. So we might have a different bug here.

I'll set up a reproduction and find the real cause for #2723 later when I have time.

Issue #2626, opened 2024-05-06
`headers` and `multiValueHeaders` are not set properly when using `hono/aws-lambda` with ALB


yiss:
### What version of Hono are you using?

^4.3.2

### What runtime/platform is your app running on?

Node, AWS Lambda

### What steps can reproduce the bug?

When using `hono/aws-lambda`adapter with a Lambda that has an ALB in front of it. It shouldn't use the `multiValueHeaders` as default, as this feature is not default when using ALB with Lambda target. Moreover, the `set-cookie` behavior in the adapter might result in bugs in some cases.

Example : 

```typescript
import { handle } from "hono/aws-lambda";
import { getCookie, setCookie } from "hono/cookie";
import { Hono } from "hono";

const app = new Hono();

app.get("/set-headers", (ctx) => {
  setCookie(ctx, "foo", "bar", {
    secure: process.env.NODE_ENV === "production",
    domain: `.${process.env.COOKIE_DOMAINE}`,
    httpOnly: true,
    maxAge: 60 * 10,
    sameSite: "Lax",
  });
  setCookie(ctx, "fizz", "buzz", {
    secure: process.env.NODE_ENV === "production",
    domain: `.${process.env.COOKIE_DOMAINE}`,
    httpOnly: true,
    maxAge: 60 * 10,
    sameSite: "Lax",
  });
  return ctx.redirect("/get-headers");
});

app.get("/get-headers", (ctx) => {
  const fooCookie = getCookie(ctx, "foo");
  const fizzCookie = getCookie(ctx, "fizz");
  return ctx.json({
    fooCookie,
    fizzCookie,
  });
});

export const handler = handle(app);
```

And CDK minimal config :
```typescript
import * as cdk from '@aws-cdk/core';
import * as ec2 from '@aws-cdk/aws-ec2';
import * as lambda from '@aws-cdk/aws-lambda';
import * as lambdaNodejs from '@aws-cdk/aws-lambda-nodejs';
import * as elbv2 from '@aws-cdk/aws-elasticloadbalancingv2';

export class CdkAlbLambdaStack extends cdk.Stack {
  constructor(scope: cdk.Construct, id: string, props?: cdk.StackProps) {
    super(scope, id, props);

    // Create VPC
    const vpc = new ec2.Vpc(this, 'MyVpc', {
      maxAzs: 2
    });

    // Create Lambda function
    const lambdaFunction = new lambdaNodejs.NodejsFunction(this, 'MyLambda', {
[...]

PR #2751, opened 2024-05-20
feat(aws-lambda): add support for alb multiValueQueryStringParameters

yiss:
When working on #2626, I realized that when `multiValueHeaders` is enabled, the query string parameters are passed as an array as well. More info : https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html#multi-value-headers

> If you enable multi-value headers, the load balancer uses both key values sent by the client and sends you an event that includes query string parameters using multiValueQueryStringParameters. For example:
`"multiValueQueryStringParameters": { "myKey": ["val1", "val2"] },`

This PR adds support for `multiValueQueryStringParameters` when it's available on the event

- [x] Add tests
- [x] Run tests

### DIFF
--- a/runtime_tests/lambda/index.test.ts
+++ b/runtime_tests/lambda/index.test.ts
@@ -108,6 +108,16 @@ describe('AWS Lambda Adapter for Hono', () => {
     return c.json(lambdaContext)
   })
 
+  app.get('/query-params', (c) => {
+    const queryParams = c.req.query()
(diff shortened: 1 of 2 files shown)


yusukebe (approved):
LGTM!

yusukebe:
Hi! @yiss 

Thanks for the PR. I'll merge it now and soon release a patch version that includes this change.

PR #2767, opened 2024-05-22
fix(http-exception): prioritize the status code

### LINKED ISSUE #2707: HTTPException status is ignored if a Response is provided

4.3.7

Deno

```ts
import { HTTPException, Hono } from "hono";

const app = new Hono();

// Should throw 400, status is 200
app.get("/example-a", (c) => {
  throw new HTTPException(400, {
    res: new Response("An exception", { status: 200 }),
  });
});

// Should throw 401, status is 200
app.get("/example-b", (c) => {
  throw new HTTPException(401, {
    res: c.text("An exception"),
  });
});

// Correctly throws 401
app.get("/example-c", (c) => {
  throw new HTTPException(401, {
    res: c.text("An exception", 401),
  });
});

app.onError((err, c) => {
  if (err instanceof HTTPException) {
    return err.getResponse();
  }

  console.error(`${err}`);
  return c.text("Internal Server Error", 500);
});

Deno.serve(app.fetch);

```

The behavior I expected is that the status provided to `HTTPException` would take precedent over response, especially if `c.text()`, `c.json()`, or `c.html()` are used, since those all set a `200` status by default.

### What do you see instead?

The status of the response provided in the `res` argument always takes precedence.

### Additional information

Unless this is intended behavior, I would propose `HTTPException` always return a response with the status code provided in the first argument.

Note: I did some cursory research into modifying the status of a Response instance. It appears modifying a Response status isn't possible, and clones of responses have read-only statuses as well. It would seem that implementing this change would necessitate creating a new Response, then applying the body and headers from the given Response, but ignoring the status and instead using the status given to `HTTPException`.

yusukebe:
Fixes #2707

- [x] Add tests
- [x] Run tests

### DIFF
--- a/deno_dist/http-exception.ts
+++ b/deno_dist/http-exception.ts
@@ -36,7 +36,11 @@ export class HTTPException extends Error {
 
   getResponse(): Response {
     if (this.res) {
-      return this.res
+      const newResponse = new Response(this.res.body, {
+        status: this.status,
+        headers: this.res.headers,
+      })
+      return newResponse
     }
     return new Response(this.message, {
       status: this.status,
--- a/src/http-exception.test.ts
+++ b/src/http-exception.test.ts
@@ -31,4 +31,15 @@ describe('HTTPException', () => {
     expect(exception.message).toBe('Internal Server Error')
     expect(exception.cause).toBe(error)
   })
+
+  it('Should prioritize the status code over the code in the response', async () => {
+    const exception = new HTTPException(400, {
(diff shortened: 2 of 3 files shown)


yusukebe:
@usualoma 

Could you review this too?

usualoma:
@yusukebe Thank you. LGTM 👍

yusukebe:
@usualoma Thanks! Merging.

PR #2780, opened 2024-05-24
refactor(request): show user-friendly type for `c.req.param()`

usualoma:
### Before

CleanShot 2024-05-24 at 17 46 31@2x

### After

CleanShot 2024-05-24 at 17 46 49@2x

- <del>Add tests</del>
    - The [tests](https://github.com/honojs/hono/blob/main/src/hono.test.ts#L2342-L2352) we have now are sufficient.
- [x] Run tests


usualoma:
@yusukebe How about this?

yusukebe:
@usualoma 

Looks good to me!

Let's merge. But right now, shall we merge it into the `next`?

usualoma:
@yusukebe  Understood!
Merged next into this branch.

yusukebe:
@usualoma 

Yeah. `denoify` is no longer needed! Merging.

PR #2787, opened 2024-05-25
feat(middleware): Introduce IP Limit Middleware

nakasyou:
I created IP Limit Middleware.

You can limit request by IP Address.

For example, you can limit request, this server accepts local-only requests:
```ts
import { Hono } from 'hono'

const app = new Hono()

app.use('*', ipLimit(getConnInfo, {
  deny: [],
  allow: ['127.0.0.1', '::1']
}))
app.get('/', c => c.text('Hello world!'))
```
`deny` takes precedence over `allow`.

Rules supported some syntax:

| Title | example of IPv4 | example of IPv6 |
| --- | --- | --- |
| static | `0.0.0.0` | `::1` |
| CIDR | `192.168.2.1/24` | `abcd::ef01/64` |
| Wildcard | `192.*.2.*` |  |

- [x] Add tests
- [x] Run tests

### DIFF
--- a/jsr.json
+++ b/jsr.json
@@ -21,6 +21,7 @@
     "./basic-auth": "./src/middleware/basic-auth/index.ts",
     "./bearer-auth": "./src/middleware/bearer-auth/index.ts",
     "./body-limit": "./src/middleware/body-limit/index.ts",
+    "./ip-limit": "./src/middleware/ip-limit/index.ts",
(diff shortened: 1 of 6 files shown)


yusukebe:
@nakasyou 

Thanks! I'll check this after releasing the`v4.4.0` very soon.

PR #2788, opened 2024-05-25
chore(deno): change path including deno_dist

ryuapp:
Fix path including deno_dist.
`--unstable-sloppy-imports` flag is now required on benchmark due to jsr aliases.

- [ ] Add tests
- [x] Run tests


yusukebe:
@ryuapp 

Thanks!

Issue #664, opened 2022-11-16
Feature request: support for file upload


skitsanos:
I have a played a bit here with `bun` and `hono` to see how it works and realized that the request body comes there in raw format if I want to upload files.

surprisingly, getting body as a buffer worked out of the box:

```ts
 const reader = ctx.req.body.getReader()
 const result = await reader.read()
 const buffer = Buffer.from(result.value ?? [])
```

So I have my `buffer`, but the question is -- do we have anything is there any functionality to have a list of files within that buffer?

In other words, - maybe there is an example of implementing file uploading with `hono`, or at least some timeline on when it will be available.

Thank you.

yusukebe:
Hi @skitsanos !

Do you want to upload and get files with `multipart/form-data`? If so, you can use [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData), but unfortunately, it is not implemented in Bun yet. Below is an example using `FormData`.

```ts
app.post('/form-upload', async (c) => {
  const formData = await c.req.formData()
  const file = formData.get('file')
  const arr = await file.arrayBuffer()

  fs.writeFile('foo.png', Buffer.from(arr), (err) => {
    if (err) throw err
  })

  return c.text('uploaded!')
})
```

This is working well on Node.js with [node-server](https://github.com/honojs/node-server).

I'll try to figure out how to implement this without FormData. Regardless, I think the best way is that Bun supports `FormData`.

skitsanos:
Yes, Yusuke, I wanted it to run via `bun`... With `node`, it would be a simpler task because there are tons of multipart parsers out there already and the task itself is not a complicated one at all, - `boundary` comes in HTTP header, so you split the request body by line that contains boundary... This would give you form values and files that were sent...

yusukebe:
I think we have to use the web-standard API as much as possible. In the case of Bun, it seems to be difficult to import `FormData` from WebKit now.

https://github.com/oven-sh/bun/issues/621

>  so you split the request body by line that contains boundary

Hmm... :(

skitsanos:
What does parsing the request body has to do with that?:) I just explained the flow of how to make it happen :)

yusukebe:
Ah, yes. I know it:)

yusukebe:
@skitsanos 

How about doing "polyfill"? I have not used it, but there is a library like this:

https://www.npmjs.com/package/formdata-polyfill

> In hono you could add multipart parsing as wasm done from some rust lib...

😱

[...]

PR #2791, opened 2024-05-25
fix(client): allow multiple files on the same key in form

olivier-drieux:
In addition to the [#664](https://github.com/honojs/hono/issues/664), the client did not manage multiple files on a single key in FormData. It's quite the same code as `query`.

- [ ] Add tests
- [x] Run tests


yusukebe:
Hi @olivier-drieux 

Looks good! I've added the test (and updated the test library `msw`). This is okay. Merging now. Thanks!

PR #2795, opened 2024-05-26
feat(jsx/dom): introduce react-dom/client APIs and React.version

### LINKED ISSUE #2792: Missing pieces for React compatibility
### version

In React we can get the version string as follows, but not in hono/jsx

```ts
import { version } from 'react';
```

https://github.com/facebook/react/blob/main/packages/shared/ReactVersion.js#L15

### react/dom/client

The createRoot and hydrateRoot APIs are not implemented in hono.

```ts
import { createRoot, hydrateRoot } from 'react-dom/client';
```

usualoma:
Fixes #2792

### version

There seems to be a library that does conditional branching on the major version number of `React.version`.

This is like a UserAgent of a web browser (e.g. Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/125.0.0.0 Safari/537. 36), which is a string for compatibility and somewhat unpleasant to use, and we do not recommend its use, but if this improves compatibility, it should be added.

### createRoot / hydrateRoot

Some options are not yet supported, but they can be rendered in the way recommended by React since version 18, as shown in the following diff.

https://github.com/usualoma/hono-react-compat-demo/commit/38773555179ef54f8564ca1963720ac16f720987

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/jsx/dom/client.ts
+++ b/src/jsx/dom/client.ts
@@ -3,11 +3,17 @@
  * This module provides APIs for `hono/jsx/dom/client`, which is compatible with `react-dom/client`.
  */
 
+import type { Child } from '../base'
 import { useState } from '../hooks'
 import { buildNode, renderNode } from './render'
 import type { NodeObject } from './render'
 
-type CreateRootOptions = Record<string, unknown>
+export interface Root {
+  render(children: Child): void
+  unmount(): void
+}
+export type RootOptions = Record<string, unknown>
+
 /**
  * Create a root object for rendering
  * @param element Render target
@@ -16,8 +22,8 @@ type CreateRootOptions = Record<string, unknown>
  */
 export const createRoot = (
   element: HTMLElement | DocumentFragment,
-  options: CreateRootOptions = {}
-) => {
+  options: RootOptions = {}
+): Root => {
   let setJsxNode:
     | undefined // initial state
     | ((jsxNode: unknown) => void) // rendered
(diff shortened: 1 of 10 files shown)


usualoma:
@yusukebe How about including this in v4.4?

src/jsx/dom/client.ts
```diff
@@ -0,0 +1,73 @@
+import { useState } from '../hooks'
```
yusukebe, line 1: Add a [module doc](https://jsr.io/docs/writing-docs#module-documentation) at the top, like #2796.

yusukebe (commented):
I've left a comment. And can you update the `exports` field in `jsr.json`?

yusukebe:
> @yusukebe How about including this in v4.4?

Nice! Let's include this.

src/jsx/dom/client.ts
usualoma: OK, I added it in 61e2f7e.

yusukebe:
@usualoma 

One thing. Please update the `exports` field in [the `jsr.json`](https://github.com/honojs/hono/blob/next/jsr.json)!

usualoma:
I'm fixing it, just give me a minute.

usualoma:
Sorry I'm late, fixed. 2e0dbca

yusukebe (approved):
LGTM!

yusukebe:
@usualoma 

Thank you! Now, I'll merge this into the next. The `v4.4` will be released tomorrow!

Issue #2762, opened 2024-05-22
Setting cookie with Max-Age > 400 days throws


hmnd:
### What version of Hono are you using?

4.3.9

### What runtime/platform is your app running on?

Cloudflare Workers

### What steps can reproduce the bug?

Set cookie with Max-Age > 400 days.

### What is the expected behavior?

Cookie is set without throwing an error.

### What do you see instead?

An error is thrown.

### Additional information

This has caused an issue with [supabase's auth library](https://github.com/supabase/auth-helpers/pull/776) and could affect similar usage of cookies in other libs/projects.

Given the RFC is targeted primarily at User Agents ([_The user agent MUST limit the maximum value of the Max-Age attribute_](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-13#section-4.1.2.2-2)), I don't believe it's wise to force this onto users of Hono.

EdamAme-x:
can you tell me details of error?

hmnd:
@EdamAme-x here: https://github.com/honojs/hono/blob/main/src%2Futils%2Fcookie.ts#L152-L158

EdamAme-x:
Thanks @hmnd .
Understood.
I think we can choose to remove the "Max-Age" restriction, 
since specifying more than 400 days does not result in an explicit error.

EdamAme-x:
ready just now @hmnd

ryuapp:
Is there any good in being able to set max-age to more than 400 days?

Chrome will automatically change the max-age to 400 days if it exceeds 400 days. 
In other words, it behaves differently than the value set in the webframework.
So I don't think restrictions should be removed if there is no benefit.

The restriction was added in the following PR: https://github.com/honojs/hono/pull/2314

Related:
https://developers.google.com/privacy-sandbox/blog/progress-in-the-privacy-sandbox-2022-04#additional_cookie_updates

hmnd:
@ryuapp I think the more important question is, is there any benefit to throwing when a max age > 400 is set? If you review my earlier comments, you can see this has caused me grief trying to use Hono with Supabase Auth: https://github.com/supabase/auth-helpers/issues/441#issuecomment-2046645189

EdamAme-x:
hi @ryuapp .
I don't see the need to throw an error if it is automatically 400 days, as I think it is.

ryuapp:
[...]

PR #2797, opened 2024-05-26
fix(utils): remove limit of max-age

EdamAme-x:
see: https://github.com/honojs/hono/issues/2762

- [ ] Add tests
- [x] Run tests


EdamAme-x:
RE:
> I think we can choose to remove the "Max-Age" restriction, since specifying more than 400 days does not result in an explicit error.

PR #2800, opened 2024-05-26
refactor: change to short-hand

EdamAme-x:
For readability
I don't change "if syntax" of "return with handler syntax" and "assign syntax".

- [ ] Add tests
- [x] Run tests


EdamAme-x:
I check config of eslint again.

PR #2806, opened 2024-05-27
fix(hono-base): remove any of .use

EdamAme-x:

- [ ] Add tests
- [x] Run tests


EdamAme-x:
base by https://github.com/honojs/hono/pull/2802

yusukebe:
Hi @EdamAme-x 

Thanks for the PR.

I think this fix is not needed. It should not be `E`, and it's okay to use `any` at the line since it's just for implementation, not type definition.

The actual type definitions are on `types.ts`, and `MiddlewareHandlerInterface` is more complex.

EdamAme-x:
I understood. OK

PR #2799, opened 2024-05-26
feat(context): c.redirect support punycode and URL object

EdamAme-x:
### The author should do the following, if applicable

- [x] Add tests
- [x] Run tests
- [x] `bun run format:fix && bun run lint:fix` to format the code

PR #2808, opened 2024-05-27
feat(context): c.redirect support punycode and URL object

EdamAme-x:

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/context.test.ts
+++ b/src/context.test.ts
@@ -52,6 +52,18 @@ describe('Context', () => {
     expect(res.headers.get('Location')).toBe('https://example.com/destination')
   })
 
+  it('c.redirect() - URL', async () => {
+    let res = c.redirect(new URL("https://hono.dev/"))
+    expect(res.status).toBe(302)
+    expect(res.headers.get('Location')).toBe('https://hono.dev/')
+  })
+
+  it('c.redirect() - Punycode', async () => {
+    let res = c.redirect(new URL("https://🔥.hono.dev"))
+    expect(res.status).toBe(302)
+    expect(res.headers.get('Location')).toBe('https://xn--4v8h.hono.dev/')
+  })
+
   it('c.header()', async () => {
     c.header('X-Foo', 'Bar')
     const res = c.body('Hi')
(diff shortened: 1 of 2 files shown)


EdamAme-x:
based by https://github.com/honojs/hono/pull/2799

yusukebe:
Hi @EdamAme-x 

Thank you for the PR.

Supporting `URL` seems good. But what if it does not create a URL object when it's a `string`?

```diff
diff --git a/src/context.ts b/src/context.ts
index ff4299f..41dee41 100644
--- a/src/context.ts
+++ b/src/context.ts
@@ -558,11 +558,6 @@ export class Context<
   redirect = (location: string | URL, status: RedirectStatusCode = 302): Response => {
     if (location instanceof URL) {
       location = location.href
-    } else {
-      try {
-        const cachedURL = new URL(location)
-        location = cachedURL.href
-      } catch (_e) {} // Do nothing
     }
     this.#headers ??= new Headers()
     this.#headers.set('Location', location)
```

With this implementation, the performance degradation is less if the argument is `string`.

@usualoma Any thoughts?

EdamAme-x:
thanks @yusukebe .
If it is passed as "string", it would eliminate performance problems if it could be branched out into whether it is a URL or just a path.

EdamAme-x:
There is a way to simply check if the "://" is included, what do you think?

EdamAme-x:
However, this still reduces performance in cases such as "https://hono.dev".
Is a delay of 20 μs within the range that can be overlooked?

EdamAme-x:
I noticed that I can determine if punycode should be used when using regex.

yusukebe:
@EdamAme-x 

If the user wants to use Punycode, the user can use a URL object instead of `string`. So, I think we don't have to support it when it's `string`.

> Is a delay of 20 μs within the range that can be overlooked?

Indeed, it's a short time, but in some cases, the `redirect` is used heavily. We want to avoid creating `URL` as much as possible.

EdamAme-x:
> @EdamAme-x
> 
> If the user wants to use Punycode, the user can use a URL object instead of `string`. So, I think we don't have to support it when it's `string`.
> 
> > Is a delay of 20 μs within the range that can be overlooked?
> 
> Indeed, it's a short time, but in some cases, the `redirect` is used heavily. We want to avoid creating `URL` as much as possible.

I agree. I'll write document this

PR #2813, opened 2024-05-27
feat(middleware): Introduce IP Limit Middleware

nakasyou:
Recreated #2807 
***

I created IP Limit Middleware.

You can limit request by IP Address.

For example, you can limit request, this server accepts local-only requests:
```ts
import { Hono } from 'hono'

const app = new Hono()

app.use('*', ipLimit(getConnInfo, {
  deny: [],
  allow: ['127.0.0.1', '::1']
}))
app.get('/', c => c.text('Hello world!'))
```
`deny` takes precedence over `allow`.

Rules supported some syntax:

| Title | example of IPv4 | example of IPv6 |
| --- | --- | --- |
| static | `0.0.0.0` | `::1` |
| CIDR | `192.168.2.1/24` | `abcd::ef01/64` |
| Wildcard | `192.*.2.*` |  |

- [x] Add tests
- [x] Run tests


EdamAme-x:
I think it would be good to have `*` automatically when `allow` is empty or undefined.
And I can also suggest a middleware called `request-limit` based on this.
https://github.com/nakasyou/hono/pull/1

nakasyou:
Hi @EdamAme-x, thank you for comment and PR.
> I think it would be good to have * automatically when allow is empty or undefined.

Yes, I agree.

> And I can also suggest a middleware called request-limit based on this.

I think that renaming is not needed. And what do you think about allowing to receive `(conninfo: ConnInfo) => boolean` as `IPLimitRule`? It's smarter than your suggestion in PR for me.

Also, your PR includes 2 suggestions about allowing to add `*` by automatic and adding `denyHandler` and `validHandler`. Could you separation that?

EdamAme-x:
thanks for reply
I understood. Wait a moment...

EdamAme-x:
> I think that renaming is not needed. And what do you think about allowing to receive (conninfo: ConnInfo) => boolean as IPLimitRule? It's smarter than your suggestion in PR for me.  

nice suggestion! I like this

EdamAme-x:
RFC (translated)
https://tex2e.github.io/rfc-translater/html/rfc9110.html#15-5-4--403-Forbidden

EdamAme-x:
@nakasyou I just ready to review
https://github.com/nakasyou/hono/pull/3

usualoma:
Hi @nakasyou 
Great middleware!

### Wildcard

Is “Wildcard” an important feature of this middleware? This is because, although the use of wildcards is seen as customary, I do not believe that such a notation is defined as an official specification.

Patterns with trailing `*`, such as `192.168.2.*`, we can specify in the CIDR. 
The pattern `192.*.2.*` is, um, probably not used in real-world operations.

If you have a special preference for “Wildcard," you may support It, but otherwise, I think that not supporting it will reduce the amount of code and make future maintenance easier.

EdamAme-x:
@usualoma 
In my opinion, it needs to be a feature, just like Cloudflare supports wildcards.
However, I would prefer moving the function to another file. (e.x., “. /utils.ts”)

usualoma:
@EdamAme-x 
Thanks!
Sorry for my lack of knowledge, can you tell me in what scenarios Cloudflare allows you to use `*` for IP addresses? (I can't seem to find the information)

EdamAme-x:
sorry, I guess I didn't check enough,
wildcard is already obsolete and CIDR notation seems to be in use.

ryuapp:
Hi @nakasyou. Nice PR :)

How about reconsidering your naming conventions?
IMO, it is necessary to improve naming convention, since it is APIs that is exposed externally as part of utils.
e.g.)
```
ip-limit // this helper actually looks like an ip access manager.
distinctionRemoteAddr // "distinction" is a noun.
expandIPv6, ipV6ToBinary // variation in spelling(IPv6, ipV6).
```
I have a few other opinions, I'll leave them later.

fzn0x:
> How about reconsidering your naming conventions?

Agreed on this, I think we should go with `ip-rate-limit`, because it offers similar functionality.

nakasyou:
@usualoma, I removed wildcard as you said.

nakasyou:
Hi @ryuapp, thank you for review.
I fixed some spelling errors from your comment.

Also I'd like to discuss about name of the middleware.
@fzn0x should name the middleware `rate-limit`, but I think the middleware doesn't do rate-limit. I think rate-limit is things that limit the number of accesses, what do you think?

fzn0x:
> Hi @ryuapp, thank you for review. I fixed some spelling errors from your comment.
> 
> Also I'd like to discuss about name of the middleware. @fzn0x should name the middleware `rate-limit`, but I think the middleware doesn't do rate-limit. I think rate-limit is things that limit the number of accesses, what do you think?

I don't have other suggestions other than `ip-rate-limit`, maybe you have an idea for other names.

johnforte:
> I don't have other suggestions other than `ip-rate-limit`, maybe you have an idea for other names.

Why not just ip-deny-middleware? Those make more sense than an ip-rate-limit.

nakasyou:
Hi @johnforte, thank you for proposing.

> Why not just ip-deny-middleware? Those make more sense than an ip-rate-limit.

It maybe suitable. However naming it `middleware` is not smart, so I think we should name it `ip-deny` if we name it `ip-deny-middleware`

usualoma:
Hi @nakasyou 

I think `ip-limit` is fine and a good enough option, although some of the others' suggestions get a nod.

### Another option

By the way, I was thinking, what about turning this middleware into a middleware for "access control" that is not limited to IP addresses?

Assuming that the first level of support is restriction by IP address, it also supports authorization by authorization headers and cookies. We envisage something that would also be renamed to "access-control" and could be used as follows.

It could be used in production environment to "allow access if either the IP address or another authentication method matches".

```ts
import { Hono } from 'hono'
import { basic as acBasic, cookie as acCookie, accessControl } from 'hono/access-control'

const app = new Hono()

app.use(
  '*',
  accessControl(getConnInfo, {
    allow: [                         // allow if satisfy any of the following
      '192.168.0.2',                 // allow by IP address
      acCookie('auth', secretValue), // allow by cookie
      acBasic('user', 'password'),   // allow by basic auth
    ],
  })
)
app.get('/', (c) => c.text('Hello world!'))
```

Although the functionality overlaps with existing "basic-auth" and "bearer-auth", it may be possible to co-operate reasonably while sharing code.

If this option is adopted, control by cookies and request headers can be considered in a separate PR. In this PR, only the name should be changed and remain "middleware that restricts by IP address."

EdamAme-x:
> ### 別のオプション
> ちなみに、このミドルウェアを、IP アドレスに限定されない「アクセス制御」用のミドルウェアにしたらどうだろうと考えていました。
> 
> 最初のレベルのサポートが IP アドレスによる制限であると仮定すると、認証ヘッダーと Cookie による認証もサポートされます。また、名前が「アクセス制御」に変更され、次のように使用できるものを想定できます。
> 

In my opinion, restrictions in cookies should be left to other middleware.

EdamAme-x:
I would like to be able to pass IP address rules in this format as well. (for DataBase)
`(remote: NetAddrInfo) => boolean`.

nakasyou:
Hi @usualoma, thank you for proposing, but I don't agree.
I have 2 reasons.

First, as @EdamAme-x said, 
> But in my opinion, restrictions in cookies and headers should be left to other middleware.
Hono-users may be confused as to which middleware to use.

I agree that. It is reasonable to separate the IP address from other authentication methods for me.
Currently we can limit by cookies and basic auth using middleware.

Second, users maybe not using IP address control, users may use only `acCookie` or `acBasic`. In this case, `getConnInfo` is not needed. Also it doesn't when code runs on runtime not supported `getConnInfo` such as Node.js or Vercel.

So I don't agree about your proposing.

fzn0x:
> ```ts
> controlIPAddress(getConnInfo, ['192.168.0.2']),    // allow by IP address
> ```

Hi @nakasyou it sounds like you can control the IP address.

Actually I have a new idea for the name!

It is....

`ip-restrict` 🎉 

Because as I remember `restriction` is the most common naming when it comes to limit access to a network by IP.

fzn0x:
ip-limit sounds like the IP have some limits.

other than ip-restrict we can try `limit-ip-access`

usualoma:
@EdamAme-x @nakasyou Thanks for the explanation!

Indeed, my proposal was mixing too many things.

I was hoping to have a control similar to Apache's Satisfy Directive. As it is currently not possible to get `Satisfy Any` behaviour just by combining the middleware provided in hono.
https://httpd.apache.org/docs/2.4/en/mod/mod_access_compat.html#Satisfy


### I calmed down and thought about it.

As for `Satisfy Any`,  I have come to the conclusion that it is better not to mix it with this PR. I think This PR should remain as it is now, simple IP address restriction only. (Sorry for the confusion.)

It would be nice if `Satisfy Any` could be provided in a separate, independent middleware and combined with other middleware. I envisage the following APIs.

```ts
import { Hono } from 'hono'
import { basicAuth } from 'hono/basic-auth'
import { satisfyAny } from 'hono/access-control' // New middleware

const app = new Hono()

app.use(
  '*',
  satisfyAny([
    ipLimit(getConnInfo, { allow: [ '192.168.0.2' ] },
    basicAuth({ username: 'hono', password: 'acoolproject' })
  })
)
app.get('/', (c) => c.text('Hello world!'))
```

```ts
import { Hono } from 'hono'
import { bearerAuth } from 'hono/bearer-auth'
import { rateLimit } from '@app/my-rate-limit'
import { satisfyAny, satisfyAll } from 'hono/access-control' // New middleware

const app = new Hono()

// Very complex access restriction rules.
app.use(
  '*',
  satisfyAny([
    satisfyAll([
      ipLimit(getConnInfo, { allow: [ '192.168.0.2' ] },
      bearerAuth({ token })
    ]), // If both are satisfied, rateLimit is not necessary.
    rateLimit()
  })
)
app.get('/', (c) => c.text('Hello world!'))
```

ryuapp:
I'm sorry for being late, allow me to leave comments.

## Property name
Sorry for pointing out names again and again, but `allowList/denyList` expresses the nature better than `allow/deny`.


## Custom Error
There is no need to implement it with this pull request, but it would be more convenient to be able to set a custom error.

## Middleware name
`limit` means a restriction on the size or amount of something permissible or possible.
IMO, `restriction` is better than `limit`. The problem is that the name is long.

## Dependency on another middleware
It is undesirable to depend on `getConnInfo()`.
This middleware exists only for IP restriction, so I don't think there is any need to depend on other middleware.
I think users can use this middleware in any way they want by needing only for an IP.
```ts
import { Hono } from 'hono'
import { ipRestriction } from 'hono/ip-restriction'

const app = new Hono()

app.use('*', ipRestriction('127.0.0.1', {
  deny: [],
  allow: ['127.0.0.1', '::1']
}))
app.get('/', c => c.text('Hello world!'))
```

nakasyou:
@usualoma
> It would be nice if Satisfy Any could be provided in a separate, independent middleware and combined with other middleware. I envisage the following APIs.

Yes, I agree for it.

@ryuapp
> ## Property name

It looks good for me, but it is longer, so discussing is needed.

> ## Custom Error

It's good if we can do that.


> ## Middleware name
> limit means a restriction on the size or amount of something permissible or possible.
IMO, restriction is better than limit. The problem is that the name is long.

I had thought that `ip-limit` is good, but I read your comment, I agree for your opinion.
@fzn0x, could you tell me what you think?

> ## Dependency on another middleware

Hmm, I can't understand.
Can you explain in more detail?

ryuapp:
@nakasyou

> ## Dependency on another middleware
> Hmm, I can't understand.
> Could you explain in more detail?

I thought about it again, I'm not really worried about dependencies, but rather the use of `getConnInfo()`.
`getConInfo()` only works on certain runtimes. Therefore, this middleware only works with certain runtimes(Bun, Deno, Cloudflare workers).
In order to support many runtimes, wouldn't it be better to only require an IP address?

nakasyou:
Hi @ryuapp
> In order to support many runtimes, wouldn't it be better to only require an IP address?

I don't understand your thinking well. Could you show me some examples about that?

nakasyou:
Hi @ryuapp 
Currently, we can define custom ConnInfo getter.
```ts
import type { GetConnInfo } from 'hono/conninfo'

const getConnInfo: GetConnInfo = (c) => {
  return {
    remote: {
      address: c.req.header('X-Forwarded-For')
    }
  }
}
```
If you feel it is so long, what do you think about receiving both:
```ts
app.use('*', ipRestriction(getIP, {
  deny: [],
  allow: ['127.0.0.1', '::1']
}))

app.use('*', ipRestriction(getConnInfo, {
  deny: [],
  allow: ['127.0.0.1', '::1']
}))
```

ryuapp:
@nakasyou 
> Currently, we can define custom ConnInfo getter.

You are right, we can difine the custom.

The question for me is why do I use `getConnInfo()` for this middleware?
This middleware can be implemented as long as we can get an IP address.
ConnInfo has address type(ipv4 or ipv6), but this currently only works with Bun.
That looks a weak reason to use ConnInfo.

I would respect it if ConnInfo object is based on web-standards, but if it's not, my opinion is that it shouldn't depend on any particular helper.

> If you feel it is so long, what do you think about receiving both:
> 
> ```ts
> app.use('*', ipRestriction(getIP, {
>   deny: [],
>   allow: ['127.0.0.1', '::1']
> }))
> 
> app.use('*', ipRestriction(getConnInfo, {
>   deny: [],
>   allow: ['127.0.0.1', '::1']
> }))
> 
> ```

It means that the middleware requires a function to get an IP address or ConnInfo, right?
This is a proposal based on convenience. Looks good to me.

nakasyou:
@ryuapp 
> I would respect it if ConnInfo object is based on web-standards, but if it's not, my opinion is that it shouldn't depend on any particular helper.I would respect it if ConnInfo object is based on web-standards, but if it's not, my opinion is that it shouldn't depend on any particular helper.

I see, and I agree for that. I understood!

> It means that the middleware requires a function to get an IP address or ConnInfo, right?

Yes. 

> This is a proposal based on convenience. Looks good to me.

Thank you for agreeing, I'll implement based on that.

nakasyou:
I renamed `ipLimit` to `ipRestriction`, also I made it possible to receive `(c: Context) => string)` instead of `getConnInfo`.

@ryuapp
> Sorry for pointing out names again and again, but allowList/denyList expresses the nature better than allow/deny.

By TypeScript definition, users can quickly know that `allow` and `deny` are arrays, so `allow`/`deny` is not bad, right? What do you think?

ryuapp:
@nakasyou 
> By TypeScript definition, users can quickly know that allow and deny are arrays, so allow/deny is not bad, right? What do you think?

You're right, we can notice that they are arrays.
However, since `allow`/`deny` are verbs, I find it odd that they refer to arrays.
So I think `allowList`/`denyList` is better. I'll respect `allow`/`deny` if there's any other benefit.

nakasyou:
@ryuapp I see. I understood your opinion.
I'll change for it.

nakasyou:
I changed

nakasyou:
I implemented dynamic rules, custom error handler, and changed some names.
@ryuapp @EdamAme-x @usualoma @fzn0x and @johnforte, do you have what you want to discuss?

src/middleware/ip-restriction/index.test.ts
```diff
@@ -0,0 +1,86 @@
+import { Hono } from '../../hono'
+import { Context } from '../../context'
+import { HonoRequest } from '../../request'
+import type { GetConnInfo } from '../../helper/conninfo'
+import { ipRestriction, isMatchForRule } from '.'
+
+describe('ipLimit middleware', () => {
```
usualoma, line 7:
```suggestion
describe('ipRestriction middleware', () => {
```

we should also replace s/ipLimit/ipRestriction/

src/middleware/ip-restriction/index.ts
```diff
@@ -0,0 +1,147 @@
...
+  const allowLength = allowList.length
+
+  const blockError = (): HTTPException =>
+    new HTTPException(403, {
+      res: new Response('Unauthorized', {
+        status: 403,
+      }),
+    })
+
+  return createMiddleware(async (c, next) => {
```
usualoma, line 113: It is sometimes helpful to be able to display the middleware name when debugging, so we use named functions for functions included in the hono repository.

```suggestion
  return async function ipRestriction(c, next) {
```

usualoma (commented):
@nakasyou Thank you. I have made some comments, please check them.

src/middleware/ip-restriction/index.test.ts
nakasyou: Thank you! I fixed it.

src/middleware/ip-restriction/index.ts
nakasyou: I agreed! I changed.

ryuapp (commented):
@nakasyou 
Good job. Allow me to leave some comments.

src/middleware/ip-restriction/index.test.ts
usualoma: 👍

src/middleware/ip-restriction/index.ts
usualoma: 👍

usualoma:
I have created a pull request for optimisation, as I believe IP address restrictions are required to be high performance. Please check.

https://github.com/nakasyou/hono/pull/4

src/utils/ipaddr.ts
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+import type { AddressType } from '../helper/conninfo'
+
+/**
+ * Expand IPv6 Address
+ * @param ipV6 Shorten IPv6 Address
+ * @return expanded IPv6 Address
+ */
+export const expandIPv6 = (ipV6: string): string => {
```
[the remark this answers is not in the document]
nakasyou: Thank you for advice! I changed.

src/middleware/ip-restriction/index.test.ts
```diff
@@ -0,0 +1,86 @@
...
+  })
+  it('Static Rules', () => {
+    expect(isMatchForRule({ addr: '192.168.2.1', type: 'IPv4' }, '192.168.2.1')).toBeTruthy()
+    expect(isMatchForRule({ addr: '1234::5678', type: 'IPv6' }, '1234::5678')).toBeTruthy()
+  })
+  it('Function Rules', () => {
+    expect(isMatchForRule({ addr: '0.0.0.0', type: 'IPv4' }, () => true)).toBeTruthy()
+    expect(isMatchForRule({ addr: '0.0.0.0', type: 'IPv4' }, () => false)).toBeFalsy()
+
+    const randIP = Math.random().toString()
```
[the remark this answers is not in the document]
nakasyou: Thank you for advice! I could learn about that from your comment, thank you and I fixed.

src/middleware/ip-restriction/index.ts
```diff
@@ -0,0 +1,146 @@
...
+ *
+ * ### IPv4
+ * - `192.168.2.0` static
+ * - `192.168.2.0/24` CIDR Notation
+ *
+ * ### IPv6
+ * - `::1` static
+ * - `::1/10` CIDR Notation
+ */
+export type IPRestrictRule = string | ((addr: { addr: string; type: AddressType }) => boolean)
```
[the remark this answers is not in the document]
nakasyou: Thank you! I fixed it.

MathurAditya724:
Amazing middleware! A small suggestion I would like to add, can we have a custom error handler function in the options?

nakasyou:
@MathurAditya724, thank you. Yes, the middleware has.

```ts
app.use('*', ipLimit(getConnInfo, {
  deny: [],
  allow: ['127.0.0.1', '::1']
}, c => {
  // error handler
  return c.text('Invalid IP address', 403)
}))
```

src/middleware/ip-restriction/index.ts
```diff
@@ -0,0 +1,175 @@
...
+  onError?: (remote: { addr: string; type: AddressType }) => Response | Promise<Response>
+): MiddlewareHandler => {
+  const allowLength = allowList.length
+
+  const denyMatcher = buildMatcher(denyList)
+  const allowMatcher = buildMatcher(allowList)
+
+  const blockError = (): HTTPException =>
+    new HTTPException(403, {
+      res: new Response('Unauthorized', {
```
yusukebe, line 140: I think `Forbidden` is better than `Unauthorized`.

src/middleware/ip-restriction/index.ts
```diff
@@ -0,0 +1,175 @@
...
+  getIP: GetIPAddr,
+  { denyList = [], allowList = [] }: IPRestrictionRules,
+  onError?: (remote: { addr: string; type: AddressType }) => Response | Promise<Response>
+): MiddlewareHandler => {
+  const allowLength = allowList.length
+
+  const denyMatcher = buildMatcher(denyList)
+  const allowMatcher = buildMatcher(allowList)
+
+  const blockError = (): HTTPException =>
```
yusukebe, line 138: When I tried it on Bun, I faced the error page like the following:

CleanShot 2024-07-06 at 17 10 28@2x

This may be caused by the fact that it does not specify the content type. So, how about using the context?

```ts
const blockError = (c: Context): HTTPException => {
  return new HTTPException(403, {
    res: c.text('Forbidden', 403),
  })
}
```

src/middleware/ip-restriction/index.ts
```diff
@@ -0,0 +1,175 @@
...
+
+/**
+ * IP Limit Middleware
+ *
+ * @param getIP function to get IP Address
+ */
+export const ipRestriction = (
+  getIP: GetIPAddr,
+  { denyList = [], allowList = [] }: IPRestrictionRules,
+  onError?: (remote: { addr: string; type: AddressType }) => Response | Promise<Response>
```
yusukebe, line 131: How about passing the context as the 2nd arg?

```ts
export const ipRestriction = (
  getIP: GetIPAddr,
  { denyList = [], allowList = [] }: IPRestrictionRules,
  onError?: (
    remote: { addr: string; type: AddressType },
    c: Context // <=== 
  ) => Response | Promise<Response>
) => {/** ... */}
```

yusukebe:
Hi @nakasyou !

Sorry for my late reply. This is awesome middleware! I've left some comments. Check them!

src/middleware/ip-restriction/index.ts
nakasyou: I fixed it.

src/middleware/ip-restriction/index.ts
nakasyou: It is good, I implmented.

src/middleware/ip-restriction/index.ts
nakasyou: I changed it.

yusukebe:
@nakasyou 

Great! It looks good to me. The CI is falling, but we can remove the error. Then, I'll merge this into the `next` later.

yusukebe:
The time has come! Merging into the `next`. Thank you all!

PR #2657, opened 2024-05-10
feat(aws-lambda): add alb event processor

yiss:
### The author should do the following, if applicable
This change is made to address the issue raised in #2626. The change adds the following :
- Create a custom `albProcessor` for `ALBProxyEvent`
- Refactor code and introduced new `getHeaders` method
- Added test for certain cases of the ALB event when `multiValueHeaders` is present

- [x] Add tests
- [x] Run tests
- [ ] `bun denoify` to generate files for Deno
- [ ] `bun run format:fix && bun run lint:fix` to format the code

PR #2828, opened 2024-05-28
refactor(aws-lambda): remove unused setHeadersToResult

exoego:
`setHeadersToResult` was added in https://github.com/honojs/hono/pull/2657 but seems not used.

@yiss Please confirm it should be deleted or called.
If it should be called, I'll add tests and fix implementation.

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/src/adapter/aws-lambda/handler.ts
+++ b/src/adapter/aws-lambda/handler.ts
@@ -379,26 +379,7 @@ class ALBProcessor extends EventProcessor<ALBProxyEvent> {
     }
     return headers
   }
-  protected setHeadersToResult(
-    event: ALBProxyEvent,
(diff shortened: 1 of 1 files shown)


yiss:
Hi @exoeg, it should be called. Let me just check quickly if I forgot something in my PR

yiss:
Yeah it seems that I forgot to remove it. LGTM the code of this function is handled by :
```typescript
res.headers.forEach((value, key) => {
      result.headers[key] = value
      if (event.multiValueHeaders && result.multiValueHeaders) {
        result.multiValueHeaders[key] = [value]
      }
    })
```

yiss (approved):

PR #2837, opened 2024-05-28
fix(middleware/body-limit): set default duplex option for readable stream

### LINKED ISSUE #2627: hono.request method does not work in combination with hono/body-limit

4.3.2

Node

```typescript
import {Hono} from 'hono'
import {bodyLimit} from 'hono/body-limit'

const app = new Hono()
app.use(bodyLimit({maxSize: 100 * 1024})) // 100kb

app.post('/hello', async (context) => {
  return context.json({message: 'Hello, World!'})
})

const response = await app.request('/hello', {
  method: 'POST',
  body: JSON.stringify({foo: 'bar'}),
})
console.log('response:', response.status)
```

```
response: 200
```

### What do you see instead?
```
TypeError: RequestInit: duplex option is required when sending a body.
    at new Request (node:internal/deps/undici/undici:5082:19)
    at bodyLimit2 (file:///.../node_modules/hono/dist/middleware/body-limit/index.js:48:17)
    at dispatch (file:///.../node_modules/hono/dist/compose.js:29:23)
    at file:///.../node_modules/hono/dist/compose.js:6:12
    at file:///.../node_modules/hono/dist/hono-base.js:188:31
    at Hono.dispatch (file:///.../node_modules/hono/dist/hono-base.js:198:7)
    at Hono.fetch (file:///.../node_modules/hono/dist/hono-base.js:201:17)
    at Hono.request (file:///.../node_modules/hono/dist/hono-base.js:213:17)
    at <anonymous> (/.../sandbox.ts:12:28)
    at ModuleJob.run (node:internal/modules/esm/module_job:235:25)
response: 500
```

### Additional information

Same app set up is working without any error when started as a node server and fetch response, see example below
```TypeScript
import {Hono} from 'hono'
import {bodyLimit} from 'hono/body-limit'
import {serve} from '@hono/node-server'

const app = new Hono()
app.use(bodyLimit({maxSize: 1024 * 1024})) // 1mb

app.post('/hello', async (context) => {
  return context.json({message: 'Hello, World!'})
})

// const response = await app.request('/hello', {
//   method: 'POST',
//   body: JSON.stringify({foo: 'bar'}),
// })
// console.log('response:', response.status)

serve({
  fetch: app.fetch,
  port: 3000,
})

const response = await fetch('http://localhost:3000/hello', {
  method: 'POST',
  body: JSON.stringify({foo: 'bar'}),
})
console.log('response:', response.status)
```

fzn0x:
Fixes #2627

In this case we do not need globalThis.Request anymore.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/hono-base.ts
+++ b/src/hono-base.ts
@@ -396,13 +396,6 @@ class Hono<E extends Env = Env, S extends Schema = {}, BasePath extends string =
     }
     input = input.toString()
     const path = /^https?:\/\//.test(input) ? input : `http://localhost${mergePath('/', input)}`
-
-    // Always adds default duplex option on Hono.request
-    if (requestInit) {
-      // @ts-expect-error - duplex do not have any type yet
-      requestInit.duplex = 'half'
-    }
-
     const req = new Request(path, requestInit)
     return this.fetch(req, Env, executionCtx)
   }
--- a/src/hono-base.ts
+++ b/src/hono-base.ts
@@ -396,6 +396,13 @@ class Hono<E extends Env = Env, S extends Schema = {}, BasePath extends string =
     }
     input = input.toString()
     const path = /^https?:\/\//.test(input) ? input : `http://localhost${mergePath('/', input)}`
+
+    // Always adds default duplex option on Hono.request
+    if (requestInit) {
+      // @ts-expect-error - duplex do not have any type yet
+      requestInit.duplex = 'half'
+    }
+
     const req = new Request(path, requestInit)
     return this.fetch(req, Env, executionCtx)
   }
--- a/src/middleware/body-limit/index.test.ts
+++ b/src/middleware/body-limit/index.test.ts
@@ -1,17 +1,6 @@
 import { Hono } from '../../hono'
 import { bodyLimit } from '.'
 
-const GlobalRequest = globalThis.Request
-globalThis.Request = class Request extends GlobalRequest {
-  constructor(input: Request | string, init: RequestInit) {
-    if (init) {
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      ;(init as any).duplex ??= 'half'
-    }
-    super(input, init)
-  }
-} as typeof GlobalRequest
-
 const buildRequestInit = (init: RequestInit = {}): RequestInit => {
   const headers: Record<string, string> = {
     'Content-Type': 'text/plain',
@@ -24,6 +13,8 @@ const buildRequestInit = (init: RequestInit = {}): RequestInit => {
     headers,
     body: null,
     ...init,
+    // @ts-expect-error - duplex do not have any type yet
+    duplex: 'half',
   }
 }
 
--- a/src/middleware/body-limit/index.test.ts
+++ b/src/middleware/body-limit/index.test.ts
@@ -82,6 +82,30 @@ describe('Body Limit Middleware', () => {
         expect(await res.text()).toBe('abc')
       })
 
+      it('should return 200 response without user-defined duplex option', async () => {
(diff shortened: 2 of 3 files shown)


fzn0x:
Tested without `buildRequestInit` default duplex is working!

usualoma:
Hi @fzn0x 
Very good! It works fine on all runtimes.

By the way, it seems that ec04a03 alone would solve the problem—do we need a734bf2? I don't see the need to add code to src/hono-base.ts for testing.

fzn0x:
> Hi @fzn0x Very good! It works fine on all runtimes.
> 
> By the way, it seems that [ec04a03](https://github.com/honojs/hono/commit/ec04a03b8997afa051914977b318da81ef6d0f17) alone would solve the problem—do we need [a734bf2](https://github.com/honojs/hono/commit/a734bf233b660339b4398c97ead7e36d96abe743)? I don't see the need to add code to src/hono-base.ts for testing.

Looks like we need to always add duplex by default from now though it's only one available value `half` for now, this approach was used by [`ky`](https://github.com/sindresorhus/ky/blob/2af72bfa7a391662a8ee6b1671979069f7f20737/source/core/Ky.ts#L179) and some other famous Request wrapper libraries.

fzn0x:
-- To avoid user manually adds `duplex` for their ReadableStream requests.

usualoma:
### Avoid @ts-expect-error

Rather than using `@ts-expect-error`, I prefer to avoid errors in the type definition.

```diff
diff --git a/src/middleware/body-limit/index.test.ts b/src/middleware/body-limit/index.test.ts
index 3da88445..60fae554 100644
--- a/src/middleware/body-limit/index.test.ts
+++ b/src/middleware/body-limit/index.test.ts
@@ -1,7 +1,7 @@
 import { Hono } from '../../hono'
 import { bodyLimit } from '.'
 
-const buildRequestInit = (init: RequestInit = {}): RequestInit => {
+const buildRequestInit = (init: RequestInit = {}): RequestInit & { duplex: 'half' } => {
   const headers: Record<string, string> = {
     'Content-Type': 'text/plain',
   }
@@ -13,7 +13,6 @@ const buildRequestInit = (init: RequestInit = {}): RequestInit => {
     headers,
     body: null,
     ...init,
-    // @ts-expect-error - duplex do not have any type yet
     duplex: 'half',
   }
 }
diff --git a/src/middleware/body-limit/index.ts b/src/middleware/body-limit/index.ts
index 282f4d61..7f636830 100644
--- a/src/middleware/body-limit/index.ts
+++ b/src/middleware/body-limit/index.ts
@@ -103,8 +103,8 @@ export const bodyLimit = (options: BodyLimitOptions): MiddlewareHandler => {
       },
     })
 
-    // @ts-expect-error - duplex do not have any type yet
-    c.req.raw = new Request(c.req.raw, { body: reader, duplex: 'half' })
+    const requestInit: RequestInit & { duplex: 'half' } = { body: reader, duplex: 'half' }
+    c.req.raw = new Request(c.req.raw, requestInit as RequestInit)
 
     await next()
 
```

fzn0x:
> For duplex, it is better if the user explicitly specifies 'half' (or in the future 'full').

Since the Request throws `duplex` errors once users sent ReadableStream requests without specifying it, I think this is makes sense.

Revert and types fix is applied!

usualoma:
@fzn0x Thank you, LGTM 👍

yusukebe (approved):
LGTM!

yusukebe:
@fzn0x @usualoma Thank you both! It's an interesting discussion. I fully agree. Looks good merging.

PR #2839, opened 2024-05-28
refactor(websocket): remove unused condition

fzn0x:
Looking at WSEvents type, we don't really need extra AND condition here, just make sure `undefined` is not invoked by specifying `?.()`
```ts
// EVENT is onOpen, onClose, onMessage, onError
(property) WSEvents.EVENT?: ((evt: CloseEvent, ws: WSContext) => void) | undefined
```

Also removes console logs in websocket test.

- [x] Add tests
- [x] Run tests


yusukebe (approved):
LGTM!

yusukebe:
Yes. We can use `?.()`. Thanks.

PR #2845, opened 2024-05-28
fix(jsx/dom): Fixed to not add "px" for certain properties, even if numeric value is given

### LINKED ISSUE #2774: adding 'px' in css variable automatically

4.0.9

Node.js

create a basic layout file and generating tiles and I want to change diffrent background color using hue property as a variable but when I run and see the element panel; it change hue value from number to string and sppend px behind it. this is trange.

here is basic code

### package.json 
```json
{
  "scripts": {
    "dev": "tsx watch src/index.tsx"
  },
  "dependencies": {
    "@hono/node-server": "^1.8.2",
    "@xkeshav/alphabet": "^1.2.1",
    "hono": "^4.0.9"
  },
  "devDependencies": {
    "@types/node": "^20.11.17",
    "tsx": "^3.12.2"
  }
}

```

### index.tsx

```tsx
import { Hono } from 'hono';
import { showRoutes } from 'hono/dev';
import { logger } from 'hono/logger';
import { Alphabet } from './Alphabet';

import { alphabet } from '@xkeshav/alphabet'; 

// above return  ['A', 'B', .... 'Z'] 

const app = new Hono() // .basePath('/api/v1/');

app.use(logger())

app.get('/', (c) => {
  return c.html(
  <div>
    <a href="/alphabet">Alphabet</a>
  </div>
  )
})

app.get('/alphabet', (c) => {
  return c.html(<Alphabet list={alphabet}/>)
});

app.onError((err, c) => {
  console.error(`${err}`)
  return c.text('Custom Error Message', 500)
})


const port = 3000
console.log(`Server is running on port ${port}`);

showRoutes(app);

serve({
  fetch: app.fetch,
  port
})
```

### Layout.tsx

```tsx
import { Style, css } from 'hono/css';
import { FC } from 'hono/jsx';

export const Layout: FC = (props) =>  (
    <html>
      <head>
     <Style>{
				css`
        @property --hue {
          syntax: "<number>";
          inherits: true;
          initial-value: 10;
        }
          html {
            font-family: Arial, Helvetica, sans-serif;
          }
					:root {
						--hue: 10;
					}
        `}
				</Style>
				</head>
      <body>{props.children}</body>
    </html>
  );

```

### Alphabet.tsx

```tsx
import { css } from 'hono/css';
import { FC } from 'hono/jsx';
import { Layout } from './Layout';

const parentContainer = css`
  display:grid;
  grid-gap: 4px;
  grid-template-columns: repeat(5,1fr);
`
const itemClass = css`
  display: grid;
  height: 10rem;
  color: var(--text, white);
  background-color: hwb(var(--hue) 0% 0%);
  place-content: center;
  font-size: 4rem;
}
`
const Alphabet: FC<{ list: string[] }> = ({ list }) => (
  <Layout>
    <h1>Alphabets</h1>
    <div class={parentContainer}>
      {list.map(((item, id) => <div class={itemClass} style={{ "--hue": id*10 + 20, "--text": id < 20 ? '#000' : '#FFF' }}>{item}</div>))}
    </div>
  </Layout>
);


export { Alphabet };
```

Screenshot from 2024-05-23 18-56-02

checking in Ubuntu 24.04 with firefox v 126.0  and chrome v 124.x browser

every tile must have diffrent color as per changed hue and it was working earlier (checked in non ubuntu system)  but suddenly stopped working 

### What do you see instead?

all Tile have same color because the varibel --hue assigned a string value 

### Additional information

 in chrome text color changed as per logic  but background remain same while in firefox all tile have same color and background-color

usualoma:
Fix #2774

Change in #2553 to add "px" when a value of type number is passed. However, it was found that some properties need to output number as is, as defined in React as follows.

https://github.com/facebook/react/blob/46339720d75337ae1d1e113fd56ac99e7fd1a0b3/packages/react-dom-bindings/src/shared/isUnitlessNumber.js#L14-L84

Keeping the entire list would increase the code, so this PR implementation was made smaller by excluding 'deprecated' and 'vendor-prefixed' items from the React list, and then grouping them into regular expressions.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/jsx/utils.test.ts
+++ b/src/jsx/utils.test.ts
@@ -1,0 +1,91 @@
+import { styleObjectForEach } from './utils'
+
+describe('styleObjectForEach', () => {
+  describe('Should output the number as it is, when a number type is passed', () => {
+    test.each`
+      property
+      ${'animationIterationCount'}
+      ${'aspectRatio'}
+      ${'borderImageOutset'}
+      ${'borderImageSlice'}
+      ${'borderImageWidth'}
+      ${'columnCount'}
+      ${'columns'}
+      ${'flex'}
+      ${'flexGrow'}
+      ${'flexPositive'}
+      ${'flexShrink'}
+      ${'flexNegative'}
+      ${'flexOrder'}
+      ${'gridArea'}
+      ${'gridRow'}
+      ${'gridRowEnd'}
+      ${'gridRowSpan'}
+      ${'gridRowStart'}
+      ${'gridColumn'}
+      ${'gridColumnEnd'}
+      ${'gridColumnSpan'}
+      ${'gridColumnStart'}
+      ${'fontWeight'}
+      ${'lineClamp'}
+      ${'lineHeight'}
+      ${'opacity'}
+      ${'order'}
+      ${'orphans'}
+      ${'scale'}
+      ${'tabSize'}
+      ${'widows'}
+      ${'zIndex'}
+      ${'zoom'}
+      ${'fillOpacity'}
+      ${'floodOpacity'}
+      ${'stopOpacity'}
+      ${'strokeDasharray'}
+      ${'strokeDashoffset'}
+      ${'strokeMiterlimit'}
+      ${'strokeOpacity'}
+      ${'strokeWidth'}
+    `('$property', ({ property }) => {
+      const fn = vi.fn()
+      styleObjectForEach({ [property]: 1 }, fn)
+      expect(fn).toBeCalledWith(
+        property.replace(/[A-Z]/g, (m: string) => `-${m.toLowerCase()}`),
+        '1'
+      )
+    })
+  })
+  describe('Should output with px suffix, when a number type is passed', () => {
+    test.each`
+      property
+      ${'borderBottomWidth'}
+      ${'borderLeftWidth'}
+      ${'borderRightWidth'}
+      ${'borderTopWidth'}
+      ${'borderWidth'}
+      ${'bottom'}
+      ${'fontSize'}
+      ${'height'}
+      ${'left'}
+      ${'margin'}
+      ${'marginBottom'}
+      ${'marginLeft'}
+      ${'marginRight'}
+      ${'marginTop'}
+      ${'padding'}
+      ${'paddingBottom'}
+      ${'paddingLeft'}
+      ${'paddingRight'}
+      ${'paddingTop'}
+      ${'right'}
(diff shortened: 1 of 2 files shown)


usualoma:

yusukebe:
Hi @usualoma !

I asked my ChatGPT, and it answered there are more CSS properties, and we can add the test properties:

```diff
diff --git a/src/jsx/utils.test.ts b/src/jsx/utils.test.ts
index 0f6fd7b..3fa1231 100644
--- a/src/jsx/utils.test.ts
+++ b/src/jsx/utils.test.ts
@@ -45,6 +45,12 @@ describe('styleObjectForEach', () => {
       ${'strokeMiterlimit'}
       ${'strokeOpacity'}
       ${'strokeWidth'}
+      ${'animationDuration'}
+      ${'animationDelay'}
+      ${'animationTimingFunction'}
+      ${'transitionDuration'}
+      ${'transitionDelay'}
+      ${'transitionTimingFunction'}
     `('$property', ({ property }) => {
       const fn = vi.fn()
       styleObjectForEach({ [property]: 1 }, fn)
@@ -79,6 +85,18 @@ describe('styleObjectForEach', () => {
       ${'right'}
       ${'top'}
       ${'width'}
+      ${'letterSpacing'}
+      ${'wordSpacing'}
+      ${'textIndent'}
+      ${'maxHeight'}
+      ${'maxWidth'}
+      ${'minHeight'}
+      ${'minWidth'}
+      ${'outlineWidth'}
+      ${'perspective'}
+      ${'perspectiveOrigin'}
+      ${'columnGap'}
+      ${'rowGap'}
     `('$property', ({ property }) => {
       const fn = vi.fn()
       styleObjectForEach({ [property]: 1 }, fn)
```

It will fail, so you can add more regexp rules. Can you check them?

usualoma:
@yusukebe Thank you for reviewing.

* animationDuration
* animationDelay
* animationTimingFunction
* transitionDuration
* transitionDelay
* transitionTimingFunction

These properties are indeed incorrect for “px" suffix, while number is also incorrect. It is used as `${number}s`, for example.
Even in React, I don't think any user would specify number for these properties as it is not a valid value.

Alternatively, Hono's unique convenience (add “s” suffix?) could be considered.

For the time being, I think the current regular expression is giving the result we are aiming for.

yusukebe:
@usualoma 

Thank you for your explanation. `${number}s` - Undestood well. I'll merge this now!

PR #2848, opened 2024-05-29
docs: change to shell highlight

EdamAme-x:
hightlight for shell code

```shell
cd /hono
```

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -24,13 +24,13 @@ The `honojs/hono` project uses [Bun](https://bun.sh/) as its package manager. De
 
 After that, please install the dependency environment.
 
-```txt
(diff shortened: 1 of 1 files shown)


yusukebe:
What about using `console`.

EdamAme-x:
```shell
cd /a
bun install
```

```console
cd /a
bun install
```

nice suggestion!

EdamAme-x:
Astro is using ````bash` (as shell).
what do you think? @yusukebe

yusukebe:
@EdamAme-x 

Ahh, `console` is not good. Then, I also `bash` is better.

EdamAme-x:
ok. I ready to be

yusukebe (approved):
LGTM!

PR #2850, opened 2024-05-29
refactor: added paths-ignore for ignore files of dont need tests

EdamAme-x:
docs, .vscode, ... don't need tests


yusukebe:
Thanks!

PR #2852, opened 2024-05-29
feat(hono-base): add `replaceRequest` option for `app.mount`

### LINKED ISSUE #2781: app.mount() removing the subpath from the request url is causing routing error in the inner routers

4.3.11

Cloudflare workers

`app.mount()` removes the subpath, and this is causing some issues in the inner mounted router.
I think the current behavior works for almost all applications, and in my opinion it should continue to be the default behavior, but in some particular routers, like `itty-router-openapi` that auto generates some routes for users, is causing some routes to point to non-existing routes.

This issue was raised in itty-router-openapi [repo here](https://github.com/cloudflare/itty-router-openapi/issues/150) and the linked issue provides some more information on this.

This example uses itty-router-openapi, but it would work the same way in other routers
```ts
import { OpenAPIRouter } from '@cloudflare/itty-router-openapi'
import { Hono } from 'hono'

const router = OpenAPIRouter()
router.get('/example', (request) => new Response(request.url))

const app = new Hono()
app.mount('/api', router.handle)

export default app
```

Before this issue was raised in itty-router-openapi repository, i wasn't aware of this `request.url` change from Hono, so my expected behavior before knowing this, would be that a user would just define the `base` parameter in itty-router-openapi and everything would work correctly, like this example

```ts
import { OpenAPIRouter } from '@cloudflare/itty-router-openapi'
import { Hono } from 'hono'

const router = OpenAPIRouter({ base: '/api' })
router.get('/example', (request) => new Response(request.url))

const app = new Hono()
app.mount('/api', router.handle)

export default app
```

In this example, calling `http://localhost:8787/api/example` would print `http://localhost:8787/api/example`.

### What do you see instead?

Running the script in the "What steps can reproduce the bug", then calling `http://localhost:8787/api/example` prints `http://localhost:8787/example`.

An example of the auto generated routes not working is to open this url `http://localhost:8787/api/docs` in the browser, that is trowing an error, because the lib expects the openapi spec to be an `/openapi.json` but it actually is in `/api/openapi.json` because of the url re-write.

### Additional information

I think that this should continue to be the default way of mounting other routers in Hono, as probably some routers don't support defining base paths and makes the developer experience better for users.

My initial suggestion was to add a parameter when calling the inner router in [this line](https://github.com/honojs/hono/blob/main/src/hono-base.ts#L275), and this would make "it work" for users, and users wouldn't need to change anything, then on itty-router-openapi side I could expect this new parameter and work around it.
But after talking to @yusukebe he suggested creating an additional option to disable this behavior, maybe something like this `app.mount('/api', router.handle, { subpathRewrite: false })`

yusukebe:
Resolves #2781

This PR introduces the `rewritePath` option to `app.mount()`. With the current spec, if you use `app.mount()`, the path to pass to your mounted application will be changed. The base path will be removed.

```ts
app.mount('/sub-app', subApp.handle)
```

If you access `/sub-app/hello/foo`, the subApp receives `/hello/foo`. This is good behavior for some apps, but it's bad for some apps like #2781.

With the `rewritePath` option, you can specify the path to pass to the sub-application. If you don't want to remove the base path, you can write the following:

```ts
app.mount('/sub-app', subApp.handle, {
  rewritePath: (path) => path,
})
```

### The pain point

The logic will be a little bit complex. As for this, the application size will be slightly increased. Eitherway, I think we can reduce the size in other places.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/hono-base.ts
+++ b/src/hono-base.ts
@@ -305,7 +305,7 @@ class Hono<E extends Env = Env, S extends Schema = {}, BasePath extends string =
    *
    * @param {string} path - base Path
    * @param {Function} applicationHandler - other Request Handler
-   * @param {Function | undefined} optionHandler - other Request Handler with Hono Context
+   * @param {MountOptions} [options] - options of `.mount()`
    * @returns {Hono} mounted Hono instance
    *
    * @example
@@ -319,6 +319,15 @@ class Hono<E extends Env = Env, S extends Schema = {}, BasePath extends string =
    *
    * const app = new Hono()
    * app.mount('/itty-router', ittyRouter.handle)
+   * ```
+   *
+   * @example
+   * ```ts
+   * const app = new Hono()
+   * // Send the request to another application without modification.
+   * app.mount('/app', anotherApp, {
+   *   replaceRequest: (req) => req,
+   * })
    * ```
    */
   mount(
--- a/src/hono-base.ts
+++ b/src/hono-base.ts
@@ -81,6 +81,15 @@ export type HonoOptions<E extends Env> = {
   getPath?: GetPath<E>
 }
 
+type MountOptionHandler = (c: Context) => unknown
+type MountRewritePath = (path: string) => string
+type MountOptions =
+  | MountOptionHandler
+  | {
+      optionHandler?: MountOptionHandler
+      rewritePath?: MountRewritePath
+    }
+
 class Hono<E extends Env = Env, S extends Schema = {}, BasePath extends string = '/'> {
   // Theses methods are dynamically initialized in the constructor.
   get!: HandlerInterface<E, 'get', S, BasePath>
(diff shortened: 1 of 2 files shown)


yusukebe:
Hey @usualoma ! What do you think about this feature?

yusukebe:
Or, I've created [the Mounting Helper](https://github.com/honojs/hono/pull/2853):

```ts
const app = new Hono()

app.all(
  '/sub-app/*',
  toHandler(subApp.fetch, {
    basePath: '/sub-app',
  })
)
```

This is interesting since the sub-app will act as a handler. It can be separated from `hono-base.ts`, and we can obsolete `app.mount`.

But I've noticed this is redundant.

yusukebe:
Or simply, it is okay to be `boolean` since I can't imagine the use case of the function if `rewritePath` is a function.

```ts
app.mount('/sub-app', subApp.handle, {
  rewritePath: false,
})
```

yusukebe:
@usualoma Sorry for pinging again!

In summary, there are three ways to resolve #2781.

1. `rewritePath` option - `(path: string) => string`
2. `rewritePath` option - `boolean`
3. [Mounting Helper](https://github.com/honojs/hono/pull/2853)

What do you think about it?

usualoma:
@yusukebe Sorry, I missed it. I will check now.

usualoma:
@yusukebe 
This PR approach is good! I think the following option is better.

> `rewritePath` option - `(path: string) => string`

Here, when `(path) => path`, the `Request` replacement is unnecessary, and the original `c.req.raw` can be passed directly to the sub-application. Given that, it seems that we should be providing not a handler for `path replacement`, but a handler for `request replacement`. That way, you could do things like (I wonder how much demand there would be) 'add headers in requests to sub-application'.

The following.
https://github.com/honojs/hono/pull/2873

mgrithm:
hey @usualoma, I sent an email for you, so could you see?

yusukebe:
@usualoma 

Thank you for the suggestion and the PR! `replaceRequest` is good! I've merged the PR to this branch. It's ready to integrate into the main. Maybe I'll release it as a patch release since it's only an update of the options.

usualoma:
@yusukebe 
Thank you!
I think it's good.
> Maybe I'll release it as a patch release since it's only an update of the options.

PR #2864, opened 2024-05-30
fix(websocket): the onopen event cannot be triggered during delayed operations in deno

JetLua:
```ts
const app = new Hono()

app.get(
  '/ws',
  upgradeWebSocket(async (c) => {
    await new Promise(resolve => setTimeout(resolve, 5e2)) // Simulate time-consuming operation
    return {
      onOpen: () => console.log('opened') // can't be trigger
    }
  })
)
```
Because `socket.onopen` was triggered before `createEvents`.
Swapping L7 and L9 can solve this problem：

https://github.com/honojs/hono/blob/3d6820b46e51622435aeb0cacc6434d722581aa2/src/adapter/deno/websocket.ts#L7-L9

### DIFF
--- a/runtime_tests/deno/hono.test.ts
+++ b/runtime_tests/deno/hono.test.ts
@@ -34,22 +34,3 @@ Deno.test('environment variables', () => {
   const { NAME } = env<{ NAME: string }>(c)
   assertEquals(NAME, 'Deno')
 })
-
-Deno.test('trigger the onOpen event of upgradeWebSocket', async () => {
-  const app = new Hono()
-  const open = new Promise<boolean>(resolve => {
-    app.get('/ws', upgradeWebSocket(async (c) => {
-      await new Promise(resolve => setTimeout(resolve, 5e2))
-      return {
-        onOpen: () => resolve(true)
-      }
-    }))
-    const server = Deno.serve(app.fetch)
-    const socket = new WebSocket('http://127.1:8000/ws')
-    socket.onopen = () => {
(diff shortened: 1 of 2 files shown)


yusukebe:
@JetLua 

Thank you for the PR!

@nakasyou Could you review this?

yusukebe:
@JetLua @nakasyou This is looks good. But, we'd like to have a test, if it is possible. Do you have any ideas?

JetLua:
@yusukebe Previously, I added the following test code, but it requires `--allow-net` permission.

```ts
Deno.test('trigger the onOpen event of upgradeWebSocket', async () => {
  const app = new Hono()
  const open = new Promise<boolean>(resolve => {
    app.get('/ws', upgradeWebSocket(async (c) => {
      await new Promise(resolve => setTimeout(resolve, 5e2))
      return {
        onOpen: () => resolve(true)
      }
    }))
    const server = Deno.serve(app.fetch)
    const socket = new WebSocket('http://127.1:8000/ws')
    socket.onopen = () => {
      server.shutdown()
      socket.close()
    }
  })
  assertEquals(await open, true)
})
```

yusukebe:
@JetLua 

Thank you for sharing! Hmm, it's a good test, but we don't want to run the actual server in the test phase.

yusukebe:
@JetLua 

Okay! We might add tests, but I'll merge this now. Thanks!

cc: @nakasyou If you have any good ideas in the future, please share them.

PR #2867, opened 2024-05-30
feat(stream): Options compress/decompress

watany-dev:
https://developer.mozilla.org/ja/docs/Web/API/Compression_Streams_API

- [ ] Add tests
- [ ] Run tests


watany-dev:
Sorry, this is `app.use(compress({encoding: 'gzip'}))`
The behavior is the same as when adding , so it seems there is no need to implement a new one.

PR #2873, opened 2024-05-31
feat(hono-base): introduce `replaceRequest` option instead of `rewritePath`

usualoma:


src/hono-base.ts
```diff
@@ -258,43 +258,43 @@ class Hono<E extends Env = Env, S extends Schema = {}, BasePath extends string =
...
+      if (typeof options === 'function') {
+        optionHandler = options
+      } else {
+        optionHandler = options.optionHandler
+        replaceRequest = options.replaceRequest
       }
+    }
 
-      const applicationOptions = optionHandler ? [optionHandler(c)] : [c.env, executionContext]
-      const queryStrings = getQueryStrings(c.req.url)
```
yusukebe, line 280: Indeed, we don't need `getQueryStrings`!

PR #2875, opened 2024-05-31
ci: change name of workflow jobs

EdamAme-x:
I think this is more appropriate than before.

- [ ] Add tests
- [ ] Run tests


yusukebe:
I think `jsr` is better.

EdamAme-x:
thank for reply!
I'll change it!

yusukebe (approved):
LGTM!

PR #2851, opened 2024-05-29
ci: include node22 tests

Jayllyz:
Node.js 22 added to the test matrix.

ref : https://github.com/honojs/node-server/pull/170

### The author should do the following, if applicable

- [ ] Add tests
- [ ] Run tests
- [ ] `bun run format:fix && bun run lint:fix` to format the code
- [ ] Add [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code

PR #2876, opened 2024-05-31
ci: set ".github/**" at paths-ignore

EdamAme-x:
I don't think “.github” needs "ci" of actions either.
https://github.com/honojs/hono/pull/2850

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,7 @@ on:
     paths-ignore:
       - 'docs/**'
       - '.vscode/**'
+      - '.github/**'
       - 'README.md'
       - '.gitignore'
       - 'LICENSE'


exoego:
I am against this change. 

`.github/workflows` contains CI job definitions.
CI should run when CI job definitions get updated (e.g. #2851).

EdamAme-x:
> I am against this change.
> 
> `.github/workflows` contains CI job definitions. CI should run when CI job definitions get updated (e.g. #2851).

Thank for your reply.
I apologize for my lack of knowledge. 
I think the CI after the changes are made is executed after the commit (merged), and there is no problem if it is not executed at the time of the PR, isn't it?

exoego:
>  CI after the changes are made is executed after the commit (merged), and there is no problem if it is not executed at the time of the PR, isn't it?

No, it is a problem.
Consider a PR that adds a new version of runtime (e.g. Node.js 24 or Bun2) that introduces breaking changes to Hono.
If CI is skipped, we can not be aware of such breaking changes, and the PR might be merged.
CI on the PRs created after the PR will fail due to the missed breaking changes.
I think it is terrible for contributors/maintainers, because breaking changes sometime requires time to fix.
We can revert the change easily, but I don't think it can be an excuse.

> We can also ignore other than .github/workflows. What do you think?

I personally think such "smart" configuration should be avoided.
`.github` can contain many types of files and GitHub may add more that might be related to CI.
Maintainers/contributors might get surprised when they update something in `.github` but CI is skipped.

----

I personally think "safety" (error prevention) should be prioritized.
So  `paths-ignore` should be used very carefully when:

- CI is very very slow
- CI consumes too much money

This appears not to apply to the Hono repo right now: the slowest CI job succeeds in 1m, which is acceptable.
image

EdamAme-x:
@exoego thank for your reply!
I understood. I'll close this

PR #2880, opened 2024-05-31
chore: reestructure `message` of tests

mgrithm:
Improved the assertion messages in the tests for better clarity and understanding

image

### DIFF
--- a/src/middleware/basic-auth/index.test.ts
+++ b/src/middleware/basic-auth/index.test.ts
@@ -28,7 +28,7 @@ describe('Basic Auth by Middleware', () => {
       password,
     })
   )
-  
+
   app.use('/auth/*', async (c, next) => {
     c.header('x-custom', 'foo')
     await next()
--- a/src/middleware/basic-auth/index.test.ts
+++ b/src/middleware/basic-auth/index.test.ts
@@ -28,7 +28,7 @@ describe('Basic Auth by Middleware', () => {
       password,
     })
   )
-  // Test multiple handlers
+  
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @mgrithm 

Unifying the descriptions is good. But some are considering things, so I don't merge them immediately.

* If we accept this PR, we have to rewrite all the tests.
* There will be a lot of diffs. Then, you or someone who is only doing a rewrite of the existing things has many commits. This is also the case with factoring, but it is also concerning that these people are valued more than others as contributors.
* We don't have the guidelines for the test message currently.
* This PR removes the tests. Not only fixing the descriptions.

mgrithm:
Hi @yusukebe 

Some tests don't seem to make sense because the main one is already tested. Using `should` in tests is standard and we could start using it

mgrithm:
If you want, I can work to update the name of all tests

yusukebe:
> Some tests don't seem to make sense because the main one has already been tested. 

Indeed, now we can measure the coverage so we can know the tests we can remove.

> Using `should` in tests is standard and we could start using it

I know it, but I can't say "please do it" right now for some of the above reasons. Please wait.

mgrithm:
All right! Thanks for the answer. I will work elsewhere in Hono

mgrithm:
All right! Thanks for the answer. I will work elsewhere in Hono

yusukebe:
@mgrithm 

Your contribution to Hono is very much appreciated. And we are getting many contributors. So, I'll create "guidelines," like updating the contribution guide. Thanks!

PR #2885, opened 2024-06-02
perf(hono-base): don't import `HTTPException`

yusukebe:
This is like a `refactor` but is made a `perf` since reducing the bundle size for the performance.

With this PR, the `hono-base.ts` does not need to import the `HTTPException` class directly. So, the bundle size for the app that does not use `HTTPException` will be small. To enable it, it does not use `instanceOf`; it will detect by checking the property named a symbol of `IS_HTTP_EXCEPTION.`

Result:

CleanShot 2024-06-02 at 18 40 35@2x

It was reduced by 260 bytes. It seems small, but reducing size is important for edge environments such as Cloudflare Workers.

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/src/hono-base.ts
+++ b/src/hono-base.ts
@@ -7,7 +7,6 @@
 import { compose } from './compose'
 import { Context } from './context'
 import type { ExecutionContext } from './context'
-import { HTTPException } from './http-exception'
 import { HonoRequest } from './request'
 import type { Router } from './router'
 import { METHODS, METHOD_NAME_ALL, METHOD_NAME_ALL_LOWERCASE } from './router'
@@ -34,12 +33,18 @@ import { getPath, getPathNoStrict, getQueryStrings, mergePath } from './utils/ur
  */
 export const COMPOSED_HANDLER = Symbol('composedHandler')
 
+/**
+ * Symbol for the property name to mark whether it is HTTPException.
+ */
+export const IS_HTTP_EXCEPTION = Symbol('isHTTPException')
+
 const notFoundHandler = (c: Context) => {
   return c.text('404 Not Found', 404)
 }
 
 const errorHandler = (err: Error, c: Context) => {
-  if (err instanceof HTTPException) {
+  if ((err as any)[IS_HTTP_EXCEPTION]) {
+    //@ts-expect-error `getResponse()` is not typed.
     return err.getResponse()
   }
   console.error(err)
--- a/src/http-exception.ts
+++ b/src/http-exception.ts
@@ -3,6 +3,7 @@
  * This module provides the `HTTPException` class.
  */
 
+import { IS_HTTP_EXCEPTION } from './hono-base'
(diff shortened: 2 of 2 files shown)


yusukebe:
Hey @usualoma. I want to know your thoughts on this.

usualoma:
@yusukebe 
Ah I see, this is very nice!

Alternatively, how about the following. It simplifies the definition and opens up the possibility for users themselves to use "errors implementing getResponse()", other than hono’s `HTTPException`. The behaviour would change if an existing application used "errors implementing getResponse()", but, well, there would be no such application.

```diff
diff --git a/src/hono-base.ts b/src/hono-base.ts
index f3e9ad03..858d65db 100644
--- a/src/hono-base.ts
+++ b/src/hono-base.ts
@@ -7,7 +7,6 @@
 import { compose } from './compose'
 import { Context } from './context'
 import type { ExecutionContext } from './context'
-import { HTTPException } from './http-exception'
 import { HonoRequest } from './request'
 import type { Router } from './router'
 import { METHODS, METHOD_NAME_ALL, METHOD_NAME_ALL_LOWERCASE } from './router'
@@ -21,6 +20,7 @@ import type {
   MergeSchemaPath,
   MiddlewareHandler,
   MiddlewareHandlerInterface,
+  HTTPResponseError,
   Next,
   NotFoundHandler,
   OnHandlerInterface,
@@ -38,8 +38,8 @@ const notFoundHandler = (c: Context) => {
   return c.text('404 Not Found', 404)
 }
 
-const errorHandler = (err: Error, c: Context) => {
-  if (err instanceof HTTPException) {
+const errorHandler = (err: Error | HTTPResponseError, c: Context) => {
+  if ('getResponse' in err) {
     return err.getResponse()
   }
   console.error(err)
diff --git a/src/types.ts b/src/types.ts
index 2178e5c1..87691b01 100644
--- a/src/types.ts
+++ b/src/types.ts
@@ -91,8 +91,12 @@ export type H<
 > = Handler<E, P, I, R> | MiddlewareHandler<E, P, I>
 
 export type NotFoundHandler<E extends Env = any> = (c: Context<E>) => Response | Promise<Response>
+
+export interface HTTPResponseError extends Error {
+  getResponse: () => Response
+}
 export type ErrorHandler<E extends Env = any> = (
-  err: Error,
+  err: Error | HTTPResponseError,
   c: Context<E>
 ) => Response | Promise<Response>
 ```

yusukebe:
@usualoma Thank you for the suggestion!

I also thought it was okay to check `getResponse` only first. As you said, we can ignore other applications that have "errors implementing getResponse()" since they are very rare or do not exist. So, I'll follow your idea.

yusukebe:
@usualoma 

Can you create the new PR with your patch?

yusukebe:
This is not necessary now since #2898 has been merged.

Issue #2897, opened 2024-06-03
Allow throwing responses (like Remix) and automatically wrap them in HttpException


ajaishankar:
### What is the feature you are proposing?

Having a conditional response like the following breaks type inference as a generic Response is inferred

```ts
app.get("/profile", (c) => {
  return c.var.user ? c.json(c.var.user) : c.redirect("/login");
});
```

Feel the documentation could be enhanced to avoid conditional responses and use HttpException instead

```ts
app.get("/profile", (c) => {
  if (!c.var.user) {
    throw new HTTPException(undefined, {
      res: c.redirect("/login"),
    });
  }
  return c.json(c.var.user);
});
```

Since we are throwing an exception for the redirect, the return type is inferred correctly for the happy path.

Even better maybe hono could support throwing responses similar to Remix?

https://remix.run/docs/en/main/utils/redirect
https://remix.run/docs/en/main/guides/not-found#how-to-send-a-404

```ts
app.get("/profile", (c) => {
  if (!c.var.user) {
    throw c.redirect("/login")
  }
  return c.json(c.var.user);
});
```

This can be achieved with a simple middleware, but feel this _might_ be a good enhancement?

mgrithm:
hey @yusukebe, could you take a look at that?

yusukebe:
Hi @ajaishankar 

One of the reasons why it occurs is the `c.redirect()` does not support the RPC. We can make it support it. How about like this?

```ts
const app = new Hono()

const flag = true

const routes = app.get('/profile', (c) => {
  if (!flag) {
    return c.redirect('/', 301)
  }
  return c.json({ name: 'abc' }, 200)
})

const client = hc<typeof routes>('/')

const res = await client.profile.$get()

if (res.status === 301) {
  console.log(res.headers.get('Location'))
} else {
  const data = await res.json()
  //    ^ { name: string }
}
```

molives:
Will Hono implement this soon?

yusukebe:
@molives yes.

@ajaishankar Can you check #2908?

ajaishankar:
@yusukebe 

Commented on the PR, looks good...

Redirect is a common enough use case and nice to have support for this in hono RPC.

For something like Remix thrown response I'd suggested very easy to accomplish with a simple middleware. Again Hono makes everything super simple! 🙂

```ts
const thrownResponseToHttpException = createMiddleware(async (c, next) => {
  try {
    await next();
  } catch (e) {
[...]

PR #2908, opened 2024-06-05
feat(context): `c.redirect()` supports `TypedResponse`

yusukebe:
Related to #2897

This PR enables `c.redirect()` to return `TypedResponse`. So, you can handle redirect responses in the RPC mode.

```ts
const routes = app.get('/profile', (c) => {
  if (!flag) {
    return c.redirect('/')
  }
  return c.json({ name: 'abc' }, 200)
})

const client = hc<typeof routes>('/')

const res = await client.profile.$get()

if (res.status === 302) {
  console.log(res.headers.get('Location'))
} else {
  const data = await res.json()
  //    ^ { name: string }
}
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -10,7 +10,7 @@ import { parse } from '../utils/cookie'
 import type { Equal, Expect } from '../utils/types'
 import { validator } from '../validator'
 import { hc } from './client'
-import type { InferRequestType, InferResponseType } from './types'
+import type { ClientResponse, InferRequestType, InferResponseType } from './types'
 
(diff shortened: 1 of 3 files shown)


yusukebe:
@NamesMT What do you think of this PR?

ajaishankar:
Thanks @yusukebe 

Looks good 👍  feel redirect is a common enough use case to support in hono RPC...

yusukebe:
@ajaishankar Thanks!

Let's merge it and release it as **patch** since it is a `feat` but a tiny change.

PR #2910, opened 2024-06-05
fix(filepath): allow suffix includes `-` and `_`

yusukebe:
Fixes https://github.com/honojs/hono/issues/2251#issuecomment-2150930169

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/utils/filepath.test.ts
+++ b/src/utils/filepath.test.ts
@@ -56,6 +56,9 @@ describe('getFilePath', () => {
     expect(getFilePath({ filename: 'foo', root: 'bar', defaultDocument: 'index.txt' })).toBe(
       'bar/foo/index.txt'
     )
+
+    expect(getFilePath({ filename: 'filename.suffix_index' })).toBe('filename.suffix_index')
+    expect(getFilePath({ filename: 'filename.suffix-index' })).toBe('filename.suffix-index')
   })
 })
 
--- a/src/utils/filepath.ts
+++ b/src/utils/filepath.ts
@@ -16,7 +16,7 @@ export const getFilePath = (options: FilePathOptions): string | undefined => {
   if (filename.endsWith('/')) {
     // /top/ => /top/index.html
     filename = filename.concat(defaultDocument)
-  } else if (!filename.match(/\.[a-zA-Z0-9]+$/)) {
+  } else if (!filename.match(/\.[a-zA-Z0-9_-]+$/)) {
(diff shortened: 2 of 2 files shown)


yusukebe:

usualoma:
Hi @yusukebe 

As I understand it, the problem is that `console.warn()` is called when the file does not exist, in Deno.

It is not called `console.warn()` in bun.
https://github.com/usualoma/hono/blob/121746b8727d740cf6e36433ae55721f74b78353/src/adapter/bun/serve-static.ts#L13-L14

node-server checks for the existence of the file first.
https://github.com/usualoma/node-server/blob/a0be98b95ed5221d96c988565e5f5e6fa74228ee/src/serve-static.ts#L56-L59

Therefore I think we should be to silently return null if the file does not exist, even in deno.

```diff
diff --git a/src/adapter/deno/serve-static.ts b/src/adapter/deno/serve-static.ts
index b037466b..9fb577bd 100644
--- a/src/adapter/deno/serve-static.ts
+++ b/src/adapter/deno/serve-static.ts
@@ -16,6 +16,11 @@ export const serveStatic = <E extends Env = Env>(
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
         return file ? (file.readable as any) : null
       } catch (e) {
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
+        if (e instanceof Deno.errors.NotFound) {
+          return null // do not report error
+        }
         console.warn(`${e}`)
       }
     }
```


I don't think we need to change the extension definition now.

yusukebe:

> As I understand it, the problem is that `console.warn()` is called when the file does not exist, in Deno.

Indeed, Bun's and Deno's implementations differ. Bun's serve static does not show the error.

But I intended it to show the not found error, making the user quickly debug, though Bun's is not showing it now.

In this case, the problem is if accessing the URLs will be access to the following file paths:

* `/static/foo.bar_index` => `./static/foo.bar_index/index.html` // not found
* `/static/foo.bar-index` => `./static/foo.bar-index/index.html` // not found

Ideally, if access to `/static/foo.bar_index`, `./static/foo.bar_index` will be looked up. So, the change in this PR is preferred.

usualoma:
@yusukebe 
I understood the situation from your explanation. Sorry for making you spend so much time.
I think it's good!

yusukebe:
@usualoma Thanks!

PR #2911, opened 2024-06-05
fix: Read all chunks when hashing a readable stream

jpdutoit:
createHash was only reading the first chunk when hashing a readable stream. This breaks etags for longer responses.

On bun the first chunk seems to always be 16384 bytes long, so I added a test there. Could not reproduce on node

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/utils/crypto.ts
+++ b/src/utils/crypto.ts
@@ -34,11 +34,19 @@ export const createHash = async (data: Data, algorithm: Algorithm): Promise<stri
   if (data instanceof ReadableStream) {
     let body = ''
     const reader = data.getReader()
-    await reader?.read().then(async (chuck) => {
-      const value = await createHash(chuck.value || '', algorithm)
+    let chunks = 0
+    let chunk = await reader?.read()
+    while (chunk && !chunk.done) {
+      chunks++
+      const value = await createHash(chunk.value || '', algorithm)
       body += value
-    })
-    return body
+
+      chunk = await reader?.read()
+    }
+    if (chunks === 1) {
+      return body
+    }
+    data = body
   }
   if (ArrayBuffer.isView(data) || data instanceof ArrayBuffer) {
     sourceBuffer = data
--- a/runtime_tests/bun/index.test.tsx
+++ b/runtime_tests/bun/index.test.tsx
@@ -9,6 +9,7 @@ import { Hono } from '../../src/index'
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 import { jsx } from '../../src/jsx'
 import { basicAuth } from '../../src/middleware/basic-auth'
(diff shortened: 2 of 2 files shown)


yusukebe:
Hi @jpdutoit 

Thank you for the PR! Can you write a unit test in ``src/utils/crypto.test.ts`?

src/utils/crypto.ts
```diff
@@ -34,11 +34,19 @@ export const createHash = async (data: Data, algorithm: Algorithm): Promise<stri
...
+    while (chunk && !chunk.done) {
+      chunks++
+      const value = await createHash(chunk.value || '', algorithm)
       body += value
-    })
-    return body
+
+      chunk = await reader?.read()
+    }
+    if (chunks === 1) {
```
exoego, line 46: Why not 
```suggestion
    if (chunks > 0) {
```

usualoma:
Hi @jpdutoit 
Thanks for creating the PR!

Your point about "etag" is correct, and the current hono "middleware/etag" behavior is incorrect. I also believe that this PR will fix the "middleware/etag" behavior.

### sha1

```ts
import { sha1 } from '... /... /utils/crypto'.
```

I think it is an incorrect result for the name `sha1` that the value changes depending on how the chunk is split, even if the same data is passed. I think the same hash value should always be returned for the same data. I think it would be good if the behaviour of the following test would be.

https://github.com/honojs/hono/pull/2918/files#diff-8dde9c6181523f43830fbad5a7e2eeda12b860c5d712ec628e883bd58ad4904aR44-R59


### etag

However, since "etag" does not require a strict "sha1 value for data", your method of generating "etag" is suitable in terms of saving memory usage about "etag".

For that reason, I believe it is better to resolve the problem in "middleware/etag" regarding the generation of "etag" when split into multiple chunks.

jpdutoit:
This is all good comments, haven't had time to revisit this yet...

jpdutoit:
For fixing the sha1 part, it would be better with a solution that supports true incremental hashing instead of needing to load the whole thing into memory, not sure what portable options there are. Unfortunately subtle.crypto does not support that..

usualoma:
Hi @jpdutoit
Can this pull request be updated?

As you say, incremental hashing would be best. However, `crypto.subtle` does not support incremental hashing, and even if we implemented it ourselves in TypeScript, the performance would not be that good.

Since we have received a report from [another person](https://github.com/honojs/hono/issues/3536), it would be good to solve this problem temporarily using the following approach, but what do you think?

https://github.com/honojs/hono/pull/2918

jpdutoit:
Sorry for the delay, I'm not using Hono at the moment...

But looks like you fixed it in the meantime! So I'll close this PR.

PR #2915, opened 2024-06-06
fix(types): #2912: interfaces array's respond typed as `never`

### LINKED ISSUE #2912: RPC output type becomes `never` when returning an array of interface

4.4.2

Bun

```typescript
import { Hono } from "hono";

// common
interface TestInterface {
  test: string
}
const testValue: TestInterface[] = [{ test: 'test' }]

// server

const testApi = new Hono()
  .get(
    '/test',
    async c => {
      return c.json(testValue)
    }
  )

export type Test = typeof testApi

// client
type WrongType = {
  wrong: string
}

const testFunction = (test: WrongType) => {
  console.log(test)
}

const client = hc<Test>('http://localhost')
const resp = await client.test.$get()
const result = await resp.json()

testFunction(testValue) // this case reports an error
testFunction(result)  // this case does not report any errors(type of result is never)
```

Type checking should be failed in both case.

### What do you see instead?

The latter case does not report any type error.

### Additional information

The problems does not reproduce when using type alias instead of interface.

NamesMT:
Resolves #2912

This PR introduces `SimplifyDeepArray` type to handle the case when user response with an array of interfaces.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -461,6 +461,13 @@ describe('Merge path with `app.route()`', () => {
         ok: true,
       })
     }),
+    http.get('http://localhost/api/searchArray', async () => {
+      return HttpResponse.json([
+        {
+          ok: true,
+        },
+      ])
+    }),
     http.get('http://localhost/api/foo', async () => {
       return HttpResponse.json({
         ok: true,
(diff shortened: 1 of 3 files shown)


yusukebe (approved):
LGTM!

yusukebe:
Thank you for the quick fix!

PR #2916, opened 2024-06-06
feat(jsx): support htmlfor attribute alias

akira-tsuno:

- [x] Add tests
- [x] Run tests

To enhance compatibility with React in hono/jsx, we have added support for the `htmlFor` alias. (The relevant code in React can be found at [this file](https://github.com/facebook/react/blob/3730b40e9bbacef0279f6d120b344c1544cb38ba/packages/react-dom-bindings/src/shared/getAttributeAlias.js#L12))

By supporting `htmlFor` in hono/jsx, we gain the following benefits:

Composite components utilizing HTML-only toggles, such as a `<label htmlFor="toggle">` and `<input type="checkbox" id="toggle"/>` combination, will function properly.

For instance, in daisyUI, which works purely with Tailwind, the Drawer component will operate as intended without any code modifications when used with hono/jsx.

### DIFF
--- a/src/jsx/utils.test.ts
+++ b/src/jsx/utils.test.ts
@@ -1,4 +1,74 @@
-import { styleObjectForEach } from './utils'
+import { normalizeIntrinsicElementProps, styleObjectForEach } from './utils';
+
+
+describe('normalizeIntrinsicElementProps', () => {
+  it('should convert className to class', () => {
+    const props: Record<string, unknown> = {
+      className: 'test-class',
+      id: 'test-id',
+    };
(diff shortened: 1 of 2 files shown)


yusukebe:
@akira-tsuno Thanks for the PR. Looks good to me!

@usualoma, Can you review this, too?

usualoma:
Hi @akira-tsuno 
Thank you. I think it's good!

yusukebe:
@akira-tsuno @usualoma Thanks!

Merging now.

PR #2920, opened 2024-06-06
fix(types): `SimplifyDeepArray` should now actually be "deep"

### LINKED ISSUE #2878: With typescript, type inference is lost if the json returned is not an object but an array

4.3.11"

node

``` js
const app = new Hono()
  .get("/", async (c) => {
    const data = await readData();
    return c.json(data);
  })
  .get(
    "/filtration",
    zValidator(
      "query",
      z.object({
        pageSize: z.number().default(10),
      })
    ),
    async (c) => {
      const { pageSize } = c.req.valid("query");
      const data = await readData();
      const result = await splitArrayBySize(data, pageSize);
      return c.json(result);
    }
  );

export default app;

```

``` js
  client.api.issues.filtration.$get({ query: {} }).then((res) => {
    const value = res.json();
// never
  });
```

Hopefully, the result will be returned correctly

### What do you see instead?

_No response_

### Additional information

If you return objects instead, everything is fine

``` js
const app = new Hono()
  .get("/", async (c) => {
    const data = await readData();
    return c.json(data);
  })
  .get(
    "/filtration",
    zValidator(
      "query",
      z.object({
        pageSize: z.number().default(10),
      })
    ),
    async (c) => {
      const { pageSize } = c.req.valid("query");
      const data = await readData();
      const result = await splitArrayBySize(data, pageSize);
      return c.json({result});
    }
  );

export default app;
```

NamesMT:
Resolves #2878

#2915 added `SimplifyDeepArray` but it wasn't actually deep so nested interface array was still bugged as `never`.
It is now actually deep and should fix that.

I have also split JSONRespond's return declaration to `JSONRespondReturn` type for DRY-er code structure because of some weird bug with typescript yelling `type is possibly infinite` at line ~664 when `ReturnType<JSONRespond>` is used as return declaration,
We could remove this type-split commit and put the full type declaration at line ~668 if you prefer not splitting it.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -552,8 +552,12 @@ describe('Merge path with `app.route()`', () => {
     interface ExtraDeepInterface {
       l4: DeepInterface
     }
-    type verifyDeepInterface = Expect<Equal<SimplifyDeepArray<DeepInterface> extends JSONValue ? true : false, true>>
-    type verifyExtraDeepInterface = Expect<Equal<SimplifyDeepArray<ExtraDeepInterface> extends JSONValue ? true : false, true>>
+    type verifyDeepInterface = Expect<
+      Equal<SimplifyDeepArray<DeepInterface> extends JSONValue ? true : false, true>
+    >
+    type verifyExtraDeepInterface = Expect<
+      Equal<SimplifyDeepArray<ExtraDeepInterface> extends JSONValue ? true : false, true>
+    >
   })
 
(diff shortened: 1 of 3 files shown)


NamesMT:
@yusukebe please help me review it, thanks!

yusukebe (approved):
LGTM 💯

yusukebe:
@NamesMT Great work! Thanks!

PR #2922, opened 2024-06-06
refactor(middleware/serve-static): call getContent only once if the file does not exist

usualoma:
Refactored so that `getContent()` was called twice in the same path when 404, so it is now called once.

- [x] Add tests
- [x] Run tests


yusukebe:
@usualoma It's a nice refactoring. Thanks!

PR #2926, opened 2024-06-06
fix(aws-lambda): handle multiple cookies in streaming responses

### LINKED ISSUE #2921: Setting multiple cookies at once does not work using lambda streamHandle

4.4.3

AWS Lambda

Create a response of a request which is setting multiple cookies with `Set-Cookie` header.

All cookies are sent on headers to the client response.

### What do you see instead?

Only one header is sent to the client response.

### Additional information

[This line in the code](https://github.com/honojs/hono/blob/main/src/adapter/aws-lambda/handler.ts#L130-L133) does overwrite multiple headers with the same name.

[As documented here](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html) the cookies should be returned as an array of strings to the lambda runtime.

If you accept this, I'll provide a PR to fix this.

KnisterPeter:
This fix does address that cookies have to be handled
specially in lambda responses.

Fixes #2921

### DIFF
--- a/src/adapter/aws-lambda/handler.ts
+++ b/src/adapter/aws-lambda/handler.ts
@@ -126,10 +126,21 @@ export const streamHandle = <
           context,
         })
 
+        const headers: Record<string, string> = {}
+        const cookies: string[] = []
+        res.headers.forEach((value, name) => {
+          if (name === 'set-cookie') {
+            cookies.push(value)
+          } else {
+            headers[name] = value
+          }
+        })
+
         // Check content type
         const httpResponseMetadata = {
           statusCode: res.status,
-          headers: Object.fromEntries(res.headers.entries()),
+          headers,
+          cookies,
         }
 
         // Update response stream
--- a/runtime_tests/lambda/stream-mock.ts
+++ b/runtime_tests/lambda/stream-mock.ts
@@ -1,0 +1,43 @@
+import { Writable } from 'node:stream'
+import { vi } from 'vitest'
+import type {
+  APIGatewayProxyEvent,
+  APIGatewayProxyEventV2,
+} from '../../src/adapter/aws-lambda/handler'
+import type { LambdaContext } from '../../src/adapter/aws-lambda/types'
+
+type StreamifyResponseHandler = (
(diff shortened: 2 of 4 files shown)


KnisterPeter:
It fails in the coverage check because of some GitHub rate-limit. 😿

src/adapter/aws-lambda/handler.ts
```diff
@@ -126,10 +126,21 @@ export const streamHandle = <
...
+            headers[name] = value
+          }
+        })
+
         // Check content type
         const httpResponseMetadata = {
           statusCode: res.status,
-          headers: Object.fromEntries(res.headers.entries()),
+          headers,
+          cookies,
```
watany-dev, line 143: ok.

https://github.com/jill64/types-lambda/blob/main/types/streaming/ResponseStreamMetadata.ts

KnisterPeter:
@yusukebe @usualoma Can you share an update on this bugfix? Will it be able to get merged? If not what do I need to provide to have it fixed?

yusukebe:
@KnisterPeter 

Sorry for not responding to you! It seems to be good. But if you can, can you write a unit test?

yusukebe:
@watany-dev Do you have any ideas for testing `streamHandle`?

KnisterPeter:
@yusukebe Found a way to test the cookies (and other headers) in the stream. Can you accept that solution?

KnisterPeter:
@yusukebe What else can I do to get this merged?

yusukebe (approved):
LGTM!

yusukebe:
@KnisterPeter 

Sorry for my late reply. This is great! I appreciate you added the test. Let's go with it. I'll merge it and release a minor version, including it later. Thanks.

KnisterPeter:
Hi @yusukebe, can you do a minor release with this patch? We are seeing this issue and waiting for a fix.

yusukebe:
@KnisterPeter 

I've released the `4.4.6` which includes this fix now! Thanks!

Issue #2925, opened 2024-06-06
Improve React compability with NPM Aliases


gaetan-puleo:
### What is the feature you are proposing?

We could create a package to alias hono/jsx/dom.

This package will only reexport everything coming from hono/jsx/dom.

Inspiration : https://www.npmjs.com/package/@preact/compat

It will be possible to do something like this `yarn add react@npm:@hono/jsx-dom-compat`, `bun add react@npm:@hono/jsx-dom-compat` to create a React alias without Vite.

It's native and use npm aliases https://bun.sh/guides/install/npm-alias

yusukebe:
@gaetan-puleo Thank you for the request!

@usualoma I'll check it later, but I want you to take a look. It would be super cool if we could do it.

gaetan-puleo:
On my side I created a POC to test the idea.

https://www.npmjs.com/package/hono-jsx-dom-compat

I get the Vite Starter page working with Hono

![image](https://github.com/honojs/hono/assets/12990773/e59b27b9-5717-494e-baca-892e1d4d8ca5)

To test the POC, you can  create a new vite project with react and do `yarn add react@npm:hono-jsx-dom-compat react-dom@npm:hono-jsx-dom-compat`


I will describe the steps I used to make it works

gaetan-puleo:
First I cloned https://github.com/honojs/middleware

I duplicated the hello middleware.


I created three files in the src folder.
index.ts, jsx-dev-runtime.ts and client.ts

Here the content of the files.

client.ts
```ts
export * from 'hono/jsx/dom/client';
import {hydrateRoot, createRoot} from 'hono/jsx/dom/client'

export default {hydrateRoot, createRoot}
```

index.ts
```ts
export * from 'hono/jsx/dom';
export * as default from 'hono/jsx/dom'
```

and jsx-dev-runtime.ts
```ts
export * from 'hono/jsx/dom/jsx-dev-runtime';
```

After I updated the package.json
```json
  "name": "hono-jsx-dom-compat",
  "version": "0.1.8",
  "description": "hono react compatibility layer",
  "main": "dist/index.js",
  "module": "dist/index.mjs",
  "types": "dist/index.d.ts",
  "files": [
    "dist"
  ],
  "scripts": {
    "test": "vitest --run",
    "build": "tsup ./src/index.ts src/client.ts src/jsx-dev-runtime.ts --format esm,cjs --dts",
    "publint": "publint",
[...]

PR #2929, opened 2024-06-07
feat(jsx/dom): export createRoot and hydrateRoot from jsx/dom/client as default

usualoma:
Small additions to make #2925 easier to implement.

- [x] Add tests
- [x] Run tests


yusukebe (approved):
LGTM!

yusukebe:
Thank @usualoma !

I'm wondering which version we should include this change in. So, keeping not merging for a while.

PR #2937, opened 2024-06-08
fix(jsx): rename `Hono` to `JSX` and export `JSX` namespace

### LINKED ISSUE #2927: HTMX example is not working per v4.4

4.4.4

Bun

- `bun create hono@latest` -> `x-basic`
- `bun i -D typed-htmx`
- Add the snippets from <https://hono.dev/snippets/htmx#typed-htmx>:
  ```ts
  import "typed-htmx";

  declare global {
    namespace Hono {
      interface HTMLAttributes extends HtmxAttributes {
      }
    }
  }
  ```

When working with '.tsx' files, `hx-` attributes should be available.

### What do you see instead?

Typescript declaration merging failure.

### Additional information

Was working on v4.3.

yusukebe:
Fixes #2927

Currently, the definition for HTML tags elements and attributes are exported as a namespace named `Hono` in `jsx/intrinsic-elements.ts`. This is not bad, but the namespace `Hono` is not exported for users. So, it's impossible to extend the values such as `HTMLAttributes`.  In this PR, it will be exported from `hono/jsx`. And I renamed `Hono` to `JSX` because the name `Hono` will cause confusion. That namespace was not used in a global. So I think it will not be a breaking change.

With this PR, you can write the JSX with the HTMX attribute with the following code:

```tsx
// You may have to write `import "typed-htmx"`

declare module 'hono/jsx' {
  namespace JSX {
    interface HTMLAttributes extends HtmxAttributes {}
  }
}

app.get('/', (c) => {
  return c.render(
    <div hx-get="/foo">
      <h1>Hello!</h1>
    </div>
  )
})
```

By the way, it was exported in `global` before `v4.4.0`, but it was not valid for JSR publishing and polluted the global. So, it was changed to exported locally in the module.

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/jsx/base.ts
+++ b/src/jsx/base.ts
@@ -3,7 +3,10 @@ import { escapeToBuffer, stringBufferToString } from '../utils/html'
 import type { HtmlEscaped, HtmlEscapedString, StringBuffer } from '../utils/html'
 import type { Context } from './context'
 import { globalContexts } from './context'
-import type { Hono, IntrinsicElements as IntrinsicElementsDefined } from './intrinsic-elements'
+import type {
+  JSX as HonoJSX,
+  IntrinsicElements as IntrinsicElementsDefined,
+} from './intrinsic-elements'
 import { normalizeIntrinsicElementProps, styleObjectForEach } from './utils'
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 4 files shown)


yusukebe:
Hey @usualoma What do you think of this PR?

usualoma:
@yusukebe Thank you.
It makes sense 👍

yusukebe:
@usualoma Thanks! I'll merge it.

yusukebe:
P.S.

We can write like this instead of the document: https://hono.dev/guides/jsx#override-type-definitions

```ts
declare module 'hono/jsx' {
  namespace JSX {
    interface IntrinsicElements {
      'my-custom-element': HTMLAttributes & {
        'x-event'?: 'click' | 'scroll'
      }
    }
  }
}
```

PR #2944, opened 2024-06-09
feat(jsx/dom): Improve compatibility React (useCallback, React.StrictMode)

usualoma:
### `useCallback`

The following was a type error when trying to register a function that takes arguments, so it was corrected to the same definition as in React.

https://github.com/honojs/hono/compare/main...usualoma:hono:feat/react-compat-type-strict?expand=1#diff-8f050239e20450fd8425feb7efbda60df851ebbc32bec8139d6d9ad1b53a7bc0R1688

### `React.StrictMode`

It was simply exported as an alias of `Fragment`. There is no development mode functionality; when React users want to try "hono/jsx", they might try it in code where `StrictMode` is used, and it would be a waste to have an error at that time, so it would be better to add it.

https://ja.react.dev/reference/react/StrictMode

- [x] Add tests
- [x] Run tests


yusukebe (approved):
LGTM!

PR #2946, opened 2024-06-09
refactor(hono-base): don't check 1st argument of `app.on()`

yusukebe:
This PR removes checking the actual value of the first argument on `app.on()`. I think it is ok not to check because the argument is typed as required. And if the coming actual value is undefined, it will be checked in the initialization phase, so it's no problem just throwing an error.

- [ ] Add tests
- [ ] Run tests


yusukebe:
Hey @usualoma , what do you think of this? I think removing the lines is not a problem.

usualoma:
@yusukebe Empty strings do not make sense and should not be checked 👍

yusukebe:
@usualoma Thanks!

PR #2949, opened 2024-06-09
refactor(context): remove unnecessary initialization add add tests for Context

yusukebe:
This refactoring removed unnecessary initialization in `c.get()` and `c.set()`. And added tests for improving coverage.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -237,7 +237,7 @@ export class Context<
    * ```
    */
   env: E['Bindings'] = {}
-  private _var: E['Variables'] = {}
+  private _var: E['Variables'] | undefined
   finalized: boolean = false
   /**
    * `.error` can get the error object from the middleware if the Handler throws an error.
@@ -480,8 +480,8 @@ export class Context<
 ```
    */
   set: Set<E> = (key: string, value: unknown) => {
-    // @ts-expect-error this._var is initialized as {}
-    this._var[key] = value
+    this._var ??= {}
+    this._var[key as string] = value
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/context.ts
```diff
@@ -480,8 +480,8 @@ export class Context<
 ```
    */
   set: Set<E> = (key: string, value: unknown) => {
-    this._var ??= {}
-    this._var[key as string] = value
+    // @ts-expect-error this._var is initialized as {}
```
yusukebe, line 483: It throws a type error that points to the possibility of `undefined`. But it has been initialized above. So ignore it.

yusukebe:
@usualoma Sorry for calling you again and again. Please review this!

usualoma:
Hi @yusukebe 

Some time ago, in applications that did not use `app.set()` or `app.get()`, this was done to avoid the cost of initialising the private `this._var`. It is this PR that `this._var` is now always initialised.

https://github.com/honojs/hono/pull/1406
https://github.com/honojs/hono/pull/1406/files#diff-1c6b79733c620b41d7f8b63deb644f09e5fda79664918394b443c2a348731605L102

It is quite questionable whether this optimisation is still valid today, but if we were to reproduce the aims of that time, the following change would be appropriate.


```diff
diff --git a/src/context.ts b/src/context.ts
index e5eaa581..d067b048 100644
--- a/src/context.ts
+++ b/src/context.ts
@@ -237,7 +237,7 @@ export class Context<
    * ```
    */
   env: E['Bindings'] = {}
-  private _var: E['Variables'] = {}
+  private _var: E['Variables'] | undefined
   finalized: boolean = false
   /**
    * `.error` can get the error object from the middleware if the Handler throws an error.
```

If there is no need to reproduce previous aims, this is a good change as this PR change makes "context.ts" very simple.

yusukebe:
Hi @usualoma 

You are right! No need to initialize it. I've fixed it with your patch and will merge this now. Thanks!

PR #2958, opened 2024-06-11
feat(adaptor): Remove `unknown` from AddressType

### LINKED ISSUE #2943: Retrive client `addressType` by ip address string

Hello.

I have a feature suggestion for the great getConnInfo helper that was recently added.

Currently, in Cloudflare Workers, the addressType always seems to return `unknown`.

https://github.com/honojs/hono/blob/e72787968fd19f56e5d7cbbc5637b5872e5a2a4a/src/adapter/cloudflare-workers/conninfo.ts#L6

Maybe a function like the following could be written to determine if the addressType is ipv4 or ipv6?

```typescript
function determineAddressType(ip: stirng) {
  if (ip.includes('.')) {
    return 'IPv4'; }
  } else if (ip.includes(':')) {
    return 'IPv6'; }
  } else {
    return 'unknown'; }
  }
}
```

yasuaki640:
close #2943 

Fixed to return undefined if addressType cannot be obtained from runtime.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/adapter/bun/conninfo.test.ts
+++ b/src/adapter/bun/conninfo.test.ts
@@ -1,10 +1,17 @@
 import { Context } from '../../context'
 import { HonoRequest } from '../../request'
 import { getConnInfo } from './conninfo'
+import type { AddressType } from '../../helper/conninfo'
 
-const createRandomBunServer = () => {
-  const address = Math.random().toString()
-  const port = Math.floor(Math.random() * (65535 + 1))
+const createRandomBunServer = ({
+  address = Math.random().toString(),
+  port = Math.floor(Math.random() * (65535 + 1)),
+  family = 'IPv6',
+}: {
+  address?: string
+  port?: number
+  family?: AddressType | string
+} = {}) => {
   return {
     address,
     port,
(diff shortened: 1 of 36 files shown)


src/adapter/bun/conninfo.test.ts
```diff
@@ -1,18 +1,25 @@
...
-  const port = Math.floor(Math.random() * (65535 + 1))
+const createRandomBunServer = ({
+  address = Math.random().toString(),
+  port = Math.floor(Math.random() * (65535 + 1)),
+  family = 'IPv6',
+}: {
+  address?: string
+  port?: number
+  family?: AddressType | string
+} = {}) => {
```
yasuaki640, line 14: I was wondering if I should add a test case, but if it is not needed, I will remove it.

src/adapter/bun/conninfo.test.ts
yasuaki640, line 1: I was wondering if I should add a test case, but if it is not needed, I will remove it.

yusukebe:
@yasuaki640 Thank you for the PR!

@nakasyou @usualoma  Can you review this?

If it's accepted, I think it would be good to include it in the next minor release `v4.5.0`.

yusukebe:
Hi @yasuaki640 

I'll merge this into the `next` branch for the next minor release `v4.5.0` now. Thank you!

PR #2973, opened 2024-06-14
use correct return type for c.html depending on input

asmadsen:
I had some issues with using `c.html` with a non-promise body that had the return-type `Response | Promise<Response>`. This should use correct return-type depending on the input.

### DIFF
--- a/src/context.test.ts
+++ b/src/context.test.ts
@@ -36,7 +36,7 @@ describe('Context', () => {
   })
 
   it('c.html()', async () => {
-    const res = await c.html('<h1>Hello! Hono!</h1>', 201, { 'X-Custom': 'Message' })
+    const res: Response = c.html('<h1>Hello! Hono!</h1>', 201, { 'X-Custom': 'Message' })
     expect(res.status).toBe(201)
     expect(res.headers.get('Content-Type')).toMatch('text/html')
     expect(await res.text()).toBe('<h1>Hello! Hono!</h1>')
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @asmadsen Thank you for the PR.

Hey @usualoma Question. Can we think, if the first argument of `c.html()` is `Promise`, the return type is always `Promise<Response>`?

usualoma:
Hi @asmadsen Thank you!

@yusukebe Yes, a Promise<Response> is returned whenever a Promise<string> is given!

asmadsen:
@yusukebe As @usualoma answered, if the first argument is Promise the return type is always `Promise<Response>`.

I saw there was a type error in the tests, so I fixed that. As well as added explicit type declarations so it would give an type error if the expected type is not returned.

src/context.test.ts
```diff
@@ -36,21 +36,22 @@ describe('Context', () => {
   })
 
   it('c.html()', async () => {
-    const res = await c.html('<h1>Hello! Hono!</h1>', 201, { 'X-Custom': 'Message' })
+    const res: Response = c.html('<h1>Hello! Hono!</h1>', 201, { 'X-Custom': 'Message' })
```
yusukebe, line 39: Is the type definition `: Response` necessary? I imagined the following code:

```ts
const res = c.json({ message: 'Hello' }, 201, { 'X-Custom': 'Message' })
```
It's okay to have type declarations, but if they are inferred, I prefer to use them as-is, and Hono's code consistently does so.

src/context.test.ts
```diff
@@ -36,21 +36,22 @@ describe('Context', () => {
...
     expect(res.status).toBe(201)
     expect(res.headers.get('Content-Type')).toMatch('text/html')
     expect(await res.text()).toBe('<h1>Hello! Hono!</h1>')
     expect(res.headers.get('X-Custom')).toBe('Message')
   })
 
   it('c.html() with async', async () => {
-    const res = await c.html(
-      new Promise((resolve) => setTimeout(() => resolve('<h1>Hello! Hono!</h1>'), 0)),
+    const resPromise: Promise<Response> = c.html(
```
yusukebe, line 47: If the above is changed, it should be:

```ts
const resPromise = //...
```

src/context.test.ts
yusukebe: @asmadsen 

Sorry! I've missed your comment!

> I saw there was a type error in the tests, so I fixed that. As well as added explicit type declarations so it would give an type error if the expected type is not returned.

Understood well. Let's go with this.

yusukebe (approved):
LGTM!

yusukebe:
@asmadsen 

Thank you for your contribution! I'll merge this now.

PR #2975, opened 2024-06-15
test: introduce build tests

usualoma:
We should test for the following aspects.

* The product of the build does not contain unintended files (as a result of tree shaking).
* The file size has not increased unintentionally.

As for the file size check, information could be added as a comment to each pull request (like coverage) in the future, but this pull request does not go that far.

This pull request will enable the following outputs to be obtained.

### `vitest -c .vitest.config/build.ts --run`

```
 ✓ build_tests/build.test.ts (5)
   ✓ app-jsx (1)
     ✓ should not be included /jsx/dom/ in sources
   ✓ app-simple (2)
     ✓ should not be included http-exception in sources
     ✓ should not be included /jsx/ in sources
   ✓ jsx-dom-all (1)
     ✓ should not be included /helper/html in sources
   ✓ jsx-dom-simple (1)
     ✓ should not be included /jsx/base in sources
```

### `tsx ./build_tests/report.ts`

```
app-jsx: 8.9 KiB
app-simple: 7.5 KiB
jsx-dom-all: 5.2 KiB
jsx-dom-simple: 3.5 KiB
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -4,6 +4,6 @@ module.exports = {
     '@typescript-eslint/unbound-method': 'error',
   },
   parserOptions: {
-    project: ['./tsconfig.json', './runtime_tests/tsconfig.json'],
+    project: ['./tsconfig.json', './runtime_tests/tsconfig.json', './build_tests/tsconfig.json'],
   },
 }
--- a/.vitest.config/build.ts
+++ b/.vitest.config/build.ts
@@ -1,0 +1,9 @@
+import config from '../vitest.config'
+config.esbuild = {
+  jsx: 'automatic',
+  jsxImportSource: __dirname + '/../src/jsx',
+}
+if (config.test) {
+  config.test.include = ['build_tests/**/(*.)+(spec|test).+(ts|tsx|js)']
+}
+export default config
--- a/.vitest.config/build.ts
+++ b/.vitest.config/build.ts
@@ -5,5 +5,8 @@ config.esbuild = {
 }
 if (config.test) {
   config.test.include = ['build_tests/**/(*.)+(spec|test).+(ts|tsx|js)']
+  if (config.test.coverage) {
+    config.test.coverage.enabled = false
+  }
 }
 export default config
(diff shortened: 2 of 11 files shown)


usualoma:
Errors in the test have been fixed. The content added by this pull request is excluded from the coverage measurement.

yusukebe:
Hi @usualoma 

I also thought we had to have a measuring bundle/file size like yours. This seems great, but I have another idea. Please wait a bit until I show you it. Thanks!

yusukebe:
Hey @usualoma 

Regarding this matter, I think we should first monitor bundle sizes. Ideally, the CI will tell us the current bundle sizes and how they have increased from the main branch.

To do it, I found 3rd party library instead of implementing it by myself. There are several libraries, but the Size Limit seems to be best:

https://github.com/ai/size-limit

I've tried a little with CLI in my machine:

CleanShot 2024-06-21 at 23 47 57@2x

At first glance, I couldn't find a good way to integrate with GitHub Actions, though it has integrations with CircleCI, TravisCI, etc.

Anyway, we can use it or refer to it.

Jayllyz:
@yusukebe Regarding github action, I think this is what you are looking for : https://github.com/andresz1/size-limit-action

yusukebe:
@Jayllyz 

Ohh, thanks. Can this compare the result of PR with the main's?

Jayllyz:
I don't think so. 
A workraound would be to update the size limit during the release workflow. (I don't know if this is the solution thing to do).

usualoma:
@yusukebe Thank you!
I too think "ai/size-limit" is cool.

Bundle size is certainly the most important.

I was hoping in addition to that to be able to check that 'certain modules are not included', as achieved in the pull request https://github.com/honojs/hono/pull/2885. I think "a module is included that shouldn't be included" could be considered an obvious "bug".


However, it is expensive to keep track of 'modules that should not be included' in detail, and at the moment it would be fine to just use 'ai/size-limit'.

PR #2985, opened 2024-06-17
test: test uncovered return statement

### LINKED ISSUE #2980: Suggestion for Improving Test Coverage of client/utils.ts

In utils.test.ts it looks like early return statements are not executed in the test code.

https://app.codecov.io/gh/honojs/hono/blob/main/src%2Fclient%2Futils.ts#L40

As a suggested fix, why not modify utils.test.ts as follows?

```diff
 describe('deepMerge', () => {
   it('should return the source object if the target object is not an object', () => {
     const target = null
-    const source = { a: 1 }
+    const source = 'not object' as unknown as Record<string, unknown>
```

As another suggestion, if we assume that only objects are passed to the source argument of the deepMerge function, the conditional expression of the if statement could be as follows.

```diff
 export function deepMerge<T>(target: T, source: Record<string, unknown>): T {
-  if (!isObject(target) && !isObject(source)) {
+  if (!isObject(target)) {
     return source as T
   }
```

If this modification is necessary, I will make a PR.

yasuaki640:
close #2980 

Adjusted the value passed as the second argument to improve coverage.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/client/utils.test.ts
+++ b/src/client/utils.test.ts
@@ -97,7 +97,7 @@ describe('removeIndexString', () => {
 describe('deepMerge', () => {
   it('should return the source object if the target object is not an object', () => {
     const target = null
-    const source = { a: 1 }
+    const source = 'not an object' as unknown as Record<string, unknown>
     const result = deepMerge(target, source)
     expect(result).toEqual(source)
   })


yusukebe (approved):
LGTM!

yusukebe:
@yasuaki640 Thanks!

PR #2987, opened 2024-06-18
fix(types): env variables override ContextVariableMap

KaelWD:
```ts
declare module 'hono' {
    test: 'root'
  }
}

const middleware = createMiddleware<{
  Variables: {
    test: 'override'
  }
}>(async (c, next) => {
  c.set('test', 'override')
  await next()
})

const router = new Hono()
  .get(
    '/test',
    middleware,
    async c => {
      const test = c.get('test')
      //    ^? "root"
      return c.text(test) // 'override', type doesn't match
    },
  )

const app = new Hono()
  .use(async (c, next) => {
    c.set('test', 'root')
    await next()
  })
  .route('/', router)
```

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -69,8 +69,8 @@ export type Layout<T = Record<string, any>> = (props: T) => any
  * @template E - Environment type.
  */
 interface Get<E extends Env> {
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @KaelWD !

We have to fix it. I've added a test by myself. It looks good, so I'm merging now. Thanks!

PR #3004, opened 2024-06-20
test: add tests for buffer.ts

### LINKED ISSUE #2995: Suggestion for Improving Test Coverage of utils/buffer.ts

The equals function in utils/buffer.ts does not appear to be used.

https://app.codecov.io/github/honojs/hono/blob/main/src%2Futils%2Fbuffer.ts#L9

If it is a function used on the user side, I would suggest adding test code.
Or, if it is not going to be used in the future, it would be better to remove it.

If either of these fixes are needed, I will send out a pull request!
If no modifications are needed, the issue should be closed!

yasuaki640:
close #2995

Add some tests for uncovered lines.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/utils/buffer.test.ts
+++ b/src/utils/buffer.test.ts
@@ -1,5 +1,31 @@
 import { SHA256 as sha256CryptoJS } from 'crypto-js'
-import { bufferToFormData, bufferToString, timingSafeEqual } from './buffer'
+import { bufferToFormData, bufferToString, equal, timingSafeEqual } from './buffer'
+
+describe('equal', () => {
+  it('should return true for identical ArrayBuffers', () => {
+    const buffer1 = new ArrayBuffer(1)
+    const buffer2 = buffer1
+    expect(equal(buffer1, buffer2)).toBe(true)
+  })
+
+  it('should return false for ArrayBuffers of different lengths', () => {
+    const buffer1 = new ArrayBuffer(1)
+    const buffer2 = new ArrayBuffer(2)
+    expect(equal(buffer1, buffer2)).toBe(false)
+  })
+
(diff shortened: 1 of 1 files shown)


src/utils/buffer.test.ts
```diff
@@ -43,6 +70,7 @@ describe('buffer', () => {
     expect(await timingSafeEqual({ a: 1 }, { a: 2 })).toBe(false)
     expect(await timingSafeEqual([1, 2], [1, 2])).toBe(false)
     expect(await timingSafeEqual([1, 2], [1, 2, 3])).toBe(false)
+    expect(await timingSafeEqual('a', 'b', () => undefined)).toBe(false)
```
yasuaki640, line 73: Test covering the first if statement branch

https://github.com/honojs/hono/blob/7ee829be6bef89746a0157c007ea9668141ddce9/src/utils/buffer.ts#L41-L43

src/utils/buffer.test.ts
```diff
@@ -1,5 +1,32 @@
 import { SHA256 as sha256CryptoJS } from 'crypto-js'
-import { bufferToFormData, bufferToString, timingSafeEqual } from './buffer'
+import { bufferToFormData, bufferToString, equal, timingSafeEqual } from './buffer'
+import { expect } from 'vitest'
```
yusukebe, line 3: Is this line necessary? I think the `expect` function is already imported globally.

yasuaki640: Okay. I've deleted this import.

yusukebe (approved):
LGTM!

yusukebe:
@yasuaki640 Thanks!

PR #3017, opened 2024-06-21
feat(logger): improve automatic color support detection

NamesMT:
The current color detection is not great and the log message in CloudWatch logs is hard to read.
image

The PR adds `utils/flags.ts` and improve `getColorEnabled()` ~~using `isColorSupported` flag~~ _The flag have been moved to inline of the function to align with the current test cases_.

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/src/utils/color.ts
+++ b/src/utils/color.ts
@@ -3,14 +3,22 @@
  * Color utility.
  */
 
+import { hasTTY, isCI, isWindows } from './flags'
+
 export function getColorEnabled(): boolean {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const { process, Deno } = globalThis as any
 
+  // It is put as a function instead of a constant to only to run as needed in the ternary operator.
+  const isColorSupported = () =>
+    !globalThis.process?.env?.NO_COLOR &&
+    (Boolean(globalThis.process?.env?.FORCE_COLOR) ||
+      ((hasTTY || isWindows) && globalThis.process?.env?.TERM !== 'dumb') ||
+      isCI)
+
   const isNoColor =
     typeof process !== 'undefined'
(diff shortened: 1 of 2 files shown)


src/utils/color.ts
```diff
@@ -3,14 +3,22 @@
  * Color utility.
  */
 
+import { hasTTY, isCI, isWindows } from './flags'
+
 export function getColorEnabled(): boolean {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const { process, Deno } = globalThis as any
 
+  // It is put as a function instead of a constant to only to run as needed in the ternary operator.
```
NamesMT, line 12: Revising in 2025:

`isColorSupported` was put here instead of `flags` because Cloudflare Workers in the past does include the env variables on the global `process.env` and we couldn't check for `NO_COLOR`, `FORCE_COLOR`.

If we're reviving this PR, we move it to `flags` now as Workers have supported it.

src/utils/color.ts
```diff
@@ -3,14 +3,22 @@
...
+  const isColorSupported = () =>
```
yusukebe, line 13: Sorry if I'm misunderstanding. On Cloudflare Workers, this value returns from `isColorSupported()` is always `false`, right?

NamesMT: I think so, unless the user sets `FORCE_COLOR`, anyways, let me do a refresh update and do actual test with Lambda + Workers to see if it works normally.

NamesMT: What do you think of the added `flags.ts`, is it good to you?, or should I inline everything in the `getColorEnabled`?

yusukebe: I think adding `flags.ts` is good. Let's go with it.

PR #3025, opened 2024-06-23
PoC: feat(middleware/hook): introduce hook middleware

usualoma:
### Usage

```ts
import { Hono } from 'hono'
import { hook } from 'hook'

const app = new Hono()
app.use(
  hook({
    before(c, handler, handlerContext, traceId) {
      console.log(traceId, 'before', handler.name)
      handlerContext['start'] = Date.now()
    },
    beforeNext(c, handler, _, traceId) {
      console.log(traceId, 'beforeNext', handler.name)
    },
    afterNext(c, handler, _, traceId) {
      console.log(traceId, 'afterNext', handler.name)
    },
    after(c, handler, handlerContext, traceId) {
      const elapsed = Date.now() - handlerContext['start']
      console.log(traceId, `after ${handler.name} ${elapsed}ms`, handler.name)
    },
  })
)

app.use(async function middlewareA(c, next) {
  await new Promise((resolve) => setTimeout(resolve, 100))
  await next()
})

app.use(async function middlewareB(c, next) {
  await new Promise((resolve) => setTimeout(resolve, 200))
  await next()
})

app.get('/', async function handler(c) {
  await new Promise((resolve) => setTimeout(resolve, 200))
  return c.text('Hello, World!')
})

export default app
```

### Output

```
c0934c4a53236 before middlewareA
c0934c4a53236 beforeNext middlewareA
c0934c4a53236 before middlewareB
c0934c4a53236 beforeNext middlewareB
c0934c4a53236 before handler
c0934c4a53236 after handler 203ms handler
c0934c4a53236 afterNext middlewareB
c0934c4a53236 after middlewareB 406ms middlewareB
c0934c4a53236 afterNext middlewareA
c0934c4a53236 after middlewareA 508ms middlewareA
```

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/middleware/hook/index.ts
+++ b/src/middleware/hook/index.ts
@@ -1,0 +1,53 @@
+import type { Context } from '../../context'
+import type { Env, Handler, MiddlewareHandler, Next } from '../../types'
+
+const isWrapped = Symbol()
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type Hook = (c: Context, handler: Handler, handlerContext: Record<string, any>) => void
+export const hook = <E extends Env = Env>(
+  options: {
+    before?: Hook
+    beforeNext?: Hook
+    afterNext?: Hook
+    after?: Hook
+  } = {}
+): MiddlewareHandler => {
+  function hook(c: Context<E>, next: Next) {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    ;(c.req.matchResult[0] as unknown as [[any]][]).forEach((routeData) => {
+      if (routeData[0][0][isWrapped]) {
+        return
+      }
+
+      const handler = routeData[0][0]
+      const name = handler.name || ''
+      routeData[0][0] = {
+        [name]: function (c: Context, next: Next) {
+          const handlerContext = Object.create(null)
+
+          if (options.before) {
+            options.before?.(c, handler, handlerContext)
+          }
+          const internalNext = () => {
+            options.beforeNext?.(c, handler, handlerContext)
+            const res = next()
+            res.finally(() => options.afterNext?.(c, handler, handlerContext))
+            return res
+          }
+          const res = handler(c, internalNext)
+          if (res instanceof Promise) {
+            res.finally(() => options.after?.(c, handler, handlerContext))
+          } else {
+            options.after?.(c, handler, handlerContext)
+          }
+          return res
+        },
+      }[name]
+      routeData[0][0][isWrapped] = true
+    })
+    return next()
+  }
+  hook[isWrapped] = true
+  return hook
+}
--- a/src/middleware/hook/index.ts
+++ b/src/middleware/hook/index.ts
@@ -3,8 +3,13 @@ import type { Env, Handler, MiddlewareHandler, Next } from '../../types'
 
 const isWrapped = Symbol()
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type Hook = (c: Context, handler: Handler, handlerContext: Record<string, any>) => void
+export type Hook = (
+  c: Context,
+  handler: Handler,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  handlerContext: Record<string, any>,
+  traceId: string
+) => void
 export const hook = <E extends Env = Env>(
   options: {
     before?: Hook
@@ -14,6 +19,9 @@ export const hook = <E extends Env = Env>(
   } = {}
 ): MiddlewareHandler => {
   function hook(c: Context<E>, next: Next) {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    c.set('middleware-hook-trace-id' as any, Math.random().toString(16).slice(2))
+
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     ;(c.req.matchResult[0] as unknown as [[any]][]).forEach((routeData) => {
       if (routeData[0][0][isWrapped]) {
@@ -25,21 +33,22 @@ export const hook = <E extends Env = Env>(
       routeData[0][0] = {
         [name]: function (c: Context, next: Next) {
           const handlerContext = Object.create(null)
+          const traceId = c.get('middleware-hook-trace-id')
 
           if (options.before) {
-            options.before?.(c, handler, handlerContext)
+            options.before?.(c, handler, handlerContext, traceId)
           }
           const internalNext = () => {
-            options.beforeNext?.(c, handler, handlerContext)
+            options.beforeNext?.(c, handler, handlerContext, traceId)
             const res = next()
-            res.finally(() => options.afterNext?.(c, handler, handlerContext))
+            res.finally(() => options.afterNext?.(c, handler, handlerContext, traceId))
             return res
           }
           const res = handler(c, internalNext)
           if (res instanceof Promise) {
-            res.finally(() => options.after?.(c, handler, handlerContext))
+            res.finally(() => options.after?.(c, handler, handlerContext, traceId))
           } else {
-            options.after?.(c, handler, handlerContext)
+            options.after?.(c, handler, handlerContext, traceId)
           }
           return res
         },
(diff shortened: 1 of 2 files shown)


usualoma:
Updated 36267e2

MathurAditya724:
Quite Interesting! Some thoughts I have regarding the API - 

1. Isn't `afterNext` and `after` functions getting called one after the other, so to simplify can't we just have an `after` function?
2. Not sure about `beforeNext` as I don't like some random code running between my code which might alter the response. I believe `beforeNext` func and `before` func of the next middleware will have the same params, so what's the point of `beforeNext`?
3. Can we simplify the `hook` API and make it similar to the `middleware` function?
```ts
app.use(
  hook(async (c, handler) => {
    console.log("Before")
    await handler();
    console.log("After")
  })
)
```

These are just my viewpoints, correct me if I am wrong but I love the idea of a function running between middleware, already thinking about a lot of stuff to build with it 😂

usualoma:
@MathurAditya724 

Thank you. I understand your argument for more simplicity.

But if you write the following, the `handler` can't be a handler as it is in the original. It becomes a wrapped function. Here we have to ask, "What function are you hooking?" so we need to know, so we need to pass a bit more information.

```ts
app.use(
  hook(async (c, handler) => {
    console.log("Before")
    await handler();
    console.log("After")
  })
)
```


### Why beforeNext / afterNext

BeforeNext and afterNext are provided to allow the processing time of ` hardWorkBeforeNext()` and ` hardWorkAfterNext()` to be measured in the following code.

```ts
app.use(async (c, next) => {

  // before hook

  hardWorkBeforeNext()

  // beforeNext hook

  await next()

  // afterNext hook

  hardWorkAfterNext()

  // after hook

})
```

Code-Hex:
@usualoma Thank you very much for your prompt creation of a PoC based on the ideas you presented at the Hono Conference! The image is very close.

The idea originally came from my comment that it would be nice to have a mechanism in place to help Hono users visualize how their applications are being used. The idea on which this idea is based is [`stats.Handler`](https://pkg.go.dev/google.golang.org/grpc/stats#Handler) in grpc-go. As the name implies, if we implement an object that satisfies the `stats.Handler` interface, you can aggregate the usage by the users themselves.

`HandleRPC` is the easiest to imagine. It takes an interface called [`RPCStats`](https://pkg.go.dev/google.golang.org/grpc/stats#RPCStats) as a parameter.

```go
type RPCStats interface {
	// IsClient returns true if this RPCStats is from client side.
	IsClient() bool
}
```

The reason `IsClient` is a method is because gRPC allows client and server middleware and stats Handler to use the same interface.

8 implementations are provided to satisfy this interface. They are [`Begin`](https://pkg.go.dev/google.golang.org/grpc/stats#Begin), [`InHeader`](https://pkg.go.dev/google.golang.org/grpc/stats#InHeader), [`InPayload`](https://pkg.go.dev/google.golang.org/grpc/stats#InPayload), [`InTrailer`](https://pkg.go.dev/google.golang.org/grpc/stats#InTrailer), [`OutHeader`](https://pkg.go.dev/google.golang.org/grpc/stats#OutHeader), [`OutPayload`](https://pkg.go.dev/google.golang.org/grpc/stats#OutPayload), [`OutTrailer`](https://pkg.go.dev/google.golang.org/grpc/stats#OutTrailer), [`End`](https://pkg.go.dev/google.golang.org/grpc/stats#End).

Using these, users can create handlers in this way to collect statistical information.

```go
// HandleRPC implements per RPC tracing and stats implementation.
func (h *serverStatsHandler) HandleRPC(ctx context.Context, rs stats.RPCStats) {
	ri := getRPCInfo(ctx)
	if ri == nil {
		logger.Error("ctx passed into server side stats handler metrics event handling has no server call data present")
		return
	}
	h.processRPCData(ctx, rs, ri.ai)
}

func (h *serverStatsHandler) processRPCData(ctx context.Context, s stats.RPCStats, ai *attemptInfo) {
	switch st := s.(type) {
	case *stats.InHeader:
		if ai.pluginOptionLabels == nil && h.options.MetricsOptions.pluginOption != nil {
			labels := h.options.MetricsOptions.pluginOption.GetLabels(st.Header)
			if labels == nil {
				labels = map[string]string{} // Shouldn't return a nil map. Make it empty if so to ignore future Get Calls for this Attempt.
			}
			ai.pluginOptionLabels = labels
		}
		h.serverMetrics.callStarted.Add(ctx, 1, otelmetric.WithAttributes(otelattribute.String("grpc.method", ai.method)))
	case *stats.OutPayload:
		atomic.AddInt64(&ai.sentCompressedBytes, int64(st.CompressedLength))
	case *stats.InPayload:
		atomic.AddInt64(&ai.recvCompressedBytes, int64(st.CompressedLength))
	case *stats.End:
		h.processRPCEnd(ctx, ai, st)
	default:
	}
}
```

https://github.com/grpc/grpc-go/blob/f199062ef31ddda54152e1ca5e3d15fb63903dc3/stats/opentelemetry/server_metrics.go#L203-L232

If these handlers were actually passed as options inside the framework, they would be called this way.

```go
func (s *Server) processUnaryRPC(ctx context.Context, t transport.ServerTransport, stream *transport.Stream, info *serviceInfo, md *MethodDesc, trInfo *traceInfo) (err error) {
	shs := s.opts.statsHandlers
	if len(shs) != 0 || trInfo != nil || channelz.IsOn() {
		// ...
		var statsBegin *stats.Begin
		for _, sh := range shs {
			beginTime := time.Now()
			statsBegin = &stats.Begin{
				BeginTime:      beginTime,
				IsClientStream: false,
				IsServerStream: false,
			}
			sh.HandleRPC(ctx, statsBegin)
		}
	}

	// Processing to call handler
	// after called

	for _, sh := range shs {
		end := &stats.End{
			BeginTime: statsBegin.BeginTime,
			EndTime:   time.Now(),
		}
		if err != nil && err != io.EOF {
			end.Error = toRPCErr(err)
		}
		sh.HandleRPC(ctx, end)
	}
```

In the same file, InPayload is called immediately after parsing the request body and OutPayload is called immediately after writing the response body. The other InHeader, OutHeader, InTrailer, and OutTrailer call `HandleRPC` at similar times.

https://github.com/grpc/grpc-go/blob/f199062ef31ddda54152e1ca5e3d15fb63903dc3/server.go#L1206-L1246

---

Based on the above, I think that `TagRPC` (routing information) and `HandleRPC` (handling information) hooks are the parts that can be used for Hono. I think users can hook the middleware themselves if they work hard enough, so it may not be necessary to do so.

Since these are hooks for monitoring, they are defined on the assumption that **no errors are thrown**.

And I do not believe they should be used to hack complex business logic. In other words, I believe this should be available as something completely different from existing middleware.

usualoma:
@Code-Hex 

Thank you. That is very helpful information.

I understand that "grpc" provides the following eight,
> Begin, InHeader, InPayload, InTrailer, OutHeader, OutPayload, OutTrailer, End.

What would a generic framework, such as Hono, be able to provide? I still don't understand, when should `handleHandler(c, handlerStats)` be called?

Is there anything we can do by 'providing it as a core feature rather than middleware'?

Code-Hex:
@usualoma Thanks for reply me.

I'm writing just focus for `handleHandler`.

`start` and `end` are mandatory (metrics or trace usecase). I think this should be called immediately after Hono receives the request, and at the last layer when all processing is done and the response is returned.

> Start, End

It is right after the request headers are received and right after the response headers to be returned are **finalized**. Since multiple middleware can modify the request and response, I believe that only the core functionality of the framework can retrieve the finalized information.

- Why are headers important?
  - This is because many of the tracing and request context mechanisms are based on HTTP headers.
    - e.g. https://www.w3.org/TR/trace-context/#trace-context-http-headers-format
  - Often included in request headers via cloud vendors such as GCP.
    - https://cloud.google.com/trace/docs/trace-context

> Header

Payload is a concern of mine as well. Hono has a mechanism to validate payloads such as path parameters and form bodies by using [validator](https://hono.dev/docs/guides/validation) as middleware. `InPayload` is exactly what I thought it would be nice to call `c.req.json()` or `c.req.valid()` when the request payload is successfully parsed.

I thought that `OutPayload` should be similar to the timing of End, a fixed endpoint where the response body is not changed any more.

> Payload

Personally, I consider the Trailer to be optional. gRPC uses it to express Response Status details, so it is important, but the HTTP Web Framework in general does not consider it that important.

PR #3028, opened 2024-06-23
feat(cloudflare-pages): Add Cloudflare Pages middleware handler

### LINKED ISSUE #3015: Cloudflare Pages Middleware
Cloudflare Pages has it's own concept for middleware functions
https://developers.cloudflare.com/pages/functions/middleware/

```ts
export async function onRequest(context) {
  try {
    return await context.next();
  } catch (err) {
    return new Response(`${err.message}\n${err.stack}`, { status: 500 });
  }
}
```

`hono/cloudflare-pages` currently provides a `handle` adapter to use hono with Pages Function, and a `serveStatic` adapter to handle static assets. It would be great if there was a similar adapter for Pages middleware.

My specific use-case is using `HTMLRewriter` to rewrite static assets based on the value of a cookie.

```ts
// functions/_middleware.ts
import { handleMiddleware } from 'hono/cloudflare-pages'

export const onRequest = handleMiddleware(async (c, next) => {
  const myCookie = getCookie(c, 'my_cookie')

  const response = await next()
  // Alternative API similar to Pages middleware
  const response = await c.next()

  return new HTMLRewriter()
    .on('head', new InjectScriptHandler(myCookie))
    .transform(response)
})
```

Let me know if this is worth adding, I'd be happy to contribute

BarryThePenguin:
Adds an adapter for [Cloudflare Pages middleware](https://developers.cloudflare.com/pages/functions/middleware/)

This provides a convenient way to use `Context` and helpers from `hono` within Cloudflare Pages middleware

Fixes #3015

### DIFF
--- a/src/adapter/bun/conninfo.test.ts
+++ b/src/adapter/bun/conninfo.test.ts
@@ -49,7 +49,7 @@ describe('getConnInfo', () => {
   })
   it('should return undefined when addressType is invalid string', () => {
     const { server } = createRandomBunServer({ family: 'invalid' })
-    const c = new Context(new HonoRequest(new Request('http://localhost/')), { env: { server } })
+    const c = new Context(new Request('http://localhost/'), { env: { server } })
 
     const info = getConnInfo(c)
 
--- a/src/adapter/cloudflare-pages/handler.test.ts
+++ b/src/adapter/cloudflare-pages/handler.test.ts
@@ -1,6 +1,7 @@
+import { getCookie } from '../../helper/cookie'
 import { Hono } from '../../hono'
 import type { EventContext } from './handler'
-import { handle } from './handler'
+import { handle, handleMiddleware } from './handler'
 
 type Env = {
   Bindings: {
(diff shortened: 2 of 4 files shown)


src/adapter/cloudflare-pages/handler.test.ts
```diff
@@ -49,3 +50,75 @@ describe('Adapter for Cloudflare Pages', () => {
...
+    const request = new Request('http://localhost/api/foo', {
+      headers: {
+        Cookie: 'my_cookie=1234',
+      },
+    })
+    const env = {
+      TOKEN: 'HONOISCOOL',
+    }
+    const handler = handleMiddleware(async (c, next) => {
+      const cookie = getCookie(c, 'my_cookie')
```
BarryThePenguin, line 65: Included this as an example of my use-case, but it's not entirely necessary as part of the test

src/adapter/cloudflare-pages/handler.ts
```diff
@@ -34,6 +43,24 @@ export const handle: HandleInterface = (app: Hono) => (eventContext: EventContex
...
+      env: executionCtx.env,
+      executionCtx,
+    })
+
+    const response = await middleware(context, async () => {
+      context.res = await executionCtx.next()
+    })
+
+    return response ?? context.res
+  }
```
BarryThePenguin, line 61: This is a very naive implementation, I haven't spent much time looking at the internals of hono yet. Appreciate any feedback on how this might be improved

yusukebe:
@BarryThePenguin 

Thank you for the PR. I'll review this later.

yusukebe:
The current code can't handle an `HTTPException` thrown by a middleware like Basic Auth Middleware. The code below throws a `500` internal server error instead of a `401` authorization error.

```ts
import { basicAuth } from 'hono/basic-auth'

export const onRequest = handleMiddleware(
  basicAuth({
    username: 'foo',
    password: 'bar'
  })
)
```

To handle `HTTPException` correctly, you can write like the following:

```diff
diff --git a/src/adapter/cloudflare-pages/handler.ts b/src/adapter/cloudflare-pages/handler.ts
index 8d3325a..13e6916 100644
--- a/src/adapter/cloudflare-pages/handler.ts
+++ b/src/adapter/cloudflare-pages/handler.ts
@@ -1,5 +1,6 @@
 import { Context } from '../../context'
 import type { Hono } from '../../hono'
+import { HTTPException } from '../../http-exception'
 import { HonoRequest } from '../../request'
 import type { Env, Input, MiddlewareHandler } from '../../types'

@@ -53,9 +54,17 @@ export function handleMiddleware<E extends Env = any, P extends string = any, I
       executionCtx,
     })

-    const response = await middleware(context, async () => {
-      context.res = await executionCtx.next()
-    })
+    let response: Response | void = undefined
+
+    try {
+      response = await middleware(context, async () => {
+        context.res = await executionCtx.next()
+      })
+    } catch (e) {
+      if (e instanceof HTTPException) {
+        response = e.getResponse()
+      }
+    }

     return response ?? context.res
   }
```

src/adapter/cloudflare-pages/handler.ts
```diff
@@ -34,6 +43,24 @@ export const handle: HandleInterface = (app: Hono) => (eventContext: EventContex
   )
 }
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function handleMiddleware<E extends Env = any, P extends string = any, I extends Input = {}>(
```
yusukebe, line 47: Please export `handleMiddleware` from `src/adapter/cloudflare-pages/index.ts`.

```ts
export { handle, handleMiddleware, serveStatic } from './handler'
```

src/adapter/cloudflare-pages/handler.ts
```diff
@@ -34,6 +43,34 @@ export const handle: HandleInterface = (app: Hono) => (eventContext: EventContex
...
+    try {
+      response = await middleware(context, async () => {
+        context.res = await executionCtx.next()
+      })
+    } catch (error) {
+      if (error instanceof HTTPException) {
+        response = error.getResponse()
+      } else {
+        throw error
+      }
```
BarryThePenguin, line 67: I included the code to handle `HTTPException`, but I wasn't sure what should happen with other kinds of errors, so it will re-throw everything else

Looking at the documentation, there is also a [`Context.error`](https://hono.dev/docs/api/context#error) property. Should we handle that here too? Similar to `Context.res`

BarryThePenguin (commented):
I had a bit of trouble reconciling the changes in both the `main` and `next` branches, so I have rebased my changes onto `next`, assuming #3046 will also be in `next` soon 👀

src/adapter/cloudflare-pages/handler.ts
yusukebe: You are right. It should be re-thrown. Thanks.

> Looking at the documentation, there is also a [`Context.error`](https://hono.dev/docs/api/context#error) property. Should we handle that here too?

Yes. Please put the error to `context.error`.

yusukebe:
> I had a bit of trouble reconciling the changes in both the `main` and `next` branches, so I have rebased my changes onto `next`, assuming #3046 will also be in `next` soon 👀 

Ahh, sorry! I've merged the changes of `main` to `next`, which includes #3046.

src/adapter/cloudflare-pages/handler.ts
```diff
@@ -34,6 +43,54 @@ export const handle: HandleInterface = (app: Hono) => (eventContext: EventContex
...
+    if (context.error instanceof HTTPException) {
+      return context.error.getResponse()
+    }
+
+    if (context.error) {
+      throw context.error
+    }
+
+    return context.res
+  }
```
BarryThePenguin, line 91: I'm still not sure all of this is necessary.. I added all the test cases I could think of, so hopefully all the main use cases are covered

src/adapter/bun/conninfo.test.ts
```diff
@@ -49,7 +49,7 @@ describe('getConnInfo', () => {
   })
   it('should return undefined when addressType is invalid string', () => {
     const { server } = createRandomBunServer({ family: 'invalid' })
-    const c = new Context(new HonoRequest(new Request('http://localhost/')), { env: { server } })
```
BarryThePenguin, line 52: Looks like `next` is failing due to `HonoRequest` being removed 👀

yusukebe (approved):
LGTM!

yusukebe:
@BarryThePenguin

Awesome work! It's wonderful that we'll be able to use a lot of Hono's middleware and helpers in the Cloudflare Pages function. Thank you very much. I'll merge this into the `next` and release an RC version now.

PR #3042, opened 2024-06-25
fix(streaming): call stream.abort() explicitly when request is aborted

### LINKED ISSUE #3032: Abort does not work with bun

4.2.4

bun

With the following example inspired by the doc, I cannot get an onAbort to work using Bun, when switching to Node it works normally. I am testing with postman to make the request, and closing it in the ui.
Am I doing something wrong or something is broken ?

```
import { Hono } from "hono";
// import { serve } from "@hono/node-server";

const hono = new Hono();

let id = 0;
hono.get("t", (c) => {
  return streamSSE(c, async (stream) => {
    setTimeout(() => {
      stream.close();
    }, 10000);
    stream.onAbort(() => {
      console.log("aborted");
    });
    while (true) {
      try {
        await stream.writeSSE({
          data: `It is ${new Date().toISOString()}`,
          event: "time-update",
          id: String(id++),
        });
        await stream.sleep(1000);
      } catch (e) {
        console.error(e);
        break;
      }
    }
  });
});

// serve(hono);
export default hono;
```

'aborted' should be logged with bun and node.

### What do you see instead?

'aborted' is not logged with bun

### Additional information

_No response_

usualoma:
fixes #3032

### What is needed in bun?

In bun, when a `ReadableStream` is returned in the response, cancel() is not called if the connection is disconnected, so the hono needs to listen for the connection to be disconnected.

The following workaround is a strange piece of code, but I think it is acceptable because it would not work without it in bun.

https://github.com/honojs/hono/pull/3042/files#diff-b5c92850aac10e22281695a83a3d6d7a23f5455df59a11c597113914ef427774R14-R18

### flaky test

Even stranger still, testing requests with `Bun.serve()` sometimes fails; I tried a number of patterns, including interspersing setTimeout, but I just couldn't get it to stabilise. The instability seems to occur only in unit tests, so if it succeeds on retry, it is considered OK.

0b15369

- [x] Add tests
- [x] Run tests


usualoma:

yusukebe:
Hey @usualoma 

I may not completely understand the correct behavior, but I think it does not work well in Bun with this PR.

Bun (does not work):

https://github.com/honojs/hono/assets/10682/0eca5e0b-7138-4356-817a-b72918a048e6

Node.js (works):

https://github.com/honojs/hono/assets/10682/e7e611e2-ce23-4ec3-b721-fa34c3b98525

Btw, it has the same behavior as Node.js on Deno.

yusukebe:
@usualoma 

Is the `abort` fired when closing the browser?

yusukebe:
Ah, we can use this script:

```ts
import { Hono } from 'hono'
import { streamSSE } from 'hono/streaming'

const app = new Hono()

app.get('t', (c) => {
  return streamSSE(c, async (stream) => {
    stream.onAbort(() => {
      console.log('aborted')
      stream.close()
    })
    for (;;) {
      try {
        await stream.writeSSE({
          data: `It is ${new Date().toISOString()}`,
          event: 'message',
        })
        await stream.sleep(1000)
      } catch (e) {
        console.error(e)
        break
      }
    }
  })
})

app.get('/', (c) => {
  return c.html(`
    <html>
      <body>
        <script>
const controller = new AbortController();
const signal = controller.signal;

const eventSource = new EventSource('/t', { signal: signal });

eventSource.onmessage = function(event) {
  console.log('Message:', event.data);
};

eventSource.onerror = function(error) {
  console.log('Error:', error);
  eventSource.close();
};

setTimeout(() => {
  console.log('Aborting connection');
  eventSource.close();
}, 5000);
        </script>
      </body>
    </html>
  `)
})

export default app
//serve(app)
```

But the behaviors are the same.

usualoma:
@yusukebe Sorry, I think the `stream()` one is working, but the `streamSSE()` one was not. Please wait a bit while I fix it.

usualoma:
@yusukebe 
I finally understood what we needed.

In a bun, the destruction of `c` would start when the response was returned, so we needed to suppress that. In 55b6c8c2, I made it so that c is held until a `readableStream` is destroyed, which solved the problem. This change also made flaky tests more stable.

yusukebe (approved):
LGTM!

yusukebe:
Hi @usualoma !

Thanks for the investigation and implementation. it works well. Merging now.

PR #3058, opened 2024-06-28
fix(client): set Path as the default of Original

m-shaka:
Routings without the leading slash like `app.get('foo', ...)` works well, but it seems like `hc` fails to extract its type correctly.
You can't access `hc<typeof app>('/').foo.$get`, which works on runtime though.

- [x] Add tests
- [x] Run tests
- [ ] ~Add [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code~ not needed, I guess

### DIFF
--- a/src/client/types.test.ts
+++ b/src/client/types.test.ts
@@ -23,3 +23,37 @@ describe('WebSockets', () => {
     >().toEqualTypeOf(true)
   })
 })
+
+describe('without the leading slash', () => {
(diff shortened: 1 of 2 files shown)


yusukebe (approved):
LGTM!

yusukebe:
Looks good to me! I'll merge this now. Thank you for your contribution.

PR #3062, opened 2024-06-30
feat: Introduce Service Worker Adapter

nakasyou:
I added adapter for Service Worker on browser.

I think that Hono is good for creating Service Worker.

In the PR, you can write code like that:
```ts
import { Hono } from 'hono'
import { handle } from 'hono/service-worker'

const app = new Hono()
app.get('/', c => c.text('Hello World'))

self.addEventListener('fetch', handle(app))
```

When app responses 404, adapter fetches the original server.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/jsr.json
+++ b/jsr.json
@@ -5,7 +5,8 @@
     "lib": [
       "dom",
       "dom.iterable",
-      "deno.ns"
+      "deno.ns",
+      "webworker"
     ]
   },
   "unstable": [
--- a/jsr.json
+++ b/jsr.json
@@ -69,6 +69,7 @@
     "./vercel": "./src/adapter/vercel/index.ts",
     "./netlify": "./src/adapter/netlify/index.ts",
     "./lambda-edge": "./src/adapter/lambda-edge/index.ts",
+    "./service-worker": "./src/adapter/service-worker/index.ts",
     "./testing": "./src/helper/testing/index.ts",
     "./dev": "./src/helper/dev/index.ts",
     "./ws": "./src/helper/websocket/index.ts",
--- a/jsr.json
+++ b/jsr.json
@@ -5,8 +5,7 @@
     "lib": [
       "dom",
       "dom.iterable",
-      "deno.ns",
-      "webworker"
+      "deno.ns"
     ]
   },
   "unstable": [
--- a/package.json
+++ b/package.json
@@ -333,6 +333,11 @@
       "import": "./dist/adapter/lambda-edge/index.js",
       "require": "./dist/cjs/adapter/lambda-edge/index.js"
     },
+    "./service-worker": {
+      "types": "./dist/types/adapter/service-worker/index.d.ts",
+      "import": "./dist/adapter/service-worker/index.js",
+      "require": "./dist/cjs/adapter/service-worker/index.js"
+    },
     "./testing": {
       "types": "./dist/types/helper/testing/index.d.ts",
       "import": "./dist/helper/testing/index.js",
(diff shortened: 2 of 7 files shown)


yusukebe:
Sounds good! I'll try it later.

yusukebe:
Hi @nakasyou 

I've noticed that this adapter is so important! If we have this adapter, we can obsolete the `app.fire()` method in the next major release. The `app.fire()` is this:

```ts
// src/hono-base.ts
fire = (): void => {
  // @ts-ignore
  addEventListener('fetch', (event: FetchEventLike): void => {
    event.respondWith(this.dispatch(event.request, event, undefined, event.request.method))
  })
}
```

I think it's okay to delete it completely because this service worker style is now used only for Fastlyu Compute and Service Worker.

nakasyou:
@yusukebe It sounds good!
However, the adapter is for browsers. So if app returns 404, adapter fetches the original server. Currently replacing `app.fire` maybe not good.

yusukebe:
> However, the adapter is for browsers. So if app returns 404, adapter fetches the original server.

Yeah. I'll consider it.

yusukebe:
Hi @nakasyou !

I've created the demo project to try this adapter: https://github.com/yusukebe/service-worker-adapter-demo

And I'll suggest the code, though I'm not sure this is 100% good:

```diff
diff --git a/src/adapter/service-worker/handler.ts b/src/adapter/service-worker/handler.ts
index 2c85343..b8801e3 100644
--- a/src/adapter/service-worker/handler.ts
+++ b/src/adapter/service-worker/handler.ts
@@ -1,12 +1,12 @@
+/// <reference lib="webworker" />
 /**
  * Handler for Service Worker
  * @module
  */

 import type { Hono } from '../../hono'
-import type { FetchEventLike } from '../../types'

-type Handler = (evt: FetchEventLike) => void
+type Handler = (evt: FetchEvent) => void

 /**
  * Adapter for Service Worker
@@ -14,24 +14,16 @@ type Handler = (evt: FetchEventLike) => void
 export const handle = (
   app: Hono,
   opts: {
-    fetch: (req: Request) => Promise<Response>
+    fetch: typeof fetch | undefined
   } = {
     fetch: fetch,
   }
 ): Handler => {
-  return (evt) => {
-    const fetched = app.fetch(evt.request)
-    if (fetched instanceof Response && fetched.status === 404) {
-      return
+  return async (evt) => {
+    let res = await app.fetch(evt.request)
+    if (res.status === 404 && opts.fetch) {
+      res = await opts.fetch(evt.request)
     }
-    evt.respondWith(
-      (async () => {
-        const res = await fetched
-        if (res.status === 404) {
-          return await opts.fetch(evt.request)
-        }
-        return res
-      })()
-    )
+    evt.respondWith(res)
   }
 }
```

There are two points:

1. To correct types on the user side, this uses `FetchEvent` in the `WebWorker` for `evt`.
2. Simplify the logic. I think this is enough. The current code can't handle the 404 well, but the error will be removed in the demo project with this code.

Plus, with this code, you can pass through the handling 404 by assigning `undefined` to the `fetch` option. This means it can support Fastly Compute.

```ts
addEventListener(
  'fetch',
  handle(app, {
    fetch: undefined
  })
)
```

What do you think of them?

nakasyou:
Hi @yusukebe, I agree for your comment.
I'll change it.

nakasyou:
Hi @yusukebe, I have an issue at TypeScript. It is made by webworker types.
Do you know how to resolve?

yusukebe (approved):
LGTM!

yusukebe:
@nakasyou 

Awesome work! I'll merge this into the `next` now. Thanks!

PR #3074, opened 2024-07-01
feat(types): improve JSONParsed

m-shaka:
I've improved `JSONParsed` in order to follow the spec of `JSON.stringify` as much as possible.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#description

Here is major improvement:
- support tuple type
- handle undefined, symbol, and function type correctly(in array or object). For example
  - convert undefined to null in an array
  - drop keys whose value is undefined 

You can see some actual behaviour in [TS playground](https://www.typescriptlang.org/play/?#code/MYewdgziA2CmB00QHMAUByAZgVzMALgJbgBc6ANAAQBSAygPIBy8E+AToWMoZgJ6qoAlJQC8APkoBvAL6DhAenmVcAE1iZOsFQChQkGAiRp0EXgFsARjDJU6TFu07c+qWuavR4mEG1QAiCD85BSVVdU0dPSg4RBQMU0sYSgBrWF4AbRALACtYAgBdChoGZlYOLh5+SUp0t0TPb18AoPySSgBGWRDKPxk-XXBowzj0MI0wLUycvPxC2xKHcucqygBDNrGIyi7KRR6+gf0Yo3j3JIA3VehsWCncgqK7UscKl2r1yjqPLx9-QMEdntetJ+lEDLFjDg8ERwJRLtdbll7rNHgsyk5Kqh3m0hKIJDJAUpgaDBuCTqMwGpxpNVmw2KteHNivZ0a9+OlNhMVPk5LsiekwNhoNB8iSjsNjAkPOlafTGaiWS9lqhamcGr9moIed0-AKhSKxUMIRgoQRiGAZXSGUynosMS50rjxFJpNq+T09cLRYcjeSJqwtJR8CAngrnktMZJg08ccJnVigyGSrG8R1tsJCfsQUA)

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -11,7 +11,13 @@ import type {
 } from './types'
 import { HtmlEscapedCallbackPhase, resolveCallback } from './utils/html'
 import type { RedirectStatusCode, StatusCode } from './utils/http-status'
-import type { IsAny, JSONParsed, JSONValue, SimplifyDeepArray } from './utils/types'
+import type {
+  InvalidJSONValue,
+  IsAny,
+  JSONParsed,
+  JSONValue,
+  SimplifyDeepArray,
+} from './utils/types'
 
 type HeaderRecord = Record<string, string | string[]>
 
@@ -142,12 +148,18 @@ interface TextRespond {
  * @returns {JSONRespondReturn<T, U>} - The response after rendering the JSON object, typed with the provided object and status code types.
  */
 interface JSONRespond {
-  <T extends JSONValue | SimplifyDeepArray<unknown>, U extends StatusCode = StatusCode>(
+  <
+    T extends JSONValue | SimplifyDeepArray<unknown> | InvalidJSONValue,
+    U extends StatusCode = StatusCode
+  >(
     object: T,
     status?: U,
     headers?: HeaderRecord
   ): JSONRespondReturn<T, U>
-  <T extends JSONValue | SimplifyDeepArray<unknown>, U extends StatusCode = StatusCode>(
+  <
+    T extends JSONValue | SimplifyDeepArray<unknown> | InvalidJSONValue,
+    U extends StatusCode = StatusCode
+  >(
     object: T,
     init?: ResponseInit
   ): JSONRespondReturn<T, U>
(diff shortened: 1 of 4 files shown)


yusukebe:
Hi @m-shaka 

Awesome! Following the spec of `JSON.stringify` is great.

One thing we have to consider is performance. This change will introduce performance degradation, especially in RPC.

Try to create an app that has many routes with this script:

```ts
import { writeFile } from 'fs'

const count = 200

const generateRoutes = (count: number) => {
  const imports = `
import { Hono } from 'hono';
import { z } from 'zod';
import { zValidator } from '@hono/zod-validator';
`

  const schema = `
const schema = z.object({
  q: z.string()
})\n
`

  let routes = imports + schema + 'export const app = new Hono()\n'
  for (let i = 1; i <= count; i++) {
    routes += `  .get('/route${i}/:id', zValidator('query', schema), (c) => {
    return c.json({
      query: c.req.valid('query'),
      id: c.req.param('id'),
      ok: true
    })
  })\n`
  }
  return routes
}

const routes = generateRoutes(count)

writeFile('app.ts', routes, (err) => {
  if (err) throw err
  console.log(`${count} routes have been written to app.ts`)
})
```

Load this app in another script with `hc`:

```ts
import { app } from './app'

```

If with current version `4.4.10`, it can handle over 200 routes in my IDE:

CleanShot 2024-07-04 at 15 07 06@2x

But with this PR, it will throw the error:

CleanShot 2024-07-04 at 15 07 41@2x

Performance issues are often reported in this project, so we must address them. We want a good idea to support both the following specs and better performance. We may be able to ignore this matter and refactor other points (like `hc`) to ensure good performance for RPC. However, we don't want to decrease the performance as much as possible with these JSON things.

@m-shaka @NamesMT Any thoughts?

yusukebe:
Ah, the script includes a validator, so it's better to use this script:

```ts
import { writeFile } from 'fs'

const count = 200

const generateRoutes = (count: number) => {
  let routes = `import { Hono } from 'hono'
  export const app = new Hono()`
  for (let i = 1; i <= count; i++) {
    routes += `  .get('/route${i}/:id', (c) => {
    return c.json({
      ok: true
    })
  })\n`
  }
  return routes
}

const routes = generateRoutes(count)

writeFile('app.ts', routes, (err) => {
  if (err) throw err
  console.log(`${count} routes have been written to app.ts`)
})
```

m-shaka:
Oh, I'll have a look later 👀

m-shaka:
@yusukebe  While I've changed the behaviour of optional keys, it seems to work now.
Now `a?: string` gonna be converted to `a: string | undefined`, but I don't think it's a problem in practical use

yusukebe:
@m-shaka Thanks! I'll check it later.

yusukebe (approved):
LGTM!

yusukebe:
@m-shaka 

Great! There is little or no performance degradation. Looks good! I wonder if we should merge this into the current `main` or the `next` branch(for the next minor release, `v4.5`). I'll merge later. Thanks!

yusukebe:
Hi @m-shaka 

This is a `feat` but does not introduce a new future for users. So, I'll merge this now into the `main` and include it in a patch release. Thanks!

PR #3079, opened 2024-07-02
fix(stream): Fixed a problem that onAbort() is called even if request is normally closed in deno

usualoma:
#3042 fixed the problem in bun, but in deno, `onAbort()` was being called after a successful exit.

In deno, an "abort" event is fired for `req.signal` even if it exits successfully.

In this pull request, fix so that `onAbort()` is not called when stream finishes successfully.

```ts
import { Hono } from 'hono'

const app = new Hono()
app.get('/stream', (c) => {
  return stream(c, async (stream) => {
    stream.onAbort(() => {
      console.log('Aborted!')
    })
    await stream.write('Hello')
  })
})

export default app
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/runtime_tests/bun/index.test.tsx
+++ b/runtime_tests/bun/index.test.tsx
@@ -333,6 +333,14 @@ describe('streaming', () => {
       })
     })
   })
+  app.get('/streamHello', (c) => {
+    return stream(c, async (stream) => {
+      stream.onAbort(() => {
+        aborted = true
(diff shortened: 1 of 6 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@usualoma Thank you for the fix and also adding the tests! Merging now.

PR #3082, opened 2024-07-02
feat(middleware): introduce Request ID middleware

ryuapp:
Request ID middleware generates a unique id for a request.
The unique request id can be used to trace a request end-to-end.
```ts
const app = new Hono()

app.use(requestId())
app.get('/', (c) => {
    console.log(c.get('requestId'))
    return c.text('Hello World!')
})
```

It is also easily customizable with several options.

```ts
type RequesIdOptions = {
  // The maximum length of request id.
  // The default value is 255. 
  limitLength?: number
  // The header name used in request id.
  // The default value is 'X-Request-Id'.
  headerName?: string
  // The request id generation function.
  // The default value is 'crypto.randomUUID()'.
  generator?: (c: Context) => string
}
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -213,6 +213,11 @@
       "import": "./dist/middleware/pretty-json/index.js",
       "require": "./dist/cjs/middleware/pretty-json/index.js"
     },
+    "./request-id": {
+      "types": "./dist/types/middleware/request-id/index.d.ts",
+      "import": "./dist/middleware/request-id/index.js",
+      "require": "./dist/cjs/middleware/request-id/index.js"
+    },
     "./secure-headers": {
       "types": "./dist/types/middleware/secure-headers/index.d.ts",
       "import": "./dist/middleware/secure-headers/index.js",
--- a/package.json
+++ b/package.json
@@ -460,6 +460,9 @@
       "pretty-json": [
         "./dist/types/middleware/pretty-json"
       ],
+      "request-id": [
+        "./dist/types/middleware/request-id"
+      ],
       "streaming": [
         "./dist/types/helper/streaming"
       ],
--- a/src/middleware/request-id/request-id.ts
+++ b/src/middleware/request-id/request-id.ts
@@ -10,7 +10,7 @@ export type RequestIdVariables = {
   requestId: string
 }
 
-export type RequesIdOptions = {
+export type RequestIdOptions = {
   limitLength?: number
   headerName?: string
   generator?: (c: Context) => string
@@ -43,7 +43,7 @@ export const requestId = ({
   limitLength = 255,
   headerName = 'X-Request-Id',
   generator = () => crypto.randomUUID(),
-}: RequesIdOptions = {}): MiddlewareHandler => {
+}: RequestIdOptions = {}): MiddlewareHandler => {
   return async function requestId(c, next) {
     // If `headerName` is empty string, req.header will return the object
     let reqId = headerName ? c.req.header(headerName) : undefined
--- a/src/middleware/request-id/request-id.ts
+++ b/src/middleware/request-id/request-id.ts
@@ -1,0 +1,60 @@
+/**
+ * @module
+ * Request ID Middleware for Hono.
+ */
+
+import type { MiddlewareHandler } from '../../types'
+
+export type RequestIDVariables = {
+  requestID: string
+}
+
+export type RequesIDOptions = {
+  limitLength?: number
+  headerName?: string
+  generator?: () => string
+}
+
+/**
+ * Request ID Middleware for Hono.
+ *
+ * @param {object} options - Options for Request ID middleware.
+ * @param {number} [options.limitLength=255] - The maximum length of request id.
+ * If positive truncates the request id at the specified length.
+ * @param {string} [options.headerName=X-Request-Id] - The header name used in request id.
+ * @param {generator} [options.generator=crypto.randomUUID()] - The request id generation function.
+ *
+ * @returns {MiddlewareHandler} The middleware handler function.
+ *
+ * @example
+ * ```ts
+ * type Variables = RequestIDVariables
+ * const app = new Hono<{Variables: Variables}>()
+ *
+ * app.use(requestID())
+ * app.get('/', (c) => {
+ *   console.log(c.get('requestID')) // Debug
+ *   return c.text('Hello World!')
+ * })
+ * ```
+ */
+export const requestID = (options?: RequesIDOptions): MiddlewareHandler => {
+  const limitLength = options?.limitLength ?? 255
+  const headerName = options?.headerName ?? 'X-Request-Id'
+
+  return async function requestID(c, next) {
+    let requestId = c.req.header(headerName)
+    if (requestId) {
+      requestId = requestId.replace(/[^\w\-]/g, '')
+      requestId = limitLength > 0 ? requestId.substring(0, limitLength) : requestId
+    } else {
+      requestId = options?.generator?.() ?? crypto.randomUUID()
+    }
+
+    c.set('requestID', requestId)
+    if (headerName) {
+      c.header(headerName, requestId)
+    }
+    await next()
+  }
+}
--- a/src/middleware/request-id/request-id.ts
+++ b/src/middleware/request-id/request-id.ts
@@ -3,6 +3,7 @@
  * Request ID Middleware for Hono.
  */
 
+import type { Context } from '../../context'
 import type { MiddlewareHandler } from '../../types'
 
 export type RequestIdVariables = {
@@ -12,7 +13,7 @@ export type RequestIdVariables = {
 export type RequesIdOptions = {
   limitLength?: number
   headerName?: string
-  generator?: () => string
+  generator?: (c: Context) => string
 }
 
 /**
@@ -22,7 +23,7 @@ export type RequesIdOptions = {
  * @param {number} [options.limitLength=255] - The maximum length of request id.
  * If positive truncates the request id at the specified length.
  * @param {string} [options.headerName=X-Request-Id] - The header name used in request id.
- * @param {generator} [options.generator=crypto.randomUUID()] - The request id generation function.
+ * @param {generator} [options.generator=() => crypto.randomUUID()] - The request id generation function.
  *
  * @returns {MiddlewareHandler} The middleware handler function.
  *
@@ -38,10 +39,11 @@ export type RequesIdOptions = {
  * })
  * ```
  */
-export const requestId = (options?: RequesIdOptions): MiddlewareHandler => {
-  const limitLength = options?.limitLength ?? 255
-  const headerName = options?.headerName ?? 'X-Request-Id'
-
+export const requestId = ({
+  limitLength = 255,
+  headerName = 'X-Request-Id',
+  generator = () => crypto.randomUUID(),
+}: RequesIdOptions = {}): MiddlewareHandler => {
   return async function requestId(c, next) {
     // If `headerName` is empty string, req.header will return the object
     let reqId = headerName ? c.req.header(headerName) : undefined
@@ -50,7 +52,7 @@ export const requestId = (options?: RequesIdOptions): MiddlewareHandler => {
       reqId = reqId.replace(/[^\w\-]/g, '')
       reqId = limitLength > 0 ? reqId.substring(0, limitLength) : reqId
     } else {
-      reqId = options?.generator?.() ?? crypto.randomUUID()
+      reqId = generator(c)
     }
 
     c.set('requestId', reqId)
--- a/src/middleware/request-id/request-id.ts
+++ b/src/middleware/request-id/request-id.ts
@@ -43,7 +43,9 @@ export const requestID = (options?: RequesIDOptions): MiddlewareHandler => {
   const headerName = options?.headerName ?? 'X-Request-Id'
 
   return async function requestID(c, next) {
-    let requestId = c.req.header(headerName)
+    // If `headerName` is empty string, req.header will return the object
+    let requestId = headerName ? c.req.header(headerName) : undefined
+
     if (requestId) {
       requestId = requestId.replace(/[^\w\-]/g, '')
       requestId = limitLength > 0 ? requestId.substring(0, limitLength) : requestId
--- a/src/middleware/request-id/request-id.ts
+++ b/src/middleware/request-id/request-id.ts
@@ -21,7 +21,6 @@ export type RequestIdOptions = {
  *
  * @param {object} options - Options for Request ID middleware.
  * @param {number} [options.limitLength=255] - The maximum length of request id.
- * If positive truncates the request id at the specified length.
  * @param {string} [options.headerName=X-Request-Id] - The header name used in request id.
  * @param {generator} [options.generator=() => crypto.randomUUID()] - The request id generation function.
  *
@@ -47,11 +46,7 @@ export const requestId = ({
   return async function requestId(c, next) {
     // If `headerName` is empty string, req.header will return the object
     let reqId = headerName ? c.req.header(headerName) : undefined
-
-    if (reqId) {
-      reqId = reqId.replace(/[^\w\-]/g, '')
-      reqId = limitLength > 0 ? reqId.substring(0, limitLength) : reqId
-    } else {
+    if (!reqId || reqId.length > limitLength || /[^\w\-]/.test(reqId)) {
       reqId = generator(c)
     }
 
--- a/src/middleware/request-id/request-id.ts
+++ b/src/middleware/request-id/request-id.ts
@@ -5,11 +5,11 @@
 
 import type { MiddlewareHandler } from '../../types'
 
-export type RequestIDVariables = {
-  requestID: string
+export type RequestIdVariables = {
+  requestId: string
 }
 
-export type RequesIDOptions = {
+export type RequesIdOptions = {
   limitLength?: number
   headerName?: string
   generator?: () => string
@@ -28,34 +28,34 @@ export type RequesIDOptions = {
  *
  * @example
  * ```ts
- * type Variables = RequestIDVariables
+ * type Variables = RequestIdVariables
  * const app = new Hono<{Variables: Variables}>()
  *
- * app.use(requestID())
+ * app.use(requestId())
  * app.get('/', (c) => {
- *   console.log(c.get('requestID')) // Debug
+ *   console.log(c.get('requestId')) // Debug
  *   return c.text('Hello World!')
  * })
  * ```
  */
-export const requestID = (options?: RequesIDOptions): MiddlewareHandler => {
+export const requestId = (options?: RequesIdOptions): MiddlewareHandler => {
   const limitLength = options?.limitLength ?? 255
   const headerName = options?.headerName ?? 'X-Request-Id'
 
-  return async function requestID(c, next) {
+  return async function requestId(c, next) {
     // If `headerName` is empty string, req.header will return the object
-    let requestId = headerName ? c.req.header(headerName) : undefined
+    let reqId = headerName ? c.req.header(headerName) : undefined
 
-    if (requestId) {
-      requestId = requestId.replace(/[^\w\-]/g, '')
-      requestId = limitLength > 0 ? requestId.substring(0, limitLength) : requestId
+    if (reqId) {
+      reqId = reqId.replace(/[^\w\-]/g, '')
+      reqId = limitLength > 0 ? reqId.substring(0, limitLength) : reqId
     } else {
-      requestId = options?.generator?.() ?? crypto.randomUUID()
+      reqId = options?.generator?.() ?? crypto.randomUUID()
     }
 
-    c.set('requestID', requestId)
+    c.set('requestId', reqId)
     if (headerName) {
-      c.header(headerName, requestId)
+      c.header(headerName, reqId)
     }
     await next()
   }
--- a/jsr.json
+++ b/jsr.json
@@ -47,6 +47,7 @@
     "./method-override": "./src/middleware/method-override/index.ts",
     "./powered-by": "./src/middleware/powered-by/index.ts",
     "./pretty-json": "./src/middleware/pretty-json/index.ts",
+    "./request-id": "./src/middleware/request-id/request-id.ts",
     "./secure-headers": "./src/middleware/secure-headers/secure-headers.ts",
     "./ssg": "./src/helper/ssg/index.ts",
     "./streaming": "./src/helper/streaming/index.ts",
--- a/src/middleware/request-id/index.test.ts
+++ b/src/middleware/request-id/index.test.ts
@@ -1,12 +1,12 @@
 import { Hono } from '../../hono'
-import { requestID } from '.'
+import { requestId } from '.'
 
 const regexUUIDv4 = /([0-9a-f]{8})-([0-9a-f]{4})-([0-9a-f]{4})-([0-9a-f]{4})-([0-9a-f]{12})/
 
 describe('Request ID Middleware', () => {
   const app = new Hono()
-  app.use('*', requestID())
-  app.get('/requestId', (c) => c.text(c.get('requestID') ?? 'No Request ID'))
+  app.use('*', requestId())
+  app.get('/requestId', (c) => c.text(c.get('requestId') ?? 'No Request ID'))
 
   it('Should return random request id', async () => {
     const res = await app.request('http://localhost/requestId')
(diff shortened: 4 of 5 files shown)


yusukebe:
Hi @ryuapp 

I love this middleware since it's simple! I think this has no problem and is ready to merge.

@usualoma @nakasyou @MathurAditya724 @fzn0x and others. Do you have any thoughts on this?

src/middleware/request-id/request-id.ts
```diff
@@ -0,0 +1,60 @@
...
+import type { MiddlewareHandler } from '../../types'
+
+export type RequestIDVariables = {
+  requestID: string
+}
+
+export type RequesIDOptions = {
+  limitLength?: number
+  headerName?: string
+  generator?: () => string
```
yusukebe, line 15: What about passing the context to the `generator`?

```ts
generator?: (c:Context) => string
```

ryuapp: It looks nice.
This will allow users to handle request ID as they like.

src/middleware/request-id/request-id.ts
```diff
@@ -0,0 +1,62 @@
...
+ * })
+ * ```
+ */
+export const requestID = (options?: RequesIDOptions): MiddlewareHandler => {
+  const limitLength = options?.limitLength ?? 255
+  const headerName = options?.headerName ?? 'X-Request-Id'
+
+  return async function requestID(c, next) {
+    // If `headerName` is empty string, req.header will return the object
+    let requestId = headerName ? c.req.header(headerName) : undefined
```
fzn0x, line 47: is this really needed?

since we already provide a check for this

```ts
const headerName = options?.headerName ?? 'X-Request-Id'
```

src/middleware/request-id/request-id.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    if (requestId) {
+      requestId = requestId.replace(/[^\w\-]/g, '')
+      requestId = limitLength > 0 ? requestId.substring(0, limitLength) : requestId
+    } else {
+      requestId = options?.generator?.() ?? crypto.randomUUID()
```
fzn0x, line 53: Note that, `crypto.randomUUID()` may not be a stable method to use if Hono plan to focus on [supporting Web Browsers](https://x.com/honojs/status/1714947374886637883)

yusukebe: Ah, I also worried about `crypto.randomUUID()`. Either way, I think `crypto.randomUUID()` works on major browsers: https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID

Or am I missing something?

src/middleware/request-id/request-id.ts
ryuapp: > is this really needed?
> 
> since we already provide a check for this
> 
> ```ts
> const headerName = options?.headerName ?? 'X-Request-Id'
> ```


I haven't documented this yet, but it is for internal use only. For example, logging.
Specifying empty string means that the user expects no request ID to be present in  request/response.
And we need to check for empty string.

```ts
console.log("" ?? "string") // empty
console.log("" || "string") // string
console.log("" ? "A" : "B") // B
```

ryuapp:
I can't decide between `requestId` and `requestID` in the variable and middleware name.
The former follows the naming convention and looks better, but I thought 'L' and 'I' were confusing, so I chose the latter.

However, I realized that the above problem is a font problem and does not affect anything other than the document, so I am thinking of fixing it, but does anyone have any opinions?

src/middleware/request-id/request-id.ts
ryuapp: I think it's one of web-standards, so I don't think there's a problem.
One thing to note is that this does not work in Node v18. However, there are `generator` option for avoiding that.

fzn0x: `crypto.randomUUID()` only works in HTTPS @yusukebe @ryuapp I'm afraid it something we should thinking of

fzn0x:
I think `requestId` is better to use when typing on keyboard, and it already use since then and accepted by the majority, so maybe just go with `requestId`?

MathurAditya724:
Interesting middleware! Can already think about multiple use cases for this. And I am voting for `requestId`.

src/middleware/request-id/request-id.ts
fzn0x: Also a random stackoverflow post I've found talking about it: https://stackoverflow.com/a/77981051/16258067

src/middleware/request-id/request-id.ts
fzn0x: Ah I see, good approach!

fzn0x: You can close this comment

src/middleware/request-id/request-id.ts
MathurAditya724: Yup, `crypto.randomUUID()` doesn't work on this domain http://example.com/. But I believe the majority of the apps will be deployed on `HTTPS` so this shouldn't be a problem.

But if you want, We can do an `if else` statement to check if `randomUUID` is present in `crypto` and can either show a warning or do a simple implementation like `() => "add-a-generator"`, which doesn't require `crypto`

ryuapp: > crypto.randomUUID() only works in HTTPS

I think this problem can be ignored.
It can be assumed that only Hono playgrounds connected via HTTP are affected.
That's a very rare case.
As mentioned above, if `crypto.randomUUID()` is not possible, we can use `generator` option.

All we have to do is write the cautionary point in the docs.

yusukebe: @fzn0x @MathurAditya724 Ahh, thanks!

Does this mean that if I run an app with this Request ID middleware on localhost as a Service Worker, it does not work well? Hmm, we don't want to use anything that isn't common as much as possible.

yusukebe:
I prefer `requestId` for the same reason as @fzn0x 👍

src/middleware/request-id/request-id.ts
yusukebe: However, creating our own UUID generation logic in the `hono` package might be wasteful, so maybe the user can just pass it as an option.

ryuapp:
Thank you for all the opinions.
Many people agree with `requestId` and the reason is obvious, I will change it to `requestId`.

src/middleware/request-id/request-id.ts
ryuapp: > Does this mean that if I run an app with this Request ID middleware on localhost as a Service Worker, it does not work well?

It works well on localhost. Only 'HTTP' connection is a problem. Check it there:
https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts#when_is_a_context_considered_secure

> Locally-delivered resources such as those with http://127.0.0.1 URLs, http://localhost and http://*.localhost URLs (e.g. http://dev.whatever.localhost/), and file:// URLs are also considered to have been delivered securely.

Although it is not necessary to use UUID, I did it this way because it is minimum codes and many frameworks use UUIDv4 for request ID middleware.

yusukebe: > It works well on localhost. Only 'HTTP' connection is a problem. Check it there:

Thanks, I understood well. So I think it's using `crypto.randomUUID()` is good! Shall we go with this?

fzn0x: @yusukebe Lets ship it and wait how users will react! 🚀

fzn0x (approved):
Looks good to me

package.json
yusukebe, line 1: @ryuapp 

Can you also update the `typesVersions` field?

```js
    "typesVersions": {
    "*": {
      "types": [
        "./dist/types/types"
      ],
// ...
      "request-id": [
// ...
      ],
// ...
```

ryuapp: I missed it and fixed it. Thank you.

ryuapp:
The tests is failing on `next` regardless of this branch, I submit a PR to fix it. #3115

usualoma (commented):
Thanks for the great pull request!

### Performance improvement proposals

If the request ID can be obtained from the request header, it can be assumed that it can be used as is in most cases. `String.prototype.replace()` is heavy, so it is better not to do anything if it passes the check by the regular expression prepared in advance.

The following is an example implementation.

```diff
diff --git a/src/middleware/request-id/request-id.ts b/src/middleware/request-id/request-id.ts
index 0c11cb46..ea66b48b 100644
--- a/src/middleware/request-id/request-id.ts
+++ b/src/middleware/request-id/request-id.ts
@@ -44,13 +44,19 @@ export const requestId = ({
   headerName = 'X-Request-Id',
   generator = () => crypto.randomUUID(),
 }: RequesIdOptions = {}): MiddlewareHandler => {
+  const isValidIncomingHeaderValueRe = new RegExp(
+    `^[\\w\\-]${limitLength ? `{,${limitLength}}` : '+'}$`
+  )
+
   return async function requestId(c, next) {
     // If `headerName` is empty string, req.header will return the object
     let reqId = headerName ? c.req.header(headerName) : undefined
 
     if (reqId) {
-      reqId = reqId.replace(/[^\w\-]/g, '')
-      reqId = limitLength > 0 ? reqId.substring(0, limitLength) : reqId
+      if (!isValidIncomingHeaderValueRe.test(reqId)) {
+        reqId = reqId.replace(/[^\w\-]/g, '')
+        reqId = limitLength > 0 ? reqId.substring(0, limitLength) : reqId
+      }
     } else {
       reqId = generator(c)
     }
```

src/middleware/request-id/request-id.ts
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+import type { Context } from '../../context'
+import type { MiddlewareHandler } from '../../types'
+
+export type RequestIdVariables = {
+  requestId: string
+}
+
+export type RequesIdOptions = {
```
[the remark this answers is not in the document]
ryuapp: You are right. Thanks.

ryuapp:
Hello @usualoma .
Thank you for your proposal.
When I ran a benchmark with Deno, it seems that the cost of checking with regular expressions is greater.
```sh
benchmark               time (avg)        iter/s             (min … max)       p75       p99      p995
------------------------------------------------------------------------ -----------------------------
no-check: valid         68.51 ns/iter  14,595,900.4  (60.83 ns … 240.78 ns) 66.26 ns 136.44 ns 158.73 ns
pre check: valid        96.83 ns/iter  10,327,663.8     (85.27 ns … 154 ns) 92.97 ns 148.24 ns 149.35 ns
no-check: invalid       93.97 ns/iter  10,641,957.4  (90.13 ns … 212.15 ns) 92.73 ns 122.85 ns 146.21 ns
pre check: invalid     246.53 ns/iter   4,056,358.6 (235.76 ns … 670.25 ns) 240.52 ns 547.67 ns 621.05 ns
```
Source codes
<details>

```ts
const validId = crypto.randomUUID();
const invalidId = validId + "!";
const limitLength = 255;
const isValidIncomingHeaderValueRe = new RegExp(
  `^[\\w\\-]${limitLength ? `{0,${limitLength}}` : "+"}$`,
);

let id;
console.log(isValidIncomingHeaderValueRe);
console.log(!isValidIncomingHeaderValueRe.test(validId));
console.log(!isValidIncomingHeaderValueRe.test(invalidId));
Deno.bench({
  name: "no-check: valid",
  fn() {
    id = validId.replace(/[^\w\-]/g, "");
    id = limitLength > 0 ? id.substring(0, limitLength) : id;
  },
});

Deno.bench({
  name: "pre check: valid",
  fn() {
    if (!isValidIncomingHeaderValueRe.test(validId)) {
      id = validId.replace(/[^\w\-]/g, "");
      id = limitLength > 0 ? id.substring(0, limitLength) : id;
    }
  },
});
Deno.bench({
  name: "no-check: invalid",
  fn() {
    id = invalidId.replace(/[^\w\-]/g, "");
    id = limitLength > 0 ? id.substring(0, limitLength) : id;
  },
});

Deno.bench({
  name: "pre check: invalid",
  fn() {
    if (!isValidIncomingHeaderValueRe.test(invalidId)) {
      id = invalidId.replace(/[^\w\-]/g, "");
      id = limitLength > 0 ? id.substring(0, limitLength) : id;
    }
  },
});

```
</details>

However, as you say, `String.prototype.replace()` is heavy. The longer the requestid, the higher the load.
This validation is added to respect Ruby on Rails, but it may not be suitable for JavaScript. I also checked other frameworks (primarily written in GO) and they didn't have this kind of functionality.

I'm thinking of removing this validation, leaving only lengthLimit. What do you think about this?

usualoma:
@ryuapp 
Oooh, sorry, you are right, my version was slow. First of all, I apologise for taking up your time.
 
I also benchmarked and confirmed the following.

Checking with a regular expression like `const isValidCharsOnlyRe = /^[\w\-]+$/;`, "node" and "deno" might perform a bit better on normal systems.  In "bun", my approach did not perform well in any way.

<details>

<summary>scripts</summary>

Both benchmarks are the same.

##### deno

```ts
const validId = crypto.randomUUID();
const invalidId = validId + "!";
const limitLength = 255;
const isValidIncomingHeaderValueRe = new RegExp(
  `^[\\w\\-]${limitLength ? `{0,${limitLength}}` : "+"}$`,
);
const isValidCharsOnlyRe = /^[\w\-]+$/;


let id;
console.log(isValidIncomingHeaderValueRe);
console.log(!isValidIncomingHeaderValueRe.test(validId));
console.log(!isValidIncomingHeaderValueRe.test(invalidId));
console.log(!isValidCharsOnlyRe.test(validId));
console.log(!isValidCharsOnlyRe.test(invalidId));
Deno.bench({
  name: "no-check: valid",
  fn() {
    id = validId.replace(/[^\w\-]/g, "");
    id = limitLength > 0 ? id.substring(0, limitLength) : id;
  },
});

Deno.bench({
  name: "pre check: valid",
  fn() {
    if (!isValidIncomingHeaderValueRe.test(validId)) {
      id = validId.replace(/[^\w\-]/g, "");
      id = limitLength > 0 ? id.substring(0, limitLength) : id;
    }
  },
});

Deno.bench({
  name: "pre check w/ isValidCharsOnlyRe: valid",
  fn() {
    if (!isValidCharsOnlyRe.test(validId) || (limitLength > 0 && validId.length > limitLength)) {
      id = validId.replace(/[^\w\-]/g, "");
      id = limitLength > 0 ? id.substring(0, limitLength) : id;
    }
  },
});

Deno.bench({
  name: "no-check: invalid",
  fn() {
    id = invalidId.replace(/[^\w\-]/g, "");
    id = limitLength > 0 ? id.substring(0, limitLength) : id;
  },
});

Deno.bench({
  name: "pre check: invalid",
  fn() {
    if (!isValidIncomingHeaderValueRe.test(invalidId)) {
      id = invalidId.replace(/[^\w\-]/g, "");
      id = limitLength > 0 ? id.substring(0, limitLength) : id;
    }
  },
});
```

##### node / bun

```ts

const validId = crypto.randomUUID()
const invalidId = validId + '!'
const limitLength = 255
const isValidIncomingHeaderValueRe = new RegExp(
  `^[\\w\\-]${limitLength ? `{0,${limitLength}}` : '+'}$`
)
const isValidCharsOnlyRe = /^[\w\-]+$/

let id
console.log(isValidIncomingHeaderValueRe)
console.log(!isValidIncomingHeaderValueRe.test(validId))
console.log(!isValidIncomingHeaderValueRe.test(invalidId))

group('uuid validation', () => {
  bench('no-check: valid', () => {
    id = validId.replace(/[^\w\-]/g, '')
    id = limitLength > 0 ? id.substring(0, limitLength) : id
  })
  bench('pre check: valid', () => {
    if (!isValidIncomingHeaderValueRe.test(validId)) {
      id = validId.replace(/[^\w\-]/g, '')
      id = limitLength > 0 ? id.substring(0, limitLength) : id
    }
  })
  bench('pre check w/ isValidCharsOnlyRe: valid', () => {
    if (!isValidCharsOnlyRe.test(validId) || (limitLength > 0 && validId.length > limitLength)) {
      id = validId.replace(/[^\w\-]/g, '')
      id = limitLength > 0 ? id.substring(0, limitLength) : id
    }
  })
  bench('no-check: invalid', () => {
    id = invalidId.replace(/[^\w\-]/g, '')
    id = limitLength > 0 ? id.substring(0, limitLength) : id
  })
  bench('pre check: invalid', () => {
    if (!isValidIncomingHeaderValueRe.test(invalidId)) {
      id = invalidId.replace(/[^\w\-]/g, '')
      id = limitLength > 0 ? id.substring(0, limitLength) : id
    }
  })
})

run()
```

</details>

#### deno

```
benchmark                                   time (avg)        iter/s             (min … max)       p75       p99      p995
-------------------------------------------------------------------------------------------- -----------------------------
no-check: valid                             50.84 ns/iter  19,667,727.7       (48.6 ns … 74 ns) 51.19 ns 59.42 ns 66.02 ns
pre check: valid                            61.92 ns/iter  16,150,143.1   (59.85 ns … 74.46 ns) 62.24 ns 65.65 ns 66.93 ns
pre check w/ isValidCharsOnlyRe: valid      35.25 ns/iter  28,369,004.6   (33.86 ns … 40.23 ns) 35.56 ns 37.77 ns 38.45 ns
no-check: invalid                           76.43 ns/iter  13,083,104.8    (72.61 ns … 97.5 ns) 77.72 ns 87.95 ns 90.08 ns
pre check: invalid                         181.03 ns/iter   5,524,007.4 (174.89 ns … 193.31 ns) 183.09 ns 191.79 ns 192.09 ns
```

#### node

```
• uuid validation
------------------------------------------------------------------------------ -----------------------------
no-check: valid                          46.04 ns/iter     (43.56 ns … 131 ns)  45.51 ns   68.5 ns  80.02 ns
pre check: valid                         59.99 ns/iter   (56.76 ns … 81.89 ns)  60.77 ns  67.26 ns  73.71 ns
pre check w/ isValidCharsOnlyRe: valid   33.85 ns/iter   (32.67 ns … 66.59 ns)  34.02 ns  39.79 ns  46.81 ns
no-check: invalid                        67.87 ns/iter     (64.53 ns … 139 ns)  67.55 ns  92.77 ns    115 ns
pre check: invalid                         169 ns/iter       (159 ns … 209 ns)    169 ns    195 ns    207 ns
```

#### bun

```
• uuid validation
------------------------------------------------------------------------------ -----------------------------
no-check: valid                            1.8 ns/iter      (1.63 ns … 219 ns)   1.75 ns   2.28 ns   9.09 ns
pre check: valid                          2.29 ns/iter    (2.18 ns … 21.18 ns)   2.26 ns   2.83 ns   6.19 ns
pre check w/ isValidCharsOnlyRe: valid     2.4 ns/iter    (2.24 ns … 38.45 ns)   2.38 ns   3.01 ns   6.73 ns
no-check: invalid                          1.8 ns/iter    (1.69 ns … 99.83 ns)   1.79 ns   2.22 ns   5.33 ns
pre check: invalid                        2.02 ns/iter    (1.89 ns … 23.52 ns)   2.01 ns   2.44 ns   6.39 ns
```


I'm thinking about this now.

> I'm thinking of removing this validation, leaving only the lengthLimit What do you think about this?

usualoma:
Ah, using `const hasInvalidCharRe = /[^\w\-]/` gave the best performance for valid id, whether "node", "deno" or "bun".

<details>

<summary>script</summary>

```ts

const validId = crypto.randomUUID()
const invalidId = validId + '!'
const limitLength = 255
const isValidIncomingHeaderValueRe = new RegExp(
  `^[\\w\\-]${limitLength ? `{0,${limitLength}}` : '+'}$`
)
const isValidCharsOnlyRe = /^[\w\-]+$/
const hasInvalidCharRe = /[^\w\-]/

let id
console.log(isValidIncomingHeaderValueRe)
console.log(!isValidIncomingHeaderValueRe.test(validId))
console.log(!isValidIncomingHeaderValueRe.test(invalidId))

group('uuid validation', () => {
  bench('no-check: valid', () => {
    id = validId.replace(/[^\w\-]/g, '')
    id = limitLength > 0 ? id.substring(0, limitLength) : id
  })
  bench('pre check: valid', () => {
    if (!isValidIncomingHeaderValueRe.test(validId)) {
      id = validId.replace(/[^\w\-]/g, '')
      id = limitLength > 0 ? id.substring(0, limitLength) : id
    }
  })
  bench('pre check w/ isValidCharsOnlyRe: valid', () => {
    if (!isValidCharsOnlyRe.test(validId) || (limitLength > 0 && validId.length > limitLength)) {
      id = validId.replace(/[^\w\-]/g, '')
      id = limitLength > 0 ? id.substring(0, limitLength) : id
    }
  })
  bench('pre check w/ hasInvalidCharRe: valid', () => {
    if (hasInvalidCharRe.test(validId) || (limitLength > 0 && validId.length > limitLength)) {
      id = validId.replace(/[^\w\-]/g, '')
      id = limitLength > 0 ? id.substring(0, limitLength) : id
    }
  })
  bench('no-check: invalid', () => {
    id = invalidId.replace(/[^\w\-]/g, '')
    id = limitLength > 0 ? id.substring(0, limitLength) : id
  })
  bench('pre check: invalid', () => {
    if (!isValidIncomingHeaderValueRe.test(invalidId)) {
      id = invalidId.replace(/[^\w\-]/g, '')
      id = limitLength > 0 ? id.substring(0, limitLength) : id
    }
  })
})

run()
```

</details>

#### deno

```
• uuid validation
------------------------------------------------------------------------------ -----------------------------
no-check: valid                          47.77 ns/iter         (0 ps … 977 ns)      0 ps    977 ns    977 ns !
pre check: valid                         59.64 ns/iter         (0 ps … 977 ns)      0 ps    977 ns    977 ns !
pre check w/ isValidCharsOnlyRe: valid    36.8 ns/iter         (0 ps … 977 ns)      0 ps    977 ns    977 ns !
pre check w/ hasInvalidCharRe: valid     35.77 ns/iter         (0 ps … 977 ns)      0 ps    977 ns    977 ns !
no-check: invalid                        72.89 ns/iter         (0 ps … 977 ns)      0 ps    977 ns    977 ns !
pre check: invalid                         174 ns/iter         (0 ps … 977 ns)      0 ps    977 ns    977 ns !
```

#### node

```
• uuid validation
------------------------------------------------------------------------------ -----------------------------
no-check: valid                          46.79 ns/iter     (42.56 ns … 139 ns)  45.65 ns  67.34 ns  79.71 ns
pre check: valid                         61.04 ns/iter     (56.62 ns … 274 ns)  60.91 ns  81.44 ns    227 ns
pre check w/ isValidCharsOnlyRe: valid   33.54 ns/iter   (31.78 ns … 61.83 ns)  33.26 ns  38.43 ns  43.01 ns
pre check w/ hasInvalidCharRe: valid     32.59 ns/iter   (30.66 ns … 57.82 ns)  33.06 ns  37.76 ns  40.98 ns
no-check: invalid                        67.71 ns/iter     (62.32 ns … 123 ns)  67.59 ns  92.26 ns    108 ns
pre check: invalid                         169 ns/iter       (162 ns … 243 ns)    170 ns    201 ns    241 ns
```

#### bun

```
• uuid validation
------------------------------------------------------------------------------ -----------------------------
no-check: valid                           1.76 ns/iter      (1.59 ns … 179 ns)   1.69 ns   3.34 ns  10.23 ns
pre check: valid                           2.2 ns/iter    (2.05 ns … 18.33 ns)   2.16 ns   2.85 ns    6.9 ns
pre check w/ isValidCharsOnlyRe: valid    2.37 ns/iter    (2.24 ns … 33.63 ns)   2.34 ns   3.07 ns   7.02 ns
pre check w/ hasInvalidCharRe: valid      1.55 ns/iter    (1.44 ns … 25.37 ns)   1.55 ns   1.95 ns   5.72 ns
no-check: invalid                         1.74 ns/iter     (1.63 ns … 22.5 ns)   1.73 ns   2.18 ns   6.37 ns
pre check: invalid                        1.96 ns/iter    (1.83 ns … 26.43 ns)   1.93 ns    2.4 ns   6.73 ns
```


However...

ryuapp:
@usualoma  Thank you for investigating this.
I agree with your opinion.
Of all the ideas we can think of at the moment, I think "re-generating" makes the most sense and is a good idea.
I was concerned that Deno's benchmarks were slow, but it seems like `crypto.ramdomUUID()` on Deno is just slow.
We can ignore this issue.

ok. I adopt the proposal. Thanks so much!

usualoma:
@ryuapp Thank you!

yusukebe (approved):
LGTM!

yusukebe:
@usualoma Thank you for reviewing, @ryuapp Thank you for such great middleware.

Now, I'll merge this into the next!

PR #3090, opened 2024-07-04
test(hono): not found handled by sub

douglasduteil:
Hi

I don't know if it's a typo but I expect a sub not found handler to take priority over the parent not found.

### Case : 

```ts
const app = new Hono()
const sub = new Hono()
app.notFound((c) => {
  return c.text('404 Not Found by app', 404)
})
sub.notFound((c) => {
  return c.text('404 Not Found by sub', 404)
})
app.route('/sub', sub)
```

### DIFF
--- a/src/hono.test.ts
+++ b/src/hono.test.ts
@@ -1958,7 +1958,7 @@ describe('Hono with `app.route`', () => {
       const res = await app.request('https://example.com/sub/implicit-404')
       expect(res.status).toBe(404)
       expect(res.headers.get('explicit')).toBe(null)
-      expect(await res.text()).toBe('404 Not Found by app')
+      expect(await res.text()).toBe('404 Not Found by sub')
     })
   })
 })


yusukebe:
Hi @douglasduteil 

It's not a typo or mistake. It's intended. The `app.route('/sub', sub)` is just adding the sub app's routes to the main app, not passing the not found handler.

If you want to handle a 404, you can create a fallback endpoint:

```ts
sub.get('/ok', (c) => {
  return c.text('ok')
})

// ...

sub.all('*', (c) => {
  return c.text('404 Not Found by sub', 404)
})
```

EdamAme-x:
I understand this spec, but I believe the notFoundHandler should take over as wel　　l.
I think this is more elegant than creating endpoints for all paths.　　
I believe this could be accomplished by creating a route in `.route` that passes `app.all("*", (c) => c.notFound())` with a notFoundHandler.

EdamAme-x:
image
I think that this is a PoC that does not consider already registered routes, it does not pass the other tests, but the features mentioned in this PR can be implemented with my proposed approach.

EdamAme-x:
The implementation is simple, since all you have to do is search for the presence of `/*` in the already registered routes. If this proposal is accepted I can send a PR to this PR branch.

EdamAme-x:
I think there are two points to discuss on this. "Should we make notFoundHandler public?" and "type inference"

Issue #3091, opened 2024-07-04
Add the triage label


mvares:
### What is the feature you are proposing?

We can create a label to issue bugs. The best option is `triage`.

PR #3092, opened 2024-07-04
chore: add the triage label

mvares:
- Closes: #3091 

Changes
--- 
Firstly, it's need to create the label.

### DIFF
--- a/.github/ISSUE_TEMPLATE/1-bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/1-bug-report.yml
@@ -1,6 +1,6 @@
 name: 🐛 Bug Report
 description: Report an issue that should be fixed
-labels: [bug, triage]
+labels: [triage]
 body:
   - type: markdown
     attributes:
--- a/.github/ISSUE_TEMPLATE/1-bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/1-bug-report.yml
@@ -1,6 +1,6 @@
 name: 🐛 Bug Report
 description: Report an issue that should be fixed
-labels: [bug]
+labels: [bug, triage]
 body:
   - type: markdown
     attributes:


yusukebe (commented):
I think we don't need a `bug` label first. What do you think of it?

```yaml
labels: [triage]
```

yusukebe:
Hi @mvares 

Using a `triage` label is good for us. I've created it. Can you check the comment?

mvares:
Hi @yusukebe,

Will we maintain both the labels?

yusukebe:
@mvares 

Yes. I imagine:

1. A user creates an issue with choosing `🐛 Bug Report`
2. A `triage` label is added to the issue.
3. We - maintainers check whether it is an actual bug.
4. If it's a bug, we added a `bug` label.

I don't have any idea for the case if it's triaged, but it's not a bug.

yusukebe:
Ah, but my idea's bad point is the maintainers (almost I) should check all issues.

mvares:
> @mvares
> 
> Yes. I imagine:
> 
> 1. A user creates an issue with choosing `🐛 Bug Report`
> 2. A `triage` label is added to the issue.
> 3. We - maintainers check whether it is an actual bug.
> 4. If it's a bug, we added a `bug` label.
> 
> I don't have any idea for the case if it's triaged, but it's not a bug.

What I think:

An issue only should be assign with `triage` if to pass by triage process. 

This makes sense for you? 🚀

yusukebe:
@mvares 

Sorry, I can't fully understand.

> An issue only should be assign with `triage` if to pass by triage process. 

An issue created by `🐛 Bug Report` will be labeled as nothing? Or `bug`?

mvares:
An issue created by `🐛 Bug Report` will be labeled with the triage label. After that, some maintainer will `analyze` the possible bug. If there is a bug, the issue will receive the bug label.

yusukebe:
@mvares 

Thanks for the explanation. Understood. So [the line](https://github.com/honojs/hono/pull/3092/files#diff-7ef8d00ae79e1c2ac42d02e7069549d69e2324e6cf37f8d4373428f360091babR3) should be the following?:

```yaml
labels: [triage]
```

mvares:
> @mvares
> 
> Thanks for the explanation. Understood. So [the line](https://github.com/honojs/hono/pull/3092/files#diff-7ef8d00ae79e1c2ac42d02e7069549d69e2324e6cf37f8d4373428f360091babR3) should be the following?:
> 
> ```yaml
> labels: [triage]
> ```

Exactly!

yusukebe:
@mvares 

Okay, I'm not 100% sure we can operate with this style, but let's try it now. Thanks!

PR #3102, opened 2024-07-06
chore: update benchmark

yusukebe:
* Update the libraries.
* Add [rou3](https://github.com/unjs/rou3).


yusukebe:
Hi @usualoma 

Just let you know. I've added the [rou3](https://github.com/unjs/rou3)  ([Memoirist](https://github.com/SaltyAom/memoirist) is also fast), which is a fast router though our RegExpRouter is fastest in almost all cases. If you are interested, please run it on your machine!

PR #3103, opened 2024-07-06
chore: replace tsx with Bun

nakasyou:
I replaced tsx with Bun.
Hono uses Bun, so tsx is not needed.

- [ ] Add tests
- [x] Run tests


yusukebe:
@nakasyou 

Exactly! `tsx` is not needed now. Thanks.

PR #3117, opened 2024-07-09
fix(types): allow `string[] | File[]` for RPC form value

### LINKED ISSUE #3116: In Hono RPC, specifying an array as a FormData value will result in type error

4.4.12

Node.js

FormData values ​​can contain arrays, but if you are using Honor RPC you will get a type error.

```typescript
import { Hono } from "hono";
import { validator } from "hono/validator";

const route = new Hono().post(
  "/",

  validator("form", (value: { body?: string[] }, c) => {
    const body = value["body"];

    if (!body || !Array.isArray(body)) {
      return c.text("Invalid!", 400);
    }

    return { body };
  }),

  (ctx) => {
    return ctx.text(`Body is [${ctx.req.valid("form").body}]`)
  }
);

const client = hc<typeof route>("");

client.index.$post({
  form: {
    body: ["1", "2", "3"], /*********** TYPE ERROR ! ***********/
  },
});
```

Hono PRC allows you to specify the array.

```typescript
client.index.$post({
  form: {
    body: ["1", "2", "3"],  // No type errors
  },
});
```

### What do you see instead?

_No response_

### Additional information

Since `"hono/client"` allows you to specify an array, I guess the problem is only with type information.

https://github.com/honojs/hono/blob/51264492269c0f842697736825e47903ae868bca/src/client/client.ts#L68-L80

yusukebe:
Fixes #3116

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1927,9 +1927,11 @@ type MergeTypedResponse<T> = T extends Promise<infer T2>
 //////                             /////
 ////////////////////////////////////////
 
+export type FormValue = string | File
+
 export type ValidationTargets = {
   json: any
-  form: Record<string, string | File>
+  form: Record<string, FormValue | FormValue[]>
   query: Record<string, string | string[]>
   param: Record<string, string> | Record<string, string | undefined>
   header: Record<string, string>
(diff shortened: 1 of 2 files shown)


nakasyou:
How about replacing `File` with `Blob`?
https://developer.mozilla.org/en-US/docs/Web/API/FormData/append#parameters

yusukebe:
@nakasyou 

AH! `Blob`! You are right. `Blob` is better. I'll create another PR to change `File` to `Blob`. Thanks!

yusukebe:
@nakasyou 

Created the issue: https://github.com/honojs/hono/issues/3119 . We can work on it after merging this PR.

PR #3118, opened 2024-07-09
fix(jsx): redefine scope attribute as enum type

yasuaki640:
The scope attribute of the `th` element is defined as an enumerated type in the HTML spec.

[Reference](https://html.spec.whatwg.org/multipage/tables.html#attr-th-scope:~:text=The%20scope%20attribute%20is%20an%20enumerated%20attribute%20with%20the%20following%20keywords%20and%20states%3A)

So I redefined it. 
(To avoid destructive changes, the definition as a string type is retained.)

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/src/jsx/intrinsic-elements.ts
+++ b/src/jsx/intrinsic-elements.ts
@@ -580,7 +580,7 @@ export namespace JSX {
     colspan?: number | undefined
     headers?: string | undefined
     rowspan?: number | undefined
-    scope?: string | undefined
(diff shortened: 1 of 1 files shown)


yusukebe (approved):
LGTM!

yusukebe:
Hello @yasuaki640 

> (To avoid destructive changes, the definition as a string type is retained.)

This is fine!

I'll merge this now. Thanks!

Issue #3122, opened 2024-07-10
4.4.12 breaks client typing when using use with a path


BryanAbate:
### What version of Hono are you using?

4.4.12

### What runtime/platform is your app running on?

Node

### What steps can reproduce the bug?

Stackblitz of the reproduction: https://stackblitz.com/edit/hono-client-typing-issue-middleware?file=src%2Fapp.ts

When using `.use` with a path like `.use('*', myMiddleware)` hono client typings will break. Removing the path `'*'` solves the typing issue.

I suppose https://github.com/honojs/hono/pull/3087 introduced the regression.

EdamAme-x:
types is breaking? or `.use` is breaking?

EdamAme-x:
`.use` always generate `{ [x: string]: never }` 
![image](https://github.com/honojs/hono/assets/121654029/192c4532-58e2-4146-a6c1-50ad41c44d8b)

EdamAme-x:
generics of `.use` was marked at `any`
![image](https://github.com/honojs/hono/assets/121654029/8ea73316-cf79-4503-82f8-70da7e496d36)

EdamAme-x:
`createMiddleware` breaks
![image](https://github.com/honojs/hono/assets/121654029/4b64eaf6-abad-4c00-b250-bcd529df49cf)

EdamAme-x:
I will try to work on this issue.
https://github.com/honojs/hono/blob/main/src/types.ts#L680

EdamAme-x:
tmp
![image](https://github.com/honojs/hono/assets/121654029/24d1e183-31e0-4cf3-9a24-49c643d95d1b)

EdamAme-x:
fixed? I'll add tests
![image](https://github.com/honojs/hono/assets/121654029/fdc0fe3c-0ed0-4372-b163-2db7c430580b)

BryanAbate:
Thank you for looking into the issue @EdamAme-x !

PR #3126, opened 2024-07-11
fix(client): do not suggest unnecenssary route path

EdamAme-x:
Closes https://github.com/honojs/hono/issues/3122

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/src/client/types.ts
+++ b/src/client/types.ts
@@ -144,8 +144,8 @@ export type InferRequestOptionsType<T> = T extends (
   : never
 
 type FilterNeverKeys<T> = {
-  [K in keyof T as (T[K] extends never ? never : T ) extends never ? never : K]: T[K]
-};
+  [K in keyof T as (T[K] extends never ? never : T) extends never ? never : K]: T[K]
+}
 
 type PathToChain<
   Path extends string,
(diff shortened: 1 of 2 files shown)


EdamAme-x:
hi @yusukebe 
this pr contains the following two changes.

## 1.  Not includes never
This issue is involved only when the middleware is chained by `.use`.
When we mark `*` with middleware, the `Schema` value contains never.
```ts
{
  "*": never
} & {...}
```
This will also be suggested by hono client as a route.   
This is a serious UX concern.  
However, this PR does not solve the problem of overriding the handler's schema in the middleware path when defining middleware and handler in the same path.

(Instead of using `&` when creating Schema, I can suggest creating and using the type `mergeSchema`.
I have already built this type.)

## 2. Change Generics of createHandler/createMiddleware
Previously, Path was marked as any by default, so if you did not explicitly mark Path as `any`, it was marked as never when any was on Schema.

So I marked at `""`
image
It is working now.

EdamAme-x:
Maybe the two changes should be in different PRs.
If desired I will do so.

EdamAme-x:
hi @yusukebe 
This PR was half resolved by https://github.com/honojs/hono/pull/3128
but I think "1" is not resolved

yusukebe:
@EdamAme-x 

That is another issue.

EdamAme-x:
ok @yusukebe 
I'll create new issue

yusukebe:
@EdamAme-x 

Don't do that. I'll do it.

PR #3128, opened 2024-07-11
fix(types): `app.use(path, mw)` return correct schema type

### LINKED ISSUE #3122: 4.4.12 breaks client typing when using use with a path

yusukebe:
Fixes #3122

- [ ] Add tests
- [ ] Run tests


yusukebe:
Hey @EdamAme-x 

I think this solution is better than your PR #3126 since the diff is minimal. I've referred to your code, and I'll merge it with you as a co-author. Thanks!

PR #3136, opened 2024-07-13
feat(types): allow passing `interface`s as Bindings / Variables

### LINKED ISSUE #2981: Hono router generic type-errors if it receives an `interface` instead of a `type`

4.4.6

Cloudflare Pages

Attempt to initialise `Hono` router with an `interface`. I have attempted to do so with cloudflare pages. The interface generated by wrangler looks as follows:
```
type Env = {
	DB: D1Database;
};
```

There should be no `ts-error`. This is an example of the `type` syntax

image

### What do you see instead?

image

The following type-error appears:
```
Type '{ Bindings: Env; }' does not satisfy the constraint 'Env'.
  Types of property 'Bindings' are incompatible.
    Type 'Env' is not assignable to type 'Bindings'.
      Index signature for type 'string' is missing in type 'Env'.
```

### Additional information

_No response_

ottomated:
This changes the type of `Bindings` and `Variables` from `Record<string, unknown>` to `object`.

Pro:
- Allows `interfaces` to be used like so:
```ts
interface Env {
  KV: KVNamespace;
}
// would error before
const app = new Hono<{ Bindings: Env }>();
```

Con:
- Looser type - any non-primitive could hypothetically be passed in
```ts
// no error
const app = new Hono<{ Bindings: string[] }>();
```

In my opinion, this trade-off is worth it because Cloudflare's `wrangler types` command generates output like this:
```ts
// Generated by Wrangler
// After adding bindings to `wrangler.toml`, regenerate this interface via `npm run cf-typegen`
interface Env {
  KV: KVNamespace;
}
```

Fixes #2981 

- [x] Add tests
- [x] Run tests


yusukebe:
Hi @ottomated !

I've updated your branch to follow the `next` branch for the next minor release `v4.5.0`. I'll merge this now. Thank you for your contribution!

PR #3137, opened 2024-07-13
feat(types): allow passing `interface`s as Bindings / Variables [2]

### LINKED ISSUE #2981: Hono router generic type-errors if it receives an `interface` instead of a `type`

yusukebe:
Improved #3136.

With this PR, it does now allow looser types as Bindings or Variables:

```ts
// Error!
const app = new Hono<{ Bindings: string[] }>()
```

So, the following tests are passed:

```ts
describe('Generics for Bindings and Variables', () => {
  interface CloudflareBindings {
    MY_VARIABLE: 'my_value'
  }

  it('Should not throw type errors', () => {
    // @ts-expect-error Bindings should be Record<string, unknown> or interface
    new Hono<{
      Bindings: string[]
    }>()

    const appWithInterface = new Hono<{
      Bindings: CloudflareBindings
    }>()

    appWithInterface.get('/', (c) => {
      expectTypeOf(c.env.MY_VARIABLE).toMatchTypeOf<string>()
      return c.text('/')
    })

    const appWithType = new Hono<{
      Bindings: {
        foo: string
      }
    }>()

    appWithType.get('/', (c) => {
      expectTypeOf(c.env.foo).toMatchTypeOf<string>()
      return c.text('Hello Hono!')
    })
  })
})
```

Fixes #2981

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -265,7 +265,7 @@ export class Context<
    * ```
    */
   env: E['Bindings'] = {}
-  #var: E['Variables'] | undefined
+  #var: Map<unknown, unknown> | undefined
   finalized: boolean = false
   /**
    * `.error` can get the error object from the middleware if the Handler throws an error.
@@ -521,9 +521,9 @@ export class Context<
    * ```
 ```
    */
-  set: Set<E> = (key: string, value: unknown) => {
-    this.#var ??= {}
-    this.#var[key as string] = value
+  set: Set<E> = (key: unknown, value: unknown) => {
+    this.#var ??= new Map()
+    this.#var.set(key, value)
   }
 
   /**
@@ -539,8 +539,8 @@ export class Context<
    * })
    * ```
    */
-  get: Get<E> = (key: string) => {
-    return this.#var ? this.#var[key] : undefined
+  get: Get<E> = (key: unknown) => {
+    return this.#var ? this.#var.get(key) : undefined
   }
 
   /**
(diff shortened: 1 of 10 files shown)


yusukebe:
Hi @ottomated!

I've created another PR referring to [your PR](https://github.com/honojs/hono/pull/3136). What do you think about this? I think this is better since it does not allow looser types.

ottomated:
It's a bit of a loosing battle:
```
// No error:
const app = new Hono<{ Bindings: () => string }>()
const app = new Hono<{ Bindings: Promise<void> }>()
const app = new Hono<{ Bindings: Set<string> }>()
```

I chose not to try to prevent these because it doesn't crash anything if the user shoots themselves in the foot, and the added typescript complexity is harder to maintain and requires more CPU.

Your call, though!

yusukebe:
@ottomated 

You are right. My code is not enough. To make it strictly support it, we have to create complex and large types that detect Promise, RegExp,  Uint8Array, etc. So, your approach is good; I noticed it.

I'll close this and accept your PR #3136 later. Thanks!

ottomated:
I added your tests to my PR @yusukebe :)

PR #3138, opened 2024-07-14
fix(type): degradation of generic type handling

### LINKED ISSUE #3135: bug: JSONParsed type not correctly working for generic type

4.4.13

Nodejs

This is happening in the last 2 releases, `v4.4.12` and `v4.4.13`. It seems like the issue is occurring because of this #3074 PR. A community member has created a sample repo - https://github.com/avetisk/dummy-hono-context-helper-type/blob/main/serve.ts

image

The issue is with the generic type (`TData` in this case), if we remove the generic type everything works like normal.

```ts
ok: <TData>(
  data: TData,
  meta?: {
    count?: number;
  }
) => Response &
  TypedResponse<
    { data: TData; meta?: { count?: number } },
    StatusCode,
    "json"
  >;
```

I believe it is occurring because of this new addition in the `JSONParsed` type - 

Screenshot 2024-07-13 at 2 03 02 AM

Even if we `extend` the type, it still gives the error.

_No response_

### What do you see instead?

_No response_

### Additional information

_No response_

m-shaka:
fix #3135 

This works in the use case of the reported issue, I'm not 100% sure why this works though

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/utils/types.ts
+++ b/src/utils/types.ts
@@ -26,7 +26,9 @@ export type IfAnyThenEmptyObject<T> = 0 extends 1 & T ? {} : T
 
 export type JSONPrimitive = string | boolean | number | null
 export type JSONArray = (JSONPrimitive | JSONObject | JSONArray)[]
-export type JSONObject = { [key: string]: JSONPrimitive | JSONArray | JSONObject | object }
+export type JSONObject = {
+  [key: string]: JSONPrimitive | JSONArray | JSONObject | object | InvalidJSONValue
+}
 export type InvalidJSONValue = undefined | symbol | ((...args: unknown[]) => unknown)
 
 type InvalidToNull<T> = T extends InvalidJSONValue ? null : T
--- a/src/context.ts
+++ b/src/context.ts
@@ -180,7 +180,7 @@ type JSONRespondReturn<
       ? JSONValue extends SimplifyDeepArray<T>
         ? never
         : JSONParsed<T>
-      : JSONParsed<T>,
+      : never,
     U,
     'json'
   >
--- a/src/types.test.ts
+++ b/src/types.test.ts
@@ -2240,3 +2240,24 @@ describe('Returning type from `app.use(path, mw)`', () => {
     type verify = Expect<Equal<Expected, Actual>>
   })
 })
+describe('generic typed variables', () => {
+  type Variables = {
+    ok: <TData>(data: TData) => TypedResponse<{ data: TData }>
+  }
+  const app = new Hono<{ Variables: Variables }>()
+
+  it('Should set and get variables with correct types', async () => {
+    const route = app
+      .use('*', async (c, next) => {
+        c.set('ok', (data) => c.json({ data }))
+        await next()
+      })
+      .get('/', (c) => {
+        const ok = c.get('ok')
+        return ok('Hello')
+      })
+    type Actual = ExtractSchema<typeof route>['/']['$get']['output']
+    type Expected = { data: string }
+    expectTypeOf<Actual>().toEqualTypeOf<Expected>()
+  })
+})


src/utils/types.ts
```diff
@@ -26,7 +26,9 @@ export type IfAnyThenEmptyObject<T> = 0 extends 1 & T ? {} : T
 
 export type JSONPrimitive = string | boolean | number | null
 export type JSONArray = (JSONPrimitive | JSONObject | JSONArray)[]
-export type JSONObject = { [key: string]: JSONPrimitive | JSONArray | JSONObject | object }
+export type JSONObject = {
+  [key: string]: JSONPrimitive | JSONArray | JSONObject | object | InvalidJSONValue
```
yusukebe, line 30: A quick question. Is adding `InvalidJSONValue` to `JSONObject` our ideal definition? I feel it's weird `JSONObject` has `InvalidJSONValue`.

m-shaka: I see. That depends on the meaning of `JSONValue`.

For example, the first parameter of `Context.json` can be an object with a key whose value is `InvalidJSONValue` type.
https://github.com/honojs/hono/blob/bb14b1e5bade8ec4811be9d75d1c1112b7c85d82/src/context.ts#L723-L727

I've just given a quick look but `JSONValue` seems to be used in two ways:
- the type of a candidate value to be converted by `JSON.stringify`
- the type of a **already** converted value by `JSON.stringify`
  - like https://github.com/honojs/hono/blob/4d33e1c0d4fc1efe63e9c2da63d3a4a6b0777fae/src/types.ts#L1902-L1910

I'm not sure but I think it's not so harmful for now. We may need refactoring, though

yusukebe: Thank you for the explanation!

How about this? `undefined` is invalid, but having `undefined` as the value of a record is valid:

```diff
diff --git a/src/context.ts b/src/context.ts
index 8d0aa89e..365161fc 100644
--- a/src/context.ts
+++ b/src/context.ts
@@ -180,7 +180,7 @@ type JSONRespondReturn<
       ? JSONValue extends SimplifyDeepArray<T>
         ? never
         : JSONParsed<T>
-      : JSONParsed<T>,
+      : never,
     U,
     'json'
   >
diff --git a/src/utils/types.ts b/src/utils/types.ts
index 3fd17a94..b6cef375 100644
--- a/src/utils/types.ts
+++ b/src/utils/types.ts
@@ -26,7 +26,9 @@ export type IfAnyThenEmptyObject<T> = 0 extends 1 & T ? {} : T

 export type JSONPrimitive = string | boolean | number | null
 export type JSONArray = (JSONPrimitive | JSONObject | JSONArray)[]
-export type JSONObject = { [key: string]: JSONPrimitive | JSONArray | JSONObject | object }
+export type JSONObject = {
+  [key: string]: JSONPrimitive | JSONArray | JSONObject | object | undefined
+}
 export type InvalidJSONValue = undefined | symbol | ((...args: unknown[]) => unknown)

 type InvalidToNull<T> = T extends InvalidJSONValue ? null : T
```

m-shaka: In the meaning of the type of a candidate value, it's ok.
You can call `ctx.json({ a: undefined })` and the type of output is `{}`, which is completely correct because the fact that `JSON.stringify` drops keys with `undefined` is the specification.
Same as `ctx.json({ a: Symbol.for("") })` or `ctx.json({ a: () => {} })`.

However, in the meaning of the latter case (used as a type constraint of `TypedResponse`), it's not correct. It expects types being proceeded by `JSONParsed`, that never have keys with `undefined`, `symbol`, or `function` type.
But, my guess is that in most cases `TypedResponse` is used through `JSONRespondReturn` or like that, and `JSONValue` is used here only for deciding the response format, and types like `{ a: undefined }` should be treated as `'json'` format.
That's why I said it's not so harmful is that 

(I'm not confident if I understand what you mean. Sorry if not.)

yusukebe: @m-shaka

Understood well. You are right that the definition of `JSONObject` is good for now. Sorry for wasting your time. Thanks!

yusukebe (approved):
LGTM!

src/utils/types.ts
m-shaka: No way! Thank you for your patience!

PR #3146, opened 2024-07-16
chore: remove rimraf and use bun shell

nakasyou:
I replaced rimraf with Bun shell.
Bun has OS-independent `rm -rf`, so we can delete rimraf as a dependency.

- [ ] Add tests
- [x] Run tests


yusukebe (approved):
LGTM!

yusukebe:
Hey @nakasyou 

Interesting! It's good to reduce dependencies and actively use Bun features. Thanks. I will merge it now.

mountainash:
If you are making this change, can I suggest changing `"node": ">=16.0.0"` to bun with a version number greater than when the shell commands were enabled?

yusukebe:
Hi @mountainash 

What specific changes would you like to add?

mountainash:
While [it's not supported yet](https://github.com/oven-sh/bun/issues/5846) I'd consider adding something like, to hint to future devs that you'll need bun to run this project (not Node.JS).

```
	"engines": {
		"bun": ">>=1.1.5"
	}
```

yusukebe:
@mountainash 

Ah, it's interesting. We don't have to add it into `package.json` right now, but when Bun supports it, we should do that.

PR #3151, opened 2024-07-17
Fixes #3150 - BunJS not crashing when using Bun.build and then consume the client.

### LINKED ISSUE #3150: hono/client hc<>() breaks when using with BunJS build

4.5.0

Bun

import the hono/client hc to a file and build it using BunJS Bun.build(). Then try to consume it, this will crash the execution.

No errors.

### What do you see instead?

BunJS crash.

### Additional information

Bun related issues:
https://github.com/oven-sh/bun/issues/11079
https://github.com/oven-sh/bun/issues/11880

Migrating the client class ClientRequestImpl to a function fixes the issue.

AlbertSabate:

- [❌] Add tests
- [✔️] Run tests
- [✔️] `bun run format:fix && bun run lint:fix` to format the code
- [❌] Add [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code

While we could wait for BunJS to fix it, we could move the client to use functions instead of a class. This should not change the behavior of the client. Also, I took advantage to improve a bit further the types.

### DIFF
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -39,9 +39,9 @@ function clientRequestImpl(baseUrl: string, method: string) {
   return {
     fetch: async (
       args?: ValidationTargets<FormValue> & {
-        param?: Record<string, string>;
+        param?: Record<string, string>
       },
-      opt?: ClientRequestOptions,
+      opt?: ClientRequestOptions
     ) => {
       if (args) {
         if (args.query) {
@@ -92,8 +92,8 @@ function clientRequestImpl(baseUrl: string, method: string) {
         ...(typeof opt?.headers === 'function'
           ? await opt.headers()
           : opt?.headers
-            ? opt.headers
-            : {}),
+          ? opt.headers
+          : {}),
       }
 
       if (args?.cookie) {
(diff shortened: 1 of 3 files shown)


src/client/client.ts
```diff
@@ -183,14 +180,15 @@ export const hc = <T extends Hono<any, any, any>>(
         'ws'
       )
       const targetUrl = new URL(webSocketUrl)
-      for (const key in opts.args[0]?.query) {
-        targetUrl.searchParams.set(key, opts.args[0].query[key])
+      for (const key in opts.args[0]?.query ?? []) {
+        const value = opts.args[0].query[key]
+        targetUrl.searchParams.set(key, Array.isArray(value) ? value.join(',') : value)
```
AlbertSabate, line 185: Please someone to review this line if looks correct.

yusukebe:
@AlbertSabate 

I don't want to change the class style to the function style. So, can you remove the change for class to function, and can you keep only the type and the search params things?

AlbertSabate:
Removed the class change as mentioned. I left only the small improvements I could find. There is more room to improve the types and remove the as UnionToIntersection<Client<T>>, but I think is good enough to leave it like this.

yusukebe:
Hi @AlbertSabate 

This PR breaks the types:

CleanShot 2024-07-20 at 22 36 56@2x

And there are not enough tests. I can't accept this PR. So, I'll close this PR now and create another PR to solve the Search Param's issue with you as a co-author. Thanks.

PR #3153, opened 2024-07-18
refactor: pull out `.length` from for loop

movahhedi:
On something like the router, the smallest improvement makes a difference.

https://dev.to/siddiqus/which-for-loop-is-the-fastest-in-javascript-4hdf

### DIFF
--- a/src/router/linear-router/router.ts
+++ b/src/router/linear-router/router.ts
@@ -21,7 +21,7 @@ export class LinearRouter<T> implements Router<T> {
   match(method: string, path: string): Result<T> {
     const handlers: [T, Params][] = []
     const routesLength = this.routes.length
-    ROUTES_LOOP: for (let i = 0, len = routesLength; i < len; i++) {
+    ROUTES_LOOP: for (let i = 0; i < routesLength; i++) {
       const [routeMethod, routePath, handler] = this.routes[i]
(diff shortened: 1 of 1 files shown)


src/router/linear-router/router.ts
```diff
@@ -20,7 +20,8 @@ export class LinearRouter<T> implements Router<T> {
 
   match(method: string, path: string): Result<T> {
     const handlers: [T, Params][] = []
-    ROUTES_LOOP: for (let i = 0, len = this.routes.length; i < len; i++) {
+    const routesLength = this.routes.length
+    ROUTES_LOOP: for (let i = 0, len = routesLength; i < len; i++) {
```
EdamAme-x, line 24:
```suggestion
    const routesLength = this.routes.length
    ROUTES_LOOP: for (let i = 0; i < routesLength; i++) {
```

Is this not good enough?

yusukebe:
Hi @movahhedi 

These are the same behavior. It does not calculate the length every time in the current code.

movahhedi:
Oh my God, my eyes don't seem to see any more. 🤦‍♂️
I was looking at it in the GitHub web code viewer.
Sorry for the waste of time.

yusukebe:
@movahhedi No problem!!

Issue #3159, opened 2024-07-18
Issue tracking for experimental features


cybercoder-naj:
### What is the feature you are proposing?

I wish to use `factory.createApp()` but I see it is still experimental (in the docs). I was wondering to what extent is the instability of the function. Is there any issue track for this? What are the known bugs of using `factory.createApp()`?

EdamAme-x:
I have seen people using this in production on several occasions.

cybercoder-naj:
Here's another related issue, i think it would be great [middleware#652](https://github.com/honojs/middleware/issues/652)

yusukebe:
Hi @cybercoder-naj 

It is heavily used, the feeling is good, and we can confirm the design of API is good! I think we can remove the experimental flag. Just wait a bit to make the plan to do it. Thanks!

PR #3160, opened 2024-07-19
feat: initApp of factory support rpc

EdamAme-x:
The purpose of this PR is to add the route declared in `initApp` to `Schema` so that it can be used from RPC.  

Also, since `initApp` is only used in createApp and specifying the types in the factory constructor makes the types more circuitous,
The previous method is marked as deprecated.  

We can remove the `deprecated` and make `initApp` a `createApp` only property.

```ts
// typed
const app = factory.createApp({
  initApp: app => app.get("/factory", (c) => c.text("Hono and Factory is cool")).use(middleware)
})
```

However, since this would not achieve the objective of avoiding redundancy by the factory, we are actually considering the following approach.

```ts
const factory = createFactory().initApp(app => app...)
```

- [x] Add tests
- [x] Run tests

### Related
https://github.com/honojs/hono/issues/3159

### DIFF
--- a/src/helper/factory/index.test.ts
+++ b/src/helper/factory/index.test.ts
@@ -204,7 +204,8 @@ describe('createHandler', () => {
 
 describe('createApp', () => {
   type Env = { Variables: { foo: string } }
-  const factory = createFactory<Env>({
+  const factory = createFactory<Env>()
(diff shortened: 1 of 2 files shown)


EdamAme-x:
Close once

PR #3167, opened 2024-07-19
fix(middleware/jwt): Changed the jwt-secret type to SignatureKey

JulesVerner:
This PR allows the JWT-middleware secret to be a `SignatureKey`, instead of being limited to a `string`.
Since the secret is only passed to `Jwt.verify`, which already takes a `SignatureKey`, the parameter for the middleware should match.

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/middleware/jwt/jwt.ts
+++ b/src/middleware/jwt/jwt.ts
@@ -11,6 +11,7 @@ import type { CookiePrefixOptions } from '../../utils/cookie'
 import { Jwt } from '../../utils/jwt'
 import '../../context'
 import type { SignatureAlgorithm } from '../../utils/jwt/jwa'
+import { SignatureKey } from '../../utils/jwt/jws'
 
 export type JwtVariables = {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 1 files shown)


src/middleware/jwt/jwt.ts
```diff
@@ -11,6 +11,7 @@ import type { CookiePrefixOptions } from '../../utils/cookie'
 import { Jwt } from '../../utils/jwt'
 import '../../context'
 import type { SignatureAlgorithm } from '../../utils/jwt/jwa'
+import { SignatureKey } from '../../utils/jwt/jws'
```
DerZade, line 14:
```suggestion
import type { SignatureKey } from '../../utils/jwt/jws'
```

DerZade (changes requested):

yusukebe:
Hi @JulesVerner 

Thank you for the PR. Looks good. One thing. Could you run `bun run format:fix && bun run lint:fix` to format the code?

yusukebe (approved):
LGTM!

yusukebe:
@JulesVerner Thanks! I'll merge this now.

Issue #3194, opened 2024-07-27
Validator doesn't parse body when request uses content-type application/json; charset=utf-8


bartekbp:
### What version of Hono are you using?

4.5.1

### What runtime/platform is your app running on?

Cloudflare workers

### What steps can reproduce the bug?

There is a bit of a discussion on stack overflow about application/json content-type - if it's valid for it to indicate a charset or not (https://stackoverflow.com/questions/9254891/what-does-content-type-application-json-charset-utf-8-really-mean).

I noticed that `context.json()` returns a response with content-type with a charset included (https://github.com/honojs/hono/blob/main/src/context.ts#L733), but when we send a request with the exact content-type to the app which uses a validator, it fails. It seems to be caused by https://github.com/honojs/hono/blob/main/src/validator/validator.ts#L74 and the fact that `jsonRegex` doesn't include optional matching for a charset.

I prepared a code that demonstrates the issue and a test case.

File app.ts:
```
import { createRoute, OpenAPIHono, z } from '@hono/zod-openapi';

const app = new OpenAPIHono();

const RequestBodySchema = z
  .object({
    name: z.string(),
  })
  .openapi('RequestBody');

const ResponseBodySchema = z
  .object({
    status: z.string(),
  })
  .openapi('ResponseBody');

const route = createRoute({
  method: 'post',
  path: '/test',
  request: {
    body: {
      content: {
        'application/json': {
          schema: RequestBodySchema,
        },
      },
      required: true,
    },
  },
  responses: {
    200: {
      content: {
        'application/json': {
          schema: ResponseBodySchema,
        },
      },
      description: 'Successful response',
    },
  },
});

app.openapi(route, async (c) => {
  const { name } = c.req.valid('json');
  return c.json({
    status: 'ok',
  } satisfies z.infer<typeof ResponseBodySchema>);
});

export default app;

```

file app.test.ts:
```
import { describe, expect, test } from 'vitest';
import app from './app';

test('should return valid response for content-type application/json with a charset', async () => {
  const res = await app.request('/test', {
    method: 'post',
    body: JSON.stringify({
      name: 'John Doe',
    }),
    headers: {
      'content-type': 'application/json; charset=utf-8',
    },
  });
  console.log(await res.text());
  expect(res.status).toBe(200);
});
```

### What is the expected behavior?

I would expect the test to pass. Validator should accept both 'application/json' as well as 'application/json; charset=utf-8'

### What do you see instead?

[...]

PR #3195, opened 2024-07-27
fix(validator): Add double quotation marks to multipart checker regex

CPlusPatch:
Some runtimes, like Bun (and I assume WebKit based stuff) use double quotes for the `boundary` part of `Content-Type` headers.

Here's a real value from my app: `multipart/form-data; boundary="-WebkitFormBoundary6210507107db483aa49299da6f4298bd`.

Currently, the regex does not include double quotes. This PR fixes this issue by adding double quotes. Patch tested working in my app for Hono 4.5.2.

### DIFF
--- a/src/validator/validator.ts
+++ b/src/validator/validator.ts
@@ -24,7 +24,7 @@ export type ValidationFunction<
 type ExcludeResponseType<T> = T extends Response & TypedResponse<any> ? never : T
 
 const jsonRegex = /^application\/([a-z-\.]+\+)?json$/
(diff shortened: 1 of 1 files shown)


CPlusPatch:
Other relevant issue: #3194

yusukebe (approved):
LGTM!

yusukebe:
HI @CPlusPatch !

Thank you for the PR. I don't know if we should add a test, but this should be fixed. I'll merge it now and fix #3194 with another PR.

PR #3205, opened 2024-07-29
fix(jsx/dom): fix performance issue with adding many new node listings

### LINKED ISSUE #3201: JSX DOM render extremely slow

4.5.2

Bun

```jsx
import { render } from "hono/jsx/dom";

const root = document.getElementById("root");

render(
	<>
		<div><span></span></div>
		<div><span></span></div>
		<div><span></span></div>
		<div><span></span></div>
		<div><span></span></div>
		<div><span></span></div>
		<div><span></span></div>
		<div><span></span></div>
		<div><span></span></div>
		<div><span></span></div>
		<div><span></span></div>
		<div><span></span></div>
		<div><span></span></div>
		<div><span></span></div>
		<div><span></span></div>
		<div><span></span></div>
		<div><span></span></div>
		<div><span></span></div>
		<div><span></span></div>
		<div><span></span></div>
		<div><span></span></div>
		<div><span></span></div>
		<div><span></span></div>
		<div><span></span></div>
		<div><span></span></div>
		<div><span></span></div>
		<div><span></span></div>
		<div><span></span></div>
	</>
, root);

```

Rendering something of that complexity should be immediate

### What do you see instead?

It takes about 12 seconds to render the code above. Twice as much if I add one more `<div><span></span></div>`

### Additional information

The time it takes for the `render` function to complete increases exponentially with the amount of nodes.
From what I gathered it seems to get stuck in the [findInsertBefore](https://github.com/honojs/hono/blob/cdd48f4fd45633bf9e04a474c3c0d696c365e592/src/jsx/dom/render.ts#L273) function

usualoma:
fix #3201

* First time rendering does not need to look for elements from child nodes
** Avoid unnecessary checks with `node.pP`.
* Sibling elements are always searched, so there is no need for a for loop.
** Need only `findInsertBefore(node.vC[0])`

I have not added any additional tests as there is already sufficient testing on the expected rendering results.

For performance, although it is a simple test, I have added a test to check that I am not getting as much computation as n to the nth power.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/jsx/dom/index.test.tsx
+++ b/src/jsx/dom/index.test.tsx
@@ -117,6 +117,24 @@ describe('DOM', () => {
     expect(root.innerHTML).toBe('Hello')
   })
 
+  describe('performance', () => {
+    it('should be O(N) for each additional element', () => {
+      const App = () => (
+        <>
+          {Array.from({ length: 1000 }, (_, i) => (
+            <div>
+              <span>{i}</span>
+            </div>
+          ))}
+        </>
+      )
+      render(<App />, root)
+      expect(root.innerHTML).toBe(
+        Array.from({ length: 1000 }, (_, i) => `<div><span>${i}</span></div>`).join('')
+      )
+    })
+  })
+
   describe('attribute', () => {
     it('simple', () => {
       const App = () => <div id='app' class='app' />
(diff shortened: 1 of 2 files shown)


usualoma:
@yusukebe 
Would you please review?

yusukebe:
@usualoma Yes. I'll do it later.

yusukebe (approved):
LGTM!

PR #3213, opened 2024-07-30
feat(client): Add WebSocket Provider Integration Tests and Enhance WebSocket Initialization

naporin0624:
## Background
I use hono/client and wanted to reconnect websockets. However, the $ws method in hono/client internally creates a new WebSocket, so in order to use a websocket client with reconnection, it was necessary to define the url directly. This would lose the type safety of Hono RPC.

This change allows the use of clients that conform to WebSocket interfaces with reconnection, such as [reconnecting-websocket](https://github.com/pladaria/reconnecting-websocket), internally.

### example usage

```ts
import ReconnectingWebSocket from 'reconnecting-websocket';

const client = hc<AppType>(url, {
  webSocket(url, options) {
    return new ReconnectingWebSocket(url, options)
  },
})
client.index.$ws({ query })
```

## What was done

- Added new test cases for WebSocket provider integration to ensure correct initialization and handling of query parameters.
- Enhanced WebSocket initialization logic to support custom WebSocket implementations via client options.
- Updated `hc` function to use `establishWebSocket` method for flexible WebSocket creation.
- Modified `ClientRequestOptions` to include `webSocket` for custom WebSocket handling.

## How to Test

- Run the new and existing test suite to verify functionality and integration.
- Verify the correct inclusion and handling of query parameters in WebSocket URLs.

- [X] Add tests
- [X] Run tests
- [X] Add [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code

### DIFF
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -1250,7 +1250,7 @@ describe('WebSocket Provider Integration', () => {
         console.log(`Message from client: ${event.data}`)
         ws.send('Hello from server!')
       },
-      onClose: () => {
+      onClose() {
         console.log('Connection closed')
       },
     }))
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -1240,3 +1240,80 @@ describe('Redirect response - only types', () => {
     }
   })
 })
+
+describe('WebSocket Provider Integration', () => {
+  const app = new Hono()
+  const route = app.get(
+    '/',
+    upgradeWebSocket((c) => ({
+      onMessage(event, ws) {
+        console.log(`Message from client: ${event.data}`)
+        ws.send('Hello from server!')
+      },
+      onClose: () => {
+        console.log('Connection closed')
+      },
+    }))
+  )
+
+  type AppType = typeof route
+
+  const server = setupServer()
+  beforeAll(() => server.listen())
+  afterEach(() => {
+    vi.clearAllMocks()
+    server.resetHandlers()
+  })
+  afterAll(() => server.close())
+
+  it.each([
+    {
+      description: 'should initialize the WebSocket provider correctly',
+      url: 'http://localhost',
+      query: undefined,
+      expectedUrl: 'ws://localhost/index',
+    },
+    {
+      description: 'should correctly add query parameters to the WebSocket URL',
+      url: 'http://localhost',
+      query: { id: '123', type: 'test', tag: ['a', 'b'] },
+      expectedUrl: 'ws://localhost/index?id=123&type=test&tag=a&tag=b',
+    },
+  ])('$description', ({ url, expectedUrl, query }) => {
+    const webSocketMock = vi.fn()
+    const client = hc<AppType>(url, {
+      webSocket(url, options) {
+        return webSocketMock(url, options)
+      },
+    })
+    client.index.$ws({ query })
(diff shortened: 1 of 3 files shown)


yusukebe:
Hi @naporin0624 ! Thank you for the PR.

> Enhanced WebSocket initialization logic to support custom WebSocket implementations via client options.

Before I review the details, I would like to know the real-world use case of this option. Can you share it?

naporin0624:
@yusukebe 
I didn't explain the background. I apologize.
I use `hono/client` and wanted to reconnect websockets. However, the $ws method in `hono/client` internally creates a `new WebSocket`, so in order to use a websocket client with reconnection, it was necessary to define the url directly. This would lose the type safety of Hono RPC.

This change allows the use of clients that conform to WebSocket interfaces with reconnection, such as [`reconnecting-websocket`](https://github.com/pladaria/reconnecting-websocket), internally.

yusukebe:
@naporin0624 

That makes sense. Thanks! I'll review this.

yusukebe:
Hi @naporin0624 

One thing I want to consider. You've added an option for `webSocket` in two places:

```ts
// 1
const client = hc<AppType>(url, {
  webSocket() {}, // <==
})

// 2
const client = hc<AppType>(url)
client.index.$ws(
  undefiend,
  {
    webSocket() {}, // <==
  }
)
```

It's good, but in my opinion, it's better to have one than two, a few changes as possible. This is because it is easy to add options but difficult to delete existing ones.

If your needs can be satisfied with `1`, using only `1` is best. This would simplify the argument checking. What do you think?

naporin0624:
hi @yusukebe

Thank you for your review.
I specified two types in the same way as fetch, but I think it's not too late to add the implementation of 2 once the need arises. Given the nature of OSS, I understand that it is difficult to delete an API once it has been added, so I will change it to only implement 1, which is necessary this time.

This review has been a very valuable conversation for me. Thank you.

nakasyou:
Hi @naporin0624, I think it's a nice idea. You have to import `WebSocket` from `ws` module if you use old Node.js version, so it's helpful for that.

Then, what do you think about receiving a class, not a function?
```ts
const client = hc<AppType>(url, {
  WebSocket: ReconnectingWebSocket
})
```
Because WebSocket is just class, and ReconnectingWebSocket has an API compatible with WebSocket. I think it more intuitive for WebSocket to receive WebSocket compatible APIs just as fetch receives fetch compatible functions.

naporin0624:
@nakasyou
Thank you for your review. I also considered taking the class as an option and creating a new one at runtime, but I wanted to avoid limiting the options to classes, so I made it a function-based interface.

I'm also thinking about making it so that it can accept `retry` like `retry(new WebSocket())`.

nakasyou:
@naporin0624 I agree. As you said you can easily change options.

yusukebe (approved):
LGTM!

yusukebe:
@naporin0624 

Thank you for changing it. This is simpler and looks good. I'll merge it later. This is actually a `feat` change but is not a so big feature for users, so we can release it in a patch release. Thanks!

naporin0624:
@yusukebe Okay! Looking forward to the release!

yusukebe:
Hi @naporin0624 

I'll merge this into the `main` and release a new patch version!

PR #3218, opened 2024-07-31
fix: factory inherit basePath types

EdamAme-x:
factory don't inherit types of BasePath.

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/src/helper/factory/index.ts
+++ b/src/helper/factory/index.ts
@@ -5,7 +5,7 @@
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { Hono } from '../../hono'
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @EdamAme-x 

It exactly lacks to inherit the type but, what the situation do you want this fix?

EdamAme-x:
@yusukebe 
Assuming I created an instance to pass to `.route`, the default was `/`, which was not consistent with the actual path used.
Therefore, this PR was created.

yusukebe:
@EdamAme-x 

I see, but please provide an example code!

EdamAme-x:
ok, I'll create example later.

EdamAme-x:
I'm closing this because I just now found another way to solve the problem.

PR #3224, opened 2024-08-02
fix(jsx): corrects the type of 'draggable' attribute in intrinsic-elements.ts

### LINKED ISSUE #3223: draggable attribute causing error Deno: Type string is not assignable to type boolean | undefined

4.5.1

Deno

try to return a div with 'draggable` attribute

```
<div draggable="true"> I'M A DRAGGABLE DIV</div>
```

it should be possible to add draggable attribute:
- if you add `draggable=true` it renders correctly, but compiler complains
- if you add `draggable` compiler doesn't complain, but renders without our attribute, so you can't drag
- if you add `draggable={true}` similarly as above

### What do you see instead?

```console
TS2322 [ERROR]: Type 'string' is not assignable to type 'boolean | undefined'.
          draggable="true"
          ~~~~~~~~~
    at file:///home/karol/projects/pliker.net/repliker/client/dg/forms/available_properties.tsx:29:11

    The expected type comes from property 'draggable' which is declared here on type 'HTMLAttributes'
        draggable?: boolean | undefined;
        ~~~~~~~~~
        at https://jsr.io/@hono/hono/4.5.1/src/jsx/intrinsic-elements.ts:155:5
```

### Additional information

_No response_

yasuaki640:
closes #3223 

Since [the html specification defines the draggable attribute as an enum](https://html.spec.whatwg.org/multipage/dnd.html#the-draggable-attribute),
we have modified it to conform to that specification.

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/jsx/intrinsic-elements.ts
+++ b/src/jsx/intrinsic-elements.ts
@@ -152,7 +152,7 @@ export namespace JSX {
     contenteditable?: boolean | 'inherit' | undefined
     contextmenu?: string | undefined
     dir?: string | undefined
-    draggable?: boolean | undefined
+    draggable?: 'true' | 'false' | undefined
     enterkeyhint?: 'enter' | 'done' | 'go' | 'next' | 'previous' | 'search' | 'send' | undefined
     hidden?: boolean | undefined
     id?: string | undefined


yusukebe (approved):
LGTM!

yusukebe:
@yasuaki640 

Thanks! I didn't know it should be an enum. Merging now.

PR #3229, opened 2024-08-03
fix(types): `param` in `ValidationTargets` supports optional param

### LINKED ISSUE #3227: Types: `zValidator` `param` validation makes input optional where it should be required
Reproduction:
1. Given the exact same schema with 1 required field, create 2 validators: one for `param`, the other one for `json`:
  ```ts
const schema = z.object({ id: z.string() })

const paramValidator = zValidator("param", schema)

const jsonValidator = zValidator("json", schema)
```

2. For the `jsonValidator`, the `input` type is inferred correctly: `id` is required
CleanShot 2024-08-01 at 22 48 50@2x

3. For the `paramValidator`, however, the `id` field in the `input` type is optional for an unknown reason, while it should be marked as required since the schema says it is not optional and not nullable:
CleanShot 2024-08-01 at 22 50 45@2x

This example is not working as expected, i.e. accepts `undefined` as a param:
CleanShot 2024-08-01 at 22 51 52@2x

Suspected commit: https://github.com/honojs/middleware/commit/4875e1c53146d2c67846b8159d3630d465c2a310

[Link to TS Playground](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgCQgOwgGjgSTQMwFMoAlQgRwFdCBnGAFQE8xC4BfOfKCEOAIgAW6CHwBQoSLERwAXgDUAhgBtgAEwUxoHLj34ABIRgD0MiKoC0AN2VqN0MROjwkM9p268+p1Q-BPpAgDGbjqehhBGgSqEaDBiooHodHA0gQKEIApwALyyAHQQAEYAVoSBMAAUSGoAXPl0UMBoAOYVAJTsbaIJSfBgClAKIIoq6ppQObIjtuMVfP2DIHzYqemZXT1oycU06NNj0JPyNgdQczvoyylpGQobiVvwCmBgk2iEAO4owu15zYSVPhGXYgQhGGpqK6iOBwBZDfZ2KCYaFwBQ0RhoYIPGCEAAe8GyAD44Ni8TA8hc0HMIABrPhdDYwZisEgQSg4miTJksCD4VEvboPZJRYAxAlwIIAHlZ7NohLm9MF0VieRBhAA2nwIT4ALp5AAk-0qCBRcJAdRNMJhtTglDQqkI+CahFUcCMRjgEAEcAwKyElCUqjQAHJ4IVWMolBAPi6UWxRGw2kA)

### PR DESCRIPTION
**Pull request by yusukebe**
Fixes #3227

In the current branch, `param` will be `Record<string, string> | Record<string, string | undefined>` regardless of whether the value of path param is optional or not. With this PR, the value type will be `string` by default, but it will be `string | undefined` if it is optional.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1808,7 +1808,7 @@ type ChangePathOfSchema<S extends Schema, Path extends string> = keyof S extends
   : { [K in keyof S as Path]: S[K] }
 
 export type Endpoint = {
-  input: Partial<ValidationTargets>
+  input: any
   output: any
   outputFormat: ResponseFormat
   status: StatusCode
@@ -1930,11 +1930,11 @@ type MergeTypedResponse<T> = T extends Promise<infer T2>
 export type FormValue = string | Blob
 export type ParsedFormValue = string | File
 
-export type ValidationTargets<T extends FormValue = ParsedFormValue> = {
+export type ValidationTargets<T extends FormValue = ParsedFormValue, P extends string = string> = {
   json: any
   form: Record<string, T | T[]>
   query: Record<string, string | string[]>
-  param: Record<string, string> | Record<string, string | undefined>
+  param: Record<P, P extends `${infer _}?` ? string | undefined : string>
   header: Record<string, string>
   cookie: Record<string, string>
 }
--- a/src/validator/validator.test.ts
+++ b/src/validator/validator.test.ts
@@ -833,6 +833,11 @@ it('With path parameters', () => {
 
   const route = app.put(
     '/posts/:id',
+    validator('param', () => {
+      return {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/types.ts [hunk 1]
```diff
@@ -1808,7 +1808,7 @@ type ChangePathOfSchema<S extends Schema, Path extends string> = keyof S extends
   : { [K in keyof S as Path]: S[K] }
 
 export type Endpoint = {
-  input: Partial<ValidationTargets>
+  input: any
```
yusukebe, line 1811: Relaxing the type. It's not necessary to be strict.

PR #3241, opened 2024-08-07
fix(jsx): allow null, undefined, and boolean to be returned from function component

### LINKED ISSUE #3235: Fragment does'n work in custom JSX Component

4.5.1

Deno

- create custom JSX Component 
```ts
function MyComponent() {
  return <></>;
}
```
- use this component
```ts
  return <MyComponent></MyComponent>;
```
and the app crash.

should render Fragment without import

### What do you see instead?

```console
2024-08-06T08:11:01.631Z [ERROR] [client] TypeError: Cannot read properties of null (reading 'isEscaped')
    at JSXFunctionNode.toStringToBuffer (https://jsr.io/@hono/hono/4.5.1/src/jsx/base.ts:251:64)
    at JSXFunctionNode.toString (https://jsr.io/@hono/hono/4.5.1/src/jsx/base.ts:143:12)
    at html (https://jsr.io/@hono/hono/4.5.1/src/helper/html/index.ts:35:29)
```

### Additional information

if you import `import { Fragment } from "hono/jsx/jsx-runtime";` it works, but IDE not report any info that something is wrong

usualoma:
fixes #3235
The deno converts an empty `Fragment` to `null` when the value of the "jsx" entry is "precompile", but "hono/jsx" was not able to handle the function component returning `null`, which caused the problem #3235.

### What should be the returned value of an `FC` type?

In the current React type definition, `FunctionComponent` returns the `ReactNode`, and since `ReactNode` also includes `boolean` and `undefined`, it would be preferable if Hono's `FC` also included these (`boolean` and `undefined`). 

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L1114-L1122
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L478-L489

However, the changes to add `undefined` and `boolean` have a considerable impact on existing code, so I have chosen not to do so in this pull request. I would like to discuss this at another time.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/runtime_tests/deno-jsx/jsx.test.tsx
+++ b/runtime_tests/deno-jsx/jsx.test.tsx
@@ -28,6 +28,12 @@ Deno.test('JSX: Fragment', () => {
   assertEquals(fragment.toString(), '<p>1</p><p>2</p>')
 })
 
+Deno.test('JSX: Empty Fragment', () => {
+  const Component = () => <></>
+  const html = <Component />
+  assertEquals(html.toString(), '')
+})
+
 Deno.test('JSX: Async Component', async () => {
   const Component = async ({ name }: { name: string }) =>
     new Promise<HtmlEscapedString>((resolve) => setTimeout(() => resolve(<span>{name}</span>), 10))
--- a/src/jsx/base.ts
+++ b/src/jsx/base.ts
@@ -247,7 +247,10 @@ class JSXFunctionNode extends JSXNode {
       children: children.length <= 1 ? children[0] : children,
     })
 
-    if (res instanceof Promise) {
+    if (typeof res === 'boolean' || res == null) {
+      // boolean or null or undefined
+      return
+    } else if (res instanceof Promise) {
       if (globalContexts.length === 0) {
         buffer.unshift('', res)
       } else {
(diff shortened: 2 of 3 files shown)


usualoma:

yusukebe:
@usualoma 

Okay! I'll do it later.

yusukebe:
Hi @usualoma 

Looks good. I want to know:

> In the current React type definition, `FunctionComponent` returns the `ReactNode`, and since `ReactNode` also includes `boolean` and `undefined`, it would be preferable if Hono's `FC` also included these (`boolean` and `undefined`). 

Does this issue already exist before this PR (Or #3235)? As you said, I also think we should not include the change in this PR.

usualoma:
@yusukebe 

Thanks for the review!

The `FC` return type divergence was noticed during this work. I don't think there are any known issues.

I think it would be better to include `undefined` and `boolean`, but I think components that return `undefined` or `boolean` are very rare. Changing it would likely affect the result of `<Component/>.toString()`, so I don't think it should be changed unless there is some special motivation.

yusukebe:
@usualoma 

Thank you for your reply.

> I think it would be better to include `undefined` and `boolean`, but I think components that return `undefined` or `boolean` are very rare. Changing it would likely affect the result of `<Component/>.toString()`, so I don't think it should be changed unless there is some special motivation.

I agree 👍 

Now, let's merge this one!

yusukebe (approved):
LGTM!

PR #3274, opened 2024-08-15
fix(jsx): handle async component error explicitly and throw the error in the response

### LINKED ISSUE #3252: Broken error handling with JSX renderer middleware

4.5.4

Deno 1.45.5 on Windows 11

Throwing an exception from a JSX-rendered async component can crash the server entirely in some cases.

Specifically, the layout needs to take in additional content that is also an async component, and that takes longer to render than the broken content.

EDIT: I have also run into this when using nested layouts instead of extra layout props, but I don't have a reproduction for that one.

```tsx
#!/usr/bin/env -S deno run --allow-net

/** @jsxImportSource jsr:@hono/hono@4.5.4/jsx */

import { Hono } from "jsr:@hono/hono@4.5.4";
import { Child, FC } from "jsr:@hono/hono@4.5.4/jsx";
import { jsxRenderer } from "jsr:@hono/hono@4.5.4/jsx-renderer";

declare module "jsr:@hono/hono@4.5.4" {
    interface ContextRenderer {
        (
            content: string | Promise<string>,
            props: { extraContent: Child },
        ): Response;
    }
}

const app = new Hono();

app.get(
    "*",
    jsxRenderer((props) => {
        return (
            <>
                {props.extraContent}
                {props.children}
            </>
        );
    }),
);

const ExtraContent: FC = async () => {
    await new Promise((resolve) => setTimeout(resolve, 500));

    return <div>This is some extra content!</div>;
};

const BrokenComponent: FC = async () => {
    await new Promise((_resolve, reject) => {
        setTimeout(() => {
            reject(new Error("oopsie"));
        }, 250);
    });

    return <div>This component is broken :(</div>;
};

app.get("/test", (ctx) => {
    return ctx.render(
        <BrokenComponent />,
        {
            extraContent: <ExtraContent />,
        },
    );
});

Deno.serve(app.fetch);
```

Hono handles the exception gracefully, calling `onError` hooks, etc. and the server continues running

### What do you see instead?

The server crashes

### Additional information

I originally posted this as a Discussion: https://github.com/orgs/honojs/discussions/3244

usualoma:
fixes #3252

`Uncaught (in promise)` was occurring and the app was crashing under the following conditions.

* There are multiple asynchronous components.
* While waiting for the first Promise to be resolved, an error occurred in the subsequent Promise.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/jsx/index.test.tsx
+++ b/src/jsx/index.test.tsx
@@ -112,6 +112,39 @@ describe('JSX middleware', () => {
     expect(res.headers.get('Content-Type')).toBe('text/html; charset=UTF-8')
     expect(await res.text()).toBe('<html><body><h1>Hello from async component</h1></body></html>')
   })
+
+  it('Should handle async component error', async () => {
+    const componentError = new Error('Error from async error component')
+
+    const AsyncComponent = async () => {
+      await new Promise((resolve) => setTimeout(resolve, 10))
+      return <h1>Hello from async component</h1>
+    }
+    const AsyncErrorComponent = async () => {
+      await new Promise((resolve) => setTimeout(resolve, 0))
+      throw componentError
+    }
+
+    let raisedError: any
+    app.onError((e, c) => {
+      raisedError = e
+      return c.html('<html><body><h1>Error from onError</h1></body></html>', 500)
+    })
+    app.get('/', (c) => {
+      return c.html(
+        <>
+          <AsyncComponent />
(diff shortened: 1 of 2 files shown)


usualoma:

yusukebe (approved):
LGTM!

yusukebe:
@usualoma 

I've confirmed that this patch works well on several runtimes. Looks good! Thanks.

Issue #3264, opened 2024-08-12
Client components are very slow and cause stack overflow 


aminya:
### What version of Hono are you using?

4.5.4

### What runtime/platform is your app running on?

Chrome

### What steps can reproduce the bug?

Here's the full benchmark reproduction.
https://github.com/krausest/js-framework-benchmark/pull/1723

Hono is much slower than [Solidjs](https://www.solidjs.com/), the reference fast framework. The current result makes Hono the slowest JS framework. Up to 100 times slower in a medium number of components, and the browser stack overflows when the number of components is high (benchmarks fail).

![image](https://github.com/user-attachments/assets/1b77a8d2-0f59-48a2-b8bc-4a6d7e480a51)

The profiler confirms the inefficient implementation of client components in Hono.

![image](https://github.com/user-attachments/assets/030f06be-4462-427a-a743-ed38bde49d1d)

### What is the expected behavior?

Have an acceptable performance, at least on par with React.

https://krausest.github.io/js-framework-benchmark/2024/table_chrome_127.0.6533.72.html

### What do you see instead?

Use other frameworks like [Solidjs](https://www.solidjs.com/) for client components.

usualoma:
Hi @aminya, thank you for creating the issue!

I am investigating now, and we can probably achieve the expected performance, so wait a moment!

PR #3203, opened 2024-07-28
feat(jsx/dom): improve performance of `findInsertBefore`

ryuapp:
Fix #3201

I honestly don't know much about it, but eliminating recursive calls improves performance.
We may need accurate benchmarks or tests.

I tried with the following example:
https://github.com/ryuapp/hono-vite-template
At least increasing the issue example didn't slow it down.

### The author should do the following, if applicable

- [ ] Add tests
- [ ] Run tests
- [ ] `bun run format:fix && bun run lint:fix` to format the code
- [ ] Add [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code

PR #3288, opened 2024-08-18
perf(jsx/dom): improve performance

usualoma:
#3264

#### use for-loop instead of recursion for `findInsertBefore` 0dadfbc26c44bff5c5789a34ee20184675648b57

As @ryuapp  pointed out in #3203, the implementation with recursive calls did not perform well and sometimes resulted in errors when there were a large number of elements, so it was modified so that recursive calls were not made.

#### assign type and ref to node object directly. `Object.defineProperties` is too slow 7fd73efa4ba42002eb335c2ba2e993e482edb55d

* `reactElement.type`
* `reactElement.ref`

https://github.com/honojs/hono/compare/main...usualoma:hono:perf/jsx-dom?expand=1#diff-c9b14c1b74662b399c25e6dc0c870408c466e1c96e39a1d429903b2a3d0db51fL11-L20

The code to make the following properties referential is for React compatibility, but the implementation using `Object.defineProperties` was very expensive. This change is not pretty, but it is an acceptable redundancy when considered as a trade-off for improved performance.

#### Reduce recalculation and reapplication when properties are unchanged

In some cases, DOM attributes were being reassigned even when there were no changes to the properties, so reassignment is no longer performed when it is unnecessary.

### benchmark

Based on this branch (thank you very much @aminya!),

https://github.com/aminya/js-framework-benchmark/tree/hono

The following changes were added and benchmarked in my environment.

```diff
diff --git a/frameworks/keyed/hono/src/index.tsx b/frameworks/keyed/hono/src/index.tsx
index 2628e205..998c4b0f 100644
--- a/frameworks/keyed/hono/src/index.tsx
+++ b/frameworks/keyed/hono/src/index.tsx
@@ -141,9 +141,8 @@ const App = () => {
     <table class="table table-hover table-striped test-data">
       <tbody>
         {data.map(item => {
-          console.log(item);
           return (
-            <Row item={item} selected={selected === item.id} dispatch={dispatch} />
+            <Row key={item.id} item={item} selected={selected === item.id} dispatch={dispatch} />
           );
         })}
       </tbody>
```

Performance for creation or update is roughly equivalent to React, with smaller file sizes and memory usage, which is an expected result since "hono/jsx" aims for "an implementation with a React-compatible API that is small enough to be practical with a small amount of code".

jsx-dom-bench

- [x] Add tests
    - Only a few tests have been added, but as most of the changes are internal refactoring with no external specification changes, no further additional tests are considered necessary. 
- [x] Run tests

### DIFF
--- a/src/jsx/dom/context.ts
+++ b/src/jsx/dom/context.ts
@@ -2,7 +2,7 @@ import type { Child } from '../base'
 import { DOM_ERROR_HANDLER } from '../constants'
 import type { Context } from '../context'
 import { globalContexts } from '../context'
-import { newJSXNode, setInternalTagFlag } from './utils'
+import { setInternalTagFlag } from './utils'
 
 export const createContextProviderFunction =
   <T>(values: T[]): Function =>
@@ -33,7 +33,7 @@ export const createContextProviderFunction =
       }),
       props: {},
     })
-    const res = newJSXNode({ tag: '', props })
+    const res = { tag: '', props, type: '' }
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     ;(res as any)[DOM_ERROR_HANDLER] = (err: unknown) => {
       values.pop()
--- a/src/jsx/dom/index.test.tsx
+++ b/src/jsx/dom/index.test.tsx
@@ -736,6 +736,16 @@ describe('DOM', () => {
     })
   })
 
+  describe('dangerouslySetInnerHTML', () => {
+    it('string', () => {
+      const App = () => {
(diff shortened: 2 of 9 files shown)


usualoma:

usualoma:
Minor refactorings and tests are added. 
There is no change in behavior, but test coverage scores have improved.

yusukebe:
@usualoma 

Thank. I'll review this.

yusukebe (approved):
LGTM!

aminya:
> Based on this branch (thank you very much @aminya!),

You're welcome! It's nice to see that the benchmark effort resulted in these improvements. I'll update the benchmark based on the new version.

PR #3221, opened 2024-08-01
feat(context): Add types for `c.header`

nakasyou:
I added a types for `c.header`.
I'm sure you can type headers easier.

IMG_3616
IMG_3617

If we merged this, we can use custom header names as before.

From: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers
### The author should do the following, if applicable

- [ ] Add tests
- [x] Run tests
- [x] `bun run format:fix && bun run lint:fix` to format the code
- [x] Add [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code

PR #3289, opened 2024-08-18
feat(context): Custom header name for c.header

akku1139:
Power UP #3221 .
You can define your own headers and the editor (e.g. VSCode) can complete them.

```ts
import { Hono } from "hono"

const app = new Hono<{
  Headers: "X-Hono-Test" | "Hello-Hono"
}>()

app.get("/", (c) => {
  c.header("Hello-Hono") // Defined header will be completed
  return c.text("Hello Hono!", 100, {
  })
})
```

image

- [ ] Add tests
- [ ] Run tests


yusukebe:
Hi @akku1139 

Supporting adding extra header names is good. But I don't agree with using the generics for `Hono`. The generics for `Hono` have two types, `Bindings` and `Variables`. These are the most important types in the Hono type system. To keep the generics type definition simple, I think we don't have to add the `Headers` type for the generics. It should have only two important types.

I think it would be good to extend types in the following way:

https://hono.dev/docs/guides/jsx#override-type-definitions

yusukebe:
Regarding adding custom header types, I think it's not good to pass generics to `new Hono` because of [the above reason](https://github.com/honojs/hono/pull/3289#issuecomment-2295208742). Instead of it, we can extend the types with the following method if we can implement it:

```ts
declare module 'hono/context' {
  interface CustomResponseHeader {
    'X-Custom-Foo': 'foo' | 'foo2' | 'foo3'
    'X-Custom-Bar': 'bar' | 'bar2' | 'bar3'
  }
}
```

@nakasyou What do you think of it?

nakasyou:
I think it is better than creating a generic.
Also it's more convenient if there are multi instances of Hono. I recommend the idea.

yusukebe:
This is an interesting idea. But it's over to add generics to `new Hono()` for header names. Closing. Thanks.

PR #3292, opened 2024-08-18
feat(client): improve handling status code types

### LINKED ISSUE #580: Status code used in the third argument of zValidator is not inferred on the client side
I have the following code. When trying to handle branching on the client side based on `res.status`, it always ends up being 200. Is there any documentation or solution regarding this issue?

```ts
const sampleRoute = new Hono().get(
  '/sample',
  zValidator(
    'json',
    z.object({
      keyword: z.string().min(1),
    }),
    (result, c) => {
      if (!result.success) {
        return c.json({ error: 'error!' }, 400)
      }
    }
  ),
  async (c) => {
    return c.json({ data: 'data' }, 200)
  }
)
```

in client
スクリーンショット 2024-06-16 13 04 46

### LINKED ISSUE #2719: RPC supports middleware responses

Currently, the RPC mode does not support middleware response. The client can't handle the response from the middleware:

CleanShot 2024-05-19 at 06 46 01@2x

I'm not 100% sure, but perhaps we can implement it by modifying `types.ts`:

```diff
diff --git a/src/types.ts b/src/types.ts
index 3c7a53d..5e93fc1 100644
--- a/src/types.ts
+++ b/src/types.ts
@@ -190,13 +190,22 @@ export interface HandlerInterface<
     I3 extends Input = I & I2,
     E2 extends Env = E,
     E3 extends Env = E,
-    E4 extends Env = IntersectNonAnyTypes<[E, E2, E3]>
+    E4 extends Env = IntersectNonAnyTypes<[E, E2, E3]>,
+    // Response from Middleware
+    RM1 extends HandlerResponse<any> = any,
+    RM2 extends HandlerResponse<any> = any
   >(
     path: P,
-    ...handlers: [H<E2, MergedPath, I>, H<E3, MergedPath, I2>, H<E4, MergedPath, I3, R>]
+    ...handlers: [H<E2, MergedPath, I, RM1>, H<E3, MergedPath, I2, RM2>, H<E4, MergedPath, I3, R>]
   ): Hono<
     IntersectNonAnyTypes<[E, E2, E3, E4]>,
-    S & ToSchema<M, MergePath<BasePath, P>, I3, MergeTypedResponse<R>>,
+    S &
+      ToSchema<
+        M,
+        MergePath<BasePath, P>,
+        I3,
+        MergeTypedResponse<R> | MergeTypedResponse<RM1> | MergeTypedResponse<RM2>
+      >,
     BasePath
   >
```

CleanShot 2024-05-19 at 06 48 11@2x

It will be worth implementing, though the code will be very long.

### LINKED ISSUE #3170: Hono doesn't consider status codes from middlewares with `testClient`

4.4.9

Bun

```ts
// index.ts
import { Hono } from "hono";
import { createFactory } from "hono/factory";

const myMiddleware = () => {
  return factory.createMiddleware(async (c, next) => {
    const random = Math.random();
    if (random > 0.5) {
      c.status(500);
      return c.text("Probabilistic error", 500);
    }
    return next();
  });
};

const app = new Hono().use(logger()).get("/", myMiddleware(), async (c) => {
  return c.json({ message: "Hello, World!" }, 200);
});

export default app;

// index.test.ts
import { test, expect } from "bun:test";
import app from ".";
import { testClient } from "hono/testing";

test("GET /", async () => {
  const res = await testClient(app).index.$get();

  expect(res.status).toBe(500); <- "Argument of type '500' is not assignable to parameter of type '200'"
});
```

Consider other status codes that may be returned from middlewares or other factors.

### What do you see instead?

`Argument of type '500' is not assignable to parameter of type '200'`

### Additional information

_No response_

yusukebe:
The PR will improve the handling of TypeScript types of status codes in the Response returned by the client.

Fixes #2719
Fixes #3170
Fixes https://github.com/honojs/middleware/issues/580

### Allowing `number` for `res.status`

Imagine the following application which throws the `HTTPException` in the handler:

```ts
import { Hono } from 'hono'

const app = new Hono()
const flag = {}

const routes = app.get('/foo', (c) => {
  if (flag) {
    throw new HTTPException(500, {
      message: 'Server Error!',
    })
  }
  return c.json({ ok: true }, 200)
})
```

You will create the client with the types:

```ts

const client = hc<typeof routes>('/')
const res = await client.foo.$get()
```

The current pain point is it can't infer the response having `500` status correctly.

```ts
if (res.status === 500) {
  const data = await res.json()
}
```

The error showing in your IDE:

CleanShot 2024-08-18 at 18 34 22@2x

It should be fixed. However, handling the response content and status with an `HTTPException`  or `Error`  object is impossible. So, I've found the solution to enable accepting `number` for `res.status`.

```ts
if (res.status === 500) { // res.status can be `number`. **No** error!
  const data = await res.json()
}
```

### Accept generics for `res.json()`

That solution solves the problem of the error of `res.status`, but the content returned from `res.json()` will not be typed:

```ts
if (res.status === 500) {
  const data = await res.json() // never. not typed
}
```

Also to set the correct types is impossible for it. So, I'll introduce `res.json()` accepting generrics. You can write the following:

```ts
if (res.status === 500) {
  const data = await res.json<{ message: string }>()
}
```

This method does not allow automatic inference, but it avoids type errors.

### Other improvements

I added some modifications since other tests failed with the changes.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/client/types.test.ts
+++ b/src/client/types.test.ts
@@ -1,7 +1,12 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
 import { expectTypeOf } from 'vitest'
 import { Hono } from '..'
 import { upgradeWebSocket } from '../adapter/deno/websocket'
 import { hc } from '.'
+import { HTTPException } from '../http-exception'
+import type { Equal, Expect } from '../utils/types'
+import { setupServer } from 'msw/node'
+import { HttpResponse, http } from 'msw'
 
 describe('WebSockets', () => {
   const app = new Hono()
@@ -55,5 +60,67 @@ describe('with the leading slash', () => {
   })
   it('`foo[":id"].baz` should have `$get`', () => {
     expectTypeOf(client.foo[':id'].baz).toHaveProperty('$get')
+  })
+})
+
+describe('Status Code - test only Types', () => {
+  const server = setupServer(
+    http.get('http://localhost/foo', async () => {
+      return HttpResponse.json({})
+    })
+  )
+
+  beforeAll(() => server.listen())
+  afterEach(() => server.resetHandlers())
+  afterAll(() => server.close())
+
+  const app = new Hono()
+  const flag = {}
+
+  const routes = app.get('/foo', (c) => {
+    if (flag) {
+      throw new HTTPException(500, {
+        message: 'Server Error!',
+      })
+    }
+    if (flag) {
+      return c.json({ message: 'invalid!' }, 401)
+    }
+    if (flag) {
+      return c.redirect('/', 301)
+    }
+    return c.json({ ok: true }, 200)
+  })
+
+  const client = hc<typeof routes>('http://localhost')
+
+  it('Should handle different status codes', async () => {
+    const res = await client.foo.$get()
+
+    if (res.status === 500) {
+      const data = await res.json<{ errorMessage: string }>()
+      type Expected = { errorMessage: string }
+      type verify = Expect<Equal<Expected, typeof data>>
+    }
+
+    if (res.status === 401) {
+      const data = await res.json()
+      type Expected = { message: string }
+      type verify = Expect<Equal<Expected, typeof data>>
+    }
+
+    if (res.status === 200) {
+      const data = await res.json()
+      type Expected = { ok: boolean }
+      type verify = Expect<Equal<Expected, typeof data>>
+    }
+  })
+
+  it('Should infer union types', async () => {
+    const res = await client.foo.$get()
+
+    const data = await res.json()
+    type Expected = { message: string } | { ok: boolean }
+    type verify = Expect<Equal<Expected, typeof data>>
   })
 })
--- a/src/client/types.ts
+++ b/src/client/types.ts
@@ -1,6 +1,6 @@
 import type { Hono } from '../hono'
 import type { Endpoint, ResponseFormat, Schema } from '../types'
-import type { StatusCode, SuccessStatusCode } from '../utils/http-status'
+import type { RedirectStatusCode, StatusCode, SuccessStatusCode } from '../utils/http-status'
 import type { HasRequiredKeys } from '../utils/types'
 
 type HonoRequest = (typeof Hono.prototype)['request']
@@ -63,7 +63,15 @@ type ClientResponseOfEndpoint<T extends Endpoint = Endpoint> = T extends {
   outputFormat: infer F
   status: infer S
 }
-  ? ClientResponse<O, S extends number ? S : never, F extends ResponseFormat ? F : never>
+  ? F extends 'redirect'
+    ? ClientResponse<O, S extends RedirectStatusCode ? S : never, 'redirect'>
+    :
+        | ClientResponse<O, S extends StatusCode ? S : never, F extends ResponseFormat ? F : never>
+        | ClientResponse<
+            {},
+            S extends StatusCode ? Exclude<Exclude<StatusCode, RedirectStatusCode>, S> : never,
+            F extends ResponseFormat ? F : never
+          >
   : never
 
 export interface ClientResponse<
@@ -84,11 +92,13 @@ export interface ClientResponse<
   url: string
   redirect(url: string, status: number): Response
   clone(): Response
-  json(): F extends 'text'
+  json<JSONT>(): F extends 'text'
     ? Promise<never>
     : F extends 'json'
-    ? Promise<BlankRecordToNever<T>>
-    : Promise<unknown>
+    ? undefined extends JSONT
+      ? Promise<BlankRecordToNever<T>>
+      : Promise<JSONT>
+    : Promise<T & {}>
   text(): F extends 'text' ? (T extends string ? Promise<T> : Promise<never>) : Promise<string>
   blob(): Promise<Blob>
   formData(): Promise<FormData>
@@ -119,12 +129,14 @@ export type InferResponseType<T, U extends StatusCode = StatusCode> = InferRespo
   U
 >
 
+type WithoutEmptyObject<T> = T extends {} ? (keyof T extends never ? never : T) : T
+
 type InferResponseTypeFromEndpoint<T extends Endpoint, U extends StatusCode> = T extends {
   output: infer O
   status: infer S
 }
   ? S extends U
-    ? O
+    ? WithoutEmptyObject<O>
     : never
   : never
 


yusukebe:
Hey @m-shaka @NamesMT 

Can you share your opinion on this PR and review it?

m-shaka:
Ok, I can see that later!

NamesMT:
The idea seems good to me,  
If possible, I think adding a way to define a default type for unknown/certain status codes would be greatly appreciated, I'll try to bring up some PoC later tonight when I got back home.

m-shaka:
generic typed `json` looks nice!

However, number type `status` may cause some problems.
First, users of the client can't know which status they should handle from `status` type.
For example, users can understand 200 and 400 should be handled from the type of `status`, without reading the server code.
image
image


```typescript
import { Hono } from './src/'
import { hc } from './src/client'

const app = new Hono()
    .get('foo/:id', c => {
        if (c.req.param('id') === '1') {
            return c.json({ res: 400 as const }, 400)
        }
        return c.json({ res: 200 as const }, 200)
    })

const client = hc<typeof app>('')

client.foo[':id'].$get({ param: { id: '1' }}).then(async res => {
    switch (res.status) {
        case 200:
            return await res.json()
        case 400:
            return await res.json()
    }
})
```

Second, projects using [@typescript-eslint/switch-exhaustiveness-check](https://typescript-eslint.io/rules/switch-exhaustiveness-check/) will be broken. You need to add `default` case even if you needn't.

As @NamesMT said, a default unknown status type possibly works

NamesMT:
I pushed a sample for the `DefaultResponse` support at NamesMT/hono@9248417,  
I don't know if this is most optimal way though cause `DefaultResponse` was added to many types.

Also, now that we have a `DefaultResponse` type, we could potentially mark this PR's feature (default response routes) as opt-in: NamesMT/hono@2d0bfa7, this would prevent possible breaking of some projects that is currently using `switch` or `if/else` on the `res.status`, as well as prevent spam when user hovers on `res.status`.

`src/_demo.ts` for testing:
```ts
// server
import { HTTPException } from './http-exception'
import { Hono } from './hono'

const app = new Hono()
const flag = {}

const routes = app.get('/foo', (c) => {
  if (flag) {
    throw new HTTPException(500, {
      message: 'Server Error!',
    })
  }
  return c.json({ ok: true }, 200)
})

// client
import { hc } from './client'

type HCDefaultResponse = (
  {
    code: 'unknownErr'
    message: string
    detail: any
  } |
  {
    code: 'validationErr'
    message: string
    detail: string[][]
  }
)

const client = hc<typeof routes, HCDefaultResponse>('/') // Remove `HCDefaultResponse` to test the opt-in of default routes.

async function demo() {
  const res = await client.foo.$get()

  switch (res.status) {
    case 200:
        return await res.json().then(data => data.ok)
    default: {
      const data = await res.json() // Should be typeof `HCDefaultResponse`

      console.error(data.code === 'validationErr'
        ? { message: data.message, errorCount: data.detail.length }
        : { message: data.message })

      throw new Error(data.code)
    }
  }
}
```

yusukebe:
@m-shaka @NamesMT Thank you!

As you said, we should support the default response type.

I like @NamesMT 's implementation. My proposal is to pass the default type generics into `client.foo.$ge()` instead of `hc`:

```ts
const res = await client.foo.$get<DefaultResponseInFoo>()
```

How about this?

It includes diff for styling, but we can implement it with the following:

```diff
diff --git a/src/client/client.ts b/src/client/client.ts
index 13641299..512e7675 100644
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -134,17 +134,16 @@ class ClientRequestImpl {
   }
 }
 
-
 /**
  * @param baseUrl The base URL of the Hono instance.
  * @param options The options for the client.
  * @returns A hono API client instance.
- * 
+ *
  * **Typedoc**: <Hono, DefaultResponse>
- * 
+ *
  * - Hono: The type of the Hono instance.
- * 
- * - DefaultResponse: This is used in cases where the response is unknown,  
+ *
+ * - DefaultResponse: This is used in cases where the response is unknown,
  * e.g. checking for an unknown status code:
  * ```ts
  * // The response of status code 400 is processed by a middleware and not explicitly defined in the router type flow.
@@ -153,7 +152,7 @@ class ClientRequestImpl {
  * }
  * ```
  */
-export const hc = <H extends Hono<any, any, any>, DefaultResponse = {}>(
+export const hc = <H extends Hono<any, any, any>>(
   baseUrl: string,
   options?: ClientRequestOptions
 ) =>
@@ -229,4 +228,4 @@ export const hc = <H extends Hono<any, any, any>, DefaultResponse = {}>(
       return req.fetch(opts.args[0], args)
     }
     return req
-  }, []) as UnionToIntersection<Client<H, DefaultResponse>>
+  }, []) as UnionToIntersection<Client<H>>
diff --git a/src/client/types.ts b/src/client/types.ts
index ef54e1fe..664fb4a5 100644
--- a/src/client/types.ts
+++ b/src/client/types.ts
@@ -25,12 +25,18 @@ export type ClientRequestOptions<T = unknown> = {
       headers: T | (() => T | Promise<T>)
     })
 
-export type ClientRequest<S extends Schema, DefaultResponse = {}> = {
+export type ClientRequest<S extends Schema> = {
   [M in keyof S]: S[M] extends Endpoint & { input: infer R }
     ? R extends object
       ? HasRequiredKeys<R> extends true
-        ? (args: R, options?: ClientRequestOptions) => Promise<ClientResponseOfEndpoint<S[M], DefaultResponse>>
-        : (args?: R, options?: ClientRequestOptions) => Promise<ClientResponseOfEndpoint<S[M], DefaultResponse>>
+        ? <DefaultResponse = {}>(
+            args: R,
+            options?: ClientRequestOptions
+          ) => Promise<ClientResponseOfEndpoint<S[M], DefaultResponse>>
+        : <DefaultResponse = {}>(
+            args?: R,
+            options?: ClientRequestOptions
+          ) => Promise<ClientResponseOfEndpoint<S[M], DefaultResponse>>
       : never
     : never
 } & {
@@ -66,8 +72,8 @@ type ClientResponseOfEndpoint<T extends Endpoint = Endpoint, DefaultResponse = {
   ? F extends 'redirect'
     ? ClientResponse<O, S extends RedirectStatusCode ? S : never, 'redirect'>
     : Equal<DefaultResponse, {}> extends true
-      ? ClientResponse<O, S extends StatusCode ? S : never, F extends ResponseFormat ? F : never>
-      :
+    ? ClientResponse<O, S extends StatusCode ? S : never, F extends ResponseFormat ? F : never>
+    :
         | ClientResponse<O, S extends StatusCode ? S : never, F extends ResponseFormat ? F : never>
         | ClientResponse<
             DefaultResponse,
@@ -158,33 +164,27 @@ export type InferRequestOptionsType<T> = T extends (
   ? NonNullable<R>
   : never
 
-type PathChaining<
-  Path extends string,
-  E extends Schema,
-  DefaultResponse = {}
-> = PathToChain<Path, E, Path, DefaultResponse>
+type PathChaining<Path extends string, E extends Schema> = PathToChain<Path, E, Path>
 
 type PathToChain<
   Path extends string,
   E extends Schema,
-  Original extends string = Path,
-  DefaultResponse = {}
+  Original extends string = Path
 > = Path extends `/${infer P}`
-  ? PathToChain<P, E, Path, DefaultResponse>
+  ? PathToChain<P, E, Path>
   : Path extends `${infer P}/${infer R}`
-  ? { [K in P]: PathToChain<R, E, Original, DefaultResponse> }
+  ? { [K in P]: PathToChain<R, E, Original> }
   : {
       [K in Path extends '' ? 'index' : Path]: ClientRequest<
-        E extends Record<string, unknown> ? E[Original] : never,
-        DefaultResponse
+        E extends Record<string, unknown> ? E[Original] : never
       >
     }
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type Client<T, DefaultResponse = {}> = T extends Hono<any, infer S, any>
+export type Client<T> = T extends Hono<any, infer S, any>
   ? S extends Record<infer K, Schema>
     ? K extends string
-      ? PathChaining<K, S, DefaultResponse>
+      ? PathChaining<K, S>
       : never
     : never
   : never
```

m-shaka:
This is just my opinion, but **server side** should accept the default response type instead of client methods like `$get`, because server side is supposed to know which status code it returns.
It' best that you can know which status you have to handle from the type

However, I have a feeling that adding this functionality to server side requires much more effort than adding to client 🤔 
The default response type on client works like an escape hatch and resolves the pain point you mentioned first so I think it's ok atm

yusukebe:
@m-shaka Thank you for your comment!

@NamesMT What do you think of this API? https://github.com/honojs/hono/pull/3292#issuecomment-2295778823

yusukebe:
Hi @NamesMT 

It looks good! We should consider the naming if we go with `$withDR` or get other names, but I like the idea (I think `withDR` is not bad).

NamesMT:
Hi @yusukebe
Just wanna say that I'm super busy until October so I cannot work on this yet, holiday, trips and presenting a new project that I'm leading at work for September.

yusukebe:
@NamesMT 
 
Okay! Enjoy your time!

PR #3294, opened 2024-08-18
fix(jsx/dom): Fixed a bug that caused Script elements to turn into Style elements.

usualoma:
Script elements with empty props or only src were inserted as style elements by typo.

- [x] Add tests
- [x] Run tests


yusukebe (approved):
LGTM!

yusukebe:
@usualoma 

Thanks! Merging now.

PR #3305, opened 2024-08-21
feat(jsx): improve `type` (MIME) attribute types

ssssota:
embed, link, object, script, source, and style elements

embed
link
object

I did NOT add BaseMime to either source, script or style element. Should I do this?

source
script
style

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/src/jsx/intrinsic-elements.ts
+++ b/src/jsx/intrinsic-elements.ts
@@ -287,7 +287,7 @@ export namespace JSX {
   interface EmbedHTMLAttributes extends HTMLAttributes {
     height?: number | string | undefined
     src?: string | undefined
-    type?: string | undefined
+    type?: StringLiteralUnion<BaseMime> | undefined
     width?: number | string | undefined
   }
 
@@ -495,7 +495,7 @@ export namespace JSX {
     imagesizes?: string | undefined
     referrerpolicy?: HTMLAttributeReferrerPolicy | undefined
     sizes?: string | undefined
-    type?: string | undefined
+    type?: StringLiteralUnion<BaseMime> | undefined
     charSet?: string | undefined
 
     // React 19 compatibility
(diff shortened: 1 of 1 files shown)


ryuapp:
Nice PR. Allow me to leave comments.

In my opinion, wouldn't it be better to provide only practical types in `link` tags?
Since the purpose of this PR is supporting coding for developers, it seems more reasonable to not include types that are never used, such as `application/epub+zip` and `application/gzip`.

Even in `source` tags, it seems more reasonable to use only media resource types(image, audio, picture and video).

ssssota:
@ryuapp 

> In my opinion, wouldn't it be better to provide only practical types in link tags?

Looks good! So how can I choose "practical types"? Just `text/html` and `text/css`?
The `link` element is also used to `rel="preload"` to preload assets like this: _[Including a MIME type](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload#including_a_mime_type)_.

> Even in source tags, it seems more reasonable to use only media resource types(image, audio, picture and video).

Looks good too! I'll try it!

ryuapp:
@ssssota 
> Looks good! So how can I choose "practical types"? Just text/html and text/css?
The link element is also used to rel="preload" to preload assets like this: [Including a MIME type](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload#including_a_mime_type).

`preload` is often only used for font, css and image types, but as you pointed out, it seems that virtually all BaseMime types can be used with link tags. So, I reconsider that it would be fine for us to continue as is.

Also, the type of source tag's type attribute is awesome. Thank you.

yusukebe (commented):
LGTM!

yusukebe:
@ssssota 

Great PR, again and again! Thank you for your contribution. Merging now.

PR #3313, opened 2024-08-22
update HTMLRespond type

rossrobino:
If you pass anything with a `toString` method into `c.html` it will return the result, but the types don't currently reflect this. 

This PR adds a `HasToString` type as an option for the first argument of `c.html`. 

This enables you to do things like this and not have TS errors. Right now this works but TS errors.

```ts
const app = new Hono();

app.get("/number", (c) => c.html(1));

const obj = { toString: () => "hello world" };

app.get("/hello", (c) => c.html(obj));
```

```txt
  Overload 1 of 2, '(html: string | Promise<string>, status?: StatusCode | undefined, headers?: HeaderRecord | undefined): Response | Promise<Response>', gave the following error.
    Argument of type 'number' is not assignable to parameter of type 'string | Promise<string>'.
  Overload 2 of 2, '(html: string | Promise<string>, init?: ResponseInit | undefined): Response | Promise<Response>', gave the following error.
    Argument of type 'number' is not assignable to parameter of type 'string | Promise<string>'.ts(2769)
```

Thanks for the review, I'm not sure if there's a case I'm not considering here. 

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -9,7 +9,7 @@ import type {
   RouterRoute,
   TypedResponse,
 } from './types'
-import { HtmlEscapedCallbackPhase, resolveCallback } from './utils/html'
+import { HtmlEscapedCallbackPhase, HTMLInput, resolveCallback } from './utils/html'
 import type { RedirectStatusCode, StatusCode } from './utils/http-status'
 import type {
   InvalidJSONValue,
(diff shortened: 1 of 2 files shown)


yusukebe:
@rossrobino Thank you for the PR.

Issue #3316, opened 2024-08-23
Return a piped response with compression enabled causes an unexpected error


ceifa:
### What version of Hono are you using?

4.5.8

### What runtime/platform is your app running on?

Node

### What steps can reproduce the bug?

Enable the built-in compression:

```js
app.use(compress());
```

Inside a hono route, do the following:

```js
return await fetch(targetUrl, {
    method: req.method,
    headers,
    body: req.raw.body,
    duplex: 'half',
});
```

### What is the expected behavior?

Response piped correctly.

### What do you see instead?

TypeError: immutable
    at _Headers.delete (node:internal/deps/undici/undici:3303:17)
    at set res [as res] (file:///D:/whitewall/proxy/node_modules/hono/dist/context.js:62:31)
    at compress2 (file:///D:/whitewall/proxy/node_modules/hono/dist/middleware/compress/index.js:12:13)

### Additional information

It works on latest version of 3.x. Stopped in 4.x.

nitedani:
Related lines:

https://github.com/honojs/hono/blob/main/src/context.ts#L477
https://github.com/honojs/hono/blob/main/src/middleware/compress/index.ts#L43-L44

Potential solutions:
1. implement a proxy for modifying response.headers (recreating the response)
2.
```js
    const headers = new Headers(res.headers)
    headers.delete('Content-Length')
    headers.set('Content-Encoding', encoding)
    // Compress the response
    const stream = new CompressionStream(encoding)
    ctx.res = new Response(res.body.pipeThrough(stream), {
      status: res.status,
      statusText: res.statusText,
      headers
    })
```
and rework https://github.com/honojs/hono/blob/main/src/context.ts#L477 to avoid overwriting the headers

yusukebe:
This was closed by https://github.com/honojs/hono/pull/3318. Thanks.

PR #3317, opened 2024-08-23
feat(compress): improve compress middleware

nitedani:
Improvements:

- Content-Length threshold for compression
- Handling of already compressed responses
- Cache-Control: no-transform handling
- Handling of custom Content-Type headers

Fixes:
- https://github.com/honojs/hono/issues/3262

- [x] Add tests
- [x] Run tests
- [x] `bun run format:fix && bun run lint:fix` to format the code
- [x] Add [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code

PR #3318, opened 2024-08-24
feat(context): make fetch Response headers mutable

nitedani:
The added test fails without the PR.

Fixes:
- https://github.com/honojs/hono/issues/3316

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -480,6 +480,18 @@ export class Context<
     }
     this.#res = _res
     this.finalized = true
+
+    // fixes: https://github.com/honojs/hono/issues/3316
+    if (this.#res) {
+      const headersObj = this.#res.headers as unknown as Record<symbol, string>
+      const guardSymbol = Object.getOwnPropertySymbols(headersObj).find(
+        (symbol) => symbol.toString() === 'Symbol(guard)'
+      )
+      // Make the headers mutable again
+      if (guardSymbol && headersObj[guardSymbol] === 'immutable') {
+        headersObj[guardSymbol] = 'response'
+      }
+    }
   }
 
   /**
--- a/src/context.ts
+++ b/src/context.ts
@@ -464,24 +464,12 @@ export class Context<
    */
   set res(_res: Response | undefined) {
     this.#isFresh = false
-    if (_res) {
+    if (this.#res && _res) {
       try {
-        // Check if `_res` headers are immutable.
-        _res.headers.delete('______________________________')
-      } catch (e) {
-        if (e instanceof TypeError && e.message.includes('immutable')) {
-          // `_res` is immutable (probably a response from a fetch API), so retry with a new response.
-          this.res = new Response(_res.body, {
-            headers: _res.headers,
-            status: _res.status,
-          })
-          return
-        }
-      }
-
-      if (this.#res) {
-        this.#res.headers.delete('content-type')
         for (const [k, v] of this.#res.headers.entries()) {
+          if (k === 'content-type') {
+            continue
+          }
           if (k === 'set-cookie') {
             const cookies = this.#res.headers.getSetCookie()
             _res.headers.delete('set-cookie')
@@ -491,6 +479,17 @@ export class Context<
           } else {
             _res.headers.set(k, v)
           }
+        }
+      } catch (e) {
+        if (e instanceof TypeError && e.message.includes('immutable')) {
+          // `_res` is immutable (probably a response from a fetch API), so retry with a new response.
+          this.res = new Response(_res.body, {
+            headers: _res.headers,
+            status: _res.status,
+          })
+          return
+        } else {
+          throw e
         }
       }
     }
--- a/src/context.ts
+++ b/src/context.ts
@@ -464,34 +464,38 @@ export class Context<
    */
   set res(_res: Response | undefined) {
     this.#isFresh = false
-    if (this.#res && _res) {
-      this.#res.headers.delete('content-type')
-      for (const [k, v] of this.#res.headers.entries()) {
-        if (k === 'set-cookie') {
-          const cookies = this.#res.headers.getSetCookie()
-          _res.headers.delete('set-cookie')
-          for (const cookie of cookies) {
-            _res.headers.append('set-cookie', cookie)
+    if (_res) {
+      try {
+        // Check if `_res` headers are immutable.
+        _res.headers.delete('______________________________')
+      } catch (e) {
+        if (e instanceof TypeError && e.message.includes('immutable')) {
+          // `_res` is immutable (probably a response from a fetch API), so retry with a new response.
+          this.res = new Response(_res.body, {
+            headers: _res.headers,
+            status: _res.status,
+          })
+          return
+        }
+      }
+
+      if (this.#res) {
+        for (const [k, v] of this.#res.headers.entries()) {
+          if (k === 'set-cookie') {
+            const cookies = this.#res.headers.getSetCookie()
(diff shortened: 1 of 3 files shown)


usualoma:
Hi @nitedani, Thanks for making the pull request!

Referencing internal symbols is a very interesting implementation, but in the main body of the hono we want to ensure that it is implemented only in runtime-independent code, within the scope of the web standard.

Also, as we want to maintain high performance with simple requests, we usually do not want to include processing for ‘immutable response objects’.

About #3316, I would prefer the following approach.

```diff
diff --git a/src/context.ts b/src/context.ts
index e3dbc3c5..487e1456 100644
--- a/src/context.ts
+++ b/src/context.ts
@@ -465,16 +465,32 @@ export class Context<
   set res(_res: Response | undefined) {
     this.#isFresh = false
     if (this.#res && _res) {
-      this.#res.headers.delete('content-type')
-      for (const [k, v] of this.#res.headers.entries()) {
-        if (k === 'set-cookie') {
-          const cookies = this.#res.headers.getSetCookie()
-          _res.headers.delete('set-cookie')
-          for (const cookie of cookies) {
-            _res.headers.append('set-cookie', cookie)
+      try {
+        for (const [k, v] of this.#res.headers.entries()) {
+          if (k === 'content-type') {
+            continue
           }
+
+          if (k === 'set-cookie') {
+            const cookies = this.#res.headers.getSetCookie()
+            _res.headers.delete('set-cookie')
+            for (const cookie of cookies) {
+              _res.headers.append('set-cookie', cookie)
+            }
+          } else {
+            _res.headers.set(k, v)
+          }
+        }
+      } catch (e) {
+        if (e instanceof TypeError && e.message.includes('immutable')) {
+          // `_res` is immutable (probably a response from a fetch API), so retry with a new response.
+          this.res = new Response(_res.body, {
+            headers: _res.headers,
+            status: _res.status,
+          })
+          return
         } else {
-          _res.headers.set(k, v)
+          throw e
         }
       }
     }
```

src/context.ts
```diff
@@ -464,17 +464,33 @@ export class Context<
...
+
+      if (this.#res) {
+        for (const [k, v] of this.#res.headers.entries()) {
+          if (k === 'set-cookie') {
+            const cookies = this.#res.headers.getSetCookie()
+            _res.headers.delete('set-cookie')
+            for (const cookie of cookies) {
+              _res.headers.append('set-cookie', cookie)
+            }
+          } else if (!_res.headers.has(k)) {
```
nitedani, line 490: Why wasn't this checked before?

nitedani:
> I would prefer the following approach.

That fails the added test ☹️ I think we need to check if the response headers are immutable.
https://github.com/honojs/hono/pull/3318/files#diff-0cd594dddd6a9f2e3e26afebbb92716434437d7001c2416252aa7531e7f2b8d4R468-R480

usualoma:
@nitedani 

Thanks for the answer. I think the test needs to be modified. I don't think ‘`c.res.headers` becomes mutable when `c.res = res`’ is a result that context.ts should guarantee. The result that should be guaranteed is that `c.res = res` does not result in an error'.

```ts
    c.res = res
    c.res.headers.set('X-Custom', 'Message')
    expect(c.res.headers.get('X-Custom')).toBe('Message')
```

What is needed is a test of the following. However, it is debatable whether it is appropriate to use the external service https://jsonplaceholder.typicode.com in unit tests here.

```ts
  it('Should be able to overwrite a fetch reponse with a new response.', async () => {
    c.res = await fetch('https://jsonplaceholder.typicode.com/todos/1')
    c.res = new Response('foo', {
      headers: {
        'X-Custom': 'Message',
      },
    })
    expect(c.res.headers.get('X-Custom')).toBe('Message')
  })

  it('Should be able to overwrite a response with a fetch response.', async () => {
    c.res = new Response('foo', {
      headers: {
        'X-Custom': 'Message',
      },
    })
    c.res = await fetch('https://jsonplaceholder.typicode.com/todos/1')
    expect(c.res.headers.get('X-Custom')).toBe('Message')
  })
```

usualoma:
The following cases are problems with the code itself. It is not possible to use the `res` after `res.text()` in hono (or any other framework) as response, so I think it is not a target to be handled by context.ts.

```ts
const res = fetch(...)
await res.text()
ctx.res = res
```

src/context.ts
usualoma: I think that point is correct. I think the new response headers should be prioritised. And you're right, it would be a spec change and should be discussed in another PR.

nitedani:
> . I don't think ‘c.res.headers becomes mutable when c.res = res’ is a result that context.ts should guarantee.

I agree. Should we close the PR? Maybe we can support middleware like compression for fetch responses some other way?

usualoma:
Hi @nitedani, Thanks for your comment.

The following branch is the one with tests added to the contents of https://github.com/honojs/hono/pull/3318#issuecomment-2308567803.

https://github.com/honojs/hono/compare/main...usualoma:hono:fix/update-context-res

As for #3316, I believe this is how it should be fixed.

It is up to you whether you want to reflect this in this pull request or close this pull request once and make another pull request.

yusukebe:
Hi @nitedani @usualoma 

Sorry for the delayed response. This will be awesome PR! Sometimes, I have trouble with a matter like #3316 when I `fetch` origin content and return it in CDN edges.

The code @usualoma wrote https://github.com/honojs/hono/pull/3318#issuecomment-2308567803 and https://github.com/honojs/hono/pull/3318#issuecomment-2315331269 looks good. Would you like us to go with it?

yusukebe:
Hi @nitedani 

Can you work on this? If you can't, I'll do it.

nitedani:
@yusukebe Please take it.

yusukebe:
Hi @usualoma 

I've updated [the @nitedani 's branch](https://github.com/nitedani/hono/tree/nitedani/immutable-headerse) with applying your patch and pushing it. The test passed on my machine, but the CI is falling. Do you have any idea?

usualoma:
@yusukebe 
OK, mystery solved. Since we test against the branch being merged, we need to take into account the current contents of main.

CleanShot 2024-09-06 at 22 26 16@2x

With the changes in #3317, small data is no longer compressed, so the test needs to be large as well.
https://github.com/honojs/hono/blob/main/src/middleware/compress/index.ts#L36

```diff
diff --git a/runtime_tests/node/index.test.ts b/runtime_tests/node/index.test.ts
index b55512d5..310ec9c7 100644
--- a/runtime_tests/node/index.test.ts
+++ b/runtime_tests/node/index.test.ts
@@ -207,10 +207,11 @@ describe('streamSSE', () => {
 })
 
 describe('compress', async () => {
+  const cssContent = Array.from({ length: 60 }, () => 'body { color: red; }').join('\n')
   const [externalServer, externalPort] = await new Promise<[Server, number]>((resolve) => {
     const externalApp = new Hono()
     externalApp.get('/style.css', (c) =>
-      c.text('body { color: red; }', {
+      c.text(cssContent, {
         headers: {
           'Content-Type': 'text/css',
         },
 
@@ -242,6 +243,6 @@ describe('compress', async () => {
     const res = await request(server).get('/fetch/style.css')
     expect(res.status).toBe(200)
     expect(res.headers['content-encoding']).toBe('gzip')
-    expect(res.text).toBe('body { color: red; }')
+    expect(res.text).toBe(cssContent)
   })
 })

```

usualoma:
@yusukebe Thank you, LGTM!

PR #3326, opened 2024-08-27
chore(test): migrate to vitest v2

yasuaki640:
Migrated to vitest v2.

### Details of changes

- install vitest v2 via bun
- modified the test method to match [the vitest spec changes](https://vitest.dev/guide/migration#empty-task-in-suite-tasks-removed)
- fix tsc error

### The author should do the following, if applicable

- [ ] Add tests
- [x] Run tests
- [x] `bun run format:fix && bun run lint:fix` to format the code
- [ ] Add [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code

PR #3338, opened 2024-08-30
chore(test): replace deprecated vitest type

yasuaki640:
I forgot to replace the deprecated type in #3326 , so I replaced it.

[reference.](https://github.com/vitest-dev/vitest/blob/372a2c26a8e5ac2cfe6db961ed40529f09464e82/packages/vitest/src/public/index.ts#L130-L131)

- [ ] Add tests
- [x] Run tests


yusukebe:
@yasuaki640 

Nice! Thank you.

Issue #3319, opened 2024-08-24
Non-streaming ErrorBoundary in streaming context


Mabi19:
### What is the feature you are proposing?

Currently, `<ErrorBoundary>` will always try to stream content when the current connection is streaming, even without `<Suspense>`. I'd like to be able to disable this behavior.

I came across this issue when trying to use HTMX's SSE functionality. In short, SSE streams aren't like normal HTML streams, and this breaks the ErrorBoundary. Here's an example (run on Deno 1.46.1 on Windows 11):
```tsx
#!/usr/bin/env -S deno run --allow-net

/** @jsxImportSource jsr:@hono/hono@4.5.8/jsx */
/** @jsx react-jsx  */

import { Hono } from "jsr:@hono/hono@4.5.8";
import { html } from "jsr:@hono/hono@4.5.8/html";
import { ErrorBoundary, FC, PropsWithChildren } from "jsr:@hono/hono@4.5.8/jsx";
import { streamSSE } from "jsr:@hono/hono@4.5.8/streaming";

const app = new Hono();

const AsyncContent: FC<PropsWithChildren> = async ({ children }) => {
    await new Promise((resolve) => setTimeout(resolve, 100));
    return <div>{children}</div>;
};

app.get("/", (ctx) => {
    return ctx.html(html`
        <!DOCTYPE html>
        <html lang="en">
        <head>
            <meta charset="UTF-8" />
            <meta name="viewport" content="width=device-width, initial-scale=1.0" />
            <title>Test Document</title>
        </head>
        <body>
            <script>
                const eventSource = new EventSource("/sse");
                eventSource.onmessage = (event) => {
                    const wrapper = document.createElement("div");
                    wrapper.innerHTML = event.data;
                    document.body.appendChild(wrapper);
                }
            </script>
        </body>
        </html>
    `);
});

app.get("/sse", (ctx) => {
    return streamSSE(ctx, async (stream) => {
        // Note: Streaming HTML through SSE isn't documented. But I found that using await html`${content}` works (well, except for this issue)
        await stream.writeSSE({
            data: await html`${<AsyncContent>This is some streamed-in content!</AsyncContent>}`,
        });
        await stream.writeSSE({
            data: await html`${(
                <ErrorBoundary>
                    <AsyncContent>
                        This content is wrapped in an ErrorBoundary! It will never be visible in the browser because the
                        ErrorBoundary sends a placeholder instead as it thinks this is a regular HTML stream.
                    </AsyncContent>
[...]

PR #1913, opened 2024-01-07
POC: feat: enable to write jsx to streaming api

sor4chi:
### Author should do the followings, if applicable

- [x] Add tests
- [x] Run tests
- [x] `yarn denoify` to generate files for Deno

If JSX can be delivered via streamingAPI, as in the tweet below, it would be a nice match with the shadowrootmode of web components and allow components to be changed dynamically from the server without client JS.

https://x.com/shou_study/status/1743996737696764069?s=20

Demo:

<details><summary>Code</summary>
<p>

```tsx
import { Hono } from "hono";
import { stream } from "hono/streaming";

const app = new Hono();

app.get("/", (c) => {
  return stream(c, async (stream) => {
    await stream.write(
      <html lang="en">
        <body>
          <template shadowrootmode="open">
            <slot name="component1">
              <p>Loading...</p>
            </slot>
          </template>
        </body>
      </html>
    );
    await stream.sleep(1000);
    await stream.write(
      <div slot="component1">
        <h1>Hello Hono!</h1>
      </div>
    );
  });
});

export default app;
```

</p>
</details> 

https://github.com/honojs/hono/assets/80559385/c2e494a0-f75b-4274-9749-b680d5f301ef

PR #3344, opened 2024-08-31
feat(helper/streaming): Support Promise<string> or (async) JSX.Element in streamSSE

usualoma:
#3319

This change allows you to use "hono/jsx" to write the following in streamSSE. I don't think there are many use cases, but the changes on the `helper/streaming/sse.ts` side are also small and could be added to the implementation.

```tsx
streamSSE(c, async (stream) => {
  await stream.writeSSE({ data: <div>Hello</div> })
})
```

```tsx
streamSSE(c, async (stream) => {
  await stream.writeSSE({
    data: (
      <ErrorBoundary fallback={<div>Error</div>}>
        <AsyncComponent />
      </ErrorBoundary>
    ),
  })
})
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -844,18 +844,11 @@ export class Context<
     this.#preparedHeaders['content-type'] = 'text/html; charset=UTF-8'
 
     if (typeof html === 'object') {
-      if (!(html instanceof Promise)) {
-        html = (html as string).toString() // HtmlEscapedString object to string
-      }
-      if ((html as string | Promise<string>) instanceof Promise) {
-        return (html as unknown as Promise<string>)
-          .then((html) => resolveCallback(html, HtmlEscapedCallbackPhase.Stringify, false, {}))
-          .then((html) => {
-            return typeof arg === 'number'
-              ? this.newResponse(html, arg, headers)
-              : this.newResponse(html, arg)
-          })
-      }
+      return resolveCallback(html, HtmlEscapedCallbackPhase.Stringify, false, {}).then((html) => {
+        return typeof arg === 'number'
+          ? this.newResponse(html, arg, headers)
+          : this.newResponse(html, arg)
+      })
     }
 
     return typeof arg === 'number'
--- a/src/helper/streaming/sse.test.tsx
+++ b/src/helper/streaming/sse.test.tsx
@@ -1,3 +1,5 @@
+/** @jsxImportSource ../../jsx */
+import { ErrorBoundary } from '../../jsx'
 import { Context } from '../../context'
 import { streamSSE } from '.'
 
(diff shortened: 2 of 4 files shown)


usualoma:
@yusukebe @watany-dev @sor4chi 
What do you think?

yusukebe:
Hey @usualoma 

It is interesting that JSX can be written in SSE's `data` property. With [the HTMX SSE extension](https://github.com/bigskysoftware/htmx-extensions/blob/main/src/sse/README.md), the `data` property will be rendered as HTML. We can write the following code.

```tsx
import { Hono } from 'hono'
import { streamSSE } from 'hono/streaming'

const app = new Hono()

app.get('/', (c) => {
  return c.html(
    <html>
      <head>
        <script src="https://unpkg.com/htmx.org@2.0.2"></script>
        <script src="https://unpkg.com/htmx-ext-sse@2.2.2/sse.js"></script>
      </head>
      <body>
        <div hx-ext="sse" sse-connect="/event-source" sse-swap="EventName"></div>
      </body>
    </html>
  )
})

app.get('/event-source', (c) => {
  return streamSSE(c, async (stream) => {
    while (true) {
      await stream.writeSSE({
        data: (
          <div>
            <b>It's</b> {new Date().toISOString()}
          </div>
        ),
        event: 'EventName'
      })
      await stream.sleep(1000)
    }
  })
})

export default app
```

I like this feature.

sor4chi:
Hello, @yusukebe and @usualoma.

https://github.com/honojs/hono/pull/1913

I think when I proposed something similar in this PR before, the conclusion was that it was not necessary because it could be done by piping with `renderToReadableStream`, but what is the difference between this PR and that one?
I'm sorry I haven't fully understood it :)

Is it because you can't use `renderToReadableStream` with SSE?

usualoma:
@sor4chi  Thank youuu.
Maybe it's time to re-evaluate #1913!

### SSE

> Is it because you can't use renderToReadableStream with SSE?

Yes, the main reason is as you pointed out, SSE cannot simply use `renderToReadableStream` to output. Also, in order to execute the `script` element output by `Suspense`, code like the following comment may also be required, and I believe there are many environments where `renderToReadableStream` is not suitable.

### stream

As for the stream referenced in #1913, it can be simply piped with renderToReadableStream. And in general, script elements output by Suspense are also executed without any additional implementation.

However, there are some APPs that can be written more simply by #1913, and considering the consistency with SSE, I feel that it would be better to reopen #1913 and merge it.



(Sorry, I'm submitting this in the middle of a bit of a children's event, but that's the way I see it)

usualoma:
We should consider whether or not to re-evaluate #1913, but for now, in this pull request, I refactored the duplicated parts in context.ts and sse.ts so that one line of resolveCallback is all that is needed.

I think this refactoring will also simplify the implementation of #1913.

sor4chi:
Thank you for your answer, I see what you mean.
Let's discuss [#1913](https://github.com/honojs/hono/pull/1913) after this has been merged!

In the meantime, I'm all for this idea.

yusukebe (approved):
LGTM!

yusukebe:
@usualoma @sor4chi Thank you!

I'll create a `next` branch for the next new minor release `v4.6.0` later. After that, I'll merg this into it!

PR #3369, opened 2024-09-03
feat(context): added (optional) message argument to notFound() handler

marceloverdijk:
This PR adds an optional message argument (either `string` or `object`) to the `notFound()` handler.

E.g. this allows users to do:

```
app.get('/notfound', (c) => {
  return c.notFound('Customer not found.')
})
```

or even:

```
app.get('/notfound', (c) => {
  return c.notFound({ message: 'Customer not found.' })
})
```

The latter option (when an `object` is provided) will automatically result in a json response.

Of course the default behavior is still applied:

```
app.get('/notfound', (c) => {
  return c.notFound()
})
```

will just return in a text response containing `404 Not Found`.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -899,8 +899,8 @@ export class Context<
    * })
    * ```
    */
-  notFound = (): Response | Promise<Response> => {
+  notFound = (message?: string | object): Response | Promise<Response> => {
     this.#notFoundHandler ??= () => new Response()
-    return this.#notFoundHandler(this)
+    return this.#notFoundHandler(this, message)
   }
 }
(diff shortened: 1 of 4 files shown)


yusukebe:
Hi @marceloverdijk 

I don't like this feature. Because it should have more code for detecting objects like using `typeof`. The` hono-base.ts` should be small. And, the behavior may be unexpected if the user sets the custom not found handler with `app.notFound()`. For example, `message` arg in `c.notFound(message)` will not affect the custom handler.

marceloverdijk:
OK np, do you have other suggestions?

Or is it recommended users should simply use `c.text('my message', 404)` or  `c.json({ .. }, 404)` when a custom test / json response is desired for a `404 / Not Found`

I like the semantics of `c.notFound(..)` (but desire a custom response message.

Or would would allowing somethinglike `c.notFound(c.json({ .. }))` be an option?

PR #3370, opened 2024-09-03
feat(context): added application json constant

marceloverdijk:
This adds the `APPLICATION_JSON = 'application/json; charset=UTF-8'` constant which is used in the `.json()` helper (and now can be re-used).

Note: this is similar as the existing `TEXT_PLAIN = 'text/plain; charset=UTF-8'` constant already present.

- [x] Add tests
- [x] Run tests


yusukebe:
Hi @marceloverdijk 

I have some ideas for treating constants. I'll share it later.

PR #3371, opened 2024-09-03
feat(basic-auth): added custom response message option

marceloverdijk:

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/basic-auth/index.ts
+++ b/src/middleware/basic-auth/index.ts
@@ -9,19 +9,21 @@ import type { MiddlewareHandler } from '../../types'
 import { auth } from '../../utils/basic-auth'
 import { timingSafeEqual } from '../../utils/buffer'
 
+type MessageFunction = (c: Context) => string | object | Promise<string | object>
+
 type BasicAuthOptions =
   | {
       username: string
       password: string
       realm?: string
       hashFunction?: Function
-      invalidUserMessage?: string | object | Function
+      invalidUserMessage?: string | object | MessageFunction
     }
   | {
       verifyUser: (username: string, password: string, c: Context) => boolean | Promise<boolean>
       realm?: string
       hashFunction?: Function
-      invalidUserMessage?: string | object | Function
+      invalidUserMessage?: string | object | MessageFunction
     }
 
 /**
@@ -35,7 +37,7 @@ type BasicAuthOptions =
  * @param {string} [options.realm="Secure Area"] - The realm attribute for the WWW-Authenticate header.
  * @param {Function} [options.hashFunction] - The hash function used for secure comparison.
  * @param {Function} [options.verifyUser] - The function to verify user credentials.
- * @param {string | object | Function} [options.invalidUserMessage="Unauthorized"] - The invalid user message.
+ * @param {string | object | MessageFunction} [options.invalidUserMessage="Unauthorized"] - The invalid user message.
  * @returns {MiddlewareHandler} The middleware handler function.
  * @throws {HTTPException} If neither "username and password" nor "verifyUser" options are provided.
  *
--- a/src/middleware/basic-auth/index.ts
+++ b/src/middleware/basic-auth/index.ts
@@ -15,11 +15,13 @@ type BasicAuthOptions =
       password: string
       realm?: string
       hashFunction?: Function
+      invalidUserMessage?: string | object | Function
     }
   | {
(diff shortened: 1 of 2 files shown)


yusukebe:
@marceloverdijk Thank you for the PR!

Regarding the `invalidUserMessage` function, I think allowing returning `string | object` is not a good design. It might be useful for some users, but this should be more flexible and reduce the `if` and `type of` statement to detect the types.

I think the following is best:

```ts
type CustomMessageFunction = (c: Context) => Response | Promise<Response>
```

Returning `Response` is the most flexible, and the user can fully customize the response. Enable to this, you can change the following:

```diff
diff --git a/src/middleware/basic-auth/index.ts b/src/middleware/basic-auth/index.ts
index b9b9f359..f69cc049 100644
--- a/src/middleware/basic-auth/index.ts
+++ b/src/middleware/basic-auth/index.ts
@@ -9,19 +9,21 @@ import type { MiddlewareHandler } from '../../types'
 import { auth } from '../../utils/basic-auth'
 import { timingSafeEqual } from '../../utils/buffer'

+type CustomMessageFunction = (c: Context) => Response | Promise<Response>
+
 type BasicAuthOptions =
   | {
       username: string
       password: string
       realm?: string
       hashFunction?: Function
-      invalidUserMessage?: string | object | Function
+      invalidUserMessage?: CustomMessageFunction
     }
   | {
       verifyUser: (username: string, password: string, c: Context) => boolean | Promise<boolean>
-      invalidUserMessage?: string | object | Function
+      invalidUserMessage?: CustomMessageFunction
     }

 /**
@@ -73,10 +75,6 @@ export const basicAuth = (
     options.realm = 'Secure Area'
   }

-  if (!options.invalidUserMessage) {
-    options.invalidUserMessage = 'Unauthorized'
-  }
-
   if (usernamePasswordInOptions) {
     users.unshift({ username: options.username, password: options.password })
   }
@@ -104,23 +102,15 @@ export const basicAuth = (
     }
     // Invalid user.
     const status = 401
-    const headers = {
-      'WWW-Authenticate': 'Basic realm="' + options.realm?.replace(/"/g, '\\"') + '"',
+    ctx.status(401)
+    ctx.header('WWW-Authenticate', 'Basic realm="' + options.realm?.replace(/"/g, '\\"') + '"')
+
+    const defaultCustomMessage = (c: Context) => {
+      return c.text('Unauthorized')
     }
-    const responseMessage =
-      typeof options.invalidUserMessage === 'function'
-        ? await options.invalidUserMessage(ctx)
-        : options.invalidUserMessage
-    const res =
-      typeof responseMessage === 'string'
-        ? new Response(responseMessage, { status, headers })
-        : new Response(JSON.stringify(responseMessage), {
-            status,
-            headers: {
-              ...headers,
-              'content-type': 'application/json; charset=UTF-8',
-            },
-          })
+
+    const customResponse = options.invalidUserMessage ?? defaultCustomMessage
+    const res = await customResponse(ctx)
     throw new HTTPException(status, { res })
   }
 }
```

I'd like to hear other's opinions: cc @usualoma

marceloverdijk:
@yusukebe but in the proposed case is there some helper available in Hono to create a json `Response` with the correct `context-type` set?

marceloverdijk:
@yusukebe I created this alternative PR with the changes you suggested: https://github.com/honojs/hono/pull/3380

marceloverdijk:
@yusukebe as we decided to close #3380 , let's discuss here what is needed to get #3371 going again.

src/middleware/basic-auth/index.ts
```diff
@@ -15,11 +15,13 @@ type BasicAuthOptions =
       password: string
       realm?: string
       hashFunction?: Function
+      invalidUserMessage?: string | object | Function
```
yusukebe, line 18: It's better to write the more specific type like this instead of `Function`: 

```ts
type MessageFunction = (c: Context) => string | object | Promise<string | object>
```

yusukebe:
Hi @marceloverdijk 

I think almost all are good. I've left a comment. Check it.

src/middleware/basic-auth/index.ts
marceloverdijk: Yes that's a good idea. PR is updated.

yusukebe (approved):
LGTM!

yusukebe:
Hi @marceloverdijk 

Looks good to me! Let's go with this. I'll merge this into the `next` later.

PR #3380, opened 2024-09-04
feat(basic-auth): added custom response message option (alternative)

marceloverdijk:

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/basic-auth/index.test.ts
+++ b/src/middleware/basic-auth/index.test.ts
@@ -121,20 +121,7 @@ describe('Basic Auth by Middleware', () => {
     return c.text('verify-user')
   })
 
-  app.get('/auth-custom-invalid-user-message-string/*', (c) => {
-    handlerExecuted = true
-    return c.text('auth')
-  })
-  app.get('/auth-custom-invalid-user-message-object/*', (c) => {
-    handlerExecuted = true
-    return c.text('auth')
-  })
-  app.get('/auth-custom-invalid-user-message-function-string/*', (c) => {
-    handlerExecuted = true
-    return c.text('auth')
-  })
-
-  app.get('/auth-custom-invalid-user-message-function-object/*', (c) => {
+  app.get('/auth-custom-invalid-user-message/*', (c) => {
     handlerExecuted = true
     return c.text('auth')
   })
--- a/src/middleware/basic-auth/index.test.ts
+++ b/src/middleware/basic-auth/index.test.ts
@@ -84,38 +84,13 @@ describe('Basic Auth by Middleware', () => {
   })
 
   app.use(
-    '/auth-custom-invalid-user-message-string/*',
+    '/auth-custom-invalid-user-message/*',
     basicAuth({
       username,
       password,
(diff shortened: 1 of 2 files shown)


src/middleware/basic-auth/index.test.ts
```diff
@@ -83,6 +83,17 @@ describe('Basic Auth by Middleware', () => {
     return auth(c, next)
   })
 
+  app.use(
+    '/auth-custom-invalid-user-message/*',
+    basicAuth({
+      username,
+      password,
+      invalidUserMessage: (c) => {
+        return c.json({ message: 'Custom unauthorized message' })
```
yusukebe, line 92: First grance, it seems to be good. However, it may cause unintended behavior because the status code and `WWW-Authenticate` are already set.

For instance. If the user returns the `c.json()` without the status code, the status code will be `401`. Sometimes, it's expected, but sometimes, the user who is expecting a `200` status will be confused.

To prevent it, we make it does not pass the status code and `WWW-Authenticate` header.

```diff
diff --git a/src/middleware/basic-auth/index.ts b/src/middleware/basic-auth/index.ts
index b382a7c6..307ece12 100644
--- a/src/middleware/basic-auth/index.ts
+++ b/src/middleware/basic-auth/index.ts
@@ -102,9 +102,9 @@ export const basicAuth = (
     }
     // Invalid user.
     const status = 401
-    ctx.status(status)
-    ctx.header('WWW-Authenticate', 'Basic realm="' + options.realm?.replace(/"/g, '\\"') + '"')
     const defaultInvalidUserMessage = (c: Context) => {
+      c.status(status)
+      c.header('WWW-Authenticate', 'Basic realm="' + options.realm?.replace(/"/g, '\\"') + '"')
       return c.text('Unauthorized')
     }
     const invalidUserMessage = options.invalidUserMessage ?? defaultInvalidUserMessage
```

It is very flexible, but the user must write `WWW-Authenticate` by themselves, which is not a good point.

yusukebe: @usualoma Any thoughts?

marceloverdijk: Indeed, I expect the `WWW-Authenticate` to be included automatically.
Maybe when customizing the message/responses the user is better of copying the middleware and changing where needed?

yusukebe: @marceloverdijk 

Someone expect it includes `WWW-Authenticate` but someone may not know the header is set or not because it implicit.

I've considered it. I think #3371 is better, though I've suggested this PR's way. With #3371, users will not be confused because that it has a `WWW-Authenticate` header and a `401` status is understandable. So, sorry for taking your time. Would you like us to go with #3371?

marceloverdijk: Hi @yusukebe No reason to be sorry. It's good to experiment with alternatives and for me the original #3371 is absolutely fine. Let's close this one and discuss further in #3371 what is needed to get it going.

yusukebe: Thanks!

PR #3383, opened 2024-09-04
Adding support to absolute path. Tested on Bun.

cj-1010-1414:

- [X] Add tests
- [x] Run tests


yusukebe:
Hi @cj-1010-1414 

It is not a bug that the serve static does not support an absolute path. It's intended. Regarding supporting the absolute path, we are considering how we should support it. Either way, this will affect current user who uses this feature as a breaking change.

cj-1010-1414:
Hope soon Hono support both relative and absolute path

yusukebe:
@cj-1010-1414 

Okay. I'll work on it after releasing the next minor version.

yusukebe:
This feature is implemented in #3420. Thanks!

PR #3398, opened 2024-09-09
fix(secure-headers): optimize getPermissionsPolicyDirectives function

kbkn3:

- [x] Add tests
- [x] Run tests

I am very sorry.
I found a bug (wrong specification) and fixed it.

- When specified with `{delective}: true`, `true` means “allowed at all origins” and should be represented as `*` in the header. However, an empty parenthesis () is printed, implying a restriction.

Please let me know if there is a problem, because I don't know if I should direct the PR to the `next` branch.


yusukebe (approved):
LGTM!

yusukebe:
@kbkn3 

Thanks! No problem. I'll merge it into the `next`.

PR #3401, opened 2024-09-09
chore(path): remove unnecessary slashes

ryuapp:
Fix https://github.com/denoland/deno/issues/25532

This is a Deno compatibility issue. However we can fix it here.
It would be better if it could be bound by rules.

- [ ] Add tests
- [x] Run tests


ryuapp:
Oops. The issue seems to be a problem that cannot be fixed here , so I will temporarily close this PR.
If I have any other ideas, I will open the pr again.

Issue #3377, opened 2024-09-04
monitoring type-check performance on CI


m-shaka:
### What is the feature you are proposing?

Some changes may accidentally damage type-check performance like https://github.com/honojs/hono/pull/3074#issuecomment-2208196163

It would be great if you could know if your change has an impact on type-check performance.

What do you guys think? Please tell me your opinion and possible solutions to this.
A minimum solution that came to me is storing the result of type-check profiling on `main` branch and comparing it with the one of the PR

yusukebe:
Hi @m-shaka 

I also wanted that! First, we must create the benchmark method to measure the type-check performance. I've used this way ago:

https://github.com/microsoft/TypeScript/wiki/Performance#performance-tracing

It will be good to make some script to show the score as a concrete number: like `This took xxx ms`.

m-shaka:
found that [--diagnostics option](https://www.typescriptlang.org/tsconfig/#diagnostics
) of `tsc` is useful. It gives you info like this
```
Files:              247
Lines:           113520
Identifiers:     110226
Symbols:         256205
Types:           214588
Instantiations: 3438524
Memory used:    469468K
I/O read:         0.07s
I/O write:        0.00s
Parse time:       0.45s
Bind time:        0.22s
Check time:       5.11s
Emit time:        0.00s
Total time:       5.79s
```

`Instantiations` and `Check time` are important to Hono.

I'll make POC later

m-shaka:
Yeah, that's useful when you're looking for the bottleneck in compilation

But the first thing we need is to be able to notice the possibility that type checking performance has declined, isn't it?

`--diagnostics` option is easy to use and the result is easy to process via CLI so it's a minimum solution we should take first, I suppose

yusukebe:
> But the first thing we need is to be able to notice the possibility that type checking performance has declined, isn't it?

Yes. Exactly. We have first to measure the performance. `--diagnostics` seems to be good!

yusukebe:
We now have the GitHub action to measure performance with `tsc` with `-diagnostics` introduced in #3406. There may be other ways to do it, but now we can close this issue. Thanks!

PR #3406, opened 2024-09-11
ci: Type check perf

m-shaka:
part of #3377 
- [ ] Add tests
- [ ] Run tests

### detail
The main goal is to compare type check performances between main branch and each PR.

What CI jobs I added do is:
- generate Hono app with 200 routes
- compile a script using the app above and take diagnostics
- on main branch, store the result to GitHub actions cache
- on PR, compare the current performance with the latest stored result

I think it'd be nice to post the comparison as a comment on the PR (after making sure caching is working) 

### question
Should I do something for Codecov?


yusukebe:
Hi @m-shaka !

This is great! It's unsure how this will be effective and how to measure performance accurately, but starting to try using it is important.

One thing about naming. How about creating `perf-measures` and putting this `type-check-perf` as `type-check`? In the future, we may create other scripts for measuring performance like `perf-measures/routers`.

yusukebe:
> Should I do something for Codecov?

I'll investigate it, but it can be done after merging this.

m-shaka:
Thank you for the feedback!
Could you explain more about what kind of files you are planning to place in `type-check` and `perf-measures` respectively?

yusukebe:
> Could you explain more about what kind of files you are planning to place in `type-check` and `perf-measures` respectively?

We may not do that, and we don't have to do that, but I was thinking:

* Router speed
* Req => Res performance
* File/Bundle size like https://github.com/honojs/hono/pull/2975
* Other things in the [benchmarks](https://github.com/honojs/hono/tree/main/benchmarks)

exoego:
We can ignore coverage for type perf code by adding
```
type-check-perf/**/*.ts
```
to
https://github.com/honojs/hono/blob/dfbd717263ab8ecd7bf495fbd2c32d8c79854284/vitest.config.ts#L19-L30

yusukebe:
> We can ignore coverage for type perf code by adding

Yes, we should do it. Thank you for the advice.

m-shaka:
Got it. You just mean `mv type-check-perf perf-measures/type-check`, don't you?

yusukebe:
@m-shaka 

> Got it. You just mean `mv type-check-perf perf-measures/type-check`, don't you?

Yes. Exactly!

m-shaka:
@yusukebe moved the directory and made some refactoring!

perf-measures/type-check/.gitignore
yusukebe, line 1: How about adding a `result.txt`?

m-shaka: Done!

yusukebe (approved):
LGTM!

yusukebe:
@m-shaka 

Thanks! Let's go with this and keep improving it.

PR #3414, opened 2024-09-15
perf(serve-static): performance optimization for precompressed feature

usualoma:
* 6a675c3e79877775c9815023a24f30acdcdeba9e : Replace with faster implementation
* 7f58d8f1aaa142b2f475d0bf0acb00bc2bcaf783 : "Content-Type" is appropriate to set only if the file is found
* 9aeb1cd2522522fc821f98c100f6a2f28ce484bb : As I wrote in https://github.com/honojs/node-server/pull/199#issuecomment-2351498366, if the file is not a compressible mime type, it is better not to find a compressed file

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/utils/compress.ts
+++ b/src/utils/compress.ts
@@ -1,0 +1,10 @@
+/**
+ * @module
+ * Constants for compression.
+ */
+
+/**
+ * Match for compressible content type.
+ */
+export const COMPRESSIBLE_CONTENT_TYPE_REGEX =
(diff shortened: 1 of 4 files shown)


yusukebe (approved):
LGTM!

src/utils/compress.ts
yusukebe, line 1: I'm planning to create `utils/constants.ts` to have all constants in the `honojs/hono` project, which may include this `COMPRESSIBLE_CONTENT_TYPE_REGEX`. But it's not fixed yet. It's okay to create `utils/compress.ts`.

yusukebe:
@usualoma 

Thanks! Perfect. I'll merge it now.

PR #3415, opened 2024-09-16
fix(serve-static): use application/octet-stream if the mime type is not detected

usualoma:
Currently, if the `mimeType` could not be detected, the `Content-Type` is not set, so it falls back to “text/plain”. There may be cases where “text/plain” is acceptable, but I don't think it is appropriate for binary data to be returned as "text/plain", so I think it should be "application/octet-stream”. Even on general web servers, the default is "application/octet-stream" for files that cannot be detected from the file extension.

- [x] Add tests
- [x] Run tests


yusukebe (approved):
LGTM!

yusukebe:
@usualoma 

Thank you. You are right. It's should be `application/octet-stream` Merging.

PR #3425, opened 2024-09-19
feat(request): Make request.ts available throught JSR for frameworks that need to instantiate HonoRequest

Sorikairox:

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/jsr.json
+++ b/jsr.json
@@ -13,6 +13,7 @@
   ],
   "exports": {
     ".": "./src/index.ts",
+    "./request": "./src/request.ts",
     "./types": "./src/types.ts",
     "./hono-base": "./src/hono-base.ts",
     "./tiny": "./src/preset/tiny.ts",


yusukebe:
Hey @Sorikairox 

Thank you for the PR! The `request.ts` is not exported for the npm package, not only for Deno. I want to know the situation if you want to use the `HonoRequest` class. Can you share the use case for using the `HonoRequest` class?

Sorikairox:
@yusukebe 

Thanks for your answer !

The context is for our framework Danet (https://github.com/savory/Danet).

For the past 12 months or so until last week, we were using Hono's deno.land/x/ version, where everything was accessible. 
I decided to migrate to JSR, which is why I noticed that the class was not accessible.

To keep it simple, in order to allow our internal logic and our users to use the same code with either HTTP request or Websocket message, we instantiate a HonoRequest for the latter.

yusukebe (commented):
LGTM!

yusukebe:
@Sorikairox 

Thank you for your explanation! I understood well. I'll merge this into `main` later and include it in the next patch release. In the npm package, the `request.ts` also may be exported.

PR #3432, opened 2024-09-20
feat(jsx/precompile): Normalization and stringification of attribute values as `renderToString`

### LINKED ISSUE #3427: `style=` objects do not work with `jsx: precompile` on Deno

4.5.9

Deno 1.45.5

When using the `"precompile"` JSX option in Deno, passing an object to the `style` attribute does not work correctly. Instead, it always outputs `[object Object]`.

```jsx
return c.render(
  <div style={{ display: 'none' }} />
);
```

With `"jsx": "react-jsx"`:

```html
<div style="display:none"></div>
```

### What do you see instead?

With `"jsx": "precompile"`:

```html
<div style="[object Object]"></div>
```

### Additional information

_No response_

usualoma:
fix #3427

The “precompile” process converts attribute values in the same way as the normal `renderToString` process.
https://github.com/usualoma/hono/blob/fd80df6a0f03876cf6ac2b12b4b7b812c19ca616/src/jsx/base.ts#L180-L227

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/jsx/jsx-runtime.ts
+++ b/src/jsx/jsx-runtime.ts
@@ -9,7 +9,6 @@ export type { JSX } from './jsx-dev-runtime'
 import { html, raw } from '../helper/html'
 import type { HtmlEscapedString, StringBuffer, HtmlEscaped } from '../utils/html'
 import { escapeToBuffer, stringBufferToString } from '../utils/html'
-import { booleanAttributes } from './base'
 import { normalizeIntrinsicElementKey, styleObjectForEach } from './utils'
 
 export { html as jsxTemplate }
--- a/src/jsx/jsx-runtime.ts
+++ b/src/jsx/jsx-runtime.ts
@@ -6,13 +6,51 @@
 export { jsxDEV as jsx, Fragment } from './jsx-dev-runtime'
 export { jsxDEV as jsxs } from './jsx-dev-runtime'
 export type { JSX } from './jsx-dev-runtime'
-
 import { html, raw } from '../helper/html'
-import type { HtmlEscapedString } from '../utils/html'
+import type { HtmlEscapedString, StringBuffer, HtmlEscaped } from '../utils/html'
+import { escapeToBuffer, stringBufferToString } from '../utils/html'
+import { booleanAttributes } from './base'
+import { normalizeIntrinsicElementKey, styleObjectForEach } from './utils'
+
 export { html as jsxTemplate }
+
 export const jsxAttr = (
-  name: string,
-  value: string | Promise<string>
-): HtmlEscapedString | Promise<HtmlEscapedString> =>
-  typeof value === 'string' ? raw(name + '="' + html`${value}` + '"') : html`${name}="${value}"`
+  key: string,
+  v: string | Promise<string> | Record<string, string | number | null | undefined | boolean>
+): HtmlEscapedString | Promise<HtmlEscapedString> => {
+  key = normalizeIntrinsicElementKey(key)
+  const buffer: StringBuffer = [`${key}="`] as StringBuffer
+  if (key === 'style' && typeof v === 'object') {
+    // object to style strings
+    let styleStr = ''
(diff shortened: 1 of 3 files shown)


usualoma:
It'll need some tweaking.

usualoma:
@yusukebe 
Would you please review?

src/jsx/jsx-runtime.ts
```diff
@@ -6,13 +6,45 @@
...
 export { html as jsxTemplate }
+
 export const jsxAttr = (
-  name: string,
-  value: string | Promise<string>
-): HtmlEscapedString | Promise<HtmlEscapedString> =>
-  typeof value === 'string' ? raw(name + '="' + html`${value}` + '"') : html`${name}="${value}"`
+  key: string,
+  v: string | Promise<string> | Record<string, string | number | null | undefined | boolean>
+): HtmlEscapedString | Promise<HtmlEscapedString> => {
```
yusukebe, line 20: Can we create one common function for these lines and https://github.com/usualoma/hono/blob/fd80df6a0f03876cf6ac2b12b4b7b812c19ca616/src/jsx/base.ts#L180-L227 and use it from the both?

Like this, though, I think `objectToAttribute` is not the best name.

```ts
export const objectToAttribute = (key: string, v: any, buffer: any) => {
  if (v === null || v === undefined) {
    return
  }
  if (key === 'style' && typeof v === 'object') {
    // object to style strings
    let styleStr = ''
    styleObjectForEach(v as Record<string, string | number>, (property, value) => {
      if (value != null) {
        styleStr += `${styleStr ? ';' : ''}${property}:${value}`
      }
    })
    escapeToBuffer(styleStr, buffer)
    buffer[0] += '"'
  } else if (typeof v === 'boolean' && booleanAttributes.includes(key)) {
    return buffer
  } else if (typeof v === 'string') {
  // ...
  return buffer
}
```

src/jsx/jsx-runtime.ts
```diff
@@ -6,13 +6,45 @@
 export { jsxDEV as jsx, Fragment } from './jsx-dev-runtime'
 export { jsxDEV as jsxs } from './jsx-dev-runtime'
 export type { JSX } from './jsx-dev-runtime'
-
 import { html, raw } from '../helper/html'
-import type { HtmlEscapedString } from '../utils/html'
+import type { HtmlEscapedString, StringBuffer, HtmlEscaped } from '../utils/html'
+import { escapeToBuffer, stringBufferToString } from '../utils/html'
+import { booleanAttributes } from './base'
```
yusukebe, line 12: This `booleanAttributes` is never used. Could you remove this line?

yusukebe:
Hey @usualoma 

Thank you for the PR. I've left comments!

src/jsx/jsx-runtime.ts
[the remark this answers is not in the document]
usualoma: fixed b337b23

usualoma:
Key normalization is done by deno, so processing by hono was removed.  4405e69

src/jsx/jsx-runtime.ts
usualoma: @yusukebe 
I also considered making it common, but if it is made common, there will inevitably be overhead for both sides in the following areas.

https://github.com/usualoma/hono/blob/4405e69ae8ab419b2dab36dfd25c605a9c50fca3/src/jsx/base.ts#L202
https://github.com/usualoma/hono/blob/4405e69ae8ab419b2dab36dfd25c605a9c50fca3/src/jsx/base.ts#L205

I don't think it's a big overhead, but I want to be particularly conscious of performance when stringifying JSX, so I want to keep it separate for each implementation for optimization. What do you think?

yusukebe: @usualoma 

Understood well. Let's go with this. Thank you for spending for considering it!

yusukebe (approved):
LGTM!

yusukebe:
@usualoma 

Looks good! Thank you.

PR #3438, opened 2024-09-22
ci: post result of perf-measures/type-check

m-shaka:

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -206,11 +206,15 @@ jobs:
           } >> "$GITHUB_ENV"
         working-directory: perf-measures/type-check
       - run: echo "$COMPARISON"
-    # remove the comment out after we make sure that caching is working
-    # - uses: thollander/actions-comment-pull-request@v2
-    #   with:
-    #     comment_tag: perf-measures/type-check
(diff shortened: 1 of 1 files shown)


m-shaka:
@yusukebe I tried to post a comment from a workflow, but I got `Resource not accessible by integration` error
https://github.com/honojs/hono/actions/runs/10980454233/job/30486253901?pr=3438#step:10:47

According to the document of the action, we need to use `pull_request_target` event instead of `pull_request` to post comments to the PR from a forked repository, but I wouldn't like to do that.
https://github.com/thollander/actions-comment-pull-request/tree/v2/?tab=readme-ov-file#permissions

Now we can see the comparison in the workflow detail page, so I think posting it as a comment is not always necessary.
image

Or do you have any better idea?

yusukebe:
@m-shaka 

I also don't want to use the `pull_request_target`.

> Now we can see the comparison in the workflow detail page, so I think posting it as a comment is not always necessary.

Agree. Posting a comment is unnecessary.

Then, if we go without it, should I merge this PR?

m-shaka:
Ok! I'm closing this PR and create another one to remove the comment in ci.yml

PR #3452, opened 2024-09-26
chore(test): cover jsx attributes

exoego:
This PR is going to cover `jsx-runtime.ts` file which is covered 0% now.

https://app.codecov.io/github/honojs/hono/blob/main/src%2Fjsx%2Fjsx-runtime.ts
image

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/jsx/jsx-runtime.test.tsx
+++ b/src/jsx/jsx-runtime.test.tsx
@@ -11,12 +11,16 @@ describe('jsx-runtime', () => {
 
   it('Should render HTML strings', async () => {
     app.get('/', (c) => {
-      return c.html(<h1>Hello</h1>)
+      return c.html(
+        <h1 class='hello' style={{ margin: '1px', padding: '1px' }}>
+          Hello
+        </h1>
+      )
     })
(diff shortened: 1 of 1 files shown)


exoego:
Hmm, I thought this PR can improve coverage on https://app.codecov.io/gh/honojs/hono/pull/3452/blob/src/jsx/jsx-runtime.ts#L43 but no changes ... 🤔

usualoma:
Hi @exoego.

`jsxAttr` is a function used when Deno's `“jsx”: “precompile”` is used, and unit tests are run as follows.
https://github.com/usualoma/hono/blob/ebf9ea769bb638f4a12d553ec3ead3f3204a04e3/runtime-tests/deno-jsx/jsx.test.tsx#L114-L166

It would be good if we could reflect the coverage of Deno's runtime-tests, but as you can see from the following pull request, it's not as straightforward as it seems, and it's not currently being refrected.
https://github.com/honojs/hono/pull/3456

exoego:
Superseded by #3632

PR #3456, opened 2024-09-27
chore(test): merge deno runtime coverage

usualoma:
Checking the results in github actions...

### The author should do the following, if applicable

- [ ] Add tests
- [ ] Run tests
- [ ] `bun run format:fix && bun run lint:fix` to format the code
- [ ] Add [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code

PR #3461, opened 2024-09-28
feat(serve-static): "206 Partial Content" support

### LINKED ISSUE #3324: Support returning 206 Partial Content with ``serveStatic``

Hi!

I'm developing an app that relies on serving images and video files. It's not for a huge amount of people so serving static files from the server is enough. However when serving a video file and returning it with the status ``200 OK`` the browser will not allow seeking (or at least chromium based browsers, haven't really tested others). 

I guess that serveStatic isn't meant to be used for media because you would typically use something like an S3 bucket to serve files but for development purposes saving files to the development machine is easier than setting up a data storage bucket. That being said returning the files (or at least media, could be specified by extension or content type) with 206 Partial Content would improve the development experience of hono.

exoego:
Closes #3324 

This implements [206 Partial Content](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/206)(also called as range requests) support in `serve-static` middleware.
It will be useful specially for seeking a video file. 

`serve-static` base implementation now accepts an optional function:
```typescript
partialContentSupport?: (path: string) => Promise<{
      size: number
      getPartialContent: (start: number, end: number) => PartialContent
      close: () => void
}>
```
If `partialContentSupport` is provided and HTTP header includes valid `Range`, `serve-static` returns  subsets of content.
`getPartialContent` function encapsulates a file handle to avoid opening and closing file multiple times.
`close` function is invoked when all ranges are read, to teardown the file handle.

### 416 Range Not Satisfiable
The implementation also returns [416 Range Not Satisfiable](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/416) if `Range` headers contains many range requests (more than 10), that is potentially DoS (Denial of Service) attack.
This limitation (`10`) is just a clueless number.
The number may be changed when we understand it is too small/too big.

### Not included
- Cache-control via `If-Range`, `ETag` and so on. I think this feature is useful without having cache-control, which needs a lot work.
- Node.js implementation. Because Node.js does not share `serverStatic` implementation 😭 It is solely implemented here: https://github.com/honojs/node-server/blob/74e86a28f375e5acd52e342519ff2b1110a95c16/src/serve-static.ts#L59 
- Deno coverage is 0%: #3456 will fix it

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -175,11 +175,11 @@ export class HonoRequest<P extends string = '/', I extends Input['out'] = {}> {
   header(): Record<string, string>
   header(name?: string) {
     if (name) {
-      return this.raw.headers.get(name.toLowerCase()) ?? undefined
+      return this.raw.headers?.get(name.toLowerCase()) ?? undefined
     }
 
     const headerData: Record<string, string | undefined> = {}
-    this.raw.headers.forEach((value, key) => {
+    this.raw.headers?.forEach((value, key) => {
       headerData[key] = value
     })
     return headerData
--- a/runtime-tests/bun/index.test.tsx
+++ b/runtime-tests/bun/index.test.tsx
@@ -167,6 +167,77 @@ describe('Serve Static Middleware', () => {
     expect(await res.text()).toBe('Bun!')
     expect(onNotFound).not.toHaveBeenCalled()
   })
+
+  describe('Range Request', () => {
+    it('Should support a single range request', async () => {
+      const res = await app.request(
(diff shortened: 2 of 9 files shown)


src/request.ts
exoego, line 1: Note) `?` is added because when `serve-static` checks `Range` header, `headers` is `undefined` on some existing test cases.

yusukebe:
Hi @exoego 

I'm sorry I can't review it yet. Still a little bit busy; please wait.

yusukebe:
@exoego 

I'm sorry for being super late. I have an idea, and I'll share it with you soon.

PR #3474, opened 2024-10-01
fix(types): add key to IntrinsicAttributes

codehz:
make `arr.map(x => <CustomComponent key={x.id} />)` not emit type error

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/jsx/base.ts
+++ b/src/jsx/base.ts
@@ -30,6 +30,9 @@ export namespace JSX {
   export interface IntrinsicElements extends IntrinsicElementsDefined {
     [tagName: string]: Props
   }
+  export interface IntrinsicAttributes {
+      key?: string | number | bigint | null | undefined;
+  }
 }
 
 let nameSpaceContext: Context<string> | undefined = undefined
--- a/src/jsx/base.ts
+++ b/src/jsx/base.ts
@@ -31,7 +31,7 @@ export namespace JSX {
     [tagName: string]: Props
   }
   export interface IntrinsicAttributes {
-      key?: string | number | bigint | null | undefined;
+    key?: string | number | bigint | null | undefined
   }
 }
 


yusukebe:
Hi @codehz 

Thank you for the PR.

I just wanted to confirm it with you. Currently, it does not throw a type error on my machine. Is this what you're pointing out?

CleanShot 2024-10-01 at 18 21 59@2x

codehz:
> Hi @codehz
> 
> Thank you for the PR.
> 
> I just wanted to confirm it with you. Currently, it does not throw a type error on my machine. Is this what you're pointing out?
> 
> CleanShot 2024-10-01 at 18 21 59@2x

while in my machine
image

and the Window is defined as
```ts
function Window({
  id,
  title,
  state,
  zIndex,
}: {
  id: WindowId;
  title: WindowTitleState;
  state: WindowState;
  zIndex: number;
}) {...
}
```

codehz:
```tsx
/** @jsxImportSource hono/jsx */

import {render} from 'hono/jsx/dom'

render(<Test />, document.body)

function Test() {
    const arr = [1]
    return <div>{arr.map(x => <CustomComponent key={x} val...
```

[Playground Link](https://www.typescriptlang.org/play/?target=99&jsx=4&ts=5.6.2#code/PQKhAIAECsGcA8CSBbADgewE4BcDK6BXTAYwFNwALdAO3WDnnBGACgWBLNLbcAb01LUAJqUwBfcADNM6ZOADkVWvQTAhs+WwHDRACgA8AFVKwewAHwAacOuIFkg7ADoARuiEBPAJRtJBasTY7DTgxqa6Xnws4DHgxDSm4ACGmJjgALzgANoAjAC60bEC2ETU4PpC7ABu5rwpmE7ISai6jOnm5QDCBKaynbIY1I7gANakHum88BJVSQA2BKST0+AWXmL6atXmLGK+-oHBZd29yP1cQ9TYurzgswvkYgBct-eLT+DU9i6i4GKRvEKMWKpXKlRqvDepA2WxquyAA)

yusukebe:
Thanks @codehz !

Indeed, if we treat hono/jsx as the same as React, we should add the `key` attribute.

Initially, hono/jsx was developed as non-React JSX, which does not support `key`. If we follow that policy, we can't accept this PR. However, currently, we are saying "hono/jsx has (almost) compatibility for React." So, it may be okay to add the `key` attribute.

@usualoma, what do you think of this?

usualoma:
Hi @codehz. Thank you!

@yusukebe Compatibility with React is important now, and this addition does not affect the function for stringification, so I agree with the merge!

yusukebe (approved):
LGTM!

yusukebe:
@usualoma Thank you for the comment.

@codehz Looks good. I'll merge this now. Thank you.

PR #3477, opened 2024-10-01
fix(serve-static): support Windows

### LINKED ISSUE #3475: serveStatic root is incorrect on Windows + Bun for absolute path  (other interpreters probably too)

4.6.3

Bun

```typescript
app.use("/static/*", 
    serveStatic({ 
        root: __dirname,
        onNotFound(path, c) {
            console.log(path)
        }
    }));
```

Server serve static file from __dirname +"/static"

### What do you see instead?

**Windows:** Server always returns "resource not found". console.log in the onNotFound handler shows path starting from "./C:/Users/..."

**Linux:** Working as it should

### Additional information

That is probably is the reason of the problem
image

I think serveStatic should always expect absolute path or not modify the path. To make it safe just avoid ".." in the file name.

Also "static" (or probably the url prefix used during the middleware registration) is always added to the root path when searching for the file. It seams to be strange too.

### LINKED ISSUE #3693: root option for serveStatic on Windows

4.6.11

Deno

I'm using this middleware in this project to serve static files from a directory specified with the `root` option: https://github.com/lumeland/cms/blob/main/adapters/lume.ts#L124-L129

The `site.dest()` returns the absolute path of the root folder. In Linux/Mac it has the format `/foo/bar`. In Windows is `C:/foo/bar` (note the `/`separator instead of `\`, since Windows supports both separators (I think it's not relevant here, but just in case).

This works fine in Linux and Mac but not in Windows (issue: https://github.com/lumeland/cms/issues/34)

I expect the root variable interpreted as an absolute path to the base folder.
Inspecting [the code here](https://github.com/honojs/hono/blob/main/src/middleware/serve-static/index.ts#L46-L51) looks like it's not always treated as absolute path, only if the path starts with `/` (which is not compatible with Windows).

### What do you see instead?

404 errors for all requests

### Additional information

This is a breaking change introduced in Hono in the version 4.6.3 (this is the PR with more info: https://github.com/honojs/hono/pull/3420)
Previously, the relative path was passed but this stopped working after upgrading Hono so I had to change it (commit here: https://github.com/lumeland/cms/compare/v0.6.4...v0.6.5)

yusukebe:
Fixes #3475
Fixes #3693

- [x] Add tests
- [x] Run tests


yusukebe:
Hi @oscarotero @dmytro-peleshenko

Sorry for being late! This RP may solve your problem. If you have time, can you try this branch to confirm it works well? You can install `hono` of this branch with the following command:

```
npm i https://pkg.pr.new/hono@3477
```

@usualoma (@ryuapp) I think this is okay. Can you review this?

yusukebe:
I think we can merge this. If you have any problems, please feel free to open an issue.

PR #3491, opened 2024-10-05
ci: Display performance measurement results as custom metrics

k2tzumi:

- [ ] Add tests
- [ ] Run tests

### What's this all about?

This will be an improvement in type checking performance monitoring added by the following Pull Request.
https://github.com/honojs/hono/pull/3406

In conjunction with octocov, type checking performance will be displayed as Pull Request comments and CI summary as follows

* [Pull Request  comments](https://github.com/k2tzumi/hono/pull/2#issuecomment-2484429361)
  image
  ~~note: The bundle size is not shown as a diff since it is the first time, but it will be shown after the main branch is updated.~~
  Looks like the diff showed up just fine!

* [CI summary](https://github.com/k2tzumi/hono/actions/runs/11903458839?pr=2#summary-33170471302)
  image


perf-measures/type-check/.octocov.perf-measures.yml
```diff
@@ -0,0 +1,15 @@
...
+  if: false
+codeToTestRatio:
+  if: false
+testExecutionTime:
+  if: false
+diff:
+  datastores:
+    - artifact://${GITHUB_REPOSITORY}
+comment:
+  if: is_pull_request
```
k2tzumi, line 13: Delete this line if you do not need it to appear as a comment in every Pull Request.
You can also check the summary only.

yusukebe:
Hi @k2tzumi 

Sorry for the late reply. This is great! But please wait a bit for the discussion in PR https://github.com/honojs/hono/pull/3443 to be finished.

k2tzumi:
@yusukebe 

There is a discussion going on for #3443, is there anything we should address in our PR here?

perf-measures/type-check/.octocov.perf-measures.yml
yusukebe: How do you think we should enable or disable it? It's cool that the performance measuring result is posted to PR as a comment, but currently, it is only a TypeScript type performance. Many PRs are not related to the type definition, so it may not be important for them.

CC: @m-shaka

m-shaka: People often ignore such an automated comment, so it may not be a problem

k2tzumi: We believe that the comment display itself should always be displayed when measuring performance. Displaying comments makes people aware of performance.

> Many PRs are not related to the type definition, so it may not be important for them.

I thought that the essential answer to this question is that we should consider not nudging the `perf-measures-type-check-on-pr` Job for PRs that do not affect TypeScript type performance.

It is possible to specify a path base for the execution condition of a GHA job, but is the following the only files that affect TypeScript type performance?

* src/client/*.ts
* perf-measures/**/*
If the source itself that measures performance is changed, it is also necessary to run the job

yusukebe: @m-shaka @k2tzumi Thank you for the comment!

> We believe that the comment display itself should always be displayed when measuring performance. Displaying comments makes people aware of performance.

I see!

> It is possible to specify a path base for the execution condition of a GHA job, but is the following the only files that affect TypeScript type performance?

We have to add the following files that affect application type definitions:

* `src/types.ts`
* `src/hono-base.ts`
* `src/request.ts`
* `src/context.ts`
* `src/validator/validator.ts` - it can be used in the RPC-mode.

@m-shaka 

What do you think of this?

k2tzumi:
If this [PR](https://github.com/honojs/hono/pull/3661) is taken in, additional action may be needed.

EdamAme-x:
Hi @k2tzumi 
Thanks for nice work, 
as you said, if https://github.com/honojs/hono/pull/3661 is merged, I think we should add custom metrics.

perf-measures/type-check/.octocov.perf-measures.yml
k2tzumi: Is it possible that the list of files affecting the application type definition changes frequently? Also, is it difficult to identify these files?

Currently, the performance measurement task takes a sufficiently short time to complete. Therefore, it may be beneficial to run it every time to facilitate maintenance.

k2tzumi:
@EdamAme-x 
Thanks for the comment! Once #3661 is merged, I’d like to add the bundle size to the custom metrics!

flipvh:
Thank you all, these are nice improvements! Type performance is the only thing I mildly warn about when recommending hono to everyone! I love seeing how hono and its ecosystem is improving every week and hoping to contribute back with our own OS projects. Keep up the great work.

perf-measures/type-check/.octocov.perf-measures.yml
k2tzumi: If the bundle size of https://github.com/honojs/hono/pull/3661 is also subject to comparison output, it might be simpler to run the job each time, but what do you think?

yusukebe: > If the bundle size of https://github.com/honojs/hono/pull/3661 is also subject to comparison output, it might be simpler to run the job each time, but what do you think?

Agree! Let's make it simple by running the job each time.

yusukebe:
@k2tzumi 

I merged #3661!

k2tzumi:
I'm starting to have the desire to clean up the commit log.

perf-measures/bundle-check/scripts/process-results.ts
```diff
@@ -1,14 +1,23 @@
...
-    const currentResult = (await fs.readFile('./generated/after.js')).byteLength
-    let previousResult: number | null = null
-    try {
-        previousResult = (await fs.readFile('./generated/before.js')).byteLength
-    } catch (e) {}
-    const table = ['| | Current | Previous |', '| --- | --- | --- |']
-    table.push(`| Bundle Size | ${currentResult} | ${previousResult || 'N/A'} |`)
-    console.log(table.join('\n'))
+    const args = process.argv.slice(2);
+    const filePath = args[0];
```
k2tzumi, line 5: Allowing file paths to be specified seems to be more versatile.
Or, it would be easier to understand the connection to build in this source.

```ts
import { build, Format } from 'esbuild';

const buildOptions = {
  entryPoints: ['dist/index.js'],
  bundle: true,
  minify: true,
  format: 'esm' as Format,
  outfile: 'perf-measures/bundle-check/generated/bundle.js'
};

build(buildOptions).then(() => {
  console.log('Build succeeded.');
  // process result
}).catch(() => {
  console.error('Build failed.');
});

```

perf-measures/bundle-check/scripts/process-results.ts
```diff
@@ -1,14 +1,23 @@
...
+
+    const bundleSize = fs.statSync(filePath).size
+    const metrics = []
+    metrics.push({
+        key: 'bundle-size',
+        name: 'Bundle Size',
+        value: bundleSize,
+        unit: 'B'
```
k2tzumi, line 13: The unit of byte is not converted and is fixed as Byte.
We thought that the number of digits would not change that much, and it would be easier to check the difference without rounding off the units.

.github/workflows/ci.yml
```diff
@@ -181,88 +181,55 @@ jobs:
...
+          bun run build
+          bunx esbuild --minify --bundle dist/index.js --format=esm --outfile=perf-measures/bundle-check/generated/bundle.js
+          bun perf-measures/bundle-check/scripts/process-results.ts perf-measures/bundle-check/generated/bundle.js > perf-measures/bundle-check/size.json
+      - name: Run octocov
+        uses: k1LoW/octocov-action@v1
+        with:
+          config: perf-measures/.octocov.perf-measures.yml
+        env:
+          OCTOCOV_CUSTOM_METRICS_BENCHMARK_BUNDLE: perf-measures/bundle-check/size.json
+          OCTOCOV_CUSTOM_METRICS_BENCHMARK_TYPE: perf-measures/type-check/diagnostics.json
```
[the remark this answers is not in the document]
EdamAme-x: > The order of description changes the order of output in PullRequest comments. Should the order be bundle -> diagnostics?

I can't set priorities, but I think that's fine.

perf-measures/bundle-check/scripts/process-results.ts
```diff
@@ -1,14 +1,23 @@
...
+    })
+    const diagnostics = {
+        key: "bundle-check",
+        name: "Bundle check",
```
k2tzumi, line 17: This will be the title that appears in the comments
Is it an appropriate title? I'm not sure, so please follow up.

k2tzumi (commented):
@yusukebe @EdamAme-x 
Hi, I added the bundle size support. Can you take a look?

perf-measures/bundle-check/scripts/process-results.ts
k2tzumi: I don't understand the build pipeline, are the build options specified the same as those in the build pipeline? Just curious.

perf-measures/bundle-check/scripts/process-results.ts
EdamAme-x: Proposal `Bundle Size Check`

perf-measures/bundle-check/scripts/process-results.ts
k2tzumi: I don't know if there is a need, but we could try to find out the size for each of the major build options.

yusukebe: @k2tzumi 

For the `hono` package, distributed files are not minified. We assume the bundle size when a user minifies an application using `hono`. For example, `wrangler deploy --minify`. So, the options depend on the user. Using esbuild and the options you specified are generic, so I think that's fine.

yusukebe: > Or, it would be easier to understand the connection to build in this source.v

Does this mean we can remove this line if you add the build command in `perf-measures/bundle-check/scripts/process-results.ts`? If so, it's good.

https://github.com/honojs/hono/pull/3491/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR226

yusukebe:
Hi @k2tzumi 

I realized that the codes in `perf-measures` are not formatted. I've created the PR and merged it into the main:

https://github.com/honojs/hono/pull/3683

Can you apply this change?

perf-measures/bundle-check/scripts/process-results.ts
k2tzumi: > For the hono package, distributed files are not minified.

Would you also like to know the size of the bundled sources that are not minified?
Is it sufficient to check the size of the minified source only as it is now?

> Does this mean we can remove this line if you add the build command in perf-measures/bundle-check/scripts/process-results.ts? If so, it's good.

Yes.
We can port a BUILD in a separate command to a script.
We believe that it is better to do the size comparison with multiple BuildOptions, but on the script side.

yusukebe: > Would you also like to know the size of the bundled sources that are not minified?

No. We only have to know minified file sizes because we focus on them when they are minified.

> Yes.
> We can port a BUILD in a separate command to a script.
> We believe that it is better to do the size comparison with multiple BuildOptions, but on the script side.

I see! It makes sense!

perf-measures/bundle-check/scripts/check-bundle-size.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  try {
+    await esbuild.build({
+      entryPoints: ['dist/index.js'],
+      bundle: true,
+      minify: true,
+      format: 'esm' as esbuild.Format,
+      target: 'es2022',
+      outfile: tempFilePath,
+    })
```
k2tzumi, line 18: I have already confirmed that the results are the same as the CLI run, but is this build option appropriate?

k2tzumi (commented):
Hi. @yusukebe 

Merged #3683. Some minor tweaks were made along with.

perf-measures/bundle-check/scripts/check-bundle-size.ts
yusukebe: Yes. This is good.

yusukebe:
@k2tzumi 

Sorry! I forgot to add a linter for the `perf-measures` directory. I've added it to this PR: https://github.com/honojs/hono/pull/3686

Could you update this PR?

yusukebe:
Hi @EdamAme-x 

Also, can you review this?

k2tzumi:
I was not able to verify the execution of jobs in `perf-measures-type-and-bundle-check-on-main`, so I tried it and found several problems, so I fixed them.
I've reflected the results in the Issue comment, so please check it out.

yusukebe (approved):
LGTM!

EdamAme-x (approved):
Great work!
(I had a Ureteral stone that knocked me down, so my review was late confirming it. My apologies lol.)

It would be a good idea to show both B and KB on custom metrics (but I don't think there's any rush to implement it now).
At any rate, let's try it out and see if there is any inconvenience in actually using it.

yusukebe:
@k2tzumi @EdamAme-x

Thank you! Let's land it!

yusukebe:
Wooow, super cool!
https://github.com/honojs/hono/pull/3666#issuecomment-2484697912

PR #3507, opened 2024-10-11
fix(types): rm ExcludeEmptyObject to fix massively increased type instantiations

m-shaka:

- [ ] Add tests
- [ ] Run tests

It's super weird, but if `ExcludeEmptyObject` is outside of `hono-base.ts`, "Type instantiation is excessively deep and possibly infinite." error occurs.
Please use this script to check
https://github.com/honojs/hono/pull/3443#issuecomment-2380551431

I don't know the cause at all, it may be a kind of cache issue, though

### DIFF
--- a/src/client/types.ts
+++ b/src/client/types.ts
@@ -1,4 +1,5 @@
 import type { Hono } from '../hono'
+import type { HonoBase } from '../hono-base'
 import type { Endpoint, ResponseFormat, Schema } from '../types'
 import type { StatusCode, SuccessStatusCode } from '../utils/http-status'
 import type { HasRequiredKeys } from '../utils/types'
@@ -159,7 +160,7 @@ type PathToChain<
     }
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type Client<T> = T extends Hono<any, infer S, any>
+export type Client<T> = T extends HonoBase<any, infer S, any>
   ? S extends Record<infer K, Schema>
     ? K extends string
       ? PathToChain<K, S>
--- a/perf-measures/type-check/client.ts
+++ b/perf-measures/type-check/client.ts
@@ -1,4 +1,4 @@
-import type { app } from './generated/app'
 import { hc } from '../../src/client'
+import type { app } from './generated/app'
 
 const client = hc<typeof app>('/')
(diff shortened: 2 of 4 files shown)


src/client/types.ts
```diff
@@ -159,7 +160,7 @@ type PathToChain<
     }
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type Client<T> = T extends Hono<any, infer S, any>
+export type Client<T> = T extends HonoBase<any, infer S, any>
```
m-shaka, line 163: Weird again.

The reason why I added `ExcludeEmptyObject` is that schema type with empty object type broke RPC type.
For example, `T` in code below is `never` without `ExcludeEmptyObject`.
```typescript
const app = new Hono().route('/', new Hono().get('/foo', c => c.json({ ok: true })))
type T = Client<typeof app>
```

But, replacing `Hono` with `HonoBase` in `Client` fixed the problem

m-shaka: 🤔 
```typescript
const app = new Hono()
  .route('/', new Hono().get('/foo', c => c.json({ ok: true })))
type Extract1<T> = T extends Hono<any, infer S, any> ? S : never
type Extract2<T> = T extends HonoBase<any, infer S, any> ? S : never
type R1 = Extract1<typeof app> // {}
type R2 = Simplify<Extract2<typeof app>> // {} | { '/foo': ... }
```

yusukebe: @m-shaka 

Is that behavior related to the following weird stuff?:

CleanShot 2024-10-14 at 18 35 10@2x

CleanShot 2024-10-14 at 18 36 12@2x

m-shaka: I think that's a matter of display.
The return type of `route` is `HonoBase` and it's correct, but in `src/hono-base.ts` its name is `Hono`. If you rename class `Hono` with `HonoBase` in `src/hono-base.ts`, you see always `HonoBase` in the widget. Presumably, the compiler is kindly renaming it to distinguish `Hono` and `HonoBase`.

image

yusukebe: @m-shaka 

Ah, understand. Thanks!

yusukebe (approved):
LGTM!

yusukebe:
@m-shaka 

Okay! Let's go with it. Thank you.

PR #3510, opened 2024-10-12
fix(cors): avoid setting `Access-Control-Allow-Origin` if there is no matching origin

uki00a:

- [x] Add tests
- [x] Run tests

Even if the `Origin` header of a request does not match a origin specified by `origin` option, CORS middleware always sets `Access-Control-Allow-Origin` header. I don't see this as a serious problem, but in some cases it may unintentionally expose a server setting.

### References

- https://github.com/expressjs/cors/blob/v2.8.5/test/test.js#L251-L262 - `cors` package seems not to set the `Access-Control-Allow-Origin` if the `Origin` header of a request does not match allowed origins
- https://fetch.spec.whatwg.org/#cors-check - A CORS check returns failure if the `Access-Control-Allow-Origin` header is missing
  - [Currently, CORS middleware returns a origin specified by `origin` option as the `Access-Control-Allow-Origin` if the option does not match the `Origin` header of a request](https://github.com/honojs/hono/blob/v4.6.4/src/middleware/cors/index.ts#L70-L76), so changes in this PR would have no particular negative impact

### DIFF
--- a/src/middleware/cors/index.test.ts
+++ b/src/middleware/cors/index.test.ts
@@ -40,6 +40,12 @@ describe('CORS by Middleware', () => {
       origin: 'http://example.com',
     })
   )
+  app.use(
+    '/api6/*',
+    cors({
+      origin: 'http://example.com',
+    })
+  )
 
   app.get('/api/abc', (c) => {
     return c.json({ success: true })
@@ -83,7 +89,10 @@ describe('CORS by Middleware', () => {
   })
 
   it('Preflight with options', async () => {
-    const req = new Request('https://localhost/api2/abc', { method: 'OPTIONS' })
+    const req = new Request('https://localhost/api2/abc', {
+      method: 'OPTIONS',
+      headers: { origin: 'http://example.com' },
+    })
     const res = await app.request(req)
 
     expect(res.headers.get('Access-Control-Allow-Origin')).toBe('http://example.com')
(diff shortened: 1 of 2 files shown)


src/middleware/cors/index.test.ts
```diff
@@ -34,6 +34,12 @@ describe('CORS by Middleware', () => {
 
   app.use('/api5/*', cors())
```
uki00a, line 36: This was added in #1129, but seems to have been lost at some point :eyes:

yusukebe:
Hi @uki00a 

Thanks! I've left one comment. Please check it!

src/middleware/cors/index.test.ts
uki00a: Thanks for the review! :pray:

I believe the addition of the following code in #1129 is to verify that duplicate headers are not set:

```typescript
  app.use(
    '/api6/*',
    cors({
      origin: 'http://example.com',
    })
  )
  app.use(
    '/api6/*',
    cors({
      origin: 'http://example.com',
    })
  )
```

If I remove one of them, should I also remove the following test?

https://github.com/honojs/hono/blob/331166471c499e5a8f432fcb77936c3d4f699cfe/src/middleware/powered-by/index.test.ts#L21-L26

yusukebe: @uki00a 

Got it! Let's leave these lines. Thank you for the explanation!

yusukebe (approved):
LGTM!

yusukebe:
@uki00a 

Looks good! I'll merge it and release the new version soon (maybe tomorrow). Thank you for your contribution!

PR #3514, opened 2024-10-14
fix(build): remove private fields

### LINKED ISSUE #2955: Weird `#private` property in some classes

4.4.2

Node

I've encountered a weird bug when using Hono in a monorepo with client and server.

```ts
// client
import { type IAppType } from "../../../../../server/api";

const client = hc<IAppType>("http://localhost:3000");
```
On the `hc<IAppType>` it gives the error:
```
Type 'Hono<IHonoEnv, { "/ping": { $get: { input: {}; output: "pong"; outputFormat: "text"; status: 100 | 101 | 102 | 103 | 200 | 201 | 202 | 203 | 204 | 205 | 206 | 207 | 208 | 226 | 300 | 301 | ... 47 more ... | -1; }; }; } & { ...; } & { ...; } & { ...; }, "/">' does not satisfy the constraint 'Hono<any, any, any>'.
  Property '#private' is missing in type 'Hono<IHonoEnv, { "/ping": { $get: { input: {}; output: "pong"; outputFormat: "text"; status: 100 | 101 | 102 | 103 | 200 | 201 | 202 | 203 | 204 | 205 | 206 | 207 | 208 | 226 | 300 | 301 | ... 47 more ... | -1; }; }; } & { ...; } & { ...; } & { ...; }, "/">' but required in type 'Hono<any, any, any>'.ts(2344)
hono-base.d.ts(65, 5): '#private' is declared here.
```

I investigated more, checking both the "Hono repo (source)", and the "hono package from NPM (build)", it seems the word `#private` is NOT in the source, but is there in the build.
In the build it can be found in these files:
```
3 results - 3 files

dist\types\context.d.ts:
  159  export declare class Context<E extends Env = any, P extends string = any, I extends Input = {}> {
  160:     #private;
  161      /**

dist\types\hono-base.d.ts:
  64  declare class Hono<E extends Env = Env, S extends Schema = {}, BasePath extends string = '/'> {
  65:     #private;
  66      get: HandlerInterface<E, 'get', S, BasePath>;

dist\types\request.d.ts:
  15  export declare class HonoRequest<P extends string = '/', I extends Input['out'] = {}> {
  16:     #private;
  17      /**
```

I've also read the discussion https://github.com/orgs/honojs/discussions/2215 but, not surprisingly, moving it to devDependencies didn't work.

_No response_

### What do you see instead?

_No response_

### Additional information

_No response_

nakasyou:
Fixes #2955

To resolve it, build scripts remove private fields from .d.ts using TypeScript compiler API.

However, the PR doesn't resolve a problem when using JSR and npm at the same time.

- [ ] Add tests
- [x] Run tests


yusukebe:
@nakasyou 

Wow, very interesting approach! I'll review it later.

package.json
yusukebe, line 1: @nakasyou 

Can you reduce the diff lines? Maybe it's formatted unintendedly.

nakasyou: Thanks for discovering. I fixed it.

build.ts
```diff
@@ -81,4 +84,9 @@ const esmBuild = () =>
 
 Promise.all([esmBuild(), cjsBuild()])
 
-exec(`tsc ${isWatch ? '-w' : ''} --emitDeclarationOnly --declaration --project tsconfig.build.json`)
+await $`tsc ${isWatch ? '-w' : ''} --emitDeclarationOnly --declaration --project tsconfig.build.json`.nothrow()
+
+console.log('Removing private fields...')
```
yusukebe, line 89: Currently, this `log` shows only once.

It's okay but, it will be better show per file because it spends little long time.

https://github.com/user-attachments/assets/a1811e2b-c962-4cc4-b9dc-309c62c034f1

How do you think about it? If you agree,  remove this line and add the following inside `for await`.

```ts
console.log(`Removing private fields... ${entry}`)
```

yusukebe:
@nakasyou 

You use `scripts` to put `remove-private-fields.ts` but I think the name `scripts` is too general. We have "*scripts*" in `benchmarks` and `perf-measures` not only it. So how about creating `build` directory and put it and also `build.ts`? It will be good for sorting out.

yusukebe:
Oops. Accidentally closed.

build.ts
nakasyou: Yeah, it looks cool for me. I implemented it.

nakasyou:
@yusukebe 
> You use scripts to put remove-private-fields.ts, but I think the name scripts is too general. We have "scripts" in benchmarks and perf-measures not only it. So, how about creating a build directory and putting it and also build.ts in? It will be suitable for sorting out.

I agree. I changed it.

yusukebe (approved):
LGTM!

yusukebe:
@nakasyou 

Thank you! I'll merge this now.

Issue #3483, opened 2024-10-03
Move Node.js serveStatic to this repo


yusukebe:
### What is the feature you are proposing?

I'll propose to move the [Node.js Server's Serve Static Middleware](https://github.com/honojs/node-server?tab=readme-ov-file#serve-static-middleware) in [`honojs/node-server`](https://github.com/honojs/node-server) to this `honojs/hono` repo.

## Create `serveStatic` in `hono/nodejs`

As mentioned in the comment: https://github.com/honojs/node-server/issues/203, we have a base implementation of Serve Static Middleware, but Node.js Server(`@hono/node-server` developed in `honojs/node-server`) does not use it. That implements the same logic, and there are lots of duplicated stuff.

We can use the base Serve Static Middleware in the `honojs/node-server`, but there are problems:

* The base Serve Static will be modified in the future.
* The Serve Static in `honojs/node-server` should follow it.
* If so, we have to update the `peerDependecies`(`hono` package) in `package.json` of `honojs/node-server`.

This is the pain point. So I considered and and found that it's good to move the Node.js Server's Server Static to this repo like:

```ts
import { serve } from '@hono/node-server'
import { Hono } from 'hono'
import { serveStatic } from 'hono/nodejs'


const app = new Hono()

app.use('/static/*', serveStatic({ root: './' }))

serve(app)
```

This means we create a Node.js Adapter (not an "Adapter Server") like the adapters for Cloudflare Workers, Deno, and Bun.

## Should The Node.js Server be in the independent repo?

@exoego said [in the comment](https://github.com/honojs/node-server/issues/203#issuecomment-2384893525):

> it is much simpler if Node.js adapter is in Hono main repo.
> *In this sentence, `Node.js adapter` means `Node.js (Adapter) Server` in `honojs/node-server`.

I can understand that opinion, but the Node.js Server should be `honojs/node-server` though the Serve Static for Node.js will be in this main repo.

Node.js almost completely has Web Standards APIs like `fetch`, `Request`,` Respoense`, etc. But Node.js does not have an "entry point" for Web Standards APIs.

For Cloudflare Workers, Deno, and Bun, we can run the following app with one command:

```ts
export default {
  fetch: () => new Response("Hi!")
}
```

Below is Bun case:

```bash
bun run hello.ts
```

[...]

PR #3523, opened 2024-10-17
[PoC] refactor: implement serve-static middleware with Node.js API

### LINKED ISSUE #3324: Support returning 206 Partial Content with ``serveStatic``

usualoma:
This PR also fixes #3324. Although it is less thorough than #3461 or #3516, it implements the minimum necessary processing of the Renge request header. (And I think the performance is not bad.)

### What is this?

This also has something to do with the organization of serve-static and its integration with node-serve, but as for serve-static, I think that by importing “node:*”, we can use node-serve's serve-static.ts as-is in deno and bun. (I haven't checked whether it works with older versions of Deno and Bun, but I think it should work with the current versions.)
https://github.com/honojs/node-server/blob/main/src/serve-static.ts

Currently, the serve-static is abstracting to bridge the differences in runtime. Still, I think eliminating the abstraction would be simpler, and I would stop trying to absorb the differences with Hono.
https://github.com/honojs/hono/blob/main/src/middleware/serve-static/index.ts#L36-L40

If serve-static in cloudflare-workers is deprecated, I think it would be a good idea to consolidate middleware/serve-static when a major version is released and unify it into a version that uses “node:*”.

Alternatively, I think it might be a good idea to have an “adapter for serve-static that proxies to cloud storage such as R2, S3 and GCP Cloud Storage”. However, if you were to implement this, I think it would be better to have independent code without the current abstraction layer. I think the “Worse is Better” approach would be appropriate here.

- [ ] Add tests
- [x] Run tests


usualoma:
According to this PR, we could say that the following policies apply.
* Web requests and responses follow web standards
* Access to the file system follows the Node.js API

usualoma:
This is a draft of one approach to the following comments.

https://github.com/honojs/hono/pull/3516#issuecomment-2415983393

yusukebe:
Hi @usualoma !

Thank you for the PR. Using the Node.js API, `node:*`, for Deno and Bun is a good approach!

However, though I think you already noticed it, this `hono` repo still has runtime-specific APIs, including for WebSockets and ConnInfo. In my thoughts, the ideal solution **may** be to remove runtime-specific APIs from this `hono` repo and use only WebStandards APIs. If so, we must create another namespace for modules like this serve-static middleware.

We should consider where to put runtime-specific things.

yusukebe:
In addition, if we merge this, this issue can be resolved: https://github.com/honojs/hono/issues/3483

PR #3526, opened 2024-10-18
perf(router):  improve performance of router

EdamAme-x:
router performance improvements

Explicitly cached lengths, adjusted the order of evaluation of logical expressions, (without regexp router, for readability)
 and removed debugging properties to speed up and use other unique values.  (trie-router)

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/src/router/linear-router/router.test.ts
+++ b/src/router/linear-router/router.test.ts
@@ -49,4 +49,17 @@ describe('LinearRouter', () => {
       expect(res[0][0]).toBe('GET /book')
     })
   })
+
+  describe('Skip part', () => {
+    const router = new LinearRouter<string>()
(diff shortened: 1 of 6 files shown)


usualoma:
Hi @EdamAme-x, Thank you!

The performance difference between "for-of" and "for loop" is currently quite small (depending on the runtime), so I don't think it needs to be actively rewritten.

However, looking at the benchmarks for loops, it seems that "for loop" is still the fastest, so the changes in this pull request are reasonable 👍

yusukebe (approved):
LGTM!

yusukebe:
@EdamAme-x 

Thanks! I'll merge this now.

PR #3541, opened 2024-10-21
fix(client): return query params in $url

### LINKED ISSUE #3532: `$url` return doesn't include query params

4.6.5

Bun

run `bun src/client.ts` in this repository: https://github.com/alex-grover/hono-rpc-url-query

output is included in the `client.ts` file in a comment as well.

I would expect the returned URL object to include query params.

Specifically, I want to use the $url as a key for SWR, and the query params are necessary for this to work correctly. It’s particularly confusing since it’s a full URL object that doesn’t fully match the URL of the request being sent.

### What do you see instead?

The URL object contains the origin and pathname, but not the query params.

### Additional information

It looks like this can be fixed by applying [this logic](https://github.com/honojs/hono/blob/90833d24670a8a679112fbcae0b8b0a09683333d/src/client/client.ts#L52-L66) in `$url()` [here](https://github.com/honojs/hono/blob/90833d24670a8a679112fbcae0b8b0a09683333d/src/client/client.ts#L174-L179)

If this is a desired change, I'm happy to open a PR!

alex-grover:
Resolves #3532. I'm not sure if there's a more concise way to write the type definition, or to write a test for it, but I was able to validate that it's working as expected locally.

## PR checklist

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -847,6 +847,27 @@ describe('$url() with a param option', () => {
   })
 })
 
+describe('$url() with a query option', () => {
+  const app = new Hono().get(
+    '/posts',
+    validator('query', () => {
+      return {} as { filter: 'test' }
+    }),
+    (c) => c.json({ ok: true })
+  )
+  type AppType = typeof app
+  const client = hc<AppType>('http://localhost')
+
+  it('Should return the correct path - /posts?filter=test', async () => {
+    const url = client.posts.$url({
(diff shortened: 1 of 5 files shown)


yusukebe (approved):
LGTM!

yusukebe:
Hi @alex-grover 

Awesome! This looks good to me. I'll merge it and release the new patch version. Thank you!

PR #3542, opened 2024-10-21
fix(serve-static): silence `NotFound` warning on Deno

pablo-abc:
Silences the `NotFound` warning when a file is not found using `serveStatic` for Deno.

```
NotFound: No such file or directory (os error 2)
```

Noticed that this is already the default behaviour for Bun's adapter according to #1810 but it is not with Deno due to Deno not allowing to check directly for a file's existence. Ignoring the error is the only way to avoid the warning.

- [ ] Add tests
- [x] Run tests


yusukebe (approved):
LGTM!

yusukebe:
Hi @pablo-abc !

Looks good to me. Merging now. Thanks.

PR #3546, opened 2024-10-22
Support custom secret keys for signing JWTs

kaandok:
- Expanded key type check to include 'secret' for JWT signing. This aligns importPrivateKey method with importPublicKey method for handling "secret" type keys.

- Added a test case to sign, verify, and decode using a custom secret, ensuring the correct flow and error handling with mismatched keys.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/utils/jwt/jws.ts
+++ b/src/utils/jwt/jws.ts
@@ -50,7 +50,7 @@ async function importPrivateKey(key: SignatureKey, alg: KeyImporterAlgorithm): P
     throw new Error('`crypto.subtle.importKey` is undefined. JWT auth middleware requires it.')
   }
   if (isCryptoKey(key)) {
-    if (key.type !== 'private') {
+    if (key.type !== 'private' && key.type !== 'secret') {
       throw new Error(`unexpected non private key: CryptoKey.type is ${key.type}`)
     }
     return key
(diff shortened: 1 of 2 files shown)


src/utils/jwt/jws.ts
```diff
@@ -50,7 +50,7 @@ async function importPrivateKey(key: SignatureKey, alg: KeyImporterAlgorithm): P
     throw new Error('`crypto.subtle.importKey` is undefined. JWT auth middleware requires it.')
   }
   if (isCryptoKey(key)) {
-    if (key.type !== 'private') {
+    if (key.type !== 'private' && key.type !== 'secret') {
       throw new Error(`unexpected non private key: CryptoKey.type is ${key.type}`)
```
yusukebe, line 54: How about changing the error message to clarify the `secret` is also expected? Like this:

```ts
throw new Error(`unexpected key type: CryptoKey.type is ${key.type}, expected 'private' or 'secret'`)
```

yusukebe:
Hi @kaandok 

Nice PR! One thing I've left it as a comment.

src/utils/jwt/jws.ts
kaandok: thanks for checking it out. updated the comment.

yusukebe (commented):
LGTM!

yusukebe:
@kaandok Thank you for the quick fix. Merging now!

PR #3556, opened 2024-10-24
Use bun:test

fel1x-developer:

`bun:test` is the first-party test runner on bun. It is faster than vitest.

- [ ] Add tests
- [x] Run tests


runtime-tests/bun/vitest.config.ts
yusukebe, line 1: This is a breaking change. If we remove this, we can't get the coverage of this Bun runtime test.

exoego: As you can see in CodeCov report, branch coverage is decreased a bit due to this change.
If you trust Bun's test, we can switch test runner from Vitest to Bun.

yusukebe: @exoego 

Thanks! If the CodeCov result will be changed, I think we don't have to merge this PR.

yusukebe:
Hi @fel1x-developer

As this comment: https://github.com/honojs/hono/pull/3556#discussion_r1816251202, the CodeCov result will be changed with this PR. There is no big reason to switch `vitest` to `bun:test`. We can not merge this now. Thanks.

PR #3557, opened 2024-10-24
refactor(timing/utils-body): a few minor edits

sharma-shray:

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/middleware/timing/timing.ts
+++ b/src/middleware/timing/timing.ts
@@ -30,7 +30,9 @@ interface TimingOptions {
 const getTime = (): number => {
   try {
     return performance.now()
-  } catch {}
+  } catch (e) {
+    console.warn('performance.now() is not available. Falling back to Date.now()', e)
+  }
   return Date.now()
 }
 
@@ -75,13 +77,11 @@ const getTime = (): number => {
  */
 export const timing = (config?: TimingOptions): MiddlewareHandler => {
   const options: TimingOptions = {
-    ...{
-      total: true,
-      enabled: true,
-      totalDescription: 'Total Response Time',
-      autoEnd: true,
-      crossOrigin: false,
-    },
+    total: true,
+    enabled: true,
+    totalDescription: 'Total Response Time',
+    autoEnd: true,
+    crossOrigin: false,
     ...config,
   }
   return async function timing(c, next) {
--- a/src/middleware/timing/timing.ts
+++ b/src/middleware/timing/timing.ts
@@ -30,9 +30,7 @@ interface TimingOptions {
 const getTime = (): number => {
   try {
     return performance.now()
-  } catch (e) {
-    console.warn('performance.now() is not available. Falling back to Date.now()', e)
-  }
+  } catch {}
   return Date.now()
 }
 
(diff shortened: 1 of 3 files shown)


src/middleware/timing/timing.ts
```diff
@@ -30,7 +30,9 @@ interface TimingOptions {
 const getTime = (): number => {
   try {
     return performance.now()
-  } catch {}
+  } catch (e) {
+    console.warn('performance.now() is not available. Falling back to Date.now()', e)
```
usualoma, line 34: There are already several `console.warn()` in this file, but these are displayed when “middleware is not set up correctly”.

This fallback to `Date.now()` is displayed when running in an environment where `performance.now()` is not available, but I don't think users would be happy if they were shown a warning for every request.

If we display a warning when it is not available, it would be better to do it like this.

```ts
diff --git a/src/middleware/timing/timing.ts b/src/middleware/timing/timing.ts
index d387f43b..f341592f 100644
--- a/src/middleware/timing/timing.ts
+++ b/src/middleware/timing/timing.ts
@@ -27,12 +27,9 @@ interface TimingOptions {
   crossOrigin?: boolean | string | ((c: Context) => boolean | string)
 }
 
-const getTime = (): number => {
-  try {
-    return performance.now()
-  } catch {}
-  return Date.now()
-}
+const getTime: () => number =
+  performance?.now?.bind(performance) ??
+  (console.warn('performance.now is not supported, using Date.now'), Date.now.bind(Date))
 
 /**
  * Server-Timing Middleware for Hono.
```

usualoma:
Hi @sharma-shray, thanks for the suggestion.

It may seem like a change that is obvious to you and doesn't need to be explained, but from the perspective of maintenance, it would be appreciated if the “why” was written in the commit message. Please use the following format as a reference for writing messages in future commits.

https://www.conventionalcommits.org/en/v1.0.0/

src/middleware/timing/timing.ts
sharma-shray: Hello! thanks for you response

Now I really think that was unnecessary, was planning to add some outputs about possible bugs, but thanks for your comments!


> There are already several `console.warn()` in this file, but these are displayed when “middleware is not set up correctly”.
> 
> This fallback to `Date.now()` is displayed when running in an environment where `performance.now()` is not available, but I don't think users would be happy if they were shown a warning for every request.
> 
> If we display a warning when it is not available, it would be better to do it like this.
> 
> ```ts
> diff --git a/src/middleware/timing/timing.ts b/src/middleware/timing/timing.ts
> index d387f43b..f341592f 100644
> --- a/src/middleware/timing/timing.ts
> +++ b/src/middleware/timing/timing.ts
> @@ -27,12 +27,9 @@ interface TimingOptions {
>    crossOrigin?: boolean | string | ((c: Context) => boolean | string)
>  }
>  
> -const getTime = (): number => {
> -  try {
> -    return performance.now()
> -  } catch {}
> -  return Date.now()
> -}
> +const getTime: () => number =
> +  performance?.now?.bind(performance) ??
> +  (console.warn('performance.now is not supported, using Date.now'), Date.now.bind(Date))
>  
>  /**
>   * Server-Timing Middleware for Hono.
> ```

yusukebe (approved):
LGTM!

yusukebe:
Hi @sharma-shray 

Thank you for the PR. Looks good. But next time, please follow the [@usualoma 's comment](https://github.com/honojs/hono/pull/3557#issuecomment-2436520634).

PR #3561, opened 2024-10-26
fix(middleware/bearer-auth): return 403 Forbidden when credentials are invalid

ryanccn:
This PR modifies the [Bearer Auth Middleware](https://hono.dev/docs/middleware/builtin/bearer-auth) to return a 403 Forbidden response instead of a 401 Unauthorized response when a Bearer access token is provided in the request but is not allowed by the middleware configuration. According to spec,

> a 403 is returned when a request contains valid credentials, but the client does not have permissions to perform a certain action ([MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401))

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/bearer-auth/index.test.ts
+++ b/src/middleware/bearer-auth/index.test.ts
@@ -264,8 +264,8 @@ describe('Bearer Auth by Middleware', () => {
     req.headers.set('Authorization', 'Bearer invalid-token')
     const res = await app.request(req)
     expect(res).not.toBeNull()
-    expect(res.status).toBe(401)
-    expect(await res.text()).toBe('Unauthorized')
+    expect(res.status).toBe(403)
+    expect(await res.text()).toBe('Forbidden')
     expect(res.headers.get('x-custom')).toBeNull()
   })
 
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @ryanccn 

I think `401 Unauthorized` is okay because the condition you pointed out throws a return response when the token user set mismatches with the token in the request header. This means "Unauthorize".

ryanccn:
According to [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110#section-15.5.4), a 403 Forbidden response should be returned if:

> If authentication credentials were provided in the request, the server considers them insufficient to grant access. The client SHOULD NOT automatically repeat the request with the same credentials.

I think this is a good match for a Bearer token that is provided by the user but isn't accepted by the middleware configuration.

yusukebe:
@ryanccn 

Based on the RFC, the correct behavior is to use `401 Unauthorized` when the token is invalid. `401` is appropriate for "authentication failure", whereas `403 Forbidden` should be used when "authentication succeeded, but the client lacks permission."

In this case, since the client’s token doesn’t match the server’s expectation, it’s an "authentication failure," making `401` the right response.

ryanccn:
> In this case, since the client’s token doesn’t match the server’s expectation, it’s an "authentication failure," making 401 the right response.

@yusukebe I think that's a good point! I am convinced that 401 is probably the right choice here.

On a different note, do you think it would be a good idea to allow functions that return `Response`s in `noAuthenticationHeaderMessage/invalidAuthenticationHeaderMessage/invalidTokenMessage`? I feel like that would allow more flexibility (with headers, etc.)

Issue #3567, opened 2024-10-27
jsx/dom: All components are memoized.


usualoma:
### What version of Hono are you using?

v4.6.7

### What runtime/platform is your app running on?

Any environment.

### What steps can reproduce the bug?

Rrender the following app.

```ts
const RandomValue = () => <div>{Math.random()}</div>
const RandomApp = () => {
  const [count, setCount] = useState(0)
  return (
    <>
      <RandomValue />
      <div>{count}</div>
      <button onClick={() => setCount(count + 1)}>+</button>
    </>
  )
}
```

### What is the expected behavior?

Since it is not memoized, a random value should be displayed each time it is updated.

https://github.com/user-attachments/assets/6d355ee0-e0ce-441f-9433-d456f98f073d

### What do you see instead?

The value is not updated.

https://github.com/user-attachments/assets/546b47ea-26d9-49f0-a007-7e62789990ca

### Additional information

This bug is due to the fact that the fix was applied incorrectly when checking the behavior of the React Compiler. In order to be memoized by the React Compiler, a different approach is needed to handle memoization.
https://github.com/honojs/hono/pull/3049

Issue #3473, opened 2024-10-01
jsx memo break hooks in client side


codehz:
### What version of Hono are you using?

4.6.3

### What runtime/platform is your app running on?

Bun

### What steps can reproduce the bug?

```jsx
const Counter = memo(function Counter() {
  const [count, setCount] = useState(0);
  return (
    <div>
      <p>Count: {count}</p>
      <button onClick={() => setCount(count + 1)}>Increment</button>
    </div>
  );
});
```

### What is the expected behavior?

the hook should keep working after memoization (to compatible with react) on client side

### What do you see instead?

click increment not working, it seems the behavior is different between hono and react, is it expected?

PR #3568, opened 2024-10-27
fix(jsx/dom): fix `memo` for DOM renderer

usualoma:
#3473
#3567

The `jsx/dom` memoization was generally behaving incorrectly, so I'm fixing it.

### 12ee8293ccb06dc3e0ae41116ebc412d925d3bac

In the past, the code used the same approach as for `toString(),` “holding memoized data in JavaScript memory,” but when rendering to the DOM, I found that memoization could be achieved by “not re-evaluate components” in the renderer, so the processing within the renderer was changed accordingly.

Also, looking at the code generated by React Compiler, it seemed that React expected the ReactNode object generated by `jsx()` to be cached and reused, and to be memoized, so I made it behave in the same way. I applied this branch to the code at https://github.com/sor4chi/explore-react-compiler-with-hono-jsx and confirmed that it was memoized.

### 38a3c7e5438c08b9036c3988e8932c2457bfb7b6 / 3b45b574d08890c8ead9bc8d8a093d4f6909b782

I changed the implementation of `memo` so that memoization is done using a new mechanism.

- [x] Add tests
- [x] Run tests


yusukebe:
Hi @usualoma 

Looks good to me!

@sor4chi Are you interested in reviewing this PR? If so, please!

yusukebe:
I'll merge this now, though @sor4chi does not review it yet. Thanks!

Fixes #3473
Fixes #3567

sor4chi:
I was too busy to see it, sorry...
Good test of react-compiler. I could confirm that it is working correctly for me too!

usualoma:
@sor4chi Thanks for the review!

Issue #3485, opened 2024-10-03
Hono RPC client - types inferred incorrectly


pjeziorowski:
### What version of Hono are you using?

4.6.2

### What runtime/platform is your app running on?

Bun

### What steps can reproduce the bug?

```ts
export type PreferencesDTO = {
  userId: string
  newMessages: boolean
  newConnectionRequests: boolean
  matchSuggestionsOrUpdates: boolean
  productUpdates: boolean
  createdAt: string
  updatedAt: string | null
}

const preferences: PreferencesDTO = {...}

return c.json({
  data: preferences
}, 200) ----> The "output" of this gets inferred as "any" in the client:
```

### What is the expected behavior?

Types for the basic JS types are inferred correctly

### What do you see instead?

Output type in the RPC client is of type `any`

<img width="440" alt="image" src="https://github.com/user-attachments/assets/9944acff-c75e-4a5f-a69a-a815ca8aafe8">

### Additional information

If you cast your output type like this

```
return c.json({
  data: preferences
}, 200) as {
        newMessages: boolean
        newConnectionRequests: boolean
        matchSuggestionsOrUpdates: boolean
        productUpdates: boolean
      } 
 ```
 
 then it's fine, but it's super inconvenient to be forced to cast types like this

**Comment by pjeziorowski** [comment]
I found the issue!

I'm using a monorepo with this structure:

- apps
  - backend
  - frontend
- packages
  - shared

In the shared package, I had stored some basic utility types. It turned out that importing these into the backend and returning data that had been wrapped or transformed by these types/functions was causing problems with type inference. Relocating the shared code directly into the backend resolved the issue.

Question - Do I have to set something up for the type inference to work fine without moving the shared code directly to the backend part?

**Comment by yusukebe** [comment]
HI @pjeziorowski 

I need to look into your actual project to find the answer. But, it's not a bug of `hono` and the problem of setting the project though it's a bad issue for us. So, honestly, I was hoping you could talk about this on [Discussions](https://github.com/orgs/honojs/discussions) or Discord.

**Comment by detarkende** [comment]
I have a similar situation in [this repo](https://github.com/detarkende/stremio-ncore-addon/tree/f5ca898299de1c29593235b3d3dc8204095f16e1).

When I create a file `src/client/api/api-client.ts`, with this content:
```ts
import { hc } from 'hono/client';
import type { ApiRoutes } from 'src/server/index';

export const client = hc<ApiRoutes>('/');
```
then I see this issue:
[...]

PR #3580, opened 2024-10-29
fix(types): replace Hono with HonoBase to get better consistency of types

m-shaka:

- [ ] Add tests
- [x] Run tests

I don't understand the cause well, but this change resolved https://github.com/honojs/hono/issues/3485#issuecomment-2440386025

Also, I removed mutual dependency between `Hono` and `HonoBase`:

```mermaid
graph TD;
    HonoBase-->HandlerInterface;
    HandlerInterface-->Hono;
    Hono-->HonoBase;
```

There are subclasses of `HonoBase` other than `Hono`, so return types of `HandlerInterface` should not be `Hono`.


yusukebe:
@m-shaka Thanks! Great!

What I find a little curious is that the type of the value returned by `app.get()` is not `HonoBase`, but `Hono`. This is ideal behavior, but I couldn't figure out why this is the following.

CleanShot 2024-10-29 at 14 47 36@2x

yusukebe:
@m-shaka 

The behavior [I commented](https://github.com/honojs/hono/pull/3580#issuecomment-2443276200) is ideal, so if it is the correct behavior you expected, let's go with it.

m-shaka:
> What I find a little curious is that the type of the value returned by app.get() is not HonoBase, but Hono.

I guess this is because `HonoBase` is named as `Hono` when it's defined. If you rename `HonoBase` `Hono`, you'll see `HonoBase` there
https://github.com/honojs/hono/blob/e7732a5f23c66ade29531acecc1469a1aa397088/src/hono-base.ts#L102

yusukebe (approved):
LGTM!

yusukebe:
@m-shaka Thanks! Let's go with this.

PR #3585, opened 2024-10-30
perf(compose): check if it is a instance only once

EdamAme-x:

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -37,6 +37,8 @@ export const compose = <C extends ComposeContext, E extends Env = Env>(
 ): ((context: C, next?: Function) => Promise<C>) => {
   return (context, next) => {
     let index = -1
+    const isContext = context instanceof Context
+
     return dispatch(0)
 
     /**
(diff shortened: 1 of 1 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@EdamAme-x 

Nice.

I've tried to benchmark for `instanceof` with this script:

```ts
import { bench, group, run } from 'mitata'
import { Context } from '../../src/context'

const c = new Context(new Request('http://localhost'))

bench('noop', async () => {})

group(() => {
  bench('instaceof', async () => {
    c instanceof Context
  })
  bench('noop', async () => {})
})

run()
```

There is no difference in Bun, but there is a slight difference on Node.js:

```ts
-------------------------------------- -------------------------------
instaceof                54.87 ns/iter  54.20 ns █▅
                 (51.39 ns … 92.19 ns)  81.03 ns ██▃▂▂▄▂▁▁▁▁▁▁▁▂▂▁▁▁▁▁
noop                     47.66 ns/iter  48.45 ns ▅ █
                 (42.72 ns … 90.24 ns)  73.71 ns █▃█▅▄▃▃▃▁▁▁▁▁▂▁▂▁▁▁▁▁
```

Interesting. Let's merge this. Thanks!

PR #3595, opened 2024-10-31
fix(hono-base): remove `.matchRoute` and reduce bundle size

EdamAme-x:
`.matchRoute` is a private method that is only called from one place and does not wrap a complex process, so I think it can be removed.  
This will reduce the bundle size.  

- [ ] Add tests
- [x] Run tests


yusukebe (approved):
LGTM!

yusukebe:
@EdamAme-x 

You are right! We can remove it. Thanks.

PR #3596, opened 2024-10-31
refactor: use `#` for private methods to reduce the minified file size

yusukebe:
This PR reduces the bundle size using `#` for the private methods instead of the `private` keyword when the code is minified.

The result for the "Hello World" app with the `hono/tiny` preset. This is minified with `esbuild --minify` command:

```
-rw-r--r--@  1 yusuke  staff  12167 10 31 16:13 current.js
-rw-r--r--@  1 yusuke  staff  11688 10 31 16:18 next.js
```

`479 bytes` will be reduced without changing any logic and without performance degradation!

### `#` vs `private`

For example, you have the code:

```ts
class A {
  private myPrivateMethod() {}
  myPublicMethod() {
    this.myPrivateMethod()
  }
}
```

If you use the `private` keyword for the private method, the function name is not minifed with esbuild

```js
"use strict";class A{myPrivateMethod(){}myPublicMethod(){this.myPrivateMethod()}}
```

Instead of `private`, you can use `#` for the private method.

```ts
class A {
  #myPrivateMethod() {}
  myPublicMethod() {
    this.#myPrivateMethod()
  }
}
```

Then, the function name will be minified!

```js
"use strict";class A{#t(){}myPublicMethod(){this.#t()}}
```

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -690,6 +690,8 @@ export class Context<
     })
   }
 
+  #newResponse = this.newResponse
+
   /**
    * `.body()` can return the HTTP response.
    * You can set headers with `.header()` and set HTTP status code with `.status`.
@@ -717,8 +719,8 @@ export class Context<
     headers?: HeaderRecord
   ): Response => {
     return typeof arg === 'number'
-      ? this.newResponse(data, arg, headers)
-      : this.newResponse(data, arg)
+      ? this.#newResponse(data, arg, headers)
+      : this.#newResponse(data, arg)
   }
 
   /**
@@ -750,8 +752,8 @@ export class Context<
     this.#preparedHeaders['content-type'] = TEXT_PLAIN
     // @ts-expect-error `Response` due to missing some types-only keys
     return typeof arg === 'number'
-      ? this.newResponse(text, arg, headers)
-      : this.newResponse(text, arg)
+      ? this.#newResponse(text, arg, headers)
+      : this.#newResponse(text, arg)
   }
 
   /**
@@ -779,7 +781,7 @@ export class Context<
     this.#preparedHeaders['content-type'] = 'application/json; charset=UTF-8'
     /* eslint-disable @typescript-eslint/no-explicit-any */
     return (
-      typeof arg === 'number' ? this.newResponse(body, arg, headers) : this.newResponse(body, arg)
+      typeof arg === 'number' ? this.#newResponse(body, arg, headers) : this.#newResponse(body, arg)
     ) as any
   }
 
@@ -794,14 +796,14 @@ export class Context<
     if (typeof html === 'object') {
       return resolveCallback(html, HtmlEscapedCallbackPhase.Stringify, false, {}).then((html) => {
         return typeof arg === 'number'
-          ? this.newResponse(html, arg, headers)
-          : this.newResponse(html, arg)
+          ? this.#newResponse(html, arg, headers)
+          : this.#newResponse(html, arg)
       })
     }
 
     return typeof arg === 'number'
-      ? this.newResponse(html as string, arg, headers)
-      : this.newResponse(html as string, arg)
+      ? this.#newResponse(html as string, arg, headers)
+      : this.#newResponse(html as string, arg)
   }
 
   /**
@@ -825,7 +827,7 @@ export class Context<
   ): Response & TypedResponse<undefined, T, 'redirect'> => {
     this.#headers ??= new Headers()
     this.#headers.set('Location', location)
-    return this.newResponse(null, status ?? 302) as any
+    return this.#newResponse(null, status ?? 302) as any
   }
 
   /**
--- a/src/context.ts
+++ b/src/context.ts
@@ -623,11 +623,11 @@ export class Context<
     return Object.fromEntries(this.#var)
   }
 
-  #newResponse: NewResponse = (
+  #newResponse(
     data: Data | null,
     arg?: StatusCode | ResponseInit,
     headers?: HeaderRecord
-  ): Response => {
+  ): Response {
     // Optimized
     if (this.#isFresh && !headers && !arg && this.#status === 200) {
       return new Response(data, {
--- a/src/context.ts
+++ b/src/context.ts
@@ -623,7 +623,7 @@ export class Context<
     return Object.fromEntries(this.#var)
   }
 
-  newResponse: NewResponse = (
+  #newResponse: NewResponse = (
     data: Data | null,
     arg?: StatusCode | ResponseInit,
     headers?: HeaderRecord
@@ -689,10 +689,7 @@ export class Context<
     })
   }
 
-  #newResponse(...args: unknown[]) {
-    // @ts-expect-error Type mismatch for args in newResponse call
-    return this.newResponse(...args)
-  }
+  newResponse: NewResponse = (...args) => this.#newResponse(...(args as Parameters<NewResponse>))
 
   /**
    * `.body()` can return the HTTP response.
--- a/src/context.ts
+++ b/src/context.ts
@@ -690,7 +690,10 @@ export class Context<
     })
   }
 
-  #newResponse = this.newResponse
+  #newResponse(...args: unknown[]) {
+    // @ts-expect-error Type mismatch for args in newResponse call
+    return this.newResponse(...args)
+  }
 
   /**
    * `.body()` can return the HTTP response.
--- a/src/hono-base.ts
+++ b/src/hono-base.ts
@@ -130,11 +130,11 @@ class Hono<E extends Env = Env, S extends Schema = {}, BasePath extends string =
         if (typeof args1 === 'string') {
           this.#path = args1
         } else {
-          this.addRoute(method, this.#path, args1)
+          this.#addRoute(method, this.#path, args1)
         }
         args.forEach((handler) => {
           if (typeof handler !== 'string') {
-            this.addRoute(method, this.#path, handler)
+            this.#addRoute(method, this.#path, handler)
           }
         })
         return this as any
@@ -147,7 +147,7 @@ class Hono<E extends Env = Env, S extends Schema = {}, BasePath extends string =
         this.#path = p
         for (const m of [method].flat()) {
           handlers.map((handler) => {
-            this.addRoute(m.toUpperCase(), this.#path, handler)
+            this.#addRoute(m.toUpperCase(), this.#path, handler)
           })
         }
       }
@@ -163,7 +163,7 @@ class Hono<E extends Env = Env, S extends Schema = {}, BasePath extends string =
         handlers.unshift(arg1)
       }
       handlers.forEach((handler) => {
-        this.addRoute(METHOD_NAME_ALL, this.#path, handler)
+        this.#addRoute(METHOD_NAME_ALL, this.#path, handler)
       })
       return this as any
     }
@@ -174,7 +174,7 @@ class Hono<E extends Env = Env, S extends Schema = {}, BasePath extends string =
     this.getPath = strict ? options.getPath ?? getPath : getPathNoStrict
   }
 
-  private clone(): Hono<E, S, BasePath> {
+  #clone(): Hono<E, S, BasePath> {
     const clone = new Hono<E, S, BasePath>({
       router: this.router,
       getPath: this.getPath,
@@ -183,8 +183,8 @@ class Hono<E extends Env = Env, S extends Schema = {}, BasePath extends string =
     return clone
   }
 
-  private notFoundHandler: NotFoundHandler = notFoundHandler
-  private errorHandler: ErrorHandler = errorHandler
+  #notFoundHandler: NotFoundHandler = notFoundHandler
+  #errorHandler: ErrorHandler = errorHandler
 
   /**
    * `.route()` allows grouping other Hono instance in routes.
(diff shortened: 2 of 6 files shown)


yusukebe:
Hi @usualoma @EdamAme-x 

What do you think of this?

EdamAme-x:
I think it's a nice idea.  
If we focus only on the core package, the other thing that could be changed is the `private` of router.  

https://github.com/search?q=repo%3Ahonojs%2Fhono+lang%3ATypeScript++%22private+%22+path%3A%2F%5Esrc%5C%2Frouter%5C%2F%2F&type=code

yusukebe:
I'm not sure how much this change will affect real world examples, but parsing time may be reduced, and the file upload time to the production environment will be faster, although super slightly.

yusukebe:
@EdamAme-x 

Good idea! Updated https://github.com/honojs/hono/pull/3596/commits/24d4fb157bdbdce0d38717e9ef4d55446f54dcea

yusukebe:
@usualoma 

Thanks! I was wondering how to define the `#newResponse`. How about this "class B" style?

https://github.com/honojs/hono/pull/3596/commits/7f05f7d66535728abb6a1aaf07ab0c95f17933fa

usualoma:
@yusukebe Thank you!

### benchmark - 2

Here is a version of the benchmark with a few additions.

* Add `class C` to define and assign functions outside
* Not only `new`, but also `method()` calls

The results are almost the same for `class B` and `class C` in any runtime, with `class A` performing worse, maybe JIT is less effective or something.

```ts

bench('noop', () => {})

class A {
  method = () => {}
}

class B {
  method() {
    // noop
  }
}

const method = () => {}
class C {
  method = method
}

group('create instance', () => {
  bench('class A', () => new A().method())
  bench('class B', () => new B().method())
  bench('class C', () => new C().method())
})

run()
```

```
cpu: Apple M2 Pro
runtime: node v22.2.0 (arm64-darwin)

benchmark      time (avg)             (min … max)       p75       p99      p999
noop           70 ps/iter        (20 ps … 339 ns)     82 ps     82 ps    163 ps !

• create instance
class A       168 ns/iter       (162 ns … 268 ns)    167 ns    212 ns    263 ns
class B       110 ps/iter      (61 ps … 67.81 ns)    122 ps    122 ps    183 ps !
class C       112 ps/iter        (61 ps … 146 ns)    122 ps    143 ps    203 ps !

summary for create instance
  class B
   1.02x faster than class C
   1525.21x faster than class A
```

```
cpu: Apple M2 Pro
runtime: deno 2.0.0 (aarch64-apple-darwin)

benchmark      time (avg)             (min … max)       p75       p99      p999
noop          499 ps/iter     (427 ps … 22.91 ns)    509 ps    590 ps    732 ps

• create instance
class A      2.48 ns/iter      (1.85 ns … 188 ns)   2.36 ns   4.35 ns  36.01 ns
class B       549 ps/iter       (488 ps … 119 ns)    549 ps    611 ps   1.65 ns
class C       556 ps/iter     (488 ps … 72.79 ns)    550 ps    692 ps   1.44 ns

summary for create instance
  class B
   1.01x faster than class C
   4.52x faster than class A
```

```
cpu: Apple M2 Pro

benchmark      time (avg)             (min … max)       p75       p99      p999
noop           49 ps/iter       (0 ps … 19.88 ns)     61 ps     82 ps    122 ps !

• create instance
class A     33.61 ns/iter     (29.07 ns … 302 ns)   31.7 ns    169 ns    234 ns
class B        48 ps/iter         (0 ps … 205 ns)     41 ps     61 ps    122 ps !
class C        45 ps/iter         (0 ps … 178 ns)     41 ps     61 ps    122 ps !

summary for create instance
  class C
   1.06x faster than class B
   739.63x faster than class A
```

### c.newRespons vs. c.#newRespons

`c.newResponse` is not called in many apps. On the other hand, `c.#newResponse` is often called multiple times internally. Therefore, I think it would be better to prioritize improving the performance of `c.#newResponse`.
Also, even though it is for internal use, I think it would be better to add some type to `c.#newResponse`.



For the reasons given above, why not give `c.#newResponse` priority with the following changes?

```diff
diff --git a/src/context.ts b/src/context.ts
index a591db72..363b6556 100644
--- a/src/context.ts
+++ b/src/context.ts
@@ -624,11 +624,11 @@ export class Context<
     return Object.fromEntries(this.#var)
   }
 
-  newResponse: NewResponse = (
+  #newResponse(
     data: Data | null,
     arg?: StatusCode | ResponseInit,
     headers?: HeaderRecord
-  ): Response => {
+  ): Response {
     // Optimized
     if (this.#isFresh && !headers && !arg && this.#status === 200) {
       return new Response(data, {
@@ -690,10 +690,7 @@ export class Context<
     })
   }
 
-  #newResponse(...args: unknown[]) {
-    // @ts-expect-error Type mismatch for args in newResponse call
-    return this.newResponse(...args)
-  }
+  newResponse: NewResponse = (...args) => this.#newResponse(...(args as Parameters<NewResponse>))
 
   /**
    * `.body()` can return the HTTP response.
```

yusukebe:
@usualoma 

> For the reasons given above, why not give `c.#newResponse` priority with the following changes?

That's a very interesting idea!! I've embraced it: https://github.com/honojs/hono/pull/3596/commits/3ac8d6216642303f6d62fc30221f8cdcbbdbc895

I think we can go now. Can you review this again?

usualoma:
@yusukebe Thank you!

Do you think this change is something you don't want to add? (Because it's a change that's in my comments but not in 3ac8d62)

Well, it may not be necessary, but from the benchmark, performance decreases with each additional `#method = () => {}`, so I think it's better to use `#method() {}` for hard private methods, even if it means compromising on some typing.

```diff
diff --git a/src/context.ts b/src/context.ts
index 3bb9ceda..86124e00 100644
--- a/src/context.ts
+++ b/src/context.ts
@@ -623,11 +623,11 @@ export class Context<
     return Object.fromEntries(this.#var)
   }
 
-  #newResponse: NewResponse = (
+  #newResponse(
     data: Data | null,
     arg?: StatusCode | ResponseInit,
     headers?: HeaderRecord
-  ): Response => {
+  ): Response {
     // Optimized
     if (this.#isFresh && !headers && !arg && this.#status === 200) {
       return new Response(data, {
```

yusukebe:
@usualoma Thanks!

Or like this?

```diff
diff --git a/src/context.ts b/src/context.ts
index 3bb9ceda..6a2f8f47 100644
--- a/src/context.ts
+++ b/src/context.ts
@@ -623,11 +623,11 @@ export class Context<
     return Object.fromEntries(this.#var)
   }

-  #newResponse: NewResponse = (
+  #newResponse(
     data: Data | null,
     arg?: StatusCode | ResponseInit,
     headers?: HeaderRecord
-  ): Response => {
+  ): Response {
     // Optimized
     if (this.#isFresh && !headers && !arg && this.#status === 200) {
       return new Response(data, {
@@ -689,7 +689,9 @@ export class Context<
     })
   }

-  newResponse: NewResponse = (...args) => this.#newResponse(...(args as Parameters<NewResponse>))
+  newResponse(...args: Parameters<NewResponse>) {
+    return this.#newResponse(...args)
+  }

   /**
    * `.body()` can return the HTTP response.
```

usualoma:
@yusukebe Thank you for your consideration.

I think that would mean that the intended type would no longer be available for the public API `newResponse()`.

CleanShot 2024-11-02 at 19 24 59@2x

yusukebe:
@usualoma 

Thanks! Probably, the type definition is not good. How about changing `interface NewResponse`?

```diff
diff --git a/src/context.ts b/src/context.ts
index 3bb9ceda..3bf30cef 100644
--- a/src/context.ts
+++ b/src/context.ts
@@ -107,8 +107,7 @@ interface Set<E extends Env> {
  * Interface for creating a new response.
  */
 interface NewResponse {
-  (data: Data | null, status?: StatusCode, headers?: HeaderRecord): Response
-  (data: Data | null, init?: ResponseInit): Response
+  (data: Data | null, arg?: StatusCode | ResponseInit, headers?: HeaderRecord): Response
 }

 /**
@@ -623,11 +622,11 @@ export class Context<
     return Object.fromEntries(this.#var)
   }

-  #newResponse: NewResponse = (
+  #newResponse(
     data: Data | null,
     arg?: StatusCode | ResponseInit,
     headers?: HeaderRecord
-  ): Response => {
+  ): Response {
     // Optimized
     if (this.#isFresh && !headers && !arg && this.#status === 200) {
       return new Response(data, {
@@ -689,7 +688,9 @@ export class Context<
     })
   }

-  newResponse: NewResponse = (...args) => this.#newResponse(...(args as Parameters<NewResponse>))
+  newResponse(...args: Parameters<NewResponse>) {
+    return this.#newResponse(...args)
+  }

   /**
    * `.body()` can return the HTTP response.
```

usualoma:
@yusukebe Thank you!

I don't think that's what we wanted to do with the original definition, because it would accept the following.

CleanShot 2024-11-02 at 20 45 45@2x

If we prioritize performance over being DRY, I think the only way to write it is as follows

```diff
diff --git a/src/context.ts b/src/context.ts
index 3bb9ceda..cd60ef1f 100644
--- a/src/context.ts
+++ b/src/context.ts
@@ -623,11 +623,11 @@ export class Context<
     return Object.fromEntries(this.#var)
   }
 
-  #newResponse: NewResponse = (
+  #newResponse(
     data: Data | null,
     arg?: StatusCode | ResponseInit,
     headers?: HeaderRecord
-  ): Response => {
+  ): Response {
     // Optimized
     if (this.#isFresh && !headers && !arg && this.#status === 200) {
       return new Response(data, {
@@ -689,7 +689,13 @@ export class Context<
     })
   }
 
-  newResponse: NewResponse = (...args) => this.#newResponse(...(args as Parameters<NewResponse>))
+  newResponse(data: Data | null, status?: StatusCode, headers?: HeaderRecord): Response
+  newResponse(data: Data | null, init?: ResponseInit): Response
+  newResponse(
+    ...args: [data: Data | null, arg?: StatusCode | ResponseInit, headers?: HeaderRecord]
+  ): Response {
+    return this.#newResponse(...args)
+  }
 
   /**
    * `.body()` can return the HTTP response.
```

Or, we could go one step further and do this.

```diff
diff --git a/src/context.ts b/src/context.ts
index 3bb9ceda..72069c1e 100644
--- a/src/context.ts
+++ b/src/context.ts
@@ -103,19 +103,6 @@ interface Set<E extends Env> {
   <Key extends keyof ContextVariableMap>(key: Key, value: ContextVariableMap[Key]): void
 }
 
-/**
- * Interface for creating a new response.
- */
-interface NewResponse {
-  (data: Data | null, status?: StatusCode, headers?: HeaderRecord): Response
-  (data: Data | null, init?: ResponseInit): Response
-}
-
-/**
- * Interface for responding with a body.
- */
-interface BodyRespond extends NewResponse {}
-
 /**
  * Interface for responding with text.
  *
@@ -623,11 +610,11 @@ export class Context<
     return Object.fromEntries(this.#var)
   }
 
-  #newResponse: NewResponse = (
+  #newResponse(
     data: Data | null,
     arg?: StatusCode | ResponseInit,
     headers?: HeaderRecord
-  ): Response => {
+  ): Response {
     // Optimized
     if (this.#isFresh && !headers && !arg && this.#status === 200) {
       return new Response(data, {
@@ -689,7 +676,13 @@ export class Context<
     })
   }
 
-  newResponse: NewResponse = (...args) => this.#newResponse(...(args as Parameters<NewResponse>))
+  newResponse(data: Data | null, status?: StatusCode, headers?: HeaderRecord): Response
+  newResponse(data: Data | null, init?: ResponseInit): Response
+  newResponse(
+    ...args: [data: Data | null, arg?: StatusCode | ResponseInit, headers?: HeaderRecord]
+  ): Response {
+    return this.#newResponse(...args)
+  }
 
   /**
    * `.body()` can return the HTTP response.
@@ -712,11 +705,9 @@ export class Context<
    * })
    * ```
    */
-  body: BodyRespond = (
-    data: Data | null,
-    arg?: StatusCode | ResponseInit,
-    headers?: HeaderRecord
-  ): Response => {
+  body(data: Data | null, init?: ResponseInit): Response
+  body(data: Data | null, status?: StatusCode, headers?: HeaderRecord): Response
+  body(data: Data | null, arg?: StatusCode | ResponseInit, headers?: HeaderRecord): Response {
     return typeof arg === 'number'
       ? this.#newResponse(data, arg, headers)
       : this.#newResponse(data, arg)
```

usualoma:
I think the proposal described in the following comments is a good compromise that does not involve major changes to the current code and is unlikely to cause a drop in performance.

https://github.com/honojs/hono/pull/3596#issuecomment-2452809673

yusukebe:
@usualoma 

Sorry to bother you. Let's go as you suggested. Can you review it?

usualoma (approved):
Very refactoring 👍

yusukebe:
Okay! Merging now.

EdamAme-x:
It is irrelevant, but I will note why this works for later contributors.  

TypeScript makes all properties with private modifiers public properties when transpiling them to JavaScript.  

`private fire() {}` => `fire() {}`

### Playground

`https://www.typescriptlang.org/play/?#code/MYGwhgzhAEASD2A7e0DeAoaXoAcBOAlgG5gAuAptAGYF7kAUAlGgL7otA`

image

Therefore, the minified is recognized as public property; it does not minify the property name.

Incidentally, there is almost no difference in performance between public and private properties.  
The difference is barely noticeable after several million accesses.  
So there is no need to worry about it.

PR #3606, opened 2024-11-01
refactor(utils/mime): make object and type commonality

EdamAme-x:
In the previous state, we were writing the same thing twice. This PR defines it in only one place and then creates the type definition from the union of the object's values.

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/src/utils/mime.ts
+++ b/src/utils/mime.ts
@@ -32,58 +32,9 @@ export { baseMimes as mimes }
 /**
  * Union types for BaseMime
  */
-export type BaseMime =
-  | 'audio/aac'
-  | 'video/x-msvideo'
-  | 'image/avif'
-  | 'video/av1'
-  | 'application/octet-stream'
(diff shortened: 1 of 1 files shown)


src/utils/mime.ts
```diff
@@ -139,4 +90,6 @@ const baseMimes: Record<string, BaseMime> = {
   '3g2': 'video/3gpp2',
   gltf: 'model/gltf+json',
   glb: 'model/gltf-binary',
-}
+} as const
+
+const baseMimes: Record<string, string> = _baseMimes
```
yusukebe, line 95: Should this be `Record<string, BaseMime>`?

EdamAme-x: thanks!

yusukebe (approved):
LGTM!

yusukebe:
@EdamAme-x 

Thanks! Looks good. Let's go.

PR #3619, opened 2024-11-03
[PoC] feat: generate TanStack Query Options from Hono Client functions

laurentlahmy:
It seems from my research that Hono misses that last tiny bit of functionality for a seamless integration of Hono RPC with TanStack Query.

### Problem:

The "problem" now is that we need to write [queryKeys](https://tanstack.com/query/v5/docs/framework/react/guides/query-keys) and [queryFunctions](https://tanstack.com/query/v5/docs/framework/react/guides/query-functions) for all routes, as is done [here](https://github.com/betterstack-community/betternews-hono-tanstack/blob/main/frontend/src/lib/api.ts). It seems cumbersome and redundant to add React Query helper functions for all routes when they are all available on the Hono client.

According to [this comment](https://github.com/honojs/hono/issues/727#issuecomment-1378814366) this is missing to a few people and a [selling point for tRPC](https://trpc.io/docs/client/react) over Hono RPC. 

### Proposed solution:

I tried to generate Tanstack [queryKey](https://tanstack.com/query/v5/docs/framework/react/guides/query-keys), [queryFn](https://tanstack.com/query/v5/docs/framework/react/guides/query-functions) and [queryOptions](https://tanstack.com/query/v5/docs/framework/react/guides/query-options) by reusing as much of the autocompletion that Hono client provides.

In your frontend code you can do the following: 
```tsx
queryClient.fetchQuery(
    getQueryOptions(
        () => api.dashboard.deployments.user[':userId'].$get({ param: { userId } }), // the RPC endpoint you're targeting
        [userId] // a complement you might want to add to the queryKey 
    )
)
``` 
and this will seamlessly generate the query key and function pair that you need when using React Query:
```tsx
{
  queryKey: [
    "dashboard.deployments.user[\":userId\"].$get({ param: { userId } })", // this is static (variables are passed by name, like userId here)
    "6h8s62e7uppe4ee" // this is dynamic (variables are passed by value, userId has been replaced by "6h8s62e7uppe4ee")
  ],
  queryFn: async (): Promise<ResType> => {
    const res = await api.dashboard.deployments.user[':userId'].$get({ param: { userId } };
    if (!res.ok) {
      throw new Error("server error");
    }
    const data = await res.json();
    return data;
  }
}
```

### Further improvement:

Add a parameter to pass extra [queryOptions](https://tanstack.com/query/latest/docs/framework/react/reference/useQuery) to `getQueryOptions`. 
I think it means we need to import types from React Query, which is probably not wanted in this repo. 

### Caveat:

I'm aware that my approach is a bit basic, that I'm not a typescript ninja, and that I probably put the code in the wrong place. Also, I only tested it using React.

I'm very open to your feedback and to the code being improved.

### Thanks

Please let me know what you think, I think it would be a great addition for the many, many Tanstack Query users.

- [ ✅ ] Add tests
- [ ✅ ] Run tests
- [ ✅ ] `bun run format:fix && bun run lint:fix` to format the code
- [ ✅ ] Add [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code

### DIFF
--- a/src/client/utils.test.ts
+++ b/src/client/utils.test.ts
@@ -1,6 +1,11 @@
+import { Hono } from '../hono'
+import { hc } from './client'
 import {
   buildSearchParams,
   deepMerge,
+  getQueryFn,
+  getQueryKey,
+  getQueryOptions,
   mergePath,
   removeIndexString,
   replaceUrlParam,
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @laurentlahmy 

Sorry. We can't accept this PR. The client should not have an external library-specific function.

EdamAme-x:
I think it can be released as a 3rd party library

laurentlahmy:
Started work on [hono-query](https://github.com/laurentlahmy/hono-query).

Also spotted an implementation that seems much more advanced [here](https://github.com/HegarGarcia/hono-react-query).

PR #3630, opened 2024-11-05
bench: add kitajs to benchmarks

douglasduteil:

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/benchmarks/jsx/package.json
+++ b/benchmarks/jsx/package.json
@@ -14,6 +14,7 @@
     "node-html-parser": "^6.1.11"
   },
   "dependencies": {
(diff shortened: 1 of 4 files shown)


usualoma:
Hi @douglasduteil
Thanks for creating the pull request.

I think it would be good if the files under “benchmarks” contained “benchmarks necessary for discussions on hono development”.

Comparisons with other libraries are important, and I have no doubt that "kitajs" is a great library. Still, at the moment, I don't think there is any need to compare with "kitajs" within this repository.

If the aim is to compare library performance, it would be better to do so in a neutral repository independent of this one.

douglasduteil:
Alright :)

PR #3636, opened 2024-11-06
fix(jsr): fix missing exports

EdamAme-x:
Closes: https://github.com/honojs/hono/issues/3635

PR #3638, opened 2024-11-06
chore(build): validate if exporting is correct in `package.json` and `jsr.json`

EdamAme-x:
Related: https://github.com/honojs/hono/pull/3636

If there is a difference between the two exports, the build will fail.


EdamAme-x:
It fails as follows.
image

The test succeeds when this PR is merged.
https://github.com/honojs/hono/actions/runs/11704963747/job/32598601119?pr=3638

EdamAme-x:
Fixed some missing exports

EdamAme-x:
@yusukebe 
Could you review it?

yusukebe:
Hi @EdamAme-x !

This PR means you added a function to validate the configuration of exporting modules in `package.json` and `jsr.json` with `validateExports`, right?

EdamAme-x:
That is what I mean.
If an endpoint is missing that should be exported to either, throw an error.

yusukebe:
@EdamAme-x 

Thanks! Seems to be good. But adding a test for `validateExports` like `remove-private-fields.test.ts` is better.

EdamAme-x:
thanks, okay.

EdamAme-x:
@yusukebe 
ready for review

build/remove-private-fields.test.ts
yusukebe, line 1: Can you remove this file's change? It is unrelated to this PR.

EdamAme-x: okay, It looks like the format that IDE foramatter did.

EdamAme-x:
@yusukebe 
Re-ready for review

yusukebe:
Hi @nakasyou Can you review this?

nakasyou:
It looks good to me. An idea, what do you think about generating `exports` in `package.json` from `jsr.json`?

build/build.ts
```diff
@@ -7,21 +7,30 @@
...
 
 const args = arg({
   '--watch': Boolean,
 })
 
 const isWatch = args['--watch'] || false
 
+const readJsonExports = (path: string) => JSON.parse(fs.readFileSync(path, 'utf-8')).exports
+
+const [packageJsonExports, jsrJsonExports] = ['./package.json', './jsr.json'].map(readJsonExports)
```
nakasyou, line 28: I think it also should validate `typesVersions`.

EdamAme-x:
> It looks good to me.
Thanks.
> An idea, what do you think about generating `exports` in `package.json` from `jsr.json`?
I think there is room for consideration, let's create a new issue and discuss it there.

yusukebe (approved):
LGTM!

build/build.ts
EdamAme-x: thanks, will work on it.

yusukebe: @nakasyou Ah, you are right. But perhaps validating `typeVersions` is another issue? We can check whether it's correct without `jsr.json`.

@EdamAme-x What do you think of it?

EdamAme-x: I was able to write the code, but you are right, it may be another problem.　　
I will consider this after this PR is merged.

yusukebe: Okay. Let's merge this first.

yusukebe:
@EdamAme-x @nakasyou 

Thanks! Let's go with this.

Issue #3634, opened 2024-11-06
Fix bug of serveStatic of deno adapter


oussamasf:
### What version of Hono are you using?

4.6.9

### What runtime/platform is your app running on? (with version if possible)

Deno

### What steps can reproduce the bug?

```
const publicFiles = serveStatic({ root: './public/' })
app.use('/.well-known/*', publicFiles)
```
when `./public/.well-known/ ` is an empty folder 

### What is the expected behavior?

404

### What do you see instead?

curl: (18) transfer closed with outstanding read data remaining

### Additional information

This is a bug fix from issues #3628 

Opening an empty directory with Deno.open returns a ReadableStream, which is not valid for directories. This leads to that behavior and lead also to status code 200 although it supposed to be 404.

To prevent this undesired behavior, I have implemented a guard using Deno.stat to check if the path is a file before attempting to open it.

p.s: the guard added is tested and tests are passed 
May I have permission to submit a pull request with this fix?

EdamAme-x:
I think it's good.

oussamasf:
[pr](https://github.com/honojs/hono/pull/3639)

Issue #3628, opened 2024-11-04
serveStatic invalid response when index.html is not found


alexgleason:
### What version of Hono are you using?

4.6.9

### What runtime/platform is your app running on? (with version if possible)

Deno

### What steps can reproduce the bug?

```ts
const publicFiles = serveStatic({ root: './public/' });
app.use('/.well-known/*', publicFiles);
```

Ensure that `./public/.well-known/` exists, but does not contain an `index.html`, then:

```sh
➜  curl http://0.0.0.0:4036/.well-known/
curl: (18) transfer closed with outstanding read data remaining
➜  curl http://0.0.0.0:4036/.well-known
curl: (18) transfer closed with outstanding read data remaining
```

### What is the expected behavior?

It should return a valid 404 response.

### What do you see instead?

It does not return a valid response.

PR #3639, opened 2024-11-06
fix(serveStatic): add guard to prevent reading empty folders

oussamasf:
https://github.com/honojs/hono/issues/3628

This PR addresses an issue where attempting to open an empty directory with `Deno.open` returns a `ReadableStream`, causing unexpected behavior. Specifically, it leads to a status code `200` instead of the expected 404 for directories.

- [X] Add tests
- [X] Run tests

### DIFF
--- a/runtime-tests/deno/middleware.test.tsx
+++ b/runtime-tests/deno/middleware.test.tsx
@@ -139,6 +139,22 @@ Deno.test('Serve Static middleware', async () => {
   res = await app.request('http://localhost/static-absolute-root/plain.txt')
   assertEquals(res.status, 200)
   assertEquals(await res.text(), 'Deno!')
+
+  res = await app.request('http://localhost/static')
+  assertEquals(res.status, 404)
+  assertEquals(await res.text(), '404 Not Found')
+
+  res = await app.request('http://localhost/static/dir')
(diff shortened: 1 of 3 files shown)


EdamAme-x:
I think this function is supposed to determine if the path is a directory...
If so, this may be a bug in serve-static as a whole.

https://github.com/honojs/hono/blob/main/src/adapter/deno/serve-static.ts#L25

EdamAme-x:
@oussamasf 

Adding the following to the beginning of your pull request is useful because it will automatically close the issue when it is merged!  
`Closes: https://github.com/honojs/hono/issues/3634`

yusukebe:
Hi @oussamasf !

Thank you for the PR. Seems to be good. But, we need to write tests. Could you add a test for it? You can put it to `runtime-tests/deno/middleware.test.tsx`.

oussamasf:
@EdamAme-x I replaced stat with the existing function 
@yusukebe I added the minimum pattern (..as instructed in the comments of the test file)

yusukebe (approved):
LGTM!

yusukebe:
@oussamasf 

Looks good to me! Thank you for your contribution.

PR #3647, opened 2024-11-08
perf(trie-router): optimize and remove unnecessary processes

EdamAme-x:
1. The following process is not necessary and has been removed.

https://github.com/honojs/hono/pull/3647/files#diff-bf1549ba9405f87c376d2d338883f54bfeb6a91f40b28e62dd2ce5e9d4edd4e5L66-L69

2. Since it will be minified, I changed the name to be easy to understand.

Related: https://github.com/honojs/hono/pull/3596

https://github.com/honojs/hono/pull/3647/files#diff-bf1549ba9405f87c376d2d338883f54bfeb6a91f40b28e62dd2ce5e9d4edd4e5L85

3. Since only the type number is used for the key, an empty object should be created instead of `Object.create(null)`. It is faster and smaller.

Node and Bun (8x~ faster)
```llvm
runtime: node 20.18.0 (x64-linux)

benchmark              avg (min … max) p75   p99    (min … top 1%)
-------------------------------------- -------------------------------
= Object.create(null)    16.67 ns/iter  15.13 ns  █                   
                 (13.54 ns … 96.35 ns)  42.71 ns ▇█▂▁▂▂▁▁▂▁▁▁▁▁▁▁▁▁▁▁▁
= {}                      2.41 ns/iter   2.22 ns  █                   
                  (1.84 ns … 95.40 ns)   8.21 ns ▁█▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁
```

```llvm
runtime: deno 2.0.5 (x86_64-unknown-linux-gnu)

benchmark              avg (min … max) p75   p99    (min … top 1%)
-------------------------------------- -------------------------------
= Object.create(null)    16.64 ns/iter  14.93 ns  █                   
                (12.03 ns … 137.93 ns)  33.21 ns ▁██▃▁▁▁▁▁▁▁▁▁▁▅▂▁▁▁▁▁
= {}                      2.70 ns/iter   2.19 ns █                    
                  (2.03 ns … 62.84 ns)  16.17 ns █▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁
```
In Bun, the speed of both is almost the same.

https://github.com/honojs/hono/pull/3647/files#diff-bf1549ba9405f87c376d2d338883f54bfeb6a91f40b28e62dd2ce5e9d4edd4e5L95

~~4. Multiple index accesses are eliminated and the size is reduced.~~

~~https://github.com/honojs/hono/pull/3647/files#diff-bf1549ba9405f87c376d2d338883f54bfeb6a91f40b28e62dd2ce5e9d4edd4e5R130~~

5. Remove as it is not a necessary check

https://github.com/honojs/hono/pull/3647/files#diff-bf1549ba9405f87c376d2d338883f54bfeb6a91f40b28e62dd2ce5e9d4edd4e5L178

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/src/router/trie-router/node.ts
+++ b/src/router/trie-router/node.ts
@@ -154,7 +154,7 @@ export class Node<T> {
           // '/hello/*/foo' => match /hello/bar/foo
           if (pattern === '*') {
             const astNode = node.children['*']
-            if (node.children['*']) {
+            if (astNode) {
               handlerSets.push(
                 ...this.#getHandlerSets(astNode, method, node.params, Object.create(null))
               )
--- a/src/router/trie-router/node.ts
+++ b/src/router/trie-router/node.ts
@@ -63,10 +63,6 @@ export class Node<T> {
       curNode = curNode.children[p]
     }
 
-    if (!curNode.methods.length) {
-      curNode.methods = []
-    }
-
     const m: Record<string, HandlerSet<T>> = Object.create(null)
 
     const handlerSet: HandlerSet<T> = {
@@ -82,7 +78,7 @@ export class Node<T> {
   }
 
   // getHandlerSets
-  #gHSets(
+  #getHandlerSets(
     node: Node<T>,
     method: string,
     nodeParams: Record<string, string>,
@@ -92,7 +88,7 @@ export class Node<T> {
     for (let i = 0, len = node.methods.length; i < len; i++) {
       const m = node.methods[i]
       const handlerSet = (m[method] || m[METHOD_NAME_ALL]) as HandlerParamsSet<T>
-      const processedSet: Record<string, boolean> = Object.create(null)
+      const processedSet: Record<number, boolean> = {}
       if (handlerSet !== undefined) {
         handlerSet.params = Object.create(null)
         for (let i = 0, len = handlerSet.possibleKeys.length; i < len; i++) {
@@ -131,12 +127,15 @@ export class Node<T> {
           nextNode.params = node.params
           if (isLast) {
             // '/hello/*' => match '/hello'
-            if (nextNode.children['*']) {
+            const astNode = nextNode.children['*']
+            if (astNode) {
               handlerSets.push(
-                ...this.#gHSets(nextNode.children['*'], method, node.params, Object.create(null))
+                ...this.#getHandlerSets(astNode, method, node.params, Object.create(null))
               )
             }
-            handlerSets.push(...this.#gHSets(nextNode, method, node.params, Object.create(null)))
+            handlerSets.push(
+              ...this.#getHandlerSets(nextNode, method, node.params, Object.create(null))
+            )
           } else {
             tempNodes.push(nextNode)
           }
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @EdamAme-x !

It looks awesome! I'd like to know how much this change affects performance. Did you benchmark it? And how much did you reduce the bundle size?

PS.

To benchmark the req/res performance, you can follow the method used in this repo: https://github.com/SaltyAom/bun-http-framework-benchmarkhttps://github.com/SaltyAom/bun-http-framework-benchmark

EdamAme-x:
# Report

Revert (4): The bundle size is smaller, but we got strange benchmark results that one index access is slower than multiple index accesses, so we reverted.
```llvm
runtime: deno 2.0.5 (x86_64-pc-windows-msvc)

benchmark              avg (min … max) p75   p99    (min … top 1%)
-------------------------------------- -------------------------------
Once                    193.31 ps/iter 195.31 ps ▇  █
                (170.90 ps … 70.87 ns) 341.80 ps █▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁
Multi                   136.00 ps/iter 146.48 ps     █
                 (97.66 ps … 22.39 ns) 219.73 ps ▁▁▁▁█▁▁▁▄▁▁▁▂▁▁▁▂▁▁▁▁
```

## Bundle Size

19.34kb => 19.28kb (1kb = 1024b)
```llvm
gitpod@honojs-hono-0oouxesc90f:/workspace/hono/sandbox$ ls compare -l
total 40
-rw-r--r-- 1 gitpod gitpod 19288 Nov  9 07:09 after.js
-rw-r--r-- 1 gitpod gitpod 19344 Nov  9 07:11 before.js
```

## Benchmark

### In Deno
There was not much change.
(+0.03%)
```llvm
|  Framework       | Runtime | Average | Ping       | Query      | Body       |
| ---------------- | ------- | ------- | ---------- | ---------- | ---------- |
| after | deno | 41,225.407 | 50,386.78 | 38,903.5 | 34,385.94 |
| before | deno | 41,079.073 | 50,923.38 | 39,116.56 | 33,197.28 |
```

yusukebe:
@EdamAme-x 

Have you ever tried to benchmark frameworks on Bun, not only Deno?

EdamAme-x:
I'll benchmark with several approaches.

EdamAme-x:
I measured the performance of router itself using node, bun, and mitata, but did not see any significant performance improvement that we could see.

src/router/trie-router/node.ts
```diff
@@ -151,8 +154,10 @@ export class Node<T> {
           // '/hello/*/foo' => match /hello/bar/foo
           if (pattern === '*') {
             const astNode = node.children['*']
-            if (astNode) {
-              handlerSets.push(...this.#gHSets(astNode, method, node.params, Object.create(null)))
+            if (node.children['*']) {
```
yusukebe, line 157: Is this intended? `node.children['*']` can be `astNode`.

EdamAme-x: Sorry, I made a mistake in changing it.

yusukebe:
@EdamAme-x 

Thanks. Anyway, this is a good improvement. I'll merge this now.

PR #3661, opened 2024-11-12
ci: compare bundle size

EdamAme-x:

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -228,13 +228,12 @@ jobs:
 
   perf-measures-bundle-check-on-pr:
     name: 'Bundle Check on PR'
+    needs: main
     runs-on: ubuntu-latest
     if: github.event_name == 'pull_request'
     steps:
       - uses: actions/checkout@v4
       - uses: oven-sh/setup-bun@v2
-      - run: bun install
-      - run: bun run build
       - run: bunx esbuild --minify --bundle dist/index.js --format=esm --outfile=perf-measures/bundle-check/generated/after.js
       - uses: actions/cache/restore@v4
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -182,6 +182,7 @@ jobs:
           path: coverage/
 
   perf-measures-type-check-on-pr:
+    name: 'Type Check on PR'
     runs-on: ubuntu-latest
     if: github.event_name == 'pull_request'
     steps:
@@ -209,6 +210,7 @@ jobs:
         name: display comparison
 
   perf-measures-type-check-on-main:
+    name: 'Type Check on Main'
     runs-on: ubuntu-latest
     if: github.ref == 'refs/heads/main'
     steps:
@@ -225,6 +227,7 @@ jobs:
           key: type-check-perf-previous-result-${{ github.sha }}
 
   perf-measures-bundle-check-on-pr:
+    name: 'Bundle Check on PR'
     runs-on: ubuntu-latest
     if: github.event_name == 'pull_request'
     steps:
@@ -250,6 +253,7 @@ jobs:
         name: display comparison
 
   perf-measures-bundle-check-on-main:
+    name: 'Bundle Check on Main'
     runs-on: ubuntu-latest
     if: github.ref == 'refs/heads/main'
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -228,12 +228,13 @@ jobs:
 
   perf-measures-bundle-check-on-pr:
     name: 'Bundle Check on PR'
-    needs: main
     runs-on: ubuntu-latest
     if: github.event_name == 'pull_request'
     steps:
       - uses: actions/checkout@v4
       - uses: oven-sh/setup-bun@v2
+      - run: bun install
+      - run: bun run build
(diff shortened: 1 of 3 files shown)


EdamAme-x:
if this pr is merged, `before.js` will also be output to show the change in bundle size.

image

EdamAme-x:
Hi @m-shaka, could you please check this?

perf-measures/bundle-check/generated/.gitkeep
yusukebe, line 1: We should add a `perf-measures/bundle-check/.gitignore` like: https://github.com/honojs/hono/blob/main/perf-measures/type-check/.gitignore

yusukebe:
Hi @k2tzumi! Can you review this, you too?

.github/workflows/ci.yml
```diff
@@ -223,3 +225,44 @@ jobs:
         with:
           path: perf-measures/type-check/previous-result.txt
           key: type-check-perf-previous-result-${{ github.sha }}
+
+  perf-measures-bundle-check-on-pr:
+    name: 'Bundle Check on PR'
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request'
```
k2tzumi, line 232: q: Does the bundle size change every time?
Or does it change only when a specific file is changed?

Should I integrate workflow jobs when importing into https://github.com/honojs/hono/pull/3491? Should we keep them separate? I am pondering.

.github/workflows/ci.yml
```diff
@@ -223,3 +225,44 @@ jobs:
...
+      - uses: actions/cache/restore@v4
+        with:
+          path: perf-measures/bundle-check/generated/before.js
+          restore-keys: |
+            perf-measures-bundle-check-previous-file-
+          key: perf-measures-bundle-check-previous-file-
+      - run: |
+          {
+            echo 'COMPARISON<<EOF'
+            ls -l ./generated | column -s '|' -t
```
k2tzumi, line 248: It seems that it is not enough to simply parse the result of the ls command to get output like the following
https://github.com/honojs/hono/actions/runs/11818332808/job/32925651835?pr=3665#step:9:24

The `column -s '|' -t` command formats `|` delimited characters, so you need to pipe at least `|` delimited output

ex)
```
| | Current | Previous |
| --- | --- | --- |
| Files | 256 | 256 |
| Lines | 115024 | 115024 |
| Identifiers | 112049 | 112049 |
| Symbols | 243346 | 243346 |
| Types | 204684 | 204684 |
| Instantiations | 3044052 | 3044052 |
| Memory used | 422387K | 422692K |
| I/O read | 0.01s | 0.02s |
| I/O write | 0.00s | 0.00s |
| Parse time | 0.28s | 0.28s |
| Bind time | 0.11s | 0.11s |
| Check time | 2.40s | 2.42s |
| Emit time | 0.00s | 0.00s |
| Total time | 2.79s | 2.81s |
```

.github/workflows/ci.yml
EdamAme-x: > q: Does the bundle size change every time?
> Or does it change only when a specific file is changed?

`it change only when a specific file is changed` is right.

Only when a core file is changed.

> Should I integrate workflow jobs when importing into https://github.com/honojs/hono/pull/3491? Should we keep them separate? I am pondering.  

I have limited knowledge about it, sorry.

.github/workflows/ci.yml
EdamAme-x: thanks, any idea to format the output?

EdamAme-x: thanks, any idea to format the output?

k2tzumi: How about the following format?

| | Current | Previous |
| --- | --- | --- |
|  Bundle size | 19206 | 19206 |

The number of bytes assumes that the number of digits will not change in the future.

yusukebe: Looks good to me!

EdamAme-x:
what happen, ddos...?
image

.github/workflows/ci.yml
EdamAme-x: I would like to keep this as a TODO/NOTE as it seems a bit difficult to implement.

https://stackoverflow.com/questions/70708306/github-actions-run-step-job-in-a-workflow-if-changes-happen-in-specific-folde

.github/workflows/ci.yml
EdamAme-x: image
@k2tzumi what do you think?

EdamAme-x: It is simple, but it is good enough for now.
Let's add custom metrics later.

EdamAme-x:
hi @yusukebe, this pr should be ready.

k2tzumi (approved):
Looks good to me!

m-shaka (approved):
Sorry for being late. Great work!

It would be nicer if you could see a unit(like KB or MB) of bundle size, but it's OK to add it later

EdamAme-x:
> It would be nicer if you could see a unit(like KB or MB) of bundle size, but it's OK to add it later

Agreed!  
However, we will add custom metrics later, so it seems like a good idea to add them at that time.

yusukebe (approved):
LGTM!

yusukebe:
@EdamAme-x @k2tzumi @m-shaka

Thank you all! Let's go with this.

PR #3663, opened 2024-11-13
fix(middleware/combine): prevent `c.req.routeIndex` from being changed

### LINKED ISSUE #3657: `every` middleware does not pass `param` to the wrapped middleware if global middleware is present

4.6.9

Node, but same thing happens on Bun and Deno

Consider this test:

```ts
  it('Should pass the path params to middlewares', async () => {
    const app = new Hono();
    app.use('*', nextMiddleware)
    const paramMiddleware: MiddlewareHandler = async (c) => {
      return c.json(c.req.param(), 200)
    }

    app.use('/:id', every(paramMiddleware))
    app.get('/:id', (c) => {
      return c.text('Hello World')
    })

    const res = await app.request('http://localhost/123')
    expect(await res.json()).toEqual({ id: '123' })
  })
```

This test will currently fail, the result will be `{}`. Somehow because of the global middleware (`app.use('*', nextMiddleware)`) `c.req.param` returns an empty object when used on a route with route params. It only happens when it's wrapped in `every` middleware.

`c.req.param()` should return `{ id: '123' }` in the example above.

### What do you see instead?

`c.req.param()` returns `{}` despite being used in a route with route params.

### Additional information

The test will pass if we either remove the middleware applied to every route, or remove the `every` middleware (i.e. use `paramMiddleware` directly. This is naturally a trivial example, but a real life example would be if we e.g. wrapped a bunch of `validator` middlewares in `every`. Validation will fail in that case.

I'm pretty sure this is caused by `combine` but I don't know how to fix it myself; I'm wondering if passing `c.req.param` in `every` to `combine` might've been related to this issue but it's not used in `combine` at all.

usualoma:
fixes #3657

As I also refactored the relevant parts, there are a lot of changes, but the key point is that c.req.routeIndex must not be changed’ in the following line.

https://github.com/honojs/hono/compare/main...usualoma:hono:fix/conbine-every-route-index?expand=1#diff-0127ebff640ea2834e4434792875961100daac21bcdaae2d033a6a59815da3bfR98

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -1,5 +1,4 @@
 import { Context } from './context'
-import type { ParamIndexMap, Params } from './router'
 import type { Env, ErrorHandler, NotFoundHandler } from './types'
 
 /**
@@ -31,7 +30,7 @@ interface ComposeContext {
  * @returns {(context: C, next?: Function) => Promise<C>} - A composed middleware function.
  */
 export const compose = <C extends ComposeContext, E extends Env = Env>(
-  middleware: [[Function, unknown], ParamIndexMap | Params][],
+  middleware: [[Function, unknown], unknown][] | [[Function]][],
   onError?: ErrorHandler<E>,
   onNotFound?: NotFoundHandler<E>
 ): ((context: C, next?: Function) => Promise<C>) => {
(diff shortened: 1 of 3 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@usualoma 

Ah, `c.req.routeIndex`! Indeed, we have to care about it. Looks good. Thanks!

PR #3670, opened 2024-11-14
feat(helper/factory): Calculate Context<Env> Type in createHandlers

### LINKED ISSUE #3467: Context<Env> from middlewares are not inferred in `factory.createHandlers`

4.6.3

Node.js / Bun

The `Env`s from middleware is inferred in the handlers which are created as `app.get('/foo', middleware, ...)`, but it isn't when you create them using `factory.createHandlers(middleware, ...)`.

Here's an example based on the [doc](https://hono.dev/docs/helpers/factory#factory-createhandlers):

```ts
import { Hono } from 'hono';
import { createFactory, createMiddleware } from 'hono/factory'

const app = new Hono();
const factory = createFactory()

const middleware = createMiddleware<{ Variables: { foo: string } }>(async (c, next) => {
  c.set('foo', 'bar')
  await next()
})

const handlers = factory.createHandlers(logger(), middleware, async (c) => {
  // I suppose `c` is `Context<{ Variables: { foo: string } }, ...>`, but actually `Context<any, any, {}>`

  return c.json(c.get('foo'));
});

app
  .get('/api', ...handlers)
  .get('/api', logger(), middleware, async (c) => {
    // Here `c` is inferred as expected.
    return c.json(c.get('foo'));
  })

```

I attach how it looks in VSCode for reference
image

As I wrote in the code comment, the context type of `c` in `createHandlers` is inferred as `Context<{ Variables: { foo: string } }, ...>` so I can use `c.get('foo')` or `c.var.foo` with the inferred type, as like as it works for the handler which directly attached to `app`.

### What do you see instead?

The type of `c` is `Context<any, any, {}>` so I cannot get benefit from the type. Although it works as expected for the directly attached handler.

### Additional information

I am not 100% sure if my assumption that the inference works for the `createHandlers`. If it is not expected as design (or TypeScript constraint), sorry for my misunderstanding.

I have searched the issue and found #3202 (and its dups - #3341, #3198) is discussing the inference for middlewares, but I guess it is actually not relevant.

sushichan044:
closes #3467

This PR fixes a problem in createHandlers where the expected type information was not passed to subsequent handlers 
when middlewares changed the Context<Env> type.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/helper/factory/index.test.ts
+++ b/src/helper/factory/index.test.ts
@@ -227,6 +227,29 @@ describe('createHandler', () => {
       expectTypeOf(routes).toEqualTypeOf<Expected>()
     })
   })
+
+  describe('Types - Context Env with Multiple Middlewares', () => {
+    const factory = createFactory()
+
+    const mw1 = createMiddleware<{ Variables: { foo: string } }>(async (c, next) => {
+      c.set('foo', 'bar')
+      await next()
+    })
+
+    const mw2 = createMiddleware<{ Variables: { bar: number } }>(async (c, next) => {
+      c.set('bar', 1)
+      await next()
+    })
+
+    it('Should set the correct type for context from multiple middlewares', () => {
+      factory.createHandlers(mw1, mw2, (c) => {
+        expectTypeOf(c.var.foo).toEqualTypeOf<string>()
+        expectTypeOf(c.var.bar).toEqualTypeOf<number>()
+
+        return c.json({ foo: c.var.foo, bar: c.var.bar })
+      })
+    })
+  })
 })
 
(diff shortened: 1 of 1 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@sushichan044 

Awesome! Thank you for your first contribution. Merging now.

mecab:
I raised #3467 and I'm so happy to hear it is fixed. It is superhelpful. I will try it soon.

Thank you @sushichan044 and @yusukebe !

Rykuno:
I brought this issue up in the discord. Thank you!

Issue #3671, opened 2024-11-14
Cannot access invalid private field 'app.#errorHandler'


TiBianMod:
### What version of Hono are you using?

4.6.10

### What runtime/platform is your app running on? (with version if possible)

bun

### What steps can reproduce the bug?

related of using `bull-board` with `Hono`
`open issues`: https://github.com/felixmosh/bull-board/issues/843

### What do you see instead?

```sh
84 |   #errorHandler = errorHandler;
85 |   route(path, app) {
86 |     const subApp = this.basePath(path);
87 |     app.routes.map((r) => {
88 |       let handler;
89 |       if (app.#errorHandler === errorHandler) {
               ^
TypeError: Cannot access invalid private field (evaluating 'app.#errorHandler')
      at /Users/tt/Code/bull-board/examples/with-hono/node_modules/hono/dist/hono-base.js:89:11
      at map (1:11)
      at route (/Users/tt/Code/bull-board/examples/with-hono/node_modules/hono/dist/hono-base.js:87:16)
      at /Users/tt/Code/bull-board/examples/with-hono/index.js:55:7
```

yusukebe:
@TiBianMod 

This is not a bug. The `errorHandler` was a `private` function, though it was introduced in #3596:

https://github.com/honojs/hono/pull/3596/files#diff-f42b11987a9538ff250a31ba4eb66ee557f7512860fae88c5020a83b5e97c28cL187

Users should not use it.

TiBianMod:
i know is introduced in https://github.com/honojs/hono/pull/3596
`#errorHandler` is not accessible from `app`
https://github.com/honojs/hono/blob/96006d733a3bfe27f5215f7304a85da5859ec0d9/src/hono-base.ts#L219

TiBianMod:
@yusukebe please checkout PR https://github.com/honojs/hono/pull/3673

PR #3673, opened 2024-11-14
fix: Private name "#errorHandler" must be declared in an enclosing class

TiBianMod:
This PR fixes https://github.com/honojs/hono/issues/3671

on `hono-base.ts` the `route` method accepts `app: Hono<SubEnv, SubSchema, SubBasePath>`, so when using the `app` instance we don't have access on private methods.

before the `errorHandler` propery was `private errorHandler` (only valid on typescript) but now is `#errorHandler`, so is not  accessible  anymore.

related: https://github.com/honojs/hono/pull/3596

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/src/hono-base.ts
+++ b/src/hono-base.ts
@@ -184,6 +184,10 @@ class Hono<E extends Env = Env, S extends Schema = {}, BasePath extends string =
   #notFoundHandler: NotFoundHandler = notFoundHandler
   #errorHandler: ErrorHandler = errorHandler
 
+  get errorHandler() {
+    return this.#errorHandler
+  }
+
   /**
    * `.route()` allows grouping other Hono instance in routes.
    *
@@ -214,11 +218,11 @@ class Hono<E extends Env = Env, S extends Schema = {}, BasePath extends string =
     const subApp = this.basePath(path)
     app.routes.map((r) => {
       let handler
-      if (app.#errorHandler === errorHandler) {
+      if (app.errorHandler === errorHandler) {
         handler = r.handler
       } else {
         handler = async (c: Context, next: Next) =>
-          (await compose<Context>([], app.#errorHandler)(c, () => r.handler(c, next))).res
+          (await compose<Context>([], app.errorHandler)(c, () => r.handler(c, next))).res
         ;(handler as any)[COMPOSED_HANDLER] = r.handler
       }
 


yusukebe:
@TiBianMod Thanks!

> on `hono-base.ts` the `route` method accepts `app: Hono<SubEnv, SubSchema, SubBasePath>`, so when using the `app` instance we don't have access on private methods.

Does it throw an error? If so, can you share the minimal code to reproduce it?

TiBianMod:
@yusukebe Thanks,

please checkout https://github.com/TiBianMod/3673-reproduction

but like i set, `#errorHandler` is not accessible from `app`

yusukebe:
@TiBianMod 

As you said, `app.#errorHandler` should not be accessible. But the tests passed and did not throw errors. It's weird.

usualoma:
From the document, there is no problem with the current code, as the JavaScript specification allows access to private properties of other instances of the same class within the class.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties

TiBianMod:
@yusukebe yes is weird, (i tried some tests but no luck)

for me you can this merge this PR (is totally safe and valid)
or we can make the property again `private errorHandler: ErrorHandler = errorHandler` that is valid only on typescript.

usualoma:
Hi @TiBianMod. Thank you for the explanation.

The `errorHandler` property was private by definition even before the change in #3596, so the code you showed us at https://github.com/honojs/hono/pull/3673#issuecomment-2477452591 would result in an error, which is the expected result of the specification.

However, before #3596, the transpiler was exporting `errorHandler` as a public property in dist/hono-base.js, so if you weren't doing type checking in TypeScript, the error wouldn't have been detected and it wouldn't have been an error at runtime either.

Given this situation, although #3596 contains a change that is effectively incompatible, `errorHandler` has always been a private property in Hono, so I think it would be appropriate to fix the application side.

yusukebe:
@usualoma Thanks for the information!

@TiBianMod Can we close the issue and this PR?

TiBianMod:
> @TiBianMod Can we close the issue and this PR?

sure

felixmosh:
The following code is wrong... (it is located in `hono-base.js`)
image

When there is a usage of the following pattern, 
```js
const app = new Hono();
const innerApp = new Hono();
app.route('/', innerApp);
```

it will fail, since the code in the image is accessing a private member of the "passed" `app` (which in our example it is `innerApp`) therefore, the runtime will throw an error.

yusukebe:
@felixmosh 

All our tests pass. Please share your test codes that fail.

yusukebe:
If '#' bad, we can implement using 'private'.

EdamAme-x:
We suppose you could use private in the code and prefix it at build time for readability and all that, but that might be overkill.

felixmosh:
The code above throws an error since it access `private` property of a given `app` param.
So there is some code that tries to access it, therefore, it is not private.

Private means that only the class itself uses it...

felixmosh:
@yusukebe reproduction repo
https://github.com/felixmosh/hono-private-bug

It is related to the fact that the inner Hono app is exported with `commonjs` style

yusukebe:
@TiBianMod I have ideas. Please wait a bit!

PR #3691, opened 2024-11-19
aws-lambda adapter - add alb event requestContext undefined check for testing convenience

ospatil:
### Issue Summary

While testing a lambda using `aws lambda invoke` or from the AWS Console,  the current `aws-lambda` adapter `isProxyEventALB` implementation results in following error:

```sh
{
    "errorType": "TypeError",
    "errorMessage": "Cannot convert undefined or null to object",
    "stack": [
        "TypeError: Cannot convert undefined or null to object",
        "    at Function.hasOwn (<anonymous>)",
        "    at isProxyEventALB (/var/task/index.js:1816:17)",
        "    at getProcessor (/var/task/index.js:1807:7)",
        "    at Runtime.handler (/var/task/index.js:1620:23)",
        "    at Runtime.handleOnceNonStreaming (file:///var/runtime/index.mjs:1173:29)"
    ]
}
```

This can be reproduced by scaffolding a lambda project using `npm create hono@latest` and deploying it in an AWS account without any changes to the code.

### Benefits of the fix

A fix by adding an undefined check for `event.requestContext` will significantly ease the testing of lambda using simple tools like `aws lambda invoke` and AWS Console.

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/src/adapter/aws-lambda/handler.ts
+++ b/src/adapter/aws-lambda/handler.ts
@@ -407,7 +407,7 @@ class ALBProcessor extends EventProcessor<ALBProxyEvent> {
 
   protected getQueryString(event: ALBProxyEvent): string {
     // In the case of ALB Integration either queryStringParameters or multiValueQueryStringParameters can be present not both
-    /* 
+    /*
       In other cases like when using the serverless framework, the event object does contain both queryStringParameters and multiValueQueryStringParameters:
       Below is an example event object for this URL: /payment/b8c55e69?select=amount&select=currency
       {
@@ -471,7 +471,10 @@ export const getProcessor = (event: LambdaEvent): EventProcessor<LambdaEvent> =>
 }
 
 const isProxyEventALB = (event: LambdaEvent): event is ALBProxyEvent => {
-  return Object.hasOwn(event.requestContext, 'elb')
+  if (event.requestContext) {
+    return Object.hasOwn(event.requestContext, 'elb')
+  }
+  return false
 }
 
 const isProxyEventV2 = (event: LambdaEvent): event is APIGatewayProxyEventV2 => {


yusukebe:
Hi @ospatil 

Thank you for the PR. Looks good, but one thing. Can you write a test for this change? There may not be enough tests for this AWS Lambda adapter currently, but we have to add a test for the patch.

ospatil:
@yusukebe I have added tests for `getProcessor` function. Since the `Processor` class types are not exported, I had to limit the assertions to the processor being defined after calling the function.

A big shout-out to you and the team for this amazing project 🙌🏼 .

src/adapter/aws-lambda/handler.ts
```diff
@@ -471,7 +471,10 @@ export const getProcessor = (event: LambdaEvent): EventProcessor<LambdaEvent> =>
 }
 
 const isProxyEventALB = (event: LambdaEvent): event is ALBProxyEvent => {
-  return Object.hasOwn(event.requestContext, 'elb')
+  if (event.requestContext) {
+    return Object.hasOwn(event.requestContext, 'elb')
+  }
+  return false
```
exoego, line 477: This line is not covered

ospatil: That's from the report before I updated the PR with the tests. After adding the tests, the coverage report has not run as it's awaiting maintainer approval.

I ran the coverage locally (`bun coverage`) after adding the tests and it's all green.
hono-lambda-coverage

exoego: I see.
New tests are not executed since approval needed

runtime-tests/lambda/index.test.ts
```diff
@@ -967,3 +971,53 @@ describe('streamHandle function', () => {
...
+      isBase64Encoded: false,
+      requestContext: {
+        elb: {
+          targetGroupArn:
+            'arn:aws:elasticloadbalancing:us-east-1:123456789012:targetgroup/lambda-279XGJDqGZ5rsrHC2Fjr/49e9d65c45c6791a',
+        },
+      },
+    }
+
+    const processor = getProcessor(event)
```
yusukebe, line 990: What do you think of whether we should test whether the `processor` is an instance of `ALBProcessor,` `EventV2Processor`, or `EventV1Processor`?

ospatil: I think that would be more robust assertion. Let me know if it's ok to export `EventProcessor`, `EventV1Processor`, `EventV2Processor`, and `ALBProcessor` classes from `handler.ts`. I can add `toBeInstanceOf` assertions in the tests in that case.

yusukebe: > I think that would be more robust assertion.

Yes. It's okay to export these classes, and you can add the assertions. Can you proceed with it?

ospatil: Done!

yusukebe (approved):
LGTM!

yusukebe:
@ospatil 

Nice job! I'll merge this and release a new version. Thank you for your contribution!

PR #3692, opened 2024-11-19
fix(base): define `errorHandler` with `private` not use `#`

### LINKED ISSUE #3671: Cannot access invalid private field 'app.#errorHandler'

yusukebe:
With this PR, defining `errorHandler` with `private` not using `#`  in `hono-base.ts`:

```ts
private errorHandler: ErrorHandler = errorHandler
```

This will fix #3671

Since the property defined with `#` will be a completely private member that can't be accessed from outside in the JavaScript runtime phase, problem #3671 occurs. But, if you define it with `private` keyword without `#` it can be accessed from outside in the JavaScript runtime phase.

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/hono-base.ts
+++ b/src/hono-base.ts
@@ -182,7 +182,8 @@ class Hono<E extends Env = Env, S extends Schema = {}, BasePath extends string =
   }
 
   #notFoundHandler: NotFoundHandler = notFoundHandler
-  #errorHandler: ErrorHandler = errorHandler
+  // Cannot use `#` because it requires visibility at JavaScript runtime.
+  private errorHandler: ErrorHandler = errorHandler
 
   /**
    * `.route()` allows grouping other Hono instance in routes.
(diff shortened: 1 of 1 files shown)


yusukebe:
@usualoma 

Can you review this?

usualoma:
Hi @yusukebe, thank you 👍 

Okay, so #3671 indicates that the problem is with apps that use a mixture of esm and cjs.
I'm not sure if this (mixing ems and cjs is OK) is within the scope of what Hono supports, but I think this change will fix it.

yusukebe:
@usualoma 

> I'm not sure if this (mixing ems and cjs is OK) is within the scope of what Hono supports

I'm also thinking about this thing! But currently, we have to accept this PR.

PR #3696, opened 2024-11-21
ci(perf-measures): support `KB`

EdamAme-x:
We can check `kb`

### DIFF
--- a/perf-measures/bundle-check/scripts/check-bundle-size.ts
+++ b/perf-measures/bundle-check/scripts/check-bundle-size.ts
@@ -31,7 +31,7 @@ async function main() {
       key: 'bundle-size-kb',
       name: 'Bundle Size (KB)',
       value: parseFloat((bundleSize / 1024).toFixed(2)),
-      unit: 'KB',
+      unit: 'K',
     })
 
     const benchmark = {
--- a/perf-measures/bundle-check/scripts/check-bundle-size.ts
+++ b/perf-measures/bundle-check/scripts/check-bundle-size.ts
@@ -30,7 +30,7 @@ async function main() {
     metrics.push({
       key: 'bundle-size-kb',
       name: 'Bundle Size (KB)',
-      value: (bundleSize / 1024).toFixed(2),
+      value: (bundleSize / 1024).toFixed(2).toString(),
       unit: 'KB',
     })
 
--- a/perf-measures/bundle-check/scripts/check-bundle-size.ts
+++ b/perf-measures/bundle-check/scripts/check-bundle-size.ts
@@ -30,7 +30,7 @@ async function main() {
     metrics.push({
       key: 'bundle-size-kb',
       name: 'Bundle Size (KB)',
-      value: (bundleSize / 1024).toFixed(2).toString(),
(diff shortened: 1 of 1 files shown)


EdamAme-x:
hmm?
https://github.com/honojs/hono/actions/runs/11945102648/job/33297187610?pr=3696#step:7:95
@k2tzumi What do you know?

EdamAme-x:
> * [JS Bundle Analysis](https://docs.codecov.com/docs/javascript-bundle-analysis) - Avoid shipping oversized bundles

Interesting 👀

perf-measures/bundle-check/scripts/check-bundle-size.ts
```diff
@@ -19,12 +19,21 @@ async function main() {
...
+      key: 'bundle-size-b',
+      name: 'Bundle Size (B)',
       value: bundleSize,
       unit: 'B',
     })
+
+    metrics.push({
+      key: 'bundle-size-kb',
+      name: 'Bundle Size (KB)',
+      value: (bundleSize / 1024).toFixed(2),
```
k2tzumi, line 33: I don't know, as I have never tried it, but I was curious about the following warning message.

https://github.com/honojs/hono/actions/runs/11945102648/job/33297187610?pr=3696#step:7:94
`Skip collecting custom metrics: json: float unexpected end of JSON input`

What if we set the value as a string?

EdamAme-x: This is the only issue that has the same error, not octocov.  
So this error itself is unusual...  
Hmmm...  

https://github.com/pluja/whishper/issues/78

perf-measures/bundle-check/scripts/check-bundle-size.ts
```diff
@@ -19,12 +19,21 @@ async function main() {
...
+      key: 'bundle-size-b',
+      name: 'Bundle Size (B)',
       value: bundleSize,
       unit: 'B',
     })
+
+    metrics.push({
+      key: 'bundle-size-kb',
+      name: 'Bundle Size (KB)',
+      value: (bundleSize / 1024).toFixed(2).toString(),
```
k2tzumi, line 33:
```suggestion
      value: parseFloat((bundleSize / 1024).toFixed(2)),
```
It looks like it had to be Number in the custom metrics schema.
https://github.com/k1LoW/octocov/blob/main/report/custom_metrics_schema.json#L51-L53

EdamAme-x:
It seems to work well

yusukebe:
Hi @EdamAme-x

Good PR! One thing I noticed.

When displaying Bundle Size, I think it would be easier to compare "K" with "B":

CleanShot 2024-11-22 at 09 21 29@2x

In the above, the "B" and "KB" digits appear to be aligned. `ls -lh` command also uses the method I suggest, so it's okay.

CleanShot 2024-11-22 at 09 25 58@2x

EdamAme-x:
thanks for suggestions

yusukebe (commented):
LGTM!

yusukebe:
@EdamAme-x 

Thank! Merging.

PR #3702, opened 2024-11-25
feat(logger): include query params

ryuapp:
The default value of built-in logger is not very desirable for API servers that use query params.
e.g.) https://hono-hacker-news.deno.dev/

It's a bit disruptive for logging, so we probably won't need to add it right away.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/logger/index.ts
+++ b/src/middleware/logger/index.ts
@@ -80,10 +80,9 @@ function log(
  */
 export const logger = (fn: PrintFunc = console.log): MiddlewareHandler => {
   return async function logger(c, next) {
-    const { method } = c.req
+    const { method, url } = c.req
 
-    const url = new URL(c.req.raw.url)
-    const path = url.pathname + url.search
+    const path = url.slice(url.indexOf('/', 8))
 
     log(fn, LogPrefix.Incoming, method, path)
 
--- a/src/middleware/logger/index.ts
+++ b/src/middleware/logger/index.ts
@@ -5,7 +5,6 @@
 
 import type { MiddlewareHandler } from '../../types'
 import { getColorEnabled } from '../../utils/color'
-import { getPath } from '../../utils/url'
 
 enum LogPrefix {
   Outgoing = '-->',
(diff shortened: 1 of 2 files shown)


src/middleware/logger/index.ts
```diff
@@ -83,7 +82,8 @@ export const logger = (fn: PrintFunc = console.log): MiddlewareHandler => {
   return async function logger(c, next) {
     const { method } = c.req
 
-    const path = getPath(c.req.raw)
+    const url = new URL(c.req.raw.url)
+    const path = url.pathname + url.search
```
EdamAme-x, line 86: If a large size of query parameters are intentionally sent by an attacker, I believe this should be optional as it will contaminate the standard output.

ryuapp: The same is true even if path is simply long.
So there is room for optimization here, but I don't think you need to worry about URL length.

EdamAme-x: Yes, it is true.  
It might be nice to have the following features.

`/too/longggggggg~/paths` => `/too/longgggggg...`
`/?too=longggggggg~&pathssssssssss~` => `/?too=longg...`

ryuapp: This is optimized on https://github.com/honojs/hono/pull/3702/commits/5a8553ba75068148c97b9906a9ca806896fa31c1 for performance.

> It might be nice to have the following features.
> 
> /too/longggggggg~/paths => /too/longgggggg...
> /?too=longggggggg~&pathssssssssss~ => /?too=longg...

It's an issue that potentially already exists. So that seems outside the scope of this PR.
You may want to create a new PR. Thanks.

EdamAme-x: okay, thanks

yusukebe (approved):
LGTM!

yusukebe:
Hi @ryuapp 

Sorry for the super late response. Looks good! Let's include this feature in the next minor version `v4.7`! Thank you!

PR #3707, opened 2024-11-28
feat: warn when json validator does not find a content type in the request

### LINKED ISSUE #841: zod-validator should show a more useful error when JSON body is missing content-type header
I created a very simple zod-validator to use in my test suite.

```typescript
import { zValidator } from '@hono/zod-validator'
import { z } from 'zod'

export const testBodyValidator = zValidator(
  'json',
  z.object({
    key: z.string().min(1),
  })
)
```

My route looked like

```typescript
router.post('/', testBodyValidator, async (c) => {
  return c.json(c.req.valid('json'))
})
```

My test looked like this

```typescript
describe('POST /testing', () => {
  it('should return the request body', async () => {
    const res = await app.request('/testing', {
      method: 'POST',
      body: JSON.stringify({ key: 'value' }),
    })
    expect(await res.json()).toStrictEqual({ key: 'value' })
  })
})
```

However, I kept getting errors about the `key` field being missing.

After hours of debugging I noticed one example had the content-type header set in the `app.request()`

```
      headers: new Headers({ 'Content-Type': 'application/json' }),
```

That turns out to be what the problem was. While I understand why it happened, it wasn't obvious that was the issue. It would be really useful if zValidator could return an error if the validation target is `json` but the content-type header is missing or does not contain `application/json`.

Soviut:
When a validator has a target 'json' it will now warn if a request is missing a `Content-Type: application/json` header.

Closes https://github.com/honojs/middleware/issues/841

Added some docs too https://github.com/honojs/website/pull/540

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/validator/validator.ts
+++ b/src/validator/validator.ts
@@ -72,8 +72,7 @@ export const validator = <
     switch (target) {
       case 'json':
         if (!contentType || !jsonRegex.test(contentType)) {
-          const message = 'missing Content-Type header "application/json"'
-          throw new HTTPException(415, { message })
+          throw new HTTPException(415, { message: ERROR_MESSAGE_JSON })
           break
         }
         try {
--- a/src/validator/validator.ts
+++ b/src/validator/validator.ts
@@ -64,7 +64,6 @@ export const validator = <
     E,
     P2
   >,
-  { warn = true } = {}
 ): MiddlewareHandler<E, P, V> => {
   return async (c, next) => {
     let value = {}
@@ -73,11 +72,8 @@ export const validator = <
     switch (target) {
       case 'json':
         if (!contentType || !jsonRegex.test(contentType)) {
-          if (warn) {
-            console.warn(
-              'Validator target is "json" but request is missing a Content-Type header containing "application/json"'
-            )
-          }
+          const message = 'missing a Content-Type header "application/json"'
+          throw new HTTPException(415, { message })
           break
         }
         try {
@@ -92,11 +88,8 @@ export const validator = <
           !contentType ||
           !(multipartRegex.test(contentType) || urlencodedRegex.test(contentType))
         ) {
-          if (warn) {
-            console.warn(
-              'Validator target is "form" but request is missing a Content-Type header containing "multipart/form-data" or "application/x-www-form-encoded"'
-            )
-          }
+          const message = 'missing a Content-Type header "application/x-www-form-encoded"'
+          throw new HTTPException(415, { message })
           break
         }
 
--- a/src/validator/validator.ts
+++ b/src/validator/validator.ts
@@ -63,7 +63,7 @@ export const validator = <
     OutputType,
     E,
     P2
-  >,
+  >
 ): MiddlewareHandler<E, P, V> => {
   return async (c, next) => {
     let value = {}
--- a/src/validator/validator.ts
+++ b/src/validator/validator.ts
@@ -27,8 +27,8 @@ const jsonRegex = /^application\/([a-z-\.]+\+)?json(;\s*[a-zA-Z0-9\-]+\=([^;]+))
 const multipartRegex = /^multipart\/form-data(;\s?boundary=[a-zA-Z0-9'"()+_,\-./:=?]+)?$/
 const urlencodedRegex = /^application\/x-www-form-urlencoded(;\s*[a-zA-Z0-9\-]+\=([^;]+))*$/
 
-const ERROR_MESSSAGE_JSON = 'missing Content-Type header "application/json"'
-const ERROR_MESSSAGE_FORM = 'missing Content-Type header "multipart/form-data" or "application/x-www-form-urlencoded"'
+const ERROR_MESSAGE_JSON = 'missing Content-Type header "application/json"'
+const ERROR_MESSAGE_FORM = 'missing Content-Type header "multipart/form-data" or "application/x-www-form-urlencoded"'
 
 export const validator = <
   InputType,
--- a/src/validator/validator.ts
+++ b/src/validator/validator.ts
@@ -72,7 +72,9 @@ export const validator = <
     switch (target) {
       case 'json':
         if (!contentType || !jsonRegex.test(contentType)) {
-          console.warn('Validator target is "json" but request is missing a Content-Type header containing "application/json"')
+          console.warn(
+            'Validator target is "json" but request is missing a Content-Type header containing "application/json"'
+          )
           break
         }
         try {
--- a/src/validator/validator.ts
+++ b/src/validator/validator.ts
@@ -72,7 +72,7 @@ export const validator = <
     switch (target) {
       case 'json':
         if (!contentType || !jsonRegex.test(contentType)) {
-          const message = 'missing a Content-Type header "application/json"'
+          const message = 'missing Content-Type header "application/json"'
           throw new HTTPException(415, { message })
           break
         }
@@ -88,7 +88,7 @@ export const validator = <
           !contentType ||
           !(multipartRegex.test(contentType) || urlencodedRegex.test(contentType))
         ) {
-          const message = 'missing a Content-Type header "application/x-www-form-encoded"'
+          const message = 'missing Content-Type header "application/x-www-form-encoded"'
           throw new HTTPException(415, { message })
           break
         }
--- a/src/validator/validator.ts
+++ b/src/validator/validator.ts
@@ -72,6 +72,7 @@ export const validator = <
     switch (target) {
       case 'json':
         if (!contentType || !jsonRegex.test(contentType)) {
+          console.warn('Validator target is "json" but request is missing a Content-Type header containing "application/json"')
           break
         }
         try {
--- a/src/validator/validator.ts
+++ b/src/validator/validator.ts
@@ -89,6 +89,9 @@ export const validator = <
           !contentType ||
           !(multipartRegex.test(contentType) || urlencodedRegex.test(contentType))
         ) {
+          console.warn(
+            'Validator target is "form" but request is missing a Content-Type header containing "multipart/form-data" or "application/x-www-form-encoded"'
+          )
           break
         }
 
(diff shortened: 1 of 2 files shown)


yusukebe:
Thanks! I'll check it later.

yusukebe:
Hi @Soviut 

Thank you for the PR!

Do you think that we should show a similar message for `form`?

https://github.com/honojs/hono/pull/3707/files#diff-e3ae0a2a0f7c0f9209a9627ba4da7f499f23e1873275b8d0c0f028e77d9cdee4R87-R93

Soviut:
@yusukebe Yes, let's do that. I'll add the change. I should be able to figure out the regex, but just in case, what are the content types we need for the form? `multipart/form-data`, etc.

EdamAme-x:
I think that this may be an option.

Because a threat actor's malicious request could cause a warning to be logged to the console in a production environment

Soviut:
@EdamAme-x @yusukebe this seems to have stalled. Is there a security issue? if not, I'd love to see it merged.

EdamAme-x:
> What is the concern with that? Wouldn't that help identify threat actors more easily if lots of warnings start showing up in logs?

Apologies if this is misplaced.

For example, if someone with malicious intent deliberately sends a request with the Content-Type set incorrectly to the root where the validator is located, it might trigger a warning.

This could result in the logs being affected, which may lead to increased load.

While the administrator would likely notice the attack in the logs, it seems there might not be a clear way to fully address the issue. Even if the IP is blocked, the potential risk of attack would still remain.

EdamAme-x:
I think the solution to this is to allow option to decide whether to log or not.

The user can choose whether or not to generate warnings by environment variables, etc.

In the production environment, the log should not be emitted, but in the development environment, it should be emitted.

```ts
validator(
  "json",
  (...) => {...},
  {
    showWarning: process.env.NODE_ENV === "development"
  }
)
```

EdamAme-x:
@yusukebe What do you think?

Soviut:
@EdamAme-x An option to enable the warnings could be added, but I think that's out of scope for this particular PR. This PR is to address the issue that the invalid body is ignored, making it hard to debug.

I think a follow-up ticket is in order since it would require some additional decisions, docs updates, etc.

EdamAme-x:
Okay, and for now, let's consider the following PRs.

Soviut:
@yusukebe What should I do to get this merged?

yusukebe:
@Soviut @EdamAme-x 

I'm sorry for not getting back to you sooner. I've considered it, and I believe we should not accept this PR.  As @EdamAme-x mentioned, this change will allow the expected warning messages to be shown. This is a vulnerability, and while usability is important, avoiding vulnerabilities is more important.

Soviut:
@yusukebe @EdamAme-x Okay, I'd like to adjust this PR to include the setting that was mentioned earlier. Instead of `showWarning` I'm thinking just `warn` would be cleaner. This would be set to `true` by default.

EdamAme-x:
Assuming we accept this PR, it is doubtful that users will notice this option.
Perhaps we should consider setting the default to false.

EdamAme-x:
If it depends on Adapter Helper, it may not be stylish, although it could be determined from `NODE_ENV` etc.

Soviut:
@EdamAme-x Defaulting to false ultimately ends up back where we started, with a difficult to debug scenario that the warning solves if we set it to true by default.

I'm willing to write documentation about this that should be discoverable by anybody searching for the warning and the associated options to turn it off.

We could also update the warning messages to include a hint about setting warn to false to mute the warnings.

EdamAme-x:
I see, however, it may cause problems that are unnoticeable due to dependency updates, etc.
So it may be extreme, but there may be options such as introducing it in v5.

Soviut:
@EdamAme-x I see. Is there any reason we couldn't return a `415 Unsupported Media Type` error from this? Then the only person receiving it is the sender who ultimately is the one who needs to know their headers are set incorrectly.

We could ignore the validation function passed as the 2nd arg and return the 415, preventing the validation function from returning misleading errors of its own.

> 6.5.13. 415 Unsupported Media Type
> The 415 (Unsupported Media Type) status code indicates that the
origin server is refusing to service the request because the payload
is in a format not supported by this method on the target resource.
The format problem might be due to the request's indicated
Content-Type or Content-Encoding, or as a result of inspecting the
data directly.

EdamAme-x:
Interesting idea.
It might make more sense to return a response than to issue a warning.

@yusukebe What do you think?

Soviut:
@EdamAme-x glad to hear. I agree, the warning was my 2nd choice. I originally wanted to return an error like this but I think I phrased it as an exception earlier. An error response is what I meant.

Soviut:
@EdamAme-x I've updated to the PR to replace the console warnings with `HttpException(415, ...)` and truncated error messages. This is in keeping with similar error messages in other parts of the validator.

src/validator/validator.ts
```diff
@@ -72,6 +72,8 @@ export const validator = <
     switch (target) {
       case 'json':
         if (!contentType || !jsonRegex.test(contentType)) {
+          const message = 'missing Content-Type header "application/json"'
+          throw new HTTPException(415, { message })
```
EdamAme-x, line 76:
```suggestion
          throw new HTTPException(415, { message: ERROR_MESSAGE_JSON })
```

In `bodyLimit` the error message is defined as follows. How about doing the same?
Also, it would be better to start with a capital letter.

https://github.com/honojs/hono/blob/02402400381647cc51bc749e611cfac71c191c25/src/middleware/body-limit/index.ts#L10

```ts
const ERROR_MESSAGE_JSON = 'Missing Content-Type header "application/json"'
```

Soviut: I'm not against it if it's consistent but though the rest of the file doesn't follow that approach.

At the very least we'd need one for the form exception as well.

Soviut: I've moved those two messages to their own consts.

EdamAme-x:
@Soviut Can you also write tests on this?

Soviut:
@EdamAme-x I've updated your suggestion. Let me know what you think.

Just note that there are prior error messages like

```typescript
const message = 'Malformed JSON in request body'
```

and this one that gets dynamically built

```typescript
let message = 'Malformed FormData request.'
message += e instanceof Error ? ` ${e.message}` : ` ${String(e)}`
```

Soviut:
> @Soviut Can you also write tests on this?

Yeah, once we think this looks good I can write some tests. Hopefully some already exist for these cases and just need to be updated.

EdamAme-x:
> @EdamAme-x I've updated your suggestion. Let me know what you think.

I think the current one is fine.

As for the test, I think it could be changed.
I don't think it can be a destructive change if the validator is used correctly.

Soviut:
@EdamAme-x I've updated the tests. The JSON ones only needed the status code updated. However, the FormData tests were (as far as I could tell) missing the same tests, so I replicated them.

EdamAme-x (approved):
LGTM

Soviut:
> LGTM

Great. We'll need to kick off the checks again.

@yusukebe let me know if anything else needs to be done.

src/validator/validator.ts
```diff
@@ -27,6 +27,9 @@ const jsonRegex = /^application\/([a-z-\.]+\+)?json(;\s*[a-zA-Z0-9\-]+\=([^;]+))
 const multipartRegex = /^multipart\/form-data(;\s?boundary=[a-zA-Z0-9'"()+_,\-./:=?]+)?$/
 const urlencodedRegex = /^application\/x-www-form-urlencoded(;\s*[a-zA-Z0-9\-]+\=([^;]+))*$/
 
+const ERROR_MESSAGE_JSON = 'missing Content-Type header "application/json"'
+const ERROR_MESSAGE_FORM = 'missing Content-Type header "multipart/form-data" or "application/x-www-form-urlencoded"'
```
EdamAme-x, line 31:
```suggestion
const ERROR_MESSAGE_JSON = 'Missing Content-Type header "application/json"'
const ERROR_MESSAGE_FORM = 'Missing Content-Type header "multipart/form-data" or "application/x-www-form-urlencoded"'
```

How about this?

EdamAme-x:
Good night, @yusukebe.
The PR direction has been changed to return a response with status code 415.
What do you think?

Soviut:
@EdamAme-x something I've realized is that this approach doesn't quite work when multiple validators are used. Because you can have the following...

```ts
  app.post(
    '/',
    validator('json', (value) => value),
    validator('form', (value) => value),
    async (c) => {
      const jsonData = c.req.valid('json')
      const formData = c.req.valid('form')
      return c.json({
        json: jsonData,
        form: formData,
      })
    }
  )
```

...each validator validates in isolation, meaning if you test for the json target, the form validator will fail.

This seems to be why the validators were initially failing silently; so they could pass-through. I'm going to see if I can come up with a way of handling this.

EdamAme-x:
I wonder if there is a use case for a single endpoint to receive json/form in parallel.
(This is not the case if it is used as middleware for abstract paths, but it is the case if it is used as middleware for abstract paths.)

EdamAme-x:
This would be a destructive change, but what about telling the same route to use a different target validator?
https://hono.dev/docs/middleware/builtin/combine#some

EdamAme-x:
@Soviut What do you think?

Soviut:
I’ll have a look later this evening.On Dec 14, 2024, at 10:58 PM, EdamAmex ***@***.***> wrote:﻿
@Soviut What do you think?

Soviut:
@EdamAme-x I could definitely update the tests to work with `some()` where appropriate.

It is a breaking change so I suppose it might need to wait for Hono 5.0? If so, is there a roadmap? I searched the issues but the last roadmap was for 3.1.0.

EdamAme-x:
Hi @usualoma, what do you think about the idea of using `some` of these use cases?
(It would be a disruptive change, so it might be realistic to release it in Hono v5)

```ts
  app.post(
    '/',
    some(
      validator('json', (value) => value),
      validator('form', (value) => value),
    ),
    async (c) => {
      // ...
    }
  )
```

Soviut:
@EdamAme-x combine middleware definitely could use some more docs because it affects how `c.req.valid()` works.

image

EdamAme-x:
It seems that Combine needs to better support combining validator types, etc.

PR #3711, opened 2024-11-29
refactor: clean up and reduced code size

EdamAme-x:
Some cleanups and commonization of `Object.create(null)` to reduce the amount of code.

`18.5kb -> 18.1kb`

- [ ] Add tests
- [x] Run tests


EdamAme-x:
Oddly, coverage shows a line of code that doesn't exist 👀

yusukebe:
Not only this, but we would like to reduce the amount of code by using a common single method, e.g., `new Response()`. Please wait while I think about how to do it.

EdamAme-x:
Okay.
Common file or put them together at build time.

PR #3717, opened 2024-12-02
fix(aws-lambda): Fix query string handling for v1

Holi0317:

- [x] Add tests
- [x] Run tests

---

Fix handling of multi-value query string on API Gateway v1 event.

Similar to #2782, but for API Gateway v1 event.

API Gateway v2 already handles this case because v2 event uses the raw query string directly.

https://github.com/honojs/hono/blob/190e122180dda4b51e46cbe21a301ccbec1b1f3d/src/adapter/aws-lambda/handler.ts#L285


exoego (commented):
LGTM

yusukebe (approved):
LGTM!

yusukebe:
Hi @Holi0317 

Thank you for the PR. I'll merge this into the main now.

PR #3719, opened 2024-12-03
fix(middleware/cors): explicitly return No Content for the statusText when handling an OPTIONS request

shawncarr:
When the `cors` middleware processes an `OPTIONS` request by default, it returns the response status text from the context, `c.res.statusText`. However, in most cases, the middleware is executed before any routes or other middlewares that could modify the response `statusText`, leading to a `Not Found` response status. To align with the explicit `204` being returned, the response should also include `No Content`, similar to how the `etag` middleware handles setting the `statusText` to `Not Modified`.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/cors/index.test.ts
+++ b/src/middleware/cors/index.test.ts
@@ -81,6 +81,7 @@ describe('CORS by Middleware', () => {
     const res = await app.request(req)
 
     expect(res.status).toBe(204)
+    expect(res.statusText).toBe('No Content')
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @shawncarr 

Thank you for the PR. I think this is good.

usualoma:
Hi @shawncarr, thank you.
LGTM!

yusukebe (commented):
LGTM!

yusukebe:
@shawncarr @usualoma Thanks!

Merging now.

Issue #3723, opened 2024-12-04
ipRestriction not catching ::ffff:127.0.0.1


jnatmorris:
### What version of Hono are you using?

4.6.12

### What runtime/platform is your app running on? (with version if possible)

Node

### What steps can reproduce the bug?

import { serve } from "@hono/node-server";
import { Hono } from "hono";
import { logger } from "hono/logger";
import { ipRestriction } from "hono/ip-restriction";
import { getConnInfo } from "@hono/node-server/conninfo";

const app = new Hono()
	.use(logger())
	.use(async (c, next) => {
		console.log(getConnInfo(c));
		await next();
	})
	.use(
		"*",
		ipRestriction(getConnInfo, {
			denyList: [],
			allowList: ["127.0.0.1", "::1", "::ffff:127.0.0.1"],
		}),
	)
	.get("/", (c) => {
		return c.text("will not reach");
	})

### What is the expected behavior?

HTTP requests from any of the allowed list shall be able to reach the `.get("/",(c)=>{...}`.

### What do you see instead?

When making HTTP requests from the same computer (localhost), a 403 is returned. 

![Screenshot 2024-12-04 at 20 37 44](https://github.com/user-attachments/assets/48dc10f6-dfb6-4a6a-931a-9d647885dd52)

### Additional information

At first I was only making my HTTP requests from a nextjs server component. This did not work. I then used an HTTP making client, and am running into the same issue.

EdamAme-x:
In internal, `::ffff:127.0.0.1` converted to `::ffff:127`
maybe this is bug

PR #3725, opened 2024-12-05
fix: correctly determine ipv4 mapped ipv6

EdamAme-x:
Fixes https://github.com/honojs/hono/issues/3723

- [x] Add tests
- [x] Run tests


usualoma:
Hi @EdamAme-x, thanks for creating the pull request.

As you pointed out, I also think that the cause is that the results of `expandIPv6` are incorrect.

> In internal, ::ffff:127.0.0.1 converted to ::ffff:127

I think that the following specifications should also match, so

https://github.com/usualoma/hono/commit/ca5e479afb1a4d4ac30e0f88ca5a47c2d88add0c#diff-5abe8f1 72d81691c561fcacf25bd0a5b1b4cfa4217ec43203f0d2053529234d3R106

I think it would be better to fix `expandIPv6` and `convertIPv6BinaryToString` as follows.

https://github.com/usualoma/hono/commit/ca5e479afb1a4d4ac30e0f88ca5a47c2d88add0c

What do you think?

EdamAme-x:
That fix would be cooler.
Thanks!

PR #3726, opened 2024-12-05
feat(router): Introduce `OptimizeRouter`

EdamAme-x:
A router that caches static routes to objects.
```ts
  #routes: Record<string, Record<string, number[]>> = Object.create(null)
```

Like SmartRouter, it can specify child routers.
```ts
new OptimizeRouter({
    router: new TrieRouter(),
})
```

... (wait a moment)

- [x] Add tests
- [x] Run tests


aquapi:
@EdamAme-x U should use a `Map` instead
Objects are only fast if the request handler doesn't access unknown paths, which then the property access gets optimized out.

EdamAme-x:
In my benchmarks, using an Object was faster than using a Map.

aquapi:
@EdamAme-x Can u show me the benchmark source?
I have my own router benchmark in which routers that use `Map` for static paths are slightly faster in Node and slightly slower in Bun (Likely cuz JSC lacks optimizations for `Map`).
https://github.com/mapljs/router/tree/main/bench/routers

One potential issue that can occur in your benchmark is that your object isn't deoptimized.
For example if you only access known paths:
```ts
const routes = {
  '/': 0,
  '/user': 1
};

// Fast access using hidden class
routes['/'];
routes['/user'];
```

But if you access unknown paths at some point (10 unknown props I think) the engine will stop creating a hidden class and treat it like a hashmap, which access slower.
So you have to warm up the benchmark using unknown keys first.

PR #3727, opened 2024-12-05
fix(utils/ipaddr): support IPv6-mapped IPv4 address

### LINKED ISSUE #3723: ipRestriction not catching ::ffff:127.0.0.1

usualoma:
fixes #3723

### Return value of `getConniInfo(c)`

In all the runtimes, an IPv6-mapped IPv4 address was returned as the remote address, so I think it is appropriate to handle this by fixing `convertIPv6BinaryToString`.

```ts
import { Hono } from './src/'
import { getConnInfo } from './src/adapter/deno/conninfo' // or bun, node-server

export default new Hono().get('/', (c) => {
  return c.json(getConnInfo(c))
})
```

```ts
$ deno serve --unstable-sloppy-imports --host [::] app.ts
$ bun run app.ts
```

```
% curl http://[::ffff:127.0.0.1]:8000/
{"remote":{"address":"::ffff:127.0.0.1","port":64504,"transport":"tcp"}}
% curl http://[::ffff:7f00:1]:8000/
{"remote":{"address":"::ffff:127.0.0.1","port":64519,"transport":"tcp"}}
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/ip-restriction/index.test.ts
+++ b/src/middleware/ip-restriction/index.test.ts
@@ -100,6 +100,10 @@ describe('isMatchForRule', () => {
   it('Static Rules', async () => {
     expect(await isMatch({ addr: '192.168.2.1', type: 'IPv4' }, '192.168.2.1')).toBeTruthy()
     expect(await isMatch({ addr: '1234::5678', type: 'IPv6' }, '1234::5678')).toBeTruthy()
+    expect(
+      await isMatch({ addr: '::ffff:127.0.0.1', type: 'IPv6' }, '::ffff:127.0.0.1')
+    ).toBeTruthy()
+    expect(await isMatch({ addr: '::ffff:127.0.0.1', type: 'IPv6' }, '::ffff:7f00:1')).toBeTruthy()
   })
   it('Function Rules', async () => {
     expect(await isMatch({ addr: '0.0.0.0', type: 'IPv4' }, () => true)).toBeTruthy()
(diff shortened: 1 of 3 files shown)


usualoma:
@nakasyou @yusukebe 

Would you please review?

nakasyou:
Hi @usualoma, it looks good to me.

yusukebe (approved):
LGTM!

PR #3730, opened 2024-12-05
perf(pattern-router): improve performance when create null object

EdamAme-x:
Reused null object

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/router/pattern-router/router.ts
+++ b/src/router/pattern-router/router.ts
@@ -3,6 +3,8 @@ import { METHOD_NAME_ALL, UnsupportedPathError } from '../../router'
 
 type Route<T> = [RegExp, string, T] // [pattern, method, handler, path]
 
+const emptyParams = Object.create(null)
+
 export class PatternRouter<T> implements Router<T> {
   name: string = 'PatternRouter'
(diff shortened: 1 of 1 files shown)


src/router/pattern-router/router.ts
```diff
@@ -46,7 +48,7 @@ export class PatternRouter<T> implements Router<T> {
       if (routeMethod === method || routeMethod === METHOD_NAME_ALL) {
         const match = pattern.exec(path)
         if (match) {
-          handlers.push([handler, match.groups || Object.create(null)])
+          handlers.push([handler, match.groups || emptyParams])
```
yusukebe, line 51: @EdamAme-x 

A quick question. Is there a possibility that the content of `emptyParams` will be changed? Like `emptyParams.someKey = "newValue"`.

EdamAme-x: Since params should be abstracted by `c.req.param`, so I think there should be no such use case.

yusukebe: Okay. Thanks!

yusukebe (approved):
LGTM!

PR #3740, opened 2024-12-11
feat(aws-lambda): Hono-Lambda Client

watany-dev:
When making requests to Lambda behind CloudFront, you may need to implement payload calculation logic as described in the article below. 
https://dev.classmethod.jp/articles/cloudfront-lambda-url-with-post-put-request/

This involves unnecessary tedious work, which can be eliminated by wrapping Hono's hc. However, a drawback is that the code becomes quite redundant, so you need to be careful to keep it updated.

- [x] Add tests
- [x] Run tests


watany-dev:
Alternatively, we might just provide the calculateSHA256 function and implement it like this:

```typescript
import { calculateSHA256 } from 'hono/lambda'

const payload = { name: 'Alice', message: 'Hello World' };

// Send POST request
const res = await client.echo.$post(
  {
    json: payload,
  },
  {
    headers: {
      'x-amz-content-sha256': await calculateSHA256(JSON.stringify(payload)),
    },
  }
);
```

yusukebe:
Hi @watany-dev 

Sorry for my super late response!

It will be useful. But I'm not sure whether we should introduce this. We have never put specialized features like this Hono-Lambda Client into `adapter/*`. And it will cost maintenance since it depends on the platform.

I think providing `calculateSHA256` is a good idea. Can you use a function with the same logic, `calculateSHA256` from existing utils?

watany-dev:
@yusukebe 
Thank you. I can completely agree with your opinion. Once I close this PR I will consider I/F of `calculateSHA256`. Thanks.

Issue #3751, opened 2024-12-15
Hono Logger ("hono/logger")


kingmesal:
### What is the feature you are proposing?

Currently the color status coding is set with this: 
`  const isNoColor = typeof Deno?.noColor === "boolean" ? Deno.noColor : process !== void 0 ? "NO_COLOR" in process?.env : false;`

Can something be added to check it this is running in a cloudflare worker to disable coloring?

My logs spit out `[32m200[0m 923ms`, which is terrible.

The `c.env` could be passed in and if `NO_COLOR === true` then we are set.

StuartMoncrieff:
I was going to raise a feature request related to disabling status code coloring too!

The Debug Console in VS Code does not support the ANSI escape codes that are used to add the colors, so the output looks like ```--> GET / \x1b[32m200\x1b[0m 7ms```.

Looking through the source code, I found the same thing that @kingmesal did: to disable _status code coloring_, you can set a "NO_COLOR" environment variable, which is described at <https://no-color.org/>.

The way to get coloring _working_ in the VS Code Debug Console (rather than disabling it) was to add ```"outputCapture": "std"``` to my configuration in ```launch.json```.

For @kingmesal's original problem...

I haven't written code to run on CloudFlare Workers before (so this might be a bad work-around), but I understand that they have an empty ```process.env``` object. Could you enable ```"nodejs_compat"``` in your ```wrangler.toml```
and add ```process.env['NO_COLOR'] = 'asdf'``` to your ```index.ts``` file (instead of changing the logger middleware to check for ```c.env.NO_COLOR```)?

yusukebe:
> I haven't written code to run on CloudFlare Workers before (so this might be a bad work-around), but I understand that they have an empty ```process.env``` object. Could you enable ```"nodejs_compat"``` in your ```wrangler.toml```
and add ```process.env['NO_COLOR'] = 'asdf'``` to your ```index.ts``` file (instead of changing the logger middleware to check for ```c.env.NO_COLOR```)?

Wow, interesting idea! This is okay, and we are working on [another solution](https://github.com/honojs/hono/pull/3752#issuecomment-2545086356).

albert-92:
Thanks for implementing this. This is a huge saver, especially when logging to files and not having to postprocess the logs with regex.

kingmesal:
Am I reading the `getColorEnabled()` method correctly that is process === undefined then coloring is disabled? 

If so, that in and of itself solves turning coloring off in Cloudflare.

[...]

PR #3752, opened 2024-12-16
feat(utils/color): support cloudflare workers (edge)

EdamAme-x:
Closes: https://github.com/honojs/hono/issues/3751

- [x] Add tests
- [x] Run tests

### DIFF
--- a/runtime-tests/deno/utils.test.ts
+++ b/runtime-tests/deno/utils.test.ts
@@ -1,0 +1,13 @@
+import { assertEquals } from '@std/assert'
+import { Context } from '../../src/context'
+import { getColorEnabled } from '../../src/utils/color'
+
+Deno.test('getColorEnabled() - With colors enabled', () => {
+  assertEquals(getColorEnabled(), true)
+})
+
+Deno.test('getColorEnabled() - With Deno.noColor set', () => {
+  const mockContext = new Context(new Request('http://localhost/'))
+
+  assertEquals(getColorEnabled(mockContext), true)
+})
--- a/runtime-tests/workerd/index.test.ts
+++ b/runtime-tests/workerd/index.test.ts
@@ -68,20 +68,9 @@ describe('workerd with WebSocket', () => {
   })
 })
 
-describe('worked with getColorEnabled()', () => {
-  let worker: UnstableDevWorker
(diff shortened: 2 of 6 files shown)


yusukebe:
Hi @EdamAme-x 

Thank you for the PR. But I can't agree on this way using `NO_COLOR` in `c.env`. It will be confusing if someone uses `NO_COLOR` in `c.env` for another purpose. I think the better way is to create `colorized` option for the Logger middleware and write it like this:

```ts
const app = new Hono<{
  Bindings: {
    NO_COLOR: boolean
  }
}>()

app.use(async (c, next) => {
  return logger(undefined, {
    colorized: !c.env.NO_COLOR,
  })
})
```

But, I'm not sure it is best to add options for the second arg of `logger`.

yusukebe:
We use `NO_COLOR` variable in `process.env` since it's de facto standard on the Node.js environment:

https://no-color.org/

But, I think we don't have the manner in Cloudflare Workers.

EdamAme-x:
@kingmesel
What do you think?

kingmesal:
My perspective is that the simpler and more consistent, the solution the better.

I think if it were on the `c.env` it is the simplest, but I can understand the argument that it is possible (albeit highly unlikely) that someone has used an `c.env.NO_COLOR` somewhere.

Passing in options is useful as it makes it pretty easy for people to find and alter the logger functionality, which I think is probably the better reason to use that approach over the `c.env` directly.

StuartMoncrieff:
I have created a PR to update the website with a tip on using the `NO_COLOR` environment variable to disable ANSI escape codes in the output from the build-in logger middleware (<https://github.com/honojs/website/pull/553>).

@kingmesal has [noted](https://github.com/honojs/hono/issues/3751) that the default logging behaviour on CloudFlare Workers is to _not_ have colored log output, which makes this PR unnecessary (sorry @EdamAme-x!).

Note that the default behavior of the logger on CloudFlare Workers is different to Node/Bun/Deno, which default to colored status codes.

EdamAme-x:
okay, I will close this

yusukebe:
@StuartMoncrieff @EdamAme-x Thanks!

PR #3758, opened 2024-12-19
chore(jsx): export ComponentWithChildren type

yasuaki640:
When I run `npm run type-check` on [this repository](https://github.com/yasuaki640/gpt-web-client), I get the following type error.

```shell
npm run type-check 

> type-check
> tsc

src/middleware/layout.ts:4:45 - error TS2345: Argument of type 'FC' is not assignable to parameter of type 'ComponentWithChildren'.
  Type 'HtmlEscapedString | Promise<HtmlEscapedString> | null' is not assignable to type 'HtmlEscapedString | Promise<HtmlEscapedString>'.
    Type 'null' is not assignable to type 'HtmlEscapedString | Promise<HtmlEscapedString>'.

4 export const LayoutMiddleware = jsxRenderer(Layout);
                                              ~~~~~~


Found 1 error in src/middleware/layout.ts:4
```

I think the easiest way to fix this is to export the `ComponentsWithChildren` type.

https://github.com/honojs/hono/blob/4eb101dd03e66e20dfda86e6e5ebce8045d61d05/src/middleware/jsx-renderer/index.ts#L28

Since it can happen that a jsx renderer argument is defined in a different jsx file, I think exporting the type is a good idea, what do you think ?

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/jsx-renderer/index.ts
+++ b/src/middleware/jsx-renderer/index.ts
@@ -25,7 +25,7 @@ type Component = (
   c: Context
 ) => HtmlEscapedString | Promise<HtmlEscapedString>
 
-type ComponentWithChildren = (
+export type ComponentWithChildren = (
   props: PropsWithChildren<PropsForRenderer & { Layout: FC }>,
   c: Context
 ) => HtmlEscapedString | Promise<HtmlEscapedString>


yusukebe:
Hey @yasuaki640 

In that case, you can use [`PropsWithChildren`](https://hono.dev/docs/guides/jsx#propswithchildren), and it will be fixed:

```ts
import type { PropsWithChildren } from 'hono/jsx'

export const Layout = ({ children }: PropsWithChildren) => (
// ...
)
```

In most cases, it is recommended to use `PropsWithChildren` for function components rather than `FC` because it will add a type to `children`(It might be better to rewrite the documentation).

If this solves the problem, I don't think we need this PR.

yasuaki640:
@yusukebe 
Error fixed!
Thanks!

Issue #1518, opened 2023-09-28
Cannot response HTTP code 204 when using `context.json()`


PlayerNguyen:
### What version of Hono are you using?

3.7.2

### What runtime/platform is your app running on?

Node

### What steps can reproduce the bug?

- Setup a Node/Hono server
- Implement below code
```ts
const app = new Hono();
app.delete('...', async (context: Context) => {
  const categoryId = context.req.param("category-id");

 const 

  // If not found the category
  if (response === 0) {
    context.status(204);
    return context.json(
        { data: [], message: t("routes.video_categories.category_not_found") }
    );
  }
  return context.json({ id: categoryId });
});

// Hosting the app
```
- Request DELETE / and an exception was occurred
```
TypeError: Response constructor: Invalid response status code 204

```

### What is the expected behavior?

I expected this implemented code should response 204 (No content) data

### What do you see instead?

```
TypeError: Response constructor: Invalid response status code 204

```

vanodevium:
When status code is 204, Response can't contain any content. Body is empty always.

datashaman:
HTTP Status 204 signifies that there is no content in the response body.
https://http.dev/204

Given your code, what you want to use is status 404 instead, which is usually used when you cannot find a record.
https://http.dev/404

yusukebe:
Hi @PlayerNguyen 

Is this problem resolved? If you wan to return 204 with no body, you can return `c.body(null, 204)`.

yusukebe:
Closing. If you have any problems, please reopen

askorupskyy:
@yusukebe just encountered this. While it's an easy fix, this was missed in one of my PRs and caused a few errors in prod. Do you think it could be a good idea to omit the `204` from acceptable response codes in `c.json()` and whatnot?

EdamAme-x:
Yes @askorupskyy.
You can delete body

askorupskyy:
What I had in mind is doing something like 

```typescript
json: JSONRespond = <
    T extends JSONValue | SimplifyDeepArray<unknown> | InvalidJSONValue,
    U extends Exclude<StatusCode, '204'> = Exclude<StatusCode, '204'>
  >
```

EdamAme-x:
Interesting idea

yusukebe:
Hi @askorupskyy @EdamAme-x 

Throwing a TypeScript error if the Response has a status code that can't have a `null` body is a good idea. If so, it should support other status codes, not only `204`:

[...]

Issue #242, opened 2022-05-17
`c.body` accept `null` for 204 response


yusukebe:
```ts
return c.text('no content', 204) // <--- Error! - 204 response must not have body
```

But now, `c.body` not accept `null`.

PR #3763, opened 2024-12-22
feat(hono/context): contentful status code typing

askorupskyy:
### Contentful status code

Resolves https://github.com/honojs/hono/issues/1518

All in all, prohibits returning `c.json()` or `c.text()` or `c.html()` with 101, 204, 205, and 304 status codes.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -11,7 +11,7 @@ import type {
 } from './types'
 import type { ResponseHeader } from './utils/headers'
 import { HtmlEscapedCallbackPhase, resolveCallback } from './utils/html'
-import type { RedirectStatusCode, StatusCode } from './utils/http-status'
+import type { ContentfulStatusCode, RedirectStatusCode, StatusCode } from './utils/http-status'
 import type { BaseMime } from './utils/mime'
 import type {
   InvalidJSONValue,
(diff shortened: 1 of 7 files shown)


yusukebe:
Hi @askorupskyy 

Thank you for this PR. I'll also review, but @EdamAme-x, can you review this?

yusukebe (approved):
LGTM!

yusukebe:
@askorupskyy Thanks! Looks good to me.


@EdamAme-x Pinging again. Can you review this?

EdamAme-x (approved):
Sorry for late, LGTM.

yusukebe:
@askorupskyy @EdamAme-x

Thanks! I'll merge this now and include this change in the next patch release since this is `feat` change but it's slight.

lewisedc:
I would consider this a breaking change and don't think it should have been pushed with a patch release.

This change is causing all my routes that respond with a `ContentlessStatusCode` (101 | 204 | 205 | 304) to now throw a type error because I am using the `Context.json` function as follows: `return c.json(null, 204);`. This API was specifically made to work this way in issue #242.

The type errors in my code can easily be fixed by updating my routes to use `Context.body` for `ContentlessStatusCode` responses, as follows: return c.body(null, 204);. However, making this change breaks typing in the Hono client, which has been brought up in issue #3798 and, as such, is not a usable solution for me.

bombillazo:
Yeah this is a breaking change, specially since an int value was previously acceptable while now we need to have typed num that is valid for `ContentfulStatusCode`.

luiz00martins:
Agreed, this is a breaking change.

PR #3767, opened 2024-12-24
fix(jsr): exclude unused markdown files

ryuapp:
These may have been included because they could not be linked to files on GitHub before on JSR, but now they are linked correctly like npm. Therefore, it is better to exclude them.

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/jsr.json
+++ b/jsr.json
@@ -110,9 +110,6 @@
       "jsr.json",
       "LICENSE",
       "README.md",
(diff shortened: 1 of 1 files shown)


yusukebe (approved):
LGTM!

yusukebe:
Hi @ryuapp 

Nice! Looks good to me.

PR #3770, opened 2024-12-25
feat(context): `ResponseInit` accepts generics `StatusCode` for `status`

yusukebe:
From merging #3763, `c.json()` or `c.text()` can no longer have a contentless status code as the second argument.

CleanShot 2024-12-26 at 15 24 21@2x

In this PR, if a method such as `c.json()` has `ResponseInit` as its second argument, specifying contentless code for the status property will cause a type error.

CleanShot 2024-12-26 at 15 29 20@2x

This is the correct behavior, and I think this PR should be accepted. However, it causes minor performance degradation in TypeScript, so we need to be careful about it.

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -108,7 +108,7 @@ interface Set<E extends Env> {
  */
 interface NewResponse {
   (data: Data | null, status?: StatusCode, headers?: HeaderRecord): Response
-  (data: Data | null, init?: ResponseInit): Response
+  (data: Data | null, init?: ResponseOrInit): Response
 }
 
 /**
@@ -118,7 +118,8 @@ interface BodyRespond {
   // if we return content, only allow the status codes that allow for returning the body
   (data: Data, status?: ContentfulStatusCode, headers?: HeaderRecord): Response
   (data: null, status?: StatusCode, headers?: HeaderRecord): Response
-  (data: Data | null, init?: ResponseInit): Response
+  (data: Data, init?: ResponseOrInit<ContentfulStatusCode>): Response
+  (data: null, init?: ResponseOrInit): Response
 }
 
 /**
@@ -142,7 +143,7 @@ interface TextRespond {
   ): Response & TypedResponse<T, U, 'text'>
   <T extends string, U extends ContentfulStatusCode = ContentfulStatusCode>(
     text: T,
-    init?: ResponseInit
+    init?: ResponseOrInit<U>
   ): Response & TypedResponse<T, U, 'text'>
 }
 
@@ -173,7 +174,7 @@ interface JSONRespond {
     U extends ContentfulStatusCode = ContentfulStatusCode
   >(
     object: T,
-    init?: ResponseInit
+    init?: ResponseOrInit<U>
   ): JSONRespondReturn<T, U>
 }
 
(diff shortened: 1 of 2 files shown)


src/context.ts
```diff
@@ -257,12 +259,14 @@ type ResponseHeadersInit =
...
 
-interface ResponseInit {
+interface ResponseInit<T extends StatusCode = StatusCode> {
   headers?: ResponseHeadersInit
-  status?: number
+  status?: T
   statusText?: string
 }
 
+type ResponseOrInit<T extends StatusCode = StatusCode> = ResponseInit<T> | Response
```
yusukebe, line 268: It has `Response | ResponseInit` because `ResponseInit` no longer satisfies the type of `Response`.

CleanShot 2024-12-26 at 15 37 15@2x

yusukebe:
Hey @askorupskyy @EdamAme-x 

What do you think of this PR?

askorupskyy:
@yusukebe i like this PR & the code looks good to me. Is there any way to measure the TS performance degradation this can cause?

EdamAme-x:
I think it is a good feature.
However, as you are concerned, type performance may be an issue.

yusukebe:
Thanks for the comments!

You can refer to the`Compiler Diagnostics` on [the comment](https://github.com/honojs/hono/pull/3770#issuecomment-2561824868), but it's difficult to measure the impact of real-world performance.

yusukebe:
Hey, @m-shaka. What do you think of the change in TypeScript performance due to this PR?

askorupskyy:
@yusukebe @EdamAme-x I think the subject of types is a part of a much greater discussion, but to put it simply I think _less inference overall = better_. I think the Hono Context as a concept makes it hard to make Hono scalable enough for the Typescript compiler. In this particular case, if each context method, such as .text() or .json() were to return a set of properties such as `status`, `headers`, `content`, this would be a much simpler and scalable fix

askorupskyy:
That being said, I think this particular PR will not change things much in terms of TS performance, but we have to be careful overall about populating  Hono's context with more TS

yusukebe:
Speaking of this change, I think it is necessary to make `ResponseInit` behave in the same way as #3770. So I think it is okay to merge it, ignoring the performance issue.

As for the performance of TypeScript, it is ideal to optimize it as a separate issue.

askorupskyy:
@yusukebe makes sense. From what I can say from the performance metrics above, the numbers we have to pay attention to most is `# of Types` and `Check Time`. In this PR the diff for those is minimal so I suggest we merge this.

m-shaka:
Our test case doesn't use the second argument of `json`, so I reckon it's not a suitable one to check the impact of this change

https://github.com/honojs/hono/blob/b1335f9fb296cfa4b947367bbad0512c7739c5f5/perf-measures/type-check/scripts/generate-app.ts#L8-L18

m-shaka:
I added `{ status: 200 }` to handlers of the test case on my machine and here are the results, which look ok.

```
// main
Files:              256
Lines:           115062
Identifiers:     112327
Symbols:         244896
Types:           206318
Instantiations: 3047508
Memory used:    429779K
I/O read:         0.02s
I/O write:        0.00s
Parse time:       0.33s
Bind time:        0.12s
Check time:       2.97s
Emit time:        0.00s
Total time:       3.42s

// this branch
Files:              256
Lines:           115035
Identifiers:     112200
Symbols:         245525
Types:           206794
Instantiations: 3048953
Memory used:    425269K
I/O read:         0.02s
I/O write:        0.00s
Parse time:       0.33s
Bind time:        0.13s
Check time:       3.04s
Emit time:        0.00s
Total time:       3.50s
```

`Instantiations` increased but just slightly

yusukebe:
@askorupskyy @EdamAme-x @m-shaka Thank you!

Let's go with this. Merging now.

PR #3773, opened 2024-12-26
fix(hono-base): don't use Symbol for `COMPOSED_HANDLER`

### LINKED ISSUE #3760: `COMPOSED_HANDLER` mismatch when using different version of `hono`

4.6.14

any

Sorry to reproduce and explain this problem is super difficult.

For example, this problem occurs when using the SSG helper. Suppose there are two projects in a monorepo: one called `app`, which defines the application, and one called `ssg`, which is used for SSG. If they are both using the same version of `hono`, there will be no problem. However, if they are using different versions, for example, `app` is using `4.6.13` and `ssg` is using `4.6.14`, the SSG helper will not generate the files.

Repro: https://github.com/yusukebe/composed-handler-mismatch

CleanShot 2024-12-20 at 20 01 11@2x

This happens when `onError` is defined in `subApp` within the `app` application.

```ts
// projects/app/src/index.ts
import { Hono } from 'hono'

const subApp = new Hono()

subApp.get('/', (c) => c.html('Hi'))
subApp.onError((_e, c) => c.html('Error', 500))

const app = new Hono()
app.route('/', subApp)

export default app
```

This is because the handler to handle `GET /` request is marked as middleware not as a handler in the following code:

https://github.com/honojs/hono/blob/4eb101dd03e66e20dfda86e6e5ebce8045d61d05/src/helper/ssg/utils.ts#L65-L68

The reason is that the Symbol `COMPOSED_HANDLER` is recognized as a different key if the version is different in the following locations.

https://github.com/honojs/hono/blob/4eb101dd03e66e20dfda86e6e5ebce8045d61d05/src/utils/handler.ts#L9-L15

`COMPOSED_HANDLER` is defined in `hono-base.ts`.

https://github.com/honojs/hono/blob/4eb101dd03e66e20dfda86e6e5ebce8045d61d05/src/hono-base.ts#L34

If you change the value of this `COMPOSED_HANDLER` to a string, this problem will be solved.

```ts
export const COMPOSED_HANDLER = '__COMPOSED_HANDLER'
```

### Should it be a Symbol?

How about using a string like `__COMPOSED_HANDLER` instead of the Symbol? I think it is not a problem.

### Is this a rare case?

Yes, it's a rare case. But this causes the issues: https://github.com/honojs/honox/issues/234 https://github.com/honojs/honox/issues/236 in HonoX. The issue is not caused by a difference in the version of `hono`, but it has deep nested dependencies. That will be resolved if making it string.

yusukebe:
Fixes #3760

This PR resolves the problem described in #3760. It uses a string value for `COMPOSED_HANDLER` instead of a Symbol. We should add a test for a `fix` change, but in this PR, it's challenging to reproduce the problem in #3760, so it doesn't include a test.

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/src/hono-base.ts
+++ b/src/hono-base.ts
@@ -26,12 +26,8 @@ import type {
   RouterRoute,
   Schema,
 } from './types'
+import { COMPOSED_HANDLER } from './utils/constants'
 import { getPath, getPathNoStrict, mergePath } from './utils/url'
-
-/**
- * Symbol used to mark a composed handler.
- */
-export const COMPOSED_HANDLER = Symbol('composedHandler')
 
 const notFoundHandler = (c: Context) => {
   return c.text('404 Not Found', 404)
--- a/src/utils/constants.ts
+++ b/src/utils/constants.ts
@@ -1,0 +1,4 @@
+/**
+ * Constant used to mark a composed handler.
+ */
+export const COMPOSED_HANDLER = '__COMPOSED_HANDLER'
--- a/src/utils/handler.ts
+++ b/src/utils/handler.ts
@@ -3,7 +3,7 @@
  * Handler utility.
  */
 
-import { COMPOSED_HANDLER } from '../hono-base'
+import { COMPOSED_HANDLER } from './constants'
 
 export const isMiddleware = (handler: Function) => handler.length > 1
 export const findTargetHandler = (handler: Function): Function => {


yusukebe:
Hey @usualoma. Can you review this?

yusukebe:
I think merging this without waiting for others to review is OK. If you have any suggestions, feel free to create a PR to resolve them. I'll merge now.

Issue #3765, opened 2024-12-24
SSE does not work again... 


damianpumar:
### What version of Hono are you using?

latest

### What runtime/platform is your app running on? (with version if possible)

node

### What steps can reproduce the bug?

The sse does not work again in the latest version...
I feel super frustrated after spending 1 hour and not understanding what was happening with my code, when I moved to the version "@hono/node-server": "1.8.2" and  "hono": "4.1.1", it works correctly, I have already fixed this same error in the past here https://github.com/honojs/hono/commit/b827efb3fb3fa988e74a5b96d1cfcd75d415f2af

### What is the expected behavior?

SSE should works

### What do you see instead?

-

### Additional information

-

yusukebe:
Hi @damianpumar 

Can you fix it? I can't understand that behavior only with your information.

damianpumar:
Yes, I can fix it again, but it was fixed by me some month ago... 
Basically, using the latest version, the SSE server does not work...

yusukebe:
This issue is not described well and reproduced. We can close this.

PR #2529, opened 2024-04-18
fix(sse): close sse stream on end

domeccleston:
Adds a finally block to the SSEStream class. Otherwise streamSSE will error whenever used in Cloudflare Workers.

PR #3779, opened 2024-12-27
fix: :fire: Fix closed event source

damianpumar:
Fix this issue: https://github.com/honojs/hono/issues/3765

The main idea of ​​an SSE is to keep the stream open, if we close it we lose the connection with our clients.

Just in case, I have a repo here, with latest version doesn't work... https://github.com/damianpumar/vote please take a look if you want, and also you can ping me on discord if you have any doubt about my fix!

Thanks!

### DIFF
--- a/src/helper/streaming/sse.ts
+++ b/src/helper/streaming/sse.ts
@@ -55,8 +55,6 @@ const run = async (
     } else {
       console.error(e)
     }
-  } finally {
-    stream.close()
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/helper/streaming/sse.ts
```diff
@@ -55,8 +55,6 @@ const run = async (
     } else {
       console.error(e)
     }
-  } finally {
-    stream.close()
```
yusukebe, line 59: Removing these lines means you ignore the following PR.

https://github.com/honojs/hono/pull/2529

PR #3786, opened 2024-12-31
feat(factory): Allow HonoOptions<E> with factory

miyaji255:
When using factory, options such as `strict: false` cannot be specified. This Pull Request allows you to specify options when calling `createFactory` and `createApp`.

```ts
const factory = createFactory({ defaultOptions: { strict: false } });
const app = factory.createApp({ options: { strict: true } }); // Override defaultOptions
```

- [x] Add tests
- [X] Run tests

### DIFF
--- a/src/helper/factory/index.ts
+++ b/src/helper/factory/index.ts
@@ -5,6 +5,7 @@
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { Hono } from '../../hono'
+import type { HonoOptions } from '../../hono-base'
 import type {
   Env,
   H,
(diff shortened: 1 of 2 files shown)


src/helper/factory/index.ts
```diff
@@ -285,13 +286,19 @@ export interface CreateHandlersInterface<E extends Env, P extends string> {
 
 export class Factory<E extends Env = any, P extends string = any> {
   private initApp?: InitApp<E>
+  #defaultOptions?: HonoOptions<E>
 
-  constructor(init?: { initApp?: InitApp<E> }) {
+  constructor(init?: { initApp?: InitApp<E>; defaultOptions?: HonoOptions<E> }) {
```
yusukebe, line 291: `Factory` does not provide only `createApp()` but it provides `createMiddleware()` and `createHandlers()`. So the name `defaultOptions` may introduce confusion because the user may understand the option is not for `createApp()`. So I think it should have a different name, such as `defaultAppOptions`.

yusukebe:
Hi @miyaji255 

Thank you for the PR. I think this is a good feature. I've left the comment.

src/helper/factory/index.ts
miyaji255: Thank you for pointing that out. I completely overlooked the fact that it's not just about the `app`.

yusukebe (approved):
LGTM!

yusukebe:
Hi @miyaji255 

Sorry for the delayed review. Looks good! I wonder if we should include this in a patch release or minor release, but I'll decide on and merge it later. Thanks!

PR #3792, opened 2024-12-31
fix(jsx/dom): should not return memoized result when context is changed

### LINKED ISSUE #3778: [jsx] Hydrated useContext does not receive new value from context with HonoX

4.6.14

Bun, but it's probably runtime independent

1. Use `create-hono` and choose `x-basic`
2. Prepare provider and consumer as island components
    - Define the value passed by the context containing a value and an updater function, similar to the return value of useState.
4. Interact with rendered page

When the value passed to the Provider is updated, the value received by its descendant useContext will be in an updated state.

### What do you see instead?

When the value passed to the Provider is updated, the value received by its descendant useContext will remain the same as the first.

### Additional information

It does not occur with Hono 4.4, but occurs with Hono 4.5.0-rc.2 or later.
Occurs only when using with HonoX, but depends on the Hono version, so write here.
repro: https://berlysia.github.io/hono-jsx-context-hydration-issue/

I also tried with React renderer, it works as expected

usualoma:
fixes #3778

In v4.5 and later versions, nested function components were over-cached and not re-rendered even when the context was updated. This PR will fix this bug.

- [x] Add tests
- [x] Run tests


usualoma:
@yusukebe 
Happy New Year! I hope that this year will be a good one for Hono.

yusukebe (approved):
LGTM!

yusukebe:
Hey @usualoma

Happy New Year! 🎍 It should be a good year! Looks good. Merging.

Issue #3720, opened 2024-12-04
How to return Node.js `Buffer` with TypeScript 5.7?


yusukebe:
### Discussed in https://github.com/orgs/honojs/discussions/3700

<div type='discussions-op-text'>

<sup>Originally posted by **marvinruder** November 24, 2024</sup>
The code

```ts
import { serve } from '@hono/node-server'
import { Hono } from 'hono'
import { etag } from 'hono/etag'

const app = new Hono()
app.use(etag())
app.get('/', (c) => c.body(Buffer.from('Hello Hono!')))
serve({ fetch: app.fetch })
```

runs as expected and shows no errors in TypeScript 5.6, but the following error is shown with TypeScript 5.7:

```
No overload matches this call.
  Overload 1 of 2, '(data: Data | null, status?: StatusCode | undefined, headers?: HeaderRecord | undefined): Response', gave the following error.
    Argument of type 'Buffer<ArrayBuffer>' is not assignable to parameter of type 'Data | null'.
      Type 'Buffer<ArrayBuffer>' is missing the following properties from type 'ArrayBuffer': maxByteLength, resizable, resize, detached, and 2 more.
  Overload 2 of 2, '(data: Data | null, init?: ResponseInit | undefined): Response', gave the following error.
    Argument of type 'Buffer<ArrayBuffer>' is not assignable to parameter of type 'Data | null'.
      Type 'Buffer<ArrayBuffer>' is missing the following properties from type 'ArrayBuffer': maxByteLength, resizable, resize, detached, and 2 more.
```

This is because a `Buffer` is no longer an `ArrayBuffer` (see https://devblogs.microsoft.com/typescript/announcing-typescript-5-7/#typedarrays-are-now-generic-over-arraybufferlike, https://github.com/microsoft/TypeScript/pull/59417 for details), and the `Data` type can only be a `string`, `ArrayBuffer`, or `ReadableStream`.

Returning the `ArrayBuffer` within the `Buffer` using

```ts
app.get('/', (c) => c.body(Buffer.from('Hello Hono!').buffer))
```

shows no type error, but leads to unexpected results, as responses are 8192 bytes long and contain lots of unrelated data. The reason for this is the shared memory pool from the `Buffer` module:

[...]

PR #3813, opened 2025-01-08
feat(hono/context): add buffer returns

### LINKED ISSUE #3729: Missing `Buffer<ArrayBufferLike>` type in the parameter of `c.body`

4.6.12

Node.js 22

```ts
import { readFileSync } from 'node:fs'
import { Hono } from 'hono'

const app = new Hono()

const ZIP_NAME = '<FILENAME>.zip'
const ZIP_FILEPATH = '/' + ZIP_NAME

app.get('/downloadZip', (c) => {
  c.header('Content-Disposition', `attachment; filename="${ZIP_NAME}"`)
  c.header('Content-Type', 'application/zip-compressed')

  return c.body(readFileSync(ZIP_FILEPATH))
})

const port = 3000

serve(
  {
    fetch: app.fetch,
    port
  },
  () => console.log(`Server running at http://localhost:${port}`)
)
```

`readFileSync(zipPath)` is of type `Buffer<ArrayBufferLike>`.
`c.body` accept data having `string | ArrayBuffer | ReadableStream` types.

Maybe `Buffer<ArrayBufferLike>` type should be added to data parameter of `c.body`.

No TypeScript error

### What do you see instead?

TypeScript error:

```console
  Overload 1 of 2, '(data: Data | null, status?: StatusCode | undefined, headers?: HeaderRecord | undefined): Response', gave the following error.
    Argument of type 'Buffer<ArrayBufferLike>' is not assignable to parameter of type 'Data | null'.
      Type 'Buffer<ArrayBufferLike>' is missing the following properties from type 'ArrayBuffer': maxByteLength, resizable, resize, detached, and 2 more.
  Overload 2 of 2, '(data: Data | null, init?: ResponseInit | undefined): Response', gave the following error.
    Argument of type 'Buffer<ArrayBufferLike>' is not assignable to parameter of type 'Data | null'.
      Type 'Buffer<ArrayBufferLike>' is missing the following properties from type 'ArrayBuffer': maxByteLength, resizable, resize, detached, and 2 more.ts(2769)
```

### Additional information

Workaround:
```diff
-return c.body(readFileSync(ZIP_FILEPATH))
+return c.body(readFileSync(ZIP_FILEPATH) as unknown as ReadableStream)`
```

Maybe related:
- #935
- https://github.com/honojs/node-server/pull/63

### Question
There is any better/right alternative to download a _zip_ file from browser?

askorupskyy:
# Context

I needed to be able to return Buffers from my Hono app, and was not able to do so.

Resolves #3729 #3720

- [x] Add tests
- [x] Run tests


askorupskyy:
Tested this PR with the sample repo, using `typescript@5.7` – seems to be working fine

Screenshot 2025-01-08 at 6 18 47 PM

src/context.ts
```diff
@@ -1,3 +1,5 @@
+import type { Buffer } from 'node:buffer'
```
yusukebe, line 1: Is this `import type` necessary? In my environment, it does not throw an error without this.

askorupskyy: I added this for Deno. Deno does not have a Buffer exposed to the global scope

yusukebe: It is importing just types, but I think we should not use `node:*` modules in the core of `hono`. Actually, `Buffer` is not in Web Standards API.

askorupskyy: I think using `node:buffer` is fine here, especially as a Type.  There's no way to polyfill types for specific platforms as far as I am aware (so Deno will continue to fail). In addition to that, `node:buffer` is either the default implementation for a lot of runtimes, such as Cloudflare Workers, Deno, etc, or is already defined as a global replicating Node.js API. I see no problem with using the type from Node, since all runtimes are pretty much using the same API as node.

EdamAme-x: It might be a good option to have a helper function to convert from `node:buffer` to the web standard accepted body format (For example, `ReadbleStream`).
If you think it is a good idea, we can change the implementation policy of this PR.

askorupskyy: Let me see if i can make this work with a uint8array

askorupskyy:
@yusukebe @EdamAme-x looks like `Buffer` indeed is a Node-only construct (they made it before Uint8Array was invented), so to be compliant with Web Standards I moved to `Uint8Array`, for which the Buffer is a superset. Added one more test and it looks like everything works

EdamAme-x (approved):
a

EdamAme-x:
One thing I was wondering, is there a reason why `URLSearchParams` and `FormData` are not accepted?

[TypeScript Playground](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgCQgOwnAvnAZlCEOAIgAt0JiAoKgY3QGd4BDMMOAXjjQFMB3FBQAUAShqswAOgDmPGENqcAfIipx1cKHICuUNHFqSARhAAmATyG8BAVQBKAGQDKPZlFqkACm+YgGQgG01DRCA4gZtWloeBgZiABo4GChtHgBdYPURMSwRIA)

askorupskyy:
@EdamAme-x first of all, `BodyInit` is what's used by `ClientRequestImpl`, which is part `hc`. That's to parse whatever input you give to `hc` – this is why your second example doesn't work. I don't think BodyInit matters in the context of this PR

askorupskyy:
@yusukebe can you review this once again pls?

yusukebe (approved):
LGTM!

yusukebe:
@askorupskyy 

Looks good to me! We can merge this. I'm considering whether this change should be included in a patch release or a minor release. I'll decide later. Thanks!

PR #3819, opened 2025-01-11
feat(types): Add DefaultEnv for zero runtime factory

### LINKED ISSUE #3793: Zero Runtime Factory
### Description
Since it is common for projects to use a single `Env` type, I propose using [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) as an alternative approach.

I believe this change can simplify type management and improve the developer experience, and this approach can also support `presets`.

### Proposed Method
To address this, I propose modifying the `Hono` class as follows:

```diff
- class Hono<E extends Env = Env, ...>
+ class Hono<E extends Env = DefaultEnv, ...>

+ interface DefaultEnv { }
+ interface EnvBindings { }
+ interface EnvVariables { }
```

This allows developers to define and extend `DefaultEnv` for their projects without needing to specify `Env` types in every instance:

```ts
declare namespace "hono" {
  interface DefaultEnv {
    Bindings: EnvBindings;
    Variables: EnvVariables;
  }
  interface EnvBindings {
    lambdaContext: LambdaContext;
  }
  interface EnvVariables {
    foo: string;
  }
}

const app = new Hono();
const middleware: MiddlewareHandler = (c, next) => ...
```

### Example for AWS Lambda
For example, adapters could automatically set `Bindings` when using AWS Lambda:

```ts
// aws-lambda/handler.ts
declare namespace "hono" {
  interface DefaultEnv {
    Bindings: EnvBindings
  }
  interface EnvBindings {
    event: LambdaEvent
    lambdaContext: LambdaContext
  }
}
export const handler ...
```

miyaji255:
Resolves #3793

This is BREAKING CHANGE.

I specified `DefaultEnv` for those that take `Env` as a type argument. By extending `DefaultEnv` with interface merging, you can change the default `Env` for various types.

This eliminates the need to use hono/factory in simple cases. However, if multiple `Bindings` are used, they must be explicitly specified.

```ts
interface DefaultEnv {}
interface DefaultBindings {}
interface DefaultVariables {}

export class Hono<E extends Env = DefaultEnv> {
```

### Usage

By extending DefaultEnv, library users do not need to specify it explicitly.

```ts
// library
declare module "hono" {
  interface DefaultEnv {
    Bindings: DefaultBindings;
  }
  interface DefaultBindings {
    Foo: string;
  }
}

export function handle(app: Hono<any>): ...
```

You no longer need to explicitly specify Env.

```ts
import { Hono } from "hono";
import { handle } from "hono/aws-lambda/api-gateway";

const middleware: MiddlewareHandler = (c, next) => {
  c.env.event // APIGatewayProxyEvent
  c.env.lambdaContext // LambdaContext
  return next();
}

const app = new Hono()
  .get("/", c => {
    c.env.event // APIGatewayProxyEvent
    c.env.lambdaContext // LambdaContext
    return c.text(c.env.event.path);
  });
export default handle(app);
```


yusukebe:
Hi @miyaji255 

Thank you for the PR! I also think it would be good if we could set the *env* with `declare module 'hono'`. I'll start to consider this feature.

miyaji255:
Thank you very much. I will also think about whether I can fix errors and design a better API.

PR #3824, opened 2025-01-13
fix(helper/factory): Reduce the code size of createMiddleware

miyaji255:
`createMiddleware` depends on `Hono`, so `hono/tiny` users cannot use `createMiddleware` and libraries which use `createMiddleware`. `createMiddleware` should not use `Factory`

`createMiddleware` -> `Factory` -> `Hono` (`hono/hono`)

### Libraries which use `createMiddleware`

- @hono/bun-transpiler
- @hono/cloudflare-access
- @hono/esbuild-transpiler
- @hono/oidc-auth
- @hono/prometheus
- @hono/tsyringe

### DIFF
--- a/src/helper/factory/index.ts
+++ b/src/helper/factory/index.ts
@@ -316,4 +316,4 @@ export const createMiddleware = <
   I extends Input = {}
 >(
   middleware: MiddlewareHandler<E, P, I>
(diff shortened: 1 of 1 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@miyaji255 

You are right! The fix is good. I'll merge this into the main. Thank you!

Issue #3658, opened 2024-11-11
Support JWKs validation


JoaquinGimenez1:
### What is the feature you are proposing?

JSON Web Key Sets are becoming more and more common these days. While Hono offers a JWT validation middleware it would be very nice to extend the current middleware to also validate JWTs that are signed with a public key. This should help any hono application that uses Cloudflare Zero Trust Access/Okta/other solutions. 

Cloudflare Ref: https://developers.cloudflare.com/cloudflare-one/identity/authorization-cookie/validating-json/

G4brym:
This was just released [@hono/cloudflare-access](https://www.npmjs.com/package/@hono/cloudflare-access)

Issue #2589, opened 2024-05-02
Support for JWT with JWKS


qoomon:
### What is the feature you are proposing?

currently the `jwt` middleware requires a static secret value, however in my use case (see below) it's not a static secret instead it is retrieved from an JWKS. Therefore it would be nice if the `jwt` middleware also support an `async function(payload: any) : string` as secret parameter.

Use Case:

I'm working on a server endpoint that is called by GitHub actions to authenticate I'm using the Github Actions OIDC tokens.

Currently I'm using my own middleware by using `fast-jwt` with `get-jwks` like this
```typescript
import {createVerifier, DecodedJwt, KeyFetcher} from 'fast-jwt'
import buildJwks from 'get-jwks'

const jwks = buildJwks({providerDiscovery: true})
const keyFetcher: KeyFetcher = async (jwt: DecodedJwt) => jwks.getPublicKey({
  kid: jwt.header.kid,
  alg: jwt.header.alg,
  domain: jwt.payload.iss,
})

const jwtVerifier = createVerifier({
  key: keyFetcher,
})

app.use(async (c, next) => {
  // ... get jwtTokenString value
  await jwtVerifier(jwtTokenString)
  // ... handle verification errors
  await next()
})


```

however it would be nice if I can switch to an official middleware.

qoomon:
probably related to https://github.com/honojs/hono/issues/672

sparanoid:
[Apple Music API](https://developer.apple.com/documentation/applemusicapi) also needs `kid` header which is not compatible with current jwt middleware:

expected headers and payload:

```json
{
     "alg": "ES256",
     "kid": "ABC123DEFG"
}
{
     "iss": "DEF123GHIJ",
     "iat": 1437179036,
     "exp": 1493298100
}
```

actual:

```json
{
     "alg": "ES256",
     "typ": "JWT"
}
{
     "iss": "DEF123GHIJ",
     "iat": 1437179036,
     "exp": 1493298100
}
```

maxpetretta:
also have need for `kid` in headers, only reason i have to stick with a separate jwt library now

ngdbao:
It's better if we can custom header before signing jwt

Issue #672, opened 2022-11-21
Support for asymmetric JWT validation (e.g. Auth0 or Clerk)


cd-slash:
The existing JWT middleware validates tokens where the secret is available (i.e. symmetric validation) but services like Auth0, Clerk and others use asymmetric tokens signed by the service, and not made available for validating the token. These tokens need to be validated with the public key.

Is this functionality that should be supported in Hono? It seems like it would be very useful for anyone who wants to build an API accessible by end users.

I put together a quick proof of concept middleware for Clerk tokens, which grabs the public key(s), caches them in a Cloudflare workers KV and uses them to validate the token. The token's claims are made available to the context using the `c.set()` function. This code could be refined if it's something that should be supported, and note that the code as drafted introduces a dependency on the jose library for key import and validation.

```ts
import type { MiddlewareHandler } from "hono"
import { Context } from "hono"
import { KeyLike, importJWK, jwtVerify } from "jose"

const getPublickKeys = async (
  c: Context
): Promise<{ keys: KeyLike[] } | null> => {
  const cache: KVNamespace = c.env.PUBLIC_JWK_CACHE
  const cachedKeys: string | null = await cache.get("public-keys")
  if (cachedKeys != null) {
    return JSON.parse(cachedKeys)
  } else {
    try {
      const res = await fetch(`${c.env.AUTH_DOMAIN}/.well-known/jwks.json`)
      const freshKeys: { keys: KeyLike[] } = await res.json()
      try {
        await cache.put("public-keys", JSON.stringify(freshKeys), {
          expirationTtl: 432000,
        })
      } catch (e) {
        // don't fail as the validation can still occur if keys can't be cached
        console.log("Error caching public keys: ", e)
      }
      return freshKeys
    } catch (e) {
      console.log("Error getting fresh keys: ", e)
      return null
    }
  }
}

export const asymmetricJwt = (): MiddlewareHandler => {
  return async (c, next) => {
    let jwt: string
    let errors: string[] = []
    const authorization = c.req.header("Authorization")
    if (authorization == null) {
      return c.json({ error: "No auth token found" }, 401)
    } else {
      jwt = authorization.replace(/Bearer\s+/i, "")
    }

    // get public keys from cache, or get new keys and cache them (5 day TTL)
    const publicKeys = await getPublickKeys(c)
    if (publicKeys == null) {
      return c.json({ error: "Failed to fetch public keys" }, 500)
    }

    const [token] = await Promise.all(
[...]

PR #3826, opened 2025-01-13
feat(hono/jwk): JWK Auth Middleware

Beyondo:
Hello, I recently needed a JWK middleware for my projects, but I figured contributing it to hono has the potential to save me and others a lot of time.

### Middleware Features:
- Set `options.keys` to a static array of public keys `HonoJsonWebKey[]` in code.
- Set `options.keys` to an async function that returns a `Promise<HonoJsonWebKey[]>` for flexibility
- Set `options.jwks_uri` to fetch JWKs from a URI, after which it appends those fetched keys to provided `keys` if any
- Set an optional `init` parameter (only used for `jwks_uri`)—useful if your host supports caching through custom init options.

#### Added extra:
- Added `JwtHeaderRequiresKid` exception. Since the middleware requires presence of a `kid` field in the header in order to select the correct key.
- Added `Jwt.verifyFromJwks` util function _(batteries included)_.

### Addressed issues:
- https://github.com/honojs/hono/issues/3658
- https://github.com/honojs/hono/issues/2589
- https://github.com/honojs/hono/issues/672

### Other code changes:
- Typescript-extended `JsonWebKey` to have `kid?: string` (This is a standard: https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.4)
- Added `kid?: string` to TokenHeader
- Fixed a bug in `Jwt.sign(payload, privateKey, alg)` where `privateKey.alg` was always ignored and only `alg` parameter was considered.

**Suggestion:** Removal of the default `'HS256'` value from `Jwt.sign` and make `alg` optional, so that if neither `privateKey.alg` nor `alg` is defined, throw an error. My 'philosophy' for this is that it would be misleading for hono to display `= 'HS256'` when the privateKey can internally have a different algorithm, such as when it is of type `JsonWebKey` or `CryptoKey` (Though we only care about `JsonWebKey` here). Making `alg` a "fallback" explicit parameter _could_ be better.

#### Using `jwks_uri`
```ts
import { jwk } from "hono/jwk";
app.use("/backend-api/*", jwk({ jwks_uri: "https://example-backend.hono.dev/.well-known/jwks.json" }));
```

#### Using `jwks_uri` and an optional `init`
```ts
import { jwk } from "hono/jwk";
app.use("/backend-api/*", jwk({
  jwks_uri: "https://example.hono.dev/.well-known/jwks.json"
}, {
  headers: {
    "My-Custom-Header": "Hello Hono!"
  },
  cf: { cacheTtl: 3600, cacheEverything: true } // Only for Cloudflare Workers
}));
```

#### Using `keys` from an array
```json
[
  {
    "kid": "VpNx_Ar8pXsh50UjqBD_XtYLj4k9Z72iakeOQFJYIpg",
    "kty": "RSA",
    "use": "sig",
    "alg": "RS256",
    "e": "AQAB",
    "n": "4anIpNGZ-C5rJfRty8qQt8wTheFDj9wchPA..."
  },
  {
    "kid": "Rw6zp1oTEbQ9qUIFnhHGadtstXhKYrCmkqgkNEG8uUc",
    "kty": "RSA",
    "use": "sig",
    "alg": "RS256",
    "e": "AQAB",
    "n": "vKetN8i4_ORAWAUVR1pCjVMFkNzwC1Qjy..."
  }
]
```
```ts
import keys from './keys.json';
import { jwk } from "hono/jwk";
app.use("/backend-api/*", jwk({ keys: keys }));
```

#### Using `keys` as an async function (custom logic / caching)
```ts
$router.use("/auth/*", jwk({
  keys: async () => {
    /* ... Custom caching logic ... */
    return cached_response.keys;
  }
}));
```

### Tests:

I adapted a lot from JWT's own units tests to make sure some basic JWT validation exists, then added the actual validation for the JWK middleware's own functionality which is mostly: (1) getting keys through various methods, (2) selecting the appropriate JWK based on `kid` (without fully decoding the JWT—only the header), and then (3) verifying using the utility `Jwt.verify`. Step 2 and 3 are done by `Jwt.verifyFromJwks`.

#### JWK Middleware Test
```
hono> npx src/middleware/jwk/index.test.ts

 DEV  v2.1.1 V:/Repos/hono
      Coverage enabled with v8

 ✓ src/middleware/jwk/index.test.ts (16)
   ✓ JWK (16)
     ✓ Credentials in header (10)
       ✓ Should not authorize requests with missing access token
       ✓ Should authorize from a static array passed to options.keys (key 1)
       ✓ Should authorize from a static array passed to options.keys (key 2)
       ✓ Should authorize with Unicode payload from a static array passed to options.keys
       ✓ Should authorize from a function passed to options.keys
       ✓ Should authorize from a URI remotely fetched from options.jwks_uri
       ✓ Should not authorize requests with invalid Unicode payload in header
       ✓ Should not authorize requests with malformed token structure in header
       ✓ Should not authorize requests without authorization in nested JWK middleware
       ✓ Should authorize requests with authorization in nested JWK middleware
     ✓ Credentials in cookie (5)
       ✓ Should not authorize requests with missing access token
       ✓ Should authorize from a static array passed to options.keys
       ✓ Should authorize with Unicode payload from a static array passed to options.keys
       ✓ Should not authorize requests with invalid Unicode payload in cookie
       ✓ Should not authorize requests with malformed token structure in cookie
     ✓ Error handling with `cause` (1)
       ✓ Should not authorize

 Test Files  1 passed (1)
      Tests  16 passed (16)
```

#### Full Test
```
hono> npm test
...
...
...
 Test Files  9 passed (9)
      Tests  415 passed | 3 skipped (418)
   ```
---

- [✅] Add tests (src/middleware/jwk/index.test.ts)
- [✅] Run tests (Above)
- [✅] `bun run format:fix && bun run lint:fix` to format the code (Done)
- [✅] Add [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code (src/middleware/jwk/jwk.ts)

---

Feel free to contribute more unit tests, extend and/or modify for the better.

### CODE REVIEW

**Comment by Beyondo** [comment]
Note that the test code `jwk/index.test.ts` was 519 lines so it inflated the actual aggregated changes of this PR, + jsdoc comments, and + the `keys.test.json` file and so on. The **core changes** are probably < 200 lines and thus much more mangeable to review.

I removed some commented out tests to give a clearer picture (Now at 307 lines of test code). But they did remind me—we need some **_signed cookie_ unit tests** if we're gonna make sure to support everything in `jwt/index.test.ts` too. Though I don't think it's exactly a priority and could be implemented later.

**Comment by yusukebe** [comment]
Hi @Beyondo 

Thanks! I'll check this later.

**Comment by Beyondo** [comment]
@yusukebe No problem! I originally named the new interface `ExtendedJsonWebKey`, though I do think `HonoJsonWebKey` might also work better since the only addition is the optional `kid?`, making it function just like the normal `JsonWebKey` if needed, so we might need some opinion on this.

Take your time! I'm using the fork temporarily, so no rush at all.

**Comment by yusukebe** [comment]
Hey @Code-Hex ! I'll review this later, but can you take a look at it too?

File: src/middleware/jwk/index.test.ts [hunk 1]
```diff
@@ -0,0 +1,306 @@
...
+import { Jwt } from '../../utils/jwt'
+import * as test_keys from './keys.test.json'
+import { jwk } from '.'
+
+const verify_keys = test_keys.public_keys
+
+describe('JWK', () => {
+  const resource_server = new Hono()
+  resource_server.get('/.well-known/jwks.json', (c) => c.json({ keys: verify_keys }))
+  const server = serve({ fetch: resource_server.fetch })
```
yusukebe, line 14: It's better not to use an actual server. For example, if you run another server on the `3000` port, the test will fail since the port is a collision. Using `msw` to mock the HTTP server that is used in `src/client/client.test.ts` is good:

```diff
diff --git a/src/middleware/jwk/index.test.ts b/src/middleware/jwk/index.test.ts
index 528b8216..b8e90646 100644
--- a/src/middleware/jwk/index.test.ts
+++ b/src/middleware/jwk/index.test.ts
@@ -1,5 +1,5 @@
-import { serve } from '@hono/node-server'
-import type { AddressInfo } from 'net'
+import { HttpResponse, http } from 'msw'
+import { setupServer } from 'msw/node'
 import { Hono } from '../../hono'
 import { HTTPException } from '../../http-exception'
 import { Jwt } from '../../utils/jwt'
@@ -9,10 +9,14 @@ import { jwk } from '.'
 const verify_keys = test_keys.public_keys

 describe('JWK', () => {
-  const resource_server = new Hono()
-  resource_server.get('/.well-known/jwks.json', (c) => c.json({ keys: verify_keys }))
-  const server = serve({ fetch: resource_server.fetch })
-  const port = (server.address() as AddressInfo).port
+  const server = setupServer(
+    http.get('http://localhost/.well-known/jwks.json', () => {
+      return HttpResponse.json({ keys: verify_keys })
+    })
+  )
+  beforeAll(() => server.listen())
+  afterEach(() => server.resetHandlers())
+  afterAll(() => server.close())

   describe('Credentials in header', () => {
     let handlerExecuted: boolean
@@ -33,7 +37,7 @@ describe('JWK', () => {
       '/auth-with-keys-fn/*',
       jwk({
         keys: async () => {
-          const response = await fetch(`http://localhost:${port}/.well-known/jwks.json`)
+          const response = await fetch('http://localhost/.well-known/jwks.json')
           const data = await response.json()
           return data.keys
         },
@@ -42,7 +46,7 @@ describe('JWK', () => {
     app.use(
       '/auth-with-jwks_uri/*',
       jwk({
-        jwks_uri: `http://localhost:${port}/.well-known/jwks.json`,
+        jwks_uri: 'http://localhost/.well-known/jwks.json',
       })
     )
```

Code-Hex:
@yusukebe I'm sorry I missed this PR.
I will check tomorrow 🙏

src/middleware/jwk/index.test.ts
Beyondo: Good point! I tried to use a dynamic port to workaround the collision issue, but after applying `msw`, it is definitely much cleaner for our simple resource server

src/middleware/jwk/index.ts
```diff
@@ -0,0 +1,2 @@
+export { jwk } from './jwk'
+import type {} from '../..'
```
Code-Hex, line 2:
```suggestion
```

src/middleware/jwk/jwk.ts
```diff
@@ -0,0 +1,171 @@
...
+        token = getCookie(ctx, options.cookie)
+      } else if (options.cookie.secret) {
+        if (options.cookie.prefixOptions) {
+          token = await getSignedCookie(
+            ctx,
+            options.cookie.secret,
+            options.cookie.key,
+            options.cookie.prefixOptions
+          )
+          console.log('TOK', token, options, ctx.req.raw.headers)
```
Code-Hex, line 86:
```suggestion
```

src/middleware/jwk/jwk.ts
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    let keys = typeof options.keys === 'function' ? await options.keys() : options.keys
+
+    if (options.jwks_uri) {
+      const response = await fetch(options.jwks_uri, init)
+      if (!response.ok) {
+        throw new Error(`failed to fetch JWKS from ${options.jwks_uri}`)
+      }
+      const data = (await response.json()) as { keys?: JsonWebKey[] }
```
Code-Hex, line 118: I think you should really add validation, but what do you think?

Code-Hex (commented):
It was mostly LGTM!

src/middleware/jwk/jwk.ts
Beyondo: Whoops; thanks for catching that 💯

src/middleware/jwk/jwk.ts
Beyondo: Just added a small validation to check if `keys` is an **array**—though iterating through **each key** in that array to validate is pointless (as-in stateless backends will have to re-do it every request). For example, missing fields like `e` or `alg` would not be a problem as that means it would then be a _**different key**_, always resulting in **401** in `Jwt.verify`, while extra fields/metadata are always **ignored** in `Jwt.verify`. Wrong types/values (e.g., `alg: 951624`) also cause **401** in `Jwt.verify`.

So based on that, the internal validation with `Jwt.verify(token, keys[i])` _should_ be enough, and devs also should only use **trusted** resource servers in any case

Code-Hex (approved):
LGTM

@yusukebe

Beyondo:
Hey, @yusukebe ! I improved test coverage by around +15% focusing on testing the important parts, such as signed cookies and missing `kid` exception and a few others, while the rest of unconvered lines are fairly trivial just by looking at them but also a bit tedious to write tests for; I will cover more of them as soon as I got time

yusukebe:
@Beyondo 

Thank you so much! If it's ready, please ping me. I'll review it.

Beyondo:
@yusukebe Yo, Everything, at least JWK-related, is covered & all hono tests pass.

I think it should be decent enough for review now

src/utils/jwt/jws.ts
```diff
@@ -18,7 +18,13 @@ type KeyAlgorithm =
   | (EcdsaParams & EcKeyImportParams)
   | HmacImportParams
 
-export type SignatureKey = string | JsonWebKey | CryptoKey
+// Extending the JsonWebKey interface to include the "kid" property.
+// https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.4
+export interface HonoJsonWebKey extends JsonWebKey {
```
yusukebe, line 23: I also wonder if it should be `HonoJsonWebKey` or `ExtendedJsonWebKey`, but `HonoJsonWebKey` is okay because it's a Hono specific type. Let's go with this.

src/middleware/jwk/index.test.ts
```diff
@@ -0,0 +1,639 @@
...
+    it('Should throw an error if the middleware is missing both keys and jwks_uri (empty)', async () => {
+      expect(() => app.use('/auth-with-empty-middleware/*', jwk({}))).toThrow(
+        'JWK auth middleware requires options for either "keys" or "jwks_uri"'
+      )
+    })
+
+    it('Should throw an error when crypto.subtle is missing', async () => {
+      const subtleSpy = vi.spyOn(global.crypto, 'subtle', 'get').mockReturnValue({
+        importKey: undefined,
+      } as any)
```
yusukebe, line 158: It's okay to use `any` but can you apply the patch to remove warnings?

```diff
diff --git a/src/middleware/jwk/index.test.ts b/src/middleware/jwk/index.test.ts
index 4d8bf8a1..bf591ca1 100644
--- a/src/middleware/jwk/index.test.ts
+++ b/src/middleware/jwk/index.test.ts
@@ -155,6 +155,7 @@ describe('JWK', () => {
     it('Should throw an error when crypto.subtle is missing', async () => {
       const subtleSpy = vi.spyOn(global.crypto, 'subtle', 'get').mockReturnValue({
         importKey: undefined,
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
       } as any)
       expect(() => app.use('/auth-with-bad-env/*', jwk({ keys: verify_keys }))).toThrow()
       subtleSpy.mockRestore()
@@ -228,6 +229,7 @@ describe('JWK', () => {
         const encodedPayload = encodeJwtPart(payload)
         const signaturePart = await signing(
           privateKey,
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
           privateKey.alg as any,
           utf8Encoder.encode(encodedPayload)
         )
@@ -256,6 +258,7 @@ describe('JWK', () => {
         const partialToken = `${encodedHeader}.${encodedPayload}`
         const signaturePart = await signing(
           privateKey,
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
           privateKey.alg as any,
           utf8Encoder.encode(partialToken)
         )
```

yusukebe:
@Beyondo 

Thank you for your hard work! Almost done. I've left comments.

src/utils/jwt/jws.ts
Beyondo: Yup! Less confusing this way. Also this reminded me that we should probably export it externally so I did that in `src/utils/jwt/types.ts` since it's already used for `JWTPayload` and other types as well.
```diff
 }
+
+export type { HonoJsonWebKey } from './jws'
```

yusukebe (approved):
LGTM!

yusukebe:
Hi @Beyondo 

Looks good. Thank you for your hard work! I'll merge it into the `next` branch for the next minor version and release it. Maybe soon!

Beyondo:
@nikeee I would personally attempt to separate anonymous endpoints, but addressing the question—you could create a custom middleware, say `jwkOptional`, that wraps around `jwk` but resumes on failure to achieve the _optional_ authentication.
```js
const jwkOptional = (options, init) => {
  const jwkMiddleware = jwk(options, init);
  return async (c, next) => {
    try {
      await jwkMiddleware(c, next);
    } catch (err) {
      await next();
    }
  };
};
```
With types
```ts
const jwkOptional = (options: {
  keys?: HonoJsonWebKey[] | (() => Promise < HonoJsonWebKey[] > );
  jwks_uri?: string;
  cookie?: string | {
    key: string;
    secret?: string | BufferSource;
    prefixOptions?: CookiePrefixOptions;
  };
}, init?: RequestInit) => {
  const jwkMiddleware = jwk(options, init);
  return async (c: Context, next: Next) => {
    try {
      await jwkMiddleware(c, next);
    } catch (err) {
      await next();
    }
  };
};
```
You could also modify it to set a boolean such as `is_authenticated` in the request's Context, but I suppose checking if `jwtPayload` is defined is enough too.

flexchar:
Could we add a function support to `jwks_uri` to allow fetching `backendServer` dynamically from ENV as shown in the docs when using, for example, cloudflare workers? 🙏

https://hono.dev/docs/middleware/builtin/jwk#usage

To support this:
```ts
const app = new Hono()

app.use(
  '/auth/*',
  jwk({
    jwks_uri: (c: Context) => `https://${c.env.backendServer}/.well-known/jwks.json`,
  })
)

app.get('/auth/page', (c) => {
  const payload = c.get('jwtPayload')
  return c.json(payload) // eg: { "sub": "1234567890", "name": "John Doe", "iat": 1516239022 }
})
```

Beyondo:
@flexchar That's a good idea and I see how that could be useful for a lot of people to have access to the `Context` inside the callbacks. I've created a PR for it + also @nikeee's requirement for an optional JWK **only if a token is present** without any hacky wrapper-middleware exception dissection workarounds. 💯

kyeotic:
I would also love to see the keys from the `jwks_uri` get cached for some period of time. Fetching the key on every request adds a lot of avoidable latency

varkarrus0:
@kyeotic I think the caching part could already could be done (it was mentioned in this PR); basically, some serverless adapters (say Cloudflare Workers) allow caching if you set the `cf` object in the `RequestInit` object with some options when using the JS `fetch`, see here: https://developers.cloudflare.com/workers/examples/cache-using-fetch/

I modified the example code a bit here to match what I am actually using:
```js
$router.use("/api/*",
    jwk({
        jwks_uri: (c) => c.env.AUTH_ORIGIN + "/.well-known/jwks.json",
        allow_anon: true
    }, {
        cf: { cacheTtl: 3600, cacheEverything: true }
    })
);
```
This will cache the `jwks.json` for the datacenter the cloudflare worker was called in for 1 hour as far as I know.

If you're not using Cloudflare Workers but another serverless platform, say Vercel, I think they also have their own options too such as `cache` property (instead of `cf`). If no caching options are available for your adapter (cause the platform doesn't support them), some serverless platforms might respect the caching headers you send in `fetch` anyways, so you could try setting some options in `headers` instead; check the docs of the serverless platform you use on caching.

If you're not using a serverless platform (but your own server), then most likely your headers would be respected! You just need to configure your server if it doesn't. But if all fails (somehow), you could also implement your own custom caching using the `keys` param (as a last resort) in the jwk middleware which I find very flexible; simply write your entire caching logic there and return the keys at the end, but again, this is a last resort.

kyeotic:
I would be happy with a simple in-memory cache by default.

buckett:
@varkarrus0 With something like:

> ```js
>         cf: { cacheTtl: 3600, cacheEverything: true }
> );
> ```

Just watch out caching everything means if the remote server ever has a blip and returns a 500 temporarily you'll end up caching that for as hour as well (even if it recovers quickly). Adjusting the cache timeout depending on the HTTP status allows good caching but also good recovery in the event of a problem on the remote server. For example:
```js
            cf: { cacheTtlByStatus: { "200-299": 86400, "404": 1, "500-599": 0 } }
```

varkarrus0:
@buckett This was a very good advice, thank you very much

Issue #3829, opened 2025-01-14
Allow for custom hashing methods to be used to etag


Gobd:
### What is the feature you are proposing?

Allow the user to pass a func to use a custom hashing method that would override `generateDigest`. This currently uses SHA1 maybe the user wants to use wyhash, xxhash, or something else.

PR #3832, opened 2025-01-15
feat(etag): allow for custom hashing methods to be used to etag

EdamAme-x:
resolve: https://github.com/honojs/hono/issues/3829

```ts
const app = new Hono()

app.use(
  '/etag/*',
  etag({
    generateDigest: (body: Uint8Array) =>
      crypto.subtle.digest({
          name: 'SHA-256',
        },
        body
      ),
  })
)
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/etag/digest.ts
+++ b/src/middleware/etag/digest.ts
@@ -9,13 +9,21 @@ const mergeBuffers = (buffer1: ArrayBuffer | undefined, buffer2: Uint8Array): Ui
 }
 
 export const generateDigest = async (
-  stream: ReadableStream<Uint8Array> | null
+  stream: ReadableStream<Uint8Array> | null,
+  generateDigest?: (body: Uint8Array) => ArrayBuffer | Promise<ArrayBuffer>
 ): Promise<string | null> => {
   if (!stream || !crypto || !crypto.subtle) {
     return null
   }
 
   let result: ArrayBuffer | undefined = undefined
+  generateDigest ||= (body: Uint8Array) =>
+    crypto.subtle.digest(
+      {
+        name: 'SHA-1',
+      },
+      body
+    )
 
   const reader = stream.getReader()
   for (;;) {
@@ -24,12 +32,7 @@ export const generateDigest = async (
       break
     }
 
-    result = await crypto.subtle.digest(
-      {
-        name: 'SHA-1',
-      },
-      mergeBuffers(result, value)
-    )
+    result = await generateDigest(mergeBuffers(result, value))
   }
 
   if (!result) {
--- a/src/middleware/etag/digest.ts
+++ b/src/middleware/etag/digest.ts
@@ -10,20 +10,13 @@ const mergeBuffers = (buffer1: ArrayBuffer | undefined, buffer2: Uint8Array): Ui
 
 export const generateDigest = async (
   stream: ReadableStream<Uint8Array> | null,
-  generateDigest?: (body: Uint8Array) => ArrayBuffer | Promise<ArrayBuffer>
+  generateDigest: (body: Uint8Array) => ArrayBuffer | Promise<ArrayBuffer>
 ): Promise<string | null> => {
   if (!stream || !crypto || !crypto.subtle) {
     return null
   }
 
   let result: ArrayBuffer | undefined = undefined
-  generateDigest ||= (body: Uint8Array) =>
-    crypto.subtle.digest(
-      {
-        name: 'SHA-1',
-      },
-      body
-    )
 
   const reader = stream.getReader()
   for (;;) {
--- a/src/middleware/etag/digest.ts
+++ b/src/middleware/etag/digest.ts
@@ -10,7 +10,7 @@ const mergeBuffers = (buffer1: ArrayBuffer | undefined, buffer2: Uint8Array): Ui
 
 export const generateDigest = async (
   stream: ReadableStream<Uint8Array> | null,
-  generateDigest: (body: Uint8Array) => ArrayBuffer | Promise<ArrayBuffer>
+  generator: (body: Uint8Array) => ArrayBuffer | Promise<ArrayBuffer>
 ): Promise<string | null> => {
   if (!stream || !crypto || !crypto.subtle) {
     return null
@@ -25,7 +25,7 @@ export const generateDigest = async (
       break
     }
 
-    result = await generateDigest(mergeBuffers(result, value))
+    result = await generator(mergeBuffers(result, value))
   }
 
   if (!result) {
--- a/src/middleware/etag/index.test.ts
+++ b/src/middleware/etag/index.test.ts
@@ -25,7 +25,7 @@ describe('Etag Middleware', () => {
     app.use(
       '/etag/*',
       etag({
-        generateDigest: (body: Uint8Array) =>
+        generateDigest: (body) =>
           crypto.subtle.digest(
             {
               name: 'SHA-256',
--- a/src/middleware/etag/index.test.ts
+++ b/src/middleware/etag/index.test.ts
@@ -20,6 +20,39 @@ describe('Etag Middleware', () => {
     expect(res.headers.get('ETag')).toBe('"4515561204e8269cb4468d5b39288d8f2482dcfe"')
   })
 
+  it('Should return etag header with another algorithm', async () => {
+    const app = new Hono()
+    app.use(
+      '/etag/*',
+      etag({
+        generateDigest: (body: Uint8Array) =>
+          crypto.subtle.digest(
+            {
+              name: 'SHA-256',
+            },
+            body
+          ),
+      })
+    )
+    app.get('/etag/abc', (c) => {
+      return c.text('Hono is cool')
+    })
+    app.get('/etag/def', (c) => {
+      return c.json({ message: 'Hono is cool' })
+    })
+    let res = await app.request('http://localhost/etag/abc')
+    expect(res.headers.get('ETag')).not.toBeFalsy()
+    expect(res.headers.get('ETag')).toBe(
+      '"ee7e84f92c4f54fec768123ac23003a6eb8437db95bcfbfc35db477af1ccb49e"'
+    )
+
+    res = await app.request('http://localhost/etag/def')
+    expect(res.headers.get('ETag')).not.toBeFalsy()
+    expect(res.headers.get('ETag')).toBe(
+      '"6ae7438c67f07b60b2ab069dbce206b00d3528c690840a77e0222d37398a8547"'
+    )
+  })
+
   it('Should return etag header - binary', async () => {
     const app = new Hono()
     app.use('/etag/*', etag())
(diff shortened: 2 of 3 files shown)


EdamAme-x:
Can you review this?
@Gobd @usualoma

EdamAme-x:
@Gobd Sorry for ping, how?

EdamAme-x:
@Gobd Sorry for ping, how?

usualoma:
Hi @EdamAme-x!

Thank you for creating the PR. I'm sorry for the very late response. Please wait a little while as I check it.

usualoma:
@EdamAme-x 
This method of PR is very good because users can generate the etag value simply by passing a custom hash function. Users don't need to be aware of the stream, and it's very simple.

However, the current etag of hono is limited by the fact that `crypto.subtle` does not support incremental generation, so it is in a situation where it is forced to repeat the generation of the hash value. If a custom function supports incremental generation, it would be more advantageous in terms of performance to use that. (Or, depending on the app, it may be the case that generating a hash value from the entire body is not necessary, and that generating a hash value from the first chunk is sufficient.) Considering this, I think there is a possibility that a generator that receives a stream would be better.
For example, it could be done in the following way.
https://github.com/honojs/hono/compare/main...usualoma:hono:feat/etag-generator-option

However, if you are going to be this flexible, it seems better to handle it in the middleware for each user rather than as an option.
```diff
diff --git a/src/middleware/etag/index.test.ts b/src/middleware/etag/index.test.ts
index a1c67c60..f2a90f7f 100644
--- a/src/middleware/etag/index.test.ts
+++ b/src/middleware/etag/index.test.ts
@@ -1,3 +1,4 @@
+import { createHash } from 'crypto'
 import { Hono } from '../../hono'
 import { RETAINED_304_HEADERS, etag } from '.'
 
@@ -261,4 +262,74 @@ describe('Etag Middleware', () => {
       expect(res.headers.get('ETag')).toBeNull()
     })
   })
+
+  it('Should generate etag from custom generator middleware', async () => {
+    const customGenerator = async (stream: ReadableStream) => {
+      // create sha1 hash incrementally by Node.js crypto
+      const hash = createHash('sha1')
+      const reader = stream.getReader()
+
+      while (true) {
+        const { done, value } = await reader.read()
+        if (done) {
+          break
+        }
+        hash.update(value)
+      }
+
+      return hash.digest('hex')
+    }
+
+    const app = new Hono()
+    app.use('/etag/*', etag(), async (c, next) => {
+      await next()
+
+      if (c.res.headers.get('ETag')) {
+        return
+      }
+
+      const stream = c.res.clone().body
+      if (!stream) {
+        return
+      }
+
+      const hash = await customGenerator(stream)
+      c.res.headers.set('ETag', `"${hash}"`)
+    })
+
+    app.get('/etag/stream1', (c) => {
+      return c.body(
+        new ReadableStream({
+          start(controller) {
+            controller.enqueue(new TextEncoder().encode('Hello '))
+            controller.enqueue(new TextEncoder().encode('World'))
+            controller.close()
+          },
+        })
+      )
+    })
+
+    app.get('/etag/stream2', (c) => {
+      return c.body(
+        new ReadableStream({
+          start(controller) {
+            controller.enqueue(new TextEncoder().encode('Hello'))
+            controller.enqueue(new TextEncoder().encode(' '))
+            controller.enqueue(new TextEncoder().encode('World'))
+            controller.close()
+          },
+        })
+      )
+    })
+
+    app.get('/etag/text', (c) => {
+      return c.text('Hello World')
+    })
+
+    for (const path of ['/etag/stream1', '/etag/stream2', '/etag/text']) {
+      const res = await app.request(path)
+      expect(res.status).toBe(200)
+      expect(res.headers.get('ETag')).toBe('"0a4d55a8d778e5022fab701977c5d840bbc486d0"')
+    }
+  })
 })
```


What do you think?

I think this PR approach is fine with a simple API. However, I am a little concerned about the fact that it involves going through a loop of repeatedly applying a hash function, which is being done because (the code I wrote) crypto.subtle does not support incremental generation.

https://github.com/honojs/hono/blob/ecf5a0b2aeaf29c26ffe1eaf41520f867510cb17/src/middleware/etag/digest.ts#L21-L33

EdamAme-x:
Hmmm...
Either one sounds excellent.

EdamAme-x:
I think to provide a simple api is fine in this context.

If the user is concerned about maximizing performance, it may be better to have them consider creating their own.

usualoma:
@EdamAme-x Thank you for your response.

What kind of use case is this function actually used for?

Since etag does not require a cryptographic hash function, I think the use case is ‘I want to make it faster to generate.’

However, in the first place, in cases where speed is an issue, ‘loading the entire response body and calculating the hash value at the time of response’ is not a good approach, so I think it is difficult to aim for ultra-fast with the current hono etag design.
If speed is essential, I think it is better to use the last modified date of the resource or calculate it in advance.
https://httpd.apache.org/docs/2.4/mod/core.html#fileetag

There may be cases where you want to replace the hash generation function to speed up the process while keeping the design of the hono etag, but if you go through the loop of non-incremental generation, it can't be said to be ultra fast. I feel that this would never be used in production and would be nothing more than a ‘try’.

I have a question about how much demand there is for ‘replacing the hash function to speed up hono's etag’. If it's not that much, I think it's not necessary to add the function, and it would be enough to guide people to the following approach in the documentation, etc.

```ts
app.use('/etag/*', etag(), async (c, next) => {
  await next()
  const etag = await customETagGenerator(c.res.clone().body)
  c.res.headers.set('ETag', `"${etag}"`)
})
```

What do you think, @yusukebe?

usualoma:
If we want to neatly organize it as an API, I think there are also ways to introduce APIs like Cloudflare's `crypto.DigestStream`. However, I don't think we have the motivation to do that.

https://developers.cloudflare.com/workers/runtime-apis/web-crypto/#constructors

EdamAme-x:
That is certainly true.
Would it be a good move to receive a callback function that would wrap the part that receives the `ReadbleStream` and generates the hash value...?

```ts
etag({
  generateDigest: ([prev, cur]) => { ... }
})
```

@Gobd What do you think?

Gobd:
After filing that issue what I ended up doing was copying the middleware code into a pre-existing wrapper function gets the body response object, JSON stringifies it, then runs `Bun.hash.wyhash(body).toString(16)` and uses `c.body` to send the response. Based on my testing that is faster than what was happening here but I am not a JS dev so I'm not sure. In reality I think anyone going for performance would probably do something similar and actually look at the middleware code, see how simple it is, and copy/paste if and modify it for theit needs.

yusukebe:
Hey @usualoma @EdamAme-x 

Ideally, I think it would be nice to be able to change the ETag value instead of only changing the hash generation function.

By the way, is this line really needed (though I think I wrote this code 😳 )? 

https://github.com/honojs/hono/blob/d40fffbec624bae50f8fc3b619125dda9a8c50e2/src/middleware/etag/index.ts#L74

If we can remove this, it will be fast.

yusukebe:
Hi @usualoma @EdamAme-x 

I considered it, and the `generateDigest` feature is good. As @usualoma says, using a simple value is faster than generating the hash value. However, as @Gobd says, it might be possible to speed up the hash value generation using an algorithm other than SHA1. Then I think it is good to provide the `generateDigest.` Of course, if there is no need to generate the hash value from the body, then the user just need not use the tag middleware.

usualoma:
> Ideally, I think it would be nice to be able to change the ETag value instead of only changing the hash generation function.
> 
> By the way, is this line really needed (though I think I wrote this code 😳 )?

🤔 
It certainly looks like it's not necessary.

usualoma:
Hi @yusukebe 
Thanks for your reply.

Assuming that we can replace `generateDigest`, which signature do you think we should be using?

* `(body: Uint8Array) => ArrayBuffer | Promise<ArrayBuffer>`
    * It's really simple. However, if the body is returned in multiple chunks (if it's large), the overhead will be large, just like with the current "etag" implementation.
* `(stream: ReadableStream<Uint8Array>) => string | null | Promise<string | null>`
    * It is not simple because we have to process the stream. However, you can use ‘a hash function that can add values incrementally’. It also becomes possible to do things like ‘calculate the correct SHA1 for the entire body, even if it is large’.
* Other things like `crypto.DigestStream`

EdamAme-x:
or support both...?

```js
etag({
  generateDigest: ...,
  // or
  generateDigestStream: ...,
})
```

usualoma:
I see, we can also implement both.

yusukebe:
@usualoma 

I'll comment tomorrow! Please wait.

yusukebe:
@EdamAme-x @usualoma 

I feel that `generateDigestStream` is outside the scope of this PR. So, how about we implement only `generateDigest` in this PR?

And it's my opinion. I want to keep the ETag Middleware simple; only `generateDigest` is okay. Users who wish to require high-performance ETag logic can implement it by not calculating the response content, so the `generateDigestStream` may not be used often.

usualoma:
OK, then, let's keep going!

src/middleware/etag/digest.ts
```diff
@@ -9,7 +9,8 @@ const mergeBuffers = (buffer1: ArrayBuffer | undefined, buffer2: Uint8Array): Ui
 }
 
 export const generateDigest = async (
-  stream: ReadableStream<Uint8Array> | null
+  stream: ReadableStream<Uint8Array> | null,
+  generateDigest: (body: Uint8Array) => ArrayBuffer | Promise<ArrayBuffer>
```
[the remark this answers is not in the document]
EdamAme-x: okay, thanks for your advice.
I just changed this to `generator`.

EdamAme-x:
@yusukebe maybe, ready for merge

src/middleware/etag/index.test.ts
```diff
@@ -20,6 +20,39 @@ describe('Etag Middleware', () => {
     expect(res.headers.get('ETag')).toBe('"4515561204e8269cb4468d5b39288d8f2482dcfe"')
   })
 
+  it('Should return etag header with another algorithm', async () => {
+    const app = new Hono()
+    app.use(
+      '/etag/*',
+      etag({
+        generateDigest: (body: Uint8Array) =>
```
yusukebe, line 28: Is this the type `Uint8Array` necessary? I think the following is okay.

```diff
diff --git a/src/middleware/etag/index.test.ts b/src/middleware/etag/index.test.ts
index 2795be08..296c74bb 100644
--- a/src/middleware/etag/index.test.ts
+++ b/src/middleware/etag/index.test.ts
@@ -25,7 +25,7 @@ describe('Etag Middleware', () => {
     app.use(
       '/etag/*',
       etag({
-        generateDigest: (body: Uint8Array) =>
+        generateDigest: (body) =>
           crypto.subtle.digest(
             {
               name: 'SHA-256',
```

EdamAme-x: okay!

yusukebe (approved):
LGTM!

usualoma:
@EdamAme-x @yusukebe 
Thanks! LGTM!

PR #3833, opened 2025-01-15
fix(compress): don't compress server-sent events

dmitri-gb:
Using both `compress` middleware and the `streamSSE` helper from `hono/streaming` doesn't work correctly -- the SSE messages are not streamed, instead they get buffered by the compression mechanism.

According to [expressjs/compression](https://github.com/expressjs/compression#server-sent-events) it should be possible to compress server-sent events, if the underlying compression mechanism supports manual _flushing_. As far as I can tell though, `CompressionStream` (used by Hono's `compress` middleware) doesn't offer this functionality, so the best we can do is not compress SSE streams at all.


EdamAme-x:
Nice work
I think need to add tests about this

yusukebe:
Hi @dmitri-gb 

Thank you for the PR. I also realized that the problem and your idea are good! As @EdamAme-x mentioned, you need to add a test. I think it's a little unsuitable if you add it to `src/helper/streaming/sse.test.tsx` because this fix is about Compress Middleware. So it will be okay to test if it will not be compressed if the content type is `text/event-stream` in `src/middleware/compress/index.test.ts`.

dmitri-gb:
Hi! I added a test in `src/middleware/compress/index.test.ts` that uses the `streamSSE` helper.

yusukebe (approved):
LGTM!

yusukebe:
@dmitri-gb 

Great! Thank you for your contribution.

mfulton26:
related issues for tracking ability to control flushing on `CompressionStream`:
- https://github.com/whatwg/compression/issues/56
- https://github.com/whatwg/compression/issues/22

PR #3843, opened 2025-01-21
feat(validator): Introduce `responseValidator`

### LINKED ISSUE #2439: Response validation

Certain frameworks (eg: [fastify](https://fastify.dev/docs/latest/Reference/Validation-and-Serialization/#serialization)) provide response validation for success or error payloads.
Is there support for such feature in hono? If not, is it something planned?

EdamAme-x:
Resolve: https://github.com/honojs/hono/issues/2439
Related: https://github.com/honojs/middleware/pull/184

This is proposal.
It has not been tested, nor is it fully molded.

```ts
import { Hono } from '../preset/quick'
import { responseValidator } from '.'

const app = new Hono()

app
  .use(
    '*',
    responseValidator('text', (value, c) => {
      if (!/\d/.test(value)) {
      }
    })
  )
  .get('/', (c) => {
    return c.text(c.validate('hello world at ' + Date.now()))
  })
  .get('/invalid', (c) => {
    return c.text(c.validate('hello world at ' + NaN))
  })

export default app
```

The data is kept intact in the context.

No Prepared

- [ ] Add tests
- [ ] Run tests


EdamAme-x:
What do you think about this API? @yusukebe 
If we can allow this

EdamAme-x:
The only code added to context is the following, with no major changes to the bundle size.
`this.validate=a=>(this.validateData=a,a);`

EdamAme-x:
In the future, it may be necessary to properly infer the output type for RPC.

belgattitude:
@EdamAme-x just bringing to your attention that there's an initiative of a [standard schema specification](https://github.com/standard-schema/standard-schema) that seems to be implemented in latest zod, valibot beta, arktypes...

I don't know if it affects your PR but in case it might be good to have a look.

BTW I'm just starting some internal projects with hono and nextjs and it looks great !

EdamAme-x:
@belgattitude 
Thanks for letting us know!

There is already an interesting discussion about standard schema in the honojs/middleware repository, if you want to check it out!

PR #3844, opened 2025-01-21
chore: delete misleading comments

EdamAme-x:
This is not a problem because the IETF is the organization that manages RFCs for the Internet.


ryuapp:
The intention of that comments seems to be replace correct links when rfc6265bis is standardized.

> Note: I'm adding link to latest draft in each comment, but should be updated to RFC link.
https://github.com/honojs/hono/pull/2314#issue-2169898096

So there's no need to delete the comments yet, we will have to wait for the right time.

EdamAme-x:
I see, I understand.
That interpretation seems to be correct.

PR #3856, opened 2025-01-26
fix(helper/adapter): `env` should set `c` type correctly

### LINKED ISSUE #3854: typings problem with env in hono/adapter

4.6.18

node

The `env` function seems to have incorrect typings- it returns `T & C["Env"]` where it should be `T & C["env"]`

repro case:
```
import { Hono } from "hono";
import { env } from "hono/adapter";

type Env = {
  Bindings: {
    MY_VAR: string;
  };
};

const app = new Hono<Env>();

app.get("/api/clock", (c) => {
  const { MY_VAR } = env<{MY_VAR: string}>(c);

  return c.json({
    var: MY_VAR,
    time: new Date().toLocaleTimeString(),
  });
});
```

I expect the return value of `env` to have correct typings

### What do you see instead?

The typescript error is on the 'c' in `env<{M_VAR: string>(c)` with message:
```
Argument of type 'Context<Env, "/api/clock", BlankInput>' is not assignable to parameter of type 'Context<{}, any, {}>'.
  Types of property 'set' are incompatible.
    Type 'Set<Env>' is not assignable to type 'Set<{}>'.
      Property 'Bindings' is missing in type '{}' but required in type 'Env'.
```

### Additional information

in `context.d.ts` the `Context` class has a member `env: E["Bindings"]`, so I think the `env` typings have a typo

yusukebe:
Fixes #3854

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/helper/adapter/index.test.ts
+++ b/src/helper/adapter/index.test.ts
@@ -1,9 +1,29 @@
-import { getRuntimeKey } from '.'
+import { Hono } from '../../hono'
+import { env, getRuntimeKey } from '.'
 
 describe('getRuntimeKey', () => {
   it('Should return the current runtime key', () => {
     // Now, using the `bun run test` command.
     // But `vitest` depending Node.js will run this test so the RuntimeKey will be `node`.
     expect(getRuntimeKey()).toBe('node')
+  })
+})
+
+describe('env', () => {
+  describe('Types', () => {
+    type Env = {
+      Bindings: {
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @arolson101

I think this will resolve the problem. Can you review this?

arolson101:
looks good to me; thank you!

yusukebe:
@arolson101 

Thanks! Merging.

PR #3859, opened 2025-01-27
refactor(helper/streaming): Avoid attaching AbortSignal on newer versions of Bun

Jarred-Sumner:
While investigating a memory leak, I noticed that the streams helpers in hono unconditionally attach an AbortSignal to each request. As of Bun v1.1.27, [we call `cancel` on ReadableStream](https://bun.sh/blog/bun-v1.1.27#the-cancel-method-in-readablestream) so this should be unnecessary.

I verified that on Bun v1.1.26, hardcoding the `isOldBunVersion` function to return false caused the abort-related tests to hang and that on Bun v1.2 hardcoding the `isOldBunVersion` function to return false those tests pass.

There is duplicate code but it felt less invasive to duplicate the code.

- [ ] Add tests
- [X] Run tests

### DIFF
--- a/src/helper/streaming/sse.test.tsx
+++ b/src/helper/streaming/sse.test.tsx
@@ -76,6 +76,11 @@ describe('SSE Streaming helper', () => {
   })
 
   it('Check streamSSE Response if aborted by abort signal', async () => {
+    // Emulate an old version of Bun (version 1.1.0) for this specific test case
+    // @ts-expect-error Bun is not typed
+    global.Bun = {
+      version: '1.1.0',
+    }
     const ac = new AbortController()
     const req = new Request('http://localhost/', { signal: ac.signal })
     const c = new Context(req)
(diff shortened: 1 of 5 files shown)


yusukebe:
@Jarred-Sumner Thank you for the PR.

Hey @sor4chi, The CI is falling. Should we update the tests?

src/helper/streaming/sse.test.tsx
```diff
@@ -76,6 +76,11 @@ describe('SSE Streaming helper', () => {
   })
 
   it('Check streamSSE Response if aborted by abort signal', async () => {
+    // Emulate an old version of Bun (version 1.1.0) for this specific test case
+    // @ts-expect-error Bun is not typed
+    global.Bun = {
+      version: '1.1.0',
```
yusukebe, line 82: To pass the test, specify the old version at the global.

yusukebe (approved):

PR #3860, opened 2025-01-27
fix(middleware/cors): export CORSOptions type

aryasaatvik:
Exports the CORSOptions interface to allow external usage of the CORS middleware configuration type. 

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/middleware/cors/index.ts
+++ b/src/middleware/cors/index.ts
@@ -6,7 +6,7 @@
 import type { Context } from '../../context'
 import type { MiddlewareHandler } from '../../types'
 
-type CORSOptions = {
+export type CORSOptions = {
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @aryasaatvik 

Basically, we don't export option types of built-in middleware like `CORSOptions,` `BasicAuthOptions,` `CompressionOptions`, and others. So, honestly, I don't want to export the type as much as we can. If we allow exporting `CORSOptions`, we may have to export other option types. This may be fine, but "exporting something" means increasing maintenance costs. So, I can't decide if we merge this immediately.

yusukebe:
@aryasaatvik 

If you want to get the type, you can use `Parameters` without importing `CORSOptions`.

```ts
type Options = Parameters<typeof cors>[0]
```

PR #3873, opened 2025-01-31
fix(client): correct inferring empty object from`c.json({})`

### LINKED ISSUE #3868: Hono treats empty object `{}` as type `never` for RPC

4.6.19

Bun

Create any route and simply return `c.json({})`. It will be inferred as type `never`.

```ts
const app = new Hono()
app.get("/", async (c) => c.json({}))
```

This behavior is specified in the source code:

```ts
{
  json(): F extends "text" 
    ? Promise<never> 
    : F extends "json" 
      ? Promise<BlankRecordToNever<T>> 
      : Promise<unknown>
}
```

This makes no sense as `{}` is valid JSON, and the response at the other end is in fact `{}` and not `never`.

This behavior is very surprising because this is valid JSON being mistyped. This behavior should at least be very clearly specified in the docs.

### What do you see instead?

_No response_

### Additional information

_No response_

yusukebe:
Fixes #3868

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -50,6 +50,7 @@ describe('Basic - JSON', () => {
     )
     .get('/hello-not-found', (c) => c.notFound())
     .get('/null', (c) => c.json(null))
+    .get('/empty', (c) => c.json({}))
 
   type AppType = typeof route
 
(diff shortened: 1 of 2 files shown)


yusukebe:
@meows Can you review this?

PR #3883, opened 2025-02-01
fix(adapter/aws-lambda): APIGWProxyResult should contain either of headers or multiValueHeaders, not both

### LINKED ISSUE #3870: aws-lambda duplicate headers in API Gateway V1 response

4.6.17

AWS Lambda/NodeJS 20

If an incoming request event has `multiValueHeaders` set, then headers will be set in both `headers` and `multiValueHeaders` in the response, resulting in duplicated header values in the final HTTP response. The [AWS docs describe this behaviour](https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-output-format):

> If you specify values for both headers and multiValueHeaders, API Gateway merges them into a single list. If the same key-value pair is specified in both, only the values from multiValueHeaders will appear in the merged list.

This whole block looks suspicious to me:
https://github.com/honojs/hono/blob/d72aa4b6d77c7b3150bf2b7bae001e6635fe98ae/src/adapter/aws-lambda/handler.ts#L240-L254

1. The initialization of `multiValueHeaders` in `result` seems wrong, why does the incoming event alter the structure of the outgoing response? Surely `multiValueHeaders` should be initialized if there's multi-valued headers in `res.headers`.
2. If `result.multiValueHeaders` is initialized (because the incoming event had `multiValueHeaders`) then the subsequent loop will set a header in both `result.headers` and `result.multiValueHeaders` which is incorrect. API Gateway will merge both these fields in the final HTTP response.

I think the correct behaviour should be to either:

1. Detect if there a header has multiple values in `res.headers` and only set it in `multiValueHeaders`, otherwise only set it in `headers`. Never set it in both.
2. Just set every header in `multiValueHeaders` even if there's only a single value, and never set `headers`.

The [AWS docs](https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-output-format) suggest (2) is acceptable:

> The multiValueHeaders key can contain multi-value headers as well as single-value headers. **You can use the multiValueHeaders key to specify all of your extra headers, including any single-value ones.**

Headers should not be duplicated.

### What do you see instead?

_No response_

### Additional information

_No response_

exoego:
Closes #3870

- `APIGWProxyResult` now contains `headers` or `multiValueHeaders` mutually-exclusively.
- `handle` is tweaked so the response type can be inferred from the request event type.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/runtime-tests/lambda/index.test.ts
+++ b/runtime-tests/lambda/index.test.ts
@@ -249,6 +249,7 @@ describe('AWS Lambda Adapter for Hono', () => {
     expect(response.statusCode).toBe(200)
     expect(response.body).toBe('Hello Lambda!')
     expect(response.headers['content-type']).toMatch(/^text\/plain/)
+    expect(response.multiValueHeaders).toBeUndefined()
     expect(response.isBase64Encoded).toBe(false)
   })
 
@@ -268,6 +269,7 @@ describe('AWS Lambda Adapter for Hono', () => {
     expect(response.statusCode).toBe(200)
     expect(response.body).toBe('RmFrZSBJbWFnZQ==')
     expect(response.headers['content-type']).toMatch(/^image\/png/)
+    expect(response.multiValueHeaders).toBeUndefined()
     expect(response.isBase64Encoded).toBe(true)
   })
 
@@ -289,6 +291,7 @@ describe('AWS Lambda Adapter for Hono', () => {
     expect(response.statusCode).toBe(200)
     expect(response.body).toBe('Hello Lambda!')
     expect(response.headers['content-type']).toMatch(/^text\/plain/)
+    expect(response.multiValueHeaders).toBeUndefined()
     expect(response.isBase64Encoded).toBe(false)
   })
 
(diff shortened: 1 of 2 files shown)


exoego:
I would say this is "bug" since APIGW's headers merging is confusing to users.

In addition to AWS APIGateway, ALB should be taken care of, since both shares implementation.

> https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html#multi-value-headers-response
> You must use `multiValueHeaders` if you have enabled multi-value headers and `headers` otherwise.

It appears that `headers` and `multiValueHeaders` are mutually-exclusive in ALB response.
So opened https://github.com/honojs/hono/pull/3883

But I need to check it with the actual ALB in weekday.

rown89:
Hello, i have the same duplicate header issue, someone can take care of this PR and maybe merge ? :)

yusukebe (approved):
LGTM!

yusukebe:
@exoego 

Looks good to me! I'll merge now. Thank you!

PR #3885, opened 2025-02-03
fix(helper/adapter): correct `env` type

yusukebe:
Fixes https://github.com/honojs/website/pull/577

### The author should do the following, if applicable

- [ ] Add tests
- [ ] Run tests
- [ ] `bun run format:fix && bun run lint:fix` to format the code
- [ ] Add [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code

PR #3886, opened 2025-02-03
wip: perf(types): replace to interface

EdamAme-x:
This is an experiment.

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/build/remove-private-fields-worker.ts
+++ b/build/remove-private-fields-worker.ts
@@ -1,6 +1,6 @@
 import * as ts from 'typescript'
 
-export type WorkerInput = {
+export interface WorkerInput {
   file: string
   taskId: number
 }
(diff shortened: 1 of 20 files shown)


EdamAme-x:
This is an experiment.

Then, one thing I have noticed.

It appears that PRs other than yusukebe's do not show this.
https://github.com/honojs/hono/pull/3885#issuecomment-2630343714

yusukebe:
@EdamAme-x 

I've tried the same things that convert `type` to `interface` before. But the performance did not change at that time.

> It appears that PRs other than @yusukebe's do not show this.

Maybe it's a security issue. We have to investigate it.

EdamAme-x:
Thanks.
I will think another approach.

PR #3888, opened 2025-02-04
feat(router): support greedy matches with subsequent static components.

### LINKED ISSUE #3851: Paths fail to match when using regex match

4.6.18

Cloudflare workers but this happens in plain node as well

Run the following:

```ts
import { Hono } from 'hono';

const app = new Hono({ strict: false });

app.get('/:first{.+}/middle-a/:reference?', (c) => {
  const first = c.req.param('first');
  const reference = c.req.param('reference');
  return c.text(first + ':' + reference);
});

app.get('/:first{.+}/middle-b/end-c/:uuid', async (c) => {
  return c.text('if this path enabled, and next path enabled, no paths work');
});

// if this path is commented out, then we get matching
app.get('/:first{.+}/middle-b/:digest', async (c) => {
  return c.text('if this path and last path enabled, no paths work');
});

let res = await app.request('/part1/part2/middle-a/latest');
console.log(await res.text()); // 404

res = await app.request('/part1/middle-b/end-c/latest');
console.log(await res.text()); // 404
```

https://stackblitz.com/edit/hono-regex?file=src%2Fapp.ts

The requests would succeed and return the result:

```
part1/part2:latest
if this path enabled, and next path enabled, no paths work
```

### What do you see instead?

The requests fail and return the following:

```
404 Not Found
404 Not Found
```

### Additional information

If you comment out either the second or third path, the remaining paths work. So it appears there is some internal conflict with the matcher that is affecting all paths.

I've noticed it doesn't affect other paths, and it doesn't affect any paths if the "greedy" matcher isn't used, namely `{.+}`.

I've run into this because it's a genuine requirement for a project I'm working on.

usualoma:
Fixes #3851

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/router/common.case.test.ts
+++ b/src/router/common.case.test.ts
@@ -497,6 +497,41 @@ export const runTest = ({
       })
     })
 
+    describe('Capture multiple directories', () => {
+      beforeEach(() => {
+        router.add('GET', '/:dirs{.+}/file.html', 'file.html')
+      })
+
+      it('GET /foo/bar/file.html', () => {
+        const res = match('GET', '/foo/bar/file.html')
+        expect(res.length).toBe(1)
+        expect(res[0].handler).toEqual('file.html')
+        expect(res[0].params['dirs']).toEqual('foo/bar')
+      })
+    })
+
+    describe('Capture multiple directories and optional', () => {
+      beforeEach(() => {
+        router.add('GET', '/:prefix{.+}/contents/:id?', 'contents')
+      })
+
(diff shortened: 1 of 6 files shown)


yusukebe (approved):
LGTM!

yusukebe:
Hi @usualoma 

Great work! Thank you!

yusukebe:
This is not a breaking change but I will not include it in a patch rerelease and will include the next minor version.

PR #3903, opened 2025-02-07
feat(logger): max length of log can be specified

EdamAme-x:
This can mitigate the risk of attack.

Related: https://github.com/honojs/hono/pull/3702#discussion_r1857433017

```js
logger({
  printFunc: logFunc,
  maxLength: 64,
})
// No breaking changes are made because past formats are also supported.
logger(logFunc)
```

```llvm
--> GET /hhhhhhhhhhhhhhh... 200 0ms
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/logger/index.ts
+++ b/src/middleware/logger/index.ts
@@ -56,11 +56,11 @@ const trim = (str: string, maxLength: number) => {
 
 function log(
   printFunc: PrintFunc,
+  maxLength: number,
   prefix: string,
   method: string,
   path: string,
   status: number = 0,
-  maxLength: number = Infinity,
   elapsed?: string
 ) {
   const trimmedMethod = trim(method, maxLength)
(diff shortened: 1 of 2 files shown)


EdamAme-x:
@ryuapp @yusukebe 
Can you review this?

src/middleware/logger/index.ts
```diff
@@ -47,27 +47,43 @@ const colorStatus = (status: number) => {
 
 type PrintFunc = (str: string, ...rest: string[]) => void
 
+const trim = (str: string, maxLength: number) => {
```
yusukebe, line 50: I wonder if the `maxLength` should be the length of the trimmed string or number that this method should trim.

For example, is the following the expected result?

```ts
const trimmed = trim('abcd', 3)
console.log(trimmed.length) // 6 is expected?
```

yusukebe:
Hey @usualoma 

What do you think of this feature? I think it's okay, but on the other side, the user can set the custom `printFunc`. So, if we don't want to add more options, we should not implement this.

src/middleware/logger/index.ts
EdamAme-x: That is exactly what I expected.

But which would be more reasonable...?
I don't know, but to keep the layout strict, I have learned that there is an implementation that returns including `...`

EdamAme-x:
Thank you for your opinion.

In my opinion, unlike the Apache-like logs for analysis, this is also used for visual collection of what routes are currently being accessed.

So I think I can accept the option of shortening the length of the log, but I can sympathize with the opinion that it needs to be handled on a different layer and kept simple.

EdamAme-x:
We can close this once.
We can choose to re-open it in the future if users actually request it.

yusukebe:
@usualoma Thank you for the comment

I also never heard of the web server where the user can specify the log length. Agree to handle the problem in another layer.

@EdamAme-x What about closing this PR, though you've done well?

EdamAme-x:
Closing

PR #3905, opened 2025-02-07
fix(combine): halt middleware evaluation if error in `next()`

### LINKED ISSUE #3898: [hono/combine] Function skipped in some() inside every() is still called

4.7.0

Node.js

Here is a reproducible example:
https://stackblitz.com/edit/node-dtmqxwzv?file=index.ts

```ts
import { Hono } from 'hono';
import { every, some } from 'hono/combine';

const middleware = every(
  some(
    async (_, next) => {
      console.log('middleware 1');
      await next();
    },
    async (_, next) => {
      console.log('middleware 2', 'should not be called'); // 👈️ Called after "error thrown"
      await next();
    }
  ),
  () => {
    console.log('error thrown');
    throw new Error('some error');
  }
);

const app = new Hono();
app.use(middleware);
app.get('/', (c) => {
  return c.text('ok');
});

serve(app);
```

`middleware 2` should be skipped.

### What do you see instead?

`middleware 2`  called after "error thrown".

### Additional information

This issue is also reproducible with the following code. It seems that middleware skipped by some() is called later, regardless of whether every() is used.

```ts
const middleware: MiddlewareHandler = async (c, _next) => {
  await some(
    async (_, next) => {
      console.log('middleware 1');
      await next();
    },
    async (_, next) => {
      console.log('middleware 2', 'should not be called');
      await next();
    }
  )(c, async () => {
    console.log('error thrown');
    throw new Error('some error');
  });
};
```

usualoma:
fixes #3898 

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/combine/index.ts
+++ b/src/middleware/combine/index.ts
@@ -48,7 +48,7 @@ export const some = (...middleware: (MiddlewareHandler | Condition)[]): Middlewa
       try {
         const result = await handler(c, wrappedNext)
         if (result === true && !c.finalized) {
-          await next()
+          await wrappedNext()
         } else if (result === false) {
           lastError = new Error('No successful middleware found')
           continue
--- a/src/middleware/combine/index.ts
+++ b/src/middleware/combine/index.ts
@@ -37,20 +37,16 @@ type Condition = (c: Context) => boolean
  */
 export const some = (...middleware: (MiddlewareHandler | Condition)[]): MiddlewareHandler => {
   return async function some(c, next) {
-    let nextError: unknown
-    const tryNext: Next = async () => {
-      try {
-        await next()
-      } catch (error) {
-        nextError = error
-        throw error
-      }
+    let isNextCalled = false
+    const wrappedNext = () => {
+      isNextCalled = true
+      return next()
     }
 
     let lastError: unknown
     for (const handler of middleware) {
       try {
-        const result = await handler(c, tryNext)
+        const result = await handler(c, wrappedNext)
         if (result === true && !c.finalized) {
           await next()
         } else if (result === false) {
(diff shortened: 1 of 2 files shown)


satoshun00:
@usualoma 
Thank you for the quick response! I was also looking at this part of the code, and I think this approach makes sense.  

One small point—since `some()` processes multiple middleware using a for-loop, it might be worth verifying in the test code that the call count for the next function is 1.

usualoma:
Hi @satoshun00 
Thank you for your comment!
For the following reasons, I don't think it's necessary to add a test.

* If `next` is called multiple times, the following code will cause an error
    * https://github.com/usualoma/hono/blob/e0cf1566bb7727116683a1641a2cf914c8ef90de/src/compose.ts#L51-L53
* How `next` is handled depends on the internal implementation.

usualoma:
@yusukebe
Would you please review this?

src/middleware/combine/index.ts
```diff
@@ -51,7 +57,9 @@ export const some = (...middleware: (MiddlewareHandler | Condition)[]): Middlewa
         break
       } catch (error) {
         lastError = error
-        continue
+        if (isNextCalled) {
+          break
+        }
```
yusukebe, line 62: I think these lines `#L61 - L62` are covered, though Codecov says they are not covered. Should you add more tests?

usualoma: Thanks for your comment.

The refactoring of the e0cf156 test had failed 😢  Fixed in 033d48b.

yusukebe: I see!

yusukebe (approved):
LGTM!

yusukebe:
@usualoma Thank you for the quick fix!

@satoshun00 Thank you for raising the issue!

PR #3910, opened 2025-02-09
fix(utils/url): calculate ends with slash on every iteration

### LINKED ISSUE #3909: mergePath incorrectly adds a trailing slash when first path segment is `/`

4.7.0

Cloudflare Workers

This issue mainly happens when you combine Hono with Hono's Zod Openapi, but since the issue is coming from Hono itself i have created the issue here.

By using this code here:

```typescript
const app = new OpenAPIHono<HonoContext>();

const v1 = new OpenAPIHono<HonoContext>().basePath("/v1");

v1.openapi(
  createRoute({
    method: "get",
    path: "/",
    responses: {
      200: {
        description: "Get message",
      },
    },
  }),
  (c) => c.json({ message: "Hello" }),
);

app.route("/", v1);

app.doc("/doc", {
  openapi: "3.0.0",
  info: {
    version: "1.0.0",
    title: "My API",
  },
});
```

You can request `/v1` which will respond correctly, but in the final openapi document the route is generated as the following: `/v1/` which results in scalar, swagger and other such tools to use the wrong path for requesting the endpoint.

On the Zod Openapi middleware it will provide the following paths to `mergePath`: `mergePath("/", "/v1", "/")` which results in this `/v1/`.

It seems the issue comes from these lines, since after the first iteration the p variable is now `/`, and then this condition is true. 
https://github.com/honojs/hono/blob/6ceb1ab22e70d5232e092bfea7313f5f1f100198/src/utils/url.ts#L145-L148

The openapi document should not have `/v1/` but instead `/v1`.

### What do you see instead?

That the openapi document has `/v1/` and not `/v1`

### Additional information

It seems to only happen when there is a route with only `/` as the path.

luxass:
This fixes #3909

I don't know if this is the best way of doing this, if you can find a better/faster way let me know! :) 

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/utils/url.test.ts
+++ b/src/utils/url.test.ts
@@ -152,6 +152,8 @@ describe('url', () => {
       expect(mergePath('/book', 'hey', 'say')).toBe('/book/hey/say')
       expect(mergePath('/book', '/hey/', '/say/')).toBe('/book/hey/say/')
       expect(mergePath('/book', '/hey/', '/say/', '/')).toBe('/book/hey/say/')
+      expect(mergePath('/book', '/hey', '/say', '/')).toBe('/book/hey/say')
+      expect(mergePath('/', '/book', '/hey', '/say', '/')).toBe('/book/hey/say')
 
       expect(mergePath('book', '/')).toBe('/book')
       expect(mergePath('book/', '/')).toBe('/book/')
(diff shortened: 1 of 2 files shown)


yusukebe (approved):
LGTM!

yusukebe:
Hi @luxass !

Looks good. Thank you so much!

Issue #3914, opened 2025-02-12
Usage of "node:crypto" in aws-lambda (LLRT runtime)


atoko:
Hi team,

I'm building a Hono app intended to run in an AWS Lambda environment, and so far it works great. Recently, I've started investigating the LLRT runtime for lower start times. However, trying to build the "hono-lambda" package quickly fails due to the reference to "node:crypto". 

While I'm sure there's more to do with regards to LLRT compatibility,  can we look at removing the direct crypto reference? I'd like to hear the context on the lambda adapter's requirements before continuing on this.

Thanks 

[1] LLRT (https://github.com/awslabs/llrt)

watany-dev:
AWS Lambda Adapter is for ManagedRuntime (Node.js vXX). Therefore, support for LLRT is not included in the scope.

EdamAme-x:
Hi @watany-dev 
I just want to ask if there is an intention for the `node:crypto` to be added.

@atoko 
I assume this is for compatibility since the etag middleware etc.  
uses the crypto api, but since this seems to be causing compatibility problems in the opposite direction, I would like to consider whether there is another way.

EdamAme-x:
I have created a PR that simply deletes it, but this could be a change that would destroy whatever compatibility we have.
It may be better to consider a different approach.

watany-dev:
This adapter is only for AWS Lambda ManagedRuntime (Node.js), so compatibility should be a priority over LLRT, the runtime under development. Of course, if simply removing the import is all that is needed, it should be implemented.

atoko:
Thanks for the quick response! I'll use the PR as a starting point and report any feedback as I try again

watany-dev:
@usualoma 
This import does not mean much, it just follows what was originally in the AWS Lambda Adaptor at the time of implementation. I was hesitant to remove it because I could not determine the decisions made at design time.Last year when I attended, v16 may have still been active.

You are correct in that v18 is the minimum runtime that AWS Lambda assumes. Therefore, I see no reason to object to its removal.

EdamAme-x:
okay.
I'll revert.

PR #3915, opened 2025-02-12
fix(adapter/aws-lambda): remove unneeded import and compatibility for LLRT

EdamAme-x:
this pr fixes https://github.com/honojs/hono/issues/3914 

- [ ] Add tests
- [x] Run tests


watany-dev:
The only reason `node:crypto` is declared in the Lambda@edge Adaptor is because it was declared in the AWS Lambda Adaptor at the time of creation.


By the way, the runtime of Lambda@Edge is fixed to Python or Node.js, so there may be no need to delete it.

EdamAme-x:
Close this once.

EdamAme-x:
@watany-dev I changed the approach.
Users can use compatible code from `hono/lambda-edge/llrt`.

EdamAme-x:
There was an error in the revised section.

EdamAme-x:
ref: https://github.com/honojs/hono/issues/3914#issuecomment-2653049604

yusukebe (commented):
LGTM!

yusukebe:
@EdamAme-x 

Thank you! I think this is ready to merge. Let's go.

PR #3916, opened 2025-02-12
fix(etag): change where to check for `crypto`

EdamAme-x:
This PR allows users to specify a hash generation function even in environments where there is no crypto, so the place to check for `crypto` changes.
https://github.com/honojs/hono/pull/3832

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/etag/digest.ts
+++ b/src/middleware/etag/digest.ts
@@ -10,9 +10,9 @@ const mergeBuffers = (buffer1: ArrayBuffer | undefined, buffer2: Uint8Array): Ui
 
 export const generateDigest = async (
   stream: ReadableStream<Uint8Array> | null,
-  generator?: (body: Uint8Array) => ArrayBuffer | Promise<ArrayBuffer>
+  generator: (body: Uint8Array) => ArrayBuffer | Promise<ArrayBuffer>
 ): Promise<string | null> => {
-  if (!stream || !generator) {
+  if (!stream) {
     return null
   }
 
--- a/src/middleware/etag/digest.ts
+++ b/src/middleware/etag/digest.ts
@@ -10,9 +10,9 @@ const mergeBuffers = (buffer1: ArrayBuffer | undefined, buffer2: Uint8Array): Ui
 
 export const generateDigest = async (
   stream: ReadableStream<Uint8Array> | null,
-  generator: (body: Uint8Array) => ArrayBuffer | Promise<ArrayBuffer>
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @EdamAme-x 

Do you mean this PR will resolve #3914? If so, I think this approach is good.

EdamAme-x:
The context is slightly different.

The main direction of this PR is to solve the problem of checking for the presence of a crypto api, even though it uses a hash generation method that does not use a crypto api.

Of course, as you may be thinking, this PR may improve compatibility as a side effect.

yusukebe:
@EdamAme-x 

I understood! But, to fix #3914, it will be good to delete `import crypto from 'node:crypto'` following the [comment](
https://github.com/honojs/hono/issues/3914#issuecomment-2653049604)

EdamAme-x:
Okay.
@yusukebe This PR should be able to merge.
Can you review this?

src/middleware/etag/digest.ts
```diff
@@ -10,9 +10,9 @@ const mergeBuffers = (buffer1: ArrayBuffer | undefined, buffer2: Uint8Array): Ui
 
 export const generateDigest = async (
   stream: ReadableStream<Uint8Array> | null,
-  generator: (body: Uint8Array) => ArrayBuffer | Promise<ArrayBuffer>
+  generator?: (body: Uint8Array) => ArrayBuffer | Promise<ArrayBuffer>
```
usualoma, line 13: It depends on the situation, but in general, I think it's better to avoid optional arguments in internal APIs.

src/middleware/etag/index.ts
```diff
@@ -57,15 +57,19 @@ function etagMatches(etag: string, ifNoneMatch: string | null) {
...
+    if (crypto && crypto.subtle) {
+      generator = (body: Uint8Array) =>
+        crypto.subtle.digest(
+          {
+            name: 'SHA-1',
+          },
+          body
+        )
+    }
+  }
```
usualoma, line 72: The initialisation conditions are complicated, so I think it would be better to make it a separate function.

usualoma (commented):
Hi, @EdamAme-x. Thank you for creating a pull request.

There is no need to adopt my branch as they are, but I would like to convey the following in my comments.

https://github.com/EdamAme-x/hono/compare/fix/etag-if-bug...usualoma:hono:fix/etag-if-bug-2

EdamAme-x:
ready

usualoma:
@EdamAme-x Thank you!

yusukebe (approved):
LGTM!

Issue #3919, opened 2025-02-15
`helper/proxy` performance


BarryThePenguin:
### What version of Hono are you using?

4.7.0

### What runtime/platform is your app running on? (with version if possible)

Cloudflare Workers

### What steps can reproduce the bug?

I was looking for any changes to performance metrics after adopting `hono/proxy` for a Cloudflare Worker. Ideally a proxy would have minimal overhead, but there wasn't any real observable change. Nice work! 🎉

Is there any way to ensure the proxy helper doesn't add any unnecessary overhead?

This is a basic example of the original handler..

```js
app.post('/api/v1/example', (c) => {
  const honoRequest = c.req
  const request = c.req.raw
  const toforward = honoRequest.query('toforward')
  const exampleUrl = `https://example.com${toforward}`
  const headers = new Headers(request.headers)
  headers.delete('cookie')
  return fetch(exampleUrl, { body: request.body, headers, method: 'post' })
})
```

..and after adopting `hono/proxy`..

```js
app.post('/api/v1/example', (c) => {
  const toforward = c.req.query('toforward')
  const exampleUrl = new URL(toforward, 'https://example.com')
  return proxy(exampleUrl, {
    ...c.req,
    headers: {
      ...c.req.header(),
      Cookie: undefined,
    },
  })
})
```

### Additional information

While investigating, I did notice a few other issues

- [x] #3920
- [x] #3924
- [x] #3926

usualoma:
Hi @BarryThePenguin
 
Thanks for trying out the proxy helper. Since it is difficult to imagine that the current implementation, including the spread operator, would cause any observable overhead, we do not plan to make any changes to improve performance.

BarryThePenguin:
There's still a few things I plan to revisit in the future, but I'm happy with the recent improvements that have been made. Thanks for your assistance! 👋🏻

PR #3921, opened 2025-02-16
test(helper/proxy): does not propagate undefined request headers

### LINKED ISSUE #3920: `helper/proxy` not removing `undefined` headers from `ProxyRequestInit`
Using the code from the [Proxy Helper documentation](https://hono.dev/docs/helpers/proxy#proxy-helper)..

```ts
  proxy(`https://example.com/${c.req.param('path')}`, {
    ...c.req,
    headers: {
      Authorization: undefined,
    },
  })
)
```

..the `Authorization` header isn't removed from the proxied request headers

BarryThePenguin:
- [x] Adds a failing test using the documentation as an example
- [x] Remove `undefined` headers from `ProxyRequestInit`
- [ ] Handle duplicate headers from `Request` and `ProxyRequestInit`

Fixes #3920

### DIFF
--- a/src/helper/proxy/index.test.ts
+++ b/src/helper/proxy/index.test.ts
@@ -191,16 +191,19 @@ describe('Proxy Middleware', () => {
           ...c.req,
           headers: {
             Authorization: undefined,
+            'X-Request-Id': '123',
           },
         })
       )
       const res = await app.request('/proxy/no-propagate', {
         headers: {
           Authorization: 'Bearer 123',
+          'X-Request-Id': '123',
         },
       })
       const req = (global.fetch as ReturnType<typeof vi.fn>).mock.calls[0][0]
       expect(req.headers.get('Authorization')).toBeNull()
+      expect(req.headers.get('X-Request-Id')).toBe('123, 123')
 
       expect(res.status).toBe(200)
       expect(res.headers.get('Authorization')).toBeNull()
--- a/src/helper/proxy/index.test.ts
+++ b/src/helper/proxy/index.test.ts
@@ -44,6 +44,10 @@ describe('Proxy Middleware', () => {
               },
             })
           )
+        } else if (req.url === 'https://example.com/no-propagate') {
+          return Promise.resolve(
+            new Response(`request authorization header: ${req.headers.get('authorization')}`)
+          )
         }
         return Promise.resolve(new Response('not found', { status: 404 }))
       })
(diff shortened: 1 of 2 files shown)


src/helper/proxy/index.ts
```diff
@@ -87,16 +111,7 @@ const buildRequestInitFromRequest = (
...
-  const { raw, ...requestInit } = proxyInit ?? {}
-
-  const req = new Request(
-    input,
-    // @ts-expect-error `headers` in `requestInit` is not compatible with HeadersInit
-    {
-      ...buildRequestInitFromRequest(raw),
-      ...requestInit,
-    }
-  )
```
BarryThePenguin, line 99: Re: #3919, my basic understanding of the spread operator is that new objects are created and all the keys are copied. This avoids that extra overhead

src/helper/proxy/index.ts
```diff
@@ -29,14 +29,37 @@ interface ProxyFetch {
...
+
 const buildRequestInitFromRequest = (
-  request: Request | undefined
+  request: Request | undefined,
+  requestInit: ProxyRequestInit | undefined
 ): RequestInit & { duplex?: 'half' } => {
   if (!request) {
-    return {}
+    // @ts-expect-error `headers` in `requestInit` is not compatible with HeadersInit
+    return requestInit
```
BarryThePenguin, line 58: This is what lines 94 and 97 were doing previously..

https://github.com/honojs/hono/blob/e7fbc3e2d60a657f5a330daa59b068073f061d5e/src/helper/proxy/index.ts#L92-L99

src/helper/proxy/index.ts
```diff
@@ -45,6 +68,7 @@ const buildRequestInitFromRequest = (
     method: request.method,
     body: request.body,
     duplex: request.body ? 'half' : undefined,
+    ...requestInit,
     headers,
```
BarryThePenguin, line 72: Previously the `headers` from `buildRequestInitFromRequest(raw)` were being overwritten by `...requestInit` on line 97

https://github.com/honojs/hono/blob/e7fbc3e2d60a657f5a330daa59b068073f061d5e/src/helper/proxy/index.ts#L95-L98

src/helper/proxy/index.ts
```diff
@@ -29,14 +29,37 @@ interface ProxyFetch {
...
+    requestHeaders = Object.entries(init.headers)
+  }
+
+  if (requestHeaders) {
+    for (const [name, value] of requestHeaders) {
+      if (value === undefined) {
+        headers.delete(name)
+      } else {
+        headers.append(name, value)
+      }
```
BarryThePenguin, line 47: Unsure if this is right..

It's deleting a header when the value is `undefined`, otherwise the value is appended. This means values in both the `ProxyRequestInit` and the request will be merged, rather than the value from the request overwriting the value from `ProxyRequestInit`, for example.

src/helper/proxy/index.test.ts
```diff
@@ -191,16 +191,19 @@ describe('Proxy Middleware', () => {
...
       )
       const res = await app.request('/proxy/no-propagate', {
         headers: {
           Authorization: 'Bearer 123',
+          'X-Request-Id': '123',
         },
       })
       const req = (global.fetch as ReturnType<typeof vi.fn>).mock.calls[0][0]
       expect(req.headers.get('Authorization')).toBeNull()
+      expect(req.headers.get('X-Request-Id')).toBe('123, 123')
```
BarryThePenguin, line 206: Duplicate request headers using `Header#append`. Alternatively, using `Header#set`

src/helper/proxy/index.ts
BarryThePenguin, line 72: Looking at this again, maybe this should actually be..

```suggestion
    ...requestInit,
    method: request.method,
    body: request.body,
    duplex: request.body ? 'half' : undefined,
    headers,
```

..to avoid `requestInit` overwriting any of `method`, `body`, `duplex` and `headers`?

PR #3922, opened 2025-02-16
fix(proxy): fix header overwrite

### LINKED ISSUE #3920: `helper/proxy` not removing `undefined` headers from `ProxyRequestInit`

usualoma:
fixes #3920

* header key is case-insensitive
* delete header if value is null or undefined

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/helper/proxy/index.test.ts
+++ b/src/helper/proxy/index.test.ts
@@ -5,7 +5,9 @@ describe('Proxy Middleware', () => {
   describe('proxy', () => {
     beforeEach(() => {
       global.fetch = vi.fn().mockImplementation(async (req) => {
-        if (req.url === 'https://example.com/compressed') {
+        if (req.url === 'https://example.com/ok') {
+          return Promise.resolve(new Response('ok'))
+        } else if (req.url === 'https://example.com/compressed') {
           return Promise.resolve(
             new Response('ok', {
               headers: {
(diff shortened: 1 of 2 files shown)


usualoma:
Hi @BarryThePenguin. Thank you very much for the report. I appreciate it.

I think you're also thinking about performance in #3921, but I'd like to consider that in a separate PR. I'd like to fix the problem of not being overwritten correctly in this PR, but what do you think?

usualoma:
@yusukebe
Would you please review this?

BarryThePenguin (approved):
> I think you're also thinking about performance in #3921, but I'd like to consider that in a separate PR. I'd like to fix the problem of not being overwritten correctly in this PR, but what do you think?

Sounds good 👍🏻 

I was having a hard time keeping the PR focused. This PR solves the immediate issue without changing anything else

usualoma:
@BarryThePenguin Thank you for your comment!

As you pointed out, the proxy helper behaves as follows.

> only type Record<string, string | undefined> can be used to remove specific headers from a request.

This is precisely what is intended, as it is ideal for the pattern described in the documentation: ' to allow for easy removal or overwriting in a declarative way using `...c.req.header()`’.

https://hono.dev/docs/helpers/proxy

If `type Headers | string[][]` is passed, it is assumed that the code is not written in a simple declarative way like `...c.req.header()`. In such cases, I think we can assume that the user is deliberately setting up a complex configuration, and we should not perform any modifications to headers.

yusukebe (approved):
LGTM!

yusukebe:
@usualoma @BarryThePenguin Thank you both! Looks good. Merging.

PR #3925, opened 2025-02-17
fix(proxy): include original request signal in proxy request

### LINKED ISSUE #3924: `hono/proxy` does not preserve `request.signal`
It doesn't look like the original [`Requst#signal`](https://developer.mozilla.org/en-US/docs/Web/API/Request/signal) is included in the proxy request

BarryThePenguin:
This includes the original [`Request#signal`](https://developer.mozilla.org/en-US/docs/Web/API/Request/signal) in the `RequestInit` options use to create the proxy request

Currently Cloudflare doesn't expose the `abort` event to Workers, but it looks like there is work to implement it, see:

- https://github.com/cloudflare/workerd/issues/3033
- https://github.com/cloudflare/workerd/pull/3488

I'm unsure about other platforms..

fixes #3924

### DIFF
--- a/src/helper/proxy/index.test.ts
+++ b/src/helper/proxy/index.test.ts
@@ -7,6 +7,18 @@ describe('Proxy Middleware', () => {
       global.fetch = vi.fn().mockImplementation(async (req) => {
         if (req.url === 'https://example.com/ok') {
           return Promise.resolve(new Response('ok'))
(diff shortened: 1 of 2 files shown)


usualoma:
Hi @BarryThePenguin. Thank you. LGTM!

yusukebe (approved):
LGTM!

PR #3927, opened 2025-02-17
fix(helper/proxy): missing proxy response status

### LINKED ISSUE #3926: `hono/proxy` does not preserve `response.status`
It doesn't look like the original [Response#status](https://developer.mozilla.org/en-US/docs/Web/API/Response/status) is included in the proxy response

I'm unsure if this is an issue with `hono`, or an issue somewhere else eg; `node`, `undici`, or `miniflare`? It's even an issue when running in Cloudflare Workers

BarryThePenguin:
- [x] add failing test
- [x] find out why the response status is missing (default value of 200)

Turns out that `Request` and `Response` classes do not have enumerable properties, so using the spread operator yields an empty object 🤯 
```js
console.log({ ...new Request('http://a') }) // {}
console.log({ ...new Response(null) }) // {}
```
I haven't found where this is mentioned in the [standards documentation](https://fetch.spec.whatwg.org/) 😩 

fixes #3926

### DIFF
--- a/src/helper/proxy/index.test.ts
+++ b/src/helper/proxy/index.test.ts
@@ -227,7 +227,7 @@ describe('Proxy Middleware', () => {
       expect(await res.text()).toBe('client disconnect')
     })
 
-    it.fails('not found', async () => {
+    it('not found', async () => {
       const app = new Hono()
       app.get('/proxy/:path', (c) => proxy(`https://example.com/${c.req.param('path')}`))
       const res = await app.request('/proxy/404')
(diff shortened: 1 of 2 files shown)


yusukebe (approved):
LGTM!

yusukebe:
Hi @BarryThePenguin 

Looks good to me. But it has conflict. Can you resolve it?

usualoma:
Thank you!

yusukebe:
@BarryThePenguin 

Thanks! Merging now.

PR #3928, opened 2025-02-17
fix(proxy): remove spread operator from `Request` and `Response` classes

BarryThePenguin:
Removes usage of the spread operator (`...`) on `Request` and `Response` classes

Follow-up from #3927

### DIFF
--- a/src/helper/proxy/index.ts
+++ b/src/helper/proxy/index.ts
@@ -108,12 +108,10 @@ const preprocessRequestInit = (requestInit: RequestInit): RequestInit => {
  * })
  * ```
  */
-export const proxy: ProxyFetch = async (input, proxyInit) => {
-  const { raw, ...requestInit } = proxyInit ?? {}
-
+export const proxy: ProxyFetch = async (input, proxyInit = {}) => {
   const req = new Request(input, {
-    ...buildRequestInitFromRequest(raw),
-    ...preprocessRequestInit(requestInit as RequestInit),
+    ...buildRequestInitFromRequest(proxyInit.raw),
+    ...preprocessRequestInit(proxyInit as RequestInit),
   })
   req.headers.delete('accept-encoding')
 
--- a/src/helper/ssg/ssg.test.tsx
+++ b/src/helper/ssg/ssg.test.tsx
@@ -701,14 +701,14 @@ describe('Combined Response hooks - modify response content', () => {
   const prependContentAfterResponseHook = (prefix: string): AfterResponseHook => {
     return async (res: Response): Promise<Response> => {
       const originalText = await res.text()
-      return new Response(`${prefix}${originalText}`, { ...res })
+      return new Response(`${prefix}${originalText}`, res)
     }
   }
 
   const appendContentAfterResponseHook = (suffix: string): AfterResponseHook => {
     return async (res: Response): Promise<Response> => {
       const originalText = await res.text()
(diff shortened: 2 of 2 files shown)


usualoma:
Hi @BarryThePenguin. Thank you.

Please do not modify the `proxyInit` object received as an argument, as follows.

```diff
diff --git a/src/helper/proxy/index.ts b/src/helper/proxy/index.ts
index 4c010a2d..93a7543b 100644
--- a/src/helper/proxy/index.ts
+++ b/src/helper/proxy/index.ts
@@ -49,13 +49,13 @@ const buildRequestInitFromRequest = (
   }
 }
 
-const preprocessRequestInit = (requestInit: RequestInit): RequestInit => {
+const preprocessRequestInit = (requestInit: Readonly<ProxyRequestInit>): RequestInit => {
   if (
     !requestInit.headers ||
     Array.isArray(requestInit.headers) ||
     requestInit.headers instanceof Headers
   ) {
-    return requestInit
+    return requestInit as RequestInit
   }
 
   const headers = new Headers()
@@ -67,8 +67,10 @@ const preprocessRequestInit = (requestInit: RequestInit): RequestInit => {
       headers.set(key, value)
     }
   }
-  requestInit.headers = headers
-  return requestInit
+  return {
+    ...requestInit,
+    headers,
+  }
 }
 
 /**
@@ -111,7 +113,7 @@ const preprocessRequestInit = (requestInit: RequestInit): RequestInit => {
 export const proxy: ProxyFetch = async (input, proxyInit = {}) => {
   const req = new Request(input, {
     ...buildRequestInitFromRequest(proxyInit.raw),
-    ...preprocessRequestInit(proxyInit as RequestInit),
+    ...preprocessRequestInit(proxyInit),
   })
   req.headers.delete('accept-encoding')
 
```


Also, if possible, since the changes to src/helper/proxy/index.ts are different from the following, it would be helpful if you could include only one purpose in one PR.

> Removes usage of the spread operator (`...`) on `Request` and `Response` classes

BarryThePenguin:
> Also, if possible, since the changes to src/helper/proxy/index.ts are different from the following, it would be helpful if you could include only one purpose in one PR.

I was trying to avoid using the spread operator on `proxyInit`

https://github.com/honojs/hono/blob/6ae021371a38216a02f6fbddc06eebff506dfd04/src/helper/proxy/index.ts#L113

As the current API for `proxy` makes it very easy to pass `Request` as `proxyInit`, for example
```ts
app.get('/proxy/:path', (c) => {
  return proxy(`http://${originServer}/${c.req.param('path')}`, c.req.raw)
})
```
Though I'm not sure if this is intentional or not. I'll revert the changes to `src/helper/proxy/index.ts`

usualoma:
Hi @BarryThePenguin. Thank you for getting back to me.

I see, as you say, it would be better if the following code (although it is not the recommended way to use it) processed `c.req.raw` as `RequestInit`. Or, if it is not processed, it would be better if it caused an error.

```ts
app.get(‘/proxy/:path’, (c) => {
  return proxy(`http://${originServer}/${c.req.param(‘path’)}`, c.req.raw)
})
```

Let's discuss it in another PR.

yusukebe (approved):
LGTM!

yusukebe:
@BarryThePenguin 

Thank you. Looks good. I'll merge this now.

PR #3932, opened 2025-02-18
refactor(compose): remove unnecessary complexity and improve testing

EdamAme-x:
- ### About
Overall `compose` processing has been improved.

The biggest improvement was to minimize Context compatibility.

For some reason loose Context was allowed, but since it was a 3 year old commit and only used for testing, I decided that the current code base could be rewritten.

Although I did not actually measure it, I believe that performance increased in no small way.

- ### Others

- I improved the tests and created some additional tests.
- Rearranged the conditions of the if statement to maximize performance.

- #### Is this a breaking change?

Basically, I don't think it can be a breaking change since we can expect the complete Context to be used.

- ### The author should do the following, if applicable

- [x] Run tests
- [x] Add tests

### DIFF
--- a/src/compose.test.ts
+++ b/src/compose.test.ts
@@ -2,12 +2,6 @@ import { compose } from './compose'
 import { Context } from './context'
 import type { Params } from './router'
 
-type C = {
-  req: Record<string, string>
-  res: Record<string, string>
-  finalized: boolean
-}
-
 type MiddlewareTuple = [[Function, unknown], Params]
 
 class ExpectedError extends Error {}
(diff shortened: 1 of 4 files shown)


usualoma:
Hi @EdamAme-x. Thank you!

I didn't change it because there was no difference when I measured it, but since `compose` receives something that is not `Context` only in the test, I too thought it was safe to assume that it was `Context` for use in the hono core.
I agree with this change.

yusukebe:
@EdamAme-x 

That's a good point! I also this change is okay since `compose.ts` is used for `hono` internal only. I've confirmed that the bundle size will decrease slightly with this change.

Let's go with this.

yusukebe (approved):
LGTM!

yusukebe:
This change does not introduce a new feature or fix the problem. So, I'll label this as `refactor`.

PR #3957, opened 2025-02-25
feat(client/POC): RPC support for SSE

### LINKED ISSUE #3309: RPC support for SSE

We require using EventSource to work with SSE routes, but can we have Hono's RPC to have this functionality as well?

```ts
const client = hc<typeof app>(`https://${BASE_URL}`)
client.sse.$get({
  onMessage(event) { /* do something with this */ },
  // other callbacks
}, {
  headers: {
    Authorization: 'Bearer booyah'
  }
})
```

sor4chi:
resolve #3309 

## Todo

- [x] SSE rpc handler
- [ ] FetchEvent (polyfill?)
  - [ ] LICENCE inclusion
- [ ] Typed Data interface for JSON SSE

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -11,6 +11,8 @@ import type { Equal, Expect, JSONValue, SimplifyDeepArray } from '../utils/types
 import { validator } from '../validator'
 import { hc } from './client'
 import type { ClientResponse, InferRequestType, InferResponseType } from './types'
+import { streamSSE } from '../helper/streaming'
+import { EventSource, EventSourceInit } from './eventsource'
 
 describe('Basic - JSON', () => {
   const app = new Hono()
@@ -1333,5 +1335,38 @@ describe('WebSocket Provider Integration', () => {
     })
     client.index.$ws({ query })
     expect(webSocketMock).toHaveBeenCalledWith(expectedUrl, undefined)
+  })
+})
+
+describe('SSE Provider Integration', () => {
+  it('should initialize the SSE provider correctly', async () => {
+    const app = new Hono()
+
+    const route = app.get('/sse-endpoint', (c) =>
+      streamSSE(c, async (stream) => {
+        await stream.writeSSE({
+          data: 'Hello from server!',
+          event: 'message',
+        })
+      })
+    )
+
+    type AppType = typeof route
+
+    const client = hc<AppType>('http://localhost')
+    assertType<(args: EventSourceInit) => EventSource>(client['sse-endpoint'].$sse)
+    const es = client['sse-endpoint'].$sse()
+
+    expect(es.url).toBe('http://localhost/sse-endpoint')
+
+    const eventData = await new Promise<MessageEvent>((resolve) => {
+      es.onmessage = (data) => {
+        expect(data).toBeInstanceOf(MessageEvent)
+        resolve(data)
+      }
+    })
+
+    expect(eventData.data).toBe('Hello from server!')
+    expect(eventData.type).toBe('message')
   })
 })
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -2,6 +2,7 @@ import type { Hono } from '../hono'
 import type { FormValue, ValidationTargets } from '../types'
 import { serialize } from '../utils/cookie'
 import type { UnionToIntersection } from '../utils/types'
+import { EventSource } from './eventsource'
 import type { Callback, Client, ClientRequestOptions } from './types'
 import {
   buildSearchParams,
@@ -192,6 +193,9 @@ export const hc = <T extends Hono<any, any, any>>(
       }
 
       return establishWebSocket(targetUrl.toString())
+    }
+    if (method === 'sse') {
+      return new EventSource(url, opts.args[0])
     }
 
     const req = new ClientRequestImpl(url, method)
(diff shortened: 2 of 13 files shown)


sor4chi:
Hi, @yusukebe

This time, I would like to introduce Typed SSE that bypasses JSON as a feature of SSE RPC. Typed SSE is very useful, so I would like to introduce it. Currently, the data field of SSE can specify `string | Promise<string>`, and the reason why it is not simply a string is that #3344 implemented by @usualoma allows Hono Stream SSE to send Hono JSX (HTML). The problem is with the string. Most Hono Stream SSE users probably specify `JSON.stringify` in the data field. However, if we implement it straightforwardly, the JSON serialization and deserialization operations will be performed in the Hono RPC. Therefore, the client needs to distinguish between the JSON.stringify data and the data serialized by the RPC as Typed SSE, and we are currently exploring how to implement this while maintaining backward compatibility. If you have any good ideas, I would appreciate it if you could let me know.

nakasyou:
> The problem is with the string. Most Hono Stream SSE users probably specify JSON.stringify in the data field. However, if we implement it straightforwardly, the JSON serialization and deserialization operations will be performed in the Hono RPC. Therefore, the client needs to distinguish between the JSON.stringify data and the data serialized by the RPC as Typed SSE, and we are currently exploring how to implement this while maintaining backward compatibility. If you have any good ideas, I would appreciate it if you could let me know.

How about adding `streamJSONRPC` and dividing `.$sse` into `.$sse` and `.$sseJSON`?

yusukebe:
Hi @sor4chi 

I don't want to introduce the polyfill of EventSource Parser. It increases package size and maintenance costs. Can you make it work with extending global objects like `globalThis.document`? Or any good solution?

sor4chi:
I can't think of a way to do this at the moment other than to cut it out as a sub-package and make it load only in environments where EventSource is available, such as a browser, and have the user explicitly switch between the two.

nakasyou:
> It increases package size and maintenance costs.

This is an idea to reduce package size. What do you think of making $sse a function instead of a method. e.g:
```ts
import { sse, hc } from 'hono/client'

const source = sse(client.stream.get())
```
By doing that, the codebase will tree-shakeable. And we will be able to use POST requests to receive streaming. For example, most of LLM APIs such as OpenAI use POST method to stream content. It means that we can't use EventSource. Similarly, there are many users who want to use POST method to stream content using Hono RPC I guess.

yusukebe:
@sor4chi @nakasyou 

Thank you for the comments!

The problem I'm having is not the bundle size but the package size. In other words, the amount of code included in the `hono` package is increasing (already that is fat).

In this PR, you are using the `eventsource` and `eventsource-parser` code, so I think it is almost the same as using external libraries. So, how about making it a 3rd party middleware called `@hono/sse-client` based on @nakasyou's idea? This can depend on `eventsource` and `eventsource-parser`.

```ts
import { sse } from '@hono/sse-client'

const source = sse(client.stream.get())
```

yusukebe:
@sor4chi Thank you for your understanding!

RobertAron:
@yusukebe @sor4chi 

I think I have a working package that will work... any recommendations/guidance on publishing this, or general 3rd party packages in general?


Screenshot 2025-07-23 at 11 52 16 PM

yusukebe:
Hi @RobertAron 

Does that depend on other libraries, such as `eventsource`?

RobertAron:
It does depend on `eventsource`. My plan is to have two different exports. One for `server` which can be used in runtimes that need the polyfill, and another export that doesn't import that dependency. It still has the issues you mentioned previously - the package size will be non-trivial. It should bundle nicely though.

RobertAron:
[Here's the repo](https://github.com/RobertAron/hono-type-sse/tree/main). I haven't actually published the package but I think I'm 95% ready. I'll probably make a few more tiny tweaks tomorrow and post about it in the other thread.

yusukebe:
@RobertAron I see!

> [Here's the repo](https://github.com/RobertAron/hono-type-sse/tree/main). I haven't actually published the package but I think I'm 95% ready. I'll probably make a few more tiny tweaks tomorrow and post about it in the other thread.

Oh, [the repo](https://github.com/RobertAron/hono-type-sse/) returns `404`.

RobertAron:
@yusukebe sorry about that! Should be good now.

yusukebe:
@RobertAron 

Thank you for the project.

Ideally, we would like to be able to use the syntax `client[‘sse-endpoint’].$sse()` and have it fully integrated into `hono`. Therefore, your library cannot be included in the `hono` package.

You should publish the package as is and link to it from the Hono website `https://hono.dev` or add documentation to [`examples`](https://hono.dev/examples/).

destroyer22719:
Hi there, does anyone have any updates on this? What you're working on seems really amazing

RobertAron:
@destroyer22719 I just updated the package to be a bit more stable. Bun actually [removed eventsource](https://github.com/oven-sh/bun/pull/26127) but now there is a new pull request to re-implement it.

I think my current implementation should be mostly solid. If I manage to find time at work I might try it with a more serious codebase and see if it works in most conditions. Until then It's pretty experimental.

All the logic is in 2 pretty small files though. If this is something you need you could probably copy the code from my repo.

PR #3961, opened 2025-02-27
feat(hono/jwk): Extended with `allow_anon` option & passing `Context` to callbacks

Beyondo:
Adding two main useful features people have been asking for in the [original PR](https://github.com/honojs/hono/pull/3826) + some slight changes.

### New features:
- **Allow anonymous requests** by setting the new **`allow_anon`** option to `true`
  - https://github.com/honojs/hono/pull/3826#issuecomment-2656576941
- **Optional callback for the `jwks_uri`** instead of only a hard-coded string
- **`Context`** is now **included in all callbacks**, so you can **e.g** pull information from `c.env` such as auth server
  - https://github.com/honojs/hono/pull/3826#issuecomment-2687416251
- Callbacks now can be either **async or sync**

**Example:**
```ts
app.use(
  '/auth/*',
  jwk({
    jwks_uri: (c) => `https://${c.env.authServer}/.well-known/jwks.json`,
    allow_anon: true
  })
)
```
**Note 1**: This PR allows having a single endpoint for both authenticated and non-authenticated requests through `allow_anon`. The variable `c.get("jwtPayload")` can be used **to differentiate authenticated requests from anonymous requests** since hono requires requests to have `exp` and `iat` in the payload (in the `verify` stage) which means the payload would always be defined for authenticated requests and be reliably used as indicator.

**Note 2**: Setting `allow_anon` to `true` means requests without a token present (in either header/cookie) would be allowed to pass through the middleware successfully.

**Tests**: Added 3 extra tests + modified one test to cover `jwks_uri` being a function.
```
✓ src/middleware/jwk/index.test.ts (34)
 ```
 
This PR breaks no code at all except for `Jwt.verifyFromJwks(...)`—specifically removed callbacks support from it since they're redundant and can be called outside, and also since only the middleware can provide the `Context` variable to the callbacks.

### DIFF
--- a/src/middleware/jwk/index.test.ts
+++ b/src/middleware/jwk/index.test.ts
@@ -43,7 +43,7 @@ describe('JWK', () => {
     })
   })
 
-  describe('options.skip_if_no_token = true', () => {
+  describe('options.allow_anon = true', () => {
     let handlerExecuted: boolean
 
     beforeEach(() => {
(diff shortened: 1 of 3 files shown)


Beyondo:
Hmm, I don't think that Deno error is caused by my PR as the errored part was pulled from the upstream, but I will attempt to fix it.

flexchar:
Looks very promising, thank you a lot for your prompt contribution, wise person across the world! 👌

yusukebe:
@Beyondo 

Sorry for being late. It looks good! I'll merge this later into the next branch for the next minor version and release it. Thanks!

PR #3972, opened 2025-03-02
fix(factory): support `defaultAppOptions` after calling `createApp`

### LINKED ISSUE #3970: defaultAppOptions not applying for the second app instance

4.7.2

Deno,Node

this code's assertions fail :
```ts
    const factory = createFactory({ defaultAppOptions: { strict: false } })
    const firstApp = factory.createApp()
    const secondApp = factory.createApp()
    secondApp.get('/hello', (c) => c.text('hello'))
    const res = await secondApp.request('/hello/')
    expect(res.status).toBe(200)
    expect(await res.text()).toBe('hello')
```

`strict: false` also applies to the second Hono instance

### What do you see instead?

the second Hono instance does not include `strict:false` option in it

### Additional information

the workaround for this issue is to simply pass an empty object to `createApp`

actually I've faced this bug when using a nested route in hono but I couldn't have reproduced the issue the same way that I faced it , but I've read the code and located the issue then I created this examplel

yusukebe:
Fixes #3970

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/helper/factory/index.test.ts
+++ b/src/helper/factory/index.test.ts
@@ -383,6 +383,16 @@ describe('createFactory', () => {
     expect(res.status).toBe(200)
     expect(await res.text()).toBe('hello')
   })
+
+  it('Should apply the default app options after calling createApp', async () => {
+    const factory = createFactory({ defaultAppOptions: { strict: false } })
+    factory.createApp() // Call once
+    const secondApp = factory.createApp()
+    secondApp.get('/hello', (c) => c.text('hello'))
+    const res = await secondApp.request('/hello/')
+    expect(res.status).toBe(200)
+    expect(await res.text()).toBe('hello')
+  })
 })
 
 describe('Lint rules', () => {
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @ehsan2003

Thank you for raising issue #3970. This PR can fix the problem. Can you review this?

ehsan2003:
Hi

thanks for you're attention

although this is going to fix the issue with factory it won't be fixing the root cause 

the main issue rises from the `Hono` constructor where it mutates it's inputs which may result in unexpected results 
for example the code bellow doesn't work as expected:

```ts
const opts = {strict:false}
const a1 = new Hono(opts)
const a2 = new Hono(opts)

a2.get('/hello',c=>c.text("hello"));

const res = await a2.request('/hello/')
expect(res.status).toBe(200)
expect(await res.text()).toBe('hello')
```
the assertion would fail because hono constructor deletes the strict option here 


I already opened #3971 can't it be merged ?

yusukebe:
@ehsan2003 

Thank you for the explanation and for showing the test code.

As you mentioned, this is the `src/hono-base.ts` problem, not only for `src/helper/factory/index.ts`. So I created PR #3975 with the proper title and correct test place to focus on fixing `src/hono-base.ts` by adding you as a co-author.

PR #3975, opened 2025-03-04
fix(hono-base): prevent options object mutation

### LINKED ISSUE #3970: defaultAppOptions not applying for the second app instance

yusukebe:
Fixes #3970
Based on #3971

As the [comment](https://github.com/honojs/hono/pull/3972#issuecomment-2698529533), the following test does not pass.

```ts
const opts = { strict: false }
const a1 = new Hono(opts)
const a2 = new Hono(opts)

a2.get('/hello', (c) => c.text('hello'))

const res = await a2.request('/hello/')
expect(res.status).toBe(200)
expect(await res.text()).toBe('hello')
```

This means `opts` passed to Hono's options will be mutated. This PR will fix it.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/hono-base.ts
+++ b/src/hono-base.ts
@@ -162,10 +162,9 @@ class Hono<E extends Env = Env, S extends Schema = {}, BasePath extends string =
       return this as any
     }
 
-    const strict = options.strict ?? true
-    delete options.strict
-    Object.assign(this, options)
-    this.getPath = strict ? options.getPath ?? getPath : getPathNoStrict
+    const { strict, ...optionsWithoutStrict } = options
+    Object.assign(this, optionsWithoutStrict)
+    this.getPath = strict ?? true ? options.getPath ?? getPath : getPathNoStrict
   }
 
   #clone(): Hono<E, S, BasePath> {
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @usualoma !

Can you review this?

yusukebe:
@usualoma Thanks!

Hi @ehsan2003 

You have created [the PR](https://github.com/honojs/hono/pull/3971), but I will merge this PR, making you a co-author.  I appreciate your contribution!

Issue #3872, opened 2025-01-31
Get encoded path params


kawmra:
### What is the feature you are proposing?

Currently, with the following route, there is no way to differentiate between accessing `/kv/foo/bar` and `/kv/foo%2Fbar`:

```ts
app.get("/kv/:key{.+}", async (c) => {
  console.log(c.req.param("key")) // => Always returns "foo/bar" in both cases
})
```

It seems that since #659, `c.req.param()` now returns a decoded string by default.
But I want to distinguish whether the input parameter contained a literal `/` or if it was an encoded `%2F`.

I know that `c.req.path` can be used to get the raw path, but manually parsing it is cumbersome.
And if the route hierarchy (`/kv`) changes, the position of `:key` would shift, and there's no type checking to detect such changes, which is concerning.

A method like `c.req.rawParam()` would be useful I think.

yusukebe:
Hi @kawmra 

I think you can try the new route helper: https://hono.dev/docs/helpers/route

kawmra:
I think Route Helper can add some resilience against changes in path structures! Thank you @yusukebe !

yusukebe:
@kawmra 

Great! Can we close this issue?

kawmra:
@yusukebe 
I don’t think Route Helper fully solves this issue because, even with `routePath()`, we can’t reference parameters by name, and it still depends on the path structure.
But I saw PR #4000, if you decide not to approve this feature request, please feel free to close it!

yusukebe:
@kawmra

Yes. #4000 is the same issue and we decided not to approve it. Please follow the result. I'll close this now. Thank you for using Hono!

PR #4000, opened 2025-03-15
feat(request): add `rawParam()`

Abhishek-jha-96:
Following ISSUE-#3872 
- Added rawParam() method to retrieve raw (encoded) path parameters without decoding percent-encoded characters.
- Implemented getRawParam() (private method) to fetch raw parameter values from #matchResult.

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -74,6 +74,7 @@ export class HonoRequest<P extends string = '/', I extends Input['out'] = {}> {
 
   /**
    * `.req.param()` gets the path parameters.
+   * `.req.rawParam()` gets the raw (encoded) path parameters.
    *
    * @see {@link https://hono.dev/docs/api/routing#path-parameter}
    *
@@ -82,6 +83,8 @@ export class HonoRequest<P extends string = '/', I extends Input['out'] = {}> {
    * const name = c.req.param('name')
    * // or all parameters at once
    * const { id, comment_id } = c.req.param()
+   * 
+   * const rawKey = c.req.rawParam('key') // Encoded
    * ```
    */
   param<P2 extends ParamKeys<P> = ParamKeys<P>>(key: P2 extends `${infer _}?` ? never : P2): string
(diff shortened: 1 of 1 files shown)


yusukebe:
@Abhishek-jha-96 Thank you for the PR. We have to be careful about adding code to `request.ts` because this is the core module of `hono.`

@usualoma, What do you think of this feature? I'd like to know your opinion on this feature without implementation things.

Abhishek-jha-96:
@yusukebe I totally agree that modifying `request.ts` should be done cautiously and only when necessary.
 I feel rawParam() can be useful in some cases like:
 - Signed URLs & API authentication where it requires handling encoded values exactly as they are received.
 
 we could consider adding this as a utility method or exploring other ways.
 Let me know your thoughts on the best approach.

usualoma:
Hi @yusukebe, @Abhishek-jha-96, Thank you.

Hmm, this is a surprisingly difficult issue.

In the hono default, the path passed to the router is decoded using `getPath,` so it's a bit of a grey area as to whether the value obtained in this PR can be called `raw.`

Also, if possible, I don't want to add a method to `HonoRequest`.

> Signed URLs

I understand what you're saying, but I think that "the signature target is contained in a path component below a specific path" (e.g., `/path/to/:key{.+}` ) is a minority case. In most cases where a URL is the signature target, I think it's the whole thing, such as `c.req.path`.

I can understand that there are use cases where a key in a key-value store is needed, as in #3872, but in that case, I think it would be better to have the request side encodeURIComponent the part corresponding to `:key.`

If an app really needs the raw value, I think it would be better to get it from `c.req.path`.

Abhishek-jha-96:
@usualoma Thank you for your view on this. I completely agree with your point that if we truly need the raw value, it's better to extract it from the path itself, as this approach is followed by other frameworks as well.

Abhishek-jha-96:
I now also feel that the need for this feature is quite niche, and it's better not to proceed further with it. @yusukebe I feel we can close this issue. Let me know your thoughts!

yusukebe:

@aantthony I can understand why you want this feature, but as you said, this is a niche, and we don't want to add the function to `HonoRequest`. But you have a way to get the value, though it's a little bit annoying.

Okay, I'll close this now. Thank you for suggesting it and understanding!

PR #4006, opened 2025-03-17
perf(logger): cache ColorEnabled value

markH5:
Maybe no need to check  it in loop ?
Just check one

### DIFF
--- a/src/middleware/logger/index.ts
+++ b/src/middleware/logger/index.ts
@@ -13,7 +13,8 @@ enum LogPrefix {
 }
 
 const humanize = (times: string[]) => {
-  const [delimiter, separator] = [',', '.']
+  const delimiter = ','
+  const separator = '.'
 
   const orderTimes = times.map((v) => v.replace(/(\d)(?=(\d\d\d)+(?!\d))/g, '$1' + delimiter))
 
@@ -25,8 +26,7 @@ const time = (start: number) => {
   return humanize([delta < 1000 ? delta + 'ms' : Math.round(delta / 1000) + 's'])
 }
 
-const colorStatus = (status: number) => {
-  const colorEnabled = getColorEnabled()
+const colorStatus = (status: number, colorEnabled: boolean) => {
   if (colorEnabled) {
     switch ((status / 100) | 0) {
       case 5: // red = error
@@ -52,13 +52,14 @@ function log(
   prefix: string,
   method: string,
   path: string,
+  colorEnabled: boolean,
   status: number = 0,
   elapsed?: string
 ) {
   const out =
     prefix === LogPrefix.Incoming
       ? `${prefix} ${method} ${path}`
-      : `${prefix} ${method} ${path} ${colorStatus(status)} ${elapsed}`
+      : `${prefix} ${method} ${path} ${colorStatus(status, colorEnabled)} ${elapsed}`
   fn(out)
 }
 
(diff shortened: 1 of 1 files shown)


src/middleware/logger/index.ts
```diff
@@ -13,7 +13,8 @@ enum LogPrefix {
 }
 
 const humanize = (times: string[]) => {
-  const [delimiter, separator] = [',', '.']
```
yusukebe, line 16: These lines should not be changed.

src/middleware/logger/index.ts
```diff
@@ -52,13 +52,14 @@ function log(
...
   path: string,
+  colorEnabled: boolean,
   status: number = 0,
   elapsed?: string
 ) {
   const out =
     prefix === LogPrefix.Incoming
       ? `${prefix} ${method} ${path}`
-      : `${prefix} ${method} ${path} ${colorStatus(status)} ${elapsed}`
+      : `${prefix} ${method} ${path} ${colorStatus(status, colorEnabled)} ${elapsed}`
```
yusukebe, line 62: I like the following implementation. This is because the `colorStatus` argument can be left as one. I think it's better not to have the `colorEnabled` argument here.

```diff
diff --git a/src/middleware/logger/index.ts b/src/middleware/logger/index.ts
index 5bb45383..a358e4c5 100644
--- a/src/middleware/logger/index.ts
+++ b/src/middleware/logger/index.ts
@@ -26,18 +26,15 @@ const time = (start: number) => {
 }

 const colorStatus = (status: number) => {
-  const colorEnabled = getColorEnabled()
-  if (colorEnabled) {
-    switch ((status / 100) | 0) {
-      case 5: // red = error
-        return `\x1b[31m${status}\x1b[0m`
-      case 4: // yellow = warning
-        return `\x1b[33m${status}\x1b[0m`
-      case 3: // cyan = redirect
-        return `\x1b[36m${status}\x1b[0m`
-      case 2: // green = success
-        return `\x1b[32m${status}\x1b[0m`
-    }
+  switch ((status / 100) | 0) {
+    case 5: // red = error
+      return `\x1b[31m${status}\x1b[0m`
+    case 4: // yellow = warning
+      return `\x1b[33m${status}\x1b[0m`
+    case 3: // cyan = redirect
+      return `\x1b[36m${status}\x1b[0m`
+    case 2: // green = success
+      return `\x1b[32m${status}\x1b[0m`
   }
   // Fallback to unsupported status code.
   // E.g.) Bun and Deno supports new Response with 101, but Node.js does not.
@@ -52,13 +49,14 @@ function log(
   prefix: string,
   method: string,
   path: string,
+  colorEnabled: boolean,
   status: number = 0,
   elapsed?: string
 ) {
   const out =
     prefix === LogPrefix.Incoming
       ? `${prefix} ${method} ${path}`
-      : `${prefix} ${method} ${path} ${colorStatus(status)} ${elapsed}`
+      : `${prefix} ${method} ${path} ${colorEnabled ? colorStatus(status) : status} ${elapsed}`
   fn(out)
 }
```

yusukebe: @usualoma, What do you think of this implementation?

yusukebe:
Hi @markH5 

Thank you for the PR. I think this is a good improvement. This will affect the performance issue, so I labeled this PR as `perf`.

src/middleware/logger/index.ts
usualoma: > ${colorEnabled ? colorStatus(status) : status}

That's good, I was just thinking the same thing!

src/middleware/logger/index.ts
markH5: this change doesn't align with the minor tweak mentioned in the title.
I believe modern JS engines will likely inline it after a few iterations anyway. That said, it still introduces unnecessary runtime overhead.

usualoma: It might be better to use a different PR, but I think it would be better to do the following because there is a lot of waste in the current `humanize()`.

```ts
/**
 * Humanize the time.
 * @param {number} time - The time to humanize.
 * @returns {string} The humanized time.
 * 
 * @example
 * ```ts
 * humanize(1000) // => 1,000
 * ```
 */
const humanize = (time: number): string =>
  time < 1000 ? String(time) : String(time).replace(/(\d)(?=(\d\d\d)+(?!\d))/g, '$1' + ',')

const time = (start: number): string => {
  const delta = Date.now() - start
  return delta < 1000 ? `${delta}ms` : `${humanize(Math.round(delta / 1000))}s`
}
```

usualoma: Or rather, `humanize()` only does its job when the request is over 1000 seconds long, but since there are no such long requests normally, I feel that `humanize()` itself is unnecessary.

PR #4008, opened 2025-03-18
fix: fix absolute path definition for current implementation

usualoma:


usualoma:
36243d1 was necessary to make it work with `wrangler`.
I would have preferred to find a smarter way to get around this, such as changing the settings in `wrangler`, but I couldn't find a good, concise way to do it.

yusukebe:
Thanks! Merging.

PR #4012, opened 2025-03-18
PoC: feat: Allow handlers and middleware to return functions of the same type as themselves

usualoma:
This PR will add a convenient feature to hono, but it will also increase the complexity, so please be careful. It is currently at the PoC stage.

### What does this enable us to do?

We can dynamically configure middleware with very clean code.
Of course, I know that authentication middleware has the ability to dynamically determine credentials, but with this feature, it is easy to pass settings dynamically even if the middleware does not support it.

```ts
const app = new Hono()
  .get('/shortcut-icon.svg', () => {
    const path = 'dynamic-deterministic-filename.svg'
    return serveStatic({ path })
  })
  .get('/', (c) => {
    return c.text('Hello World')
  })
```

```ts
const app = new Hono()
  .post('*', (c) => {
    const limit = new URL(c.req.url).host === 'localhost' ? 1024 * 1024 * 10 : 1024 * 1024
    return bodyLimit({ maxSize: limit })
  })
  .post('/upload', uploadHandler)
```

```ts
const app = new Hono()
  .post('*', (c) => {
    const username = c.req.header('Authorization')?.split(' ')[1] ?? ''
    return basicAuth({ username, password: passwordMap[username] })
  })
  .get('/', (c) => {
    return c.text('Hello World')
  })
```

### Isn't this just syntax sugar?

Yes, we can call it with `(c, next)` and return it as it is now, so it is safe to think of it as syntax sugar that omits `(c, next)`.

```ts
const app = new Hono()
  .get('/shortcut-icon.svg', (c, next) => {
    const path = 'dynamic-deterministic-filename.svg'
    return serveStatic({ path })(c, next)
  })
  .get('/', (c) => {
    return c.text('Hello World')
  })
```

### When is it particularly useful?

We will be able to solve issues that can be solved simply by using existing middleware in a clean way.

https://github.com/honojs/node-server/issues/205

### Is there any degradation in performance?

I don't think there is any decrease in performance at runtime.
When I compared them in "benchmarks/handle-event", there was no decrease in node and bun.

```
itty-router x 376,693 ops/sec ±4.26% (85 runs sampled)
sunder x 465,962 ops/sec ±3.09% (86 runs sampled)
worktop x 224,003 ops/sec ±4.51% (85 runs sampled)
Hono current main x 513,444 ops/sec ±5.52% (81 runs sampled)
Hono feat:returns-middleware x 525,498 ops/sec ±5.83% (80 runs sampled)
Fastest is Hono feat:returns-middleware,Hono current main
```

There may also be an impact on performance for "types", but I don't know the details of that, so I haven't been able to measure it.

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/adapter/cloudflare-pages/handler.ts
+++ b/src/adapter/cloudflare-pages/handler.ts
@@ -60,41 +60,48 @@ export function handleMiddleware<E extends Env = {}, P extends string = any, I e
       executionCtx,
     })
 
-    let response: Response | void = undefined
+    let response: Response | Function | void = undefined
 
-    try {
-      response = await middleware(context, async () => {
-        try {
-          context.res = await executionCtx.next()
-        } catch (error) {
-          if (error instanceof Error) {
-            context.error = error
-          } else {
-            throw error
+    for (;;) {
+      try {
+        response = await middleware(context, async () => {
+          try {
+            context.res = await executionCtx.next()
+          } catch (error) {
+            if (error instanceof Error) {
+              context.error = error
+            } else {
+              throw error
+            }
           }
+        })
+
+        if (typeof response === 'function') {
+          middleware = response as typeof middleware
+          continue
         }
-      })
-    } catch (error) {
-      if (error instanceof Error) {
-        context.error = error
-      } else {
-        throw error
+      } catch (error) {
+        if (error instanceof Error) {
+          context.error = error
+        } else {
+          throw error
+        }
       }
-    }
 
-    if (response) {
-      return response
-    }
+      if (response) {
+        return response as Response
+      }
 
-    if (context.error instanceof HTTPException) {
-      return context.error.getResponse()
-    }
+      if (context.error instanceof HTTPException) {
+        return context.error.getResponse()
+      }
 
-    if (context.error) {
-      throw context.error
-    }
+      if (context.error) {
+        throw context.error
+      }
 
-    return context.res
+      return context.res
+    }
   }
 }
 
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -46,28 +46,35 @@ export const compose = <E extends Env = Env>(
         handler = (i === middleware.length && next) || undefined
       }
 
-      if (handler) {
-        try {
-          res = await handler(context, () => dispatch(i + 1))
-        } catch (err) {
-          if (err instanceof Error && onError) {
-            context.error = err
-            res = await onError(err, context)
-            isError = true
-          } else {
-            throw err
+      for (;;) {
+        if (handler) {
+          try {
+            res = await handler(context, () => dispatch(i + 1))
+          } catch (err) {
+            if (err instanceof Error && onError) {
+              context.error = err
+              res = await onError(err, context)
+              isError = true
+            } else {
+              throw err
+            }
+          }
+
+          if (typeof res === 'function') {
+            handler = res
+            continue
+          }
+        } else {
+          if (context.finalized === false && onNotFound) {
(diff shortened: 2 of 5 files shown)


yusukebe:
First, regarding the performance of TypeScript. I've measured the performance in [the same way](https://github.com/honojs/hono/blob/5473682ba0d768c345e2e06e6620fa6d52035fd5/.github/workflows/ci.yml#L210
) of the CI (it does not work on this PR because of a permission issue).

The `main`:

```
Files:              263
Lines:           118435
Identifiers:     116371
Symbols:         307723
Types:           216430
Instantiations: 3093926
Memory used:    709000K
I/O read:         0.02s
I/O write:        0.00s
Parse time:       0.28s
Bind time:        0.12s
Check time:       2.88s
Emit time:        0.00s
Total time:       3.27s
```

This PR #4012: 

```
Files:              263
Lines:           118454
Identifiers:     116392
Symbols:         309125
Types:           217913
Instantiations: 3098926
Memory used:    703547K
I/O read:         0.02s
I/O write:        0.00s
Parse time:       0.29s
Bind time:        0.12s
Check time:       2.91s
Emit time:        0.00s
Total time:       3.32s
```

The result is `Instantiations` will increase `5000` from the `main`. As expected, the value increased, but it wasn't huge. If this feature is good, I think the performance degradation due to the type can be ignored.

yusukebe:
Hi @usualoma 

Thank you for the proposal and the PoC.

First of all, I like this feature.

Until now, I had always thought that handlers would always return a `Response`. That's reasonable, and the great thing about Hono is that you can return a response using the raw API, like `new Response('Hello')` instead of `c.text('Hello')` etc.

However, I realized the only didn't need to be to return a `Response.` Also, I've always thought forcing users to add `(c, next)` was not a good idea. It's particularly confusing for new users.

One point that concerns me is that, until now, handlers have been defined simply as returning a `Response`. However, this PR changes that definition. Therefore, we need to tell users the type of handler. The definition of the handler in the current `type.ts` is very complicated, so I thought it would be better to present users with a simplified type like the following (I added the definition of middleware).

```ts
// Simplified:
type Handler = (c: Context, next: Next) => Response | Promise<Response> | Handler
type MiddlewareHandler = (c: Context, next: Next) => Promise<Response | void | MiddlewareHandler>
```

Anyway, this is a big feature that will change the definition of Hono's handler and middleware. We may be able to leave it for a while.

yusukebe:
Another reason I like this feature is that it allows me to write **neatly**. This may be a matter of personal taste, but it makes sense because I have always placed importance on being able to write neatly when developing Hono.

```ts
// Verbose
app.get('/admin/*', (c, next) => {
  return basicAuth({ username, password })(c, next)
})

// Neat!
app.get('/admin/*', () => {
  return basicAuth({ username, password })
})
```

usualoma:
Hi @yusukebe, Thank you for your response!

As you say, describing it as ‘neatly’ might be better.

I think the following are two of Hono's core values (there are several).
* middleware
    * easy to write
    * the core has all the essential things
    * there are many reusable middleware, including honojs/middleware and other than the core.
* neatly
    * when using Hono, we can write application code neatly

This PR will expand the range of middleware use and further strengthen Hono's core values.

### Will ‘neatness’ continue to be important?

In the Vibe Coding Era, when writing code with AI Agents, neatness may become less important than it is now. There is a possibility that, rather than being able to write concisely, people will prefer to write in a slightly more verbose but non-omitting way.

However, highly abstract and neat code will remain valuable because it is easy to read.

PR #4025, opened 2025-03-28
WebSocket: Optional Sec-WebSocket-Protocol Header

abegehr:
This is the only header browsers can pass on `new WebSocket()`-initialization: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-WebSocket-Protocol
It can be (mis-)used for auth f.e.: https://stackoverflow.com/questions/4361173/http-headers-in-websockets-client-api/77060459#77060459

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -191,6 +191,9 @@ export const hc = <T extends Hono<any, any, any>>(
         return new WebSocket(...args)
       }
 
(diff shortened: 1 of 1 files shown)


abegehr:
Nvm, I'll use the `webSocket` option on hc's init.

PR #4027, opened 2025-03-29
fix(compress): avoid compressing if transfer-encoding is set

### LINKED ISSUE #3991: Transfer-Encoding Chunked content should not be compressed

4.7.4

node

Create a request handler which sends a response with transfer-encoding chunked.
If the hono app is configured with the compression middleware, the response is compressed but must not.

The respones should not be compressed, since it's already transfer-encoded.

### What do you see instead?

Skip the compression.

### Additional information

Hi,

according to the [HTML spec](https://greenbytes.de/tech/webdav/draft-ietf-httpbis-p1-messaging-13.html#rfc.section.6.2.1) transfer-encoding is the last transform which can be applied to a response stream.

That means if a response is already transfer encoded, it must not be compressed by the hono middleware.

Are you open for a bugfix from my side?

usualoma:
fix #3991

As is the case when Transfer-Encoding is chunked, I think it is better not to re-encode in other cases as well, so I have made it so that if Transfer-Encoding is present, compressing is skipped.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Transfer-Encoding

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/compress/index.test.ts
+++ b/src/middleware/compress/index.test.ts
@@ -46,6 +46,18 @@ describe('Compress Middleware', () => {
     c.header('Content-Length', '1024')
     return c.body(new Uint8Array(1024)) // Simulated compressed data
   })
+  app.get('/transfer-encoding-deflate', (c) => {
+    c.header('Content-Type', 'application/octet-stream')
+    c.header('Transfer-Encoding', 'deflate')
+    c.header('Content-Length', '1024')
+    return c.body(new Uint8Array(1024)) // Simulated deflate data
+  })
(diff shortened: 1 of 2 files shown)


KnisterPeter:
👍

usualoma:
@yusukebe

Thank you for releasing "@hono/eslint-config@1.1.1"!

Would you please review this?

yusukebe (approved):
LGTM!

yusukebe:
@usualoma 

Thank you! Looks good. Merging.

PR #4036, opened 2025-04-01
fix(websocket): add `Uint8Array` to `WSMessageReceive`

yusukebe:
This will fix the issue pointed out in https://github.com/honojs/middleware/pull/1077#pullrequestreview-2730692813

In [@hono/node-ws](https://github.com/honojs/middleware/tree/main/packages/node-ws), the `data` in the link can be `Buffer`:

https://github.com/honojs/middleware/blob/b18f24379bb4a1350d0b1da0cb109787dddc5193/packages/node-ws/src/index.ts#L116

But currently, `WSMessageReceive` can not be `Buffer` and the type error throws:

CleanShot 2025-04-01 at 17 51 07@2x

So I added the `Uint8Array`, which is the superclass of `Buffer` to `WSMessageReceive`.

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/helper/websocket/index.ts
+++ b/src/helper/websocket/index.ts
@@ -82,7 +82,7 @@ export class WSContext<T = unknown> {
   }
 }
 
-export type WSMessageReceive = string | Blob | ArrayBufferLike
+export type WSMessageReceive = string | Blob | ArrayBufferLike | Uint8Array
 
 export const createWSMessageEvent = (source: WSMessageReceive): MessageEvent<WSMessageReceive> => {
   return new MessageEvent<WSMessageReceive>('message', {


yusukebe:
Hey @nakasyou 

Can you review this?

nakasyou:
I think it is not [Web Standard](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/message_event#event_properties). And existing code may break. For example:
```ts
upgradeWebSocket({
  async onMessage(evt) {
    let text: string
    if (evt.data instanceof ArrayBuffer) {
      text = decoder.decode(new Uint8Array(evt.data))
    } else {
      // process as blob
      text = await evt.data.text() // TypeScript will throw an error
    }
  }
})
```

So, an idea, how about using this approach?
https://github.com/nakasyou/hono-middleware/blob/53dfda6828ba560183a515ba641fd826f515d4c4/packages/node-ws/src/index.ts#L122

yusukebe:
@nakasyou 

> So, an idea, how about using this approach?

Great! But the test will fail:

CleanShot 2025-04-01 at 18 23 36@2x

But I think `receivedMessage` should not be `Buffer`. So it's good to rewrite the test? If all is okay, can you create a PR for `@hono/node-ws`?

nakasyou:
@yusukebe 

> So it's good to rewrite the test? 

Yes, I agree with you.

> If all is okay, can you create a PR for @hono/node-ws?

Of course! Check https://github.com/honojs/middleware/pull/1094.

yusukebe:
We follow the https://github.com/honojs/middleware/pull/1094 instead of this PR. Closing.

PR #4045, opened 2025-04-02
Resolve issue when proxying 204

ErreMalote:
when proxying 204 the body is null since it has no content. this makes Hono interpret the response as a 200. However, it is a 204.

this is something I found while using Hono. 

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/helper/proxy/index.ts
+++ b/src/helper/proxy/index.ts
@@ -128,7 +128,9 @@ export const proxy: ProxyFetch = async (input, proxyInit) => {
     // Content-Length is the size of the compressed content, not the size of the original content
     resHeaders.delete('content-length')
   }
-
+  // Enforce 204 has no body
+  const body = res.status === 204 ? null : res.body;
+  
   return new Response(res.body, {
     status: res.status,
     statusText: res.statusText,


yusukebe:
Hi @ErreMalote 

If the origin server returns a null body and sets the status code to the `204` property, the Proxy Helper returns the correct 204 response. So, the following test will pass. I think it's okay.

```diff
diff --git a/src/helper/proxy/index.test.ts b/src/helper/proxy/index.test.ts
index 7aeb1767..a3d195cd 100644
--- a/src/helper/proxy/index.test.ts
+++ b/src/helper/proxy/index.test.ts
@@ -58,6 +58,12 @@ describe('Proxy Middleware', () => {
               },
             })
           )
+        } else if (req.url === 'https://example.com/no-content') {
+          return Promise.resolve(
+            new Response(null, {
+              status: 204,
+            })
+          )
         }
         return Promise.resolve(new Response('not found', { status: 404 }))
       })
@@ -213,6 +219,14 @@ describe('Proxy Middleware', () => {
       expect(req.headers.get('Authorization')).toBeNull()
     })

+    it('no content', async () => {
+      const app = new Hono()
+      app.get('/proxy/:path', (c) => proxy(`https://example.com/${c.req.param('path')}`))
+      const res = await app.request('/proxy/no-content')
+      expect(res.status).toBe(204)
+      expect(res.body).toBe(null)
+    })
+
     it('client disconnect', async () => {
       const app = new Hono()
       const controller = new AbortController()
```

Or do you have any reason for adding this fix?

ErreMalote:
Hi @yusukebe 

yea, I noticed that 204 calls in my server were being CORS blocked so I had to add this and that solved the issue.

```
uniquePaths.forEach((path) => {
	app.all(`${path}/*`, async (c) => {
		// Log all request parameters
		const queryParams = c.req.query();
		const queryString = new URLSearchParams(queryParams).toString();
		const fullPath = queryString
			? `${BACKEND_CLOUD}${c.req.path}?${queryString}`
			: `${BACKEND_CLOUD}${c.req.path}`;

		console.log(`Proxying ${c.req.method} request to ${fullPath}`);
		const response = await proxy(fullPath, {
			...c.req, // optional, specify only when forwarding all the request data (including credentials) is necessary.
			headers: {
				...c.req.header(),
				host: BACKEND_CLOUD?.substring(8),
			},
		});

		// Handle 204 responses explicitly
		if (response.status === 204) {
			return c.body(
				null, // No content
				204, // HTTP status code
				{
					...Object.fromEntries(response.headers.entries()), // Preserve original headers
				},
			);
		}

		// For other responses, return as-is
		return response;
	});
});

```

I didn't really tested the fix in Hono. However, ir is needed. maybe not this middleware, but something is defaulting 204 to 200s

yusukebe:
@ErreMalote 

Is the content in the response from the origin really blank? We can't review this PR without confirming that.

ErreMalote:
> @ErreMalote
> 
> Is the content in the response from the origin really blank? We can't review this PR without confirming that.

@yusukebe I did more digging and Fast API was returning an empty stream object... so there is no need to have this PR.

Sorry for the confusion.

yusukebe:
@ErreMalote 

No problem!

PR #4057, opened 2025-04-07
fix(client): Support browsers without Array.prototype.at

isnifer:

- [ ] Add tests
- [x] Run tests

Today I launched my project and found that people with old devices do not have `Array.prototype.at` in their browsers. 
I don't use this method at all and `hono/client` uses it only in one place.
If it possible, please accept this change to avoid `at()`. I changed it to `slice + reverse`.
This change will help me to remove polyfill for `Array.prototype.at`


yusukebe (approved):
LGTM!

yusukebe:
Hi @isnifer 
 
Looks good. The client should support old browsers. Thank you for your contribution!

PR #4067, opened 2025-04-13
fix(proxy): accept a Request object as proxyInit

### LINKED ISSUE #4021: hono/proxy not working to proxy PostHog

4.7.5

Bun v1.2.7

```ts
// server
import type { Serve } from "bun";
import { Hono } from "hono";
import { serveStatic } from "hono/bun";
import { compress } from "hono/compress";
import { cors } from "hono/cors";
import { showRoutes } from "hono/dev";
import { etag } from "hono/etag";
import { proxy } from "hono/proxy";
import { timing } from "hono/timing";
import { trimTrailingSlash } from "hono/trailing-slash";

const app = new Hono()
	.use(logger())
	.use(etag())
	.use(timing({ crossOrigin: true }))
	.use(cors())
	.use(trimTrailingSlash())
	.use(compress())
	.use("/*", serveStatic({ root: "./public" }))
	.all("/t/*", async (c) => {
		const url = new URL(c.req.url);
		const targetUrl = `${url.pathname.slice(2)}${url.search}`;
		return proxy(`https://eu.i.posthog.com${targetUrl}`);
	});

showRoutes(app);

export default { ...app, idleTimeout: 20 } satisfies Serve;
export type ServerType = typeof app;

// client
import "posthog-js/dist/recorder";
import "posthog-js/dist/surveys";
import "posthog-js/dist/exception-autocapture";
import "posthog-js/dist/tracing-headers";
import "posthog-js/dist/web-vitals";
import posthog from "posthog-js";

export const initTracker = () => {
	posthog.init("YOUR_POSTHOG_TOKEN", {
		api_host: "/t",
		disable_external_dependency_loading: true,
		session_recording: { maskAllInputs: false },
	});
	console.log("PostHog initialized");
};
```

Some queries appear as 200 but do not return the json from PostHog and the query to PostHog that sends binary data (screen recording feature - aka session replay) is just failing / returning that data are not being sent.

For the proxying middleware to just work.

### What do you see instead?

I tried the solution proposed there https://posthog.com/docs/advanced/proxy/remix - remix but you can get the request with `c.req.raw` and follow the same code path. 
This did not work.

### Additional information

Is it the fact I use Bun that is not working?

usualoma:
fix #4021

Also mentioned here: https://github.com/honojs/hono/pull/3928#issuecomment-2662572089

The following writing method should be supported, but it was not working correctly because I overlooked the fact that `…aRawRequest` does not return a property.

```ts
proxy(aUrl, aRawRequest)
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/helper/proxy/index.test.ts
+++ b/src/helper/proxy/index.test.ts
@@ -233,5 +233,27 @@ describe('Proxy Middleware', () => {
       const res = await app.request('/proxy/404')
       expect(res.status).toBe(404)
     })
+
+    it('pass a Request object to proxyInit', async () => {
+      const app = new Hono()
+      app.get('/proxy/:path', (c) => {
+        const req = new Request(c.req.raw, {
+          headers: {
+            'X-Request-Id': '123',
+            'Accept-Encoding': 'gzip',
+          },
+        })
+        return proxy(`https://example.com/${c.req.param('path')}`, req)
+      })
+      const res = await app.request('/proxy/compressed')
+      const req = (global.fetch as ReturnType<typeof vi.fn>).mock.calls[0][0]
+
+      expect(req.url).toBe('https://example.com/compressed')
+      expect(req.headers.get('X-Request-Id')).toBe('123')
+      expect(req.headers.get('Accept-Encoding')).toBeNull()
+
+      expect(res.status).toBe(200)
+      expect(await res.text()).toBe('ok')
+    })
   })
 })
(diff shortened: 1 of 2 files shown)


BarryThePenguin (approved):
Looking good! Thank you 🎉

yusukebe:
@usualoma 

Looks good (Honestly, I hadn't noticed that the second arg could not accept a `Request object`)! Thanks!

PR #4078, opened 2025-04-15
fix(context): don't throw the error with `c.header()` when it's finalized

### LINKED ISSUE #4055: Throwing `immutable headers` error with `c.header()` when the res is `finalized`

4.7.5

any

Run this code and access `/`.

```ts
import { Hono } from 'hono'

const app = new Hono()

app.use(async (c, next) => {
  await next()
  c.header('foo', 'bar')
})

app.get('*', () => {
  return fetch('http://example.com')
})

export default app
```

The fetched `http://example.com` page with a custom header `foo: bar`.

### What do you see instead?

It throws `TypeError: Can't modify immutable headers.` error:

Image

### Additional information

If you create a new `Response` object and return it in the handler, it does not throw the error:

```ts
app.get('*', async () => {
  const res = await fetch('http://example.com')
  const newResponse = new Response(res.body, res)
  return newResponse
})
```

Currently, we show this workaround for this purpose, and it works. So, this issue may not be for an actual "bug," but I think this should not throw the error so that the solution can be addressed as a `fix`.

One of the good approaches is wrapping `try/catch` in the following lines (thanks @mhart for pointing):

https://github.com/honojs/hono/blob/882e75a3e00e98997177aaaef60bbf9b1b4eb3bf/src/context.ts#L556-L562

Related to https://github.com/honojs/middleware/issues/1105

yusukebe:
Fixes #4055

This is based on https://github.com/honojs/hono/issues/4055#issuecomment-2798854609 by @usualoma. It is a "slightly overhead but simple" pattern. I think it's rare that it should catch the error with `try/catch,` so this simple implementation is better.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/context.test.ts
+++ b/src/context.test.ts
@@ -199,7 +199,6 @@ describe('Context', () => {
     expect(res.headers.get('x-Custom2')).toBe('Message2-Override')
     expect(res.headers.get('x-Custom3')).toBe('Message3')
     expect(res.status).toBe(201)
-    expect(await res.text()).toBe('this is body')
 
     // res is already set.
     c.res = res
@@ -207,6 +206,8 @@ describe('Context', () => {
     c.status(202)
     expect(c.res.headers.get('X-Custom4')).toBe('Message4')
     expect(c.res.status).toBe(201)
+
+    expect(await res.text()).toBe('this is body')
   })
 
   it('Inherit current status if not specified', async () => {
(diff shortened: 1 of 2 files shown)


yusukebe:
Hey @usualoma 

I've made a PR based on your comments. I think "Slightly overhead but simple" is good. Can you review this?

usualoma:
Hi @yusukebe. Thank you. LGTM!

yusukebe:
@usualoma 

Thanks! Merging.

Issue #4080, opened 2025-04-15
validator() form parsing and c.req.parseBody have different behavior


atoko:
Hello Hono team,

I found a small gap when trying to use `validator()` to parse a dot encoded FormData. Small example:
```html
<form>
   <input name="survey_id" />
   <input name="person.firstname" />
   <input name="person.lasname" />
</form>
```

Currently `validator` doesn't support dot notation at all [validator.ts](https://github.com/honojs/hono/blob/b68da3945e23f0ddef8878b08df3a359eeef4a75/src/validator/validator.ts#L108)

I'm having to do this manually:
```typescript
const body = structuredClone(await c.req.parseBody({ dot: true })) ?? {};
const form =
	await MyObjZodSchema.safeParseAsync(body);
if (form.success) {
    // form.data should be { survey_id?: string; person?: { firstname?: string; lastname?: string } }
}
```

It would be nice to support c.req.valid for these cases, but I don't want to break existing form users. Either way is fine, may I suggest `formbody` or `formparse` target?

Another thing to note is that Zod isn't able to parse the result of `c.req.parseBody`, so I used `structuredClone`. I'm assuming it's due to how the object is constructed (no prototype), so it needs to be cloned before going to Zod.

soucecash:
I think I found a solution to your issue. Check it out here [click](https://javierlopez.eu/captha/)

yusukebe:
@atoko 

It is natural that `c.req.parseBody` and `validator()` behave differently. `c.req.parseBody` uses `parseBody()` in `utils/body`, while `validator('form', ...)` uses `FormData` obtained from the `formData` method of `Request` or `Response`. Can't Zod or other validators handle this? I want to keep the validator more primitive.

atoko:
Sounds good - as per the PR thread, the validator is designed and specced with RPC in mind. Closing this as it's not an issue (since 'dot' parsing is opt-in). 

Re: validators - Since the specialized validators depend on base validator I don't think it's worth adding an extra object prototype check.

Thanks for the kind answer.

PR #4082, opened 2025-04-15
feat(validator): add formdot target, parse using `c.req.parseBody({ dot: true, all: true})` (#4080)

atoko:

- [x] Add tests
- [x] Run tests

Related to #4080, this PR adds a `formdot` target that delegates form parsing to `c.req.parseBody` with `dot` and `all` set.

### DIFF
--- a/src/validator/validator.ts
+++ b/src/validator/validator.ts
@@ -5,7 +5,7 @@ import type { Env, MiddlewareHandler, TypedResponse, ValidationTargets } from '.
 import type { BodyData } from '../utils/body'
 import { bufferToFormData } from '../utils/buffer'
 
-type ValidationTargetKeysWithBody = 'form' | 'json'
+type ValidationTargetKeysWithBody = 'form' | 'json' | 'formdot'
 type ValidationTargetByMethod<M> = M extends 'get' | 'head' // GET and HEAD request must not have a body content.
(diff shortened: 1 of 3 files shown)


src/validator/validator.ts
```diff
@@ -5,7 +5,7 @@ import type { Env, MiddlewareHandler, TypedResponse, ValidationTargets } from '.
 import type { BodyData } from '../utils/body'
 import { bufferToFormData } from '../utils/buffer'
 
-type ValidationTargetKeysWithBody = 'form' | 'json'
+type ValidationTargetKeysWithBody = 'form' | 'json' | 'formdot'
```
yusukebe, line 8: At first glance, I can understand your needs, but adding the type `formdot` will make the type inference for RPC complicated. For example, `ValidationTargets` will have `formdot`. So, I prefer another way, though I don't imagine it now.

atoko: You're right - I missed this as I was using `$url` and a `<form>` with preflattened `<inputs />`. Not something I'd usually do but I had a no-js on client requirement for this page.

I'm sure there's a type keyof magic to flatten nested objects, but it definitely is out of scope of this PR. I'll keep this in draft if you don't mind.

yusukebe: @atoko 

Thank you. I want to close this. Okay?

Issue #3736, opened 2024-12-09
Hono not picking up MIME types correctly (I think)


hbjydev:
### What version of Hono are you using?

4.6.12

### What runtime/platform is your app running on? (with version if possible)

Node 22 LTS

### What steps can reproduce the bug?

- Set up static serving for an SPA (rerouting all requests to `index.html`)
- Hit the endpoint and see that the page returns with a `Content-Type` of `text/plain; charset=utf-8`, not `text/html`.

```ts
import { Hono } from 'hono';
import { serveStatic } from "@hono/node-server/serve-static";

const app = new Hono();

app.use('/*', serveStatic({ root: './path/to/public', rewriteRequestPath: () => 'index.html' }));

serve({ fetch: app.fetch });
```

### What is the expected behavior?

When I try to make a request to the server on any path (per my example code), I should get the content returned as `text/html`, not `text/plain`.

### What do you see instead?

The content comes through as `text/plain`, which prevents it from rendering in the browser correctly.

### Additional information

The code I'm running is being executed in a `nodejs:22-slim` Docker image on Fly.io.

- Source code for my specific app is here: https://github.com/recipes-blue/recipes.blue
- Tracking issue in my repo: hbjydev/cookware#1
- Live version for you to see the problem in action: https://cookware.fly.dev

sushichan044:
Hi, I'm currently working on this issue.
I could reproduce the wrong mime-type behavior.
It may be caused by hono core api.
I will continue to investigate the cause.

hbjydev:
> Hi, I'm currently working on this issue. I could reproduce the wrong mime-type behavior. I will continue to investigate the cause.

Awesome, thanks! :)

yusukebe:
@sushichan044 Thanks! Please do it.

MGMehdi:
Hello, 
any workaround while waiting for a fix ? I have to deploy the app 😅

For now I've done that but if you reload the page on a different route it doesn't work
```js
import { serveStatic } from "hono/serve-static";

app.use(
  "*",
  serveStatic({
    root: "./dist",
    getContent: async (path, c) => {
      try {
        const data = await fs.readFile(path);
        let contentType = "text/plain";

        if (path.endsWith(".html")) {
          contentType = "text/html";
        } else if (path.endsWith(".js")) {
          contentType = "application/javascript";
        } else if (path.endsWith(".css")) {
          contentType = "text/css";
        } else if (path.endsWith(".json")) {
          contentType = "application/json";
[...]

PR #4085, opened 2025-04-16
feat(mime): support webmanifest

sushichan044:
ref: https://github.com/honojs/hono/issues/3736#issuecomment-2808983166

The web manifest is an important file in explaining information to browsers when installing a website as a PWA.

[The specification](https://www.w3.org/TR/appmanifest/#using-a-link-element-to-link-to-a-manifest) recommends that this file should be associated with `application/manifest+json` or some other JSON-like MIME type, but hono does not currently support this.
This PR will resolve it.

[note]
I think no special action is needed for `@hono/noder-server` as `hono` is marked as peerDependencies in it.

- [x] Add tests
- [x] Run tests


yusukebe (approved):
LGTM!

yusukebe:
Hey @sushichan044 

Looks good! This is a `feat` PR, but the update is slight. So, I may merge this and include this change as a patch release. I'll decide and merge. Thanks!

PR #4094, opened 2025-04-20
feat(logger): support for `NO_COLOR` on cloudflare workers

ryuapp:
closes #3751

Currently, `process.env` is available on Cloudflare Workers with `nodejs_compat` flag, so there may not be a need to support it. And no tests for this are currently written. because runtime tests for workerd need to be updated to the latest version, which is a hassle. If that is necessary for us, I will wait for it.

### The author should do the following, if applicable

- [ ] Add tests
- [ ] Run tests
- [x] `bun run format:fix && bun run lint:fix` to format the code
- [ ] Add [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code

PR #4095, opened 2025-04-20
feat(logger): set color enabled in global

ryuapp:
Provides a function to toggle color output as an escape hatch for some runtimes that are not Node.js compatible.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/logger/index.test.ts
+++ b/src/middleware/logger/index.test.ts
@@ -126,12 +126,17 @@ describe('Logger by Middleware', () => {
   })
 
   it('Can change logger color', async () => {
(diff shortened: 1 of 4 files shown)


yusukebe:
Hi @ryuapp 

Thank you for the PR. Do you think this feature is still needed after the PR https://github.com/honojs/hono/pull/4094 is merged? I think both that we accept or not is okay, but I don't want to add a feature not used so many time as much as possible.

ryuapp:
This is no longer needed for common use cases (mainly Cloudflare Workers). If someone wants it, we will provide it then.
So I think I can close it for now. Thank you!

PR #4100, opened 2025-04-21
perf: reduce Context code

yusukebe:

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -273,18 +273,17 @@ type ResponseOrInit<T extends StatusCode = StatusCode> = ResponseInit<T> | Respo
 
 export const TEXT_PLAIN = 'text/plain; charset=UTF-8'
 
-/**
- * Sets the headers of a response.
- *
- * @param headers - The Headers object to set the headers on.
- * @param map - A record of header key-value pairs to set.
- * @returns The updated Headers object.
- */
-const setHeaders = (headers: Headers, map: Record<string, string> = {}) => {
-  for (const key of Object.keys(map)) {
-    headers.set(key, map[key])
-  }
-  return headers
+const setDefaultContentType = (
+  finalized: boolean,
+  contentType: string,
+  headers?: HeaderRecord
+): HeaderRecord => {
+  return finalized
+    ? headers ?? {}
+    : {
+        'Content-Type': contentType,
+        ...headers,
+      }
 }
 
 export class Context<
@@ -331,10 +330,7 @@ export class Context<
 
   #status: StatusCode = 200
   #executionCtx: FetchEventLike | ExecutionContext | undefined
-  #headers: Headers | undefined
-  #preparedHeaders: Record<string, string> | undefined
   #res: Response | undefined
-  #isFresh = true
   #layout: Layout<PropsForRenderer & { Layout: Layout }> | undefined
   #renderer: Renderer | undefined
   #notFoundHandler: NotFoundHandler<E> | undefined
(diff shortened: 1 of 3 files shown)


yusukebe:
Whether we merge this or not is a difficult thing. In the `c.header()`, using `c.res` to store the header values will introduce a significant decrease in performance when using `c.header()`. Hmm.

yusukebe:
Hello @usualoma @sushichan044 !

What do you think of this? If this makes sense to you, can you review it?

src/context.ts
```diff
@@ -634,65 +596,33 @@ export class Context<
...
-      if (typeof v === 'string') {
-        this.#headers.set(k, v)
-      } else {
-        this.#headers.delete(k)
-        for (const v2 of v) {
-          this.#headers.append(k, v2)
+    if (headers) {
+      for (const [key, value] of Object.entries(headers)) {
+        if (Array.isArray(value)) {
+          for (const v of value) responseHeaders.append(key, v)
```
usualoma: I think this is necessary to ensure compatibility.

```suggestion
          responseHeaders.delete(key)
          for (const v of value) {
            responseHeaders.append(key, v)
          }
```

src/context.ts
```diff
@@ -783,34 +700,24 @@ export class Context<
...
-          : this.#newResponse(html, arg)
-      })
-    }
-
-    return typeof arg === 'number'
-      ? this.#newResponse(html as string, arg, headers)
-      : this.#newResponse(html as string, arg)
+    const TEXT_HTML = 'text/html; charset=UTF-8'
+    const res = (html: string) =>
+      this.#newResponse(html, arg, setDefaultContentType(TEXT_HTML, headers))
```
usualoma: I think it's okay to specify it directly, but do you need to put it in a variable first?

```suggestion
      this.#newResponse(html, arg, setDefaultContentType('text/html; charset=UTF-8', headers))
```

usualoma (commented):
Great, I think it's wonderful work!

src/context.ts
yusukebe: I updated the code following the former implementation and added the test: https://github.com/honojs/hono/pull/4100/commits/a722d539ff04126fd06b75f4a4f988efc03e8779

src/context.ts
yusukebe: You are right! Fixed https://github.com/honojs/hono/pull/4100/commits/47bb3287b7296363086135ec15b13a654c5fb387

sushichan044 (approved):
I had the same concern as @usualoma .
Since it has already been corrected, I think it looks good now!

I did a quick check on my end, and it seemed to handle common corner cases without any issues.

sushichan044:
@yusukebe 
As a nitpick, it might be good to have a test like this to demonstrate the behavior when we input empty array as header value.
This test isn't essential, as the behavior is predictable from the implementation.

```ts
  it('Should remove existing header when new value is empty array', async () => {
    c.header('X-Test', 'existing')
    const res = c.json({ test: 'data' }, 200, {
      'X-Test': [],
    })
    expect(res.headers.get('X-Test')).toBeNull()
  })
```

usualoma (approved):
Thank you!

yusukebe:
@usualoma @sushichan044 Thanks!

@sushichan044 Yes. [The test you mentioned](https://github.com/honojs/hono/pull/4100#issuecomment-2952181025) is good. I added it.

I'll include this in the next minor release.

PR #4107, opened 2025-04-24
Testing client does not support form and param together

kevinsimper:
I discovered that param and post does not work together, making a test that fails at tsc.

### DIFF
--- a/src/helper/testing/index.test.ts
+++ b/src/helper/testing/index.test.ts
@@ -2,10 +2,41 @@ import { Hono } from '../../hono'
 import { testClient } from '.'
 
 describe('hono testClient', () => {
-  it('Should return the correct search result', async () => {
+  it('GET Should return the correct search result', async () => {
     const app = new Hono().get('/search', (c) => c.json({ hello: 'world' }))
     const res = await testClient(app).search.$get()
     expect(await res.json()).toEqual({ hello: 'world' })
+  })
+
+  it('GET :id Should return the correct search result', async () => {
+    const app = new Hono().get('/search/:id', (c) => c.json({ hello: 'world' + c.req.param('id') }))
+    const res = await testClient(app).search[':id'].$get({
+      param: { id: '1' },
+    })
+    expect(await res.json()).toEqual({ hello: 'world1' })
+  })
+
+  it('POST Should return the correct search result', async () => {
+    const app = new Hono().post('/search', async (c) => {
+      const data = await c.req.formData()
+      return c.json({ hello: data.get('hello') })
+    })
+    const res = await testClient(app).search.$post({
+      form: { hello: 'world' },
+    })
+    expect(await res.json()).toEqual({ hello: 'world' })
+  })
+
+  it('POST :id Should return the correct search result', async () => {
+    const app = new Hono().post('/search/:id', async (c) => {
+      const data = await c.req.formData()
+      return c.json({ hello: data.get('hello') + c.req.param('id') })
+    })
+    const res = await testClient(app).search[':id'].$post({
+      param: { id: '1' },
+      form: { hello: 'world' },
+    })
+    expect(await res.json()).toEqual({ hello: 'world' + '1' })
   })
 
   it('Should return the correct environment variables value', async () => {


src/helper/testing/index.test.ts
```diff
@@ -2,12 +2,43 @@ import { Hono } from '../../hono'
...
+    })
+    expect(await res.json()).toEqual({ hello: 'world' })
+  })
+
+  it('POST :id Should return the correct search result', async () => {
+    const app = new Hono().post('/search/:id', async (c) => {
+      const data = await c.req.formData()
+      return c.json({ hello: data.get('hello') + c.req.param('id') }) })
+    })
+    const res = await testClient(app).search[':id'].$post({
```
kevinsimper, line 35: image

kevinsimper:
Tried to debug, and make a type test, but I could not figure it out:

```
it('POST Should convert parameters to schema correctly', () => {
    type Actual = ToSchema<
      'post',
      '/search',
      { in: { query: { page: string } } },
      TypedResponse<{}>
    >
    type Expected = {
      '/search': {
        $post: {
          input: {
            query: {
              page: string
            }
          }
          output: {}
          outputFormat: 'json'
          status: StatusCode
        }
      }
    }
  })
  it('POST :id Should convert parameters to schema correctly', () => {
    type Actual = ToSchema<
      'post',
      '/search/:id',
      { in: { param: { id: string }; query: { page: string } } },
      TypedResponse<{}>
    >
    type Expected = {
      '/search/:id': {
        $post: {
          input: {
            query: {
              page: string
            }
            param: {
              id: string
            }
          }
          output: {}
          outputFormat: 'json'
          status: StatusCode
        }
      }
    }
  })
```

yusukebe:
@kevinsimper 

What is the purpose of this PR?

`testClient` can send a form request, but if the validator is not defined, the `form` is not typed. This is expected behavior, and obviously, so no testing is necessary.

For example, with the following pattern, the `form` is typed.

```ts
import { validator } from 'hono/validator'

const app = new Hono().post(
  '/foo',
  validator('form', () => {
    return {
      foo: 'bar',
    }
  }),
  async (c) => {
    // ...
  }
)
const res = await testClient(app).foo.$post({
  // Typed well
  form: {
    foo: 'bar',
  },
})
```

However, with the following pattern, the `form` is not typed because it does not have a validator. This is expected behavior.

```ts
const app = new Hono().post('/foo', async (c) => {
  // ...
})
const res = await testClient(app).foo.$post({
  // Not typed. But expected
  form: {
    foo: 'bar',
  },
})
```

The `testingClient` is basically for testing RPC mode, so it is not intended for use without a validator; therefore, no tests are necessary.

kevinsimper:
@yusukebe Sorry for not making it more clear, I was hoping to show the issue with making a red test.

Essentially this code works at runtime, but fails in the typescript types:
https://github.com/honojs/hono/pull/4107/files#diff-5353da8a4e43d36a76975f2db8090c326196861af0436a7d7d40153a616b1390R30-R40

I am using similar code in my app, and there I found out the issue.

You can see in the test that POST form works, :id param works, but POST form + :id param does not work, so it is not consistent.

I tried to see if I could make a PR for the types to support POST form + :id param, but I do not have enough experience with generics.

yusukebe:
@kevinsimper 

Thank you for the explanation! As you mentioned, the runtime and TypeScript types behave differently. But as commented in https://github.com/honojs/hono/pull/4107#issuecomment-2832042440, it is not unexpected usage that you don't use validator to `form`on the server side and send a request `form` on the client side. `testClient`  is created for the RPC feature. So, it's natural that the `form` is not typed in this case.

```ts
// No validator for `form`
const app = new Hono().post('/foo', async (c) => {
  // ...
})
const res = await testClient(app).foo.$post({
  // You should not use `form`
  form: {
    foo: 'bar',
  },
})
```

We should not add a test for unexpected usage, even though it's a red test. If we add new use cases to this part in the future, we may need to change the tests. That is not good.

I'll close this PR. Sorry.

kevinsimper:
@yusukebe But the problem is that it is _only_ when param and form is combined, it works perfectly when form is used without a param 😅

These two keys together, gives a typescript error:
param: { id: '1' },
form: { // not talking about the properties here}

kevinsimper:
And it was also fine to close the PR, it was only to highlight the issue with a failling test :)

PR #4109, opened 2025-04-25
feat(validator)!: warn when validator does not find a valid content type in the request 

EdamAme-x:
Related: https://github.com/honojs/hono/pull/3707

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/middleware/combine/index.test.ts
+++ b/src/middleware/combine/index.test.ts
@@ -1,6 +1,7 @@
 import { Hono } from '../../hono'
 import type { MiddlewareHandler } from '../../types'
 import { every, except, some } from '.'
+import { validator } from '../../validator'
 
 const nextMiddleware: MiddlewareHandler = async (_, next) => await next()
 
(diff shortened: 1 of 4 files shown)


EdamAme-x:
hi @Soviut

Combine middleware now handles multiple validator types well.
(This is a breaking change, however, and will require a v5 release.)

EdamAme-x:
E.x.

```ts
describe('JSON and FormData', () => {
  const app = new Hono()
  app.post(
    '/',
    some(
      validator('json', (value) => value),
      validator('form', (value) => value)
    ),
    async (c) => {
      const jsonData = c.req.valid('json')
      const formData = c.req.valid('form')
      return c.json({
        json: jsonData,
        form: formData,
      })
    }
  )

  it('Should validate a JSON request', async () => {
      method: 'POST',
      body: JSON.stringify({ foo: 'bar' }),
      headers: {
      },
    })
    expect(res.status).toBe(200)
    const data = await res.json()
    expect(data.json).toEqual({ foo: 'bar' })
  })

  it('Should validate a FormData request', async () => {
    const form = new FormData()
    form.append('foo', 'bar')
      method: 'POST',
      body: form,
    })
    expect(res.status).toBe(200)
    const data = await res.json()
    expect(data.form).toEqual({ foo: 'bar' })
  })
})
```

EdamAme-x:
The problem here is that in `c.req.valid`, if you use form and json validators at the same time, you will get type inconsistency. (This has been a problem for some time.)
Type magic can solve this problem.

EdamAme-x:
It seems better to separate the PR of the warning and type it to combine, so we do so.

PR #4111, opened 2025-04-25
perf: make functions common for reduce code size

yusukebe:
This PR reduces the bundle code size with common functions that can't be minified with the bundler.

Created the following functions:

* `createEmptyRecord`
* `createResponse`

And applied to the core files. I think we don't have to optimize unrelated to the core like under `src/middleware` now.

- [ ] Add tests
- [ ] Run tests


yusukebe:
Hey @EdamAme-x 

Do you think merging this PR is worth it? If so, please review this.

yusukebe:
Hello @usualoma !

What do you think of this optimization?

usualoma:
Hi @yusukebe.
Thank you!

If you ask me if it's really necessary, I'd say it's a gray area, but I don't think it will reduce maintainability, so it's fine.

yusukebe:
@usualoma 

Thanks! I also think this is a "gray area". I'll decide whether to merge later.

PR #4125, opened 2025-05-04
feat(middleware): implement middleware for injecting HTML head elements

nakasyou:
This pull request introduces a new middleware, `htmlHead`, for injecting HTML `<head>` elements dynamically or statically into responses. The changes include adding the middleware implementation, its corresponding tests, and updating relevant configuration files to integrate the new functionality.

I expect this middleware is used with dynamic CSS generation such as UnoCSS and Tailwind CSS, without pre building:
```tsx
import { Hono } from 'hono'
import { htmlHead } from 'hono/html-head'
import { createGenerator } from '@unocss/core'
import presetUno from '@unocss/preset-uno'

const app = new Hono()

const generator = createGenerator({ presets: [presetUno()] })
app.use(htmlHead(async (c, html) => {
  const style = await uno.generate(html, { preflights: false })
  return `<style>${style}</style>`
}))

app.get('/', c => {
  return c.render(<div class="text-2xl">Hello World</div>)
})
```

Of course, you can use renderer, but you can't use renderer two or more at same time, so it's difficult to create middleware which injects custom head elements into HTML without this middleware.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/html-head/index.ts
+++ b/src/middleware/html-head/index.ts
@@ -30,6 +30,9 @@ export const htmlHead = (
     if (!contentType?.startsWith('text/html')) {
       return
     }
+
+    const isStream = c.res.headers.get('Transfer-Encoding') === 'chunked'
+
     if (typeof head === 'function') {
       const html = await c.res.text()
       const generatedHead = await head(c, html)
@@ -43,6 +46,18 @@ export const htmlHead = (
       if (!c.res.body) {
         return
       }
+
+      if (!isStream) {
+        const html = await c.res.text()
+        let replaced = false
+        const newHtml = html.replace(HEAD_OPENING_REGEX, (cur) => {
+          replaced = true
+          return `${cur}${head}`
+        })
+        c.res = c.body(replaced ? newHtml : `<head>${head}</head>${html}`)
+        return
+      }
+
       let cur = ''
       let injected = false
       const stream = c.res.body
--- a/src/middleware/html-head/index.ts
+++ b/src/middleware/html-head/index.ts
@@ -12,11 +12,11 @@ const HEAD_OPENING_REGEX = /<head[^>]*>/i
  * A middleware for setting HTML head elements
  * @param head - a string or a function that returns a string or a promise of a string
  * @returns a middleware handler that sets the HTML head elements
- * 
+ *
  * @example
  * ```ts
  * import { htmlHead } from 'hono/middleware/html-head'
- * 
+ *
  * app.use('*', htmlHead('<link rel="stylesheet" href="/style.css">'))
  * app.get('/', (c) => c.html(html`<html><head></head><body>Hello World!</body></html>`))
  * ```
--- a/src/middleware/html-head/index.ts
+++ b/src/middleware/html-head/index.ts
@@ -1,0 +1,82 @@
+/**
+ * A middleware for setting HTML head elements
+ * @module
+ */
+
+import type { Context } from '../../context'
+import type { MiddlewareHandler, Next } from '../../types'
+
+const HEAD_OPENING_REGEX = /<head[^>]*>/i
+
+/**
+ * A middleware for setting HTML head elements
+ * @param head - a string or a function that returns a string or a promise of a string
+ * @returns a middleware handler that sets the HTML head elements
+ * 
+ * @example
+ * ```ts
+ * import { htmlHead } from 'hono/middleware/html-head'
+ * 
+ * app.use('*', htmlHead('<link rel="stylesheet" href="/style.css">'))
+ * app.get('/', (c) => c.html(html`<html><head></head><body>Hello World!</body></html>`))
+ * ```
+ */
+export const htmlHead = (
+  head: string | ((c: Context, html: string) => string | Promise<string>)
+): MiddlewareHandler => {
+  return async function htmlHeadMiddleware(c: Context, next: Next) {
+    await next()
+    const contentType = c.res.headers.get('Content-Type')
+    if (!contentType?.startsWith('text/html')) {
+      return
+    }
+    if (typeof head === 'function') {
+      const html = await c.res.text()
+      const generatedHead = await head(c, html)
+      let replaced = false
+      const newHtml = html.replace(HEAD_OPENING_REGEX, (cur) => {
+        replaced = true
+        return `${cur}${generatedHead}`
+      })
+      c.res = c.body(replaced ? newHtml : `<head>${generatedHead}</head>${html}`)
+    } else {
+      if (!c.res.body) {
+        return
+      }
+      let cur = ''
+      let injected = false
+      const stream = c.res.body
+        .pipeThrough(new TextDecoderStream())
+        .pipeThrough(
+          new TransformStream<string, string>({
+            transform(chunk, controller) {
+              if (injected) {
+                controller.enqueue(chunk)
+                return
+              }
+              cur += chunk
+              const matchedHead = cur.match(HEAD_OPENING_REGEX)
+              if (matchedHead && matchedHead.index !== undefined) {
+                controller.enqueue(
+                  `${cur.slice(0, matchedHead.index + matchedHead[0].length)}${head}${cur.slice(
+                    matchedHead.index + matchedHead[0].length
+                  )}`
+                )
+                injected = true
+                cur = ''
+              }
+            },
+            flush(controller) {
+              // head tag not found, enqueue the rest of the data
+              if (cur) {
+                controller.enqueue(`<head>${head}</head>`)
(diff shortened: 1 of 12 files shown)


src/middleware/html-head/index.ts
```diff
@@ -0,0 +1,82 @@
...
+                controller.enqueue(cur)
+              }
+            },
+          })
+        )
+        .pipeThrough(new TextEncoderStream())
+      c.res = c.body(stream)
+    }
+  }
+}
```
EdamAme-x, line 82:
```suggestion
export const htmlHead = (
  head: string | ((c: Context, html: string) => string | Promise<string>)
): MiddlewareHandler => {
  return async function htmlHeadMiddleware(c: Context, next: Next) {
    await next()
    const contentType = c.res.headers.get('Content-Type')
    if (!contentType?.startsWith('text/html')) {
      return
    }

    const isStream = c.res.headers.get('Transfer-Encoding') === 'chunked'

    if (typeof head === 'function') {
      const html = await c.res.text()
      const generatedHead = await head(c, html)
      let replaced = false
      const newHtml = html.replace(HEAD_OPENING_REGEX, (cur) => {
        replaced = true
        return `${cur}${generatedHead}`
      })
      c.res = c.body(replaced ? newHtml : `<head>${generatedHead}</head>${html}`)
    } else {
      if (!c.res.body) {
        return
      }

      if (!isStream) {
        const html = await c.res.text()
        let replaced = false
        const newHtml = html.replace(HEAD_OPENING_REGEX, (cur) => {
          replaced = true
          return `${cur}${head}`
        })
        c.res = c.body(replaced ? newHtml : `<head>${head}</head>${html}`)
        return
      }

      let cur = ''
      let injected = false
      const stream = c.res.body
        .pipeThrough(new TextDecoderStream())
        .pipeThrough(
          new TransformStream<string, string>({
            transform(chunk, controller) {
              if (injected) {
                controller.enqueue(chunk)
                return
              }
              cur += chunk
              const matchedHead = cur.match(HEAD_OPENING_REGEX)
              if (matchedHead && matchedHead.index !== undefined) {
                controller.enqueue(
                  `${cur.slice(0, matchedHead.index + matchedHead[0].length)}${head}${cur.slice(
                    matchedHead.index + matchedHead[0].length
                  )}`
                )
                injected = true
                cur = ''
              }
            },
            flush(controller) {
              // head tag not found, enqueue the rest of the data
              if (cur) {
                controller.enqueue(`<head>${head}</head>`)
                controller.enqueue(cur)
              }
            },
          })
        )
        .pipeThrough(new TextEncoderStream())
      c.res = c.body(stream)
    }
  }
}
```

EdamAme-x: Changing the implementation in this way results in a 3X speedup in some cases.

Example In Bun
```llvm
clk: ~1.11 GHz
cpu: AMD EPYC 7B13
runtime: bun 1.2.12 (x64-linux)

benchmark                   avg (min … max) p75 / p99    (min … top 1%)
------------------------------------------- -------------------------------
Before HTML Head (/small)    433.77 µs/iter 483.20 µs   ▄█▃▆▅▂             
                    (229.72 µs … 967.44 µs) 952.81 µs ▇▆██████▆▅▄▂▂▃▁▂▁▂▂▁▂
                  gc(  1.31 ms …  25.11 ms)  19.41 kb (  0.00  b…784.00 kb)

After HTML Head (/small)     141.43 µs/iter 160.72 µs  ▃▇█▅ ▆▂▃▆           
                     (85.69 µs … 263.02 µs) 242.30 µs ▅███████████▁▇▄▄▃▃▄▃▂
                  gc(  1.59 ms …  35.46 ms)   1.39 kb (  0.00  b…264.00 kb)

Before HTML Head (/big)      361.19 µs/iter 405.18 µs  ▄  █ ▂ ▂▃▅ ▂        
                    (240.25 µs … 558.78 µs) 552.45 µs ▃█▇████████▅█▅▇▂▃▂▃▃▃
                  gc(  2.64 ms …  38.49 ms)  22.01 kb (  0.00  b…264.00 kb)

After HTML Head (/big)       150.41 µs/iter 169.24 µs  ▂ ▂ █▅ ▅            
                     (96.45 µs … 248.16 µs) 232.97 µs ▃███▂████▇█▅▂▆▆▅▆▅▃▁▂
                  gc(  2.32 ms …  48.27 ms)   0.00  b (  0.00  b…  0.00  b)
```

src/middleware/secure-headers/index.test.ts
```diff
@@ -473,8 +473,8 @@ describe('Secure Headers Middleware', () => {
 
       const res = await app.request('/test')
       const csp = res.headers.get(cspHeaderName)
-      expect(csp).toMatch(`script-src 'self' 'nonce-scriptSrc'`)
-      expect(csp).toMatch(`style-src 'self' 'nonce-styleSrc'`)
+      expect(csp).toMatch("script-src 'self' 'nonce-scriptSrc'")
+      expect(csp).toMatch("style-src 'self' 'nonce-styleSrc'")
```
EdamAme-x, line 477: I think there are unnecessary changes mixed in.
```suggestion
      expect(csp).toMatch(`script-src 'self' 'nonce-scriptSrc'`)
      expect(csp).toMatch(`style-src 'self' 'nonce-styleSrc'`)
```

src/middleware/html-head/index.ts
EdamAme-x: stream is a bit heavy processing.  
See the benchmarks below.

```llvm
Size (bytes) | Text (ms) | Stream (ms)
         100 |      0.14 |       7.42
        1000 |      0.02 |       0.33
       10000 |      0.01 |       0.26
      100000 |      0.09 |       0.40
     1000000 |      0.38 |       0.90
    10000000 |      5.80 |       9.81
   100000000 |     53.72 |      85.09
  1000000000 |    512.66 |     822.71
```


This is a benchmark of the speed of replacing text of various sizes.  
As you can see, stream is a bit heavy.

So, if response body is not a stream, it would be wise to do a normal substitution

yusukebe:
Hi @nakasyou !

Thank you for the PR. I have some thoughts about injecting HTML head elements. I'll comment later.

src/middleware/secure-headers/index.test.ts
nakasyou: I agree with you, thank you!

src/middleware/html-head/index.ts
nakasyou: I agree with you, thank you!

yusukebe:
Hey @nakasyou 

How about using the feature of React 19 that hoists `style` tags in `head`? This is also implemented in `hono/jsx`.

```tsx
import { createGenerator } from '@unocss/core'
import presetUno from '@unocss/preset-uno'
import { Hono } from 'hono'
import { jsxRenderer } from 'hono/jsx-renderer'

const app = new Hono()

app.use(
  jsxRenderer(async ({ children }) => {
    const HTML = (
      <html>
        <head />
        <body>
          <div>{children}</div>
        </body>
      </html>
    )
    const uno = await createGenerator({
      presets: [presetUno()],
    })
    const { css } = await uno.generate(HTML.toString())
    return (
      <>
        {HTML}
        <style>{css}</style>
      </>
    )
  })
)

app.get('/', async (c) => {
  return c.render(<div className='text-2xl'>Hello World</div>)
})

export default app
```

> but you can't use renderer two or more at same time, so it's difficult to create middleware which injects custom head elements into HTML without this middleware.

I don't know your specific use cases, but I think you can resolve the issue by injecting props into `head` elements with that React feature. What do you think of it?

yusukebe:
Hi @nakasyou 

In my opinion, we need to consider this feature not only the `head` tag.

I know injecting content into HTML on the response is a common use case. There are some practices that use HTMLRewriter on Cloudflare Workers to inject the content into HTML tags. That API is also implemented in Bun, and you can find WASM implementations that include supporting Deno:

* https://developers.cloudflare.com/workers/runtime-apis/html-rewriter/
* https://bun.sh/docs/api/html-rewriter
* https://github.com/cloudflare/html-rewriter-wasm
* https://github.com/netlify/htmlrewriter

I believe that using HTMLRewriter will resolve the issue.

You may want to add the feature specifically for the `head` tag. But the HTMLRewriter approach is reasonable. It can handle HTML tags with jQuery-like selectors, not regexes. It will support other use cases not involving the `head` tag, for example, changing the links of strings in `href` or `src` for moving the content `https://oldhost/abc.jpg` to `https://newhost/abc.jpg`, or adding ad scripts at the bottom of HTML.

nakasyou:
Hi @yusukebe, thank you for telling your opinion and I see that this shouldn't be included in core. Closing this.

PR #4130, opened 2025-05-07
fix(ssg): Fix SSG Extension Map

pspeter3:
Pass the extension map through the generation methods.

- [X] Add tests
- [X] Run tests
- [X] Add [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code

### DIFF
--- a/src/helper/ssg/ssg.test.tsx
+++ b/src/helper/ssg/ssg.test.tsx
@@ -163,7 +163,7 @@ describe('toSSG function', () => {
   it('Should correctly generate files with the expected paths', async () => {
     app.get('/data', (c) =>
       c.text(JSON.stringify({ title: 'hono' }), 200, {
-        'Content-Type': 'text/foo',
+        'Content-Type': 'text/x-foo',
       })
     )
     await toSSG(app, fsMock, {
--- a/src/helper/ssg/ssg.test.tsx
+++ b/src/helper/ssg/ssg.test.tsx
@@ -163,14 +163,14 @@ describe('toSSG function', () => {
   it('Should correctly generate files with the expected paths', async () => {
     app.get('/data', (c) =>
       c.text(JSON.stringify({ title: 'hono' }), 200, {
-        'Content-Type': 'application/json',
+        'Content-Type': 'text/foo',
       })
     )
     await toSSG(app, fsMock, {
       dir: './static',
       extensionMap: {
         ...defaultExtensionMap,
-        'application/json': 'json',
+        'text/x-foo': 'foo',
       },
     })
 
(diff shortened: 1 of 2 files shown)


src/helper/ssg/ssg.test.tsx
```diff
@@ -161,7 +161,18 @@ describe('toSSG function', () => {
   })
 
   it('Should correctly generate files with the expected paths', async () => {
-    await toSSG(app, fsMock, { dir: './static' })
+    app.get('/data', (c) =>
```
yusukebe, line 164: This is an actual bug. But this test will not fail if it does not have your fix. So I think the test should be like this, using a custom content type:

```diff
diff --git a/src/helper/ssg/ssg.test.tsx b/src/helper/ssg/ssg.test.tsx
index 8bf04251..0ffd2eab 100644
--- a/src/helper/ssg/ssg.test.tsx
+++ b/src/helper/ssg/ssg.test.tsx
@@ -162,15 +162,15 @@ describe('toSSG function', () => {

   it('Should correctly generate files with the expected paths', async () => {
     app.get('/data', (c) =>
-      c.text(JSON.stringify({ title: 'hono' }), 200, {
-        'Content-Type': 'application/json',
+      c.text('foo', 200, {
+        'Content-Type': 'text/x-foo',
       })
     )
     await toSSG(app, fsMock, {
       dir: './static',
       extensionMap: {
         ...defaultExtensionMap,
-        'application/json': 'json',
+        'text/x-foo': 'foo',
       },
     })

@@ -184,7 +184,7 @@ describe('toSSG function', () => {
     expect(fsMock.writeFile).toHaveBeenCalledWith('static/about.html', expect.any(String))
     expect(fsMock.writeFile).toHaveBeenCalledWith('static/bravo.html', expect.any(String))
     expect(fsMock.writeFile).toHaveBeenCalledWith('static/Charlie.html', expect.any(String))
-    expect(fsMock.writeFile).toHaveBeenCalledWith('static/data.json', expect.any(String))
+    expect(fsMock.writeFile).toHaveBeenCalledWith('static/data.foo', expect.any(String))
   })

   it('should modify the request if the hook is provided', async () => {
```

pspeter3: I need to chase down the bug more. Changing to `.foo` actually produces `.html`.

yusukebe: The `.json` with the content type `application/json` is defined `src/utils/mime.ts`. So that can handle it. If you want to use JSON with the content type  `application/json`, you don't need this change.

yusukebe (approved):
LGTM!

yusukebe:
@pspeter3 Looks good! I'll merge and release a new version soon. Thank you so much!

PR #4139, opened 2025-05-13
fix(hono-base): copy notfound and error handlers in `#clone()`

### LINKED ISSUE #4138: basepath with onerror

4.7.8

bun

If `basePath` is after `onError` then `onError` is never called and generates the `Internal Server Error`, however if you move it in front of the `onError` it works as expected.

```ts
const app = new Hono()

  .onError((error, c) => {
    console.log('Error caught in handler:', error)
    return c.json(
      { success: false, message: '>>>> Something went wrong!' },
      500,
    )
  })
  .basePath('/api')
  .get('/test', async () => {
    throw new Error('API Test error')
  })
  .get('/hello', async (c) => {
    return c.text('hello')
  })
```

To work without issues.

### What do you see instead?

_No response_

### Additional information

_No response_

yusukebe:
Fixes  #4138

With the current implementation, the `errorHandler` and `#notFoundHandler` are not copied in the `#clone()`. So the issue #4138 occurs. In this PR, it copies them.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/utils/body.test.ts
+++ b/src/utils/body.test.ts
@@ -199,8 +199,12 @@ describe('Parse Body Util', () => {
 
     const req = createRequest(FORM_URL, 'POST', data)
 
-    expect(await parseBody(req, { all: true, dot: true })).toEqual({
-      file: { file1, file2 },
+    const result = await parseBody(req, { all: true, dot: true })
+    expect(result).toMatchObject({
+      file: {
+        file1: { name: 'file1', type: 'application/octet-stream' },
+        file2: { name: 'file2', type: 'application/octet-stream' },
+      },
     })
   })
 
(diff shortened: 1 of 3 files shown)


src/utils/body.test.ts
```diff
@@ -199,8 +199,12 @@ describe('Parse Body Util', () => {
 
     const req = createRequest(FORM_URL, 'POST', data)
 
-    expect(await parseBody(req, { all: true, dot: true })).toEqual({
-      file: { file1, file2 },
+    const result = await parseBody(req, { all: true, dot: true })
```
yusukebe, line 202: This test didn't pass with the following message:

CleanShot 2025-05-14 at 08 07 05@2x

I'm not sure why this fails at this time, but I've fixed the test by comparing only the necessary elements.

src/utils/body.test.ts
```diff
@@ -217,8 +221,12 @@ describe('Parse Body Util', () => {
 
     const req = createRequest(FORM_URL, 'POST', data)
 
-    expect(await parseBody(req, { all: true })).toEqual({
-      file: [file1, file2],
+    const result = await parseBody(req, { all: true })
```
yusukebe, line 224: Same reason as the above.

yusukebe:
Hey @usualoma !

Can you review this? I think #4138 is not expected behavior and a bug.

usualoma:
Hi @yusukebe, Thank you.
LGTM!

yusukebe:
@usualoma 

Thanks! Merging.

Issue #4031, opened 2025-03-30
ERROR: Body is unusable: Body has already been read


cjnoname:
### What version of Hono are you using?

4.7.5

### What runtime/platform is your app running on? (with version if possible)

NodeJS on AWS Lambda

### What steps can reproduce the bug?

The code below randomly throws this error with a probability of 1 in 100 to 1 in 1000.

```json
{
    "errorType": "TypeError",
    "errorMessage": "Body is unusable: Body has already been read",
    "stack": [
        "TypeError: Body is unusable: Body has already been read",
        "    at consumeBody (node:internal/deps/undici/undici:5712:15)",
        "    at _Response.text (node:internal/deps/undici/undici:5662:18)",
        "    at s.createResult (file:///var/task/lambda.mjs:595:10175)",
        "    at Runtime.handler (file:///var/task/lambda.mjs:595:9620)"
    ]
}
```

```ts
import { Hono } from "hono";
import { secureHeaders } from "hono/secure-headers";
import { renderPage } from "./render";
import { randomBytes } from "crypto";
import { serveStatic } from "@hono/node-server/serve-static";
import { etag } from "hono/etag";

const app = new Hono();

app.use(etag({ weak: true }));

app.use(serveStatic({ root: "./static" }));

app.use((c, next) => {
  const nonce = randomBytes(16).toString("base64");
  c.set("secureHeadersNonce", nonce);

  return secureHeaders({
    contentSecurityPolicy: {
      scriptSrc: [`'self'`, `'strict-dynamic'`, `'nonce-${nonce}'`]
    }
  })(c, next);
});

app.get("*", async c => {
  const nonce = c.get("secureHeadersNonce");
  return c.html(await renderPage(c.req.path, nonce));
});

export default app;
```

cjnoname:
It's the same issue as 

https://github.com/honojs/node-server/issues/233

Plopix:
oh nice of you could reproduce it! I couldn't!

cjnoname:
> oh nice of you could reproduce it! I couldn't!

I found that the issue is with the ETag middleware. It works fine without it. I'm still doing more research, and I suspect it's a compatibility issue between the ETag middleware，serve-static middleware and the Lambda runtime. 

There was a slight inaccuracy in my statement. I mentioned that the probability of this occurring is low, but in reality, it is quite high. However, it primarily happens during the Lambda warm up phase. Once the warm up is complete and there are enough warm Lambdas, this issue no longer occurs.

[...]

PR #4166, opened 2025-05-26
fix(lambda): resolve ETag issue causing "Body has already been read"

cjnoname:
This PR fixes the etag middleware issue in AWS Lambda environments, where the response body cannot be read due to the "Body has already been read" error. This occurs when the Lambda runtime disallows re-reading the body stream, leading to failure when attempting to generate or validate ETags.

https://github.com/honojs/hono/issues/4031

### DIFF
--- a/src/middleware/etag/digest.ts
+++ b/src/middleware/etag/digest.ts
@@ -9,9 +9,6 @@ export const generateDigest = async (
   let result: ArrayBuffer | undefined = undefined
 
   if (input instanceof ArrayBuffer) {
-    if (input.byteLength === 0) {
-      return null
-    }
     result = await generator(new Uint8Array(input))
   } else {
     const chunks: Uint8Array[] = []
--- a/src/middleware/etag/digest.ts
+++ b/src/middleware/etag/digest.ts
@@ -1,38 +1,59 @@
-const mergeBuffers = (buffer1: ArrayBuffer | undefined, buffer2: Uint8Array): Uint8Array => {
-  if (!buffer1) {
-    return buffer2
-  }
-  const merged = new Uint8Array(buffer1.byteLength + buffer2.byteLength)
-  merged.set(new Uint8Array(buffer1), 0)
-  merged.set(buffer2, buffer1.byteLength)
-  return merged
-}
-
 export const generateDigest = async (
-  stream: ReadableStream<Uint8Array> | null,
+  input: ReadableStream<Uint8Array> | ArrayBuffer | null,
   generator: (body: Uint8Array) => ArrayBuffer | Promise<ArrayBuffer>
 ): Promise<string | null> => {
-  if (!stream) {
+  if (!input) {
     return null
   }
 
   let result: ArrayBuffer | undefined = undefined
 
-  const reader = stream.getReader()
-  for (;;) {
-    const { value, done } = await reader.read()
-    if (done) {
-      break
+  if (input instanceof ArrayBuffer) {
+    if (input.byteLength === 0) {
+      return null
     }
+    result = await generator(new Uint8Array(input))
+  } else {
+    const chunks: Uint8Array[] = []
+    let totalLength = 0
+    const reader = input.getReader()
 
-    result = await generator(mergeBuffers(result, value))
+    try {
+      while (true) {
+        const { value, done } = await reader.read()
+        if (done) break
+
+        if (value.length > 0) {
+          chunks.push(value)
+          totalLength += value.length
+        }
+      }
+
+      if (totalLength === 0) {
+        return null
+      }
+
+      const accumulated = new Uint8Array(totalLength)
+      let offset = 0
+      for (const chunk of chunks) {
+        accumulated.set(chunk, offset)
+        offset += chunk.length
+      }
+
+      result = await generator(accumulated)
+    } finally {
+      reader.releaseLock()
+    }
   }
 
-  if (!result) {
+  if (!result || result.byteLength === 0) {
     return null
   }
 
-  return Array.prototype.map
-    .call(new Uint8Array(result), (x) => x.toString(16).padStart(2, '0'))
-    .join('')
+  const bytes = new Uint8Array(result)
+  const hex = new Array(bytes.length)
+  for (let i = 0; i < bytes.length; i++) {
+    hex[i] = bytes[i].toString(16).padStart(2, '0')
+  }
+  return hex.join('')
 }
--- a/src/middleware/etag/index.test.ts
+++ b/src/middleware/etag/index.test.ts
@@ -293,5 +293,237 @@ describe('Etag Middleware', () => {
       expect(res.status).toBe(200)
       expect(res.headers.get('ETag')).toBeNull()
     })
+
+    it('Should not generate etag with custom generator when crypto is unavailable', async () => {
+      const app = new Hono()
+      app.use('/etag/*', etag({
+        generateDigest: undefined
+      }))
+      app.get('/etag/no-custom-digest', (c) => c.text('Hono is hot'))
+      const res = await app.request('/etag/no-custom-digest')
+      expect(res.status).toBe(200)
+      expect(res.headers.get('ETag')).toBeNull()
+    })
+  })
+
+  describe('AWS Lambda environment', () => {
+    let originalEnv: NodeJS.ProcessEnv
+
+    beforeAll(() => {
+      originalEnv = { ...process.env }
+    })
+
+    afterAll(() => {
+      process.env = originalEnv
+    })
+
+    it('Should handle etag generation in Lambda environment with AWS_LAMBDA_FUNCTION_NAME', async () => {
+      process.env.AWS_LAMBDA_FUNCTION_NAME = 'test-function'
+      delete process.env.AWS_EXECUTION_ENV
+      delete process.env.LAMBDA_TASK_ROOT
+
+      const app = new Hono()
+      app.use('/etag/*', etag())
+      app.get('/etag/lambda', (c) => {
+        return c.text('Hono in Lambda')
+      })
+
+      const res = await app.request('http://localhost/etag/lambda')
+      expect(res.status).toBe(200)
+      expect(res.headers.get('ETag')).not.toBeFalsy()
+      expect(await res.text()).toBe('Hono in Lambda')
+    })
+
+    it('Should handle etag generation in Lambda environment with AWS_EXECUTION_ENV', async () => {
+      delete process.env.AWS_LAMBDA_FUNCTION_NAME
+      process.env.AWS_EXECUTION_ENV = 'AWS_Lambda_nodejs18.x'
+      delete process.env.LAMBDA_TASK_ROOT
+
+      const app = new Hono()
+      app.use('/etag/*', etag())
+      app.get('/etag/lambda-exec', (c) => {
+        return c.text('Lambda execution env')
+      })
+
+      const res = await app.request('http://localhost/etag/lambda-exec')
+      expect(res.status).toBe(200)
+      expect(res.headers.get('ETag')).not.toBeFalsy()
+      expect(await res.text()).toBe('Lambda execution env')
+    })
+
+    it('Should handle etag generation in Lambda environment with LAMBDA_TASK_ROOT', async () => {
+      delete process.env.AWS_LAMBDA_FUNCTION_NAME
+      delete process.env.AWS_EXECUTION_ENV
+      process.env.LAMBDA_TASK_ROOT = '/var/task'
+
+      const app = new Hono()
+      app.use('/etag/*', etag())
+      app.get('/etag/lambda-task', (c) => {
+        return c.json({ message: 'Lambda task root', data: [1, 2, 3] })
+      })
+
+      const res = await app.request('http://localhost/etag/lambda-task')
+      expect(res.status).toBe(200)
+      expect(res.headers.get('ETag')).not.toBeFalsy()
+      const jsonData = await res.json()
+      expect(jsonData.message).toBe('Lambda task root')
+      expect(jsonData.data).toEqual([1, 2, 3])
+    })
+  })
+
+  describe('Edge cases and additional coverage', () => {
+    it('Should handle ReadableStream with mixed empty and non-empty chunks', async () => {
+      const app = new Hono()
+      app.use('/etag/*', etag())
+      app.get('/etag/mixed-chunks', (c) => {
+        const stream = new ReadableStream({
+          start(controller) {
+            controller.enqueue(new Uint8Array(0)) // Empty chunk
+            controller.enqueue(new Uint8Array([0xaa, 0xbb])) // Non-empty chunk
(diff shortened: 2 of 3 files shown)


yusukebe:
Hi @cjnoname 

Thank you for the PR. I'll review this, but can you add proper tests for this change?

cjnoname:
> Hi @cjnoname
> 
> Thank you for the PR. I'll review this, but can you add proper tests for this change?

Hey mate, I've added tests to bring the coverage up to 100%. Feel free to make any changes if needed.

yusukebe:
Does this really solve #4031? If PR #4198 can't do that, I think this PR also can't resolve this.

Though this is okay, I don't want to add runtime/platform-specific code like the following:

https://github.com/honojs/hono/blob/95e813bdaa7244bb6517d20775ebdd8ca3c62b3b/src/middleware/etag/index.ts#L93-L97

cjnoname:
> Does this really solve #4031? If PR #4198 can't do that, I think this PR also can't resolve this.
> 
> Though this is okay, I don't want to add runtime/platform-specific code like the following:
> 
> https://github.com/honojs/hono/blob/95e813bdaa7244bb6517d20775ebdd8ca3c62b3b/src/middleware/etag/index.ts#L93-L97

This solves the problem. We just need to ensure that res.clone().body is not consumed in a Lambda environment, because in Lambda, it consumes the "actual" body instead of the "cloned" one.

Your PR won’t work because the error isn’t thrown in the ETag file, but in the subsequent files. Any of those later files that try to consume the body will find that it has already been used, which causes the error.

If you prefer to avoid runtime/platform-specific code, you can use ArrayBuffer in all environments. It’s slightly slower than using clone().body, but the difference is hardly noticeable. That’s why I wrote the code this way—to ensure maximum speed in non-Lambda environments.

yusukebe:
@cjnoname 

So, is it possible that the related problem occurs not only in ETag Middleware?

cjnoname:
> you prefer to avoid runtime/platform-specific code

Good question. I searched for .body usage across the entire application, and currently, the ETag middleware is the only place where res.clone().body is being used. In the Lambda environment, the body can only be consumed once, so for now, this issue only affects the ETag logic. However, if another middleware in the future also tries to consume res.clone().body, it will encounter the same problem in AWS Lambda.

The core difference is: in a browser environment, clone() duplicates both the response object and the underlying stream, allowing the stream to be consumed multiple times.

In contrast, in a Lambda environment, clone() only duplicates the response object—not the underlying stream.

To avoid writing runtime/platform-specific code, there are two ways to solve this:

1. Consume the body only once – In this case, the ETag middleware should use res.arrayBuffer() instead of res.clone().body.

2. Consume res.body once globally, cache it in memory, and reuse it wherever needed.

I believe the second option is the one you prefer, but it requires a more global understanding of the changes. Unfortunately, I’m unable to help with that, as I don’t have enough context on how the response body is being used throughout the application.

yusukebe:
@cjnoname 

Thank you for the explanation. I don't want to change the internal logic, but how about adding the following option, like `generateDigestFromContext` or `generateDigestFromResponse`?

```ts
type ETagOptions = {
  retainedHeaders?: string[]
  weak?: boolean
  generateDigest?: (body: Uint8Array) => ArrayBuffer | Promise<ArrayBuffer>
  generateDigestFromContext?: (c: Context) => ArrayBuffer | Promise<ArrayBuffer> // Add
  // or `generateDigestFromResponse`
  // generateDigestFromResponse?: (res: Response) => ArrayBuffer | Promise<ArrayBuffer>
}
```

If you want to use ETag Middleware on AWS Lambda, you can pass the function to generate a hash from the `Context` object or the `Response` object. It will be a little bit bothersome to you, but this is more generic, and there will not be runtime/platform-specific code and performance degradation.

cjnoname:
> @cjnoname
> 
> So, is it possible that the related problem occurs not only in ETag Middleware?

This answer was generated by GPT-4.1. I hope it provides you with a deeper understanding of the issue.

You asked about [res.clone().body] and why, in Lambda environments, it sometimes doesn't work as expected (the stream isn't really cloned).

Explanation:

In standard environments (like browsers or Node.js with undisturbed Fetch API), [res.clone().body] gives you a fresh, readable stream. You can read the body multiple times: once from the original, once from the clone.
In AWS Lambda (or similar serverless environments), the Response object may not be a true Web Fetch Response. Sometimes, the body stream is already consumed or not properly cloneable. This means [res.clone().body] might not actually give you a new, readable stream—it could be locked, empty, or throw errors.

Why does this happen?

Lambda runtimes may wrap or polyfill the Response object, and their implementation of [clone()] may not fully support duplicating the stream.

Streams in JavaScript can only be read once. If the Lambda runtime or a previous middleware has already read the stream, cloning won't work.

How is this handled in your code?

Your code checks if it's running in Lambda ([isLambda]).
If so, it reads the body as an [ArrayBuffer] immediately ([await res.arrayBuffer()], then re-creates the Response. This ensures the body is available for hashing and for the client.
In non-Lambda environments, it safely uses [res.clone().body].

Summary:
In Lambda, [res.clone().body] is unreliable because the stream may already be consumed or not properly cloneable. The workaround is to read the body as a buffer and re-create the Response, as your code does. This ensures consistent behavior across environments.

yusukebe:
@cjnoname 

So I can understand the behavior on the AWS Lambda, but I want to know if that does https://github.com/honojs/hono/pull/4166#issuecomment-2951308753 work for you?

cjnoname:
> @cjnoname
> 
> Thank you for the explanation. I don't want to change the internal logic, but how about adding the following option, like `generateDigestFromContext` or `generateDigestFromResponse`?
> 
> ```ts
> type ETagOptions = {
>   retainedHeaders?: string[]
>   weak?: boolean
>   generateDigest?: (body: Uint8Array) => ArrayBuffer | Promise<ArrayBuffer>
>   generateDigestFromContext?: (c: Context) => ArrayBuffer | Promise<ArrayBuffer> // Add
>   // or `generateDigestFromResponse`
>   // generateDigestFromResponse?: (res: Response) => ArrayBuffer | Promise<ArrayBuffer>
> }
> ```
> 
> If you want to use ETag Middleware on AWS Lambda, you can pass the function to generate a hash from the `Context` object or the `Response` object. It will be a little bit bothersome to you, but this is more generic, and there will not be runtime/platform-specific code and performance degradation.


I cannot see the full implementation, but I’m sure any change that avoids calling res.clone().body can help.

But the best approach is still to read the full body once at the beginning and cache it so all subsequent middlewares can share it.

yusukebe:
> But the best approach is still to read the full body once at the beginning and cache it so all subsequent middlewares can share it.

Using `await res.arrayBuffer()` will affect the performance if it sends a large amount of content. I think we shoud go with https://github.com/honojs/hono/pull/4166#issuecomment-2951308753

cjnoname:
> > But the best approach is still to read the full body once at the beginning and cache it so all subsequent middlewares can share it.
> 
> Using `await res.arrayBuffer()` will affect the performance if it sends a large amount of content. I think we shoud go with [#4166 (comment)](https://github.com/honojs/hono/pull/4166#issuecomment-2951308753)

I mean, if you could read res.body() once at the beginning and cache it in memory for all upcoming uses to reuse, that would be the best approach.

For example, in your context, you could define a variable like bodyText (or any suitable name) to hold the content of res.body(). Then, middlewares like etag can simply read the body from your context instead of cloning and reading it each time.

yusukebe:
> For example, in your context, you could define a variable like bodyText (or any suitable name) to hold the content of res.body(). Then, middlewares like etag can simply read the body from your context instead of cloning and reading it each time.

Does this method really work on AWS Lambda? Can it read `res.body()` on AWS Lambda? Anyway, adding the option will be simpler and don't touch the logics. I'll create a PR.

cjnoname:
> res.body()

Lambda can read res.body(). You need to understand that the body is actually a stream, which can only be consumed once. If it is consumed in the etag middleware, it is gone and cannot be consumed again in the actual place where it is needed.

yusukebe:
Do you mean changing this `res.clone().body` to `res.body` will work?

https://github.com/honojs/hono/blob/4a1dd5f66b2b30d7576ea15171423a51b3ed1a59/src/middleware/etag/index.ts#L96

To:

```ts
const hash = await generateDigest(res.body, generator) 
```

cjnoname:
> Do you mean changing this `res.clone().body` to `res.body` will work?
> 
> https://github.com/honojs/hono/blob/4a1dd5f66b2b30d7576ea15171423a51b3ed1a59/src/middleware/etag/index.ts#L96
> 
> To:
> 
> ```ts
> const hash = await generateDigest(res.body, generator) 
> ```



> Do you mean changing this `res.clone().body` to `res.body` will work?
> 
> https://github.com/honojs/hono/blob/4a1dd5f66b2b30d7576ea15171423a51b3ed1a59/src/middleware/etag/index.ts#L96
> 
> To:
> 
> ```ts
> const hash = await generateDigest(res.body, generator) 
> ```

Lambda 環境では、時々 body のクローン（clone()）が正しく動作しないことがあります。body 自体は正常に動作していますが、もしこの時点で body を消費してしまうと、後続の処理で再び body を使用しようとした際にエラーが発生してしまいます。

yusukebe:
@cjnoname 

I can understand. But can you answer the question? If not, we can't help you. Sorry.

yusukebe:
My idea is the following:

```diff
diff --git a/src/middleware/etag/digest.ts b/src/middleware/etag/digest.ts
index 8c6b6f97..27c7941a 100644
--- a/src/middleware/etag/digest.ts
+++ b/src/middleware/etag/digest.ts
@@ -9,23 +9,25 @@ const mergeBuffers = (buffer1: ArrayBuffer | undefined, buffer2: Uint8Array): Ui
 }

 export const generateDigest = async (
-  stream: ReadableStream<Uint8Array> | null,
+  input: ReadableStream<Uint8Array> | ArrayBuffer | null,
   generator: (body: Uint8Array) => ArrayBuffer | Promise<ArrayBuffer>
 ): Promise<string | null> => {
-  if (!stream) {
+  if (!input) {
     return null
   }

   let result: ArrayBuffer | undefined = undefined
-
-  const reader = stream.getReader()
-  for (;;) {
-    const { value, done } = await reader.read()
-    if (done) {
-      break
+  if (input instanceof ArrayBuffer) {
+    result = await generator(new Uint8Array(input))
+  } else {
+    const reader = input.getReader()
+    for (;;) {
+      const { value, done } = await reader.read()
+      if (done) {
+        break
+      }
+      result = await generator(mergeBuffers(result, value))
     }
-
-    result = await generator(mergeBuffers(result, value))
   }

   if (!result) {
diff --git a/src/middleware/etag/index.ts b/src/middleware/etag/index.ts
index 613070af..9b4c3c76 100644
--- a/src/middleware/etag/index.ts
+++ b/src/middleware/etag/index.ts
@@ -3,6 +3,7 @@
  * ETag Middleware for Hono.
  */

+import type { Context } from '../../context'
 import type { MiddlewareHandler } from '../../types'
 import { generateDigest } from './digest'

@@ -10,6 +11,7 @@ type ETagOptions = {
   retainedHeaders?: string[]
   weak?: boolean
   generateDigest?: (body: Uint8Array) => ArrayBuffer | Promise<ArrayBuffer>
+  getBodyFromContext?: (c: Context) => ArrayBuffer | Promise<ArrayBuffer>
 }

 /**
@@ -93,7 +95,12 @@ export const etag = (options?: ETagOptions): MiddlewareHandler => {
       if (!generator) {
         return
       }
-      const hash = await generateDigest(res.clone().body, generator)
+      const body = options?.getBodyFromContext
+        ? await options.getBodyFromContext(c)
+        : res.clone().body
+
+      const hash = await generateDigest(body, generator)
+
       if (hash === null) {
         return
```

The usage:

```ts
const app = new Hono()
app.use(
  '/etag/*',
  etag({
    getBodyFromContext: async (c) => {
      return await c.res.arrayBuffer()
    },
  })
)
```

This does not break the logic on platforms/runtimes other than AWS Lambda.

cjnoname:
> My idea is the following:
> 
> ```diff
> diff --git a/src/middleware/etag/digest.ts b/src/middleware/etag/digest.ts
> index 8c6b6f97..27c7941a 100644
> --- a/src/middleware/etag/digest.ts
> +++ b/src/middleware/etag/digest.ts
> @@ -9,23 +9,25 @@ const mergeBuffers = (buffer1: ArrayBuffer | undefined, buffer2: Uint8Array): Ui
>  }
> 
>  export const generateDigest = async (
> -  stream: ReadableStream<Uint8Array> | null,
> +  input: ReadableStream<Uint8Array> | ArrayBuffer | null,
>    generator: (body: Uint8Array) => ArrayBuffer | Promise<ArrayBuffer>
>  ): Promise<string | null> => {
> -  if (!stream) {
> +  if (!input) {
>      return null
>    }
> 
>    let result: ArrayBuffer | undefined = undefined
> -
> -  const reader = stream.getReader()
> -  for (;;) {
> -    const { value, done } = await reader.read()
> -    if (done) {
> -      break
> +  if (input instanceof ArrayBuffer) {
> +    result = await generator(new Uint8Array(input))
> +  } else {
> +    const reader = input.getReader()
> +    for (;;) {
> +      const { value, done } = await reader.read()
> +      if (done) {
> +        break
> +      }
> +      result = await generator(mergeBuffers(result, value))
>      }
> -
> -    result = await generator(mergeBuffers(result, value))
>    }
> 
>    if (!result) {
> diff --git a/src/middleware/etag/index.ts b/src/middleware/etag/index.ts
> index 613070af..9b4c3c76 100644
> --- a/src/middleware/etag/index.ts
> +++ b/src/middleware/etag/index.ts
> @@ -3,6 +3,7 @@
>   * ETag Middleware for Hono.
>   */
> 
> +import type { Context } from '../../context'
>  import type { MiddlewareHandler } from '../../types'
>  import { generateDigest } from './digest'
> 
> @@ -10,6 +11,7 @@ type ETagOptions = {
>    retainedHeaders?: string[]
>    weak?: boolean
>    generateDigest?: (body: Uint8Array) => ArrayBuffer | Promise<ArrayBuffer>
> +  getBodyFromContext?: (c: Context) => ArrayBuffer | Promise<ArrayBuffer>
>  }
> 
>  /**
> @@ -93,7 +95,12 @@ export const etag = (options?: ETagOptions): MiddlewareHandler => {
>        if (!generator) {
>          return
>        }
> -      const hash = await generateDigest(res.clone().body, generator)
> +      const body = options?.getBodyFromContext
> +        ? await options.getBodyFromContext(c)
> +        : res.clone().body
> +
> +      const hash = await generateDigest(body, generator)
> +
>        if (hash === null) {
>          return
> ```
> 
> The usage:
> 
> ```ts
> const app = new Hono()
> app.use(
>   '/etag/*',
>   etag({
>     getBodyFromContext: async (c) => {
>       return await c.res.arrayBuffer()
>     },
>   })
> )
> ```
> 
> This does not break the logic on platforms/runtimes other than AWS Lambda.

This should work, mate. I'll try it out for a few days before merging the code.

PR #4182, opened 2025-05-31
fix(hono/context): content-type be overridden by default response when append headers

xx1124961758:
fix node-server bug https://github.com/honojs/node-server/issues/226

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/context.test.ts
+++ b/src/context.test.ts
@@ -172,6 +172,17 @@ describe('Context', () => {
     expect(foo).toBe('Bar, Buzz')
   })
 
+  it('c.body() - content-type cannot be overridden by the default response when append headers', async () => {
+    const triggerResGetter = c.res
+    c.header('Vary', 'Accept-Encoding', { append: true })
+    const res = c.body('<h1>Hi</h1>', 200, {
+      'X-Foo': ['Bar', 'Buzz'],
+      'Content-Type': 'text/html;charset=utf-8',
+    })
+    const foo = res.headers.get('Content-Type')
+    expect(foo).toBe('text/html;charset=utf-8')
+  })
+
   it('c.status()', async () => {
     c.status(201)
     const res = c.body('Hi')
--- a/src/context.test.ts
+++ b/src/context.test.ts
@@ -173,14 +173,11 @@ describe('Context', () => {
   })
 
   it('c.body() - content-type cannot be overridden by the default response when append headers', async () => {
-    const triggerResGetter = c.res
     c.header('Vary', 'Accept-Encoding', { append: true })
-    const res = c.body('<h1>Hi</h1>', 200, {
-      'X-Foo': ['Bar', 'Buzz'],
-      'Content-Type': 'text/html;charset=utf-8',
-    })
-    const foo = res.headers.get('Content-Type')
-    expect(foo).toBe('text/html;charset=utf-8')
+    c.res
+    c.header('Content-Type', 'text/html')
+    const res = c.body('<h1>Hi</h1>')
+    expect(res.headers.get('Content-Type')).toMatch('text/html')
   })
 
   it('c.status()', async () => {
--- a/src/context.ts
+++ b/src/context.ts
@@ -670,7 +670,7 @@ export class Context<
       this.#res.headers.forEach((v, k) => {
         if (k === 'set-cookie') {
           this.#headers?.append(k, v)
-        } else if (this.finalized || k !== 'content-type') {
+        } else if (k !== 'content-type') {
           this.#headers?.set(k, v)
         }
       })
--- a/src/context.ts
+++ b/src/context.ts
@@ -670,7 +670,7 @@ export class Context<
       this.#res.headers.forEach((v, k) => {
         if (k === 'set-cookie') {
           this.#headers?.append(k, v)
-        } else {
+        } else if (this.finalized || k !== 'content-type') {
           this.#headers?.set(k, v)
         }
       })


sushichan044:
@xx1124961758 @yusukebe 
This is just for your information, but it's likely related to this issue as well: 
- https://github.com/honojs/hono/issues/3736#issuecomment-2814695601

src/context.ts
```diff
@@ -670,7 +670,7 @@ export class Context<
       this.#res.headers.forEach((v, k) => {
         if (k === 'set-cookie') {
           this.#headers?.append(k, v)
-        } else {
+        } else if (this.finalized || k !== 'content-type') {
```
yusukebe, line 673: I think we can write this condition more simply:

```ts
} else if (!(k === 'content-type')) {
```

src/context.test.ts
```diff
@@ -172,6 +172,17 @@ describe('Context', () => {
     expect(foo).toBe('Bar, Buzz')
   })
 
+  it('c.body() - content-type cannot be overridden by the default response when append headers', async () => {
```
yusukebe, line 175: This test is not good since it does not fail with the old code. Please replace it with the following?

```ts
it('c.body() - content-type cannot be overridden by the default response when append headers', async () => {
  c.header('Vary', 'Accept-Encoding', { append: true })
  c.res
  c.header('Content-Type', 'text/html')
  const res = c.body('<h1>Hi</h1>')
  expect(res.headers.get('Content-Type')).toMatch('text/html')
})
```

yusukebe:
Hi @xx1124961758 Thank you for the PR. I've left the comments.

@sushichan044 Do you think this PR will fix #3736? If so, please review this. You may know the PR #4100 can fix the problem, but I'm not sure we have to merge #4100. So, currently, it's a good idea to use this PR.

sushichan044:
@yusukebe 
I also think this pull request can fix the behavior reported in the issue.
However, as you said, #4100 is necessary for a fundamental solution.
Anyway, I'll review this PR.

src/context.ts
```diff
@@ -670,7 +670,7 @@ export class Context<
       this.#res.headers.forEach((v, k) => {
         if (k === 'set-cookie') {
           this.#headers?.append(k, v)
-        } else {
+        } else if (k !== 'content-type') {
```
sushichan044, line 673: If the branching logic is implemented this way, the test below will fail.
This is because the `content-type`, even if explicitly set via `c.res.headers`, would be ignored.

```ts
  it('c.body() - content-type can set explicitly via c.res.headers', async () => {
    c.header('Vary', 'Accept-Encoding', { append: true })
    c.res.headers.set('Content-Type', 'text/html')
    const res = c.body('<h1>Hi</h1>')
    expect(res.headers.get('Content-Type')).toMatch('text/html')
  })
```

sushichan044: @xx1124961758 @yusukebe 
To avoid this kind of behavior, one could consider rewriting it into a conditional branch like this, for example.
What do you think?

```ts
this.#res.headers.forEach((v, k) => {
  if (k === 'content-type' && this.#headers?.get('content-type')) {
    return
  }

  if (k === 'set-cookie') {
    this.#headers?.append(k, v)
  } else {
    this.#headers?.set(k, v)
  }
})
```

src/context.test.ts
```diff
@@ -172,6 +172,14 @@ describe('Context', () => {
...
+    c.header('Vary', 'Accept-Encoding', { append: true })
+    c.res
+    c.header('Content-Type', 'text/html')
+    const res = c.body('<h1>Hi</h1>')
+    expect(res.headers.get('Content-Type')).toMatch('text/html')
+  })
+
```
sushichan044, line 182: Maybe we should add this test to ensure users can set `content-type` explicitly via `c.res.headers` .

```ts
  it('c.body() - content-type can set explicitly via c.res.headers', async () => {
    c.header('Vary', 'Accept-Encoding', { append: true })
    c.res.headers.set('Content-Type', 'text/html')
    const res = c.body('<h1>Hi</h1>')
    expect(res.headers.get('Content-Type')).toMatch('text/html')
  })
```

sushichan044 (commented):
It looks good for the most part!
I'm commenting because I've found a concern about `c.res.headers` usage for users.

sushichan044:
@xx1124961758 
To make things clearer, we should also note in a comment that this fix is an ad-hoc workaround for issue #3736.
This conditional branch is special handling, and it's difficult to understand the reasoning behind it without knowing the background.

src/context.test.ts
xx1124961758: @yusukebe @sushichan044 

If we mix c.res.headers.set and c.header to set the same header, and c.header is called before c.res.headers.set.
Using the logic you provided, the following test will fail

```javascript
it('c.body() - Different header settings require ensuring order', async () => {
   c.header("Vary", "Accept-Encoding", { append: true });
   c.header("Content-Type", "image/png");
   c.res.headers.set("Content-Type", "text/html");
   const res = c.body("<h1>Hi</h1>");
   expect(res.headers.get('Content-Type')).toMatch('text/html')
})
```

Can we consider maintaining all headers uniformly on the # res object.

yusukebe: @xx1124961758 @sushichan044 

Thank you for providing the test cases.

> Can we consider maintaining all headers uniformly on the # res object. 

Okay. We have to consider it. I already created the PR to make the Context simpler and prevent weird behavior: #4100

#4100 will introduce performance degradation for specific situations, but we have to consider it anyway.

yusukebe:
The #4100 was merged. This can be closed. Thank you!

PR #4183, opened 2025-06-01
fix(utils/body): normalize key names in parseBody (#4108)

### LINKED ISSUE #4108: parseBody does not wrap single‐element key[] fields into arrays nor strip the [] suffix

4.6.14

Node

When submitting a multipart/form-data or application/x-www-form-urlencoded request with any field name ending in [], the first (and only) value is treated as a scalar and the key is left with its [] suffix.

According to the docs, any form field named key[] should always be parsed as a (string | File)[]—even if the client only submits one value. However, in the current implementation:

"foo[]": "5" instead of foo: ["5"]

### What do you see instead?

"foo[]": "5"

### Additional information

_No response_

hiroki-307:
Fixes #4108 
- [x]  Add tests
- [x] Run tests

### DIFF
--- a/src/utils/body.test.ts
+++ b/src/utils/body.test.ts
@@ -232,19 +232,19 @@ describe('Parse Body Util', () => {
 
   it('should parse multiple values if key ends with `[]`', async () => {
     const data = new FormData()
-    data.append('foo[]', 'aaa')
-    data.append('foo[]', 'bbb')
+    data.append('file[]', 'aaa')
+    data.append('file[]', 'bbb')
     data.append('message', 'hello')
 
     const req = createRequest(FORM_URL, 'POST', data)
 
     expect(await parseBody(req, { all: true })).toEqual({
-      'foo[]': ['aaa', 'bbb'],
+      'file[]': ['aaa', 'bbb'],
       message: 'hello',
     })
   })
 
-  it('should parse single value as array if the key has []', async () => {
+  it('should parse single value as array if the key has `[]`', async () => {
     const data = new FormData()
     data.append('foo[]', 'bar')
     const req = createRequest(FORM_URL, 'POST', data)
--- a/src/utils/body.test.ts
+++ b/src/utils/body.test.ts
@@ -232,15 +232,25 @@ describe('Parse Body Util', () => {
 
   it('should parse multiple values if key ends with `[]`', async () => {
     const data = new FormData()
-    data.append('file[]', 'aaa')
-    data.append('file[]', 'bbb')
+    data.append('foo[]', 'aaa')
+    data.append('foo[]', 'bbb')
     data.append('message', 'hello')
 
     const req = createRequest(FORM_URL, 'POST', data)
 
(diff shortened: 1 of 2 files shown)


src/utils/body.test.ts
```diff
@@ -239,7 +239,7 @@ describe('Parse Body Util', () => {
     const req = createRequest(FORM_URL, 'POST', data)
 
     expect(await parseBody(req, { all: true })).toEqual({
-      'file[]': ['aaa', 'bbb'],
+      file: ['aaa', 'bbb'],
```
yusukebe: I think this test is wrong. We have to test the case where the value is single.

In my opinion, we should not strip `[]` because it prevents the introduction of breaking changes. So the ideal test is as follows:

```ts
it('should parse single value as array if the key has []', async () => {
  const data = new FormData()
  data.append('foo[]', 'bar')
  expect(await parseBody(req, { all: true })).toEqual({
    'foo[]': ['bar'],
  })
})
```

@sor4chi What do you think of this?

sor4chi: Hi @yusukebe 
I think so.
In issue #4108, it appears that when the postfix is [], the return value is:

- string | File when there is a single element
- (string | File)[] when there are two or more elements

However, the type is defined as `(string | File)[]`, which does not match the actual return value when there is only one element.
Screenshot 2025-06-02 at 14 51 14

hiroki-307:
I've updated the code to always return an array when the key ends with ```[]```, and added a corresponding test.
Also fixed the test key name (was file, now reflects the actual case).

Do we need a test for when the value is a File and the key ends with ```[]```?

src/utils/body.test.ts
```diff
@@ -232,18 +232,28 @@ describe('Parse Body Util', () => {
 
   it('should parse multiple values if key ends with `[]`', async () => {
```
yusukebe: I think you don't need to change the name of the key. You can keep `file` as is.

yusukebe:
Thanks @hiroki-307, it's almost good. I commented.

@sor4chi Can you also review this?

src/utils/body.test.ts
```diff
@@ -232,18 +232,28 @@ describe('Parse Body Util', () => {
...
     const req = createRequest(FORM_URL, 'POST', data)
 
     expect(await parseBody(req, { all: true })).toEqual({
-      'file[]': ['aaa', 'bbb'],
+      'foo[]': ['aaa', 'bbb'],
       message: 'hello',
     })
   })
 
+  it('should parse single value as array if the key has []', async () => {
```
sor4chi:
```suggestion
  it('should parse single value as array if the key has `[]`', async () => {
```
nit

yusukebe (approved):
LGTM!

yusukebe:
Hi @hiroki-307 !

Looks good to me! I'll merge it and include the next minor version (since the behavior has changed slightly). Thank you for your contribution.

PR #4189, opened 2025-06-02
Add test for POST method in testClient

erik-beus:

- [x] Add tests
- [x] Run tests

From the docs, it was not clear to me how to pass a body to the `testClient`. I thought I would start by adding a test case since all existing tests for the `testClient` are all `GET` requests. Later, I plan to open a PR in the docs repo to clarify how to pass a body to the `testClient`.

See docs here https://hono.dev/docs/helpers/testing#testclient

### DIFF
--- a/src/helper/testing/index.test.ts
+++ b/src/helper/testing/index.test.ts
@@ -17,6 +17,22 @@ describe('hono testClient', () => {
     expect(await res.json()).toEqual({ hello: 'world' })
   })
 
+  it('Should support POST methods', async () => {
+    const app = new Hono().post('/search', async (c) => {
+      const body = await c.req.json()
+      return c.json({ echo: body.hello })
+    })
+    const res = await testClient(app).search.$post(
+      {},
+      {
+        headers: { 'Content-Type': 'application/json' },
+        init: { body: JSON.stringify({ hello: 'world' }) },
+      }
+    )
+    expect(res.status).toBe(200)
+    expect(await res.json()).toEqual({ echo: 'world' })
+  })
+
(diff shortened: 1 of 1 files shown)


erik-beus:
I've now added a PR in the website repository for adding an example https://github.com/honojs/website/pull/675

yusukebe:
@erik-beus 

The client returned from `testClient(app)` is the same as the client used in RPC.  And the `init` option is tested here: https://github.com/honojs/hono/blob/8fe1951d37a67eb101849f508e3eb3975dcfb8ed/src/client/client.test.ts#L945

I think this test is unnecessary.

Plus, if you want to mention it on the website, you don't have to add the description on https://hono.dev/docs/helpers/testing, but you can only add the link to the `init` option section of the RPC: https://hono.dev/docs/guides/rpc#init-option

erik-beus:
> @erik-beus
> 
> The client returned from `testClient(app)` is the same as the client used in RPC. And the `init` option is tested here:
> 
> https://github.com/honojs/hono/blob/8fe1951d37a67eb101849f508e3eb3975dcfb8ed/src/client/client.test.ts#L945
> 
> I think this test is unnecessary.
> 
> Plus, if you want to mention it on the website, you don't have to add the description on [hono.dev/docs/helpers/testing](https://hono.dev/docs/helpers/testing), but you can only add the link to the `init` option section of the RPC: [hono.dev/docs/guides/rpc#init-option](https://hono.dev/docs/guides/rpc#init-option)

Thank you. That makes sense. I will close this PR and update the PR for the docs with a link to the init option 👍

PR #4198, opened 2025-06-04
fix(etag): fallback if `res.clone()` is not supported

### LINKED ISSUE #4031: ERROR: Body is unusable: Body has already been read

yusukebe:
This PR handles the environment that does not support it. I'm not sure the Lambda does not support `res.clone()`, but if so, this PR will fix #4031.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/etag/digest.ts
+++ b/src/middleware/etag/digest.ts
@@ -9,23 +9,25 @@ const mergeBuffers = (buffer1: ArrayBuffer | undefined, buffer2: Uint8Array): Ui
 }
 
 export const generateDigest = async (
-  stream: ReadableStream<Uint8Array> | null,
+  input: ReadableStream<Uint8Array> | ArrayBuffer | null,
   generator: (body: Uint8Array) => ArrayBuffer | Promise<ArrayBuffer>
 ): Promise<string | null> => {
-  if (!stream) {
+  if (!input) {
     return null
   }
 
   let result: ArrayBuffer | undefined = undefined
-
-  const reader = stream.getReader()
-  for (;;) {
-    const { value, done } = await reader.read()
-    if (done) {
-      break
+  if (input instanceof ArrayBuffer) {
+    result = await generator(new Uint8Array(input))
+  } else {
+    const reader = input.getReader()
+    for (;;) {
+      const { value, done } = await reader.read()
+      if (done) {
+        break
+      }
+      result = await generator(mergeBuffers(result, value))
     }
-
-    result = await generator(mergeBuffers(result, value))
   }
 
   if (!result) {
--- a/src/middleware/etag/index.test.ts
+++ b/src/middleware/etag/index.test.ts
@@ -155,6 +155,29 @@ describe('Etag Middleware', () => {
     expect(res.headers.get('ETag')).toBeNull()
   })
 
+  it('Should handle clone() not supported (Lambda-like environment)', async () => {
+    const app = new Hono()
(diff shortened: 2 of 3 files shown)


yusukebe:
Hi @cjnoname

First, thank you for creating #4031! But just that if the AWS Lambda does not support `res.clone()`, this PR can fix it, and the code will be simpler. Can you see if this works as expected on the AWS Lambda?

cjnoname:
> Hi @cjnoname
> 
> First, thank you for creating #4031! But just that if the AWS Lambda does not support `res.clone()`, this PR can fix it, and the code will be simpler. Can you see if this works as expected on the AWS Lambda?

Hey mate, I’ve quickly tested your commits and everything is working as expected. Hopefully, it can be released soon. Thanks!

yusukebe:
@cjnoname Thanks! I'll merge this.

cjnoname:
> @cjnoname Thanks! I'll merge this.

Hey mate,

Unfortunately, the error occurred again this morning—sorry, it seems the test I ran yesterday didn’t work.

It appears the error wasn’t actually thrown at that point. Instead, the issue is that the .clone() didn’t take effect under Lambda; the original body was used instead. As a result, when the body was needed later, it had already been incorrectly consumed during the ETag step, causing the error.

It looks like we still need to update the code to prevent the original body from being consumed too early during the ETag process.

{
    "errorType": "TypeError",
    "errorMessage": "Body is unusable: Body has already been read",
    "stack": [
        "TypeError: Body is unusable: Body has already been read",
        "    at consumeBody (node:internal/deps/undici/undici:5712:15)",
        "    at _Response.text (node:internal/deps/undici/undici:5662:18)",
        "    at s.createResult (file:///var/task/lambda.mjs:500:123913)",
        "    at Runtime.handler (file:///var/task/lambda.mjs:500:123358)"
    ]
}

yusukebe:
@cjnoname Okay, I'll revert the PR.

PR #4202, opened 2025-06-07
feat(types): export `MergeTypedResponse` type

lazuee:
Expose MergeTypedResponse for external use cases.

- [ ] Add tests
- [ ] Run tests


yusukebe:
@lazuee 

What use case do you want to use it for? ~~Basically, you can't use the types exported only from `src/types.ts`~~

lazuee:
@yusukebe https://github.com/lazuee/hono-emitter/blob/fb95e63fde86038cf138c21d4e7cdf45470bcc10/packages/hono-emitter/src/types.ts#L34

yusukebe:
@lazuee 

Thanks.

Hmm. Regarding the "types", we don't want to export them as much as possible. They are used as utilities and may be changed without announcement.

Issue #3127, opened 2024-07-11
Mark `app.fire` as deprecated


yusukebe:
See: https://github.com/honojs/hono/pull/3062#issuecomment-2211710698

EdamAme-x:
I think it is necessary to separate the `.fire()` function from the core.  
It would be even more minimalistic and eliminate the need to put processes in the core that depend on a specific environment.  
How about using `cloudflare-workers adapter` or similar to do the following?  

```ts
import { fire } from "hono/cloudflare-workers"

fire(app)
````

**Comment by yusukebe** [comment]
Quick response. You can use the Service Worker adapter.

https://hono.dev/docs/getting-started/service-worker

**Comment by EdamAme-x** [comment]
Thank you.
Are there any plans for the `.fire()` function to be deprecated?

**Comment by yusukebe** [comment]
> Are there any plans for the `.fire()` function to be deprecated?

Yes. We should make it deprecated.

**Comment by yusukebe** [comment]
With this change, we can mark the `event` property in `Context` like `c.event` deprecated. It was used in only Service Worker syntax.

**Comment by rajsite** [comment]
> Quick response. You can use the Service Worker adapter.
> 
> https://hono.dev/docs/getting-started/service-worker

For typing, the service workers docs suggest enabling `webworker` lib in TS. I find this causes a lot of type conflicts when `DOM` lib or `jsx` tsconfig is enabled. When I used it it forced me to enable `skipLibCheck` which I want to avoid.

Also for the runtime I'm using, I needed to disable fallthrough which otherwise gets stuck in a loop and is easy to miss when using service worker:

```ts
import { handle } from 'hono/service-worker';

addEventListener('fetch', handle(app, {
  // Disable fallthrough which would get stuck in a loop
  fetch: undefined
}));
```

I'd really appreciate a service worker `fire` function similar to the above proposal to avoid the type troubles and footgun for fetch fall through:

```ts
import { fire } from "@hono/service-worker"

fire(app) // 😍
```

yusukebe:
@rajsite 

Wooow! `fire` is an awesome idea. Love it!

yusukebe:
We can close this now. See #4231. Thanks!

alex8bitw:
> [@rajsite](https://github.com/rajsite)
> 
> Wooow! `fire` is an awesome idea. Love it!

Unfortunately, this doesn't seem to work for Fastly Edge Compute.

yusukebe:
@alex8bitw 

[...]

PR #4214, opened 2025-06-13
feat(service-worker): add `fire()`

yusukebe:
This PR introduces the `fire()` method in the Service Worker adapter. It will do `addEventListener('fetch', handle(app))` for the given Hono instance. Usage:

```ts
import { Hono } from 'hono'
import { fire } from 'hono/service-worker'

const app = new Hono()

app.get('/', (c) => c.text('Hi'))

fire(app)
```

The background: https://github.com/honojs/hono/issues/3127#issuecomment-2843529188

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/src/adapter/service-worker/index.ts
+++ b/src/adapter/service-worker/index.ts
@@ -30,7 +30,7 @@ const fire = (
     fetch: undefined,
   }
 ): void => {
-  // @ts-expect-error addEventListener is not typed well
+  // @ts-expect-error addEventListener is not typed well in ServiceWorker-like contexts, see: https://github.com/microsoft/TypeScript/issues/14877
   addEventListener('fetch', handle(app, options))
 }
 


yusukebe:
Hi @rajsite!

I create the PR for the feature you may want. Can you review this?

src/adapter/service-worker/index.ts
```diff
@@ -2,4 +2,29 @@
...
+ * const app = new Hono()
+ *
+ * app.get('/', (c) => c.text('Hi'))
+ *
+ * fire(app)
+ * ```
+ */
+const fire = (app: Hono): void => {
+  // @ts-expect-error addEventListener is not typed well
+  addEventListener('fetch', handle(app))
```
rajsite: Some of the use-cases discussed in the issue were leveraging the service worker adapter for additional environments ([cloudflare-workers](https://github.com/honojs/hono/issues/3127#issuecomment-2456912838) and webassembly [via componentizejs](https://github.com/bytecodealliance/ComponentizeJS?tab=readme-ov-file#using-starlingmonkeys-fetch-event)).

I don't know for sure for cloud-flare workers but I know for sure in the webassembly componentizejs environment that the [default fallback of `handle` to trigger a fetch](https://github.com/honojs/hono/blob/main/src/adapter/service-worker/handler.ts#L20) results in the wasm module making another fetch ([see discussion on issue](https://github.com/honojs/hono/issues/3127#issuecomment-2843529188)). That default makes sense in a service worker but not a web server.

If `fire` here is intended to replace the [base fire](https://github.com/honojs/hono/blob/c18c2b2e0992d15e530a69c7db6b9ccbd6f06e61/src/hono-base.ts#L506-L520) behavior (note does not fallback to fetch) / is for web server use cases I'd instead pass undefined explicitly like I had to do for wasm componentizejs use cases before switching to base fire (example from internal repo):
image

if `fire` should be used for both service worker and web server cases I'd make sure to passthrough the handle options and put a warning for web servers switching from `base fire` to `service worker fire` that the behavior is different and likely need to go from `fire()` to `fire({fetch: undefined})`.

src/adapter/service-worker/index.ts
```diff
@@ -2,4 +2,29 @@
...
+}
```
rajsite: For compatibility with `base fetch` (and a good default for web servers but maybe not a good default for service workers):

```suggestion
const fire = (app: Hono): void => {
  // @ts-expect-error addEventListener is not typed well
  addEventListener('fetch', handle(app. {fetch: undefined}))
}
```

For usage in either web worker or web server, but a potentially unexpected default for web server:

```suggestion
const fire = (app: Hono, opts: HandleOpts): void => {
  // @ts-expect-error addEventListener is not typed well
  addEventListener('fetch', handle(app, opts))
}
```

rajsite: another idea could be to have separate methods like:
- `fire()`: which is compatible with `base fire` and passes `handle(app, {fetch: undefined})`
- and `fireWithFallback(opts)`: which can be configure for either case and uses the service worker default fallback fetch behavior

yusukebe: @rajsite Thank you for the comment.

I like this way: https://github.com/honojs/hono/pull/4214/commits/5be78838669ea6b26caea987dcf6ffa7efadeca6

rajsite: Tested the branch locally and works great! Thanks for the discussions and consideration!
image

src/adapter/service-worker/index.ts
```diff
@@ -2,4 +2,36 @@
...
+ * fire(app)
+ * ```
+ */
+const fire = (
+  app: Hono,
+  options: HandleOptions = {
+    fetch: undefined,
+  }
+): void => {
+  // @ts-expect-error addEventListener is not typed well
```
rajsite:
```suggestion
  // @ts-expect-error addEventListener is not typed well in ServiceWorker-like contexts, see: https://github.com/microsoft/TypeScript/issues/14877
```

yusukebe: @yusukebe 

Nice! Updated: https://github.com/honojs/hono/pull/4214/commits/1886afd8a302d828e59a981a963df5de1351bebb

yusukebe:
Hi @rajsite, can you review this again?

yusukebe:
@rajsite Thanks!

PR #4216, opened 2025-06-14
feat: enable to add nonce to script tag generated by Suspense and ErrorBoundary

### LINKED ISSUE #4205: CSP support for Suspense and ErrorBoundry

I'm very impressed with the Suspense and ErrorBoundry JSX, but when I tried to use them with CSP nonce.

The `<script>` generated from Suspense and ErrorBoundry cause error in browser related to CSP directive, like this.

Image

```
Refused to execute inline script because it violates the following 
Content Security Policy directive: "script-src 'nonce-LMCyRCoWBRS9OwJpEHW3eA=='".
Either the 'unsafe-inline' keyword, a hash ('sha256-UDh/om4Btj1xwKrEmuiP7pSO6PW9jlWtV0HXeqLIQto='),
or a nonce ('nonce-...') is required to enable inline execution.
```

From the source, I found the script tag is hard-coded and doesn't have ability to insert nonce attribute, like CSS ones and `Script` in honox.

Suspense
https://github.com/honojs/hono/blob/4a1dd5f66b2b30d7576ea15171423a51b3ed1a59/src/jsx/streaming.ts#L85-L88

ErrorBoundary
https://github.com/honojs/hono/blob/4a1dd5f66b2b30d7576ea15171423a51b3ed1a59/src/jsx/components.ts#L84-L87

### Expected

Both components work fine under the Content-Security-Policy header like `script-src 'nonce-LMCyRCoWBRS9OwJpEHW3eA=='; style-src 'nonce-LMCyRCoWBRS9OwJpEHW3eA=='`

### Related

* #3694

usualoma:
Fixes #4205

It may be a bit exaggerated, but how about writing it as follows?

```tsx
<StreamingContext value={{ scriptNonce: 'test-nonce' }}>
  <Suspense fallback={<p>Loading...</p>}>
    <Content />
  </Suspense>
</StreamingContext>
```

I considered a proposal that could be specified in attributes, but rejected it for the following reasons.

* I don't want to make my own extensions to React's API.
* When there are many `Suspense` nodes, I don't want to specify attributes multiple times.

```tsx
<Suspense fallback={<p>Loading...</p>} nonce="test-nonce">
  <Content />
</Suspense>
```

- [x] Add tests
- [x] Run tests


usualoma:
Hi, @yusukebe, what do you think?

yusukebe (approved):
LGTM!

yusukebe:
Hi @usualoma 

This is good! I wondered if it might be better to consider the other attributes rather than `nonce`, but I can't imagine the `script` should have another for `Suspense`. If so, we can add it later.

Issue #3572, opened 2024-10-27
Custom headers with JWT


noobnooc:
### What is the feature you are proposing?

I am building something with APN (Apple Push Notification) and it needs a token signed by JWT with some headers and payload. I noticed that the JWT related helpers in Hono do not support for custom headers but only fixed `{alg, typ}`, so I cannot sign the APN token with JWT helpers in Hono.

I finally implemented it by myself, it's not too complex. But I still hope to be able to implement it with Hono, it will be more convenient.

kunalbhagawati:
Added a PR for this: https://github.com/honojs/hono/pull/4218

kunalbhagawati:
PR is merged 
https://github.com/honojs/hono/pull/4218

PR #4218, opened 2025-06-14
feat(jwt): Add custom header location

kunalbhagawati:
Add a custom header location to the JWT middleware. 
Follows from https://github.com/honojs/hono/issues/3572.

---

## Possible Alternatives

I wasn't sure how much would be too much, so I stuck to the current changes for now.

But we could also explore other alternatives:

### Function or string
```typescript

export const jwt = (options: {
  secret: SignatureKey
  cookie?:
    | string
    | { key: string; secret?: string | BufferSource; prefixOptions?: CookiePrefixOptions }
  alg?: SignatureAlgorithm
  header?: string | (() => string | Promise<string>)  // allows more flexibility
```

### Token string getter function 
We can take a function to let the user decide how to get the token. That might play well with the `bearerAuth` middleware as well.
```typescript

export const jwt = (options: {
  secret: SignatureKey
  cookie?:
    | string
    | { key: string; secret?: string | BufferSource; prefixOptions?: CookiePrefixOptions }
  alg?: SignatureAlgorithm
  header?: () => string

  // We can skip the "token parsing from the header" part if the user gives this.
  tokenFetcherFunction: () => string | Promise<string>
})
```

---

- [x] Add tests
- [x] Run tests

_NOTE_ `bun run format:fix && bun run lint:fix` changes too many files, thus, did not push those.
```
        modified:   runtime-tests/lambda-edge/index.test.ts
        modified:   src/adapter/service-worker/handler.test.ts
        modified:   src/helper/css/common.case.test.tsx
        modified:   src/helper/css/common.ts
        modified:   src/helper/html/index.test.ts
        modified:   src/jsx/dom/css.ts
        modified:   src/middleware/jwt/index.test.ts
        modified:   src/middleware/jwt/jwt.ts
        modified:   src/middleware/secure-headers/index.test.ts
        modified:   src/utils/accept.test.ts
```

### DIFF
--- a/src/middleware/jwt/jwt.ts
+++ b/src/middleware/jwt/jwt.ts
@@ -48,12 +48,12 @@ export type JwtVariables<T = any> = {
  * ```
  */
 export const jwt = (options: {
-  secret: SignatureKey,
+  secret: SignatureKey
   cookie?:
     | string
     | { key: string; secret?: string | BufferSource; prefixOptions?: CookiePrefixOptions }
-  alg?: SignatureAlgorithm,
-  headerName?: string,
+  alg?: SignatureAlgorithm
+  headerName?: string
 }): MiddlewareHandler => {
   if (!options || !options.secret) {
     throw new Error('JWT auth middleware requires options for "secret"')
@@ -65,7 +65,7 @@ export const jwt = (options: {
 
   return async function jwt(ctx, next) {
     const headerName = options.headerName || 'Authorization'
-    
+
     const credentials = ctx.req.raw.headers.get(headerName)
     let token
     if (credentials) {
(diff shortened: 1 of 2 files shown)


src/middleware/jwt/jwt.ts
```diff
@@ -46,11 +48,12 @@ export type JwtVariables<T = any> = {
...
  */
 export const jwt = (options: {
-  secret: SignatureKey
+  secret: SignatureKey,
   cookie?:
     | string
     | { key: string; secret?: string | BufferSource; prefixOptions?: CookiePrefixOptions }
-  alg?: SignatureAlgorithm
+  alg?: SignatureAlgorithm,
+  header?: string,
```
yusukebe: I think `headerName` is better than `header` with reference to the following Bearer Auth middleware.

https://github.com/honojs/hono/blob/882e75a3e00e98997177aaaef60bbf9b1b4eb3bf/src/middleware/bearer-auth/index.ts#L32

yusukebe:
Hi @kunalbhagawati 

Thank you for the PR - this is a good PR. I've left a comment regarding the name of the option.

> I wasn't sure how much would be too much, so I stuck to the current changes for now.
> But we could also explore other alternatives:

I like your ideas, but it's okay not to add them currently.

src/middleware/jwt/jwt.ts
kunalbhagawati: Agreed, and changed :)

yusukebe:
@kunalbhagawati 

It looks good! The CI is failing, but it's not related to this PR. I'll merge this later. Thank you for your contribution!

kunalbhagawati:
Thanks for accepting this! :)

PR #4236, opened 2025-06-19
fix(jsx): Fix the JSXNode validation. Allow the function type for props.ref

yayugu:

- [ ] Add tests
- [x] Run tests

According to the [React document](https://react.dev/reference/react-dom/components/common#ref-callback), `ref` should accept the function and it properly implemented. However, the validation in `toStringBuffer()` is not reflect that and causes false positive error. This PR fixes that.

1. I don't have time to implement tests for now, Feel free to add or just close my PR and create the independent fix.
2. Also, I don't checked other validations issues, maybe there is more

### DIFF
--- a/src/jsx/base.ts
+++ b/src/jsx/base.ts
@@ -217,7 +217,7 @@ export class JSXNode implements HtmlEscaped {
         buffer[0] += ` ${key}="`
         buffer.unshift('"', v)
       } else if (typeof v === 'function') {
-        if (!key.startsWith('on')) {
+        if (!key.startsWith('on') && key !== 'ref') {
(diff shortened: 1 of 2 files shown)


yusukebe:
@yayugu Thank you for the PR. Can you add a test? There is no rush.

usualoma:
Hi @yayugu, Thank you.
I think this is a good fix 👍

yusukebe (approved):
LGTM!

yusukebe:
Hi @yayugu 

I've added the test myself. It looks good. Thank you for your contribution!

yayugu:
Awesome! Thank you so much for your help 🎉

Issue #4233, opened 2025-06-18
Cannot transpile hono using esbuild due to dynamic import of cloudflare:workers when using inspectRoutes


vaderkos:
### What version of Hono are you using?

4.8.0

### What runtime/platform is your app running on? (with version if possible)

Node.js 24.0.2

### What steps can reproduce the bug?

**Description:**
As I'm using Node I do not need `cloudflare:workers` dependency. My application is transpiled using esbuild.
Using inspectRoutes fails transpilation beacuse of dynamic import.

**Steps:**
Try transpiling specified app code using specified esbuild configuration.

### What is the expected behavior?

`inspectRoutes` from `hono/dev` does not dynamically import `cloudflare:workers`. App should be transpilable when using `hono/dev`.

### What do you see instead?

```
Could not resolve "cloudflare:workers"

    ../../node_modules/.pnpm/hono@4.8.0/node_modules/hono/dist/utils/color.js:9:120:
      9 │ ...= "Cloudflare-Workers" ? "NO_COLOR" in ((await import("cloudflare:workers")).env ?? {}) : !getColorEnabled();
        ╵                                                          ~~~~~~~~~~~~~~~~~~~~

  You can mark the path "cloudflare:workers" as external to exclude it from the bundle, which will
  remove this error and leave the unresolved path in the bundle. You can also add ".catch()" here to
  handle this failure at run-time instead of bundle-time.
```

### Additional information

As a workaround I specified `cloudflare:workers` as an external dependency of `esbuild`. 

Esbuild configuration:
```ts
import { resolve } from 'node:path'
import { cwd } from 'node:process'

import { build } from 'esbuild'

const srcDirPath = resolve(cwd(), 'src')
const distDirPath = resolve(cwd(), 'dist')

const entrypoint = resolve(srcDirPath, 'index.ts')
 await build({
  format: 'esm',
  bundle: true,
  minify: true,
  outdir: distDirPath,
  outbase: srcDirPath,
  platform: 'node',
  metafile: true,
  sourcemap: true,
  keepNames: false,
  mainFields: ['module', 'main'],
  treeShaking: true,
  entryPoints: [entrypoint],
  sourcesContent: false,
  // external: ['cloudflare:workers'],
})
```
Transpiled file:
```ts
import { serve } from '@hono/node-server'
import { Hono } from 'hono'
import { inspectRoutes } from 'hono/dev'

const app = new Hono()

app.all('*', (c) => c.text(`Hello ${Date.now()}`))

[...]

PR #4239, opened 2025-06-20
fix(utils/color): avoid resolving pacakages via Bun.build

ryuapp:
follow up https://github.com/honojs/hono/issues/4233#issuecomment-2990039940

[Bun.build](https://bun.sh/docs/bundler) is smarter than other bundlers and when minifying it will try to resolve a package name, which will result in an error.

~~This PR will switch `cloudflare:workers` variable name declaration from `const` to `let` declaration to prevent the problem.~~
This PR use `async/await` to work around a bug in opennext temporary: https://github.com/honojs/hono/issues/4233#issuecomment-2992430971 Proper opennext support will be prepared later.

- [x] Add tests
- [x] Run tests


yusukebe (approved):
LGTM!

yusukebe:
@ryuapp 

Thank you for the quick work!

PR #4241, opened 2025-06-21
fix(jsx): throw new Error instead of string

### LINKED ISSUE #4235: `onError` handler couldn't handle Hono JSX errors. Also no error messages on console

4.8.0

Node 20

I found the `onError()` hasn't triggered when there is a Hono JSX error on the server side. I debugged and found the potential design inconsistency between the dispatcher and Hono JSX.

## Reproduction code
I created the repo.
https://github.com/yayugu/hono-bug-report/tree/master

The specific error code is here:
https://github.com/yayugu/hono-bug-report/blob/348c20e98fb32a84b120691f0695582d4ff13f8c/src/index.tsx#L14

```
pnpm i
pnpm dev
```

then access to https://localhost:3000

1. Return proper 500 error page to users
2. Log proper error message / stack traces to stdout.

### What do you see instead?

The blank 500 response and no stdout messages.

### Additional information

As I mentioned before, this is caused by the potential design inconsistency between the dispatcher and Hono JSX.

### Design 1. Hono JSX throws the non Error type, string object.
For my case, this line throws string.
https://github.com/honojs/hono/blob/946fcf488f3909b1f5444f70f236f0c5db78636b/src/jsx/base.ts#L221

### Design 2. Hono dispatcher only accepts Error types
https://github.com/honojs/hono/blob/946fcf488f3909b1f5444f70f236f0c5db78636b/src/compose.ts#L53

### As a result
There is no way to handle / even logging the error. This is hard to debug on the local and impossible to debug on production server.

This should be fixed by:

A. Accepts non Error type (or add the option for that)
B. Hono JSX throws Error type

usualoma:
fixes #4235

We should throw an `Error` instance instead of a string.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/jsx/base.ts
+++ b/src/jsx/base.ts
@@ -209,7 +209,7 @@ export class JSXNode implements HtmlEscaped {
         }
       } else if (key === 'dangerouslySetInnerHTML') {
         if (children.length > 0) {
-          throw 'Can only set one of `children` or `props.dangerouslySetInnerHTML`.'
+          throw new Error('Can only set one of `children` or `props.dangerouslySetInnerHTML`.')
         }
 
         children = [raw(v.__html)]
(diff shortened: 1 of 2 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@usualoma Thanks!

PR #4242, opened 2025-06-21
feat(bun/ws): allow to import `upgradeWebSocket` and `websocket` directly

nakasyou:
I've just realized that WebSocket helper didn't require `createBunWebSocket` to provide functions used to upgrade connections. 
This PR allows to write code like the code:
```ts
import { upgradeWebSocket, websocket } from 'hono/bun'

app.get('/ws', upgradeWebSocket(...))

export default {
  fetch: app.fetch,
  websocket
}
```
Codes using a WebSocket with Bun will become simpler than before.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/adapter/bun/websocket.test.ts
+++ b/src/adapter/bun/websocket.test.ts
@@ -1,6 +1,6 @@
 import { Context } from '../../context'
 import type { BunWebSocketData, BunServerWebSocket } from './websocket'
-import { createWSContext, createBunWebSocket } from './websocket'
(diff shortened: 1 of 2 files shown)


yusukebe (approved):
LGTM!

yusukebe:
Hi @nakasyou 

Looks good to me! I'll include this change in the next minor release `v4.9` since this is a `feat` change, and making the method deprecated. If we should hurry, can you tell me so?

nakasyou:
Hi @yusukebe. Thank you for reviewing. I don't hurry so releasing with `v4.9` is good.

PR #4244, opened 2025-06-22
fix(jwt): fix the `JwtTokenIssuedAt` error message

### LINKED ISSUE #4210: JwtTokenIssuedAt error has a garbled message

4.7.8

Node 22.9 (but it's not relevant)

Make a request using the JWT middleware with an invalid `iat` claim.

There should be an attractive and grammatically correct error message, e.g. `JwtTokenIssuedAt: Invalid "iat" claim, must be a valid number lower than "2000000" (iat: "2000000000")`.

### What do you see instead?

You'll see an unclear message similar to the following: `JwtTokenIssuedAt: Incorrect "iat" claim must be a older than "2000000" (iat: "2000000000")`.

### Additional information

This is the relevant line of code: https://github.com/honojs/hono/blob/c18c2b2e0992d15e530a69c7db6b9ccbd6f06e61/src/utils/jwt/types.ts#L36

yusukebe:
Fixes #4210

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/utils/jwt/types.ts
+++ b/src/utils/jwt/types.ts
@@ -33,7 +33,9 @@ export class JwtTokenExpired extends Error {
 
 export class JwtTokenIssuedAt extends Error {
   constructor(currentTimestamp: number, iat: number) {
-    super(`Invalid "iat" claim, must be a valid number lower than "${currentTimestamp}" (iat: "${iat}")`)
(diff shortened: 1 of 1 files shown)


yusukebe:
I think we can merge this. Thanks!

PR #4250, opened 2025-06-27
feat(aws-lambda): specify content-type as binary

Kanahiro:

- [x] Add tests
- [x] Run tests

## Proposal

- add `contentTypesAsBinary` to spefify which content-type should be treated as binary (will be base64 encoded)

```typescript
// content-type: application/vnd.mapbox-vector-tile will be base64-encoded and response as binary.
handle(app, { contentTypesAsBinary: ['application/vnd.mapbox-vector-tile'] })
```

## motivation

Current implementation treats some pre-defined content-types as "should be base64-encoded".
However, sometimes we need to treat some types other than them as binary (e.g. `'application/vnd.mapbox-vector-tile` https://github.com/mapbox/vector-tile-spec ).

What do yo think?

### DIFF
--- a/src/adapter/aws-lambda/handler.ts
+++ b/src/adapter/aws-lambda/handler.ts
@@ -162,11 +162,16 @@ export const streamHandle = <
   )
 }
 
+type HandleOptions = {
+  contentTypesAsBinary: string[]
+}
+
 /**
  * Accepts events from API Gateway/ELB(`APIGatewayProxyEvent`) and directly through Function Url(`APIGatewayProxyEventV2`)
  */
 export const handle = <E extends Env = Env, S extends Schema = {}, BasePath extends string = '/'>(
-  app: Hono<E, S, BasePath>
+  app: Hono<E, S, BasePath>,
+  { contentTypesAsBinary }: HandleOptions
 ): (<L extends LambdaEvent>(
   event: L,
   lambdaContext?: LambdaContext
@@ -189,7 +194,7 @@ export const handle = <E extends Env = Env, S extends Schema = {}, BasePath exte
       lambdaContext,
     })
 
-    return processor.createResult(event, res)
+    return processor.createResult(event, res, { contentTypesAsBinary })
   }
 }
 
@@ -231,13 +236,22 @@ export abstract class EventProcessor<E extends LambdaEvent> {
     return new Request(url, requestInit)
   }
 
-  async createResult(event: E, res: Response): Promise<APIGatewayProxyResult> {
+  async createResult(
+    event: E,
+    res: Response,
+    options: { contentTypesAsBinary: string[] } = { contentTypesAsBinary: [] }
+  ): Promise<APIGatewayProxyResult> {
     const contentType = res.headers.get('content-type')
     let isBase64Encoded = contentType && isContentTypeBinary(contentType) ? true : false
 
     if (!isBase64Encoded) {
       const contentEncoding = res.headers.get('content-encoding')
       isBase64Encoded = isContentEncodingBinary(contentEncoding)
+    }
+
+    if (!isBase64Encoded) {
+      // fallback: treat specific content-type as binary to be base64-encoded
+      isBase64Encoded = options.contentTypesAsBinary.includes(contentType ?? '')
     }
 
     const body = isBase64Encoded ? encodeBase64(await res.arrayBuffer()) : await res.text()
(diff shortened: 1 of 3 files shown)


src/adapter/aws-lambda/handler.ts
```diff
@@ -231,7 +236,11 @@ export abstract class EventProcessor<E extends LambdaEvent> {
     return new Request(url, requestInit)
   }
 
-  async createResult(event: E, res: Response): Promise<APIGatewayProxyResult> {
+  async createResult(
+    event: E,
+    res: Response,
+    options: { contentTypesAsBinary: string[] } = { contentTypesAsBinary: [] }
```
yusukebe: Using the type definition in `HandleOptions` for `contentTypesAsBinary` is better.

```ts
options: Pick<HandleOptions, 'contentTypesAsBinary'> = { contentTypesAsBinary: [] }
```

src/adapter/aws-lambda/handler.ts
```diff
@@ -162,11 +162,16 @@ export const streamHandle = <
   )
 }
 
+type HandleOptions = {
+  contentTypesAsBinary: string[]
```
yusukebe: How about adding the following type or replacing `string[]`?

`(contentType: string) => boolean) | undefined`

This means the `HandleOptions` will be:

```ts
type HandleOptions = {
  contentTypesAsBinary: string[] | ((contentType: string) => boolean) | undefined
}
```

Or

```ts
type HandleOptions = {
  contentTypesAsBinary: ((contentType: string) => boolean) | undefined
}
```

You can use it like this:

```ts
const result = await processor.createResult(event, response, {
  contentTypesAsBinary: (contentType) => {
    return /application\/custom\-(a|b|c)/.test(contentType)
  },
})
```

This can handle the `contentType` more flexibly. I like passing a "callback" style for options 😃

yusukebe:
Thanks @Kanahiro 

I think this is a good feature. I've left comments.

Hi @watany-dev, can you review this if you have time?

src/adapter/aws-lambda/handler.ts
Kanahiro: Thanks! I prefer this one.

```typescript
type HandleOptions = {
  contentTypesAsBinary: ((contentType: string) => boolean) | undefined
}
```

Kanahiro:
@yusukebe
Thank you advising. I refactored codes as to use `isContentTypeBinary` as default and overwrite with user-defined one if provided. I think this is simpler and more consistent.

Current interface:

```typescript
// content-type: application/vnd.mapbox-vector-tile will be base64-encoded and response as binary.
handle(app, { isContentTypeBinary: (contentType) => contentType === 'application/vnd.mapbox-vector-tile' })
```

src/adapter/aws-lambda/handler.ts
```diff
@@ -490,6 +501,12 @@ const isProxyEventV2 = (event: LambdaEvent): event is APIGatewayProxyEventV2 =>
   return Object.hasOwn(event, 'rawPath')
 }
 
+/**
+ * Check if the given content type is binary.
+ * This is a default function and may be overwritten by the user via `isContentTypeBinary` option in handler().
+ * @param contentType The content type to check.
+ * @returns True if the content type is binary, false otherwise.
+ */
 export const isContentTypeBinary = (contentType: string) => {
```
yusukebe: This is my idea. How about exporting this function, such as `defaultIsContentTypeBinary`, for users who want to add additional content types to the default values. The diff:

```diff
diff --git a/src/adapter/aws-lambda/index.ts b/src/adapter/aws-lambda/index.ts
index 74d0a09a..bbeec0d7 100644
--- a/src/adapter/aws-lambda/index.ts
+++ b/src/adapter/aws-lambda/index.ts
@@ -3,7 +3,7 @@
  * AWS Lambda Adapter for Hono.
  */

-export { handle, streamHandle } from './handler'
+export { handle, streamHandle, isContentTypeBinary as defaultIsContentTypeBinary } from './handler'
 export type { APIGatewayProxyResult, LambdaEvent } from './handler'
 export type {
   ApiGatewayRequestContext,
```

The usage:

```ts
const result = await processor.createResult(event, response, {
  isContentTypeBinary: (contentType: string) => {
    if (defaultIsContentTypeBinary(contentType)) {
      return true
    }
    return contentType === 'application/additional-custom-header'
  },
})
```

Kanahiro:
@yusukebe Thank you for comment, I completely agree with it and fixed some codes.

yusukebe (approved):
LGTM!

yusukebe:
Hi @Kanahiro 

Looks good to me! This is a new feature so that we will include it in the next minor release. However, it's just for adding the option, so it may be included in a patch release. I'll decide it and merge this later. Thank you for the contribution!

PR #4253, opened 2025-06-29
feat(jwt): added validation for the issuer (`iss`) claim

yolocat-dev:
This adds validation for the issuer claim (`iss`),
with support for simple strings and complex Regex patterns.

As this only verifies when the user has added an expected issuer, this isn't a breaking change and won't impact usage. You can however opt-in to issuer verification by simply setting the `issuer` option to your expected issuer.

- [x] Added tests
- [x] Ran tests - no errors
- [x] Ran `bun run format:fix` to format the code
- [x] Added [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code

I didn't run `bun run lint:fix` as there was existing errors, however I ran `bun run lint` and it returned `47 errors` both before and after my commit.

### DIFF
--- a/src/utils/jwt/jwt.ts
+++ b/src/utils/jwt/jwt.ts
@@ -15,6 +15,7 @@ import {
   JwtTokenExpired,
   JwtTokenInvalid,
   JwtTokenIssuedAt,
+  JwtTokenIssuer,
   JwtTokenNotBefore,
   JwtTokenSignatureMismatched,
 } from './types'
@@ -71,7 +72,8 @@ export const sign = async (
 export const verify = async (
   token: string,
   publicKey: SignatureKey,
-  alg: SignatureAlgorithm = 'HS256'
+  alg: SignatureAlgorithm = 'HS256',
+  issuer?: string | RegExp
 ): Promise<JWTPayload> => {
   const tokenParts = token.split('.')
   if (tokenParts.length !== 3) {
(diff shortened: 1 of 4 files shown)


yolocat-dev:
Codecov report looks good to me, the new error type is tested the same way as the others

yolocat-dev:
[`honojs/website#693`](https://github.com/honojs/website/pull/693) has the documentation changes for my PR

src/utils/jwt/jwt.ts
```diff
@@ -71,7 +72,8 @@ export const sign = async (
 export const verify = async (
   token: string,
   publicKey: SignatureKey,
-  alg: SignatureAlgorithm = 'HS256'
+  alg: SignatureAlgorithm = 'HS256',
+  issuer?: string | RegExp
```
yusukebe: Do you think it will be necessary to verify `sub`, `aud`, and others in the future? If so, the number of arguments will increase accordingly. Therefore, I think it would be better to combine the options into a single object.

```ts
type VerifyOptions = {
  alg?: SignatureAlgorithm
  issuer?: string | RegExp
  audience?: string | string[]
  subject?: string
  //...
}
```

And support both the new and old option patterns by using overload.

```ts
export const verify = async (
  token: string,
  publicKey: SignatureKey,
  algOrOptions?: SignatureAlgorithm | VerifyOptions
): Promise<JWTPayload> => {
  const options = typeof algOrOptions === 'string' ? { alg: algOrOptions } : algOrOptions || {}
}
```

What do you think of it?

yolocat-dev: This looks great, should I implement this or do you want to do that? I could take a look at some other claims and add them accordingly.

Edit: I'll implement it either way ;)

yusukebe: @yolocat-dev 

> Edit: I'll implement it either way ;)

Please!

yusukebe:
Hi @yolocat-dev !

Looks good to me! This is a new feature so that it can be included in the next minor release `4.9.0`. Since it may be a bit far and this feature is not a big change, we may release it as a patch release. I'll decide it.

Thank you for the contribution!

PR #4257, opened 2025-06-30
fix(jsx): cloneElement didn't copy children

yayugu:

When I use Hono JSX with [html-react-parser](https://github.com/remarkablemark/html-react-parser/blob/a71f4dc78b8a0aed88bb692704d480a46ac83cde/src/dom-to-react.ts#L53), I realized `cloneElement` didn't copy children.

I wrote fix and tests.

I only changed `jsx` (server side) implementation and  haven't checked `jsx/dom`. That might also have the issue.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/jsx/base.test.tsx
+++ b/src/jsx/base.test.tsx
@@ -1,0 +1,30 @@
+/** @jsxImportSource ./ */
+
+import type { JSXNode } from './base'
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import { cloneElement, jsx, Fragment } from './base'
+
+describe('cloneElement', () => {
+  it('should clone an element with new props', () => {
+    const element = <div className='original'>Hello</div>
+    const clonedElement = cloneElement(element, { className: 'cloned' })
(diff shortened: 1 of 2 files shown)


usualoma:
Hi @yayugu, thanks for the suggestion. It was definitely a bug. Your fix should get it working as expected!

The `jsx/dom` side looks fine as follows.
https://github.com/honojs/hono/blob/5149aa421bef92d3a495287f2f67bca4382bad39/src/jsx/dom/index.ts#L70

Regarding the code, I would like to adjust it slightly to match Hono's existing code, but first, I would like to merge this PR and then add the adjustments.

yusukebe:
@yayugu Thank you!

@usualoma Can we merge this now?

usualoma:
Hi @yusukebe

Yes, I think it's okay to merge.
I may create a refactoring PR later, but I think the current content is of sufficient quality to merge.

yusukebe:
Hi @usualoma 

Okay! I'll merge this now. Thank you.

PR #4267, opened 2025-07-02
fix(client): remove `index` string when calling `$url()`

### LINKED ISSUE #4265: In Hono client, removeIndexString should also be applied when calling $url

4.8.3

Node 24

I'm suing the Hono client's `$url` to generate my URL. For the most part, its working perfectly but the generation of the `index` URL omits to remove the "index".

For example:
```ts
 hc<MyApp>("http://fake/").index.$url().href
```
generates `http://fake/index` where it should be `http://fake/`.

On the other side:
```ts
 hc<MyApp>("http://fake/").index.$get(...)
```
will properly fetch the `http://fake` URL.

I think I identified the issue. While for the "fetch" methods (like `$get`), the `index` string is removed:
https://github.com/honojs/hono/blob/407d1bdf64048c4876b883bc1b5cf270e12791e7/src/client/client.ts#L102

it is omitted for the `$url` method:
https://github.com/honojs/hono/blob/407d1bdf64048c4876b883bc1b5cf270e12791e7/src/client/client.ts#L159-L169

_No response_

### What do you see instead?

_No response_

### Additional information

_No response_

yusukebe:
Fixes #4265

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -339,6 +339,42 @@ describe('Basic - query, queries, form, path params, header and cookie', () => {
   })
 })
 
+describe('Basic - $url()', () => {
+  const api = new Hono().get('/', (c) => c.text('API')).get('/posts/:id', (c) => c.text('Post'))
+  const content = new Hono().get(
+    '/search',
+    validator('query', () => {
+      return { page: '1', limit: '10' }
+    }),
+    (c) => c.text('Search')
+  )
+  const app = new Hono()
+    .get('/', (c) => c.text('Index'))
+    .route('/api', api)
+    .route('/content', content)
+
+  it('Should return a correct url via $url().href', async () => {
+    const client = hc<typeof app>('http://fake')
+    expect(client.index.$url().href).toBe('http://fake/')
+    expect(client.api.$url().href).toBe('http://fake/api')
+    expect(
+      client.api.posts[':id'].$url({
+        param: {
+          id: '123',
+        },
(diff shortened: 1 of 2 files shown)


src/client/client.test.ts
```diff
@@ -339,6 +339,42 @@ describe('Basic - query, queries, form, path params, header and cookie', () => {
   })
 })
 
+describe('Basic - $url()', () => {
```
yusukebe, line 342: I added more tests for `$url`.

jraoult:
@yusukebe Brilliant, pretty much the patch I implemented temporarily. Thanks for the quick turnaround!

yusukebe:
@jraoult Thanks!

jraoult:
@yusukebe I realise today that there was an oversight with this PR. For example, if `/index` is suffixed with a query parameter, `removeIndexString` doesn't work because of the Regexp expecting `/index` to be at the end of the tested string: 

https://github.com/honojs/hono/blob/31f2406336d2a6e475c590ed594832f2799c0273/src/client/utils.ts#L55-L60 

Typically:

```ts
client.content.index.$url({
        query: {
          page: '123',
          limit: '20',
        },
      }).href
```

will contain the `/index` string.

Apologies for not catching that during the review 😬. FYI, this is what I do in the meantime to fix it:

```ts
url.pathname.replace(/\/index$/, "") + url.search
```

PR #4279, opened 2025-07-11
feat(jwk): Add custom `headerName` to JWK middleware

JoaquinGimenez1:
RFC: Enable JWK middleware to read from custom header, similar to the cookie. 

This will provider more flexibility for different use cases like Cloudflare Zero Trust Access that sends the assertion token on `Cf-Access-Jwt-Assertion` header.

- [x] Add tests
- [x] Run tests


src/middleware/jwk/jwk.ts
```diff
@@ -45,6 +49,7 @@ export const jwk = (
     cookie?:
       | string
       | { key: string; secret?: string | BufferSource; prefixOptions?: CookiePrefixOptions }
+    authorization?: string
```
yusukebe: I think `headerName` is better instead of `authorization`, as it is the same as JWT middleware:
https://github.com/honojs/hono/blob/9e42fd1c9d895a655c4fb1ed82af9d2f2cb4abe7/src/middleware/jwt/jwt.ts#L56

yusukebe:
@JoaquinGimenez1 

Thank you for the PR. Can you add proper tests?

src/middleware/jwk/jwk.ts
JoaquinGimenez1: Will do! That sounds good! Thanks

JoaquinGimenez1:
Will add tests and open a proper PR, thanks for the feedback @yusukebe

JoaquinGimenez1:
Okay, all should be good to go now.

src/middleware/jwk/index.test.ts
```diff
@@ -434,6 +434,170 @@ describe('JWK', () => {
...
+      const req = new Request('http://localhost/auth-with-keys/a')
+      req.headers.set('Authorization', `Bearer ${credential}`)
+      const res = await app.request(req)
+      expect(res).not.toBeNull()
+      expect(res.status).toBe(401)
+      expect(await res.text()).toBe('Unauthorized')
+      expect(handlerExecuted).toBeFalsy()
+    })
+
+    it.only('Should authorize', async () => {
```
yusukebe: Remove `only`.

pnpm-lock.yaml
yusukebe, line 1: Remove this file.

src/middleware/jwk/index.test.ts
```diff
@@ -434,6 +434,170 @@ describe('JWK', () => {
...
+      const payload = c.get('jwtPayload')
+      return c.json(payload)
+    })
+    app.get('/auth-with-keys-nested/*', (c) => {
+      handlerExecuted = true
+      const payload = c.get('jwtPayload')
+      return c.json(payload)
+    })
+
+    it('Should not authorize', async () => {
```
yusukebe, line 454: I think you don't add so many tests to confirm this PR works. It's okay to add only a failed pattern without a custom header and a success pattern with a custom header.

JoaquinGimenez1: I see, I was checking how we were testing the JWT middleware with custom headers, will remove some of them

yusukebe:
Hi @JoaquinGimenez1 

Thanks! Looks good to me. It's a new feature, so basically it will be introduced in the next minor version, but it's a slight change. So, it may be included in the patch release. I'll decide it later.

PR #4291, opened 2025-07-20
feat(serve-static): use `join` to correct path resolution

yusukebe:

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/middleware/serve-static/index.test.ts
+++ b/src/middleware/serve-static/index.test.ts
@@ -1,3 +1,4 @@
+import path, { join } from 'node:path'
 import { Hono } from '../../hono'
 import { serveStatic as baseServeStatic } from '.'
 
@@ -12,11 +13,11 @@ describe('Serve Static Middleware', () => {
 
   const serveStatic = baseServeStatic({
     getContent,
-    pathResolve: (path) => {
-      return `./${path}`
-    },
+    join,
     isDir: (path) => {
-      return path === 'static/hello.world'
+      if (path === 'static/sub' || path === 'static/hello.world') {
+        return true
+      }
     },
     onFound: (path, c) => {
       if (path.endsWith('hello.html')) {
@@ -36,36 +37,29 @@ describe('Serve Static Middleware', () => {
     expect(res.status).toBe(200)
     expect(res.headers.get('Content-Encoding')).toBeNull()
     expect(res.headers.get('Content-Type')).toMatch(/^text\/html/)
-    expect(await res.text()).toBe('Hello in ./static/hello.html')
-    expect(res.headers.get('X-Custom')).toBe('Found the file at ./static/hello.html')
+    expect(await res.text()).toBe('Hello in static/hello.html')
+    expect(res.headers.get('X-Custom')).toBe('Found the file at static/hello.html')
   })
 
   it('Should return 200 response - /static/sub', async () => {
     const res = await app.request('/static/sub')
     expect(res.status).toBe(200)
     expect(res.headers.get('Content-Type')).toMatch(/^text\/html/)
-    expect(await res.text()).toBe('Hello in ./static/sub/index.html')
-  })
-
-  it('Should return 200 response - /static/helloworld', async () => {
-    const res = await app.request('/static/helloworld')
-    expect(res.status).toBe(200)
-    expect(res.headers.get('Content-Type')).toMatch(/^text\/html/)
-    expect(await res.text()).toBe('Hello in ./static/helloworld/index.html')
+    expect(await res.text()).toBe('Hello in static/sub/index.html')
   })
 
   it('Should return 200 response - /static/hello.world', async () => {
     const res = await app.request('/static/hello.world')
     expect(res.status).toBe(200)
     expect(res.headers.get('Content-Type')).toMatch(/^text\/html/)
-    expect(await res.text()).toBe('Hello in ./static/hello.world/index.html')
+    expect(await res.text()).toBe('Hello in static/hello.world/index.html')
   })
 
   it('Should decode URI strings - /static/%E7%82%8E.txt', async () => {
     const res = await app.request('/static/%E7%82%8E.txt')
     expect(res.status).toBe(200)
     expect(res.headers.get('Content-Type')).toMatch(/^text\/plain/)
-    expect(await res.text()).toBe('Hello in ./static/炎.txt')
+    expect(await res.text()).toBe('Hello in static/炎.txt')
   })
 
   it('Should return 404 response - /static/not-found.txt', async () => {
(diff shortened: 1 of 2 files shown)


src/adapter/cloudflare-workers/serve-static.test.ts
```diff
@@ -79,7 +79,8 @@ describe('ServeStatic Middleware', () => {
     expect(res.headers.get('Content-Type')).toBe('text/plain; charset=utf-8')
   })
 
-  it('Should return index.html', async () => {
+  // Serve static on Cloudflare Workers cannot determine whether the target path is a directory or not
+  it.skip('Should return index.html', async () => {
```
yusukebe, line 83: In the previous implementation, if the file was not found, it checked whether `index.html` existed. This PR, that process has been removed.

Since serve static for Cloudflare Workers is deprecated and not planned to be used, this change is acceptable.

yusukebe:
Hey @usualoma !

What do you think of this? If it makes sense for you, please review it!

usualoma:
Hi @yusukebe 
Thanks for all your hard work! Just give me a few days to review everything.

yusukebe:
@usualoma Okay!

src/middleware/serve-static/path.ts
```diff
@@ -0,0 +1,71 @@
+const isWindowsPath = (path: string): boolean => {
+  return /^[a-zA-Z]:/.test(path) || path.includes('\\')
+}
+
+export const defaultJoin = (...paths: string[]): string => {
+  if (paths.length === 0) return '.'
```
usualoma:
```suggestion
  if (paths.length === 0) {
    return '.'
  }
```

Until now, at hono, we have always used `{` with `if`, so I think it would be better to use it here as well.

usualoma:
Hi @yusukebe 

### Reasons for not using ‘node:path’ in deno

I think the reason is probably because we haven't used modules in the "node:" namespace in this file before, but is that correct? (We have already used them in bun.)

If that is the case, as mentioned below, `join` processing is difficult, so I think it would be safer to depend on `nodo:path` even in Deno.

### I think it's better to always use `/` for separators.

Currently, if an attacker inserts `\` into the request string, the entire string will be joined with `\` , which may cause the path restriction to fail. In a Windows environment, `/` should be joined as expected, so I think we should always use `/` .

```ts
const app = new Hono()

app.get('/', (c) => {
  return c.text('Hello Hono!')
})

app.use('/static/*', serveStatic({ root: './' }))

export default app
```

```bash
% curl http://0.0.0.0:8000/static/test%5Ctest.txt
```

When this app is run in a Linux environment, it will search for the following file.

```
static\test\test.txt
```

In a Windows environment, this file is located in the `static` directory, but in a Linux environment, it is located in the current directory as `static\test\test.txt`.

src/middleware/serve-static/path.ts
yusukebe: You are right.

src/adapter/deno/serve-static.ts
```diff
@@ -1,3 +1,4 @@
+import { join } from 'node:path'
```
yusukebe, line 1: We can use [@std/path](https://jsr.io/@std/path/doc), but it's better to use `node:path` because it's built into Deno.

src/middleware/serve-static/index.test.ts
```diff
@@ -36,36 +35,29 @@ describe('Serve Static Middleware', () => {
...
   })
 
   it('Should return 200 response - /static/sub', async () => {
     const res = await app.request('/static/sub')
     expect(res.status).toBe(200)
     expect(res.headers.get('Content-Type')).toMatch(/^text\/html/)
-    expect(await res.text()).toBe('Hello in ./static/sub/index.html')
-  })
-
-  it('Should return 200 response - /static/helloworld', async () => {
```
yusukebe, line 50: This test is the same as `Should return 200 response - /static/sub`?? We may remove this.

yusukebe:
Hi @usualoma Thank you for the accurate comment!

For Deno, as you said, we can use `node:path`'s `join`. Updated: https://github.com/honojs/hono/pull/4291/commits/b3738f7241d281e4820893dc71ef6b217132e41f

For the separators issue, I made `defaultJoin` support only Linux (posix), and it does not support Windows. This means if runtime-depends serve static-middleware (`adapter/bun/serve-static.ts` or `adapter/deno/serve-static.ts`) does not specify `path` option, it can't handle Windows paths. But, I think it's okay to cut Windows support because these runtime-dependent serve static-middleware can use `join` in the runtime API.

What do you think of this?

yusukebe:
Hi @usualoma 

If you already realized this, sorry for that. I updated this PR. Can you review this?

usualoma:
Hi @yusukebe 

Thanks for the update. The content makes sense!

As it stands, the frequency of defaultJoin usage is very low, so I think it's important to keep the code size small, so I think it would be better to do the following.

```diff
diff --git i/src/middleware/serve-static/path.ts w/src/middleware/serve-static/path.ts
index b8ae7986..0253cf5d 100644
--- i/src/middleware/serve-static/path.ts
+++ w/src/middleware/serve-static/path.ts
@@ -3,39 +3,23 @@
  * If you need Windows path support, please use `join` exported from `node:path` etc. instead.
  */
 export const defaultJoin = (...paths: string[]): string => {
-  if (paths.length === 0) {
-    return '.'
-  }
-  if (paths.length === 1 && paths[0] === '') {
-    return '.'
-  }
-
   // Join non-empty paths with '/'
   let result = paths.filter((p) => p !== '').join('/')
 
   // Normalize multiple slashes to single slash
-  result = result.replace(/\/+/g, '/')
+  result = result.replace(/(?<=\/)\/+/g, '')
 
   // Handle path resolution (. and ..)
   const segments = result.split('/')
   const resolved = []
-  const isAbsolute = result.startsWith('/')
 
   for (const segment of segments) {
-    if (segment === '' || segment === '.') {
-      continue
-    }
-    if (segment === '..') {
-      if (resolved.length > 0 && resolved[resolved.length - 1] !== '..') {
-        resolved.pop()
-      } else if (!isAbsolute) {
-        resolved.push('..')
-      }
-    } else {
+    if (segment === '..' && resolved.length > 0 && resolved.at(-1) !== '..') {
+      resolved.pop()
+    } else if (segment !== '.') {
       resolved.push(segment)
     }
   }
 
-  const final = resolved.join('/')
-  return isAbsolute ? '/' + final : final || '.'
+  return resolved.join('/') || '.'
 }
```

yusukebe:
@usualoma Thanks! That's good. I applied the patch: https://github.com/honojs/hono/pull/4291/commits/f3ee780a3541b3dc9699ac728fa44ee8422350af

Does this look good to you?

usualoma:
Hi @yusukebe
Thank you. I think it's good!

yusukebe:
@usualoma Thanks!

PR #4292, opened 2025-07-21
feat(adapter/vercel): make it possible to read `process.env` as `c.env`

EdamAme-x:
While building a web app with Hono and Next.js, I found myself wanting to access environment variables in a way that better aligns with Hono’s philosophy.  

However, to do that, I had to create an adapter with a patch similar to this one.  

After trying it out, I thought it might actually be quite useful, so I created this PR.  

Since other adapters also receive `env` in the same way, this approach should work fine with Vercel as well.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/adapter/vercel/handler.test.ts
+++ b/src/adapter/vercel/handler.test.ts
@@ -48,7 +48,7 @@ describe('Adapter for Next.js', () => {
       return c.json([
         {
           name: 'yusukebe',
-          message: c.env.MESSAGE
+          message: c.env.MESSAGE,
         },
       ])
     })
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @EdamAme-x 

Does `env` from Adapter Helper do enough for you?

https://hono.dev/docs/helpers/adapter

EdamAme-x:
Ah, I see. That could actually meet my needs

yusukebe:
@EdamAme-x Yeah. I think that is a Hono-like way.

PR #4297, opened 2025-07-23
fix(context): encode the redirect location

### LINKED ISSUE #4295: Context.redirect doesn't accept a URL which contains multibyte characters

4.8.5

Node v20.19.2

```
app.get("/", (c) => {
  return c.redirect("https://ja.wikipedia.org/wiki/小麦粉");
});
```

Here is the reproduction repo: https://github.com/yayugu/hono-bug-report/tree/2025-07-22-redirect

```bash
$ pnpm i
$ pnpm dev
# Then access to http://localhost:3000
```

Just redirect to the page.

### What do you see instead?

The error output:
```
TypeError: Cannot convert argument to a ByteString because the character at index 30 has a value of 23567 which is greater than 255.
    at webidl.converters.ByteString (node:internal/deps/undici/undici:3889:17)
    at _Headers.set (node:internal/deps/undici/undici:8921:35)
    at Context.header (file:///home/yayugu/hono-bug-report/node_modules/.pnpm/hono@4.8.5/node_modules/hono/dist/context.js:100:15)
    at Context.redirect (file:///home/yayugu/hono-bug-report/node_modules/.pnpm/hono@4.8.5/node_modules/hono/dist/context.js:167:10)
    at Array.<anonymous> (/home/yayugu/hono-bug-report/src/index.tsx:7:12)
    at #dispatch (file:///home/yayugu/hono-bug-report/node_modules/.pnpm/hono@4.8.5/node_modules/hono/dist/hono-base.js:188:38)
    at fetch (file:///home/yayugu/hono-bug-report/node_modules/.pnpm/hono@4.8.5/node_modules/hono/dist/hono-base.js:214:26)
    at Server.<anonymous> (file:///home/yayugu/hono-bug-report/node_modules/.pnpm/@hono+node-server@1.17.1_hono@4.8.5/node_modules/@hono/node-server/dist/index.mjs:487:13)
    at Server.emit (node:events:524:28)
    at parserOnIncoming (node:_http_server:1141:12)
```

#### Necessity
I found this during implementing the dynamic redirect, for example:

```javascript
const paramA = c.req.param("paramA");
c.redirect(`/new-url/${paramA}`);
```

Type check didn't help and found the bug after released the code. It's easy to wrap with encodeURI(), by it would super nice if redirect() function cares the encoding.

#### Cause and Design of the Fix
Node's Header object doesn't accept multibyte characters https://github.com/nodejs/node/issues/42579.

Context.redirect() is currently just by pass the value. It should accept any valid URL and encode that if necessary.
The complicated thing is if the path / URI is already encoded, it must not re-encode that. The pseudo code would be

```javascript
if (path.hasNonASCII() {
  path = encodeURI(path);
}
header.set(path);
```

yayugu:
Fixes #4295

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -20,6 +20,7 @@ import type {
   JSONValue,
   SimplifyDeepArray,
 } from './utils/types'
+import { safeEncodeURI } from './utils/url'
 
 type HeaderRecord =
   | Record<'Content-Type', BaseMime>
@@ -282,24 +283,6 @@ const setDefaultContentType = (contentType: string, headers?: HeaderRecord): Hea
   return {
     'Content-Type': contentType,
     ...headers,
-  }
-}
-
-/**
- * This function ensures that the string is safely encoded,
- * and not double-encoded.
- * The input string may or may not escaped by encodeURI.
- */
-const safeEncodeURI = (str: string) => {
-  try {
-    // decodeURI will throw if str contains %
-    // In that case, we can assume that str is not encoded
-    return encodeURI(decodeURI(str))
-  } catch (e) {
-    if (e instanceof URIError) {
-      return encodeURI(str)
-    }
-    throw e
   }
 }
 
--- a/src/context.ts
+++ b/src/context.ts
@@ -285,6 +285,24 @@ const setDefaultContentType = (contentType: string, headers?: HeaderRecord): Hea
   }
 }
 
+/**
+ * This function ensures that the string is safely encoded,
+ * and not double-encoded.
+ * The input string may or may not escaped by encodeURI.
+ */
+const safeEncodeURI = (str: string) => {
+  try {
+    // decodeURI will throw if str contains %
+    // In that case, we can assume that str is not encoded
+    return encodeURI(decodeURI(str))
+  } catch (e) {
+    if (e instanceof URIError) {
+      return encodeURI(str)
+    }
+    throw e
+  }
+}
+
 export class Context<
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   E extends Env = any,
@@ -746,7 +764,7 @@ export class Context<
     location: string | URL,
     status?: T
   ): Response & TypedResponse<undefined, T, 'redirect'> => {
-    this.header('Location', String(location))
+    this.header('Location', safeEncodeURI(String(location)))
     return this.newResponse(null, status ?? 302) as any
   }
 
--- a/src/context.test.ts
+++ b/src/context.test.ts
@@ -115,18 +115,6 @@ describe('Context', () => {
     )
   })
 
-  it('c.redirect() w/ unencoded, but URI invalid ASCII chars', async () => {
-    const res = c.redirect('https://example.com/%hello')
-    expect(res.headers.get('Location')).toBe('https://example.com/%25hello')
-  })
-
-  it('c.redirect() w/ encoded string', async () => {
-    const res = c.redirect('https://example.com/%E3%81%93%E3%82%93%E3%81%AB%E3%81%A1%E3%81%AF?abc')
-    expect(res.headers.get('Location')).toBe(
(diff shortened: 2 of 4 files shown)


src/context.ts
```diff
@@ -285,6 +285,24 @@ const setDefaultContentType = (contentType: string, headers?: HeaderRecord): Hea
   }
 }
 
+/**
+ * This function ensures that the string is safely encoded,
+ * and not double-encoded.
+ * The input string may or may not escaped by encodeURI.
+ */
+const safeEncodeURI = (str: string) => {
```
yusukebe: I was a little unsure about this, but I think it would be better to move `safeEncodeURI` to `src/utils/url.ts`. It is only used in `src/contex.ts`, so it is not necessary, but since all other functions related to URL/URI are in that file, it would be better.

In that case, writing unit tests for `safeEncodeURI` in `src/utils/url.test.ts` will be good.

src/context.ts
```diff
@@ -285,6 +285,24 @@ const setDefaultContentType = (contentType: string, headers?: HeaderRecord): Hea
...
+  try {
+    // decodeURI will throw if str contains %
+    // In that case, we can assume that str is not encoded
+    return encodeURI(decodeURI(str))
```
yusukebe: I think this implementation is good. Using the exception makes the code shorter.

src/context.ts
yayugu: Cool, I don't have any preference of that, moved

yusukebe (approved):
LGTM!

yusukebe:
@yayugu 

Looks good to me! I'll merge this and release a new version. Thank you!

NamesMT:
Hi @yusukebe , @yayugu 

The `safeEncodeURI` does not seems to be working perfectly.
I'm experiencing double encoding with the redirectUrl returned by [`msal-node`](https://github.com/AzureAD/microsoft-authentication-library-for-js) (Microsoft EntraID auth)

Raw url: `https://login.microsoftonline.com/redacted-id&scope=email%20openid%20profile%20offline_access&redirect_uri=https%3A%2F%2Fredacted-domain.org%2Fapi%2Fauth%2Fcallback&client-request-id=redacted-id&response_mode=query&client_info=1&x-client-SKU=msal.js.node&x-client-VER=3.6.4&x-client-OS=linux&x-client-CPU=x64&response_type=code&code_challenge=7C75MO8zdoJZn7_sIpxcFEQzPvzWIMzD2tUf-lqevR0&code_challenge_method=S256`

After `safeEncodeURI`: `https://login.microsoftonline.com/redacted-id&scope=email%20openid%20profile%20offline_access&redirect_uri=https%253A%252F%252Fredacted-domain.org%252Fapi%252Fauth%252Fcallback&client-request-id=redacted-id&response_mode=query&client_info=1&x-client-SKU=msal.js.node&x-client-VER=3.6.4&x-client-OS=linux&x-client-CPU=x64&response_type=code&code_challenge=7C75MO8zdoJZn7_sIpxcFEQzPvzWIMzD2tUf-lqevR0&code_challenge_method=S256`

The `redirect_uri` is double encoded and breaks the site's login.

yusukebe:
@NamesMT  thank you for the report. Exactly, it does not work correctly.

yusukebe:
@yayugu @NamesMT I'm working on it.

anthonymclaughlin:
I'm still getting double encoding after updating to v4.8.8.

Calling this redirect URL:

`return c.redirect('https://api.workos.com/user_management/authorize?client_id=1234&provider=authkit&redirect_uri=https%3A%2F%2Faccount.localhost.direct%3A8081%2Fapi%2Fauth%2Fcallback&response_type=code', 302);`

results in `redirect_url` being double encoded.

This code worked in v4.8.5.

yusukebe:
@anthonymclaughlin Thank you for the report. I'll work on fixing it.

yusukebe:
Hey @anthonymclaughlin 

It's fixed with the latest `4.8.9`.

PR #4299, opened 2025-07-23
fix(context/utils): redirect uri encoding (fixes #4295)

### LINKED ISSUE #4295: Context.redirect doesn't accept a URL which contains multibyte characters

patkub:

- [x] Add tests
- [x] Run tests

### Explanation

This makes the redirect function handle URI encoding. 

Express.js does not have this issue, and they use the https://github.com/pillarjs/encodeurl package to handle URI encoding. This implements the same code from pillarjs/encodeurl directly  in hono.

### DIFF
--- a/src/context.test.ts
+++ b/src/context.test.ts
@@ -108,6 +108,12 @@ describe('Context', () => {
     expect(res.headers.get('Location')).toBe('https://example.com/destination')
   })
 
+  it('c.redirect() w/ URL encoded', async () => {
+    const res = c.redirect(new URL('/小麦粉', 'https://example.com'))
+    expect(res.status).toBe(302)
+    expect(res.headers.get('Location')).toBe('https://example.com/%E5%B0%8F%E9%BA%A6%E7%B2%89')
+  })
+
   it('c.header()', async () => {
     c.header('X-Foo', 'Bar')
     const res = c.body('Hi')
--- a/src/context.ts
+++ b/src/context.ts
@@ -20,6 +20,7 @@ import type {
   JSONValue,
   SimplifyDeepArray,
 } from './utils/types'
+import { encodeNonEncodedURI } from './utils/url'
 
 type HeaderRecord =
   | Record<'Content-Type', BaseMime>
@@ -746,7 +747,7 @@ export class Context<
     location: string | URL,
     status?: T
   ): Response & TypedResponse<undefined, T, 'redirect'> => {
-    this.header('Location', String(location))
+    this.header('Location', encodeNonEncodedURI(String(location)))
     return this.newResponse(null, status ?? 302) as any
   }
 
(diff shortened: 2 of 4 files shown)


yusukebe:
Hi @patkub 

Thank you for the PR! But we will prioritize #4297, which was created at the same time. Because the implementation of #4297 is shorter and fits with Hono's policy. I appreciate your contribution.

PR #4302, opened 2025-07-24
feat(jwt): Exposing utility function `verifyWithJwks` for external use

Beyondo:
Hello! I recently needed to use our JWK middleware in a SvelteKit SSR page, but since it's not handled through Hono's Router in this case, having the currently internal `verifyFromJwks` on its own (without the `jwk({...})` middleware) turned out to be useful, which allows us to get the same exact functionality _without the router._

Only two line changes:

`src/middleware/jwt/jwt.ts`
```diff
+export const verifyFromJwks = Jwt.verifyFromJwks
export const verify = Jwt.verify
export const decode = Jwt.decode
export const sign = Jwt.sign
```
`src/middleware/jwt/index.ts`
```diff
-export { jwt, verify, decode, sign } from './jwt'
+export { jwt, verifyFromJwks, verify, decode, sign } from './jwt'
```

**Edit Update:** `verifyFromJwks` -> `verifyWithJwks`

Example:
```ts
import { verifyWithJwks } from "hono/jwt";
```
```ts
const id_payload = await verifyWithJwks(id_token, { // throws error on failure
  jwks_uri: env.AUTH_JWKS_URL,
}, {
  cf: { cacheEverything: true, cacheTtl: 3600 }
});
// successfully verified, proceeding...
```

`verifyWithJwks` would be another exposed utility function just like `verify` but specialized for JWK keys.


Beyondo:
Since we're now exposing it publicly, I do now also think that `verifyWithJwks` would be a more semantically appropriate name than `verifyFromJwks`. "From" implies you're only fetching _from_ a remote resource like `jwks_uri`, but we also have `keys` involved that you also verify _with._ **e.g**. You open a door _with keys,_ not from keys.

This PR would still within the constraints of a simple semver patch since that utility function cannot be used outside of hono yet so it's safe and non-breaking to rename it.

yusukebe (approved):
LGTM!

yusukebe:
Hi @Beyondo 

Looks good to me. Renaming is also good. This will be included in the next minor release (not so far). Thanks!

PR #4306, opened 2025-07-25
fix(utils/url): prevent double encoding in `safeEncodeURI`

yusukebe:
Fixes https://github.com/honojs/hono/pull/4297#issuecomment-3117351523

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/utils/url.ts
+++ b/src/utils/url.ts
@@ -318,11 +318,13 @@ export const decodeURIComponent_ = decodeURIComponent
  */
 export const safeEncodeURI = (str: string): string => {
   try {
-    // decodeURI will throw if str contains %
-    // In that case, we can assume that str is not encoded
-    return encodeURI(decodeURI(str))
+    // Try to decode the string
+    const decoded = decodeURI(str)
+    // If successful, encode only if it's different from the original
+    return decoded === str ? encodeURI(str) : str
   } catch (e) {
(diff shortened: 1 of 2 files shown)


yusukebe:
Hey @NamesMT @yayugu 

Can you review this?

yusukebe:
@yayugu 

Thank you for reviewing and for the comment.

> but I don't have a good solution and it can be a users' responsibility to make the arg to state A or B.

Yeah. I also think it is the users' responsibility. Anyway, this PR will address the problem with the simple implementation. Let's go with this.

src/utils/url.ts
```diff
@@ -318,11 +318,13 @@ export const decodeURIComponent_ = decodeURIComponent
  */
 export const safeEncodeURI = (str: string): string => {
   try {
-    // decodeURI will throw if str contains %
-    // In that case, we can assume that str is not encoded
-    return encodeURI(decodeURI(str))
+    // Try to decode the string
+    const decoded = decodeURI(str)
+    // If successful, encode only if it's different from the original
+    return decoded === str ? encodeURI(str) : str
```
gabrielDonnantuoni, line 324: It looks like it should be `!==` instead of `===` here @yusukebe 
```suggestion
    // If successful, encode only if it's different from the original
    return decoded !== str ? encodeURI(str) : str
```

PR #4307, opened 2025-07-25
fix(utils/url): use `isByteString` in `safeEncodeURI`

yusukebe:

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/utils/url.test.ts
+++ b/src/utils/url.test.ts
@@ -310,24 +310,28 @@ describe('url', () => {
       expect(safeEncodeURI('こんにちは')).toBe('%E3%81%93%E3%82%93%E3%81%AB%E3%81%A1%E3%81%AF')
     })
 
-    it('Encode unencoded, but URI invalid ASCII chars', async () => {
-      expect(safeEncodeURI('https://example.com/%hello')).toBe('https://example.com/%25hello')
+    it('Keep strings with invalid ASCII chars', async () => {
+      expect(safeEncodeURI('https://example.com/%hello')).toBe('https://example.com/%hello')
     })
 
-    it('Do not change the encoded string', async () => {
+    it('Keep strings with an encoded string', async () => {
       expect(
         safeEncodeURI('https://example.com/%E3%81%93%E3%82%93%E3%81%AB%E3%81%A1%E3%81%AF?abc')
       ).toBe('https://example.com/%E3%81%93%E3%82%93%E3%81%AB%E3%81%A1%E3%81%AF?abc')
     })
 
-    it('Should be the same strings', async () => {
+    it('Keep strings with encoded :// unchanged', () => {
+      expect(safeEncodeURI('https://localhost/api?redirect_uri=https%3A%2F%2Fexample.com')).toBe(
+        'https://localhost/api?redirect_uri=https%3A%2F%2Fexample.com'
+      )
+    })
+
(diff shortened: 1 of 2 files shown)


src/utils/url.test.ts
```diff
@@ -310,24 +310,28 @@ describe('url', () => {
       expect(safeEncodeURI('こんにちは')).toBe('%E3%81%93%E3%82%93%E3%81%AB%E3%81%A1%E3%81%AF')
     })
 
-    it('Encode unencoded, but URI invalid ASCII chars', async () => {
-      expect(safeEncodeURI('https://example.com/%hello')).toBe('https://example.com/%25hello')
+    it('Keep strings with invalid ASCII chars', async () => {
+      expect(safeEncodeURI('https://example.com/%hello')).toBe('https://example.com/%hello')
```
yusukebe: The purpose of `safeEncodeURI` is to prevent the value set in the `Headers` object from causing an error, so invalid ASCII strings do not need to be converted.

src/utils/url.test.ts
```diff
@@ -310,24 +310,28 @@ describe('url', () => {
...
+      expect(safeEncodeURI('https://localhost/api?redirect_uri=https%3A%2F%2Fexample.com')).toBe(
+        'https://localhost/api?redirect_uri=https%3A%2F%2Fexample.com'
+      )
+    })
+
+    it('Keep strings with encoded spaces and :// unchanged', async () => {
       expect(
         safeEncodeURI(
-          'https://login.microsoftonline.com/redacted-id&scope=email%20openid%20profile%20offline_access&redirect_uri=https%3A%2F%2Fredacted-domain.org%2Fapi%2Fauth%2Fcallback&client-request-id=redacted-id&response_mode=query&client_info=1&x-client-SKU=msal.js.node&x-client-VER=3.6.4&x-client-OS=linux&x-client-CPU=x64&response_type=code&code_challenge=7C75MO8zdoJZn7_sIpxcFEQzPvzWIMzD2tUf-lqevR0&code_challenge_method=S256'
+          'https://localhost/api?redirect_uri=https%3A%2F%2Fexample.com&scope=email%20profile'
```
yusukebe: Simplified the strings.

src/utils/url.ts
yusukebe, line 1: The purpose this time is not to implement `safeEncodeURI` correctly, but to prevent `c.redirect()` from throwing an error. Therefore, there is no need to add logic to `utils/url.ts`.

src/context.ts
```diff
@@ -747,7 +746,13 @@ export class Context<
     location: string | URL,
     status?: T
   ): Response & TypedResponse<undefined, T, 'redirect'> => {
-    this.header('Location', safeEncodeURI(String(location)))
+    const locationString = String(location)
+    this.header(
+      'Location',
+      // Multibyes should be encoded
+      // eslint-disable-next-line no-control-regex
+      !/[^\x00-\xFF]/.test(locationString) ? locationString : encodeURI(locationString)
```
yusukebe, line 754: I decided to use a simple regular expression to determine multibyte characters.

yusukebe:
Since the priority this time is to revert the different behavior, I will merge without review. If there are any problems, I will fix them immediately.

yayugu:
This code looks clean and best, thank you for your following up!

PR #4313, opened 2025-07-27
feat: add `hcParse` util to smartly parse response from a `hc` fetch

### LINKED ISSUE #3894: Hono RPC Client: add option for response parsing

It would be great if the `options` of the `hc()` would allow to add a custom response parser to centralize common code that we want to be executed on every response, e.g.:

```ts
const client = hc<AppType>('', {
  handleResponse: async (res) => {
    if (!res.ok) throw Error("error")

   return await res.json()
  }
});
```

our current solution is to wrap every client call with a custom method that checks for errors and converts to json, e.g.
`parseResponse(await client.myEndpoint.$get())`

Would it be feasible to implement this as it will likely complicate typing the client methods?

NamesMT:
Somewhat resolves #3894

- [x] Add tests
- [x] Run tests
- [ ] Add [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code - I added some doc, but not fully semantics JSDoc.

Note: this PR adds [`fetch-result-please`](https://www.npmjs.com/package/fetch-result-please) as a dependency so the codebase doesn't have to maintain as much codes, also because we're at `/client` subpath, adding dependency won't really increase the package size of main hono.

Though, if you prefer to not add any new dependency, I have another PR for that alternative version: #4314.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -16,6 +16,7 @@
         "bun-types": "^1.1.39",
         "esbuild": "^0.15.18",
         "eslint": "^9.10.0",
+        "fetch-result-please": "^0.2.0",
         "glob": "^11.0.0",
         "jsdom": "^22.1.0",
         "msw": "^2.6.0",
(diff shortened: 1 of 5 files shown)


NamesMT:
Converted to draft because #4314 seems better.

PR #4314, opened 2025-07-27
feat: add `parseResponse` util to smartly parse `hc`'s Response

### LINKED ISSUE #3894: Hono RPC Client: add option for response parsing

NamesMT:
This PR adds a `hcParse` util that will automatically parse the response from a `hc` fetch, throwing a structured error if response is not `ok`, it is also type-safe.

Somewhat resolves #3894

- [x] Add tests
- [x] Run tests
- [ ] Add [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code - I added some doc, but not fully semantics JSDoc.

Note: this is an alternative PR for #4313, this PR does not add any dependency, ~however, do note that this PR do not have a graceful fail-safe for `json` parsing from [`destr`].~

This PR reimplements `fetch-result-please` minimally that is customized for Hono and integrates with `hcParse` better., do note that while the code have a different behavior compared to `fetch-result-please`/`ofetch` (removed `blob` response, `text` for undefined content-type header because we removed `destr`), but the UX should be the same and transparent for Hono's users because Hono always set the `content-type` for us 💪.

### DIFF
--- a/src/client/utils.test.ts
+++ b/src/client/utils.test.ts
@@ -191,7 +191,7 @@ describe('hcParse', async () => {
       })
     }),
     http.get('http://localhost/rawBuffer', () => {
-      return HttpResponse.text('hono', {
+      return HttpResponse.arrayBuffer(new TextEncoder().encode('hono'), {
         headers: {
           'content-type': 'x/custom-type',
         },
--- a/src/client/utils.test.ts
+++ b/src/client/utils.test.ts
@@ -176,7 +176,7 @@ describe('hcParse', async () => {
     http.get('http://localhost/404', () => {
       return HttpResponse.text('404 Not Found', { status: 404 })
     }),
-    http.get('http://localhost', () => {
+    http.get('http://localhost/raw', () => {
       return HttpResponse.text('hello', {
         headers: {
           'content-type': '',
--- a/src/client/utils.test.ts
+++ b/src/client/utils.test.ts
@@ -149,6 +149,11 @@ describe('hcParse', async () => {
   const app = new Hono()
     .get('/text', (c) => c.text('hi'))
     .get('/json', (c) => c.json({ message: 'hi' }))
+    .get('/404', (c) => c.notFound())
+    .get('/raw', (c) => {
+      c.header('content-type', '')
+      return c.body('hello')
+    })
 
   const client = hc<typeof app>('http://127.0.0.1:3301')
 
@@ -183,6 +188,16 @@ describe('hcParse', async () => {
       const result = await hcParse(await client.json.$get())
       expect(result).toEqual({ message: 'hi' })
       type _verify = Expect<Equal<typeof result, { message: string }>>
+    }),
+    it('should throw error when the response is not ok', async () => {
+      await expect(hcParse(client['404'].$get())).rejects.toThrowError('404 Not Found')
+    }),
+    it('should parse as text for raw responses without content-type header', async () => {
+      const result = await hcParse(client.raw.$get())
+      console.log((await client.raw.$get()).headers.get('content-type'))
+      expect(result).toBe('hello')
+      // TODO: uncomment next line after `body` response are typed
+      // type _verify = Expect<Equal<typeof result, 'hello'>>
     }),
   ])
 
--- a/src/client/utils.test.ts
+++ b/src/client/utils.test.ts
@@ -154,6 +154,14 @@ describe('hcParse', async () => {
       c.header('content-type', '')
       return c.body('hello')
     })
+    .get('/rawUnknown', (c) => {
+      c.header('content-type', 'x/custom-type')
+      return c.body('hello')
+    })
+    .get('/rawBuffer', (c) => {
+      c.header('content-type', 'x/custom-type')
+      return c.body(new TextEncoder().encode('hono'))
+    })
 
   const client = hc<typeof app>('http://127.0.0.1:3301')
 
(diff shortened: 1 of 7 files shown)


NamesMT:
I will do a refactor tomorrow on the inlined `fetchRP` util to make it more minimal for our use case (removing options interface, simpler detectResponseType), as well as adding more coverage.

NamesMT:
I have updated the implementation to be more minimal, I also adds some more tests and the new code coverage is also 100% locally, @codecov bot is missing the BASE report so its reporting a lower number.

src/client/utils.test.ts
```diff
@@ -1,6 +1,12 @@
+import { HttpResponse, http } from 'msw'
```
yusukebe, line 1: I updated the test to use `msw` instead of running the actual server. This is the same way as `src/client/client.test.ts`.

src/client/utils.ts
```diff
@@ -72,3 +75,27 @@ export function deepMerge<T>(target: T, source: Record<string, unknown>): T {
...
+ * Shortcut to get a consumable response from `hc` fetch calls, with types inference.
+ *
+ * Smartly parse the response data, and automatically throw an error if the response is not ok.
+ *
+ * To handle an error, see {@link DetailedError} interface.
+ *
+ * @example hcParse(client.posts.$get())
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export async function hcParse<T extends ClientResponse<any>>(
```
yusukebe: I like the name `parseResponse` simply. `parseResponse` uses the general word `response`, but the user can understand it as `hc` related because it's exported from `hono/client`.

```ts
import { hc, parseResponse } from 'hono/client'
```

src/client/utils_fetchRP.ts
```diff
@@ -0,0 +1,80 @@
...
+  ) {
+    super(message)
+    this.log = options.log
+    this.detail = options.detail
+    this.code = options.code
+    this.statusCode = options.statusCode
+  }
+}
+
+const JSON_RE = /^application\/(?:[\w!#$%&*.^`~-]*\+)?json(?:;.+)?$/i
```
yusukebe: I like the name  `jsonRegex` rather than `JSON_RE`.

src/client/utils_fetchRP.ts
```diff
@@ -0,0 +1,80 @@
+/* eslint @typescript-eslint/no-explicit-any: 0 */
+
+// References & credits: `fetch-result-please`, `ofetch`
+
+const nullBodyResponses = new Set([101, 204, 205, 304])
+
+/**
+ * Smartly parses and return the consumable result from a fetch `Response`.
+ */
+export async function fetchRP(fetchRes: Response | Promise<Response>): Promise<any> {
```
yusukebe: I know you named `fetchRP` picking from `fetch-result-please`, and I like the name `fetch-result-please`. But in this case, the name should be explanatory. I can't find a good name now, but like `fetchWithData`.

src/client/utils_fetchRP.ts
```diff
@@ -0,0 +1,80 @@
...
+    this.statusCode = options.statusCode
+  }
+}
+
+const JSON_RE = /^application\/(?:[\w!#$%&*.^`~-]*\+)?json(?:;.+)?$/i
+function detectResponseType(response: Response): 'json' | 'text' {
+  const _contentType = response.headers.get('content-type')
+
+  // TODO: Do we actually need this line?, Hono always help the user set the `content-type`, including default for `c.body()`
+  if (!_contentType) {
```
yusukebe: I think these lines are needed.

There is a case where `c.text()` does not set a `content-type`.

https://github.com/honojs/hono/blob/955bd8f72615e183991e2628ca56e898694115ac/src/context.ts#L686

```ts
const app = new Hono().get('/', (c) => c.text('Hi'))
console.log(res.headers.get('content-type')) // null
```

src/client/utils.test.ts
NamesMT: Okay, thanks!

src/client/utils.ts
NamesMT: Currently I named `hcParse` because it is shorter and the function is specifically designed to be used with `hc`'s `ClientResponse` only, so I do think `hcParse` is a good name, user can also  potentially discover it without reading changelog while trying to import `hc` 😁.

However, feel free to update the fn name if you think it fits Hono codebase and ecosystem better.

src/client/utils_fetchRP.ts
NamesMT: Its an internal-used function only so I didn't think to change it's naming previously, but you're right, it should be changed for future contributor reading, but I'm also not sure what to name it, but maybe `parseResponse` fits this util, or `assertParseResponse`, `parseFetchResponse`, `responseResult`, `fetchResult`?

src/client/utils_fetchRP.ts
NamesMT: Adjusted!

src/client/utils.ts
yusukebe: It's a bit of a pickle. The name `hcParse` isn't bad, but sometimes the user may not know "what" to parse. Then I think I'd prefer `parseResponse`.

src/client/utils_fetchRP.ts
yusukebe: Thanks!

src/client/utils_fetchRP.ts
yusukebe: > Its an internal-used function only so I didn't think to change it's naming previously

Ah, I see. Makes sense. It's okay for internal-used. So, this is good if you add link to https://github.com/NamesMT/fetch-result-please. If so, the contributor can understand the origin.

src/client/utils_fetchRP.ts
yusukebe, line 1: It's better to rename to `src/client/fetch-result-please.ts`, if we go with the method in this file, which is `fetchRP`.

NamesMT:
Hi @yusukebe, all pending reviews are resolved, I have rename the util to `fetch-result-please.ts` and add link, as well as rename `hcParse` to `parseResponse`

yusukebe (approved):
LGTM!

yusukebe:
@NamesMT 

Cool! Looks good to me! This is a new feature. So, I'll include this in the next minor release. I'm planning to release it soon. Thank you!

PR #4321, opened 2025-07-29
ci: add editorconfig-checker

3w36zj6:
cf. https://github.com/honojs/hono/pull/4312#issuecomment-3131128988

PR #4322, opened 2025-07-29
chore: autofix ci

### LINKED ISSUE #4320: Introduce autofix.ci
I found the autofix.ci, which is helpful for `honojs/hono`, `honojs/middleware`, and others in `honojs/*` project:

https://autofix.ci/

In these projects, the CI will throw an error if lint/format is not correct. The PR author can fix this, but it is tedious to request this every time. That's where autofix.io comes in handy.

BarryThePenguin:
fixes #4320

Once merged, I can add a similar workflow to https://github.com/honojs/middleware

### DIFF
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -1,0 +1,32 @@
+name: autofix.ci
+
+on:
+  pull_request:
+  push:
+    branches: [main]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  autofix:
+    name: autofix
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'push' || !github.event.pull_request.draft }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: '1.2.19'
+      - run: bun install
+      - run: bun run format:fix
+      - run: bun run lint:fix
+      - name: Apply fixes
+        uses: autofix-ci/action@v1
+        with:
+          commit-message: 'ci: apply automated fixes'
--- a/src/index.ts
+++ b/src/index.ts
@@ -14,7 +14,7 @@
  * ```
  */
 
-import { Hono } from './hono';
+import { Hono } from './hono'
 
 /**
  * Types for environment variables, error handlers, handlers, middleware handlers, and more.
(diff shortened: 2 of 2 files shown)


autofix-troubleshooter:
Hi! I'm the <a href="https://autofix.ci">autofix logo<strong>autofix.ci</strong></a> troubleshooter bot.

It looks like you correctly set up a CI job that uses the autofix.ci GitHub Action, but the [**autofix.ci GitHub App**](https://github.com/apps/autofix-ci/) has not been installed for this repository. This means that autofix.ci unfortunately does not have the permissions to fix this pull request. If you are the repository owner, please install the app and then restart the CI workflow! 😃

.github/workflows/autofix.yml
```diff
@@ -0,0 +1,32 @@
...
+    if: ${{ github.event_name == 'push' || !github.event.pull_request.draft }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: '1.1.39'
+      - run: bun install
+      - run: bun run format:fix
+      - run: bun run lint:fix
```
yusukebe, line 28: Is it a good idea to add `bun run editorconfig-checker -format github-actions` introduced in #4321?

yusukebe:
I added the autofix.io GitHub App into this repo:

CleanShot 2025-07-30 at 14 53 35@2x

.github/workflows/autofix.yml
BarryThePenguin: From what I can see `editorconfig-checker` only reports issues in PRs, it doesn't appear to have the ability to fix those issues 👀

yusukebe: Ah, you are right! So, it's not needed.

bun.lock
yusukebe, line 1: @BarryThePenguin 

[The CI failed](https://github.com/honojs/hono/actions/runs/16614542619/job/47110119463?pr=4322) because the `bun.lock` is updated (some library versions are mismatched with the current `main`).

I think autofix.ci update it: https://github.com/honojs/hono/pull/4322/commits/8631788e9a45c88938c43efaf5e49b28d08ad86e

Can we disable it?

BarryThePenguin: It's because the version of bun used in CI isn't compatible with the `lockfileVersion` in `bun.lock`

I did some investigation in #4323, but that resulted in `bun run lint` failing 😭 

I think the two options are

- Revert `bun.lock` to use a `lockfileVersion` compatible with v1.1.39, or
- Upgrade bun and deal with any other issues that arise

yusukebe:
@BarryThePenguin 

Thank you! autofix.ci seems to be good!

PR #4331, opened 2025-08-01
refactor(ssg): mark old hook options as deprecated

3w36zj6:
The legacy hook options can now be fully represented using the plugins interface. To prepare for their eventual removal, I believe we should encourage existing users of the old hook options to migrate to the plugins approach.

- [ ] Add tests
- [ ] Run tests


yusukebe:
Hi @3w36zj6 

This is a nice idea! Looks good. This is not a new feature, but marking `deprecated` is a bit important thing. So, we can include this into the next minor release. I'll create the branch for it and merge it into that. Thank you!

PR #4344, opened 2025-08-08
feat(parseResponse): set `DetailedError.name` (+ error tests)

NamesMT:
This PR simply set the `.name` for the `DetailedError` class, which is a good pratice, so user is able to do a fast equality check in code / search for the error in log.

I forgot to add it previously and just hit a case where I need it because I have `DetailedError` classes coming from multiple source and `instanceof` is a bit wonky.

I also added tests for cases where error is thrown.

Theoretically "breaking", but should be safe to include in next patch.

- [x] Add tests
- [x] Run tests
- [ ] Add [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the codefix

### DIFF
--- a/src/client/fetch-result-please.ts
+++ b/src/client/fetch-result-please.ts
@@ -57,6 +57,7 @@ export class DetailedError extends Error {
     options: { detail?: any; code?: any; statusCode?: number; log?: any } = {}
   ) {
     super(message)
+    this.name = 'DetailedError'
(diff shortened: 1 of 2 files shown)


yusukebe:
@NamesMT 

Ah, `name` is useful. Yes. We can include this change into the next patch release. Thanks!

PR #4346, opened 2025-08-10
fix(client): improve parseResponse type inference for conditional responses

xzedx:
## Summary
Fixed a type inference issue in `parseResponse` where conditional responses (routes that can return different response types based on conditions) were incorrectly typed as union types instead of just the success response type.

## The Problem
When a route can return multiple response types like this:
```typescript
.get('/api', (c) => {
  if (someCondition) {
    return c.json({ error: 'error' }, 500)  // error response
  }
  return c.json({ data: [...] }, 200)       // success response  
})
```

`parseResponse` was inferring the return type as `{ error: string } | { data: any[] }` but it should only be `{ data: any[] }` because parseResponse throws an error for non-OK responses.

## The Solution
- Used `Extract<>` utility type to filter out only success status code responses from union types
- Added fallback logic for regular responses that don't have explicit status codes
- Updated tests to verify the correct behavior

## Test Plan
- [x] Type inference now correctly returns only success response types for conditional routes
- [x] Existing functionality for regular routes remains unchanged
- [x] All type assertion tests pass

### DIFF
--- a/src/client/utils.test.ts
+++ b/src/client/utils.test.ts
@@ -150,6 +150,12 @@ describe('parseResponse', async () => {
   const app = new Hono()
     .get('/text', (c) => c.text('hi'))
     .get('/json', (c) => c.json({ message: 'hi' }))
+    .get('/json-conditional', (c) => {
+      if (Math.random() > 0.5) {
+        return c.json({ error: 'error' }, 500)
+      }
+      return c.json({ data: [{ id: 1 }, { id: 2 }] }, 200)
+    })
     .get('/404', (c) => c.notFound())
     .get('/500', (c) => c.text('500 Internal Server Error', 500))
     .get('/raw', (c) => {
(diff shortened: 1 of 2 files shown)


src/client/utils.test.ts
```diff
@@ -223,6 +235,22 @@ describe('parseResponse', async () => {
       expect(result).toEqual({ message: 'hi' })
       type _verify = Expect<Equal<typeof result, { message: string }>>
     }),
+    it('example with conditional response of type inference', async () => {
```
yusukebe: Is this test `example with conditional response of type inference` really needed?  I think this is not related to `parseResponse`?

yusukebe:
@xzedx Thank you for the PR! This is good.

Hey @NamesMT, can you review this too?

src/client/utils.test.ts
xzedx: makes sense! i've removed it.

NamesMT:
Hi, good issue, I'm looking into it, it seems like the current PR method requires the status code to be explicitly typed right now, changing `c.json({ data: [{ id: 1 }, { id: 2 }] }, 200)` => `c.json({ data: [{ id: 1 }, { id: 2 }] })` results in type error.

The type at `parseResponse` is also a bit hard to read, I'll try to make it a bit more readable.

Will report back soon.

PR #4348, opened 2025-08-11
fix(parseResponse): should not include error responses in result

NamesMT:
Continuing from #4346, because I did a lot of updates and I can't push to the existing PR so I created a new one, I added more test cases and use another approach that is cleaner and fixes all problems.

In summary: 
* The PR adds a new type util: `FilterClientResponseByStatusCode`, which is used by `parseResponse` to exclude error responses.  
* More test cases are also added to make sure `parseResponse` is correct in use cases where a route will also return error responses.
* Some minor adjustments are also done in this PR, like:
  * Use `c.text` instead of `c.notFound` for correct `TypedResponse` infer and test
  * Add `.buffer` chain for `/rawBuffer` return, this fixes a minor type issue in newest node version.

---

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/client/utils.ts
+++ b/src/client/utils.ts
@@ -1,5 +1,6 @@
 import { fetchRP, DetailedError } from './fetch-result-please'
 import type { ClientResponse, ObjectType } from './types'
+import type { SuccessStatusCode } from '../utils/http-status'
 
 export { DetailedError }
 
@@ -87,7 +88,15 @@ export function deepMerge<T>(target: T, source: Record<string, unknown>): T {
 export async function parseResponse<T extends ClientResponse<any>>(
   fetchRes: T | Promise<T>
 ): Promise<
-  T extends ClientResponse<infer RT, infer _, infer RF>
+  Extract<T, ClientResponse<any, SuccessStatusCode, any>> extends never
+    ? T extends ClientResponse<infer RT, infer _, infer RF>
+      ? RF extends 'json'
+        ? RT
+        : RT extends string
+        ? RT
+        : string
+      : never
+    : Extract<T, ClientResponse<any, SuccessStatusCode, any>> extends ClientResponse<infer RT, infer _, infer RF>
     ? RF extends 'json'
       ? RT
       : RT extends string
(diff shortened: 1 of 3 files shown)


NamesMT:
@yusukebe @xzedx 
Hi guys, please help me review the PR.

yusukebe:
@NamesMT Nice! Looks good to me.

@xzedx Can you review this too?

src/client/utils.ts
```diff
@@ -87,13 +92,23 @@ export function deepMerge<T>(target: T, source: Record<string, unknown>): T {
...
+        T,
+        Exclude<ContentfulStatusCode, ClientErrorStatusCode | ServerErrorStatusCode>
+      > extends ClientResponse<infer RT, infer _, infer RF>
     ? RF extends 'json'
       ? RT
       : RT extends string
       ? RT
       : string
-    : never
+    : undefined
```
xzedx, line 111: As you said in #4346, this is a little hard to read. Perhaps we could defined the type separately?

``` Typescript
// Helper types for parseResponse
type SuccessResponses<T extends ClientResponse<any, any, any>> = FilterClientResponses<T, Exclude<ContentfulStatusCode, ClientErrorStatusCode | ServerErrorStatusCode>>

type ExtractResponseData<T> = T extends ClientResponse<infer RT, infer _, infer RF>
  ? RF extends 'json'
    ? RT
    : RT extends string
    ? RT
    : string
  : undefined

type ParseResponseResult<T extends ClientResponse<any, any, any>> = SuccessResponses<T> extends never
  ? undefined
  : ExtractResponseData<SuccessResponses<T>>

```

NamesMT: That looks great!, though, I think the current type is good enough, the util is still very new, so I think we shouldn't add too many standalone types from it yet, let it sit stable and robust first before we split and export the underlying types, I only add the internal `Filter` type for now because it makes the code a lot more readable.

I will keep noted of this in case of any future updates.

yusukebe:
@NamesMT @xzedx 

Thank you both! I'll merge it now.

PR #4353, opened 2025-08-14
feat(csrf): Add modern CSRF protection with Fetch Metadata support

meck93:
### Add modern CSRF protection with Fetch Metadata support

**Why this change?**

Modern browsers support [Fetch Metadata headers (Sec-Fetch-Site)](https://web.dev/articles/fetch-metadata) which provide a more reliable way to detect cross-origin requests than traditional Origin header checking alone. This enhancement adds opt-in support for Fetch Metadata based CSRF protection while maintaining full backwards compatibility.

### References and Inspiration

The implementation is inspired by:
- the go std lib (net/http) implementation: https://github.com/golang/go/blob/5dac42363ba8281a3f4f08e03af2292b763adc38/src/net/http/csrf.go#L122-L163
- the go RFC https://github.com/golang/go/issues/73626#issue-3046320918

More references:
- https://web.dev/articles/fetch-metadata
- https://developer.mozilla.org/en-US/docs/Glossary/Fetch_metadata_request_header
- https://developer.mozilla.org/en-US/docs/Web/Security/Attacks/CSRF#fetch_metadata

- [x] Add tests
- [x] Run tests
- [x] Update the documentation -> Website PR: https://github.com/honojs/website/pull/721

### DIFF
--- a/src/middleware/csrf/index.ts
+++ b/src/middleware/csrf/index.ts
@@ -8,10 +8,18 @@ import { HTTPException } from '../../http-exception'
 import type { MiddlewareHandler } from '../../types'
 
 type IsAllowedOriginHandler = (origin: string, context: Context) => boolean
-type IsAllowedSecFetchSiteHandler = (secFetchSite: string, context: Context) => boolean
+
+const secFetchSiteValues = ['same-origin', 'same-site', 'none', 'cross-site'] as const
+type SecFetchSite = (typeof secFetchSiteValues)[number]
+
+const isSecFetchSite = (value: string): value is SecFetchSite =>
+  (secFetchSiteValues as readonly string[]).includes(value)
+
+type IsAllowedSecFetchSiteHandler = (secFetchSite: SecFetchSite, context: Context) => boolean
+
 interface CSRFOptions {
   origin?: string | string[] | IsAllowedOriginHandler
-  secFetchSite?: string | string[] | IsAllowedSecFetchSiteHandler
+  secFetchSite?: SecFetchSite | SecFetchSite[] | IsAllowedSecFetchSiteHandler
 }
 
 const isSafeMethodRe = /^(GET|HEAD)$/
@@ -115,6 +123,10 @@ export const csrf = (options?: CSRFOptions): MiddlewareHandler => {
   const isAllowedSecFetchSite = (secFetchSite: string | undefined, c: Context) => {
     if (secFetchSite === undefined) {
       // denied always when sec-fetch-site header is not present
+      return false
+    }
+    // type guard to check if the value is a valid SecFetchSite
+    if (!isSecFetchSite(secFetchSite)) {
       return false
     }
     return secFetchSiteHandler(secFetchSite, c)
--- a/src/middleware/csrf/index.test.ts
+++ b/src/middleware/csrf/index.test.ts
@@ -347,6 +347,14 @@ describe('CSRF by Middleware', () => {
         )
         expect(res.status).toBe(403)
       })
+
+      it('should block unknown values', async () => {
+        const res = await app.request(
+          'http://localhost/form',
+          buildSimplePostRequestData({ secFetchSite: 'any' })
+        )
+        expect(res.status).toBe(403)
+      })
     })
 
     describe('string[]', () => {


usualoma:
Hi @meck93

Thank you for your suggestion.
I think using `sec-fetch-site` is a good idea!

However, since the current origin-based determination does not pose any security issues, I don't think this PR, which completely changes the behavior as an option, is a good approach. Furthermore, with the implementation in this PR, cross-origin requests that are not related to CSRF attacks are also targeted, which I think exceeds the checks that should be performed by this middleware.

If we use `sec-fetch-site` for determination, we can expand the scope of allowed requests in a simpler and more robust way than the current (on main branch) approach. I believe the following implementation would be preferable.

https://github.com/honojs/hono/pull/4355

meck93:
Hi @usualoma 

Thanks for your feedback. 
I agree with your idea and the proposal to make the change less fundamental and more of an opt-in approach. 

I'll adjust my PR accordingly.

usualoma:
Hi @meck93,

Thanks for the update!

### I would like to remove the addition of OPTIONS from this PR.

This middleware is intended to check for CSRF attacks and reject requests that match the following simple requests.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS#simple_requests

With this background, we specified GET and HEAD as trusted methods.

I agree with the idea of changing this middleware to treat OPTIONS as a trusted method in the future. However, this change could result in requests that were previously rejected being allowed, so I would like to discuss this in a separate PR.

usualoma:
Hi @meck93, thank you for your effort. It looks good to me!

usualoma:
Hi @yusukebe

I think it's okay to merge this, but what do you think?

src/middleware/csrf/index.ts
```diff
@@ -8,8 +8,10 @@ import { HTTPException } from '../../http-exception'
 import type { MiddlewareHandler } from '../../types'
 
 type IsAllowedOriginHandler = (origin: string, context: Context) => boolean
+type IsAllowedSecFetchSiteHandler = (secFetchSite: string, context: Context) => boolean
 interface CSRFOptions {
   origin?: string | string[] | IsAllowedOriginHandler
+  secFetchSite?: string | string[] | IsAllowedSecFetchSiteHandler
```
yusukebe: The `string` should be only `same-origin`, `same-site`, `none`, or `cross-site`, right? If so, (this is a matter of test), you can define types like the following to improve the DX:

```ts
const secFetchSiteValues = ['same-origin', 'same-site', 'none', 'cross-site'] as const
type SecFetchSite = (typeof secFetchSiteValues)[number]

// ...

interface CSRFOptions {
  origin?: string | string[] | IsAllowedOriginHandler
  secFetchSite?: SecFetchSite | SecFetchSite[] | IsAllowedSecFetchSiteHandler
}
```

CleanShot 2025-08-18 at 17 38 40@2x

One of example codes is like the following:

```diff
diff --git a/src/middleware/csrf/index.ts b/src/middleware/csrf/index.ts
index 6ce35862..70970a34 100644
--- a/src/middleware/csrf/index.ts
+++ b/src/middleware/csrf/index.ts
@@ -8,10 +8,18 @@ import { HTTPException } from '../../http-exception'
 import type { MiddlewareHandler } from '../../types'

 type IsAllowedOriginHandler = (origin: string, context: Context) => boolean
-type IsAllowedSecFetchSiteHandler = (secFetchSite: string, context: Context) => boolean
+
+const secFetchSiteValues = ['same-origin', 'same-site', 'none', 'cross-site'] as const
+type SecFetchSite = (typeof secFetchSiteValues)[number]
+
+const isSecFetchSite = (value: string): value is SecFetchSite => {
+  return (secFetchSiteValues as readonly string[]).includes(value)
+}
+
+type IsAllowedSecFetchSiteHandler = (secFetchSite: SecFetchSite, context: Context) => boolean
 interface CSRFOptions {
   origin?: string | string[] | IsAllowedOriginHandler
-  secFetchSite?: string | string[] | IsAllowedSecFetchSiteHandler
+  secFetchSite?: SecFetchSite | SecFetchSite[] | IsAllowedSecFetchSiteHandler
 }

 const isSafeMethodRe = /^(GET|HEAD)$/
@@ -117,6 +125,10 @@ export const csrf = (options?: CSRFOptions): MiddlewareHandler => {
       // denied always when sec-fetch-site header is not present
       return false
     }
+    // Type guard to check if the value is a valid SecFetchSite
+    if (!isSecFetchSite(secFetchSite)) {
+      return false
+    }
     return secFetchSiteHandler(secFetchSite, c)
   }
```

What do you think of this? Or, is this over-TypeScripted?

yusukebe:
Hi @meck93 

Thank you for the PR. I think this is a good feature to improve the CSRF Protection middleware. I've left a comment.

src/middleware/csrf/index.ts
usualoma: 👍

meck93: @yusukebe thanks for the review and the feedback. 

I thought about this too but wasn't quite sure after reading the following statement in the Fetch Metadata RFC (https://www.w3.org/TR/fetch-metadata/#sec-fetch-site-header): 

> Valid Sec-Fetch-Site values include "cross-site", "same-origin", "same-site", and "none". In order to support forward-compatibility with as-yet-unknown request types, servers SHOULD ignore this header if it contains an invalid value.

I guess we'd have to update the type if the allowed values ever change but this should be an easy fix at this point.

yusukebe (approved):
LGTM!

yusukebe:
@meck93 

Looks good to me! I like this PR since the implementation is minimal and simple.

This is a new feature, but just adding the option is slight. So, I'll merge this into main right now and include the next patch release. Thanks!

PR #4360, opened 2025-08-20
feat(proxy): add `customFetch` option to allow custom fetch function

### LINKED ISSUE #4351: Feature request: Ability to overide fetch in `hono/proxy`

Currently, `hono/proxy` uses the global `fetch,` and there is no way to override it. 

I have an instance where I am using [`@cloudflare/workers-honeycome-logger`](https://www.npmjs.com/package/@cloudflare/workers-honeycomb-logger/v/0.3.6), and I’d like to use `request.tracer.fetch` so proxied requests are included in traces.

It could be implemented as an optional `fetchImpl` field to the `proxyRequestInit`, and if present, use that; otherwise, fall back to the default implementation of fetch. 

I’m happy to work on a PR if this proposal is accepted.

yusukebe:
Resolves #4351

This PR will add the option parameter `customFetch` to allow the user to override the global `fetch` in Proxy helper.

Usage:

```ts
app.get('/', () =>
  proxy(`https://example.com/`, {
    customFetch,
  })
)
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/helper/proxy/index.ts
+++ b/src/helper/proxy/index.ts
@@ -25,8 +25,21 @@ interface ProxyRequestInit extends Omit<RequestInit, 'headers'> {
     | Record<string, string | undefined>
 }
 
+/**
+ * Options for proxy function
+ * @example
+ * { fetch: customFetch } // Use custom fetch function instead of the global `fetch`
+ */
+type ProxyOptions = {
+  fetch: typeof fetch
+}
+
 interface ProxyFetch {
-  (input: string | URL | Request, init?: ProxyRequestInit): Promise<Response>
+  (
+    input: string | URL | Request,
+    init?: ProxyRequestInit,
+    options?: ProxyOptions
+  ): Promise<Response>
 }
 
 const buildRequestInitFromRequest = (
@@ -109,7 +122,7 @@ const preprocessRequestInit = (requestInit: RequestInit): RequestInit => {
  * })
  * ```
  */
-export const proxy: ProxyFetch = async (input, proxyInit) => {
+export const proxy: ProxyFetch = async (input, proxyInit, options) => {
   const { raw, ...requestInit } =
     proxyInit instanceof Request ? { raw: proxyInit } : proxyInit ?? {}
 
@@ -119,7 +132,7 @@ export const proxy: ProxyFetch = async (input, proxyInit) => {
   })
   req.headers.delete('accept-encoding')
 
-  const res = await fetch(req)
+  const res = await (options?.fetch || fetch)(req)
   const resHeaders = new Headers(res.headers)
   hopByHopHeaders.forEach((header) => {
     resHeaders.delete(header)
--- a/src/helper/proxy/index.ts
+++ b/src/helper/proxy/index.ts
@@ -23,23 +23,11 @@ interface ProxyRequestInit extends Omit<RequestInit, 'headers'> {
     | [string, string][]
     | Record<RequestHeader, string | undefined>
     | Record<string, string | undefined>
-}
-
-/**
- * Options for proxy function
- * @example
- * { fetch: customFetch } // Use custom fetch function instead of the global `fetch`
- */
-type ProxyOptions = {
-  fetch: typeof fetch
+  customFetch: (request: Request) => Promise<Response>
 }
 
 interface ProxyFetch {
-  (
-    input: string | URL | Request,
-    init?: ProxyRequestInit,
-    options?: ProxyOptions
-  ): Promise<Response>
+  (input: string | URL | Request, init?: ProxyRequestInit): Promise<Response>
 }
 
 const buildRequestInitFromRequest = (
@@ -122,8 +110,8 @@ const preprocessRequestInit = (requestInit: RequestInit): RequestInit => {
  * })
  * ```
  */
-export const proxy: ProxyFetch = async (input, proxyInit, options) => {
-  const { raw, ...requestInit } =
+export const proxy: ProxyFetch = async (input, proxyInit) => {
+  const { raw, customFetch, ...requestInit } =
     proxyInit instanceof Request ? { raw: proxyInit } : proxyInit ?? {}
 
   const req = new Request(input, {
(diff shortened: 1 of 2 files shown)


yusukebe:
Hey @karthik2804 @riywo @usualoma , can you review this?

karthik2804 (approved):
This looks good to me! I really appreciate the swift PR. Thank you!

riywo:
LGTM as well!

src/helper/proxy/index.ts
```diff
@@ -25,8 +25,21 @@ interface ProxyRequestInit extends Omit<RequestInit, 'headers'> {
...
 
+/**
+ * Options for proxy function
+ * @example
+ * { fetch: customFetch } // Use custom fetch function instead of the global `fetch`
+ */
+type ProxyOptions = {
+  fetch: typeof fetch
+}
+
```
BarryThePenguin: I like what `oauth4webapi` does with a [`customFetch`](https://github.com/panva/oauth4webapi/blob/main/docs/variables/customFetch.md) option as a `unique symbol`

```ts
import { customFetch } from 'hono'
import { proxy } from 'hono/proxy'

proxy(`https://example.com/`, {
  [customFetch](...args) {
    return c.env.AUTH.fetch(...args)
  }
})
```

At least that way you could avoid adding a third argument?

yusukebe: Interesting! Or making super simple, how about `fetch` property in the second args?

```ts
interface ProxyRequestInit extends Omit<RequestInit, 'headers'> {
  raw?: Request
  headers?:
    | HeadersInit
    | [string, string][]
    | Record<RequestHeader, string | undefined>
    | Record<string, string | undefined>
  fetch?: typeof fetch
}
```

Because `RequestInit` never has a property named `fetch`. So you can write like the following:

```ts
app.post('/', (c) => {
  return proxy(`https://example.com/`, {
    fetch: customFetch,
    ...c.req,
  })
})
```

The diff:

```diff
diff --git a/src/helper/proxy/index.test.ts b/src/helper/proxy/index.test.ts
index a57800a9..e0c02df4 100644
--- a/src/helper/proxy/index.test.ts
+++ b/src/helper/proxy/index.test.ts
@@ -257,22 +257,23 @@ describe('Proxy Middleware', () => {
     })

     it('Should call the custom fetch method when specified', async () => {
-      const customFetch = vi.fn().mockImplementation(async () => {
-        return new Response('custom fetch response')
+      const customFetch = vi.fn().mockImplementation(async (req: Request) => {
+        const text = await req.text()
+        return new Response('custom fetch response. message:' + text)
       })
       const app = new Hono()
-      app.get('/', () =>
-        proxy(
-          `https://example.com/`,
-          {},
-          {
-            fetch: customFetch,
-          }
-        )
-      )
-      const res = await app.request('/')
+      app.post('/', (c) => {
+        return proxy(`https://example.com/`, {
+          fetch: customFetch,
+          ...c.req,
+        })
+      })
+      const res = await app.request('/', {
+        method: 'POST',
+        body: 'hi',
+      })
       expect(res.status).toBe(200)
-      expect(await res.text()).toBe('custom fetch response')
+      expect(await res.text()).toBe('custom fetch response. message:hi')
     })
   })
 })
diff --git a/src/helper/proxy/index.ts b/src/helper/proxy/index.ts
index 70c9d0b4..e11c2c30 100644
--- a/src/helper/proxy/index.ts
+++ b/src/helper/proxy/index.ts
@@ -23,23 +23,11 @@ interface ProxyRequestInit extends Omit<RequestInit, 'headers'> {
     | [string, string][]
     | Record<RequestHeader, string | undefined>
     | Record<string, string | undefined>
-}
-
-/**
- * Options for proxy function
- * @example
- * { fetch: customFetch } // Use custom fetch function instead of the global `fetch`
- */
-type ProxyOptions = {
-  fetch: typeof fetch
+  fetch?: typeof fetch
 }

 interface ProxyFetch {
-  (
-    input: string | URL | Request,
-    init?: ProxyRequestInit,
-    options?: ProxyOptions
-  ): Promise<Response>
+  (input: string | URL | Request, init?: ProxyRequestInit): Promise<Response>
 }

 const buildRequestInitFromRequest = (
@@ -122,9 +110,12 @@ const preprocessRequestInit = (requestInit: RequestInit): RequestInit => {
  * })
  * ```
  */
-export const proxy: ProxyFetch = async (input, proxyInit, options) => {
-  const { raw, ...requestInit } =
-    proxyInit instanceof Request ? { raw: proxyInit } : proxyInit ?? {}
+export const proxy: ProxyFetch = async (input, proxyInit) => {
+  const {
+    raw,
+    fetch: customFetch,
+    ...requestInit
+  } = proxyInit instanceof Request ? { raw: proxyInit } : proxyInit ?? {}

   const req = new Request(input, {
     ...buildRequestInitFromRequest(raw),
@@ -132,7 +123,7 @@ export const proxy: ProxyFetch = async (input, proxyInit, options) => {
   })
   req.headers.delete('accept-encoding')

-  const res = await (options?.fetch || fetch)(req)
+  const res = await (customFetch || fetch)(req)
   const resHeaders = new Headers(res.headers)
   hopByHopHeaders.forEach((header) => {
     resHeaders.delete(header)
```

yusukebe: @BarryThePenguin @usualoma What do you think of this^ ?

BarryThePenguin: I think the type should be..

```ts
fetch: (request: Request) => Promise<Response>
```

..as it's only ever called with a `Request`. 

Even naming it `fetch` isn't entirely correct 🤔

yusukebe: @BarryThePenguin 

Ah, you are right! It's not a complete `fetch`. So, how about below?

```ts
interface ProxyRequestInit extends Omit<RequestInit, 'headers'> {
  raw?: Request
  headers?:
    | HeadersInit
    | [string, string][]
    | Record<RequestHeader, string | undefined>
    | Record<string, string | undefined>
  customFetch: (request: Request) => Promise<Response>
}
```

Using a `unique symbol` like `oauth4webapi` is interesting, but for the casual user, using `customFetch` as a string property is easy to understand.

BarryThePenguin: Yeah, that should be fine

usualoma: 👍

yusukebe:
@karthik2804 @riywo @BarryThePenguin @usualoma 

As discussed in the discussion: https://github.com/honojs/hono/pull/4360#pullrequestreview-3135719109, I decided to go with this API:

```ts
app.get('/', () =>
  proxy(`https://example.com/`, {
    customFetch,
  })
)
```

Thank you!

PR #4362, opened 2025-08-21
chore: update `typescript` to `5.9.2`

yusukebe:
This PR updates the `typescript` version and fixes the TypeScript error caused by the update.

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/helper/ssg/ssg.test.tsx
+++ b/src/helper/ssg/ssg.test.tsx
@@ -367,7 +367,7 @@ describe('saveContentToFile function', () => {
   const gzFileArrayBuffer = gzFileBuffer.buffer.slice(
     gzFileBuffer.byteOffset,
     gzFileBuffer.byteLength + gzFileBuffer.byteOffset
-  )
+  ) as ArrayBuffer
   // PNG, red dot (1x1)
   const pngFileBuffer = Buffer.from(
     'iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAIAAACQd1PeAAAADElEQVR4nGP4z8AAAAMBAQDJ/pLvAAAAAElFTkSuQmCCAAAALw',
@@ -376,7 +376,7 @@ describe('saveContentToFile function', () => {
   const pngFileArrayBuffer = pngFileBuffer.buffer.slice(
     pngFileBuffer.byteOffset,
     pngFileBuffer.byteLength + pngFileBuffer.byteOffset
-  )
+  ) as ArrayBuffer
 
   const fileData = [
     { routePath: '/', content: 'Home Page', mimeType: 'text/html' },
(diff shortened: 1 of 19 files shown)


src/helper/ssg/ssg.test.tsx
```diff
@@ -367,7 +367,7 @@ describe('saveContentToFile function', () => {
   const gzFileArrayBuffer = gzFileBuffer.buffer.slice(
     gzFileBuffer.byteOffset,
     gzFileBuffer.byteLength + gzFileBuffer.byteOffset
-  )
+  ) as ArrayBuffer
```
yusukebe: The `as ArrayBuffer` is required because `Buffer.buffer` returns `ArrayBufferLike` in TypeScript 5.9, but we know it's always `ArrayBuffer` in Node.js context.

src/helper/websocket/index.ts
```diff
@@ -85,7 +85,12 @@ export class WSContext<T = unknown> {
   }
 }
 
-export type WSMessageReceive = string | Blob | ArrayBufferLike
+export type WSMessageReceive =
```
yusukebe: Replaced `ArrayBufferLike` with explicit types (`ArrayBuffer | SharedArrayBuffer | Uint8Array<ArrayBuffer>`) to support all buffer types.

src/utils/jwt/jwt.test.ts
```diff
@@ -339,7 +339,10 @@ describe('JWT', () => {
     }
     const secret = await crypto.subtle.importKey(
       'raw',
-      Buffer.from('cefb73234d5fae4bf27662900732b52943e8d53e871fe0f353da95de4599c21d', 'hex'),
+      Buffer.from(
+        'cefb73234d5fae4bf27662900732b52943e8d53e871fe0f353da95de4599c21d',
+        'hex'
+      ) as BufferSource,
```
yusukebe: `Buffer.from()` returns a Node.js Buffer whose type doesn't directly match `BufferSource` in TypeScript 5.9.

yusukebe:
Hey @BarryThePenguin , can you review this?

yusukebe:
@BarryThePenguin Thanks!

So, it seems that if we use the latest `@types/node` like `^24.3.0`, there are not so many type inconsistencies. Do you think updating `@types/node` is a good idea?

yusukebe:
@BarryThePenguin Ah, I confirmed if update the `@types/node`, we can remove the cast like here https://github.com/honojs/hono/pull/4362/files#diff-d5d213f0f22d927e88abccf74312feb76e5e8e89049ab1459a7305feb8346917R342-R345, but it's only for the internal test. So, I think the current is okay!

yusukebe:
I updated the `@types/*` and it will remove the some type inconsistencies: https://github.com/honojs/hono/pull/4362/commits/980e6a88a146c5f251d6627cdf0e9a0eb49391f7

I think it's okay to merge. Thanks.

PR #4365, opened 2025-08-22
chore: replace supertest with undici

BarryThePenguin:
While looking at #4362 I had a go at isolating `lib.dom` types from `@types/node`

Removing `supertest` removes a number of dependencies including `@types/supertest`, `@types/superagent`, and a stray `@types/node@20.11.4`

It was only used in one test, so was pretty straightforward to swap

### DIFF
--- a/runtime-tests/node/index.test.ts
+++ b/runtime-tests/node/index.test.ts
@@ -29,13 +29,13 @@ describe('Basic', () => {
   it('Should return 200 response', async () => {
     const res = await agent.get('/')
     expect(res.status).toBe(200)
-    expect(res.text).toBe('Hello! Node.js!')
+    await expect(res.text()).resolves.toBe('Hello! Node.js!')
   })
 
   it('Should return correct runtime name', async () => {
     const res = await agent.get('/runtime-name')
     expect(res.status).toBe(200)
-    expect(res.text).toBe('node')
+    await expect(res.text()).resolves.toBe('node')
   })
 })
 
@@ -67,14 +67,14 @@ describe('Basic Auth Middleware', () => {
   it('Should not authorize, return 401 Response', async () => {
     const res = await agent.get('/auth/a')
     expect(res.status).toBe(401)
-    expect(res.text).toBe('Unauthorized')
+    await expect(res.text()).resolves.toBe('Unauthorized')
   })
 
   it('Should authorize, return 200 Response', async () => {
     const credential = 'aG9uby11c2VyLWE6aG9uby1wYXNzd29yZC1h'
     const res = await agent.get('/auth/a', { headers: { Authorization: `Basic ${credential}` } })
     expect(res.status).toBe(200)
-    expect(res.text).toBe('auth')
+    await expect(res.text()).resolves.toBe('auth')
   })
 })
 
@@ -89,7 +89,7 @@ describe('JWT Auth Middleware', () => {
   it('Should not authorize, return 401 Response', async () => {
     const res = await agent.get('/jwt/a')
     expect(res.status).toBe(401)
-    expect(res.text).toBe('Unauthorized')
+    await expect(res.text()).resolves.toBe('Unauthorized')
   })
 
   it('Should authorize, return 200 Response', async () => {
@@ -97,7 +97,7 @@ describe('JWT Auth Middleware', () => {
       'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJtZXNzYWdlIjoiaGVsbG8gd29ybGQifQ.B54pAqIiLbu170tGQ1rY06Twv__0qSHTA0ioQPIOvFE'
     const res = await agent.get('/jwt/a', { headers: { Authorization: `Bearer ${credential}` } })
     expect(res.status).toBe(200)
-    expect(res.text).toBe('auth')
+    await expect(res.text()).resolves.toBe('auth')
   })
 })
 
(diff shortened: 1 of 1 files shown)


runtime-tests/node/index.test.ts
```diff
@@ -257,17 +262,46 @@ describe('Buffers', () => {
...
+
+function createAgent(app: Hono) {
+  const server = createAdaptorServer(app)
+  const listening = once(server.listen(), 'listening')
+
+  return {
+    async get(path: string, init?: undici.RequestInit) {
+      await listening
+      const url = new URL(path, getOrigin())
+      const response = await undici.fetch(url, init)
```
yusukebe: I know you don't want to change the many test lines, but I think it's better to make the return value a pure `Response`:

```ts
return {
  async get(path: string, init?: undici.RequestInit) {
    await listening
    const url = new URL(path, getOrigin())
    return undici.fetch(url, init)
  },
}
```

If so, `agent.get()` will return `Promise<Response>`, it is understandable rather than the current object. What do you think of it? If you add the change, you can patch the following:

```diff
diff --git a/runtime-tests/node/index.test.ts b/runtime-tests/node/index.test.ts
index aa852dae..4c7f0bfa 100644
--- a/runtime-tests/node/index.test.ts
+++ b/runtime-tests/node/index.test.ts
@@ -29,13 +29,13 @@ describe('Basic', () => {
   it('Should return 200 response', async () => {
     const res = await agent.get('/')
     expect(res.status).toBe(200)
-    expect(res.text).toBe('Hello! Node.js!')
+    expect(await res.text()).toBe('Hello! Node.js!')
   })
 
   it('Should return correct runtime name', async () => {
     const res = await agent.get('/runtime-name')
     expect(res.status).toBe(200)
-    expect(res.text).toBe('node')
+    expect(await res.text()).toBe('node')
   })
 })
 
@@ -67,14 +67,14 @@ describe('Basic Auth Middleware', () => {
   it('Should not authorize, return 401 Response', async () => {
     const res = await agent.get('/auth/a')
     expect(res.status).toBe(401)
-    expect(res.text).toBe('Unauthorized')
+    expect(await res.text()).toBe('Unauthorized')
   })
 
   it('Should authorize, return 200 Response', async () => {
     const credential = 'aG9uby11c2VyLWE6aG9uby1wYXNzd29yZC1h'
     const res = await agent.get('/auth/a', { headers: { Authorization: `Basic ${credential}` } })
     expect(res.status).toBe(200)
-    expect(res.text).toBe('auth')
+    expect(await res.text()).toBe('auth')
   })
 })
 
@@ -89,7 +89,7 @@ describe('JWT Auth Middleware', () => {
   it('Should not authorize, return 401 Response', async () => {
     const res = await agent.get('/jwt/a')
     expect(res.status).toBe(401)
-    expect(res.text).toBe('Unauthorized')
+    expect(await res.text()).toBe('Unauthorized')
   })
 
   it('Should authorize, return 200 Response', async () => {
@@ -97,7 +97,7 @@ describe('JWT Auth Middleware', () => {
       'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJtZXNzYWdlIjoiaGVsbG8gd29ybGQifQ.B54pAqIiLbu170tGQ1rY06Twv__0qSHTA0ioQPIOvFE'
     const res = await agent.get('/jwt/a', { headers: { Authorization: `Bearer ${credential}` } })
     expect(res.status).toBe(200)
-    expect(res.text).toBe('auth')
+    expect(await res.text()).toBe('auth')
   })
 })
 
@@ -151,7 +151,7 @@ describe('stream', () => {
     expect(aborted).toBe(false)
     const res = await agent.get('/streamHello')
     expect(res.status).toBe(200)
-    expect(res.text).toBe('Hello')
+    expect(await res.text()).toBe('Hello')
     expect(aborted).toBe(false)
   })
 })
@@ -206,7 +206,7 @@ describe('streamSSE', () => {
     expect(aborted).toBe(false)
     const res = await agent.get('/streamHello')
     expect(res.status).toBe(200)
-    expect(res.text).toBe('Hello')
+    expect(await res.text()).toBe('Hello')
     expect(aborted).toBe(false)
   })
 })
@@ -249,7 +249,7 @@ describe('compress', async () => {
     const res = await agent.get('/fetch/style.css')
     expect(res.status).toBe(200)
     expect(res.headers.get('content-encoding')).toBe('gzip')
-    expect(res.text).toBe(cssContent)
+    expect(await res.text()).toBe(cssContent)
   })
 })
 
@@ -267,13 +267,13 @@ describe('Buffers', () => {
   it('should allow returning buffers', async () => {
     const res = await agent.get('/')
     expect(res.status).toBe(200)
-    expect(res.text).toBe('hello')
+    expect(await res.text()).toBe('hello')
   })
 
   it('should allow returning uint8array as well', async () => {
     const res = await agent.get('/uint8array')
     expect(res.status).toBe(200)
-    expect(res.text).toBe('hello')
+    expect(await res.text()).toBe('hello')
   })
 })
 
@@ -285,13 +285,7 @@ function createAgent(app: Hono) {
     async get(path: string, init?: undici.RequestInit) {
       await listening
       const url = new URL(path, getOrigin())
-      const response = await undici.fetch(url, init)
-
-      return {
-        headers: response.headers,
-        status: response.status,
-        text: await response.text(),
-      }
+      return undici.fetch(url, init)
     },
   }
 ```

BarryThePenguin: Yup! That sounds reasonable. I got sidetracked building my own version of `supertest` using `undici` 🙈 but eventually figured out that wasn't what I needed here 😂 

I'll update the tests to use a plain response 👍🏻

BarryThePenguin: I also found [`sagetest`](https://eddienubes.github.io/sagetest), but it didn't do everything I wanted 😩 

It would also be nice if `vitest` had an assertion for `Request` and `Response` objects, I'll add that to my todo list.. 📝

yusukebe:
@BarryThePenguin 

Thanks! Merging now.

PR #4373, opened 2025-08-26
feat(cors): Allow async functions for `origin` and `allowMethods`

jobrk:
This PR updates the CORS middleware to support asynchronous functions for the `origin` and `allowMethods` options.

This allows for dynamic, asynchronous logic (e.g., database lookups) when determining CORS headers.

```ts
app.use(
  '/api/*',
  cors({
    origin: async (origin) => {
      // e.g. check an allow-list in a database
      const isAllowed = await db.isOriginAllowed(origin)
      return isAllowed ? origin : 'https://fallback.example.com'
    },
  })
)
```

This change is backward-compatible.

### DIFF
--- a/src/middleware/cors/index.test.ts
+++ b/src/middleware/cors/index.test.ts
@@ -229,7 +229,7 @@ describe('CORS by Middleware', () => {
   })
 
   it('Allow origins by promise returning function', async () => {
-    let req = new Request('http://localhost/api4/abc', {
+    let req = new Request('http://localhost/api8/abc', {
       headers: {
         Origin: 'http://subdomain.example.com',
       },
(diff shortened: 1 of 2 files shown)


src/middleware/cors/index.test.ts
```diff
@@ -202,6 +228,28 @@ describe('CORS by Middleware', () => {
     expect(res.headers.get('Access-Control-Allow-Origin')).toBe('http://example.com')
   })
 
+  it('Allow origins by promise returning function', async () => {
+    let req = new Request('http://localhost/api4/abc', {
```
yusukebe: `http://localhost/api4/abc` should be `http://localhost/api8/abc`?

jobrk: my bad, thanks & fixed

yusukebe:
@jobrk 

Looks good to me. I'll merge this and include this feature in the next patch release. Thanks!

PR #4375, opened 2025-08-27
feat(cors): Correct origin function return type asynchronously returning null or undefined for origin

jobrk:
Forgot about that in [this PR](https://github.com/honojs/hono/pull/4373)

Updates the `origin` function's return type. The `Promise` can now resolve to `undefined` or `null`, aligning it with the synchronous return type.


yusukebe:
@jobrk Thanks!

PR #4376, opened 2025-08-27
nit: Export CORSOptions type

jobrk:
This will be useful to define custom CORS middlewares while still passing down default options without using something ugly like `NonNullable<Parameters<typeof cors>[0]>` 😅

### DIFF
--- a/src/middleware/cors/index.ts
+++ b/src/middleware/cors/index.ts
@@ -6,7 +6,7 @@
 import type { Context } from '../../context'
 import type { MiddlewareHandler } from '../../types'
 
-type CORSOptions = {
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @jobrk 

We talked about the matter before. It does not cause big problems, but based on the principle of minimizing the number of APIs exposed externally, we have chosen not to export the middleware options.

jobrk:
Hey @yusukebe

Interesting decision, I am not sure to be understanding what risk/issue would it create to be completely honest.

Regardless, solved locally with:
```ts
export type CORSOptions = NonNullable<Parameters<typeof cors>[0]>
```

PR #4382, opened 2025-08-31
fix: cloning raw HonoRequest before consuming

kamaal111:
**Problem**

After using Hono validators, the `raw` Request obejct gets consumed during body parsing, making it unusable for external libraries like `better-auth`.
This results in the error:

```shell
TypeError: Cannot construct a Request with a Request object that has already been used.
```

**Root Cause**

The issue occurs in the `#cachedBody` method in `HonoRequest`. When parsing request bodies (json, text, etc.), the method directly calls parsing methods on the raw Request object, which consumes its body stream. Once consumed, the Request cannot be cloned or reused by external libraries.

**Solution**

Clone the raw Request object before consuming its body. This ensures that:
- The original `c.req.raw` remains pristine and can be used by external libraries
- Hono's internal body caching still works correctly
- No breaking changes to existing functionality

- [x] Add tests
- [x] Run tests
- [x] Add [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code (Updated a private method, not sure if applicable)

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -227,7 +227,8 @@ export class HonoRequest<P extends string = '/', I extends Input['out'] = {}> {
       })
     }
 
-    return (bodyCache[key] = raw[key]())
+    const clonedRaw = raw.clone()
+    return (bodyCache[key] = clonedRaw[key]())
   }
 
   /**
--- a/src/validator/validator.test.ts
+++ b/src/validator/validator.test.ts
@@ -1092,6 +1092,87 @@ describe('Clone Request object', () => {
       expect(await res.json()).toEqual({ foo: 'bar' })
     })
   })
+
+  describe('Raw Request preservation after validation', () => {
+    it('Should preserve raw Request after JSON validation for external libraries', async () => {
+      const app = new Hono()
+
+      app.post(
+        '/json-validation',
+        validator('json', (value) => value),
+        async (c) => {
+          function createExternalRequest() {
+            return new Request(c.req.raw.url, {
+              method: c.req.raw.method,
+              headers: c.req.raw.headers,
+            })
+          }
+
+          const externalRequest = createExternalRequest()
+          const clone1 = c.req.raw.clone()
+          const clone2 = c.req.raw.clone()
+
+          return c.json({
+            success: true,
+            externalRequestMethod: externalRequest.method,
+            originalMethod: c.req.raw.method,
+            cloned: clone1 instanceof Request && clone2 instanceof Request,
+          })
+        }
+      )
+
+      const testData = { message: 'test', userId: 123 }
+      const res = await app.request('/json-validation', {
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify(testData),
+      })
+
+      expect(res.status).toBe(200)
+
+      const result = await res.json()
+
+      expect(result.success).toBe(true)
+      expect(result.externalRequestMethod).toBe('POST')
+      expect(result.originalMethod).toBe('POST')
+      expect(result.cloned).toBe(true)
+    })
+
+    it('Should preserve raw Request after form validation for external libraries', async () => {
+      const app = new Hono()
+
+      app.post(
+        '/form-validation',
+        validator('form', (value) => value),
+        async (c) => {
+          expect(() => {
+            return new Request(c.req.raw.url, {
+              method: c.req.raw.method,
+              headers: c.req.raw.headers,
+            })
+          }, 'Test that raw request can be used by external libraries after form validation').not.toThrow()
+          expect(() => c.req.raw.clone()).not.toThrow()
+
+          return c.json({ success: true })
+        }
+      )
+
+      const formData = new FormData()
+      formData.append('username', 'testuser')
+      formData.append('email', 'test@example.com')
+
+      const res = await app.request('/form-validation', {
+        method: 'POST',
+        body: formData,
+      })
+
+      expect(res.status).toBe(200)
+
+      const result = await res.json()
+
+      expect(result.success).toBe(true)
+    })
+  })
 })
 
 describe('Async validator function', () => {
--- a/src/request.test.ts
+++ b/src/request.test.ts
@@ -187,6 +187,97 @@ const text = '{"foo":"bar"}'
 const json = { foo: 'bar' }
 const buffer = new TextEncoder().encode('{"foo":"bar"}').buffer
 
+describe('Raw Request preservation', () => {
+  test('raw Request remains unconsumed after body parsing', async () => {
+    const originalRequest = new Request('http://localhost/', {
+      method: 'POST',
+      body: JSON.stringify({ test: 'data' }),
+      headers: { 'Content-Type': 'application/json' },
+    })
+    const req = new HonoRequest(originalRequest)
+
+    await req.json()
+    await req.text()
+    await req.arrayBuffer()
+
+    expect(() => req.raw.clone(), 'The raw request should still be usable').not.toThrow()
+    expect(
+      () => new Request(req.raw.url, { method: req.raw.method, headers: req.raw.headers }),
+      'Should be able to create a new Request from the raw request'
+    ).not.toThrow()
+    expect(req.raw.method).toBe('POST')
+    expect(req.raw.url).toBe('http://localhost/')
+    expect(req.raw.headers.get('Content-Type')).toBe('application/json')
+  })
+
+  test('raw Request can be cloned multiple times after body parsing', async () => {
+    const originalRequest = new Request('http://localhost/', {
+      method: 'POST',
+      body: JSON.stringify({ data: 'test' }),
+      headers: { 'Content-Type': 'application/json' },
+    })
+    const req = new HonoRequest(originalRequest)
+
+    await req.json()
+    await req.text()
+    const clone1 = req.raw.clone()
+    const clone2 = req.raw.clone()
+    const clone3 = req.raw.clone()
+
+    expect(clone1).toBeInstanceOf(Request)
+    expect(clone2).toBeInstanceOf(Request)
+    expect(clone3).toBeInstanceOf(Request)
+    expect(clone1.method).toBe(clone2.method)
+    expect(clone1.url).toBe(clone2.url)
+    expect(clone1.method).toBe(clone3.method)
+  })
+
+  test('external libraries can use raw Request after validation', async () => {
+    const originalRequest = new Request('http://localhost/api/test', {
+      method: 'POST',
+      body: JSON.stringify({ username: 'test', password: 'secret' }),
+      headers: { 'Content-Type': 'application/json' },
+    })
+    const req = new HonoRequest(originalRequest)
+
+    const data = await req.json()
+    expect(data).toStrictEqual({ username: 'test', password: 'secret' })
+
+    function externalLibraryRequest() {
+      return new Request(req.raw.url, {
+        method: req.raw.method,
+        headers: req.raw.headers,
+      })
+    }
+
+    const newRequest = externalLibraryRequest()
+
+    expect(newRequest.method).toBe('POST')
+    expect(newRequest.url).toBe('http://localhost/api/test')
+    expect(newRequest.headers.get('Content-Type')).toBe('application/json')
+  })
+
+  test('different body types can be accessed after raw Request preservation', async () => {
+    const formData = new FormData()
+    formData.append('field1', 'value1')
+    formData.append('field2', 'value2')
+    const originalRequest = new Request('http://localhost', {
+      method: 'POST',
+      body: formData,
+    })
+
+    const req = new HonoRequest(originalRequest)
+
+    const parsedFormData = await req.formData()
+
+    expect(parsedFormData.get('field1')).toBe('value1')
+    expect(parsedFormData.get('field2')).toBe('value2')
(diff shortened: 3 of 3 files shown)


yusukebe:
Hi @kamaal111 

Why I didn't implement cloning the request is that it will introduce significant performance degradation.

The app:

```ts

const app = new Hono()

app.post('/', async (c) => {
  return c.json(data)
})

export default app
```

Benchmark command:

```
bombardier -c 125 --fasthttp --method=POST --body='{"foo":"bar"}' http://localhost:3000/
```

Result:

CleanShot 2025-09-05 at 06 03 16@2x

We should find a good way to resolve the problem without performance delegation.

kamaal111:
I understand, thank you @yusukebe!
What about instead of cloning the Request object, I instead instantiate a new Request object with what is available in `HonoRequest`, and construct that in a new method?

Something like this:

```typescript
export class HonoRequest {
    // ...
    new = () => {
        return new Request(this.url, {
            // Collect request init from this class without accessing `raw`
        })
    }
}
```

That would then get consumed as the following:

```typescript
const app = new Hono()

app.post(
  '/login',
  zValidator(
    'json',
    z.object({
      username: z.string(),
    })
  ),
  (c) => {
    const validated = c.req.valid('json')
    // `auth` from better-auth for example
    const response = await auth.handler(c.req.new());
  }
)
```

Would it be worth it to contribute this to this project?

yusukebe:
@kamaal111 

> Would it be worth it to contribute this to this project?

I'm also thinking of familiar things. Please wait until my vacation ends.

yusukebe:
@kamaal111 

I'm working on this issue, but I can't find a good solution. Can you share more details of your idea?

kamaal111:
@yusukebe  I have workaround for it in my own project that only requires a Request object with JSON and it looks something like this:

```ts
import type { HonoContext } from '../api/contexts.js';

export async function makeNewRequest(c: HonoContext, options?: { withBody?: boolean }): Promise<Request> {
  let body: string | undefined = undefined;
  if (options?.withBody) {
    const rawBody: unknown = await c.req.json();
    body = JSON.stringify(rawBody);
  }
  const requestInit: RequestInit = { method: c.req.method, headers: c.req.header(), body };

  return new Request(c.req.url, requestInit);
}
```

But for Hono I am thinking something like this:

```ts
// ...

// Type safe transformer for request init
const BODY_TYPE_TO_TRANSFORMER_MAPPING: { [K in keyof Body]: (body: unknown) => Body[K] } = {
  json: (body) => JSON.stringify(body),
  text: (body) => body as string,
  arrayBuffer: (body) => body as ArrayBuffer,
  blob: (body) => body as Blob,
  formData: (body) => body as FormData,
}

export class HonoRequest {
    // ...

  new = async (): Promise<Request> => {
    const requestInit: RequestInit = { method: this.method, headers: this.header() }
    const cachedKeys = Object.keys(this.bodyCache) as Array<keyof typeof this.bodyCache>
    // If no body is cached but the raw request has a body and it's not consumed, consume it
    if (cachedKeys.length === 0 && this.raw.body && !this.raw.bodyUsed) {
      const body = await this.raw.text()
      this.bodyCache.text = body
      requestInit.body = body
      return new Request(this.url, requestInit)
    }

    const firstKey = cachedKeys[0] as keyof Body | undefined
    // If no cached body, return request without body
    if (firstKey == null) return new Request(this.url, requestInit)

    const cachedBody = await this.bodyCache[firstKey]
    requestInit.body = BODY_TYPE_TO_TRANSFORMER_MAPPING[firstKey](cachedBody) ?? cachedBody

    return new Request(this.url, requestInit)
  }
}
```

I haven't tested this in my application yet though, just drafted it quickly, but I will write proper tests for this and also see how it behaves in my app.
Will get back to you after I have validated this change.

kamaal111:
Just confirming that the following works fine for my project now.

```typescript
const app = new Hono()

app.post(
  '/login',
  zValidator(
    'json',
    z.object({
      username: z.string(),
    })
  ),
  (c) => {
    const validated = c.req.valid('json');
    //  New Request object constructor method, that is valid even after consumption.
    const request = await c.req.new();
    // `auth` from better-auth for example
    const response = await auth.handler(request);
  }
)
```

kamaal111:
I have pushed my updated solution, let me know if its alright

yusukebe:
Hi @kamaal111 

Adding a `new()` function is a great idea that I also think is similar! And the implementation seems to be fine. But it's not good that the logic/code has increased. The core object, `HonoRequest`, should be simple and small. We have to consider them more deeply. I'll think it again.

kamaal111:
I understand @yusukebe 
I have another idea, what if `.raw` was an getter property on HonoRequest?

Excuse me if I am making mistakes writing the following, because I can only write this from my phone for the coming 2 weeks.

```ts
class HonoRequest {
  // ...
  // internal raw Request object which was previously `raw`
  private _raw: Request

  get raw() {
    // check if raw has been consumed
    // if it has been consumed then return a new Request object with cached values
    // else return raw in this class
  }
}
```

Note that we don't need a promise with the getter, because we have 2 simple cases and we don't need to think about forcing a new RequestObject if the raw object has not been consumed.

This should fix the issue of using raw while it is not consumed, but to not cause regressions in performance we likely need to reference the non getter raw prober and the getter raw property would be used for public applications.

What do you think about this solution?

yusukebe:
@kamaal111 

I think adding a raw getter is good, but how to create a new Request object if it has been consumed? The problem is that cached values are `Promise`. And the `raw` getter can not be an `async` (right?).

```ts
  get raw() {
    if (this._raw.bodyUsed) {
      // get body from the cached values?
      // but the values in this.bodyCache are Promise
      return new Request(this._raw, {
        ...this._raw,
        body, // how to get this?
      })
    }
    return this._raw
  }
```

kamaal111:
@yusukebe you're right I missed that fact.
Looking at how `#cachedBody` is used in HonoRequest, what if we cache the resolved values instead of caching the body promise?

We will then also have access the body cache directly without the convenience method.

Then we can make a `raw` property getter work

yusukebe:
Hey @kamaal111 

Nice suggestion. The performance degradation is not big with the solution.

I've created the new PR: https://github.com/honojs/hono/pull/4425

What do you think of it?

kamaal111:
Hey @yusukebe 
The change looks good at a glance, but I can't give a thorough enough review with verification yet, because I am traveling for 1 more week.
I did my best to give a mobile review on the PR

yusukebe:
@kamaal111 

Thanks! I'll check it.

usualoma:
Sorry to interrupt.

### What is `raw`?

I think `c.req.raw` is expected to return the “original object”. Returning something newly created with `new Request()` might not be ideal.

### Potential for utility methods

While having too many utility methods has its pros and cons, the scenario of “consuming the body with `c.req` and then consuming it again with `c.req.raw`” is relatively rare. Therefore, adding a utility method might be preferable to increasing the `HonoRequest` code.

Even in the current state, it's not necessarily required to change the `HonoRequest` code. I believe we could add a utility method with an implementation like the following:


```ts
const cloneRawRequest = async (req: HonoRequest) => {
  if (!req.raw.bodyUsed) {
    return req.raw.clone();
  }

  return new Request(req.raw.url, {
    method: req.raw.method,
    headers: req.raw.headers,
    body: await req.blob(),
  })
}

const app = new Hono()

app.post('/', async (c) => {
  const text = await c.req.text()
  const req = await cloneRawRequest(c.req)
  return c.json({
    textViaReq: text,
    textViaCloneRawRequest: await req.text(),
  })
})
```

kamaal111:
Thank you @usualoma for your comment!

The main problem I have been facing is that I would like to run validating middleware before I enter my handler, and my handler body would invoke better-auth's handler, as described here:

https://hono.dev/examples/better-auth

I know my use case is quite unique, but my goal is to have the endpoints documented with OpenAPI.

So for this cloning it outside of HonoRequest would not work, because the body has already been consumed before I get access to the raw Request object in my handler.

I think either the solution that @yusukebe presented in his PR would work or the solution in this PR (mine) if we want to keep changing it in Hono.

Alternatively we can also make better-auth work with a "consumed" Request object, but this is a very short term solution as there might be other third party integrations in the future that would like to use the Request object somehow.

yusukebe:
@usualoma 

Interrupting is no problem! Thank you for the idea. Making a helper is nice to keep the HonoRequest clean. I may make it without the problem @kamaal111 mentioned. I'll work on it later.

yusukebe:
Hi @usualoma 

Sorry! Your `cloneRawRequest` works fine. I think that is expected.

@kamaal111 For example, the following works without errors.

```ts
app.post(
  '/',
  validator('json', (data) => {
    return data
  }),
  async (c) => {
    const rawRequest = await cloneRawRequest(c.req)
    auth.handler(rawRequest)
    // ...
  }
)
```

Is this what you want to have? I think it's a good idea to make a utility function rather than adding logic to `HonoRequest`.

kamaal111:
Hi @yusukebe @usualoma,
Sorry of the late response, I just verified in my application that the `cloneRawRequest` also works for my use-case, thank you 🙏

Is this something that can be added to this repo, or is there another repo better suited for it?
I can also just keep it in my own project

yusukebe:
Hi @kamaal111 

We can include `cloneRawReques` into this repo! It's good to make a [helper](https://hono.dev/docs/guides/helpers) for it. Like his:

```ts
import { cloneRawRequest } from 'hono/request-helper'
```

However, I'm unsure if `hono/request-helper` is the best naming choice. `hono/request` would be a suitable alternative, but it's already used in `src/request.ts`. Another idea is `hono/clone`.

@usualoma Do you have any idea for the naming?

usualoma:
Well, generally speaking, it seems natural that both the “class” and “class-related utility methods” are exported from "hono/request".

```ts
import { HonoRequest, cloneRawRequest } from 'hono/request'
```

However, I recognize that Hono's existing helpers aren't structured that way. So, as you say, maybe `hono/request-helper`?

I feel like putting it in `hono/request` would be fine too.

yusukebe:
@usualoma 

Ah, `hono/request` is good! If so, it may be natural to add `cloneRawRequest` to `src/request.ts`. We don't need to create `src/helper/request.ts` or `src/helper/request-helper.ts`. Keeping the structure simple is important. Additionally, we can document the usage of `cloneRawRequest` on the `HonoRequest` page (https://hono.dev/docs/api/request) on the website.

Hey @kamaal111 

Will you create a PR for `cloneRawRequest`? As we mentioned, you can add `cloneRawRequest` to `src/request.ts`.

kamaal111:
@yusukebe that sounds good!
I will try to make some time for it during the week, if not I will likely have a PR ready this weekend

kamaal111:
@yusukebe I have updated this PR to export the `cloneRawRequest` util from `/request`

src/request.ts
```diff
@@ -27,6 +27,46 @@ type BodyCache = Partial<Body & { parsedBody: BodyData }>
...
+ * )
+ * ```
+ */
+export const cloneRawRequest = async (req: HonoRequest): Promise<Request> => {
+  if (!req.raw.bodyUsed) {
+    return req.raw.clone()
+  }
+
+  return new Request(req.raw.url, {
+    method: req.raw.method,
```
yusukebe: Should we copy other properties, such as `credentials`, `mode`, and `single`?

```ts
// all properties
new Request(req.raw.url, {
  method: req.raw.method,
  headers: req.raw.headers,
  body: await req.blob(),
  mode: req.raw.mode,
  credentials: req.raw.credentials,
  cache: req.raw.cache,
  redirect: req.raw.redirect,
  referrer: req.raw.referrer,
  referrerPolicy: req.raw.referrerPolicy,
  integrity: req.raw.integrity,
  keepalive: req.raw.keepalive,
})
```

@kamaal111 @usualoma What do you think of this?

kamaal111: You are right, it might cause unexpected behavior if the user is expecting these properties.
I have explicitly declared and copied over all `RequestInit` properties except for `priority` and `window` and added tests to verify those are actually copied over correctly

src/validator/validator.test.ts
```diff
@@ -1207,3 +1208,168 @@ describe('Transform', () => {
...
+    expect(result.clonedCache).toBe(result.originalCache)
+    expect(result.clonedCredentials).toBe(result.originalCredentials)
+    expect(result.clonedMode).toBe(result.originalMode)
+    expect(result.clonedRedirect).toBe(result.originalRedirect)
+    expect(result.clonedReferrerPolicy).toBe(result.originalReferrerPolicy)
+    expect(result.cloned).toBe(true)
+    expect(result.payload).toMatchObject(testData)
+  })
+
+  it('Should allow the `cloneRawRequest` util to clone the request object without validation', async () => {
```
yusukebe: @kamaal111 

Do you think the following tests are really necessary?

* `Should allow the cloneRawRequest util to clone the request object without validation`
* `Should clone GET request with all properties`

I think these are not related to "validator" and we may not add these tests since cloning the request object is already tested. What do you think of it?

kamaal111: Its not necessary, I added to check if also GET requests and/or requests without body still will work as expected, but I already have those cases covered in `src/request.test.ts`

yusukebe:
Hey @kamaal111 

Let's go with this!

I'll merge this into the `next` branch for the next minor release `v4.10.0` (maybe released soon!). Thank you for the hard work!

kamaal111:
Awesome thank you for the support 🙏

Issue #4389, opened 2025-09-01
SSG: Support redirects


3w36zj6:
### What is the feature you are proposing?

Redirects are essential for SSG to ensure users and search engines are properly directed when URLs change. This supports smooth migrations, maintains SEO value, and prevents broken links.

### Feature Idea

When an endpoint that performs a redirect is defined as follows:

```tsx
import { Hono } from 'hono'
import { toSSG } from 'hono/ssg'
import fs from 'fs/promises'

const app = new Hono()

app.get("/old", (c) => {
  return c.redirect("/new")
})

app.get("/new", (c) => {
  return c.render(
    <>
      <title>New Page</title>
      <meta name='description' content='This is the new page.' />
      new page content.
    </>
  )
})

toSSG(app, fs)

export default app
```

`old.html` is generated as follows:

```html
<!DOCTYPE html>
<title>Redirecting to: /new</title>
<meta http-equiv="refresh" content="0;url=/new" />
<meta name="robots" content="noindex" />
<link rel="canonical" href="/new" />
<body>
  <a href="/new">Redirecting from <code>/old</code> to <code>/new</code></a>
</body>
```

Alternatively, an option can be added to `ToSSGOptions` in `toSSG` to define a mapping for redirects. This is inspired by Astro[^1].

```ts
export interface ToSSGOptions {
    dir?: string;
    beforeRequestHook?: BeforeRequestHook | BeforeRequestHook[];
    afterResponseHook?: AfterResponseHook | AfterResponseHook[];
    afterGenerateHook?: AfterGenerateHook | AfterGenerateHook[];
    concurrency?: number;
    extensionMap?: Record<string, string>;
    plugins?: SSGPlugin[];
    redirects?: Record<string, string>; // 👈
}
```

[^1]: https://docs.astro.build/en/reference/configuration-reference/#redirects

yusukebe:
@3w36zj6 

Interesting! I'd like to have the option to decide whether to create a redirect HTML or not.

3w36zj6:
> I'd like to have the option to decide whether to create a redirect HTML or not.

Understood. By the way, in this approach, what do you think the behavior should be if a redirect HTML is not output?

Currently, `toSSG()` outputs an empty text file for endpoints that perform a redirect. However, I don't think this behavior is desirable for most users. It would be more helpful if either no file is output, or if there is some kind of info or warning message provided.

yusukebe:
[...]

PR #4394, opened 2025-09-03
feat(ssg): add default plugin that defines the recommended behavior

3w36zj6:
cf. https://github.com/honojs/hono/issues/4389#issuecomment-3248655086

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/helper/ssg/ssg.ts
+++ b/src/helper/ssg/ssg.ts
@@ -349,6 +349,22 @@ export interface ToSSGAdaptorInterface<
 }
 
 /**
+ * The default plugin that defines the recommended behavior.
+ *
+ * @experimental
+ * `defaultPlugin` is an experimental feature.
+ * The API might be changed.
+ */
+export const defaultPlugin: SSGPlugin = {
+  afterResponseHook: (res) => {
+    if (res.status === 301 || res.status == 302) {
+      return false
+    }
+    return res
+  },
+}
+
+/**
  * @experimental
  * `toSSG` is an experimental feature.
  * The API might be changed.
--- a/src/helper/ssg/ssg.test.tsx
+++ b/src/helper/ssg/ssg.test.tsx
@@ -9,7 +9,13 @@ import {
   onlySSG,
   ssgParams,
 } from './middleware'
-import { defaultExtensionMap, fetchRoutesContent, saveContentToFile, toSSG } from './ssg'
+import {
+  defaultExtensionMap,
+  fetchRoutesContent,
+  saveContentToFile,
+  toSSG,
+  defaultPlugin,
+} from './ssg'
 import type {
   AfterGenerateHook,
   AfterResponseHook,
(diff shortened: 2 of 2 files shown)


3w36zj6:
Should I inject `defaultPlugin` as a default argument?

```diff
diff --git a/src/helper/ssg/ssg.ts b/src/helper/ssg/ssg.ts
index 68eb0a9b..33a40330 100644
--- a/src/helper/ssg/ssg.ts
+++ b/src/helper/ssg/ssg.ts
@@ -373,7 +373,7 @@ export const toSSG: ToSSGInterface = async (app, fs, options) => {
   let result: ToSSGResult | undefined
   const getInfoPromises: Promise<unknown>[] = []
   const savePromises: Promise<string | undefined>[] = []
-  const plugins = options?.plugins || []
+  const plugins = options?.plugins || [defaultPlugin]
   const beforeRequestHooks: BeforeRequestHook[] = []
   const afterResponseHooks: AfterResponseHook[] = []
   const afterGenerateHooks: AfterGenerateHook[] = []

```

yusukebe:
@3w36zj6 

> Should I inject `defaultPlugin` as a default argument?

Yes, I think so. Users must have the default behavior if they do not set anything.

3w36zj6:
I have made the changes.

src/helper/ssg/ssg.ts
```diff
@@ -348,6 +348,22 @@ export interface ToSSGAdaptorInterface<
...
 
+/**
+ * The default plugin that defines the recommended behavior.
+ *
+ * @experimental
+ * `defaultPlugin` is an experimental feature.
+ * The API might be changed.
+ */
+export const defaultPlugin: SSGPlugin = {
+  afterResponseHook: (res) => {
```
yusukebe, line 359: Btw, do you have any other things that you want in the `defaultPlugin`? For example, showing logs.

3w36zj6: I think having log output is fine, but I can't think of a log format that would be suitable as a standard. Also, since Vite outputs logs when using @hono/vite-ssg, it's probably better not to include log output by default. If logging is needed, users can implement their own hooks to suit their particular use case.

However, it would be a good idea to introduce log output as a use case in the documentation for SSG plugins.

I considered other possible features for the default plugin, but I couldn't think of anything besides redirects.

yusukebe: @3w36zj6, thank you for sharing your thoughts.

If there is only the redirect thing in the `defaultPlugin`, we may not need to create `defaultPlugin`, but instead, we can add the redirect thing directly. What do you think of it? I may be sorry that I said about `defaultPlugin`.

3w36zj6: Certainly, the current implementation of the default plugin is very minimal. However, by separating the default behavior as a plugin, I believe we can keep the implementation of `toSSG` minimal while also gaining various advantages such as future extensibility, API consistency, and ease of testing. For these reasons, I think it is best to keep the default plugin for now.

Of course, if you feel that not adopting the default plugin system is preferable, I will respect that as well.

yusukebe: @3w36zj6 

Okay. Let's go with keeping the current implementation using `defaultPlugin`!

yusukebe (approved):
LGTM!

yusukebe:
@3w36zj6 

Thank you! It's a new feature, I wonder whether we should include a minor release or a patch release. I'll merge this if it's decided.

3w36zj6:
By the way, I noticed that `c.notFound()` also generates plaintext.

```ts
app.get('/notfound', (c) => {
  return c.notFound()
})
```

```plaintext
404 Not Found
```

Popular static site hosting services support custom 404 pages, so I think that, like redirects, it is better for SSGs not to generate anything by default.

There may be concerns that the endpoint defining a custom 404 page could be ignored by SSG. However, if we define this page with a 200 status, I believe there is no issue. When accessing the 404 page directly in a production environment, the hosting service will automatically assign a 404 status.

What do you think?

yusukebe:
@3w36zj6

Nice catch! I also think SSGs should not generate files if the status code is 404.

But, after reconsidering, perhaps, we have to handle only a `200 OK` response.

4xx (like `404`) and 5xx (like `500`) are errors, so there's no need to output files. Additionally, 3xx are redirects, so they're unnecessary. For 2xx, I don't think we need anything other than `200`, like `201 Created`, `202 Accepted`, or `204 No Content`.

What do you think of this?

3w36zj6:
Indeed, it seems reasonable to generate only 200 responses by default.

May I modify the PR?

yusukebe:
@3w36zj6 

> May I modify the PR?

Yes. It's okay!

yusukebe:
@3w36zj6 

Thanks! Looks good to me! Please wait for the merging for a while.

PR #4395, opened 2025-09-03
fix(request): return empty string for empty catch-all param

amitksingh0880:
Fix: Ensure c.req.param() returns "" (empty string) for empty path parameters

Previously, when a route parameter matched an empty segment (e.g., / for /:remaining{.*}), c.req.param() returned undefined, which did not match its type signature (string).

This change updates the parameter extraction logic to return an empty string ("") instead of undefined for empty matches, ensuring runtime behavior matches the declared types and preventing potential runtime errors.

Also adds a test to verify this behavior.


usualoma:
Hi @amitksingh0880, thank you for creating the pull request! I believe the changes you made resolve the issue, but I have a few requests.

* For `c.req.param()` without arguments, I think it would be more consistent if it returned an empty string when no value is present
* Please add tests to “src/request.test.ts”
* (This is incidental, so either way is fine) I think #getDecodedParam could be written a bit more concisely

I think it would be good to make it like #4396. Could you please consider this?

amitksingh0880:
Hi @usualoma, thank you for the detailed feedback! I’ll look into the points you mentioned.

- Updating the behavior of `c.req.param()`
- Adding the requested tests, and reviewing #getDecodedParam for conciseness. 

I’ll also make sure it aligns with #4396. Will update the PR accordingly.

yusukebe:
@amitksingh0880 Thank you for the PR!

The followings fail:

* https://github.com/honojs/hono/pull/4395/files#diff-8ac13809c9886e994d1db33943de82df4d6c5d88b73fd0270c0189804ff565c2R928
* https://github.com/honojs/hono/pull/4395/files#diff-8ac13809c9886e994d1db33943de82df4d6c5d88b73fd0270c0189804ff565c2R1630

I think these should not fail. For example, please imagine the app:

```ts
app.get('/foo/:foo', (c) => {
  const bar = c.req.param('bar')
  return c.json({ bar })
})
```

When accessing `/foo/abc`. The value of `bar` should be `undefined` because the key (`bar`) is not found in params, but in this PR it returns `""`.

So the test you wrote:

* https://github.com/honojs/hono/pull/4395/files#diff-0a2d5d2d71866cfe92f65068de678d2952c3107d24a9feb8eb97a54c82a3dda0R67
* https://github.com/honojs/hono/pull/4395/files#diff-0a2d5d2d71866cfe92f65068de678d2952c3107d24a9feb8eb97a54c82a3dda0R84

Are wrong. Those should be like the following:

```diff
diff --git a/src/request.test.ts b/src/request.test.ts
index e04d4dfc..7a063deb 100644
--- a/src/request.test.ts
+++ b/src/request.test.ts
@@ -64,7 +64,7 @@ describe('Param', () => {
     ])

     expect(req.param('id')).toBe('123')
-    expect(req.param('name')).toBe('')
+    expect(req.param('name')).toBe(undefined)

     req.routeIndex = 1
     expect(req.param('id')).toBe('123')
@@ -81,7 +81,7 @@ describe('Param', () => {
     ])

     expect(req.param('id')).toBe('123')
-    expect(req.param('name')).toBe('')
+    expect(req.param('name')).toBe(undefined)

     req.routeIndex = 1
     expect(req.param('id')).toBe('456')
```

amitksingh0880:
Hey! @yusukebe thank you for the comment.

I will look into the described changes and update the same.

yusukebe:
I've updated the logic and tests myself. This may be good.

yusukebe:
@amitksingh0880 @usualoma I'll merge this now. Thank you!

exvillager:
so there  is a diffrence between "" and undefined in design architecture. like if i have define a route /user/:id and when i try to access c.req.params('id') if id isn't in req.url then it should give "" string but if i try to access c.req.params('name') this should give undefined because we never defined /:name in our this route. 
since this issue has been resolved so it's okay othrwise i was going to code this.

PR #4396, opened 2025-09-03
fix(request): return empty string from param() as is

usualoma:

- [ ] Add tests
- [ ] Run tests


yusukebe:
Hey @usualoma 

I've never looked at the details, and I don't know which is better, but the PR with maybe the same purpose was already created: https://github.com/honojs/hono/pull/4395

usualoma:
Hi @yusukebe, thanks for the comment. Sorry for the timing.
I created this PR to request additional fixes for #4395 and to communicate the details.

PR #4399, opened 2025-09-04
Fix `parseResponse` not parsing json in react native

lr0pb:
React native uses [`whatwg-fetch`](https://github.com/JakeChampion/fetch) polyfill to implement fetch and currently `parseResponse` checks existence of `response.body` in order to parse it, but this polyfill does not implement `body` property, so check fails. However it implements `_bodyInit` (https://github.com/JakeChampion/fetch/blob/main/fetch.js#L238) and some other internal properties we can check to ensure response body exists.

I added check of `_bodyInit` and it started parsing json responses as expected. I do not add tests for this case, but if its desirable I can do this.

- [ ] Add tests
- [ ] Run tests


yusukebe:
Hi @lr0pb 

I added the test myself, and this looks good. Thank you!

PR #4402, opened 2025-09-08
Add tests of JWK token verification

buckett:
Check that by default we verify the token timestamp attributes and when specified an issuer (iss) is also verified.

- [X] Add tests
- [X] Run tests


yusukebe (approved):
LGTM!

yusukebe:
@buckett 

Thank you for the PR. Looks good!

PR #4405, opened 2025-09-11
fix(types): JSONParsed infer unknown values

### LINKED ISSUE #4368: RPC client infers object with an `unknown` property as `never`

4.9.4

Node

Hi! Great library. Thanks for all your work. I've run into the following behavior that's tripped me up a couple of times. 

When a route returns an object with a property of type `unknown`, hono client infers that return type as `never`.

While I usually don't _mean_ to return an `unknown` property, this particular behavior was unexpected to me and pretty hard to debug. I started gradually removing parts of my application until I found the problem property. 

Is this a bug or maybe something to document?

Reproduction:

```typescript
import { Hono } from "hono";

const app = new Hono()
  .get("/expected", (c) => {
    const data: { foo: string; bar: string } = { foo: "hello", bar: "world" };
    return c.json(data);
  })
  .get("/unexpected", (c) => {
    // Return an object with a property of type unknown
    const data: { foo: string; bar: unknown } = { foo: "hello", bar: 123 };
    return c.json(data);
  });

type AppType = typeof app;

const client = hc<AppType>("/");

async function testClient() {
  const expectedRes = await client.expected.$get();
  if (expectedRes.ok) {
    const data = await expectedRes.json(); // Type: { foo: string; bar: string }
    console.log(data.foo); // ✅ Works fine
  }

  const unexpectedRes = await client.unexpected.$get();
  if (unexpectedRes.ok) {
    const data = await unexpectedRes.json(); // Type: never ❌
    console.log(data.foo); // TypeScript error: Property 'foo' does not exist on type 'never'
  }
}
```

https://www.typescriptlang.org/play/?#code/PTAEFkEsDtIWwIYBtQCcCmAHVB7AJgK4DGALpDtKDgGagASFOoASgAoDCoB06AHpulLo8oEgE8BoGNXQZoRdACgQoAOoALdJQRocBEulDqE0PEllp0JAqmgBnUCaoAjAFaCSoAO6QS6x6DYOAKo4qISwo4OAOTcANbQOF7Q0QA0ymB+hmycREiQWp7Ssg5ZlnaYFHaG4pIIMTwAbrLRUvYGCCI0opqOpATI4QIAdIqK8JWhoADe9IygAL6g1LhwoABE6ozrANzjcJOes+pEi8urG1uJwHkF0CS7YyrsGAgGAXYT5o6YmN6+-h0uH0NWMngw1lsDicODcHlARAoJAQMBgAHNQLFoAkkilFIj2j8-gBeUA8LxzRIACgAlIpQKBhmirFT1sA+AIhHh1qlQFSiDTQMSAHwzekMhFVTx4N4IABcM2WOBwCrsJFQ6J2oGcCFQqvV6LOpNm1GVCs26CQSBwPO1uvNXhwqCQ3MWewllkhlCIw1cdgoVJlyJp7sWdIZTJZbO4HI8wlt-MFIrFHpUzCsNm0lFh7lI-z8ASCITC3VqhniiWS4oZBLVoCD8sVppVoDVGugaK1Or1XGxlcoS2NSpbFqtNt53YVAEYAEwAZjd1c9mYRvv90EDspD4oW28UZdAAEFfgAVCJCoboboIX57fFShH5QoXk4AHmPmDPAmFrOA6z39RiPIyzcKQ5CUAYarsE+9y0imkqErGXLpg4pIIF4KKeLchTDEhBh4MMAAkzIkLSoaQLQVJ4cIKHDDgcSCtMS61tKsoXuhmGgNReC0X6AYhqAKhfugCommarYGh2Xb2hJ7YYgszFVDg5jDNaaKbsiwzNgJKiAKDkahOnEDjUDASgMgp4osb23EoexGG+I+dwkMMMb8HGBHESy24MhRfKuZy+G0fRjGKYSDZ2Zx-nubx65kYJYDCQqTQWIAMuShf6KlqRpCBacqOkJREADKRAapgniyLgPasLgxZiJizatHgODoA4iTlbwkB1hQl6YslqDRDuigKUAA

For the route that returns `{ foo: string; bar: unknown }`, I'd expect the inferred client type to be `{ foo: string; bar: unknown }`

### What do you see instead?

Instead, the client inferred type is `never`

### Additional information

_No response_

BarryThePenguin:
Changes `JSONParsed` so it will infer `unknown` values as `JSONValue`.

I also included handling for `bigint` so `JSONParsed` will be `never`, indicating that an error will be thrown.

fixes #4368

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -194,21 +194,12 @@ interface JSONRespond {
  * @template T - The type of the JSON value or simplified unknown type.
  * @template U - The type of the status code.
  *
- * @returns {Response & TypedResponse<SimplifyDeepArray<T> extends JSONValue ? (JSONValue extends SimplifyDeepArray<T> ? never : JSONParsed<T>) : never, U, 'json'>} - The response after rendering the JSON object, typed with the provided object and status code types.
+ * @returns {Response & TypedResponse<JSONParsed<T>, U, 'json'>} - The response after rendering the JSON object, typed with the provided object and status code types.
  */
 type JSONRespondReturn<
   T extends JSONValue | SimplifyDeepArray<unknown> | InvalidJSONValue,
   U extends ContentfulStatusCode
-> = Response &
-  TypedResponse<
-    SimplifyDeepArray<T> extends JSONValue
-      ? JSONValue extends SimplifyDeepArray<T>
-        ? never
-        : JSONParsed<T>
-      : never,
-    U,
-    'json'
-  >
+> = Response & TypedResponse<JSONParsed<T>, U, 'json'>
 
 /**
  * Interface representing a function that responds with HTML content.
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -12,6 +12,16 @@ import { validator } from '../validator'
 import { hc } from './client'
 import type { ClientResponse, InferRequestType, InferResponseType } from './types'
 
+class SafeBigInt {
+  unsafe = BigInt(42)
+
+  toJSON() {
+    return {
+      value: '42n',
+    }
+  }
+}
+
 describe('Basic - JSON', () => {
   const app = new Hono()
 
@@ -51,6 +61,8 @@ describe('Basic - JSON', () => {
     .get('/hello-not-found', (c) => c.notFound())
     .get('/null', (c) => c.json(null))
     .get('/empty', (c) => c.json({}))
+    .get('/bigint', (c) => c.json({ value: BigInt(42) }))
+    .get('/safe-bigint', (c) => c.json(new SafeBigInt()))
 
   type AppType = typeof route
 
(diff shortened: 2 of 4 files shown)


src/context.ts
```diff
@@ -194,21 +194,12 @@ interface JSONRespond {
...
-  TypedResponse<
-    SimplifyDeepArray<T> extends JSONValue
-      ? JSONValue extends SimplifyDeepArray<T>
-        ? never
-        : JSONParsed<T>
-      : never,
-    U,
-    'json'
-  >
+> = Response & TypedResponse<JSONParsed<T>, U, 'json'>
```
BarryThePenguin, line 202: I think this change could be a seperate PR.

From what I understand `JSONParsed` already contains types that do the same thing as `SimplifyDeepArray`, specifically

```ts
...
  : T extends ReadonlyArray<unknown>
  ? { [K in keyof T]: JSONParsed<InvalidToNull<T[K]>> }
  : ...
```

compared with `SimplifyDeepArray`

```ts
type SimplifyDeepArray<T> = T extends any[]
  ? { [E in keyof T]: SimplifyDeepArray<T[E]> }
  : Simplify<T>
```

yusukebe: Indeed. We can remove `SimplifyDeepArray`. For now, I'll merge this PR.

yusukebe:
@BarryThePenguin 

Thank you! Looks good.

PR #4406, opened 2025-09-12
refactor(types): remove SimplifyDeepArray from json types

BarryThePenguin:
Simplifies return type of `ctx.json()` by removing `SimplifyDeepArray`

I may be missing something here.. but it appears to still work 🙈

The simplest way I could understand it was that

```ts
expectTypeOf<SimplifyDeepArray<unknown>>().toEqualTypeOf<{}>()
```

Or more specifically

```ts
type Old = Simplify<JSONValue | SimplifyDeepArray<unknown> | InvalidJSONValue>

type New = Simplify<JSONValue | {} | InvalidJSONValue>

type Actual = string | number | boolean | symbol | (JSONObject | JSONArray | JSONPrimitive)[] | {} | {
    [x: string]: object | JSONObject | JSONArray | JSONPrimitive | InvalidJSONValue;
} | {} | null | undefined

expectTypeOf<Old>().toEqualTypeOf<Actual>()

expectTypeOf<New>().toEqualTypeOf<Actual>()
```

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -13,13 +13,7 @@ import type { ResponseHeader } from './utils/headers'
 import { HtmlEscapedCallbackPhase, resolveCallback } from './utils/html'
 import type { ContentfulStatusCode, RedirectStatusCode, StatusCode } from './utils/http-status'
 import type { BaseMime } from './utils/mime'
-import type {
-  InvalidJSONValue,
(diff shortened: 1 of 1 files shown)


yusukebe:
@BarryThePenguin 

They are really not necessary! Thank you.

PR #4423, opened 2025-09-23
feat(helper/route): enable to get route path at specific index

### LINKED ISSUE #4196: Support get real route path in middleware

Now
```ts
app.use(async (c, next) => {
    console.log(c.req.method, c.req.routePath, c.req.path, c.req.url)
    //             GET           /*            /user/1      http://127.0.0.1:3000/user/1

}

app.get("/user/:id", ...)

// and curl
// curl http://127.0.0.1:3000/user/1
```
Want to
```ts
    console.log(c.req.routePath)
    //          /user/:id
```

usualoma:
fixes #4196

Currently, only routes associated with handlers can be retrieved. This PR enables passing an optional `index` parameter (which has the same meaning as `Array.prototype.at(index)`), allowing `routePath(c, -1)` to retrieve the route that matched at the end.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/helper/route/index.ts
+++ b/src/helper/route/index.ts
@@ -37,23 +37,32 @@ export const matchedRoutes = (c: Context): RouterRoute[] =>
  * Get the route path registered within the handler
  *
  * @param {Context} c - The context object
+ * @param {number} index - The index of the root from which to retrieve the path, similar to Array.prototype.at(), where a negative number is the index counted from the end of the matching root. Defaults to the current root index.
  * @returns The route path registered within the handler
  *
  * @example
  * ```ts
  * import { routePath } from 'hono/route'
  *
+ * app.use('*', (c, next) => {
+ *   console.log(routePath(c)) // '*'
+ *   console.log(routePath(c, -1)) // '/posts/:id'
+ *   return next()
+ * })
+ *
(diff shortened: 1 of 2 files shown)


usualoma:
Hi @yusukebe 

What do you think about this?

src/helper/route/index.ts
```diff
@@ -37,23 +37,32 @@ export const matchedRoutes = (c: Context): RouterRoute[] =>
...
+ * })
+ *
  * app.get('/posts/:id', (c) => {
  *   return c.text(routePath(c)) // '/posts/:id'
  * })
  * ```
  */
-export const routePath = (c: Context): string => matchedRoutes(c)[c.req.routeIndex].path
+export const routePath = (c: Context, index?: number): string =>
+  matchedRoutes(c).at(index ?? c.req.routeIndex)?.path ?? ''
```
yusukebe, line 59: It isn't easy to define the behavior if it cannot find the route. I think both `''` and `undefined` are okay. In this case, `''` is better because this way does not break that `routePath` returns value is `string` (not `string | undefined`). So this is good.

yusukebe:
Hi @usualoma 

This is a good idea! Looks good. The change is only adding the one option, so we can include this in the next patch release. Thanks!

PR #4424, opened 2025-09-23
fix(compose): preserve `c.req.routeIndex`  across `next()`

usualoma:
Reference: https://github.com/honojs/hono/issues/4196#issuecomment-3317581222

This is a bug that has existed since quite early on. After calling `next()`, `param()` and `routePath()` were not returning the correct values. I believe this needs to be fixed.

Although this usage pattern is not very common, since it involves a change to the core behavior, I think it would be appropriate to merge this change during the next major version update.

### An app that can reproduce the bug

```ts
const app = new Hono()

app.use('/:all{.*}', async (c, next) => {
  console.log(routePath(c))
  await next()
  console.log(routePath(c)) // should be '/:all{.*}'
  console.log(c.req.param('all')) // should be the request path
})

app.get('/posts/:id', (c) => {
  return c.text(routePath(c))
})

export default app
```

### Expected result

* Outputs `/:all{.*}`, and the request path

### Actual result

* Outputs `/posts/:id`, and `undefined`

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/compose.test.ts
+++ b/src/compose.test.ts
@@ -16,7 +16,9 @@ describe('compose', () => {
 
   const a = async (c: Context, next: Next) => {
     c.set('log', 'log')
+    c.set('routeIndex-before-next', c.req.routeIndex)
     await next()
+    c.set('routeIndex-after-next', c.req.routeIndex)
   }
 
   const b = async (c: Context, next: Next) => {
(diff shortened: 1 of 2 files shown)


usualoma:
Hi @yusukebe
Would you please review this?

yusukebe (approved):
LGTM!

yusukebe:
@usualoma 

Thank you. Looks good to me!

> Although this usage pattern is not very common, since it involves a change to the core behavior, I think it would be appropriate to merge this change during the next major version update.

It's hard to decide. I want to release this before the major version update, but we don't know if it will be soon or not. As you said, some users may be damaged. For now, keep it not merged.

PR #4425, opened 2025-09-23
feat(request): return new Request with cached body in raw getter

yusukebe:
This PR will fix the problem that we wanted to resolve with #4382.

### Problem

The following code will cause the error:

```ts
import { Hono } from 'hono'

const app = new Hono()

app.post('/', async (c) => {
  await c.req.raw.json() // Body already used!
  return c.json(data)
})

export default app
```

CleanShot 2025-09-23 at 17 17 54@2x

This is because when you do `c.req.raw`, the body to be used is already consumed. This is explained in https://github.com/honojs/hono/pull/4382#issue-3370569516.

### Solution

In this PR, it will create a new Request object with a cached body when `c.req.raw` is called. The reason why it does not use getter/setter and uses `Object.defineProperty` is that if it were to use getter/setter, it would fail with the spread syntax `{...c.req}`.

### Caveat

The code size will be increased.

The performance may degrade, but it's small.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/request.test.ts
+++ b/src/request.test.ts
@@ -267,6 +267,24 @@ describe('Body methods with caching', () => {
     expect(await req.text()).toEqual(text)
   })
 
+  test('req.raw.json() should work after req.json() is called', async () => {
+    const data = { foo: 'bar' }
+    const req = new HonoRequest(
+      new Request('http://localhost', {
+        method: 'POST',
+        body: JSON.stringify(data),
+      })
+    )
+
+    // First call to req.json() - consumes the body
+    const firstResult = await req.json()
+    expect(firstResult).toEqual(data)
+
+    // Second call to req.raw.json() - should work with cached body
+    const secondResult = await req.raw.json()
+    expect(secondResult).toEqual(data)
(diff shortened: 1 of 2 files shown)


src/request.test.ts
```diff
@@ -267,6 +267,24 @@ describe('Body methods with caching', () => {
...
+      })
+    )
+
+    // First call to req.json() - consumes the body
+    const firstResult = await req.json()
+    expect(firstResult).toEqual(data)
+
+    // Second call to req.raw.json() - should work with cached body
+    const secondResult = await req.raw.json()
+    expect(secondResult).toEqual(data)
```
kamaal111, line 285: Can you also verify that it still works as expected when try to get the body for the third time?

To switch it up a bit we can also use `.text` which would still result in getting the cached body and constructing a new Request object

kamaal111 (commented):
In HonoRequest, where we don't need to access the raw body, can we use the private `_raw` instead?

This might improve the performance slightly

PR #4433, opened 2025-09-28
Fix #4427 type regression by removing non-public export

### LINKED ISSUE #4427: regression in `Context`/`c.json()` type inference in `v4.9.8`

4.9.8

Bun (1.2.16), TypeScript (^5.8.2)

```typescript
import type { Context } from "hono";

export function test(c: Context) {
  return c.json("test", 200);
}
```

For this to typecheck/for TS language server to be able to infer the returned type without annotation here.

### What do you see instead?

In `4.9.8`, this gives the following type error:

```
The inferred type of 'test' cannot be named without a reference to '../../../../node_modules/hono/dist/types/context'. This is likely not portable. A type annotation is necessary.ts(2742)
```

### Additional information

Reverting to `4.9.7` allows the TS language server to infer the returned type correctly.

aantthony:
I think this fixes https://github.com/honojs/hono/issues/4427

Example test code:
```ts
import { createFactory } from 'hono/factory';
export const handlers = factory.createHandlers(c =>
	c.json({ ok: true }),
);

```

With the type exported (since v4.9.8), I was getting the error:

```
ts(2742): The inferred type of 'handlers' cannot be named without a reference to '../node_modules/hono/dist/types/context.js'. This is likely not portable. A type annotation is necessary.
```

By avoiding the export, there is no type error. Removing this export should not create any issues as far as I know, because it is not imported anywhere or publicly exposed outside hono itself.

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -190,7 +190,7 @@ interface JSONRespond {
  *
  * @returns {Response & TypedResponse<JSONParsed<T>, U, 'json'>} - The response after rendering the JSON object, typed with the provided object and status code types.
  */
-export type JSONRespondReturn<
+type JSONRespondReturn<
   T extends JSONValue | {} | InvalidJSONValue,
   U extends ContentfulStatusCode
 > = Response & TypedResponse<JSONParsed<T>, U, 'json'>


BarryThePenguin:
Whoops.. looks like I inadvertently changed that in #4406

Looks good 👍🏻

yusukebe (approved):
LGTM!

yusukebe:
@aantthony 

Thank you!

PR #4437, opened 2025-10-02
fix(aws-lambda): sanitize non-ASCII header values to prevent ByteString errors

### LINKED ISSUE #4432: aws-lambda: handle ALB lambda event that has non-ascii header values

4.7.4

AWS Lambda

The hono application is running as a lambda behind an AWS ALB

```
Cloudflare -> ALB -> Hono (lambda)
```

Cloudflare adds UTF-8 encoded `cf-ipcity` geo header that could have non-ascii characters 

https://developers.cloudflare.com/rules/transform/managed-transforms/reference/#encoding-of-non-ascii-header-values

The ALB is decoding this correctly and passing it to the `aws-lambda` handler.

But since the header has non-ascii characters it fails with an error here:

https://github.com/honojs/hono/blob/main/src/adapter/aws-lambda/handler.ts#L417

```
TypeError: Cannot convert argument to a ByteString because the character at index 2 has a value of 299 which is greater than 255.
  at webidl.converters.ByteString (node:internal/deps/undici/undici:3889:17)",
  at _Headers.set (node:internal/deps/undici/undici:8921:35)",
  at ALBProcessor.getHeaders (file:///var/task/index.mjs:201:19)",
```

Expecting Hono aws-lambda to handle ALB lambda event with non-ascii header values

### What do you see instead?

Runtime error when undici tries to set headers.

### Additional information

_No response_

yusukebe:
Fixes #4432

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/adapter/aws-lambda/handler.test.ts
+++ b/src/adapter/aws-lambda/handler.test.ts
@@ -291,78 +291,19 @@ describe('EventProcessor.createRequest', () => {
   })
 
   describe('non-ASCII header value processing', () => {
-    it('Should correctly process ALB request with non-ASCII header values', async () => {
-      const event: LambdaEvent = {
-        headers: {
-          'content-type': 'application/json',
-          'cf-ipcity': 'São Paulo', // City name with non-ASCII characters
-          'x-custom-city': 'Москва', // Cyrillic characters
-        },
-        httpMethod: 'GET',
-        path: '/',
-        body: null,
-        isBase64Encoded: false,
-        requestContext: {
-          elb: {
-            targetGroupArn:
-              'arn:aws:elasticloadbalancing:us-east-1:123456789012:targetgroup/lambda-279XGJDqGZ5rsrHC2Fjr/49e9d65c45c6791a',
-          },
-        },
-      }
-
-      const processor = getProcessor(event)
-      const request = processor.createRequest(event)
-
-      expect(request.method).toBe('GET')
-      expect(request.url).toBe('https://undefined/')
-      expect(request.headers.get('cf-ipcity')).toBeDefined()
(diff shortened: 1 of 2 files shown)


ajaishankar:
Hi @yusukebe I wanted to refer to the Cloudflare documentation

https://developers.cloudflare.com/rules/transform/managed-transforms/reference/#encoding-of-non-ascii-header-values

> Cloudflare always converts non-ASCII characters to UTF-8 (using hexadecimal character representation) in HTTP request and response header values.
>
> For example, the header value São Paulo would be encoded as S\u00c3\u00a3o Paulo.

What I _think_ this means is Cloudflare is supposed to send `São Paulo` as `'S%C3%A3o%20Paulo'` (hexadecimal).

When I raised the issue our ALB was receiving this string `"Srīperumbūdūr"` for `cf-ipcity` header from Chennai, India. 

Maybe it's a bug with Cloudflare as it did not encode this city?

As for the fix, can't Hono take an approach similar to Cloudflare?

- If there is a char > 255 simply `encodeURIComponent` the value
- Leave it to the client code to try `decodeURIComponent` if say it finds a % in the value?

Not sure if this is a good approach, but won't decoding as `latin1` as in this PR return a wrong string?

yusukebe:
Hi @ajaishankar 

I don't know whether it's a bug with Cloudflare. But how about the current code? I added the change https://github.com/honojs/hono/pull/4437/commits/c51c060398deb7a2a71f82875cad35790da71d29, simplified and used `encodeURIComponent`.

ajaishankar:
Thanks @yusukebe! This looks good 👍

yusukebe:
@ajaishankar Thank you for reviewing. Merging.

PR #4453, opened 2025-10-08
[PoC] feat: introduce hono command

usualoma:
### What is this?

Introduce the `hono` command to easily start a web server with a single shell one-liner.

The middleware included with the main package is imported by default and can be used simply by specifying it from the command line. This is where Hono's “Batteries Included” feature comes into play.

```bash
$ npx hono \
  --use 'logger()' \
  --use 'basicAuth({username: "hono", password: "hono"})' \
  --use 'serveStatic({})' 
Listening on http://localhost:3000
```

```bash
$ npx hono \
  --use 'basicAuth({username: "hono", password: "hono"})' \
  --use '(c) => proxy(`https:/ramen-api.dev${new URL(c.req.url).pathname}`)'
Listening on http://localhost:3000
```

I think it would be interesting to also support `--get` and `--post`, but it's difficult to determine what specifications would make them useful.

### Is it appropriate to include this in the `hono` package?

I feel like a separate package would be more appropriate. (I created a PR in this repository for sharing the PoC)

```
$ npx @hono/one-liner-server
# or
$ npx honos # same series as `honox`
```

Well, it's better to keep it short, and there's the advantage that `hono` is easy to remember when you boot it up after a long time.

### Support for multiple runtimes

If Deno or Bun become available, I would like to enable launching with those runtimes in the future.

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/bin/serve.ts
+++ b/bin/serve.ts
@@ -6,13 +6,7 @@ import { Hono } from '../src/hono'
 import { basicAuth } from '../src/middleware/basic-auth'
 import { logger } from '../src/middleware/logger'
 // import all the middleware and helpers
-
-[
-  basicAuth,
-  logger,
-  serveStatic,
-  proxy,
-].forEach((f) => {
+;[basicAuth, logger, serveStatic, proxy].forEach((f) => {
   if (typeof f === 'function') {
     // useless process to avoid being deleted by bundler
   }
@@ -28,7 +22,10 @@ const { values } = parseArgs({
 const app = new Hono()
 
 values.use?.forEach((use) => {
-  app.use((eval(use)))
+  app.use(async (c, next) => {
+    const evalRes = eval(use)
+    return typeof evalRes === 'function' ? evalRes(c, next) : evalRes
+  })
 })
 
 serve(


usualoma:
Hi @yusukebe, what are your thoughts?

yusukebe:
Hi @usualoma 

I've never seen the details of this PR, but I'll make a quick response. I'm also thinking `hono` command!! So, how about these sub-commands?

```
$ hono build
$ hono serve
$ hono deploy
$ hono doc
$ hono ai // ??
```

usualoma:
With 1627375, it is now possible to return a Response object directly as follows.

```bash
$ npx hono \
  --use 'basicAuth({username: "hono", password: "hono"})' \
  --use 'proxy(`https:/ramen-api.dev${new URL(c.req.url).pathname}`)'
```

usualoma:
Hi @yusukebe, thanks for your response.

I also thought a subcommand would be helpful, but considering the current restriction of “no external dependencies,” I believed it would be difficult to implement the necessary features.

(That said, this PR does have the subtlety of depending on `devDependencies` by including the contents of @hono/node-server into `dist`.)

yusukebe:
Hi @usualoma 

Thank you for the updates. I think (currently) creating an independent project for the command is good. The naming may be good if `honos` or `honoc` (`hono command`). Then, I prefer to add sub-commands. We can create a sub-command `exec` as you suggested in this PR:

```
$ npx honoc exec\
  --use 'logger()' \
  --use 'serveStatic({})' 
```

And I came up `$ hono repl`. Isn't it good? We don't hurry to line up all sub-commands. We can start one by one.

yusukebe:
> the current restriction of “no external dependencies,” I believed it would be difficult to implement the necessary features.

I also realize we have the problem that the command may depend on many libraries. Should be considered.

usualoma:
!! > `repl`
I can't quite picture what it could do yet, but it sounds really interesting.

Alright, let's think about it in `honoc`. I'll close this PR for now.

PR #4456, opened 2025-10-11
refactor:  internal structure of `PreparedRegExpRouter` for optimization and added tests

usualoma:
In this PR, I refactored the internal structure of PreparedRegExpRouter for optimization and added tests. There are no changes to its behavior.

### Description of individual commits

* 636d0099874a10186f637448414d6d3be61fe563 : The `undefined` values in the array were being converted to `null` by `JSON.stringify`, causing type mismatches upon restoration. Therefore, I modified the code to avoid using `undefined`.
* 8206632c4cf62eafcbcede91571fefa748518c7c : The wildcard route was not being used even after being registered in the `relocationMap`, so this PR skips its registration.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/router/reg-exp-router/prepared-router.test.ts
+++ b/src/router/reg-exp-router/prepared-router.test.ts
@@ -1,3 +1,4 @@
+import { METHOD_NAME_ALL } from '../../router'
 import { runTest } from '../common.case.test'
 import { buildInitParams, serializeInitParams, PreparedRegExpRouter } from './prepared-router'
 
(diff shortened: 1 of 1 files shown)


usualoma:
Hi @yusukebe 

Would you please review this?
Since this is only refactoring with no functional changes, there's no rush.

src/router/reg-exp-router/prepared-router.ts
```diff
@@ -125,17 +151,10 @@ export const buildInitParams: (params: {
...
-    ;(matcher[0] as RegExp & { toJSON: () => string }).toJSON = function () {
-      return `@${this.toString()}@`
-    }
-  }
-  // unescape the regular expression so that it can be deserialized with `eval`.
-  const matchersStr = JSON.stringify(matchers).replace(/"@(.+?)@"/g, (_, str) =>
-    str.replace(/\\\\/g, '\\')
-  )
+  // Embed the regular expression as a result of `toString()` so that it can be evaluated as JavaScript.
+  const matchersStr = JSON.stringify(matchers, (_, value) =>
```
yusukebe, line 155: Using the replacer of `JSON.stringify` is nice!

yusukebe:
@usualoma 

Thank you!! Looks good. Merging.

PR #4458, opened 2025-10-12
refactor: use protected methods instead of computed properties to allow `tree shaking`

usualoma:
The presence of computed properties led to the determination that side effects were possible, resulting in larger bundled code.

#### file

```ts
import { PreparedRegExpRouter } from './src/router/reg-exp-router'
console.log(PreparedRegExpRouter)
```

#### bundle size

main branch

```
$ npx esbuild --bundle prr.ts | wc
     464    1619   15633
```

refactor/prepared-computed-property branch

```
$ npx esbuild --bundle import.ts | wc
      92     262    2790
```

- [x] Run tests

### DIFF
--- a/src/router/reg-exp-router/prepared-router.ts
+++ b/src/router/reg-exp-router/prepared-router.ts
@@ -1,7 +1,7 @@
 import type { ParamIndexMap, Result, Router } from '../../router'
 import { METHOD_NAME_ALL } from '../../router'
 import type { HandlerData, Matcher, MatcherMap, StaticMap } from './matcher'
-import { match, buildAllMatchersKey, emptyParam } from './matcher'
+import { match, emptyParam } from './matcher'
 import { RegExpRouter } from './router'
 
 type RelocateMap = Record<string, [(number | string)[], ParamIndexMap | undefined][]>
@@ -62,7 +62,7 @@ export class PreparedRegExpRouter<T> implements Router<T> {
     }
   }
 
-  [buildAllMatchersKey](): MatcherMap<T> {
+  protected buildAllMatchers(): MatcherMap<T> {
     return this.#matchers
   }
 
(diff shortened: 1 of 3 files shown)


usualoma:
@yusukebe
Would you please review this?

src/router/reg-exp-router/prepared-router.ts
```diff
@@ -1,7 +1,7 @@
 import type { ParamIndexMap, Result, Router } from '../../router'
 import { METHOD_NAME_ALL } from '../../router'
 import type { HandlerData, Matcher, MatcherMap, StaticMap } from './matcher'
-import { match, buildAllMatchersKey, emptyParam } from './matcher'
+import { match, emptyParam } from './matcher'
 import { RegExpRouter } from './router'
```
yusukebe, line 5: This is not related to this PR, but it uses only the type so that you can write like this:

```ts
import type { RegExpRouter } from './router'
```

yusukebe:
@usualoma 

I confirmed this change affects the new project and reduces the file size without RegExpRouter. Thanks! I've left a comment.

src/router/reg-exp-router/prepared-router.ts
usualoma: Thank you for your comment.
We're using it as a base class here!

If bundling doesn't produce the intended result, I think it's possible to separate the files to minimize the impact.

https://github.com/honojs/hono/pull/4458/files/0c98e13168f338ce40553cbb7dd97d7f0b3fe131#diff-05f98af8ac89c7dd686b79c41d8167478c790c0051245d79c3b0f0554351dc63R75

yusukebe: @usualoma Ah, class! Sorry, my misunderstanding. It's okay as it is.

yusukebe:
@usualoma 

Looks good! Thank you for the great work.

PR #4459, opened 2025-10-14
fix(proxy): Correct hop-by-hop header handling per RFC 9110

sugar-cat7:
The code references [RFC 2616](https://www.rfc-editor.org/info/rfc2616), which has been obsoleted by RFC 7230 (2014) and then RFC 9110 (2022).

**Fixed hop-by-hop headers list**
- Changed `'trailers'` to `'trailer'` in `hopByHopHeaders` array
  - RFC 2616 originally listed "Trailers" but this was an error, corrected in [RFC 2616 Errata ID 4522](https://www.rfc-editor.org/errata/eid4522) (Status: Verified)
  - Subsequent RFCs (7230, 9110) use `Trailer` (singular)
- Added missing `'upgrade'` header
  - Required per [RFC 9110 Section 7.6.1](https://datatracker.ietf.org/doc/html/rfc9110#section-7.6.1)
 
**Verified Connection header parsing**
- [RFC 9110 Section 7.6.1](https://datatracker.ietf.org/doc/html/rfc9110#section-7.6.1):
  > Intermediaries MUST parse a received Connection header field before a message is forwarded and, for each connection-option in this field, remove any header or trailer field(s) from the message with the same name as the connection-option

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/helper/proxy/index.ts
+++ b/src/helper/proxy/index.ts
@@ -3,6 +3,7 @@
  * Proxy Helper for Hono.
  */
 
+import { HTTPException } from '../../http-exception'
 import type { RequestHeader } from '../../utils/headers'
 
 // https://datatracker.ietf.org/doc/html/rfc2616#section-13.5.1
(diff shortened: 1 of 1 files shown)


src/helper/proxy/index.ts
```diff
@@ -38,6 +39,17 @@ const buildRequestInitFromRequest = (
...
 
   const headers = new Headers(request.headers)
+
+  // https://datatracker.ietf.org/doc/html/rfc9110#section-7.6.1
+  // Parse Connection header and remove listed headers (MUST per RFC 9110)
+  const connectionValue = headers.get('connection')
+  if (connectionValue) {
+    connectionValue
+      .split(',')
+      .map((h) => h.trim().toLowerCase())
```
usualoma: The header name is normalized within `headers.delete()`, so I believe `.toLowerCase()` is unnecessary.

```suggestion
      .map((h) => h.trim()
```

sugar-cat7:
> > Verified Connection header parsing
> 
> Thank you for pointing that out! I wasn't aware of that specification.
> 
> However, if headers like the following can be specified in the connection value and deleted, it seems this could lead to issues classified as "Hop-by-Hop Header Injection".
> 
> * X-Forwarded-For
> * X-Real-IP
> 
> While it's desirable to have functionality that respects the "Connection" header, I feel it should only be enabled explicitly as an option.


Thank you for pointing this out! You're absolutely right about the Hop-by-Hop Header Injection risk.

I've added a `strictConnectionProcessing` option (default: `false`):
- **Default behavior**: Ignores the Connection header entirely to prevent injection attacks
- **Strict mode** (`true`): Processes Connection header per RFC 9110 (use only in trusted environments)

sugar-cat7:
> Hi @sugar-cat7, thanks for your help. I believe `strictConnectionProcessing` should be fine!
> 
> Sorry, just one more thing when you have time.
> 
> When `strictConnectionProcessing` is enabled, the following request results in an "Internal Server Error". Since this option is mainly used in trusted environments, we generally don't need to worry about invalid values. However, configuration errors could still cause problems. In those cases, I think we should either return a 400 error or just ignore it.
> 
> ```
> $ curl -H 'connection: a a' http://localhost:3000/internal-proxy/x
> ```

@usualoma 
Thank you for pointing this out!

* Added header (token) validation based on RFC 9110 Section 5.6.2, returning 400 on validation failure.

src/helper/proxy/index.ts
```diff
@@ -148,11 +191,20 @@ const preprocessRequestInit = (requestInit: RequestInit): RequestInit => {
...
+  const {
+    raw,
+    customFetch,
+    strictConnectionProcessing = false,
+    ...requestInit
+  } = proxyInit instanceof Request ? { raw: proxyInit } : proxyInit ?? {}
+
+  const result = processConnectionHeader(raw, strictConnectionProcessing)
+  if (!result.isSuccess) {
+    return new Response(result.message, { status: 400 })
```
usualoma: Sorry for not explaining earlier, but I want to throw an `HTTPException`.

https://github.com/honojs/hono/blob/4b796cfb0b105418bbf806050e788741f2739125/src/validator/validator.ts#L81

src/helper/proxy/index.ts
```diff
@@ -148,11 +191,20 @@ const preprocessRequestInit = (requestInit: RequestInit): RequestInit => {
...
-  const { raw, customFetch, strictConnectionProcessing, ...requestInit } =
-    proxyInit instanceof Request ? { raw: proxyInit } : proxyInit ?? {}
+  const {
+    raw,
+    customFetch,
+    strictConnectionProcessing = false,
+    ...requestInit
+  } = proxyInit instanceof Request ? { raw: proxyInit } : proxyInit ?? {}
+
+  const result = processConnectionHeader(raw, strictConnectionProcessing)
```
usualoma: The variable name “result” seems to be a temporary variable used to hold the function's "result," so I believe it should be renamed.

usualoma (commented):
I've left a comment, so please check it!

src/helper/proxy/index.ts
sugar-cat7: Changed to throw an error.
https://github.com/honojs/hono/blob/ea373483778bd42c34e10462470bb9def4451ffe/src/helper/proxy/index.ts#L70

src/helper/proxy/index.ts
sugar-cat7: Reverted to initial implementation and consolidated logic into buildRequestInitFromRequest since errors are now thrown.
https://github.com/honojs/hono/pull/4459/commits/ea373483778bd42c34e10462470bb9def4451ffe

yusukebe (approved):
LGTM!

yusukebe:
@sugar-cat7 @usualoma 

Thanks! Merging and releasing a new version (it's a minor bump!).

PR #4465, opened 2025-10-17
fix(types): cannot `.use` non-return mw from `createMiddleware`

### LINKED ISSUE #4462: [4.10.0] cannot `.use` middleware created from `createMiddleware` that doesn't return anything

4.10.0

*

```ts
import { Hono } from 'hono'

const sampleMiddleware = createMiddleware(async (c) => {
  console.warn(c.req.url.length)
})

const _app = new Hono()
  .use(sampleMiddleware) // Error here
```

Works normally

### What do you see instead?

Image

### Additional information

This is because the response type of the middleware handler that is created got created as `HandlerResponse<any>` instead of the default `= Response`, not sure why this happens, as the declaration looks good to me honestly, and it works fine for non-async middleware handler and async handler that returns any response.

From my test it can be fixed easily by changing from  
`export type MiddlewareHandler<E extends Env = any, P extends string = string, I extends Input = {}, R extends HandlerResponse<any> = Response>`  
to  
`export type MiddlewareHandler<E extends Env = any, P extends string = string, I extends Input = {}, R extends HandlerResponse<any> = any>`,  
Similar to how `Handler` type is declared.

NamesMT:
Resolves #4462

- [x] Add tests
- [x] Run tests

### Additional context

TS seems to does not raise an error when `R` doesn't actually `extends HandlerResponse<any>`, and implicitly type `R` as `HandlerResponse<any>` directly for cases where `R` is non-extending.

This PR checks for a case where `R` might be `void`, and assign it as non-typed `Response`, i.e.: the default value, so it prevents the created middleware being `HandlerResponse<any>` and broke some types chain like direct `.use`.

### DIFF
--- a/src/helper/factory/index.test.ts
+++ b/src/helper/factory/index.test.ts
@@ -3,7 +3,7 @@ import { expectTypeOf } from 'vitest'
 import { hc } from '../../client'
 import type { ClientRequest } from '../../client/types'
 import { Hono } from '../../index'
-import type { ExtractSchema, ToSchema, TypedResponse } from '../../types'
+import type { Env, ExtractSchema, MiddlewareHandler, ToSchema, TypedResponse } from '../../types'
 import type { ContentfulStatusCode } from '../../utils/http-status'
 import type { Equal, Expect } from '../../utils/types'
 import { validator } from '../../validator'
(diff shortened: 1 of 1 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@NamesMT Thank you!

Issue #4468, opened 2025-10-18
Lambda handler doesn't serve Office files (Excel etc) as binary


matthiasfeist:
### What version of Hono are you using?

4.10.1

### What runtime/platform is your app running on? (with version if possible)

AWS Lambda

### What steps can reproduce the bug?

Serve a response with the Content Type `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet`. The lambda handler will incorrectly detect it as a plain text response (due to the string "xml" that's part of that mime type) and therefore the file will be broken on download at the user.

### What is the expected behavior?

Treat the response as a binary file instead of text

PR #4469, opened 2025-10-18
fix(aws-lambda): serve microsoft office files as binary in lambda handler

matthiasfeist:
This PR fixes https://github.com/honojs/hono/issues/4468

# Description:
When serving Microsoft Office files, due to the existence of "xml" as part of the mimetype, the function `defaultIsContentTypeBinary` incorrectly classifies the response as plain text and not binary. Leading to a broken file download to the user.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/adapter/aws-lambda/handler.test.ts
+++ b/src/adapter/aws-lambda/handler.test.ts
@@ -81,33 +81,25 @@ const baseV2Event: LambdaEvent = {
 }
 
 describe('isContentTypeBinary', () => {
-  it('Should determine whether it is binary', () => {
-    expect(defaultIsContentTypeBinary('image/png')).toBe(true)
-    expect(defaultIsContentTypeBinary('font/woff2')).toBe(true)
-    expect(defaultIsContentTypeBinary('image/svg+xml')).toBe(false)
-    expect(defaultIsContentTypeBinary('image/svg+xml; charset=UTF-8')).toBe(false)
-    expect(defaultIsContentTypeBinary('text/plain')).toBe(false)
-    expect(defaultIsContentTypeBinary('text/plain; charset=UTF-8')).toBe(false)
-    expect(defaultIsContentTypeBinary('text/css')).toBe(false)
-    expect(defaultIsContentTypeBinary('text/javascript')).toBe(false)
-    expect(defaultIsContentTypeBinary('application/json')).toBe(false)
-    expect(defaultIsContentTypeBinary('application/ld+json')).toBe(false)
-    expect(defaultIsContentTypeBinary('application/json')).toBe(false)
-    expect(
-      defaultIsContentTypeBinary(
-        'application/vnd.openxmlformats-officedocument.wordprocessingml.document'
-      )
-    ).toBe(true)
-    expect(
-      defaultIsContentTypeBinary(
-        'application/vnd.openxmlformats-officedocument.presentationml.presentation'
-      )
-    ).toBe(true)
-    expect(
-      defaultIsContentTypeBinary(
-        'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet'
-      )
-    ).toBe(true)
+  it.each([
+    ['image/png', true],
+    ['font/woff2', true],
+    ['image/svg+xml', false],
+    ['image/svg+xml; charset=UTF-8', false],
+    ['text/plain', false],
+    ['text/plain; charset=UTF-8', false],
+    ['text/css', false],
+    ['text/javascript', false],
+    ['application/json', false],
+    ['application/ld+json', false],
+    ['application/json', false],
+    ['application/vnd.openxmlformats-officedocument.wordprocessingml.document', true],
+    ['application/msword', true],
+    ['application/epub+zip', true],
+    ['application/ld+json', false],
+    ['application/vnd.oasis.opendocument.text', true],
+  ])('Should determine whether %s it is binary', (mimeType: string, expected: boolean) => {
+    expect(defaultIsContentTypeBinary(mimeType)).toBe(expected)
   })
 })
 


src/adapter/aws-lambda/handler.ts
```diff
@@ -564,6 +564,7 @@ const isProxyEventV2 = (event: LambdaEvent): event is APIGatewayProxyEventV2 =>
  * @returns True if the content type is binary, false otherwise.
  */
 export const defaultIsContentTypeBinary = (contentType: string): boolean => {
+  if (contentType.startsWith('application/vnd.openxmlformats-officedocument.')) return true
```
web-dev-sayantan: Instead of adding an exception for a very particular format, can we possibly update the original regex on the next line to be something like this:

`^(text\/(plain|html|css|javascript|csv).*|application\/(.*json$|.*json\.|.*xml$|.*xml\.).*|image\/svg\+xml.*)$`

Added couple of additional checks, so that texts like "_application/iam+xml_", "_application/xml.blv.me_" are matched as xmls, but "_application/closedxmlcozwhynot_" is not treated as xml.

usualoma:
Hi @matthiasfeist, @web-dev-sayantan

Thanks for creating the issue ticket and the PR, and for the review. Since the original regular expression can be reorganized, I think the approach in #4470 is the way to go.

src/adapter/aws-lambda/handler.ts
matthiasfeist: Thanks. I just incorporated your suggested changes in my PR

src/adapter/aws-lambda/handler.ts
```diff
@@ -564,7 +564,7 @@ const isProxyEventV2 = (event: LambdaEvent): event is APIGatewayProxyEventV2 =>
  * @returns True if the content type is binary, false otherwise.
  */
 export const defaultIsContentTypeBinary = (contentType: string): boolean => {
-  return !/^(text\/(plain|html|css|javascript|csv).*|application\/(.*json|.*xml).*|image\/svg\+xml.*)$/.test(
+  return !/^(text\/(plain|html|css|javascript|csv).*|application\/(.*json$|.*json\.|.*xml$|.*xml\.).*|image\/svg\+xml.*)$/.test(
```
usualoma: Thank you for considering the changes.

As indicated in #4470, I believe we should make the following changes, including improvements to the current regular expression.

```suggestion
  return !/^text\/(?:plain|html|css|javascript|csv)|(?:\/|\+)(?:json|xml)(?:;|$)/test(
```

matthiasfeist: I'll leave that up @web-dev-sayantan mostly because I'm not that great in Regex syntax to really understand the tradeoffs of that change 😅

usualoma: Hi @matthiasfeist, thanks for your reply.

I understand. In that case, it might be okay for me to make the corrections later.

However, with the current regular expression, the following:

```ts
    expect(defaultIsContentTypeBinary(‘application/json; charset=UTF-8’)).toBe(false)
    expect(defaultIsContentTypeBinary(‘application/xml; charset=UTF-8’)).toBe(false)
```

usualoma: Ah, sorry, there are also cases like the these,

```ts
expect(defaultIsContentTypeBinary(‘application/json ; charset=UTF-8’)).toBe(false)
expect(defaultIsContentTypeBinary(‘application/xml ; charset=UTF-8’)).toBe(false)
expect(defaultIsContentTypeBinary(‘application/json ’)).toBe(false)
expect(defaultIsContentTypeBinary(‘application/xml ’)).toBe(false)
```


The regular expression should have been like this.

```ts
/^text\/(?:plain|html|css|javascript|csv)|(?:\/|\+)(?:json|xml)\s*(?:;|$)/
```

web-dev-sayantan: Thinking out loud, is there any real scenario where the "/" is not present for non-binary content-type?
This regex matches strings like "application+json", which I'm guessing isn't a valid json content-type.

usualoma: Hi @web-dev-sayantan, thank you for your comment!

This code does not verify whether the Content-Type is valid, including the regular expression before modification. Furthermore, since `contentType` is a response header set by the application (not a request header), we can reasonably expect it to be passed in a valid format according to the specification.

Therefore, I believe it is acceptable that `application+json` is not correctly identified as invalid, as it is an invalid format.

src/adapter/aws-lambda/handler.test.ts
```diff
@@ -93,6 +93,21 @@ describe('isContentTypeBinary', () => {
     expect(defaultIsContentTypeBinary('application/json')).toBe(false)
     expect(defaultIsContentTypeBinary('application/ld+json')).toBe(false)
     expect(defaultIsContentTypeBinary('application/json')).toBe(false)
+    expect(
```
yusukebe: I think the purpose of following three is same:

1. `application/vnd.openxmlformats-officedocument.wordprocessingml.document`
2. `application/vnd.openxmlformats-officedocument.presentationml.presentation`
3. `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet`

src/adapter/aws-lambda/handler.ts
matthiasfeist: Looking at https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types and https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types/Common_types there are no types where the slash isn't present. So I think @usualoma's last RegEx looks good and would work.

I'll add it in the PR and also extend the test suite with some more mime-types from the 2 MDM links above

matthiasfeist: OK done :) I added more test-cases and also @usualoma's updated Regex

yusukebe:
@matthiasfeist @web-dev-sayantan @usualoma Thank you all!

matthiasfeist:
Thanks @yusukebe @usualoma @web-dev-sayantan It was fun to contribute! thanks so much for your help!

PR #4470, opened 2025-10-19
fix(aws-lambda): improve regex for content type binary detection

usualoma:
The existing regular expression was somewhat broad in its matching, so I refined it to match only the intended patterns.

- [x] Add tests
- [x] Run tests


usualoma:
@yusukebe
Would you please review this?

usualoma:
Since #4469 has been changed, let's proceed with that one.

PR #4472, opened 2025-10-21
tests: Fix test case of handlers without a path

IAmSSH:
# Description

While going through the tests, I noticed this small issue in `GET http://localhost/anything is not found` test, where the description says we are testing the route `/localhost/anything`, but the test was just checking for `/localhost`.

This PR aims to fix the test case to point to the correct URL.


yusukebe (approved):
LGTM!

yusukebe:
@IAmSSH Thanks!

PR #4478, opened 2025-10-24
fix(request-id): validation accepts `=`

### LINKED ISSUE #4477: Request ID - Can't use aws cloudfront request id

4.10.12

NodeJS behind cloudfront

- Add middleware on your app
```
router.use(
  requestId({
    // AWS CloudFront adds this header to each request it forwards to the origin
    // https://docs.aws.amazon.com/fr_fr/AmazonCloudFront/latest/DeveloperGuide/RequestAndResponseBehaviorCustomOrigin.html
    headerName: "x-amz-cf-id",
  }),
);
```
- Make a request with an header x-amz-cf-id with value `wDPR-dHzm9UNK9vGrCTZe6ybr4u8Z7E3yAVZlrHgThoiTpBFq9F3aw==`

- The provided header (x-amz-cf-id) should be used as request id

### What do you see instead?

- A new request id is generated using generate function

### Additional information

We can see here (L49) https://github.com/honojs/hono/blob/main/src/middleware/request-id/request-id.ts#L49 that there is a regexp test `[^\w\-]` that call generator if request fail.

However, cloudfront request id has a pattern that don't match this regexp like `wDPR-dHzm9UNK9vGrCTZe6ybr4u8Z7E3yAVZlrHgThoiTpBFq9F3aw==`

What are the reasons of this pattern check ?

ryuapp:
closes #4477 

The middleware validation accepts `=` to use a Base64 string like AWS request id.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/request-id/index.test.ts
+++ b/src/middleware/request-id/index.test.ts
@@ -29,6 +29,19 @@ describe('Request ID Middleware', () => {
     expect(await res.text()).toBe('hono-is-hot')
   })
 
+  it('Should return custom request id with all valid characters', async () => {
+    const validRequestId = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-_='
+    const res = await app.request('http://localhost/requestId', {
+      headers: {
+        'X-Request-Id': validRequestId,
+      },
+    })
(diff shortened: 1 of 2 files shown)


yusukebe:
@ryuapp Thank you for the quick fix!

PR #4502, opened 2025-10-31
feat(types): make Hono client's $url return the exact URL type

miyaji255:
### Summary
This pull request enhances the type safety and expressiveness of the client-side URL generation in the Hono framework. The main improvement is the introduction of a new `TypedURL` type, which allows the client to generate URLs with strongly-typed protocol, host, port, path, and query parameters based on the application's route schema and the provided base URL. Several generics and utility types are introduced to propagate this type information throughout the client, ensuring that URL generation is accurate and type-checked at compile time.

esbuild v0.15.18 doesn't support const type parameters, so I updated esbuild.

### Motivation

I wanted to add types so that keys used with SWR would be easy to reuse in other parts of the codebase. However, $url did not preserve that type information, so I created this PR.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/client/types.ts
+++ b/src/client/types.ts
@@ -277,7 +277,6 @@ type PathToChain<
         >
       }
 
- 
 export type Client<T, Prefix extends string> =
   T extends HonoBase<any, infer S, any>
     ? S extends Record<infer K, Schema>


yusukebe:
Hi @miyaji255 

This is super interesting. I'll look at it deeply later.

yusukebe:
Sorry for being late. Anyway, this change is not a "fix", but a "feat". It's a new feature.

src/client/types.test.ts
```diff
@@ -30,15 +31,33 @@ describe('without the leading slash', () => {
     .get('foo', (c) => c.json({}))
     .get('foo/bar', (c) => c.json({}))
     .get('foo/:id/baz', (c) => c.json({}))
-  const client = hc<typeof app>('')
+  const client = hc<typeof app, 'http://localhost'>('http://localhost')
```
yusukebe, line 34: Can we go without the second argument like this?

```ts
const client = hc<typeof routes>('http://localhost/')
```

Or impossible caused by a TypeScript limitation?

miyaji255: TypeScript requires you to specify all other generics if you specify one generic. If you want to solve this problem, you can solve it by currying generics like this.

```ts
const hcStrict = <T extends Hono<any, any, any>>() =>
  <const Prefix extends>(prefix: Prefix, options?: ClientOptions) => hc<T, Prefix>(prefix, options)
```

yusukebe: Thank you for answering. It's a good idea. I wonder if it's the best naming, but `createHc()` might be better (but `hcStrict` also good). Anyway, first, we will not include the function in this PR.

yusukebe:
@miyaji255 

This is awesome! We can include this feature in the next minor release. So, I'll merge when preparing to release it. Thank you!

PR #4503, opened 2025-11-03
docs(CONTRIBUTING): use bun instead of yarn in local development setup

taichi-1:
I found that the "Local Development" section of `CONTRIBUTING.md` still uses `yarn install`, even though the project has moved to Bun as its package manager.

This PR updates it to use `bun install --frozen-lockfile` for consistency with the rest of the project.


yusukebe (approved):
LGTM!

yusukebe:
@taichi-1 

Thank you!

PR #4504, opened 2025-11-05
fix(middleware/cache): skip caching when `Vary: *` is present

pHo9UBenaA:

- [x] Add tests
- [x] Run tests

### Problem

The cache middleware throws an uncaught `TypeError` when attempting to cache responses with a `Vary: *` header.

Current behavior:
1. Handler sets `Vary: *` header
2. Middleware attempts to cache the response
3. `cache.put()` throws: `(error) TypeError: Cannot cache response with 'Vary: *' header.`

#### Details

The middleware's `addHeader()` function explicitly handles `Vary: *`:

https://github.com/honojs/hono/blob/df7ac8d9327cac363b14bfffe1c3f3b039a10897/src/middleware/cache/index.ts#L105-L107

However, after calling `addHeader()`, the middleware unconditionally attempts to cache:

https://github.com/honojs/hono/blob/df7ac8d9327cac363b14bfffe1c3f3b039a10897/src/middleware/cache/index.ts#L131-L137

The Cache API throws a `TypeError` when attempting to cache a response with the `Vary: *` header.
This behavior appears to be in line with Section 4.1 of RFC 9111, which states that a stored response with `Vary: *` will always fail to match.

### Solution

Added a check before `cache.put()` to detect `Vary: *` and skip caching gracefully.

This handles multiple formats:
- `Vary: *`
- `Vary: *, Accept`
- `Vary: Accept, *`
- `Vary: Accept, *, Accept-Encoding`
- `Vary:  *  `

### Reproduction

<details>

```ts
import { Hono } from "hono";
import { cache } from "hono/cache";

const app = new Hono();

app.use("*", cache({ cacheName: "cache-name", wait: true }));

app.get('/vary', (c) => {
  c.header('Vary', '*')
  return c.json({ message: 'This should not be cached' })
})

app.get('/', (c) =>
  c.text('Call GET /vary, then check the server logs.')
)

export default app;
```

</details>

### References

- RFC 9110: [Vary Header Field](https://www.rfc-editor.org/rfc/rfc9110#section-12.5.5)
- RFC 9111: [HTTP Caching](https://www.rfc-editor.org/rfc/rfc9111#section-4.1)

### DIFF
--- a/src/middleware/cache/index.ts
+++ b/src/middleware/cache/index.ts
@@ -12,7 +12,13 @@ import type { StatusCode } from '../../utils/http-status'
  */
 const defaultCacheableStatusCodes: ReadonlyArray<StatusCode> = [200]
 
-const varyWildcardRegExp = /(?:^|,)\s*\*\s*(?:,|$)/
+const shouldSkipCache = (res: Response) => {
+  const vary = res.headers.get('Vary')
+  // Don't cache for Vary: *
+  // https://www.rfc-editor.org/rfc/rfc9111#section-4.1
+  // Also note that some runtimes throw a TypeError for it.
+  return vary && vary.includes('*')
+}
 
 /**
  * Cache Middleware for Hono.
@@ -110,13 +116,6 @@ export const cache = (options: {
         c.header('Vary', vary.join(', '))
       }
     }
-  }
-
-  const shouldSkipCache = (res: Response) => {
-    const vary = res.headers.get('Vary')
-    // Don't cache for Vary: *
-    // https://www.rfc-editor.org/rfc/rfc9111#section-4.1
-    return vary && varyWildcardRegExp.test(vary)
   }
 
   return async function cache(c, next) {


usualoma:
Hi @pHo9UBenaA , thanks for creating the pull request!

I checked the situation. Since TypeErrors occurred with Cloudflare and Deno, it seems that throwing a `TypeError` for `Vary: *` is standard behavior. (Bun didn't throw an error)

I think the implementation approach is sound. However, since the function should be placed outside the middleware as a utility function, and `*` isn't typically included in header names, it's better to check with `.includes(‘*’)` for optimal performance. Also, since the need for this check depends on the Cache API implementation, I think it's better to leave a comment about the `TypeError` as well.

```diff
diff --git i/src/middleware/cache/index.ts w/src/middleware/cache/index.ts
index 467ddcaf..c6ec9484 100644
--- i/src/middleware/cache/index.ts
+++ w/src/middleware/cache/index.ts
@@ -12,7 +12,13 @@ import type { StatusCode } from '../../utils/http-status'
  */
 const defaultCacheableStatusCodes: ReadonlyArray<StatusCode> = [200]
 
-const varyWildcardRegExp = /(?:^|,)\s*\*\s*(?:,|$)/
+const shouldSkipCache = (res: Response) => {
+  const vary = res.headers.get('Vary')
+  // Don't cache for Vary: *
+  // https://www.rfc-editor.org/rfc/rfc9111#section-4.1
+  // Also note that some runtimes throw a TypeError for it.
+  return vary && vary.includes('*')
+}
 
 /**
  * Cache Middleware for Hono.
@@ -112,13 +118,6 @@ export const cache = (options: {
     }
   }
 
-  const shouldSkipCache = (res: Response) => {
-    const vary = res.headers.get('Vary')
-    // Don't cache for Vary: *
-    // https://www.rfc-editor.org/rfc/rfc9111#section-4.1
-    return vary && varyWildcardRegExp.test(vary)
-  }
-
   return async function cache(c, next) {
     let key = c.req.url
     if (options.keyGenerator) {
```

pHo9UBenaA:
Thanks a lot for the detailed review! I’ve committed the suggested changes.

usualoma:
Thank you for the changes 👍  I believe this specification and implementation are acceptable.

yusukebe (approved):
LGTM!

yusukebe:
@pHo9UBenaA @usualoma 

Thanks! I also think the behavior is good. I'll merge and release a new version.

PR #4513, opened 2025-11-11
chore: improve private field removal

BlankParticle:
Use `oxc-parser` to remove private fields from generated code, it uses a single thread and makes the build way faster (20-40ms instead of 35-40sec on single thread), even on github CI

- [x] Add tests
- [x] Run tests

### DIFF
--- a/build/remove-private-fields.ts
+++ b/build/remove-private-fields.ts
@@ -15,7 +15,9 @@ export async function removePrivateFields(files: string[]) {
   await Promise.all(
     parsed.map(async ({ file, sourceCode, ast }) => {
       const sourceCodeWithoutPrivateFields = removePrivateFieldFromSourceCode(ast, sourceCode)
-      if (sourceCodeWithoutPrivateFields) await writeFile(file, sourceCodeWithoutPrivateFields)
+      if (sourceCodeWithoutPrivateFields) {
+        await writeFile(file, sourceCodeWithoutPrivateFields)
+      }
     })
   )
   const end = performance.now()
@@ -34,7 +36,9 @@ export function removePrivateFieldFromSourceCode(ast: ParseResult, sourceCode: s
     },
   }).visit(ast.program)
 
-  if (removals.length === 0) return
+  if (removals.length === 0) {
+    return
+  }
 
   let sourceCodeWithoutPrivateFields = sourceCode
   for (const elem of removals) {


build/remove-private-fields.ts
```diff
@@ -1,40 +1,53 @@
...
-        if (taskId === currentTaskId) {
-          if (type === 'success') {
-            resolve(value)
-          } else {
-            reject(value)
-          }
+  await Promise.all(
+    parsed.map(async ({ file, sourceCode, ast }) => {
+      const sourceCodeWithoutPrivateFields = removePrivateFieldFromSourceCode(ast, sourceCode)
+      if (sourceCodeWithoutPrivateFields) await writeFile(file, sourceCodeWithoutPrivateFields)
```
yusukebe: This is a matter of taste (and we might be able to use a lint rule to unify the style), but we don't write `if (true) return` on a single line. So, can you apply the following patch?

```diff
diff --git a/build/remove-private-fields.ts b/build/remove-private-fields.ts
index df949d40..15d7e821 100644
--- a/build/remove-private-fields.ts
+++ b/build/remove-private-fields.ts
@@ -15,7 +15,9 @@ export async function removePrivateFields(files: string[]) {
   await Promise.all(
     parsed.map(async ({ file, sourceCode, ast }) => {
       const sourceCodeWithoutPrivateFields = removePrivateFieldFromSourceCode(ast, sourceCode)
-      if (sourceCodeWithoutPrivateFields) await writeFile(file, sourceCodeWithoutPrivateFields)
+      if (sourceCodeWithoutPrivateFields) {
+        await writeFile(file, sourceCodeWithoutPrivateFields)
+      }
     })
   )
   const end = performance.now()
@@ -34,7 +36,9 @@ export function removePrivateFieldFromSourceCode(ast: ParseResult, sourceCode: s
     },
   }).visit(ast.program)

-  if (removals.length === 0) return
+  if (removals.length === 0) {
+    return
+  }

   let sourceCodeWithoutPrivateFields = sourceCode
   for (const elem of removals) {
```

BlankParticle: sure, I wasn't aware of that and formatting the code didn't fix it, so I just went with it

yusukebe:
@BlankParticle 

Looks good. It became really fast. Thank you for the contribution!

PR #4518, opened 2025-11-14
feat(types): enhance NotFoundHandler to support custom NotFoundResponse type

miyaji255:
I add a `NotFoundResponse` interface and support custom response types by allowing developers to extend `NotFoundResponse` for their own use cases.

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -96,7 +96,7 @@ export type H<
  *
  * @example
  * declare module 'hono' {
- *   interface NotFoundResponse extends Response & TypedResponse<string, 'text' 404> {}
+ *   interface NotFoundResponse extends Response, TypedResponse<string, 404, 'text'> {}
  * }
  */
 export interface NotFoundResponse {}


yusukebe:
Hi @miyaji255 

Thank you for the PR. This is super interesting. I think I know the usage, but can you share a use case of this `NotFoundHandler`?

miyaji255:
This is how I'm planning to type the `notFound` response for the client, using module augmentation. I think this pattern works well.

```ts
declare module "hono" {
  interface NotFoundResponse extends Response & TypedResponse<{ message: string}, "json", 404>{}
}

export const app = new Hono()
  .get("hoge", c => c.notFound())
  .notFound(c => c.json({ message: "Not Found" }, 404));

// And use the type with hono client
const client = hc("https://example.com/");
await client.hoge.$get() // 404 { message: string }
```

yusukebe:
@miyaji255 

Thank you for the explanation.

Your example should be the following:

```ts
declare module 'hono' {
  interface NotFoundResponse extends Response, TypedResponse<{ message: string }, 404, 'json'> {}
}
```

And, I think you should add the following patch to make it correct behavior:

```diff
diff --git a/src/context.ts b/src/context.ts
index 99bf6db9..aa32ca3d 100644
--- a/src/context.ts
+++ b/src/context.ts
@@ -756,7 +756,7 @@ export class Context<
    * })
    * ```
    */
-  notFound = (): Response | Promise<Response> => {
+  notFound = (): ReturnType<NotFoundHandler> => {
     this.#notFoundHandler ??= () => new Response()
     return this.#notFoundHandler(this)
   }
diff --git a/src/index.ts b/src/index.ts
index ba20118b..84484588 100644
--- a/src/index.ts
+++ b/src/index.ts
@@ -31,6 +31,7 @@ export type {
   Schema,
   ToSchema,
   TypedResponse,
+  NotFoundResponse,
 } from './types'
 /**
  * Types for context, context variable map, context renderer, and execution context.
```

src/types.ts
```diff
@@ -91,7 +91,21 @@ export type H<
   R extends HandlerResponse<any> = any
 > = Handler<E, P, I, R> | MiddlewareHandler<E, P, I, R>
 
-export type NotFoundHandler<E extends Env = any> = (c: Context<E>) => Response | Promise<Response>
+/**
+ * You can extend this interface to define a custom `c.notFound()` Response type.
+ *
+ * @example
+ * declare module 'hono' {
```
yusukebe, line 98: This example does not work. This should be:

```ts
declare module 'hono' {
  interface NotFoundResponse extends Response, TypedResponse<string, 404, 'text'> {}
}
```

yusukebe (approved):
LGTM!

yusukebe:
Hi @miyaji255 

Looks good to me! I'll include this feature in the next minor version. I'll merge later. Thank you!

PR #4519, opened 2025-11-15
feat(timing): add wrapTime to simplify usage

PassiDel:
Hey! I added this `wrapTime` function because I always copy it as "utils" in my own projects. It makes it easier to track the duration of promises. I hope the test cases are good enough.

Best regards!

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/timing/timing.ts
+++ b/src/middleware/timing/timing.ts
@@ -247,13 +247,8 @@ export async function wrapTime<T>(
   precision?: number
 ): Promise<T> {
   startTime(c, name, description)
-  let result
   try {
-    result = await callable
-    return result
-    // eslint-disable-next-line no-useless-catch
-  } catch (e) {
-    throw e
+    return await callable
   } finally {
     endTime(c, name, precision)
   }


src/middleware/timing/timing.ts
```diff
@@ -221,3 +221,40 @@ export const endTime = (c: Context, name: string, precision?: number) => {
...
+  c: Context,
+  name: string,
+  callable: Promise<T>,
+  description?: string,
+  precision?: number
+): Promise<T> {
+  startTime(c, name, description)
+  let result
+  try {
+    result = await callable
```
yusukebe: I think you can write like this for this line:

```ts
return await callable
```

and, the `catch() {}` is really necessary?

If not, you can apply this patch:

```diff
diff --git a/src/middleware/timing/timing.ts b/src/middleware/timing/timing.ts
index 4c81285c..02fa87bb 100644
--- a/src/middleware/timing/timing.ts
+++ b/src/middleware/timing/timing.ts
@@ -247,13 +247,8 @@ export async function wrapTime<T>(
   precision?: number
 ): Promise<T> {
   startTime(c, name, description)
-  let result
   try {
-    result = await callable
-    return result
-    // eslint-disable-next-line no-useless-catch
-  } catch (e) {
-    throw e
+    return await callable
   } finally {
     endTime(c, name, precision)
   }
```

PassiDel: You're totally right!

yusukebe:
Hi @PassiDel 

This is a good feature! Looks good to me. I'll merge this later and include the feature in the next minor version `4.11`. Thank you!

Issue #4520, opened 2025-11-16
Cannot use `Blob` for RPC `form` input


NamesMT:
### What version of Hono are you using?

4.10.6

### What runtime/platform is your app running on? (with version if possible)

*

### What steps can reproduce the bug?

Server
```ts
const app = appFactory.createApp()
  .post(
    '/',
    arktypeValidator('form', type({
      file: `Blob`,
    })),
    async (c) => {
      const { file } = c.req.valid('form')
      console.log({ file })
      return c.text('OK')
    }
  )
```

Client
```ts
await $apiClient.api.modelAdmin.partIcon.$post({ form: { file: img } })
```

### What is the expected behavior?

No error

### What do you see instead?

There is type error at client.

<img width="1421" height="360" alt="Image" src="https://github.com/user-attachments/assets/72d33af4-e575-4fdb-8cfd-d5485eb5939b" />

### Additional information

According to #3119, we should be able to use `Blob` for FormData RPC input, but there is currently type error at the `hc` client, as the input is using `ParsedFormValue` which requires `File` instead of `Blob`

EdamAme-x:
This is just a guess, but although
https://github.com/honojs/hono/blob/4960c063938763a9344c554286e72dcd2545b751/src/client/client.ts#L46
is correctly typed as `ValidationTargets<FormValue>`, the `$<method>` part may not be using the correct type.

EdamAme-x:
<img width="619" height="583" alt="Image" src="https://github.com/user-attachments/assets/6bd10af9-3788-4844-815c-1fa4479dbc77" />

<img width="835" height="238" alt="Image" src="https://github.com/user-attachments/assets/e4a20e88-907b-409a-a2eb-b9c2511f6699" />

okay, i just fixed this problem.

PR #3120, opened 2024-07-10
fix(validator-types): type Alignment with Web Standards

EdamAme-x:
Close https://github.com/honojs/hono/issues/3119

### The author should do the following, if applicable

- [x] Add tests
- [x] Run tests
- [x] `bun run format:fix && bun run lint:fix` to format the code
- [ ] Add [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code

PR #4521, opened 2025-11-16
fix(validator): fix incomplete types and wrong tests

EdamAme-x:
close: https://github.com/honojs/hono/issues/4520

This PR fixes the incomplete work from the PR I created a year ago. (https://github.com/honojs/hono/issues/3120)
I forgot to update the types for the input schemas such as `$get` and `$post`, and this mistake was caused by incorrect tests.
This PR should resolve everything. Sorry about that.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/validator/validator.test.ts
+++ b/src/validator/validator.test.ts
@@ -9,8 +9,8 @@ import type {
   ErrorHandler,
   ExtractSchema,
   ExtractSchemaForStatusCode,
+  FormValue,
   MiddlewareHandler,
-  ParsedFormValue,
   TypedResponse,
   ValidationTargets,
 } from '../types'
@@ -954,7 +954,7 @@ it('With path parameters', () => {
       $put: {
         input: {
           form: {
-            title: ParsedFormValue | ParsedFormValue[]
+            title: FormValue | FormValue[]
           }
         } & {
           param: {
@@ -1000,7 +1000,7 @@ it('`on`', () => {
       $purge: {
         input: {
           form: {
-            tag: ParsedFormValue | ParsedFormValue[]
+            tag: FormValue | FormValue[]
           }
         } & {
           query: {
--- a/src/validator/validator.ts
+++ b/src/validator/validator.ts
@@ -1,7 +1,7 @@
 import type { Context } from '../context'
 import { getCookie } from '../helper/cookie'
 import { HTTPException } from '../http-exception'
-import type { Env, MiddlewareHandler, TypedResponse, ValidationTargets } from '../types'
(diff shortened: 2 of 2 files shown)


EdamAme-x:
Hi @NamesMT.
Can you review this?

NamesMT:
Hi @EdamAme-x 
Change in this PR looks good to me, but I don't quite get why did #3120 added `ParsedFormValue` and `T` generic to `ValidationTargets` yet.
What will happen if we revert #3120 changes and simply just update `FormValue` to `string | Blob` and update the related tests if needed?

EdamAme-x:
Thanks for your reply @NamesMT.
The reason I made this change is that on the client side, `FormData` allows `string | Blob` (including `File`, since it’s a subclass), but when you retrieve the value, it becomes only `string | File` (`FormDataEntryValue`). The same happens on the server side.
https://github.com/honojs/hono/blob/4960c063938763a9344c554286e72dcd2545b751/src/validator/validator.ts#L135

The important point here is that on the server side, the value is guaranteed to be a `File`, but if the type were `string | Blob`, we would need extra checks like `if (file instanceof File)` to get the file name or to confirm the type.

My PR aims to eliminate this overhead and the inconsistency with the Web standard.

source:  
https://x.gd/ENg1i

NamesMT:
Thanks for the knowledge @EdamAme-x .

yusukebe:
@EdamAme-x 

Nice! Thank you for the contribution.

NamesMT:
@yusukebe @EdamAme-x 

Hi, I just updated to `v4.10.7`, it seems like the problem is still there, can you please help check.

Damn, last time I didn't do a runtime check, maybe we should add tests for actual types of runtime `hc` client?

image

yusukebe:
@NamesMT 

I've glanced. The following code does not throw type errors.

```ts
// Server side - accepts Blob in form data
const app = new Hono().post(
  '/upload',
  validator('form', () => {
    return { file: new Blob(['hello'], { type: 'text/plain' }) }
  }),
  async (c) => {
    const { file } = c.req.valid('form')
    return c.text('OK')
  }
)

type AppType = typeof app

const client = hc<AppType>('http://localhost:8787')
const blob = new Blob(['hello'], { type: 'text/plain' })

const res = await client.upload.$post({
  form: {
    file: blob, // No type error here - Blob is accepted
  },
})
```

So, I can't reproduce it (or I have made a mistake??).

NamesMT:
That is weird, I will update later if I got any more news/reproduce, or the issue is something on my end, thanks a lot.

PR #4529, opened 2025-11-21
fix(types): correct types when middleware return Promise<void>

### LINKED ISSUE #4517: Downstream methods without `path` args break upstream method path typing

4.10.5

Bun 1.2.13

a typing edge case was [reported on discord](https://discord.com/channels/1011308539819597844/1436389937992958063)
  - below i've included an illustrative test case adapted from the existing [Env types and a path type with `app.use(path, handler...)\` - test only types](https://github.com/honojs/hono/blob/971106d132ec8a989be12ec5c8e63cfaf597cd4f/src/types.test.ts#L1905) in `/src/tests.ts`

### TL;DR

- when a handler or middleware is called with a `path` argument, the `Hono` instance [`this.#path` is updated](https://github.com/honojs/hono/blob/971106d132ec8a989be12ec5c8e63cfaf597cd4f/src/hono-base.ts#L139)
  - if the next method in the chain is also called with the `path` argument, there's no issue
  - if the next method is called _without_ the `path` argument, then:
    - the call defaults the path to the previously-set `this.#path` (expected)
    - the schema uses the previously-set path type (expected)
    - _any methods called before the middleware are also **typed** using the not-yet-set path (unexpected)_
- i don't totally understand why this happens, but i think has to do with the [type merging](https://github.com/honojs/hono/blob/971106d132ec8a989be12ec5c8e63cfaf597cd4f/src/types.ts#L2139) that `ChangePathOfSchema` does
  - in order to resolve downstream schemas, it flattens them to the `Path` key, but i think this also overwrites the correct (type) paths for  _upstream_ schemas
  - linking these PRs as they seem like relevant context:
    - https://github.com/honojs/hono/pull/3087
    - https://github.com/honojs/hono/pull/3128

### Test Case

- the expected endpoints are `GET /` and `POST /:id`
- the actual endpoints are `GET /` and `POST /:id`
- the endpoint _types_ are `GET /:id` and `POST /:id`

```typescript
type Env = {
  Variables: {
    foo: string
  }
}

const app = new Hono<Env>()
  .get((c) => c.text('before'))
  .use('/:id', async (_c, next) => {
    await next()
  })
  .post((c) => c.text('after'))

type Actual = ExtractSchema<typeof app>
/**
  {
    '/:id': {
      $get: {
        input: {};
        output: 'before';
        outputFormat: 'text';
        status: ContentfulStatusCode;
      }
    }
  } & {
    '/:id': {
      $post: {
        input: {
          param: {
            id: string;
          };
        };
        output: 'after';
        outputFormat: 'text';
        status: ContentfulStatusCode;
      }
    }
  }
*/
type Expected = {
  '/': {
    $get: {
      input: {};
      output: 'before';
      outputFormat: 'text';
      status: ContentfulStatusCode;
    }
  }
} & {
  '/:id': {
    $post: {
      input: {
        param: {
          id: string;
        };
      };
      output: 'after';
      outputFormat: 'text';
      status: ContentfulStatusCode;
    }
  }
}

// error
```

_No response_

### What do you see instead?

_No response_

### Additional information

ideally the types would be in sync with the endpoint behavior. that being said, the current solution covers a majority of cases, and i'm not sure there's a performant solution with better coverage. there are a few simple rules/workarounds
- avoid calling methods/`use` without the `path` argument
- front-load middleware at the top of the chain
- break routes up RESTfully

### Example Workaround
```typescript
const idRoute = new Hono<Env>()
  .use('/:id', async (_c, next) => {
    await next()
  })
  .post('/:id', (c) => c.text('after'))

const app = new Hono<Env>()
  .get('/', (c) => c.text('before'))
  .route('/', idRoute)
```

### Next Steps

- i'll update the docs to describe the path defaulting behavior when methods are called without a `path` argument
  - i'll include a warning about this edge-case
- i'll work on a solution when i have a bit more time, but i'm wondering whether it's worth trying to resolve at all right now

yusukebe:
Fixes #4517

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -2288,12 +2288,10 @@ export type TypedResponse<
   _format: F
 }
 
-type MergeTypedResponse<T> = IsAny<T> extends true
-  ? any
+type MergeTypedResponse<T> = T extends Promise<void>
+  ? T
   : T extends Promise<infer T2>
-  ? T2 extends void
-    ? Promise<void>
-    : T2 extends TypedResponse
+  ? T2 extends TypedResponse
     ? T2
     : TypedResponse
   : T extends TypedResponse


yusukebe:
The issue was resolved by https://github.com/honojs/hono/pull/4552. Closing.

PR #4531, opened 2025-11-21
feat(pretty-json): support force option

missinglink:
This PR adds a `force` option to `pretty-json` for users who want all responses to be pretty printed, ie. don't want to add `?pretty=1` to all queries and don't want to use a 3rd party or hand-rolled middleware.

usage:

```js
app.use('*', prettyJSON({ force: true }))
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/pretty-json/index.test.ts
+++ b/src/middleware/pretty-json/index.test.ts
@@ -55,4 +55,22 @@ describe('JSON pretty by Middleware', () => {
     const nonPrettyText = await (await app.request('?pretty')).text()
     expect(nonPrettyText).toBe('{"message":"Hono!"}')
   })
+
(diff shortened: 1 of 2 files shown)


missinglink:
IMHO the API for this middleware would be more intuitive if the *default mode* was to prettify *all* responses.

Then if the `query` option was specified it would adopt the existing behaviour of only running when the query param is present.

That ship has unfortunately sailed as it would break backwards compat but an interesting idea if there were ever a v2.

yusukebe:
Hi @missinglink 

This is a great feature! As you said, we can consider whether `force` should be the default. But currently, let's go with this.

We may include this change for the next minor bump. I'll merge this after preparing it. Thanks!

Issue #4429, opened 2025-09-26
Bun adapter: TypeError: null is not an object (evaluating 'q.address')


brenc:
### What version of Hono are you using?

4.9.8

### What runtime/platform is your app running on? (with version if possible)

Bun 1.2.22, Docker (debian:bookworm-slim)

### What steps can reproduce the bug?

Recently I deployed a Bun + Hono web app that sits behind proxy servers. Once in a while a connection is closed prematurely ending in `server.requestIP()` being set to `null` (which is how it's typed `(method) Server.requestIP(request: Request): Bun.SocketAddress | null
`). This leads to an exception within Hono's Bun adapter:
```
 9 |     throw new TypeError("server.requestIP is not a function.");
10 |   }
11 |   const info = server.requestIP(c.req.raw);
12 |   return {
{
info.address,
                    ^
null is not an object (evaluating 'q.address')
      at NH (node_modules/hono/dist/adapter/bun/conninfo.js:14:16)
      at MO (src/middleware/ip-detection.ts:30:20)
      at <anonymous> (src/middleware/ip-detection.ts:59:16)
      at <anonymous> (src/middleware/ip-detection.ts:76:16)
      at G (node_modules/hono/dist/compose.js:22:23)
      at G (node_modules/hono/dist/compose.js:40:14)
      at <anonymous> (node_modules/hono/dist/compose.js:5:12)
      at <anonymous> (node_modules/hono/dist/hono-base.js:201:31)
      at <anonymous> (node_modules/hono/dist/hono-base.js:209:40)
      at <anonymous> (src/guard.ts:291:24)
```

### What is the expected behavior?

I suspect Hono should detect when `server.requestIP()` is set to null and return that (return `ConnInfo | null`).

yusukebe:
Hi @brenc 

If you want to use the function, please use [the ConnInfo Helper](https://hono.dev/docs/helpers/conninfo):

```ts
import { Hono } from 'hono'
import { getConnInfo } from 'hono/bun'

const app = new Hono()

app.get('/', (c) => {
  const info = getConnInfo(c)
  return c.json(info)
})

export default app
```

brenc:
As you can see from the trace, I already am using the ConnInfo helper. This is where the error is coming from.

`(node_modules/hono/dist/adapter/bun/conninfo.js:14:16)`

yusukebe:
@brenc 

Can you share a minimal project to reproduce it? I can't reproduce it.

brenc:
[...]

PR #4538, opened 2025-11-24
fix(adapter/bun): fix TypeError: null is not an object (#4429)

brenc:
Hello - I was finally able to dig into #4429. The problem was that `server.requestIP(c.req.raw)` can sometimes return null. I believe this happens when the connection is closed prematurely, before the full request is completed. I believe I have created a reasonable fix for this. I made a private NPM module of my clone to test this fix, and I haven't seen any issues with it overnight.

- [x] Add tests
- [x] Run tests


brenc:
Do you need me to do anything with this? The 4 lines it mentioned are comments and newlines.

yusukebe:
Hi @brenc 

Looks good! Thank you for the contribution.

PR #4543, opened 2025-11-25
feat(context): export the Context class publicly

BlankParticle:

- [ ] Add tests
- [x] Run tests

### Motivation
While writing a rpc method on a Durable Obeject, I wanted to return a SSE stream, so I reached for the `streamSSE` api from `hono/streaming`, but it turned out I needed to provide in a context, constructing a Context from request is easy, just the issue is that `Context` is not exported as a accessible class

I managed to work around it with this immediately invoked hono router setup
```ts
import { DurableObject } from "cloudflare:workers";
import { Hono } from "hono";

export class SSEStreamer extends DurableObject {
  async stream() {
    return new Hono()
      .get("*", (c) => {
        return streamSSE(c, async (api) => {
          await api.writeSSE({
            event: "message",
            data: "Begin!",
          });
            await api.writeSSE({
              event: "message",
              data: `tick ${i}`,
            });
            await new Promise((resolve) => setTimeout(resolve, 500));
          }
          await api.writeSSE({
            event: "message",
            data: "End!",
          });
          await api.close();
        });
      })
      .fetch(new Request("http://internal"));
  }
}

export default {
  async fetch(request: Request, env: Env, ctx: ExecutionContext) {
    const stub = env.SSE.getByName("test");
    const res = await stub.stream();
    return res;
  },
} satisfies ExportedHandler<Env>;
```

But this would be lot easier with the class, just creating the context as needed
```ts
const c = new Context(request, {/* */})
```

this opens up the bridge to use hono helpers like streaming, proxy, cookie utils with normal web standard requests without having to construct a full hono router
```ts
// example of using cookie utils in a fetch method
 fetch(request: Request): Response | Promise<Response> {
   const c = new Context(request);
   setCookie(c, "test", "test-cookie");
   return c.json({ message: "Hello, world!" });
 }
 ```


BlankParticle:
hey @yusukebe, 
I made this PR a while ago, I was wondering if you could have a look at this PR, this would be really nice to have

yusukebe (approved):
LGTM!

yusukebe:
Hey @BlankParticle 

Sorry to be late! Regarding this, I was wondering whether we should export the `Context` class. Because I think in many cases exporting only the type is enough. I worried that it was used incorrectly and something unintended would happen. But let's go with this. I'll merge and release later. Thanks.

BlankParticle:
yeah every public api has potential to be abused, but I think exporting context class is fine and useful. This would be useful in testing and building libs on hono too

BlankParticle:
@yusukebe I saw all the approved PRs got merged and new version was released, I am letting you know in case you missed this PR 🫡

yusukebe:
Oops! Please wait a little

BlankParticle:
> Oops! Please wait a little

no worries! I know you are dealing with lots of spam

yusukebe:
I'll merge now. This will be included in the next patch version.

PR #4545, opened 2025-11-26
docs(jwk): clarify what the JWK middleware validates

### LINKED ISSUE #4542: Documentation on what JWK middleware validates
On https://hono.dev/docs/middleware/builtin/jwk, it states:

> Specifically, it validates tokens using the provided `keys`, retrieves keys from `jwks_uri` if specified, and supports token extraction from cookies if the `cookie` option is set.

But what exactly is being validated?

I assume it checks the integrity at the very least, using the `alg` and `kid`  in the header.

Does it also check the `exp`/`nbf` fields as part of its validation?

I'm pretty sure it doesn't check `aud` as that's not one of the options. Neither is `iss`.

There's all kinds of validation for access tokens (as an example, see https://learn.microsoft.com/en-us/entra/identity-platform/access-tokens#validate-tokens). It would be really nice to know what the middleware does, so that what it doesn't do can be implemented in a subsequent middleware.

surajkewat72:
Updated the documentation to explain what validations happen automatically vs what developers need to handle themselves. Added some examples showing different use cases including custom claim validation.

Fixes #4542

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/middleware/jwk/jwk.ts
+++ b/src/middleware/jwk/jwk.ts
@@ -16,6 +16,34 @@ import type { VerifyOptions } from '../../utils/jwt/jwt'
 /**
  * JWK Auth Middleware for Hono.
  *
+ * This middleware handles JWT validation using JSON Web Keys (JWKs). Here's what you need to know
+ * about what gets validated automatically and what you'll need to handle in your application code.
+ *
+ * ## What the middleware validates automatically:
+ *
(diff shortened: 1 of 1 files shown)


yusukebe:
@surajkewat72 

Thank you for the PR. I think we have to prioritize updating the website: https://hono.dev/docs/middleware/builtin/jwk , instead of adding a lot of description to JSDoc.

PR #4552, opened 2025-11-26
fix(types): replace schema-based path tracking with CurrentPath parameter

kosei28:
- ref: #4529

- [x] Add tests
- [x] Run tests

### DIFF
--- a/build/build.ts
+++ b/build/build.ts
@@ -9,7 +9,7 @@
 
 import arg from 'arg'
 import { $ } from 'bun'
-import { build } from 'esbuild'
+import { build, context } from 'esbuild'
 import type { Plugin, PluginBuild, BuildOptions } from 'esbuild'
 import * as glob from 'glob'
 import fs from 'fs'
@@ -67,35 +67,43 @@ const addExtension = (extension: string = '.js', fileExtension: string = '.ts'):
 })
 
 const commonOptions: BuildOptions = {
-  watch: isWatch,
   entryPoints,
   logLevel: 'info',
   platform: 'node',
 }
 
-const cjsBuild = () =>
-  build({
-    ...commonOptions,
-    outbase: './src',
-    outdir: './dist/cjs',
-    format: 'cjs',
-  })
+const cjsConfig: BuildOptions = {
+  ...commonOptions,
+  outbase: './src',
+  outdir: './dist/cjs',
+  format: 'cjs',
+}
 
-const esmBuild = () =>
-  build({
-    ...commonOptions,
-    bundle: true,
-    outbase: './src',
-    outdir: './dist',
-    format: 'esm',
-    plugins: [addExtension('.js')],
-  })
+const esmConfig: BuildOptions = {
+  ...commonOptions,
+  bundle: true,
+  outbase: './src',
+  outdir: './dist',
+  format: 'esm',
+  plugins: [addExtension('.js')],
+}
 
-Promise.all([esmBuild(), cjsBuild()])
+const runBuild = async (config: BuildOptions) => {
+  if (isWatch) {
+    const ctx = await context(config)
+    await ctx.watch()
+  } else {
+    await build(config)
+  }
+}
 
-await $`tsc ${
-  isWatch ? '-w' : ''
-} --emitDeclarationOnly --declaration --project tsconfig.build.json`.nothrow()
+await Promise.all([
+  runBuild(esmConfig),
+  runBuild(cjsConfig),
+  $`tsc ${
+    isWatch ? '-w' : ''
+  } --emitDeclarationOnly --declaration --project tsconfig.build.json`.nothrow(),
+])
 
 // Remove #private fields
 const dtsEntries = glob.globSync('./dist/types/**/*.d.ts')
--- a/bun.lock
+++ b/bun.lock
@@ -1,6 +1,5 @@
 {
   "lockfileVersion": 1,
-  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "hono",
@@ -15,7 +14,7 @@
         "arg": "^5.0.2",
         "bun-types": "^1.2.20",
         "editorconfig-checker": "6.1.1",
-        "esbuild": "^0.15.18",
+        "esbuild": "^0.27.1",
         "eslint": "9.39.1",
         "glob": "^11.0.0",
         "jsdom": "22.1.0",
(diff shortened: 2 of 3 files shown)


kosei28:
I found a part that works differently than expected outside of the test cases, so I'll put the PR in draft.

kosei28:
I fixed it according to the original specifications and added tests for them.

kosei28:
The number of instantiations is now 694,649.

kosei28:
I fixed the pathless handler's path to be a union to match the original type. However, in the actual routing, `this.#path` (which stores the last registered path) is used for the pathless handler, resulting in the following behavior. The `CurrentPath` implementation before the fix might have been better.

Also, even if the handler's return value isn't `Promise<void>`, calling `next` will route to the subsequent handler, so I don't think type branching based on the return type is necessary.

```ts
import { Hono } from "hono";

const app = new Hono()
  .use(async (c, next) => {
    await next();
    console.log(`status: ${c.res.status}`);
  })
  .get("/a", async (c, next) => {
    console.log("get /a middleware");
    await next();
  })
  .get("/b", async (c, next) => {
    console.log("get /b middleware");
    await next();
    return c.res;
  })
  .get((c) => {
    console.log(`get ${c.req.url}`);
    return c.text("");
  })
  .post(async (c) => {
    console.log(`post ${c.req.url}`);
    return c.text("");
  });

await app.request("/a", { method: "GET" });
// get /a middleware
// status: 404

await app.request("/b", { method: "GET" });
// get /b middleware
// get http://localhost/b
// status: 200

await app.request("/a", { method: "POST" });
// status: 404

await app.request("/b", { method: "POST" });
// post http://localhost/b
// status: 200
```

kosei28:
For reference, the schema generated by the code above looks like this:

`CurrentPath` approach
```ts
const app: Hono<{}, {
    "/b": {
        $get: {
            input: {};
            output: {};
            outputFormat: string;
            status: StatusCode;
        };
    };
} & {
    "/b": {
        $get: {
            input: {};
            output: "";
            outputFormat: "text";
            status: ContentfulStatusCode;
        };
    };
} & {
    "/b": {
        $post: {
            input: {};
            output: "";
            outputFormat: "text";
            status: ContentfulStatusCode;
        };
    };
}, "/", "/b">
```

`VoidMiddlewarePath` approach
```ts
const app: Hono<{}, {
    "/b": {
        $get: {
            input: {};
            output: {};
            outputFormat: string;
            status: StatusCode;
        };
    };
} & {
    "/a": {
        $get: {
            input: {};
            output: "";
            outputFormat: "text";
            status: ContentfulStatusCode;
        };
    };
} & {
    "/a": {
        $post: {
            input: {};
            output: "";
            outputFormat: "text";
            status: ContentfulStatusCode;
        };
    };
}, "/", "/a">
```

`ExtractStringKey` approach (from #4529)
```ts
const app: Hono<{}, {
    "/a": {};
} & {
    "/b": {
        $get: {
            input: {};
            output: {};
            outputFormat: string;
            status: StatusCode;
        };
    };
} & {
    "/a": {
        $get: {
            input: {};
            output: "";
            outputFormat: "text";
            status: ContentfulStatusCode;
        };
    };
} & {
    "/a": {
        $post: {
            input: {};
            output: "";
            outputFormat: "text";
            status: ContentfulStatusCode;
        };
    };
    "/b": {
        $post: {
            input: {};
            output: "";
            outputFormat: "text";
            status: ContentfulStatusCode;
        };
    };
}, "/">
```

kosei28:
The number of instantiations for the `CurrentPath` approach is 684,676.

yusukebe:
Hi @kosei28 

Thank you for your work! I want to take a look at this deeply, but I'm a little busy. Though I said "please help me", I can't work on it immediately, sorry, but wait! It will help us. You are great!

src/types.ts
```diff
@@ -2069,25 +2262,50 @@ export interface OnHandlerInterface<
...
+  ? [keyof S & string] extends [never]
+    ? BasePath
+    : keyof S & string
+  : VoidMiddlewarePath
+
+type AddVoidMiddlewarePath<Merged, VoidMiddlewarePath extends string, Path extends string> = [
+  Merged,
+] extends [Promise<void>]
+  ? VoidMiddlewarePath | Path
+  : [Merged] extends [never]
```
yusukebe: Are these lines necessary? This means you can apply the following patch:

```diff
diff --git a/src/types.ts b/src/types.ts
index 27aeefaa..5766b474 100644
--- a/src/types.ts
+++ b/src/types.ts
@@ -2303,9 +2303,7 @@ type AddVoidMiddlewarePath<Merged, VoidMiddlewarePath extends string, Path exten
   Merged,
 ] extends [Promise<void>]
   ? VoidMiddlewarePath | Path
-  : [Merged] extends [never]
-    ? VoidMiddlewarePath | Path
-    : VoidMiddlewarePath
+  : VoidMiddlewarePath
```

I think `Merged` will not be `never` in any case.

src/helper/factory/index.test.ts
```diff
@@ -388,9 +388,7 @@ describe('createFactory', () => {
 
       const routes = new Hono().use('*', middleware)
       type Actual = ExtractSchema<typeof routes>
-      type Expected = {
-        '*': {}
-      }
+      type Expected = {}
```
yusukebe, line 391: Though you change the test, I also think it's okay. It's not a problem for real-world use cases. Makes sense for me 👍

yusukebe:
Hey @kosei28 !

Sorry for being late. Awesome! Let's go with the `VoidMiddlewarePath` approach.
I've left some comments. Please check them.

kosei28:
I think the `CurrentPath` approach is simpler to implement and aligns better with the actual behavior. Could you tell me the reason for adopting the `VoidMiddlewarePath` approach?

`CurrentPath` approach: https://github.com/honojs/hono/compare/51765d1...98df8fa

src/types.ts
kosei28: I was just following `AddSchemaOrPath` from #4529, but yeah, `Merged` won't be `never`.

yusukebe: Ah, my bad!

yusukebe:
@kosei28 

Ah, you are right! The `CurrentPath` approach is best. The problem was that types did not know the "current path". Let's go with it.

kosei28:
Reverted to `CurrentPath` approach and fixed type updates for pathless `app.use()` and multi-path `app.on()`.

Note: Tests pass locally, but CI is failing due to const type parameters conflicting with the old esbuild version. #4571 will fix the build.

yusukebe:
@kosei28 

Looks good! You solve the big problem. Awesome. Thank you for your contribution!!

yusukebe:
I'll merge this to the `next` branch for the next minor version and release it soon.

PR #4553, opened 2025-11-27
fix(validator): no return in middleware

hyoban:

- [ ] Add tests
- [x] Run tests

We may not have returned a response yet. In RSSHub, we have a wrapped handler, https://github.com/DIYgod/RSSHub/blob/680abcc00820876176d34cd47c3c13a37ad564dd/lib/registry.ts#L193-L210, and I tried to add a validator. https://github.com/DIYgod/RSSHub/pull/20584 This will result in unexpected errors. Please tell me if I did anything wrong.

~~BTW, can we have the standard validator accept a nullable schema? In that case, validation can be skipped.~~ I can pass an empty middleware.

And is it possible to restrict the schema type to the parameter defined in the path (Is there a type utils for this in Hono?) This is how I wrote it now.

```
type ParamObjectFromPath<P extends string> = P extends `${string}:${infer Param}/${infer Rest}`
    ? Param extends `${infer Name}?`
        ? { [K in Name]?: string } & ParamObjectFromPath<`/${Rest}`>
        : { [K in Param]: string } & ParamObjectFromPath<`/${Rest}`>
    : P extends `${string}:${infer Param}`
      ? Param extends `${infer Name}?`
          ? { [K in Name]?: string }
          : { [K in Param]: string }
      : Record<string, never>;
```

### DIFF
--- a/src/validator/validator.ts
+++ b/src/validator/validator.ts
@@ -166,7 +166,7 @@ export const validator = <
 
     c.req.addValidatedData(target, res as never)
 
-    return (await next()) as ExtractValidationResponse<VF>
+    await next()
   }
 }
 


yusukebe:
Hey @hyoban 

Can you provide a **minimal** code or project to explain your problem?

hyoban:
> Can you provide a **minimal** code or project to explain your problem?

Like this, I want to handle the return only in the middleware.

```ts
import { Hono } from "hono";
import { validator } from "hono/validator";

const app = new Hono();

app.use(async (c, next) => {
  await next();
  return c.json({ message: c.get('message') });
});

app.get(
  "/",
  validator("query", () => {
    return 'whatever';
  }),
  (c) => {
    c.set('message', 'Hello, Hono!');
  }
);

serve(app)
```

yusukebe:
Hi @hyoban 

Thank you for the response.

I understand your problem. Actually, I/we designed the handler after the validator returns a response. So, if you don't return anything, it doesn't behave correctly, and a TypeScript type mismatch occurs.

This means I've not imagined the use case using the no response handler after validation. But we may have to accept it. I'll work on it. Thanks!

PR #4558, opened 2025-12-03
feat(csrf): Support async `IsAllowedOriginHandler`

baseballyama:
I updated `IsAllowedOriginHandler` to support `async`. In the project I’m involved in, each tenant has its own subdomain, so CORS checks may require database access. Since DB access involves asynchronous I/O, it’s useful if `IsAllowedOriginHandler` can call async functions internally.

Using `await` may incur a small performance cost because it consumes one tick, but I believe this overhead is negligible. What do you think?

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/csrf/index.ts
+++ b/src/middleware/csrf/index.ts
@@ -100,7 +100,7 @@ export const csrf = (options?: CSRFOptions): MiddlewareHandler => {
       return (origin) => optsOrigin.includes(origin)
     }
   })(options?.origin)
-  const isAllowedOrigin = async (origin: string | undefined, c: Context): Promise<boolean> => {
+  const isAllowedOrigin = async (origin: string | undefined, c: Context) => {
     if (origin === undefined) {
       // denied always when origin header is not present
       return false


src/middleware/csrf/index.ts
```diff
@@ -100,12 +100,12 @@ export const csrf = (options?: CSRFOptions): MiddlewareHandler => {
       return (origin) => optsOrigin.includes(origin)
     }
   })(options?.origin)
-  const isAllowedOrigin = (origin: string | undefined, c: Context) => {
+  const isAllowedOrigin = async (origin: string | undefined, c: Context): Promise<boolean> => {
```
yusukebe: To make a short code, you can remove the return value annotation and the `await`. Do you think of it?

```diff
diff --git a/src/middleware/csrf/index.ts b/src/middleware/csrf/index.ts
index f57b99e4..24164919 100644
--- a/src/middleware/csrf/index.ts
+++ b/src/middleware/csrf/index.ts
@@ -100,12 +100,12 @@ export const csrf = (options?: CSRFOptions): MiddlewareHandler => {
       return (origin) => optsOrigin.includes(origin)
     }
   })(options?.origin)
-  const isAllowedOrigin = async (origin: string | undefined, c: Context): Promise<boolean> => {
+  const isAllowedOrigin = async (origin: string | undefined, c: Context) => {
     if (origin === undefined) {
       // denied always when origin header is not present
       return false
     }
-    return await originHandler(origin, c)
+    return originHandler(origin, c)
   }

   const secFetchSiteHandler: IsAllowedSecFetchSiteHandler = ((optsSecFetchSite) => {
```

baseballyama: Thank you for reviewing the PR!
In conclusion, I am happy to follow your opinion, so please let me know your final decision. First, let me explain why I added `await`.

The primary reason is that without `await`, the stack trace gets cut off, which makes debugging difficult for users. A complete stack trace helps users investigate issues easily.

However, it is true that there is a performance overhead (in some runtimes). Since `async` / `await` keywords consume one microtask each, they add overhead. I ran a quick benchmark:

Node.js / Deno: About 10% overhead.
Bun: Interestingly, it was about 20% faster.

That said, I don't think the performance overhead is critical in this case. Since the CORS middleware typically runs only once per API request, I don't believe it will significantly impact the overall response performance.

## Stack trace differences REPL

```js
node -e '
const originHandler = async (origin, c) => {
  await Promise.resolve();
  throw new Error("Error inside originHandler"); 
};

const isAllowedOrigin_WithAwait = async (origin, c) => {
  if (origin === undefined) return false;
  return await originHandler(origin, c);
};

const isAllowedOrigin_NoAwait = async (origin, c) => {
  if (origin === undefined) return false;
  return originHandler(origin, c);
};

const execute = async () => {  
  const c = {};
  const origin = "https://example.com";

  console.log("\n--- [Case 1] return await originHandler(...) ---");
  try {
    await isAllowedOrigin_WithAwait(origin, c);
  } catch (e) {
    // Should see "at isAllowedOrigin_WithAwait"
    console.log(e.stack.split("\n").slice(1, 4).join("\n"));
  }

  console.log("\n--- [Case 2] return originHandler(...) ---");
  try {
    await isAllowedOrigin_NoAwait(origin, c);
  } catch (e) {
    // "at isAllowedOrigin_NoAwait" might be missing
    console.log(e.stack.split("\n").slice(1, 4).join("\n"));
  }
}

execute();
'

// --- [Case 1] return await originHandler(...) ---
//     at originHandler ([eval]:5:9)
//     at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
//     at async isAllowedOrigin_WithAwait ([eval]:11:10)

// --- [Case 2] return originHandler(...) ---
//     at originHandler ([eval]:5:9)
//     at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
```

## Performance differences REPL

```js
bun -e '
const { performance } = require("perf_hooks");

const originHandler = async () => 1;

const isAllowedOrigin_WithAwait = async () => { return await originHandler(); };
const isAllowedOrigin_NoAwait = async () => { return originHandler(); };

(async () => {
  const ITERATIONS = 5_000_000;
  console.log(`\nStarting microbenchmark (${ITERATIONS.toLocaleString()} iterations)...`);

  // --- Warmup (Important for JIT optimization) ---
  for (let i = 0; i < 10000; i++) {
    await isAllowedOrigin_WithAwait();
    await isAllowedOrigin_NoAwait();
  }

  // --- Measure: With Await ---
  const start1 = performance.now();
  for (let i = 0; i < ITERATIONS; i++) {
    await isAllowedOrigin_WithAwait();
  }
  const time1 = performance.now() - start1;
  console.log(`[With return await] ${time1.toFixed(2)} ms`);

  // --- Measure: No Await ---
  const start2 = performance.now();
  for (let i = 0; i < ITERATIONS; i++) {
    await isAllowedOrigin_NoAwait();
  }
  const time2 = performance.now() - start2;
  console.log(`[No return await]   ${time2.toFixed(2)} ms`);

  // --- Result ---
  const diff = time1 - time2;
  const ratio = (time2 / time1) * 100;
  console.log(`\nResult: Removing await was ${Math.abs(diff).toFixed(2)}ms ${diff > 0 ? "faster" : "slower"}`);
  console.log(`Ratio: (No Await) is ${ratio.toFixed(1)}% of (With Await) execution time.`);
})();
'

// ------------------------------------------------------------
// Node.js
// Starting Deno benchmark (10,000,000 iterations)...
// [With return await] 605.46 ms
// [No return await]   543.51 ms

// Result: Removing await was 61.96ms faster
// Ratio: (No Await) is 89.8% of (With Await) execution time.
// ------------------------------------------------------------

// ------------------------------------------------------------
// Deno
// Starting Deno benchmark (10,000,000 iterations)...
// [With return await] 638.45 ms
// [No return await]   551.04 ms

// Result: Removing await was 87.41ms faster
// Ratio: (No Await) is 86.3% of (With Await) execution time.
// ------------------------------------------------------------

// ------------------------------------------------------------
// Bun
// Starting microbenchmark (5,000,000 iterations)...
// [With return await] 506.50 ms
// [No return await]   622.49 ms

// Result: Removing await was 115.99ms slower
// Ratio: (No Await) is 122.9% of (With Await) execution time.
// ------------------------------------------------------------
```

yusukebe: Thank you for the explanation! I can understand it's intended.

> That said, I don't think the performance overhead is critical in this case. Since the CORS middleware typically runs only once per API request, I don't believe it will significantly impact the overall response performance.

You are right. So, adding `await` as you have done looks good!

And, how about removing the return type annotation? This means applying the following patch. It's okay to add it, but in many cases in `hono`, we don't add an annotation.

```diff
diff --git a/src/middleware/csrf/index.ts b/src/middleware/csrf/index.ts
index f57b99e4..c7656b68 100644
--- a/src/middleware/csrf/index.ts
+++ b/src/middleware/csrf/index.ts
@@ -100,7 +100,7 @@ export const csrf = (options?: CSRFOptions): MiddlewareHandler => {
       return (origin) => optsOrigin.includes(origin)
     }
   })(options?.origin)
-  const isAllowedOrigin = async (origin: string | undefined, c: Context): Promise<boolean> => {
+  const isAllowedOrigin = async (origin: string | undefined, c: Context) => {
     if (origin === undefined) {
       // denied always when origin header is not present
       return false
```

baseballyama: OK! done! https://github.com/honojs/hono/pull/4558/commits/bedcdafa68786dbeb201919ed836d5f059330912

baseballyama: And I applied the same changes to https://github.com/honojs/hono/pull/4559

yusukebe (approved):
LGTM!

yusukebe:
Hey @baseballyama 

Looks good to me! I'll merge and release a new version including this change. Thanks!

PR #4578, opened 2025-12-14
fix(types): fix app.on method array type inference

kosei28:
## Summary
- Fix type inference for `app.on` when passing a method array like `['GET', 'POST']`
- Previously, `Ms extends string[]` caused the type to widen to `string[]`, resulting in incorrect schema types
- Changed to `M extends string` with `methods: M[]` to preserve literal types

- [x] Add tests
- [x] Run tests


yusukebe:
@kosei28 

Wooooow great! It also works great for my local. Thank you so much!

PR #4599, opened 2025-12-25
feat(ssg): add redirect plugin

### LINKED ISSUE #4389: SSG: Support redirects

3w36zj6:
close #4389

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/helper/ssg/plugins.test.tsx
+++ b/src/helper/ssg/plugins.test.tsx
@@ -1,4 +1,5 @@
 import { Hono } from '../../hono'
+import type { RedirectStatusCode, StatusCode } from '../../utils/http-status'
 import * as plugins from './plugins'
 import { toSSG } from './ssg'
 import type { FileSystemModule } from './ssg'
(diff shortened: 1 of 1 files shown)


src/helper/ssg/ssg.test.tsx
```diff
@@ -254,6 +255,49 @@ describe('toSSG function', () => {
...
+    const content = writtenFiles['static/old.html']
+    // Should contain meta refresh
+    expect(content).toContain('meta http-equiv="refresh" content="0;url=/new"')
+    // Should contain canonical
+    expect(content).toContain('rel="canonical" href="/new"')
+    // Should contain link anchor
+    expect(content).toContain('<a href="/new">Redirecting from')
+  })
+
+  it('should skip generating a redirect HTML when 301/302 has no Location header', async () => {
```
3w36zj6: The specification mandates that redirect responses must include a `Location` header [^1]. Current implementations simply bypass generating the file if no `Location` header is present, though options include displaying a warning or throwing an error could be considered.

[^1]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Location

src/helper/ssg/index.ts
```diff
@@ -11,3 +11,4 @@ export {
   disableSSG,
   onlySSG,
 } from './middleware'
+export { redirectPlugin } from './plugins'
```
3w36zj6: Since built-in plugins should only handle lightweight ones, I decided to manage multiple plugins in a single `plugins.ts` file. This should make updating exports slightly more convenient.

src/helper/ssg/plugins.ts
```diff
@@ -0,0 +1,41 @@
...
+ * @experimental
+ * `redirectPlugin` is an experimental feature.
+ * The API might be changed.
+ */
+export const redirectPlugin = (): SSGPlugin => {
+  return {
+    afterResponseHook: (res) => {
+      if (res.status === 301 || res.status === 302) {
+        const location = res.headers.get('Location')
+        if (!location) return false
```
yusukebe: Please use this syntax that we use in this `hono` project.

```ts
if (!location) {
  return false
}
```

src/helper/ssg/plugins.ts
```diff
@@ -0,0 +1,41 @@
...
+        const html = generateRedirectHtml('', location)
```
yusukebe: You don't specify the `from` parameter. Is it expected?

src/helper/ssg/plugins.ts
```diff
@@ -0,0 +1,41 @@
+import type { SSGPlugin } from './ssg'
+
+const generateRedirectHtml = (from: string, to: string) => {
+  const html = `<!DOCTYPE html>
+<title>Redirecting to: ${to}</title>
```
yusukebe: It's a rare case if passing the script into `to`, but I think you should escape it. How about `html`?

```ts
import { html } from '../html'
```
And you can esapce `to` with ``${html`${to}`}``

src/helper/ssg/ssg.test.tsx
```diff
@@ -254,6 +255,53 @@ describe('toSSG function', () => {
...
+        writtenFiles[path] = typeof data === 'string' ? data : data.toString()
+        return Promise.resolve()
+      },
+      mkdir: vi.fn(() => Promise.resolve()),
+    }
+    const app = new Hono()
+    app.get('/old', (c) => c.redirect('/new'))
+    app.get('/new', (c) => c.html('New Page'))
+
+    await toSSG(app, fsMock, { dir: './static', plugins: [redirectPlugin()] })
```
yusukebe: What do you think of a situation where you use it with `defaultPlugin`? It works well if you put the `redirecPlugin` above the `defaultPlugin`. So, I think it's good to add a description about it on our website.

src/helper/ssg/plugins.ts
3w36zj6: Thank you for the suggestion. I'll update it later.

By the way, regarding code style verification, I checked the settings and noticed something important.

`@hono/eslint-config` specifies `'curly': ['error', 'all']`[^1], but this rule is actually disabled by `eslint-config-prettier`[^2]. According to the README, we need to manually override this setting if we want to enforce it[^3]. Should we make this adjustment?

[^1]: https://github.com/honojs/middleware/blob/main/packages/eslint-config/index.js#L33  
[^2]: https://github.com/prettier/eslint-config-prettier/blob/main/index.js#L12  
[^3]: https://github.com/prettier/eslint-config-prettier#curly

yusukebe: > Should we make this adjustment?

Thank you for checking. If it is possible, we should do it. Can you create a PR for it?

yusukebe:
@3w36zj6 

Please ping me if it's ready to review.

3w36zj6:
Sorry for the delayed response. Other tasks besides ESLint rule enforcement are currently unfinished.

Once I've completed the fixes and replied to comments, I'll send review request.

src/helper/ssg/plugins.ts
3w36zj6: I forgot to implement it.

The `Response` object in `afterResponseHook` does not contain information about the original redirect source, so we probably need to prepare a custom header like `x-hono-ssg-route` to inject the request path.

I also thought of another option, but this approach might break when running in parallel. (For production use, we would need to use a FIFO queue, but in this PoC I implemented it in a straightforward way.)

```ts
export const redirectPlugin = (): SSGPlugin => {
  const requestedUrls: string[] = []
  return {
    beforeRequestHook: (req) => {
      requestedUrls.push(new URL(req.url).pathname)
      return req
    },
    afterResponseHook: (res) => {
      if (res.status === 301 || res.status === 302) {
        const location = res.headers.get('Location')
        if (!location) {
          return false
        }
        const from = requestedUrls.shift()
        if (!from) {
          return false
        }
        const body = generateRedirectHtml(from, location)
        return new Response(body, {
          status: 200,
          headers: { 'Content-Type': 'text/html; charset=utf-8' },
        })
      }
      return res
    },
  }
}
```

src/helper/ssg/plugins.ts
3w36zj6: I refactored the code to use the html helper and added test cases related to escaping. Double escaping might not be necessary.

Also, it could be useful if the html helper had an option to get a minified string. If you cannot think of many use cases for it, it may not be needed.

src/helper/ssg/ssg.test.tsx
3w36zj6: I also think updating the documentation is essential.

Additionally, it might be helpful to add `name: string` and `requires?: string[]` to the plugin, so that we can validate the plugin order. What do you think? (I think dependency validation itself is an interesting idea, but we need to carefully discuss the interface and validation methods.)

src/helper/ssg/ssg.test.tsx
3w36zj6, line 1: Build-in plugins test should be separated to `plugins.test.tsx`?

src/helper/ssg/plugins.ts
3w36zj6, line 1: The default plugin is an object, but the redirect plugin is a function. Should we standardize it to functions for consistency?

src/helper/ssg/ssg.test.tsx
3w36zj6: I added TSDoc and tests for plugin ordering.

src/helper/ssg/ssg.test.tsx
yusukebe: The `ssg.test.tsx` became fat. So separating it is a good idea.

src/helper/ssg/plugins.ts
yusukebe: Ahh, I think `function` is better because the user can pass options like `redirectPlugin({ optionA })` if needed.

src/helper/ssg/plugins.ts
yusukebe: How about removing `from` from this PR?

At the very least, the user knows where it comes from because they access the URL directly. And I think the information about `from` is not so important.

src/helper/ssg/ssg.test.tsx
yusukebe: > I think dependency validation itself is an interesting idea, but we need to carefully discuss the interface and validation methods.

Yes. Let's discuss another place.

3w36zj6:
I've finished addressing the review.

This might be a nitpick, but should we also treat 303[^1], 307[^2], and 308[^3] as triggers for generating redirect HTML? Since all of them require the `Location` header, I feel it would make the implementation more consistent with the specification.

[^1]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/303
[^2]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/307
[^3]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/308

src/helper/ssg/index.ts
yusukebe: The `defaultPlugin` is moved. So, you should export the `defaultPlugin` from the `plugins.ts`:

```ts
export { defaultPlugin, redirectPlugin } from './plugins'
```

yusukebe:
@3w36zj6 

> This might be a nitpick, but should we also treat 303[^1], 307[^2], and 308[^3] as triggers for generating redirect HTML? 

I was also thinking the same thing. That is a good idea.

src/helper/ssg/index.ts
```diff
@@ -11,3 +11,4 @@ export {
   disableSSG,
   onlySSG,
 } from './middleware'
+export { defaultPlugin, redirectPlugin } from './plugins'
```
3w36zj6, line 14: I'm wondering whether I should use `export *`.

```ts
export * from './plugins'
```

yusukebe: I think the current explicit implementation is okay.

yusukebe (approved):
LGTM!

yusukebe:
@3w36zj6 

Looks good! This change is a "feature", so we will include it in the next minor release v4.12.0. It may be released not too far. Thanks!

PR #4611, opened 2026-01-05
fix(client): exclude $all from ClientRequest type

### LINKED ISSUE #3184: Type definition of the `app.all` client.

4.5.1

Deno 1.45.3

1. Clone https://github.com/nakasyou/hono-client-app-all
2. Run `deno run --allow-net main.ts`
3. Console shows `"Hello ALL!"`

TypeScript should throw the error because there is no method: `ALL`, so `$all` is invalid.

### What do you see instead?

_No response_

### Additional information

Related: https://github.com/orgs/honojs/discussions/3143
There is the runtime that runtime that can't fetch incorrect methods.

paveg:
## Summary

- Fixes the RPC client incorrectly exposing `$all` as a valid method when using `app.all()`
- Since HTTP has no "ALL" method, the client should not expose `$all`
- Users should use specific HTTP methods like `$get`, `$post`, etc.

## Changes

- Modified `ClientRequest` type to exclude `$all` using `Exclude<M, '$all'>`
- Added type tests to verify `$all` is not exposed on the client

## Related

- Closes #3184
- Related discussion: https://github.com/orgs/honojs/discussions/3143

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/client/types.ts
+++ b/src/client/types.ts
@@ -68,8 +68,14 @@ export type ClientRequest<Prefix extends string, Path extends string, S extends 
   [M in keyof ExpandAllMethod<S>]: ExpandAllMethod<S>[M] extends Endpoint & { input: infer R }
     ? R extends object
       ? HasRequiredKeys<R> extends true
-        ? (args: R, options?: ClientRequestOptions) => Promise<ClientResponseOfEndpoint<ExpandAllMethod<S>[M]>>
-        : (args?: R, options?: ClientRequestOptions) => Promise<ClientResponseOfEndpoint<ExpandAllMethod<S>[M]>>
+        ? (
+            args: R,
+            options?: ClientRequestOptions
+          ) => Promise<ClientResponseOfEndpoint<ExpandAllMethod<S>[M]>>
+        : (
+            args?: R,
+            options?: ClientRequestOptions
+          ) => Promise<ClientResponseOfEndpoint<ExpandAllMethod<S>[M]>>
       : never
     : never
 } & {


yusukebe:
Hi @paveg 

Thank you for the PR.

We have a problem to consider. If you define `app.all`, you can send any method from the client. So should `$get()` typed? In the following code, `client['all-route'].$get()` is not typed. But the client can send GET request and does not throw runtime errors.

```ts
const app = new Hono()

const routes = app.all('/all-route', (c) => {
  return c.json({ method: c.req.method })
})

const client = hc<typeof routes>('http://localhost', {
  fetch: app.request,
})

const res = await client['all-route'].$get() // $get is not typed
console.log(res.status) // 200
```

src/client/types.ts
```diff
@@ -44,7 +44,7 @@ export type ClientRequestOptions<T = unknown> = {
     })
 
 export type ClientRequest<Prefix extends string, Path extends string, S extends Schema> = {
-  [M in keyof S]: S[M] extends Endpoint & { input: infer R }
+  [M in keyof S as Exclude<M, '$all'>]: S[M] extends Endpoint & { input: infer R }
```
yusukebe: I think we have to use the constant instead of `all` hard-coded.

https://github.com/honojs/hono/blob/18324b8b6c668b602199a437456f1c22925cb803/src/router.ts#L13

yusukebe:
Hey @nakasyou 

What do you think of ths problem? https://github.com/honojs/hono/pull/4611#issuecomment-3713656217

nakasyou:
Hi @yusukebe, I agree with you about the problem you showed. An idea to solve this is creating a type to define methods such as `$get`, `$post`, `$put`, and others. Using custom HTTP methods rather than official HTTP methods, GET, POST, and others, is not popular. Therefore, this way is not bad.

paveg:
@yusukebe, @nakasyou 

Thank you for the feedback.
You're right, simply excluding `$all` doesn't expose standard HTTP methods (`$get`, `$post`, etc.) for routes defined with `app.all()`.
Here's a proposed solution that also uses constants from `router.ts` instead of hardcoding:

```ts
import { METHODS, METHOD_NAME_ALL_LOWERCASE } from '../router'

// Use constants for consistency
type MethodNameAll = `$${typeof METHOD_NAME_ALL_LOWERCASE}`  // '$all'
type StandardMethods = `$${(typeof METHODS)[number]}`        // '$get' | '$post' | '$put' | '$delete' | '$options' | '$patch'

// Expand $all into all standard HTTP methods
type ExpandAllMethod<S> =
  MethodNameAll extends keyof S
    ? { [M in StandardMethods]: S[MethodNameAll] } & Omit<S, MethodNameAll>
    : S

// Updated ClientRequest type
export type ClientRequest<Prefix extends string, Path extends string, S extends Schema> = {
  [M in keyof ExpandAllMethod<S>]: ExpandAllMethod<S>[M] extends Endpoint & { input: infer R }
    ? // ... rest of the implementation
    : never
}
```

This would make `client['all-route'].$get()`  properly typed, matching the runtime behavior.
Should I proceed with this approach?

yusukebe:
@paveg Yeees! That approch is great. I love it.

src/client/types.ts
```diff
@@ -44,11 +65,11 @@ export type ClientRequestOptions<T = unknown> = {
     })
 
 export type ClientRequest<Prefix extends string, Path extends string, S extends Schema> = {
-  [M in keyof S]: S[M] extends Endpoint & { input: infer R }
+  [M in keyof ExpandAllMethod<S>]: ExpandAllMethod<S>[M] extends Endpoint & { input: infer R }
```
paveg, line 68: ## Performance Measurement

Measured TypeScript type-checking performance before and after this change, just to make sure.

### Environment

- TypeScript: `tsc --noEmit --extendedDiagnostics`
- Machine: macOS (Darwin 25.2.0)

### Results

| Metric | Before | After | Diff |
|--------|--------|-------|------|
| Memory used | 704,748K | 713,348K | +8,600K (+1.2%) |
| Check time | 2.76s | 2.74s | -0.02s |
| Total time | 3.28s | 3.38s | +0.10s |

### Analysis

The `ExpandAllMethod<S>` type adds minimal overhead:

- Single conditional type check (`MethodNameAll extends keyof S`)
- Mapped type over 6 keys (`StandardMethods`)
- Built-in `Omit` utility type

No recursive type computation is introduced. The slight memory increase (~1.2%) is negligible and within acceptable bounds.

paveg:
@yusukebe I've implemented it (53c413a24754015a2363a7cbab791a8e777917cd), so kindly review this again. Thanks.

yusukebe (approved):
LGTM!

yusukebe:
@paveg 

Awesome work! The performance decrease is unwelcome but slight. It’s acceptable. Correcting the behavior of `$all` is worth more than that. Thank you for the contribution!

paveg:
@yusukebe
I appreciate your support for merging this improvement 👍 
I'll continue contributing to the repo.

Issue #4103, opened 2025-04-22
Validator for queries inconsistent


floratmin:
### What version of Hono are you using?

4.7.2

### What runtime/platform is your app running on? (with version if possible)

Cloudflare Workers

### What steps can reproduce the bug?

Have a route with a query parameter and add a validator for this query parameter. If I use this parameter only one time, the value to be validated is a string, otherwise an array of strings.

The problem comes from this line in the validator function:
https://github.com/honojs/hono/blob/15a83b125cae5a772f265ce448974c03efa5e3bc/src/validator/validator.ts#L126

### What is the expected behavior?

I would expect to validate only the array version for query parameters.

yusukebe:
Hi @floratmin 

I can understand your problem 50%. So, can you provide the specific code and instructions to reproduce the bug?

floratmin:
The validator has to be typed as `string | string[]`, depending on receiving the query parameter a single or multiple times. Depending how I access the value in the response, I get type `string | string[]` or type `string[]`.

```
app.get('/foo', arktypeValidator('query', type({foo: 'string|string[]'})), async (c) => {
  const foo = c.req.valid('query').foo; // type is string | string[]
  const foo2 = c.req.queries('foo'); // type is string[]
});
```

yusukebe:
@floratmin 

I can't understand the problem. The `foo` should be `string | string[]` because it's is specified with `type({foo: 'string|string[]'})`. Can you provide the test code to explain it?

floratmin:
The validator does not specify the type, but validates it. I have to validate in this way, otherwise it would not allow to handle both cases. 
The request query comes in as `string[]`, but the validator extracts the string when the array has a length of one.So I have to validate both cases when I want to allow the query parameter to be used one time and more than one time. 
If I would validate the type as `type({foo: 'string'})`, it throws when the query parameter is used more than one time. When I validate the type as `type({foo: 'string[]'})`, it throws when the query parameter is used only a single time.
So when fetching `https://api.example.com/endpoint?foo=a` and `https://api.example.com/endpoint?foo=a&foo=b` I have to validate one time as `string` and one time as `string[]`.
I think that this is confusing and extra work for validating query parameters, I have to remap the single case again to an array. But it could introduce for some APIs breaking changes.

yusukebe:
@floratmin 

Hmm. I understand the issue, but this is not a bug but a specification. I cannot find a good way to address this without breaking the current API. However, it can be handled on the validator side. How about this?

[...]

PR #4615, opened 2026-01-07
fix(validator): make query params always return string[] for consistency

BBAK-jun:
# Summary

- Fixes query parameter validator having inconsistent types: single values were string, multiple values were string[]
- Now query parameters always return string[] for consistency
- This aligns with the behavior of c.req.queries() which always returns arrays

# Changes

- validator.ts: Use c.req.queries() directly instead of converting single values to strings
- types.ts: Change ValidationTargets['query'] from Record<string, string | string[]> to Record<string, string[]>
- validator/utils.ts: Update InferInput query type to string[]

# Related

- #4103

### DIFF
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -323,7 +323,7 @@ describe('Basic - query, queries, form, path params, header and cookie', () => {
   it('Should get 200 response - query', async () => {
     const res = await client.search.$get({
       query: {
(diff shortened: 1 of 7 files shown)


yusukebe:
@BBAK-jun 

Thank you for the PR. I have another idea and will create a PR. Please wait a bit.

PR #4618, opened 2026-01-07
docs(bun/websocket): Fixed a typo in hono/bun deprecation message and updated test.

Itsnotaka:
Fix typo in createBunWebSocket deprecation message

The deprecation notice incorrectly referenced createWebSocket which doesn't exist. Changed to upgradeWebSocket which is the actual export users should migrate to.

Also fixed the test assertion which was comparing variables to themselves instead of verifying the factory returns the correct exports.

Tests pass: `bun run vitest run src/adapter/bun/websocket.test.ts`


yusukebe:
Hey @Itsnotaka 

Thanks. Nice catch!

PR #4626, opened 2026-01-13
feat(bun): export getBunServer

artemtam:
Exported `getBunServer` from `hono/bun` to be able to access Bun server instance in handlers. It was possible before via `c.env`, but this change makes it straightforward.

Also, made `getBunServer` a generic that requires a server type, so users can set it correctly. It makes more clear internally as well - `getConnInfo.ts` and `websocket.ts` specify fields they need. Let me know if you don't like it.

The usage would look like that:

```typescript
import { Hono } from 'hono';
import type { Server } from 'bun';
import { getBunServer } from 'hono/bun';

const app = new Hono();

app.post('/', (c) => {
  const server = getBunServer<Server<unknown>>(c);
 
  if (server) {
    server.publish('topic', 'message');
  }
});

```

- [x] Add tests (N/A)
- [x] Run tests

### DIFF
--- a/src/adapter/bun/server.ts
+++ b/src/adapter/bun/server.ts
@@ -6,8 +6,9 @@ import type { Context } from '../../context'
 
 /**
  * Get Bun Server Object from Context
+ * @template T - The type of Bun Server
  * @param c Context
  * @returns Bun Server
  */
-export const getBunServer = <ServerT>(c: Context): ServerT | undefined =>
-  ('server' in c.env ? c.env.server : c.env) as ServerT | undefined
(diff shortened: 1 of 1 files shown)


yusukebe:
@artemtam 

Thank you for the PR. I think this is a good feature.

Hey @nakasyou! Can you also review this?

yusukebe:
@artemtam 

Looks good! It's a new feature, but a slight change. So I'm planning to add it in a patch version. Please wait until the release.

artemtam:
@yusukebe great, thanks!

PR #4627, opened 2026-01-13
fix(url): ignore fragment identifiers in getPath()

### LINKED ISSUE #4440: Fragment identifiers cause incorrect route parameter parsing in Service Worker

4.9.9

browser service worker (Chrome)

1: Run Hono in a Service Worker context using `fire()` from `hono/service-worker`
2: Define routes
```typescript
   app.get('/users/', (c) => c.html(...))
   app.get('/users/:id', (c) => c.html(...))
```
3: User navigates to the page with a fragment: 
```
GET /users/#user-list
GET /users/1#profile-section
```

- `/users/#user-list` matches `/users/` route
- `/users/1#profile-section` matches `/users/:id` route with `id = "1"`

### What do you see instead?

- `/users/#user-list` matches `/users/:id` route with `id = "#user-list"`
- `/users/1#profile-section` matches `/users/:id` but with `id = "1#profile-section"`

This occurs because Service Workers receive the full URL including fragments, unlike traditional web servers where browsers strip fragments before sending requests. Hono's route matcher treats `#` as a regular character.

### Additional information

_No response_

sano-suguru:
## Summary

- Modified `getPath()` to strip fragment identifiers (`#`) from URLs
- This fixes incorrect route parameter parsing in Service Worker contexts

Fixes #4440

## Changes

In Service Worker contexts, URLs may contain fragment identifiers (e.g., `/users/1#profile-section`) which were incorrectly included in route parameters. This change modifies `getPath()` to strip fragment identifiers, consistent with how query strings are already handled.

### Before
```
/users/#user-list → id = "#user-list"
/users/1#profile-section → id = "1#profile-section"
```

### After
```
/users/#user-list → matches /users/
/users/1#profile-section → id = "1"
```

## Implementation

Based on the approach suggested by @usualoma in #4440:
- Added `#` (charCode 35) to the loop termination condition
- Handle fragment identifiers in the percent-encoded path branch

## Test plan

- [x] Added tests for URLs with fragments
- [x] Added tests for URLs with both query strings and fragments  
- [x] Added tests for percent-encoded paths with fragments
- [x] All existing tests pass

### DIFF
--- a/src/utils/url.test.ts
+++ b/src/utils/url.test.ts
@@ -144,6 +144,11 @@ describe('url', () => {
       expect(path).toBe('/search')
     })
 
+    it('getPath - with percent encoding only (no query or fragment)', () => {
+      const path = getPath(new Request('https://example.com/hello%20world'))
+      expect(path).toBe('/hello world')
+    })
+
     it('getPath - with percent encoding and fragment', () => {
       let path = getPath(new Request('https://example.com/hello%20world#section'))
       expect(path).toBe('/hello world')


sano-suguru:
## Fixed: Percent-encoded path with fragment containing query-like characters

The previous implementation had a bug where it would check for `?` first, and only look for `#` if `?` was not found:

```typescript
let separator = url.indexOf('?', i)
if (separator === -1) {
  separator = url.indexOf('#', i)
}
```

This caused incorrect behavior when a URL contains a fragment with query-like characters (e.g., `https://example.com/hello%20world#section?foo=bar`). In such cases, per RFC 3986, the `?` after `#` is part of the fragment, not a query delimiter.

### Fix

Changed to find whichever delimiter (`?` or `#`) comes first using `Math.min()`:

```typescript
const queryIndex = url.indexOf('?', i)
const hashIndex = url.indexOf('#', i)
const separator =
  queryIndex === -1 ? hashIndex : hashIndex === -1 ? queryIndex : Math.min(queryIndex, hashIndex)
```

### Added Test

```typescript
it('getPath - with percent encoding and fragment containing query-like chars', () => {
  const path = getPath(new Request('https://example.com/hello%20world#section?foo=bar'))
  expect(path).toBe('/hello world')
})
```

All tests pass.

sano-suguru:
Thank you for the review and the optimization suggestion! I've updated the code as you suggested.

The restructured ternary now avoids the redundant `separator === -1` check, reducing evaluations from 3 to 2 for URLs like `/path?x=1`.

I also added a test case to cover the `undefined` branch (when both queryIndex and hashIndex are -1).

(Sorry if my English is a bit rough - it's not my native language!)

src/utils/url.ts
```diff
@@ -111,13 +111,22 @@ export const getPath = (request: Request): string => {
     const charCode = url.charCodeAt(i)
     if (charCode === 37) {
       // '%'
-      // If the path contains percent encoding, use `indexOf()` to find '?' and return the result immediately.
+      // If the path contains percent encoding, use `indexOf()` to find '?' or '#' and return the result immediately.
       // Although this is a performance disadvantage, it is acceptable since we prefer cases that do not include percent encoding.
       const queryIndex = url.indexOf('?', i)
-      const path = url.slice(start, queryIndex === -1 ? undefined : queryIndex)
+      const hashIndex = url.indexOf('#', i)
+      const end =
```
yusukebe, line 118: How about the following implementation? This is more short and readable.

```ts
const end = Math.min(
  queryIndex === -1 ? url.length : queryIndex,
  hashIndex === -1 ? url.length : hashIndex
)
```

This runs `Math.min` every time, but in some cases that contain percent encoding, the path has a query.

@sano-suguru @usualoma What do you think of it?

usualoma: @yusukebe 
Sorry I missed that. Either one seems fine.
The character count after minification is probably the same, right?

```js
Math.min(a===-1?u.length:a,b===-1?u.length:b)
```

```js
a===-1?b===-1?void 0:b:b===-1?a:Math.min(a,b)
```

Performance-wise, there probably won't be a noticeable difference in production. I prefer the current implementation, but either one is fine.

yusukebe: > Sorry I missed that. Either one seems fine.

No problem!

> The character count after minification is probably the same, right?

Ah, yes. So it's just for readability. If you feel both are okay, I think we should keep the current implementation.

yusukebe:
@sano-suguru 

I've left a small comment, but let's go with this implementation. Looks good. Thanks!

PR #4635, opened 2026-01-15
feat(jsx/dom): Extend the render function to accept strings as containers

tokiya-takai:

- [x] Add tests
- [x] Run tests

### Summary

```ts
render(<App />, document.getElementById('app')!)
```

This is explicit, but I find it inconvenient in the following two respects:
1. Writing `document.getElementById()` is cumbersome, yet it must be written for each project.
2. Using an exclamation mark to assert existence is problematic, because it does not necessarily exist in all cases.

Taking these points into account, I made the following improvements:
1. By entering only the ID name, it is handled internally.
2. Since `app` becomes the default value, cases where no input is required are now possible.
3. Full compatibility with the existing specification.

Additionally, the following has not been implemented:
1. Retrieving DOM elements by class name. Since multiple elements can share the same class, using an ID is recommended.

### DIFF
--- a/src/jsx/dom/render.ts
+++ b/src/jsx/dom/render.ts
@@ -778,7 +778,6 @@ export const render = (jsxNode: Child, container: Container | string = 'app'): v
     targetContainer = container
   }
 
-   
   renderNode(
     buildNode({ tag: '', props: { children: jsxNode } } as any) as NodeObject,
     targetContainer


yusukebe:
@tokiya-takai 

Does that mean you want to write `render(<App />, 'app')`?

tokiya-takai:
@yusukebe 
Yes, that’s correct!
However, in most cases I’d like to keep things simple and just use `render(<App />)` with the default settings.

usualoma:
Hi @tokiya-takai, thank you for your feature suggestion.

I appreciate the idea, but I think keeping the `document.getElementById()` call works well for now. I'd prefer to avoid adding incompatibility with React at this point.

tokiya-takai:
@usualoma 
Thank you for your response!
I see—it follows the specifications of `ReactDOM.render()`.
Indeed, this could potentially be a drawback for React users.
I understand the intention now!

PR #4636, opened 2026-01-15
feat(client): $path

### LINKED ISSUE #4186: Support for Relative Path in Hono Client

Currently, the Hono client always returns an absolute URL object when using the $url() method. This creates issues when working with relative paths, especially in environments where the base URL might change or when building URLs that need to remain relative.

**Current Behavior**

```ts
const client = hc('/api');
const url = client.users.$url(); // Returns URL object with absolute path
```

The code in [client.js](https://github.com/honojs/hono/blob/53656e126df4855969afa55603cd063343d65552/src/client/client.ts#L169)  always returns [new URL(result)](https://github.com/honojs/hono/blob/53656e126df4855969afa55603cd063343d65552/src/client/client.ts#L169), which converts relative paths to absolute URLs based on the current origin. and throwing an error.

**Expected Behavior**

Provide an option to return relative URLs as strings instead of URL objects.

ShaMan123:

- [x] Add tests
- [x] Run tests

closes #4186

### DIFF
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -420,7 +420,7 @@ describe('Basic - $url()', () => {
     ).toBe('http://fake/content/search?page=123&limit=20')
   })
 
-  it.each(['http://fake', 'http://fake/', 'http://fake/api'])(
+  it.each(['http://fake', 'http://fake/', 'http://fake//', 'http://fake/api'])(
     'Should return a correct path via $path() regardless of %s',
     async (baseURL) => {
       const client = hc<typeof app>(baseURL)
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -176,7 +176,10 @@ export const hc = <T extends Hono<any, any, any>, Prefix extends string = string
         }
       }
       result = removeIndexString(result)
-      return method === 'url' ? new URL(result) : mergePath('', result.replace(baseUrl, ''))
+      if (method === 'url') {
+        return new URL(result)
+      }
+      return result.slice(baseUrl.replace(/\/+$/, '').length).replace(/^\/?/, '/')
     }
     if (method === 'ws') {
       const webSocketUrl = replaceUrlProtocol(


src/client/types.test.ts
```diff
@@ -58,6 +60,19 @@ describe('without the leading slash', () => {
...
+      client.foo[':id'].baz.$path({
+        param: { id: '123' },
+      })
+    ).toEqualTypeOf<'/foo/123/baz'>()
+    expectTypeOf(
+      client.foo[':id'].baz.$path({
+        param: { id: '123' },
+        query: { q: 'hono' },
+      })
+    ).toEqualTypeOf<TypedURL<'http:', 'localhost', '', '/foo/123/baz', `?${string}`>['pathname']>()
```
ShaMan123: This doesn't seem right

src/client/types.test.ts
```diff
@@ -77,3 +92,21 @@ describe('with the leading slash', () => {
...
+  const client = hc<typeof app, 'http://localhost/api'>('http://localhost/api')
+  it('$path', () => {
+    expectTypeOf(client.foo.$path()).toEqualTypeOf<'/foo'>()
+    expectTypeOf(client.foo.bar.$path()).toEqualTypeOf<'/foo/bar'>()
+    expectTypeOf(
+      client.foo[':id'].baz.$path({
+        param: { id: '123' },
+        query: { q: 'hono' },
+      })
+    ).toEqualTypeOf<'/foo/123/baz'>()
```
ShaMan123: same

src/client/types.test.ts
yusukebe: So, the correct test is below, right?

```ts
expectTypeOf(
  client.foo[':id'].baz.$path({
    param: { id: '123' },
    query: { q: 'hono' },
  })
).toEqualTypeOf<`/foo/123/baz?${string}`>()
```

yusukebe: @ShaMan123 

Try to apply this patch:

```diff
diff --git a/src/client/types.test.ts b/src/client/types.test.ts
index ce0c893e..6b186ff8 100644
--- a/src/client/types.test.ts
+++ b/src/client/types.test.ts
@@ -72,7 +72,7 @@ describe('without the leading slash', () => {
         param: { id: '123' },
         query: { q: 'hono' },
       })
-    ).toEqualTypeOf<TypedURL<'http:', 'localhost', '', '/foo/123/baz', `?${string}`>['pathname']>()
+    ).toEqualTypeOf<`/foo/123/baz?${string}`>()
   })
 })

@@ -107,6 +107,6 @@ describe('with base URL pathname', () => {
         param: { id: '123' },
         query: { q: 'hono' },
       })
-    ).toEqualTypeOf<'/foo/123/baz'>()
+    ).toEqualTypeOf<`/foo/123/baz?${string}`>()
   })
 })
diff --git a/src/client/types.ts b/src/client/types.ts
index f8f9fd3a..816c0bc9 100644
--- a/src/client/types.ts
+++ b/src/client/types.ts
@@ -81,7 +81,7 @@ export type ClientRequest<Prefix extends string, Path extends string, S extends
       | undefined = undefined,
   >(
     arg?: Arg
-  ) => BuildPathname<Path, Arg>
+  ) => BuildPath<Path, Arg>
 } & (S['$get'] extends { outputFormat: 'ws' }
     ? S['$get'] extends { input: infer I }
       ? {
@@ -134,6 +134,8 @@ type BuildPathname<P extends string, Arg> = Arg extends { param: infer Param }
   ? `${ApplyParam<TrimStartSlash<P>, Param>}`
   : `/${TrimStartSlash<P>}`

+type BuildPath<P extends string, Arg> = `${BuildPathname<P, Arg>}${BuildSearch<Arg, 'query'>}`
+
 type BuildTypedURL<
   Protocol extends string,
   Host extends string,
```

yusukebe:
Hey @ShaMan123 !

I added the code to support an edge case, but it looks good! Let's go with this. This change is a new feature; we’ll include it in the next minor version, v4.12.0. Maybe the date isn't too far away. Please wait a bit. Thank you for your contribution!

ShaMan123:
Thanks!

PR #4641, opened 2026-01-21
refactor(jwks): mark allowedAlgorithms, so the user can pass a `const…

nikeee:
Marks the parameter as readonly, to signal that the array wont be changed.

```js
const config = {
  allowedAlgorithms: ["RS256"],
} as const;

// error, because verifyWithJwks doesn't allow a const algorithm list
verifyWithJwks("foo", { allowedAlgorithms: config.allowedAlgorithms });
```

This works now. This change is non-breaking.


BarryThePenguin (approved):
Nice! 👍🏻

yusukebe:
@nikeee Thanks!

PR #4645, opened 2026-01-24
feat(basic-auth): add context key and callback options

### LINKED ISSUE #4192: Pass username to context from basicAuth

As far as I can see from [the source](https://github.com/honojs/hono/blob/53656e126df4855969afa55603cd063343d65552/src/middleware/basic-auth/index.ts#L90) (and from [the docs](https://hono.dev/docs/middleware/builtin/basic-auth)), `basicAuth` doesn't pass `username` anywhere. I need to get the username in the route handler, and it seems I have to parse the header again, duplicating half of the `basicAuth` (the `auth` util, to be exact).

What I'm proposing is to pass the username to the context.

How exactly to do so is an open quesion: should it be [.set](https://hono.dev/docs/api/context#set-get)('username', username)? should the name be configurable? How to make .get understand that that variable was set? ([this](https://github.com/honojs/hono/issues/3341) looks related). Since one can use `'username'` in their context already, it may be useful to add to `BasicAuthOptions` something like

    passUsernameToContext?: boolean | string

where boolean means "use 'username'", and string is a custom context variable. But if the types are inferred correctly, I think just `boolean` would suffice: if one used `'username'` for something else, they'll either don't need to pass the username from the middleware, or can rename their variable.

AprilNEA:
Close #4192

Add two new options to basicAuth middleware:
- `usernameContextKey`: stores authenticated username in context
  - `true`: uses default key 'basicAuthUsername'
  - `string`: uses custom key name
- `onAuthSuccess`: callback invoked after successful authentication

This allows route handlers to access the authenticated username without
re-parsing the Authorization header.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/basic-auth/index.ts
+++ b/src/middleware/basic-auth/index.ts
@@ -107,7 +107,6 @@ export const basicAuth = (
     if (requestUser) {
       if (verifyUserInOptions) {
         if (await options.verifyUser(requestUser.username, requestUser.password, ctx)) {
-          // Call onAuthSuccess callback if provided
           if (options.onAuthSuccess) {
             await options.onAuthSuccess(ctx, requestUser.username)
           }
@@ -121,7 +120,6 @@ export const basicAuth = (
             timingSafeEqual(user.password, requestUser.password, options.hashFunction),
           ])
           if (usernameEqual && passwordEqual) {
-            // Call onAuthSuccess callback if provided
             if (options.onAuthSuccess) {
               await options.onAuthSuccess(ctx, requestUser.username)
             }


src/middleware/basic-auth/index.ts
```diff
@@ -18,12 +18,34 @@ type BasicAuthOptions =
...
       verifyUser: (username: string, password: string, c: Context) => boolean | Promise<boolean>
       realm?: string
       hashFunction?: Function
       invalidUserMessage?: string | object | MessageFunction
+      /**
+       * If set, the authenticated username will be stored in the context.
+       * - `true`: stores under the key `'basicAuthUsername'`
+       * - `string`: stores under the specified custom key
+       */
+      usernameContextKey?: boolean | string
```
yusukebe: `usernameContextKey` is needed? I think it is unnecessary. You can use the following method:

```ts
basicAuth({
  username: 'hono',
  password: 'secret',
  onAuthSuccess: (c, username) => {
    c.set('basicAuthUsername', username)
  },
})
```

It's simple.

AprilNEA: `usernameContextKey` is indeed just a shortcut, and you are right, we should keep it simple.

yusukebe: @AprilNEA Can you update it?

yusukebe:
Hi @AprilNEA 

I've updated some points myself. Looks good! I'll include this change in the next minor version. Thanks!

Issue #4640, opened 2026-01-21
Build failure when included in TS project


michielbdejong:
### What version of Hono are you using?

4.11.4

### What runtime/platform is your app running on? (with version if possible)

node + npm or pnpm 

### What steps can reproduce the bug?

See https://github.com/michielbdejong/node-typescript-boilerplate/pull/1 for a minimal example where I include hono in an otherwise (almost) empty project, and both `pnpm build` and `npm run build` start failing. You can see it failing in GitHub Actions on that PR.

### What is the expected behavior?

My TS project that uses hono would build successfully, both on my laptop and in GitHub Actions

### What do you see instead?

```
node_modules/.pnpm/hono@4.11.4/node_modules/hono/dist/types/request.d.ts:19:5 - error TS1166: A computed property name in a class property declaration must have a simple literal type or a 'unique symbol' type.

19     [GET_MATCH_RESULT]: Result<[unknown, RouterRoute]>;
       ~~~~~~~~~~~~~~~~~~
```

yusukebe:
@michielbdejong 

Can you add a `compilerOptions.skipLibCheck` to the `tsconfig.json`?

```ts
{
  "compilerOptions": {
    "skipLibCheck": true
  },
}
```

michielbdejong:
Confirmed that `skipLibCheck` fixes it for 4.11.4 and also confirmed that with 4.11.7 my build succeeds even without `skipLibCheck`.

Thanks! <3

PR #4646, opened 2026-01-24
refactor(request): replace private symbol with static method

BarryThePenguin:
Replaces usage of the private `GET_MATCH_RESULT` symbol with a static `matchedRoutes()` method

This keeps `#matchResult` private while still allowing access to `matchedRoutes()`

### DIFF
--- a/src/helper/route/index.ts
+++ b/src/helper/route/index.ts
@@ -29,7 +29,7 @@ import { getPattern, splitRoutingPath } from '../../utils/url'
  * })
  * ```
  */
-export const matchedRoutes = (c: Context): RouterRoute[] => HonoRequest.matchedRoutes(c)
+export const matchedRoutes = (c: Context): RouterRoute[] => HonoRequest.matchedRoutes(c.req)
 
 /**
  * Get the route path registered within the handler
--- a/src/request.ts
+++ b/src/request.ts
@@ -1,5 +1,4 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import type { Context } from './context'
 import { HTTPException } from './http-exception'
 import type { Result } from './router'
 import type {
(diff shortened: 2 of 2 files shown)


yusukebe:
Hey @BarryThePenguin 

Thank you for the PR! One of the purposes of this PR is to fix the build problem, like https://github.com/honojs/hono/issues/4640, right?

I think it is worth accepting, but for details, we have to be careful when updating the HonoRequest, and this change will increase the bundle size very slightly.

BarryThePenguin:
> Thank you for the PR! One of the purposes of this PR is to fix the build problem, like https://github.com/honojs/hono/issues/4640, right?

Yes, I was doing some investigating and found this alternative approach. I thought I'd share it to see what others thought.

> I think it is worth accepting, but for details, we have to be careful when updating the HonoRequest, and this change will increase the bundle size very slightly.

Yeah, that makes sense. I wasn't able to find out why a symbol was chosen originally. If it works then there's no reason to change it. There's some added type safety using a static method, but it's all internal, so there's no real benefit for users.

usualoma:
hi @BarryThePenguin and @yusukebe 

Thank you for the suggestion!

`import type { Context } from './context'` in `src/request.ts` is not ideal in terms of dependency order.

I haven't fully verified if this resolves the issue, but I think fixing it in the direction suggested in #4651 would be best.

BarryThePenguin:
> `import type { Context } from './context'` in `src/request.ts` is not ideal in terms of dependency order.

Instead of passing `Context` as an argument, it could just be a `HonoRequest`. That would remove the import and any dependency ordering

> I haven't fully verified if this resolves the issue, but I think fixing it in the direction suggested in #4651 would be best.

Sounds good 👍🏻

usualoma:
Hi @BarryThePenguin,

Thank you for the update!

I apologize for leaving an incomplete comment last time. My thoughts are as follows, so I'd like to maintain the current implementation approach.

* `matchedRoutes` is strictly a helper; it should not be utility functionality held by the `HonoRequest` class.
* While `HonoRequest` internally holds `#matchResult`, this should be hidden, and ideally, no means to access it should be provided.

yusukebe:
> `matchedRoutes` is strictly a helper; it should not be utility functionality held by the `HonoRequest` class.

I agree with this!

BarryThePenguin:
Ok, thank you for the extra details

PR #4647, opened 2026-01-24
test(netlify): add handler tests

rokasta12:

- [x] Add tests
- [x] Run tests

Adds comprehensive tests for the Netlify adapter handler, covering:
- Basic request handling (GET, POST, path/query params)
- Netlify context passthrough (IP, geo data)
- Response handling (JSON, status codes, headers)
- Error handling (404, custom error handlers)
- Middleware execution
- basePath compatibility

14 test cases with 100% coverage on `handler.ts`.

### DIFF
--- a/src/adapter/netlify/handler.test.ts
+++ b/src/adapter/netlify/handler.test.ts
@@ -1,0 +1,282 @@
+import { Hono } from '../../hono'
+import { handle } from './handler'
+
+describe('Netlify Adapter', () => {
+  // Mock Netlify context
+  const createMockContext = (overrides = {}) => ({
+    geo: {
+      city: 'San Francisco',
+      country: { code: 'US', name: 'United States' },
+    },
+    ip: '192.168.1.1',
+    requestId: 'test-request-id',
+    ...overrides,
+  })
+
+  describe('Basic Request Handling', () => {
+    it('Should return 200 response for GET request', async () => {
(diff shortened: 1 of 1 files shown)


rokasta12:
Closing this in favor of #4649 (already merged), which I think covers the only contract this adapter has.

Looking at it again with fresh eyes: `handle()` in `src/adapter/netlify/handler.ts` is three lines — a passthrough wrapper around `app.fetch(req, { context })`. The tests I added here (basic GET/POST routing, path params, query params, content types, custom status codes, middleware, basePath, 404s, custom error handlers) all exercise Hono's core router and middleware pipeline. They would behave identically against `app.fetch(req, { context })` called directly, so they don't actually test the Netlify integration boundary.

The one piece of Netlify-specific contract — that the runtime context object reaches user code as `c.env.context` — is now covered by `src/adapter/netlify/conninfo.test.ts` (merged in #4649), including the `context.ip` happy path, missing `ip`, and missing `context` cases.

Adding 297 lines of duplicate Hono-core coverage isn't a net positive for the maintenance surface, especially when those tests would also need to be repeated for every other adapter on the same logic. If a real coverage gap shows up later (adapter-specific request/response transform, an actual Netlify quirk), I'd rather open a focused PR for it.

Thanks for leaving it open this long while I figured out the right scope.

PR #4650, opened 2026-01-25
refactor: use `unique symbol` for more accurate typing.

### LINKED ISSUE #4640: Build failure when included in TS project

usualoma:
fixes #4640

Explicitly annotating with `: symbol` causes type widening, converting the inferred `unique symbol` type to a general `symbol` type. By removing the type annotation, TypeScript automatically infers `unique symbol` for const declarations initialized with Symbol().

### DIFF
--- a/src/request/constants.ts
+++ b/src/request/constants.ts
@@ -1,1 +1,1 @@
-export const GET_MATCH_RESULT: symbol = Symbol()
+export const GET_MATCH_RESULT = Symbol()


usualoma:
I see, so I have to explicitly specify the type.

PR #4655, opened 2026-01-29
Feat: export crypto utilities (sha256 , sha1 , md5)

holoflash:
The crypto utilities `sha256`, `sha1`, `md5` currently used internally are perfect replacements for external libraries like crypto-js or the Node.js crypto module.

Exposing them from the main entry point enables IDE auto-imports and makes them more easily discovered.

- [ ] Add tests (not applicable)
- [x] Run tests

### DIFF
--- a/src/utils/crypto.ts
+++ b/src/utils/crypto.ts
@@ -12,18 +12,54 @@ type Algorithm = {
 
 type Data = string | boolean | number | JSONValue | ArrayBufferView | ArrayBuffer
 
+/**
+ * Calculates the SHA-256 digest of the given data.
+ *
+ * @param data - The data to calculate the SHA-256 digest for.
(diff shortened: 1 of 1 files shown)


yusukebe:
The top-level `Hono` should export them. We can't accpet this PR. Sorry.

yusukebe:
The `Hono` should export only what is related to Hono's core feature. The crypto utilities functions are not.

holoflash:
> The `Hono` should export only what is related to Hono's core feature. The crypto utilities functions are not.

Ok, I understand. I assume then that my latest push to this branch is not what you meant:
With this change, the utils can be imported like this:
`import { sha256, sha1, md5 } from "hono";`

(Not showing because PR is closed)

PR #4657, opened 2026-02-02
feat: add methodNotAllowed handler for 405 responses

### LINKED ISSUE #4633: Return `405 Method Not Allowed` when a path exists but the HTTP method is unsupported

### Problem

### Expected Behavior

### Example

```ts
```

```http
405 Method Not Allowed
Allow: GET
```

instead of:

```http
404 Not Found
```

### Benefits

codewithkenzo:
Fixes #4633

Right now Hono returns 404 when you hit a route that exists but with the wrong HTTP method. That's not quite right per the HTTP spec - it should be 405 Method Not Allowed.

This PR adds a `methodNotAllowed()` handler (similar to how `notFound()` works) so you can customize the response, and includes the `Allow` header listing which methods actually work for that path.

## What changed

- Added `methodNotAllowed()` method to customize the 405 response
- Router now distinguishes "path doesn't exist" (404) from "path exists but wrong method" (405)
- Default response includes the `Allow` header with valid methods
- Works with basePath, nested routers via `app.route()`, etc.

## Usage

```ts
const app = new Hono()

// Optional - customize the response
app.methodNotAllowed((c, allowedMethods) => {
  return c.json({ 
    error: 'Method not allowed',
    allowed: allowedMethods 
  }, 405)
})

app.get('/users', (c) => c.text('List users'))
app.post('/users', (c) => c.text('Create user'))

// DELETE /users -> 405 with Allow: GET, POST
// GET /nonexistent -> 404 (unchanged)
```

Kept it simple and followed the existing patterns. Let me know if anything needs tweaking!

### DIFF
--- a/src/hono.test.ts
+++ b/src/hono.test.ts
@@ -1403,6 +1403,27 @@ describe('Method Not Allowed', () => {
     })
   })
 
+  describe('With nested routers (app.route)', () => {
+    const app = new Hono()
+    const api = new Hono()
+    api.get('/users', (c) => c.text('GET users'))
+    api.post('/users', (c) => c.text('POST users'))
+    app.route('/api', api)
+
+    it('Should return 405 for unsupported method on nested route', async () => {
+      const res = await app.request('http://localhost/api/users', { method: 'DELETE' })
+      expect(res.status).toBe(405)
+      const allow = res.headers.get('Allow')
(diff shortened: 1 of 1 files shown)


yusukebe:
We can't accept this PR. Please read the issue at https://github.com/honojs/hono/issues/4633 in full. Closing.

PR #4658, opened 2026-02-02
feat(trailing-slash): add `alwaysRedirect` option to support wildcard routes

### LINKED ISSUE #3407: `trimTrailingSlash` doesn't work on paths with `*`

4.6.1

Cloudflare Workers

Just make a Hono project and put this in `src/index.ts`

```ts
import { Hono } from "hono";
import { trimTrailingSlash } from "hono/trailing-slash";

const app = new Hono();

app.use(trimTrailingSlash());

app.get("/my-path/*", (c) => {
  return c.text("hi");
});

export default app;

```

Navigating to `http://localhost:8787/my-path/something/else/` should redirect to `http://localhost:8787/my-path/arst/arst`

### What do you see instead?

It stays the exact same :D

### Additional information

Manually putting the `trimTrailingSlash` code inside the function will work:

```ts
app.get("/my-path/*", (c) => {
  if (c.req.path !== "/" && c.req.path[c.req.path.length - 1] === "/") {
    const url = new URL(c.req.url);
    url.pathname = url.pathname.substring(0, url.pathname.length - 1);
    return c.redirect(url.toString(), 301);
  }

  return c.text("hi");
});
```

yusukebe:
Resolves #3407

Add ~`strict`~ `alwaysRedirect` option to `trimTrailingSlash` and `appendTrailingSlash` middleware.                 
                                                                                                   
When ~`strict: true`~ `alwaysRedirect:true`, the middleware redirects before executing handlers, which fixes the issue where trailing slash handling doesn't work with wildcard routes (`*`).                           
                                                                                                   
  ```ts                                                                                            
app.use(trimTrailingSlash({ alwaysRedirect: true }))
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/trailing-slash/index.ts
+++ b/src/middleware/trailing-slash/index.ts
@@ -13,7 +13,7 @@ type TrimTrailingSlashOptions = {
    * If `false` (default), it will only redirect when the route is not found (404).
    * @default false
    */
-  eager?: boolean
+  alwaysRedirect?: boolean
 }
 
 /**
@@ -34,16 +34,16 @@ type TrimTrailingSlashOptions = {
  *
  * @example
  * ```ts
- * // With eager option for wildcard routes
+ * // With alwaysRedirect option for wildcard routes
  * const app = new Hono()
  *
- * app.use(trimTrailingSlash({ eager: true }))
+ * app.use(trimTrailingSlash({ alwaysRedirect: true }))
  * app.get('/my-path/*', (c) => c.text('Wildcard route'))
  * ```
  */
 export const trimTrailingSlash = (options?: TrimTrailingSlashOptions): MiddlewareHandler => {
   return async function trimTrailingSlash(c, next) {
-    if (options?.eager) {
+    if (options?.alwaysRedirect) {
       if (
         (c.req.method === 'GET' || c.req.method === 'HEAD') &&
         c.req.path !== '/' &&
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by codler — APPROVED** [review]

File: src/middleware/trailing-slash/index.ts [hunk 1]
```diff
@@ -5,11 +5,22 @@
...
 import type { MiddlewareHandler } from '../../types'
 
+type TrimTrailingSlashOptions = {
+  /**
+   * If `true`, the middleware will always redirect requests with a trailing slash.
+   * This is useful for routes with wildcards (`*`).
+   * If `false` (default), it will only redirect when the route is not found (404).
+   * @default false
+   */
+  strict?: boolean
```
codler: Should we invert the logic? When I hear an option strict it is more limited, hence only applied to 404. So what I mean if we implement as 
```
  /**
   * If `true` (default), it will only redirect when the route is not found (404).
   * If `false`, the middleware will always redirect requests with a trailing slash.
   * This is useful for routes with wildcards (`*`).
   * @default true
   */

yusukebe: @codler Thanks.

I thought `strict` was confusing, since I saw your comment that we can interpret it in either sense. So I propose the `eager` instead of `strict` ( https://github.com/honojs/hono/pull/4658/changes/1b4d0b86cc36dda7a56e04930e1e898c0edea859). This obviously executes before the handler with `eager:true`.

What do you think of it?

codler: It is better. How about `alwaysRedirect?: boolean`?
```node
app.use(trimTrailingSlash({ alwaysRedirect: true }))

yusukebe: @codler 

`alwaysRedirect` is good! I replaced.

codler (approved):

PR #4659, opened 2026-02-02
fix(bearer-auth): make auth-scheme case-insensitive

bytaesu:

- [x] Add tests
- [x] Run tests

---

> [!NOTE]
> It uses a case-insensitive token as a means to identify the authentication scheme, followed by additional information necessary for achieving authentication via that scheme.
> https://datatracker.ietf.org/doc/html/rfc7235#section-2.1 

This PR updates the `bearer-auth` middleware to handle auth scheme comparison case-insensitively

### DIFF
--- a/src/middleware/bearer-auth/index.test.ts
+++ b/src/middleware/bearer-auth/index.test.ts
@@ -450,6 +450,18 @@ describe('Bearer Auth by Middleware', () => {
     expect(res.headers.get('x-custom')).toBe('foo')
   })
 
+  it.each([['bearer'], ['BEARER'], ['BeArEr']])(
+    'Should authorize - prefix is case-insensitive: %s',
+    async (prefix) => {
+      const req = new Request('http://localhost/auth/a')
+      req.headers.set('Authorization', `${prefix} ${token}`)
+      const res = await app.request(req)
+      expect(res).not.toBeNull()
+      expect(res.status).toBe(200)
+      expect(handlerExecuted).toBeTruthy()
+    }
+  )
+
   it('Should not authorize - no authorization header', async () => {
     const req = new Request('http://localhost/auth/a')
     const res = await app.request(req)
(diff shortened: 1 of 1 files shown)


src/middleware/bearer-auth/index.ts
```diff
@@ -112,8 +112,8 @@ export const bearerAuth = (options: BearerAuthOptions): MiddlewareHandler => {
   }
 
   const realm = options.realm?.replace(/"/g, '\\"')
-  const prefixRegexStr = options.prefix === '' ? '' : `${options.prefix} +`
-  const regexp = new RegExp(`^${prefixRegexStr}(${TOKEN_STRINGS}) *$`)
```
yusukebe, line 116: Simply, only adding `i` to the regexp is not okay?

```ts
const regexp = new RegExp(`^${prefixRegexStr}(${TOKEN_STRINGS}) *$`, 'i')
```

This means you can apply the patch to the main:

```diff
diff --git a/src/middleware/bearer-auth/index.ts b/src/middleware/bearer-auth/index.ts
index 3ecd2e3b..122ccc55 100644
--- a/src/middleware/bearer-auth/index.ts
+++ b/src/middleware/bearer-auth/index.ts
@@ -113,7 +113,7 @@ export const bearerAuth = (options: BearerAuthOptions): MiddlewareHandler => {

   const realm = options.realm?.replace(/"/g, '\\"')
   const prefixRegexStr = options.prefix === '' ? '' : `${options.prefix} +`
-  const regexp = new RegExp(`^${prefixRegexStr}(${TOKEN_STRINGS}) *$`)
+  const regexp = new RegExp(`^${prefixRegexStr}(${TOKEN_STRINGS}) *$`, 'i')
   const wwwAuthenticatePrefix = options.prefix === '' ? '' : `${options.prefix} `

   const throwHTTPException = async (
```

bytaesu: Hi @yusukebe,

Using the `i` flag would make TOKEN_STRINGS matching case-insensitive as well.

While validation is ultimately done via `equal = await options.verifyToken(match[1], c)` using the captured original value, this change was intended to make only the prefix case-insensitive while keeping the token itself case-sensitive.

Which approach do you prefer? 🧐

yusukebe: Hey! Thank you for the explanation!

Hmmm. This is a bit difficult problem. Honestly, both are okay, and I can fully understand what you intend. But the code in this PR is verbose. So how about using `i` flag, though, it is making `TOKEN_STRINGS` case-insensitive?

Just to confirm, the final behavior will not change, right?

bytaesu: Yes, since `verifyToken()` compares the captured original value, the end result is the same. I've reverted the change, updated it, and added tests for this case 🙂

yusukebe:
@bytaesu 

Looks good! Let's go with this. Thanks!

Issue #2343, opened 2024-03-12
c.json returning empty string for undefined values


yusukebe:
### Discussed in https://github.com/orgs/honojs/discussions/2338

<div type='discussions-op-text'>

<sup>Originally posted by **marcosrjjunior** March 11, 2024</sup>
I'm mainly creating this ticket to understand a bit more about why this is happening and potentially flag an issue.

One of my functions was returning `undefined` instead of `false` and the `c.json` is forcing the response to be an "" (empty string) instead of throwing an error.
![Screenshot 2024-03-11 at 8 51 27 am](https://github.com/honojs/hono/assets/5287262/4ca77143-ce93-4643-824b-b5bf6a188368)
**Response:** ""

---

Reading the [json implementation here](https://github.com/honojs/hono/blob/4ca5f60def726731b3afbf3f6137f0a4b92d8b27/src/context.ts#L469) (hopefully that is the correct place).

I noticed that there possibly some extra things going on there:
- `JSON.stringify(object)`. Why do we need to run throguh a JSON.stringify before returning?
- Re-creating the response

Because of the issues mentioned above I replaced it to use the main [response](https://developer.mozilla.org/en-US/docs/Web/API/Response/json_static) instead.
![Screenshot 2024-03-11 at 8 51 36 am](https://github.com/honojs/hono/assets/5287262/d4c41bba-3362-47b0-8b9d-3e96411ceb1a)</div>

damianpumar:
Hello @yusukebe 
I think we can not replace JSON.stringify to Response.json, because the JSON with `undefined`returns `undefined` and Response throws an exception.
So, my proposal is create a unit test for context.json function with undefined and fix this method.

What do you think ? 
Thanks! 

yusukebe:
Hi @damianpumar 

The interesting thing about this problem is that the result of `new Response(undefined)` is different at different runtimes. although it might be expected to throw an error. So I thought of using `Response.json()` to make `c.json()` behave according to each runtime.

But I'm still trying to figure out if that's the best option.

What do you think the result of `c.json(undefined)` should be?

marcosrjjunior:
On the scenario described above the `c.json` is treating the undefined and returning an empty string ("") which forces the users to treat the client that is consuming the endpoint or fix the function to return `false` instead. [(more examples on the discussion mentioned above)](https://github.com/orgs/honojs/discussions/2338).

do we really need **extra process JSON.stringify** if we know the response is already a JSON?

[...]

PR #4665, opened 2026-02-08
fix(url): strip fragment identifiers from path in getPath()

AhmadYasser1:
## Summary

Service Workers receive the full URL including fragment identifiers (e.g., `/users/#section`), unlike traditional servers where browsers strip fragments before sending requests. This caused Hono's route matcher to treat `#` as a regular character, leading to incorrect route parameter parsing (e.g., `id = "#user-list"` instead of a 404/correct match).

This implements the approach suggested by @usualoma in the issue discussion: update `getPath()` to treat `#` as a path terminator alongside `?`, both in the fast-path character loop and the percent-encoding fallback path.

## Changes

- In the fast-path loop: break on charCode `35` (`#`) in addition to `63` (`?`)
- In the percent-encoding fallback: check for `#` when `?` is not found
- Added test cases for fragment stripping with and without percent encoding

## Checklist

- [x] Add tests
- [x] Run tests
- [x] `bun run format:fix && bun run lint:fix` to format code
- [x] Add TSDoc/JSDoc to document code (no new public APIs added)

Closes #4440

PR #4682, opened 2026-02-08
fix(context): throw TypeError for non-serializable c.json() values

AhmadYasser1:
## Summary
- `JSON.stringify()` returns `undefined` for values like `undefined` or functions, which caused `c.json()` to silently return an empty body with a `200` status and `application/json` content type
- Now throws a `TypeError('Value is not JSON serializable')` so the error handler properly returns a `500` response

## Changes
- `src/context.ts`: Added check after `JSON.stringify(object)` — if result is `undefined`, throw `TypeError`
- `src/hono.test.ts`: Added test verifying `c.json(undefined)` triggers a 500 response

Closes #2343

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -703,8 +703,12 @@ export class Context<
     arg?: U | ResponseOrInit<U>,
     headers?: HeaderRecord
   ): JSONRespondReturn<T, U> => {
+    const body = JSON.stringify(object)
+    if (body === undefined) {
+      throw new TypeError('Value is not JSON serializable')
+    }
     return this.#newResponse(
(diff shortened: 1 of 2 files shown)


ota-meshi:
I'm not the maintainer, but I happened to come across it, so I'll leave a comment.
It looks like you already have a pull request open that makes the same changes to src/context.ts. #4668
Why not merge them into one pull request?

AhmadYasser1:
@ota-meshi 

Thanks for letting me know, Ota. I received a similar comment from Yusuke in [#4665](url)

I'll be consolidating all of my open PRs into 1.

PR #4703, opened 2026-02-10
[WIP] feat: introduce Route Check Middleware

### LINKED ISSUE #4701: Add configurable route existence check before authentication middleware execution

4.11.9

Node 20.19.4 (but not important)

1. Set up a Hono app with authentication middleware
2. Add some routes with authentication
3. Make requests to non-existent paths

```typescript
import { Hono } from 'hono'
import { jwt } from 'hono/jwt'

const app = new Hono()

// Authentication middleware
app.use('*', jwt({ secret: 'my-secret' }))

// Actual route
app.get('/protected', (c) => c.text('Protected content'))

// Request to non-existent path still triggers JWT validation
// GET /non-existent-route
```

When accessing non-existent routes:

1. The server should first check if the route exists
2. If route doesn't exist, return 404 immediately
3. Authentication middleware should only run for valid routes

### What do you see instead?

1. Authentication middleware runs for ALL requests, even to non-existent routes
2. This causes unnecessary CPU-intensive authentication processing
3. Only after authentication does the framework return 404 for invalid routes

### Additional information

**Security Impact:**

- Creates a potential DoS vulnerability where attackers can spam random URLs to trigger expensive auth checks

- Particularly problematic for JWT/OAuth/other crypto-heavy authentication

**Performance Impact:**

- Wastes server resources validating authentication for invalid routes

- Can significantly impact performance under high traffic

**Suggested Solutions:**

1.  Add global config option:

```typescript
new Hono({ checkRouteBeforeAuth: true })
```

2. Provide route validation middleware:

```typescript
app.use(routeValidator()) // Checks route existence
app.use(authMiddleware)  // Then authenticates
```

3. Expose routing inspection API for custom solutions

yusukebe:
Resolve #4701

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/jsr.json
+++ b/jsr.json
@@ -50,6 +50,7 @@
     "./pretty-json": "./src/middleware/pretty-json/index.ts",
     "./request-id": "./src/middleware/request-id/request-id.ts",
     "./language": "./src/middleware/language/language.ts",
+    "./route-check": "./src/middleware/route-check/index.ts",
     "./secure-headers": "./src/middleware/secure-headers/secure-headers.ts",
     "./combine": "./src/middleware/combine/index.ts",
     "./ssg": "./src/helper/ssg/index.ts",
--- a/package.json
+++ b/package.json
@@ -255,6 +255,11 @@
       "import": "./dist/middleware/language/index.js",
       "require": "./dist/cjs/middleware/language/index.js"
     },
+    "./route-check": {
+      "types": "./dist/types/middleware/route-check/index.d.ts",
+      "import": "./dist/middleware/route-check/index.js",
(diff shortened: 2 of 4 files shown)


BelovedYaoo:
Hi @yusukebe, thanks for working on this!

I've opened #4714 as a follow-up optimization for this middleware. It introduces a WeakMap cache to store the result of the isMiddleware and findTargetHandler checks.

Since handler identities are immutable at runtime, this avoids redundant recursive lookups for every request, optimizing the performance from O(depth) to O(1) while remaining safe for dynamic routes.

Feel free to check it out :)

PR #4707, opened 2026-02-10
perf(context): add fast path to c.json() matching c.text() optimization

mgcrea:
`c.text()` already has a fast path (added in an earlier optimization) that returns `new Response(text)` directly when no status, headers, or finalized state exists. `c.json()` was missing this optimization and always went through `#newResponse()`, which unconditionally allocates a `Headers` object and iterates header entries — even for the simplest `c.json({ ... })` call.

This PR adds the same 5-condition fast path to `c.json()`.

## What changed

In `src/context.ts`, the `json()` method now checks:

```typescript
!this.#preparedHeaders && !this.#status && !arg && !headers && !this.finalized
```

When all conditions are met (the common case for simple JSON responses), it creates the Response directly with an inline header object instead of going through `#newResponse()`:

```typescript
new Response(jsonString, {
})
```

## Why this matters

In the `@hono/node-server` adapter, this has a compounding effect. The lightweight `Response` constructor stores inline header objects directly in its cache. When the adapter writes the response, it checks `header instanceof Headers` — a plain object skips the `buildOutgoingHttpHeaders()` conversion entirely, while a `Headers` instance requires iteration and conversion to `OutgoingHttpHeaders`.

### Benchmark results (nodejs)

> with corresponding `@hono/node-server` optimizations, see https://github.com/honojs/node-server/pull/301

| Benchmark | Before | After  | Change     |
| --------- | ------ | ------ | ---------- |
| Ping      | 55,183 | 59,697 | **+8.2%**  |
| Query     | 46,033 | 52,883 | **+14.9%** |
| Body      | 26,439 | 46,726 | **+76.7%** |
| Average   | 42,552 | 53,102 | **+24.8%** |

*Body benchmark improvement is primarily from `@hono/node-server` direct body reading; this PR contributes to the overall gains by eliminating allocations on the response path.*

Benchmarked with [bun-http-framework-benchmark](https://github.com/SaltyAom/bun-http-framework-benchmark) using `bombardier` at 500 concurrent connections for 10s per test on Node.js.

---

Thanks for the great lib :)

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -657,6 +657,10 @@ export class Context<
     headers?: HeaderRecord
   ): ReturnType<BodyRespond> => this.#newResponse(data, arg, headers) as ReturnType<BodyRespond>
 
+  #useFastPath(): boolean {
+    return !this.#preparedHeaders && !this.#status && !this.finalized
+  }
+
   /**
    * `.text()` can render text as `Content-Type:text/plain`.
    *
@@ -674,7 +678,7 @@ export class Context<
     arg?: ContentfulStatusCode | ResponseOrInit,
     headers?: HeaderRecord
   ): ReturnType<TextRespond> => {
-    return !this.#preparedHeaders && !this.#status && !arg && !headers && !this.finalized
+    return this.#useFastPath() && !arg && !headers
       ? (new Response(text) as ReturnType<TextRespond>)
       : (this.#newResponse(
           text,
@@ -704,7 +708,7 @@ export class Context<
     headers?: HeaderRecord
   ): JSONRespondReturn<T, U> => {
     return (
-      !this.#preparedHeaders && !this.#status && !arg && !headers && !this.finalized
+      this.#useFastPath() && !arg && !headers
         ? Response.json(object)
         : this.#newResponse(
             JSON.stringify(object),


src/context.ts
```diff
@@ -703,10 +703,13 @@ export class Context<
...
     headers?: HeaderRecord
   ): JSONRespondReturn<T, U> => {
-    return this.#newResponse(
-      JSON.stringify(object),
-      arg,
-      setDefaultContentType('application/json', headers)
+    const jsonString = JSON.stringify(object)
+    return (
+      !this.#preparedHeaders && !this.#status && !arg && !headers && !this.finalized
+        ? new Response(jsonString, {
```
yusukebe: I am wondering whether we should merge this PR, but regarding this, I think using `Response.json()` is better. It's simpler and faster: https://x.com/jarredsumner/status/2023328556210921948?s=20

```diff
diff --git a/src/context.ts b/src/context.ts
index 6b7949b9..f2a0edad 100644
--- a/src/context.ts
+++ b/src/context.ts
@@ -703,13 +703,14 @@ export class Context<
     arg?: U | ResponseOrInit<U>,
     headers?: HeaderRecord
   ): JSONRespondReturn<T, U> => {
-    const jsonString = JSON.stringify(object)
     return (
       !this.#preparedHeaders && !this.#status && !arg && !headers && !this.finalized
-        ? new Response(jsonString, {
-            headers: { 'Content-Type': 'application/json' },
-          })
-        : this.#newResponse(jsonString, arg, setDefaultContentType('application/json', headers))
+        ? Response.json(object)
+        : this.#newResponse(
+            JSON.stringify(object),
+            arg,
+            setDefaultContentType('application/json', headers)
+          )
     ) /* eslint-disable @typescript-eslint/no-explicit-any */ as any
   }
```

src/context.ts
```diff
@@ -703,10 +703,13 @@ export class Context<
     arg?: U | ResponseOrInit<U>,
     headers?: HeaderRecord
   ): JSONRespondReturn<T, U> => {
-    return this.#newResponse(
-      JSON.stringify(object),
-      arg,
-      setDefaultContentType('application/json', headers)
+    const jsonString = JSON.stringify(object)
+    return (
+      !this.#preparedHeaders && !this.#status && !arg && !headers && !this.finalized
```
yusukebe: Using a common function reduces the file size slightly. I think the overhead can be ignored.

```diff
diff --git a/src/context.ts b/src/context.ts
index 6b7949b9..eb43dcf4 100644
--- a/src/context.ts
+++ b/src/context.ts
@@ -657,6 +657,10 @@ export class Context<
     headers?: HeaderRecord
   ): ReturnType<BodyRespond> => this.#newResponse(data, arg, headers) as ReturnType<BodyRespond>

+  #shouldFastPath(): boolean {
+    return !this.#preparedHeaders && !this.#status && !this.finalized
+  }
+
   /**
    * `.text()` can render text as `Content-Type:text/plain`.
    *
@@ -674,7 +678,7 @@ export class Context<
     arg?: ContentfulStatusCode | ResponseOrInit,
     headers?: HeaderRecord
   ): ReturnType<TextRespond> => {
-    return !this.#preparedHeaders && !this.#status && !arg && !headers && !this.finalized
+    return this.#shouldFastPath() && !arg && !headers
       ? (new Response(text) as ReturnType<TextRespond>)
       : (this.#newResponse(
           text,
@@ -705,7 +709,7 @@ export class Context<
   ): JSONRespondReturn<T, U> => {
     const jsonString = JSON.stringify(object)
     return (
-      !this.#preparedHeaders && !this.#status && !arg && !headers && !this.finalized
+      this.#shouldFastPath() && !arg && !headers
         ? new Response(jsonString, {
             headers: { 'Content-Type': 'application/json' },
           })
```

yusukebe:
@mgcrea 

This PR is super cool. ~Especially if adding this fast path, we can use `Response.json()` as commented in https://github.com/honojs/hono/pull/4707#discussion_r2811843439. If so, it can reduce the bundle size and get it faster.~

Sorry! I completely misunderstood. The file size is slightly increased. But it's interesting that we can use `Response.json()` (https://github.com/honojs/hono/pull/4707#discussion_r2811843439) with this PR.

I would like to merge this. Can you handle my comments? Or, I'll fix them myself.

PR #4717, opened 2026-02-11
feat(language): add progressive locale code truncation to normalizeLanguage

### LINKED ISSUE #4294: LanguageDetector middleware does not parse locale code into language code

4.8.2

Cloudflare Worker

When an PWA app wrapped with TWA (bubblewrap) installed as APK on Android where system language is japanese visits my website, it appears that the accepter-language header may only contains a locale code.

Cloudflare Worker's log:
```
headers: {
  accept: "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.7",
  accept-encoding: "gzip, br",
  accept-language: "ja-JP",
  ...
```

I'm using hono's languageDetector in the following setting, and it seems hono does not match "ja" with "ja-JP", so the website shows english page in this case.
```ts
languageDetector({
  supportedLanguages: ["en", "ja"],
  fallbackLanguage: "en",
  order: ["cookie", "header"],
  lookupCookie: "language",
  cookieOptions: {
    sameSite: "Lax",
    secure: process.env.API_ENV !== "development",
    maxAge: 60 * 60 * 24 * 365,
    httpOnly: false,
  },
});
```

hono's languageDetector should detect the language as "ja" when supportedLanguages is "ja" and accept-language is "ja-JP".

### What do you see instead?

hono's languageDetector treats "ja" and "ja-JP" as different language, and falls back to "en".

### Additional information

_No response_

sorafujitani:
Closes #4294

## Summary

- Add RFC 4647 Lookup-based progressive truncation to `normalizeLanguage` so that locale codes like `ja-JP` match `ja` when only the base language is in `supportedLanguages`
- Exact match is always checked first; truncation is only attempted when no exact match is found

## Verification

```ts
// verify.ts
import { languageDetector } from './src/middleware/language'

const app = new Hono()
app.use('/*', languageDetector({
  supportedLanguages: ['en', 'ja'],
  fallbackLanguage: 'en',
  order: ['cookie', 'header'],
}))
app.get('/*', (c) => c.text(c.get('language')))

export default app
```

```sh
# Issue case: ja-JP → ja (was 'en' before this fix)
bunx @hono/cli request verify.ts -H "accept-language: ja-JP"
# => ja

# Exact match still works
bunx @hono/cli request verify.ts -H "accept-language: ja"
# => ja

# Unmatched locale falls back correctly
bunx @hono/cli request verify.ts -H "accept-language: ko-KR"
# => en
```

### Why progressive truncation, not leftmost-only comparison

Comparing only the primary (leftmost) subtag would cause incorrect matches for script-differentiated languages. For example, `zh-Hans` (Simplified Chinese) and `zh-Hant` (Traditional Chinese) share the primary subtag `zh` but are mutually unintelligible in written form. Progressive truncation per [RFC 4647 Section 3.4](https://www.rfc-editor.org/rfc/rfc4647.html#section-3.4) correctly tries `zh-Hant-CN` → `zh-Hant` → `zh` in order, so `zh-Hant` is matched before falling back to `zh`.

### No guessing of invalid language codes

This change only applies truncation to subtags separated by `-`. Tags without subtags (e.g., `jp`, `de`) are unaffected — if they don't exactly match a supported language, the fallback is returned as before. This PR does not attempt to guess or correct invalid language codes (e.g., `jp` → `ja`). Such mappings can be handled via the existing `convertDetectedLanguage` option.

### Loop style: `for` with early return

The same file's `detectLanguage` function (L244) uses `for...of` with `break` for early-exit iteration. The truncation loop follows this convention, using a `for` loop with early `return`. The only `let` is the loop counter `i`; all other variables are `const`.

### Singleton subtag skipping (not implemented)

RFC 4647 Section 3.4 specifies that single-character subtags (`x` for private-use, `u` for Unicode extensions, etc.) should be removed together with their closest trailing subtag during truncation:

```
zh-Hant-CN-x-private1-private2
zh-Hant-CN-x-private1
zh-Hant-CN                      ← skips 'zh-Hant-CN-x' (singleton 'x')
zh-Hant
zh
```

This PR does **not** implement singleton skipping. In practice, singleton-ending tags like `en-x` or `de-DE-u` never appear in `supportedLanguages`, so they would simply be skipped as non-matching candidates without affecting the final result. Full singleton handling can be added in a follow-up if needed.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/language/language.ts
+++ b/src/middleware/language/language.ts
@@ -107,10 +107,9 @@ export const normalizeLanguage = (
     }
 
     // Progressive truncation (RFC 4647 Lookup)
-    let candidate = compLang
-    let dashIndex: number
-    while ((dashIndex = candidate.lastIndexOf('-')) > 0) {
-      candidate = candidate.substring(0, dashIndex)
+    const parts = compLang.split('-')
+    for (let i = parts.length - 1; i > 0; i--) {
+      const candidate = parts.slice(0, i).join('-')
       const prefixIndex = compSupported.indexOf(candidate)
       if (prefixIndex !== -1) {
         return options.supportedLanguages[prefixIndex]


sorafujitani:
@yusukebe 
As mentioned in the description, we have adopted a policy to use `for let` instead of `forEach` from the perspective of early loop termination.

> Loop style: for instead of forEach
The codebase commonly uses forEach for iteration, but the truncation loop requires early return when a match is found. forEach cannot break out of the enclosing function, so a for loop is used instead. With this approach the only let is the loop counter i, and all other variables are const.

yusukebe:
Hey @sorafujitani !

The code is readable, and I think it's good.  Let's go with this.

I'll include this change in the next minor release, v4.12.0. Please wait a bit to merge. Thanks!

PR #4719, opened 2026-02-11
feat(types): Add exports field to ExecutionContext

### LINKED ISSUE #4493: ExecutionContext is missing exports filed available in Cloudflare Workers

4.10.3

Cloudflare Workers

Creating a Cloudflare Workers app with `enable_ctx_exports` compatibility flag does not gives access to new property `exports` that should list all top-level exports like DurableObjects. Please check documentation: https://developers.cloudflare.com/changelog/2025-09-26-ctx-exports/

https://github.com/honojs/hono/blob/main/src/context.ts should not declare it's own `ExecutionContext` interface but use the one provided by `worker-configuration.d.ts` generated by wrangler. It could be achieved for example by passing ExecutionContext the same way as Binding and Variables:

```ts
const routes = new Hono<{ Bindings: CloudflareBindings; Variables: HonoAuthorizationContext, ExecutionContext: ExecutionContext }>();
```

### What do you see instead?

_No response_

### Additional information

This is the functionality available behind compatibility flag, but it looks like it will be GA anytime. Also enabling devs to pass their own type is always future-proof

toreis-up:
## Summary
Add `exports` property to `ExecutionContext` for Cloudflare Workers

## Related
- Fixes #4493 
- https://developers.cloudflare.com/workers/runtime-apis/context/#exports

## Note
To Autocomplete with Wrangler's generated types, we should declare the type like:
```ts
import 'hono'

declare module 'hono' {
  interface ExecutionContext {
    readonly exports: Cloudflare.Exports
  }
}
```

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -44,6 +44,11 @@ export interface ExecutionContext {
    */
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   props: any
+  /**
+   * For compatibility with Wrangler 4.x.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  exports?: any
 }
 
 /**


src/types.ts
```diff
@@ -31,6 +31,7 @@ export type BlankEnv = {}
 export type Env = {
   Bindings?: Bindings
   Variables?: Variables
+  ExecutionContext?: unknown
```
yusukebe: I think passing a generic to define the ExecutionContext type is overkill. Basically, ExecutionContext is used only for Cloudflare Workers. We don’t have to make it platform-specific in the core.

Do you have any ideas for making it simpler?

toreis-up: @yusukebe san

My apologies for the delayed response.

You're right, this would be a runtime-specific (Cloudflare Workers) change. I didn't consider that enough.
**I'll just add an `exports` property.**

However, if the property is `any`, we can't override use `declare module` to declare the type, IntelliSense won't provide autocompletion with Wrangler's generated types.
I think this can be resolved by casting like following:
`(c.executionCtx as globalThis.ExecutionContext).exports;`

yusukebe: @toreis-up 

No problem!

As you said, not having IntelliSense working is not good. I'll consider it a little.

yusukebe: @toreis-up 

Even if the `exports` do not have types, you can specify them within `declare module` like this:

```ts
declare module 'hono' {
  interface ExecutionContext {
    readonly exports: Cloudflare.Exports
  }
}
```

Adding `ExecutionContext` to Env isn't a bad idea, but as I said, it's a little Cloudflare-specific. Additionally, the big problem is that it increases the type definition. Honestly, I don't want to make it more complex at the current major version.

So, I think we should make it `any` and encourage the use of `declare module`.

What do you think?

toreis-up: @yusukebe san

That would be nice. I agree with it.

> However, if the property is any, we can't override use declare module to declare the type, "

In my case, it turned out to be an environment issue. The code is running correctly now.
Sorry for the confusion.

yusukebe:
@toreis-up 

In the end, we use `any`, but it's good. Thank you for your contribution.

I'll include this change in the next minor version, v4.12.0. Will be released soon!

PR #4722, opened 2026-02-13
test(types): add regression tests for #4388

### LINKED ISSUE #4388: Type inference drops routes declared before use() when using method chaining

4.6.5

Cloudflare Workers (issue observed in local type inference only)

When defining endpoints with method chaining, if a "use()" method is inserted, the routes declared before that "use()" are not included in the inferred type information.

```ts
import { Hono } from "hono";

const routerA = new Hono()
  .get("/", (c) => c.text("Hello from Router A!"))
  .use("/:slug", async (c, next) => {
    return next();
  })
  .post("/:slug", (c) => c.text("Hello from Router A with slug!"));

const app = new Hono().route("/a", routerA);

export type AppType = typeof app;
export default app;
```

Chained routes defined before "use()" should still be inferred in the types.

### What do you see instead?

_No response_

### Additional information

_No response_

TyposBro:
## Summary
- Adds regression type tests for #4388 (routes declared with explicit paths before `.use()` being dropped in TypeScript inference)
- Covers the exact scenarios from the issue: `.get('/')` before `.use('/path')`, full chains with `.post()`, `.route()` wrapping, and multiple explicit-path routes
- All 4 tests pass on current `main`, confirming the fix from the `CurrentPath` parameter approach is working

## Test plan
- [x] `bun run test src/types.test.ts` — 108 tests pass (104 existing + 4 new)
- [x] `npx tsc --noEmit` — no type errors

Closes #4388

### DIFF
--- a/src/types.test.ts
+++ b/src/types.test.ts
@@ -3682,3 +3682,110 @@ describe('Handlers returning Promise<void>', () => {
     type verify = Expect<Equal<Expected, Actual>>
   })
 })
+
+// Regression tests for #4388: routes before .use() with explicit paths should not be dropped
+describe('Routes before .use() with explicit paths (#4388)', () => {
+  it('should preserve explicit-path .get() before .use() with path', () => {
+    const app = new Hono()
+      .get('/', (c) => c.text('Hello from /'))
+      .use('/noop', async (c, next) => {
+        await next()
+      })
+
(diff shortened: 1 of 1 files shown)


yusukebe:
@TyposBro 

Nice! These tests are meaningful. Looks good!

PR #4723, opened 2026-02-16
fix(request): return `string | undefined` from param() when path type is any

### LINKED ISSUE #3198: With `UpgradeWebSocket`, any path parameters become non-optional in TS type system

4.5.1

Node.js

```typescript
import { Hono } from "hono"
import { UpgradeWebSocket } from "hono/ws"

declare var upgradeWebSocket: UpgradeWebSocket // actually I'm using `@hono/node-ws` one, but `@hono/node-ws` exposing UpgradeWebSocket as-is

const app = new Hono()

app.get("/api/rooms/:roomId/info", async c => {
    const roomId: string = c.req.param("roomId")
    // @ts-expect-error
    const yetAnotherParam: string = c.req.param("yetAnotherParam")

    return c.json({"name": "stub"})
})

app.get("/api/rooms/:roomId/ws", upgradeWebSocket(c => {
    const roomId: string = c.req.param("roomId")
    // @ts-expect-error
    const yetAnotherParam: string = c.req.param("yetAnotherParam")

    return {}
}))
```

In the example code, `c.req.param("yetAnotherParam")` should return `string | undefined` in both handlers (all of `// @ts-expect-error` should be consumed).

### What do you see instead?

In the example code, `c.req.param("yetAnotherParam")` returns a `string` in the handler using `UpgradeWebSocket`.

### Additional information

_No response_

andrewdamelio:
When middleware uses Context with a default path type parameter (any), c.req.param('key') incorrectly returns string instead of string | undefined. This is because ParamKeys<any> resolves to string, matching the first overload that returns string.

Add a new overload that detects when P is any using the 0 extends (1 & P) trick and returns string | undefined, ensuring type safety in middleware.

Fixes #3198

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/types.test.ts
+++ b/src/types.test.ts
@@ -523,7 +523,7 @@ describe('Test types of Handler', () => {
       const foo = c.get('foo')
       expectTypeOf(foo).toEqualTypeOf<number>()
       const id = c.req.param('id')
-      expectTypeOf(id).toEqualTypeOf<string>()
+      expectTypeOf(id).toEqualTypeOf<string | undefined>()
       return c.text('Hi')
     }
     app.get('/', handler)
--- a/src/request.ts
+++ b/src/request.ts
@@ -91,8 +91,9 @@ export class HonoRequest<P extends string = '/', I extends Input['out'] = {}> {
    * const { id, comment_id } = c.req.param()
    * ```
    */
-  param(key: 0 extends 1 & P ? string : never): string | undefined
-  param<P2 extends ParamKeys<P> = ParamKeys<P>>(key: P2 extends `${infer _}?` ? never : P2): string
+  param<P2 extends ParamKeys<P> = ParamKeys<P>>(
+    key: string extends P ? never : P2 extends `${infer _}?` ? never : P2
+  ): string
   param<P2 extends RemoveQuestion<ParamKeys<P>> = RemoveQuestion<ParamKeys<P>>>(
     key: P2
   ): string | undefined


yusukebe:
Hi @andrewdamelio 

Is this really going to fix the issue written in https://github.com/honojs/hono/issues/3198#issue-2433757000?

Can you write a test and confirm this PR fixes it?

andrewdamelio:
> Hi @andrewdamelio
> 
> Is this really going to fix the issue written in [#3198 (comment)](https://github.com/honojs/hono/issues/3198#issue-2433757000)?
> 
> Can you write a test and confirm this PR fixes it?


Added two type-level tests to src/request.test.ts in the Param > Type describe block:

1. param() returns string | undefined when P is any — creates a HonoRequest<any> (simulating middleware context) and asserts via expectTypeOf
2. param() returns string when P is a known route string — confirms existing typed-route behaviour is unaffected

emmyasuai:
## PR Brief (Manual Bootstrap)

**TL;DR**
- PR #4723 in `honojs/hono` has a generated bootstrap summary.
- Generated from head SHA `c08143a4f01e0d593f836c0d6bdee10e3346faf6`.

**Evidence Notes**
- Evidence not provided for runtime behavior or CI outcomes.
- Summary is grounded in provided metadata only.

**Reviewer Checklist**
- Validate behavior by running relevant tests.
- Confirm impacted files and risk-sensitive paths manually.

Generated from: `c08143a4f01e0d593f836c0d6bdee10e3346faf6`

**Render Artifacts**
- Render status: `disabled`
- Video artifact: not available
- Poster artifact: not available

src/types.test.ts
```diff
@@ -523,7 +523,7 @@ describe('Test types of Handler', () => {
       const foo = c.get('foo')
       expectTypeOf(foo).toEqualTypeOf<number>()
       const id = c.req.param('id')
-      expectTypeOf(id).toEqualTypeOf<string>()
+      expectTypeOf(id).toEqualTypeOf<string | undefined>()
```
yusukebe, line 526: There may be some discussion about updating the test, but if we accept this PR, the old test is wrong. The value of `c.req.param('id')` can be `string | undefined`.

yusukebe (approved):
LGTM!

yusukebe:
Hi @andrewdamelio 

I updated the type definition and the existing test. I think the behavior is good. I'll merge this. Thank you for your help.

moroshko:
@andrewdamelio Shouldn't the `roomId` in this example be of type `string`? 
It's `string | undefined` in `hono` version `4.12.8`.

```ts
import { Hono } from 'hono'
import { upgradeWebSocket } from "hono/bun";

const app = new Hono()

app.get('/rooms/:roomId/ws', upgradeWebSocket((c) => {
  const roomId = c.req.param("roomId");

  return {
    onOpen() {
      console.log('WebSocket opened')
    },
  }
}));
```

Issue #4711, opened 2026-02-10
Bug: JSON.parse without error handling in request body parsing


RoyRoki:
## 🐛 Bug Report

**Severity:** High  
**Category:** Error Handling / DoS  
**Location:** `src/request.ts`

**Found by:** [WhiteRose](https://github.com/shakecodeslikecray/whiterose) - AI-powered bug hunter  
**Bug ID:** WR-014

---

## 📋 Description

`JSON.parse()` is called without try-catch blocks in request body parsing. Malformed JSON causes uncaught exceptions and crashes the application.

---

## 🔍 Vulnerable Code

```typescript
// src/request.ts
const body = await this.raw.text()
const parsed = JSON.parse(body)  // ❌ No error handling!
```

---

## 💥 Impact

- **Application crashes:** Malformed JSON = unhandled exception
- **DoS attacks:** Attacker sends invalid JSON to crash server
- **Poor error messages:** Users don't get helpful validation errors

---

## 🧪 Attack Example

```bash
curl -X POST http://api.example.com/data \
  -H "Content-Type: application/json" \
  -d "{invalid json here}"
```

**Result:** Server crashes with `SyntaxError: Unexpected token`

---

## ✅ Suggested Fix

Wrap `JSON.parse()` in try-catch:

```typescript
async json<T = unknown>(): Promise<T> {
  const body = await this.raw.text()
  
  try {
    return JSON.parse(body)
  } catch (e) {
    throw new HTTPException(400, {
      message: 'Invalid JSON in request body',
      cause: e,
    })
  }
}
```

---

## 🔗 References

- **WhiteRose:** https://github.com/shakecodeslikecray/whiterose
- **CWE-754:** Improper Check for Unusual or Exceptional Conditions

---

_Found automatically by WhiteRose AI bug hunter - helping make open source more secure 🛡️_

yusukebe:
This is not a bug.

The result of `c.req.json()` for an invalid input is the same as that of `c.req.raw.json()` (returning 500 error). The behavior should be the same between them to keep compatibility with the Web Standard. And if we apply your suggestion, it will introduce a breaking change. I don't know, it's a good idea.

Closing this issue.

PS.

Don't submit the issue or PR via the automated AI. If you do more, I'll block you.

PR #4726, opened 2026-02-17
fix(request): Return 400 on invalid JSON request body issue #4711

henrygoeszanin:
Handle JSON parse errors in HonoRequest.json(): the method now awaits the cached text, wraps JSON.parse in a try/catch, and throws an HTTPException(400) with message 'Invalid JSON in request body' and the original error as the cause. This prevents uncaught parse exceptions and returns a clear client error while preserving the cached-body behavior and generic typing.

- [ ] Add tests
- [X] Run tests

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -249,7 +249,15 @@ export class HonoRequest<P extends string = '/', I extends Input['out'] = {}> {
    * ```
    */
   json<T = any>(): Promise<T> {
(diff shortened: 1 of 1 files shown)


yusukebe:
This PR is unnecessary. I wrote the reason here: https://github.com/honojs/hono/issues/4711

Closing. Thanks.

PR #4729, opened 2026-02-18
fix(jsx/dom): handle empty arrays in render children loop

### LINKED ISSUE #4727: render() throws TypeError given JSX with that generates an empty array

4.11.4 and 4.11.9

Deno 2.3.2

```tsx
import { render } from "hono/jsx/dom";

function App() {
  const tags: string[] = [];
  const terms: string[] = ["hello"];
  return (
    <div>
      {tags.map((x) => <span>{x}</span>)}
      {terms.map((x) => <span>{x}</span>)}
      <input type="text" />
    </div>
  );
}

render(<App />, document.getElementById("root")!);
// TypeError: Cannot read properties of undefined (reading 'children')
```

Empty array should render as nothing and the second array should be rendered.

### What do you see instead?

```
TypeError: Cannot read properties of undefined (reading 'children')
    at build (render.ts:477)
    typeof node.tag == 'function' ? invokeTag(context, node) : toArray(node.props.children)
```

### Additional information

| Pattern | Result |
|---------|--------|
| `{[].map(...)}` `<input/>` | ✅ works |
| `{["a"].map(...)}` `{[].map(...)}` | ✅ works |
| `{[].map(...)}` `{[].map(...)}` | ❌ crashes |
| `{[].map(...)}` `{["a"].map(...)}` | ❌ crashes |

The error happens in the `build()` function's child-flattening loop in `src/jsx/dom/render.ts`:

```typescript
for (let i = 0; i < children.length; i++) {
  if (Array.isArray(children[i])) {
    children.splice(i, 1, ...(children[i] as Child[]).flat())
  }
  let child = buildNode(children[i])
```

**Suggested fix:**

Add `i--` after splicing so the loop re-examines the same index:

```typescript
for (let i = 0; i < children.length; i++) {
  if (Array.isArray(children[i])) {
    children.splice(i, 1, ...(children[i] as Child[]).flat())
+   if (i >= children.length) break  // empty array removed, nothing replaced
+   i--  // re-examine this index
+   continue
  }
  let child = buildNode(children[i])
```

mixelburg:
Fixes #4727

When rendering JSX that includes an empty array (e.g. `{[].map(...)}`) followed by a non-empty array, the `build()` function crashes with `TypeError: Cannot read properties of undefined (reading 'children')`.

**Root cause:** After `children.splice(i, 1, ...(children[i]).flat())` flattens an empty array, the element at index `i` is removed and nothing replaces it. The loop still increments `i`, skipping the next child. When the skipped index is out of bounds, `buildNode()` receives `undefined` and crashes.

**Fix:** Add `i--; continue` after the splice so the loop re-examines the same index. This correctly handles:
- Empty arrays (removed, loop continues at same position)
- Non-empty arrays (flattened, loop re-processes first flattened element)

Added a test case covering the exact reproduction from the issue.

### DIFF
--- a/src/jsx/dom/index.test.tsx
+++ b/src/jsx/dom/index.test.tsx
@@ -124,8 +124,12 @@ describe('DOM', () => {
     const terms: string[] = ['hello']
     const App = () => (
       <div>
-        {tags.map((x) => <span>{x}</span>)}
-        {terms.map((x) => <span>{x}</span>)}
+        {tags.map((x) => (
+          <span>{x}</span>
+        ))}
+        {terms.map((x) => (
+          <span>{x}</span>
+        ))}
         <input type='text' />
       </div>
     )


usualoma:
Hi @mixelburg 
Thank you! LGTM!

yusukebe (approved):
LGTM!

yusukebe:
@mixelburg Thanks!

PR #4730, opened 2026-02-18
perf(jsx/dom): flatten children once at the start to avoid repeated flattening

usualoma:
This is a pull request to add to #4729, optimizing performance and fixing the location where tests are added.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/jsx/dom/index.test.tsx
+++ b/src/jsx/dom/index.test.tsx
@@ -119,24 +119,6 @@ describe('DOM', () => {
     expect(root.innerHTML).toBe('Hello')
   })
 
-  it('render with empty array followed by non-empty array', () => {
-    const tags: string[] = []
-    const terms: string[] = ['hello']
-    const App = () => (
-      <div>
-        {tags.map((x) => (
-          <span>{x}</span>
-        ))}
-        {terms.map((x) => (
-          <span>{x}</span>
-        ))}
-        <input type='text' />
-      </div>
-    )
-    render(<App />, root)
-    expect(root.innerHTML).toBe('<div><span>hello</span><input type="text"></div>')
-  })
-
   describe('performance', () => {
     it('should be O(N) for each additional element', () => {
       const App = () => (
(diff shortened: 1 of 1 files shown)


usualoma:
@yusukebe
Would you please review this?

src/jsx/dom/render.ts
```diff
@@ -486,7 +486,7 @@ export const build = (context: Context, node: NodeObject, children?: Child[]): v
     let prevNode: Node | undefined
     for (let i = 0; i < children.length; i++) {
       if (Array.isArray(children[i])) {
-        children.splice(i, 1, ...(children[i] as Child[]).flat())
+        children.splice(i, 1, ...((children[i] as unknown[]).flat(Infinity) as Child[]))
```
yusukebe, line 489: If you use `flat(Infinity)`, perhaps you can write it more simply?

```diff
diff --git a/src/jsx/dom/render.ts b/src/jsx/dom/render.ts
index 5d2eae62..985d9f98 100644
--- a/src/jsx/dom/render.ts
+++ b/src/jsx/dom/render.ts
@@ -484,12 +484,8 @@ export const build = (context: Context, node: NodeObject, children?: Child[]): v
         : undefined
     const vChildren: Node[] = []
     let prevNode: Node | undefined
+    children = (children as unknown[]).flat(Infinity) as Child[]
     for (let i = 0; i < children.length; i++) {
-      if (Array.isArray(children[i])) {
-        children.splice(i, 1, ...((children[i] as unknown[]).flat(Infinity) as Child[]))
-        i--
-        continue
-      }
       let child = buildNode(children[i])
       if (child) {
         if (
```

usualoma: Thanks for the review!

The result will still be correct. However, doing that means a copy always occurs during the `.flat()` call, even when children doesn't contain `Array.isArray(c)`. Since I want to avoid copying when it doesn't contain it, I'd like to keep the current implementation as-is for that part.

yusukebe: @usualoma I see. Thanks!

yusukebe:
@usualoma Thank you! Merging.

PR #4732, opened 2026-02-19
fix(client): skip undefined values in form data serialization

### LINKED ISSUE #4731: Inconsistent undefined handling in RPC client: query params skip undefined, form data serializes as "undefined" string

4.11.10

Cloudflare worker

When undefined is passed as a value in the form option of a Hono RPC client request, it is serialized as the literal string "undefined" via
FormData.append(). In contrast, undefined values in query are correctly skipped.

This is because buildSearchParams in dist/client/utils.js explicitly filters out undefined:

```js
if (v === void 0) { continue; }
```

But the form data loop in dist/client/client.js does not:

```js
for (const [k, v] of Object.entries(args.form)) {
  form.append(k, v); // undefined becomes "undefined"
}
```

Server:
```js
app.post("/test", async (c) => {
  const form = await c.req.formData();
  return c.json({
    form: form.getAll("form_value_undefined"),
    query: c.req.queries("query_value_undefined"),
  });
});
```

Client:
```js
const resp = await hc().test.$post({
  form: { form_value_undefined: undefined },
  query: { query_value_undefined: undefined },
});
await resp.json();

```

undefined values in form should be skipped, consistent with query behavior. The response should be:

```json
{ "form": [], "query": [] }
```

### What do you see instead?

`query_value_undefined` is correctly omitted from the request, but `form_value_undefined` is sent as the string "undefined".

```json
{ "form": ["undefined"], "query": [] }
```

### Additional information

_No response_

aidenlx:
Closes #4731

The RPC client serializes `undefined` form values as the literal string `"undefined"` via `FormData.append()`, while query parameters correctly skip `undefined` values.

This PR adds the same `undefined` check to the form data serialization loop in `ClientRequestImpl`, matching the existing behavior in `buildSearchParams`.

## Changes

- `src/client/client.ts`: Add `if (v === undefined) { continue }` guard in the form data loop
- `src/client/client.test.ts`: Add test verifying undefined form values are skipped

### DIFF
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -447,6 +447,42 @@ describe('Form - Multiple Values', () => {
   })
 })
 
+describe('Form - Undefined Values', () => {
+  const server = setupServer(
+    http.post('http://localhost/form-undefined', async ({ request }) => {
+      const data = await request.formData()
+      return HttpResponse.json({
+        keys: [...data.keys()],
+        title: data.get('title'),
+        optional: data.get('optional'),
+      })
+    })
+  )
+
+  beforeAll(() => server.listen())
+  afterEach(() => server.resetHandlers())
+  afterAll(() => server.close())
+
+  const client = hc('http://localhost/')
+
+  it('Should skip undefined values in form data', async () => {
+    // @ts-expect-error `client['form-undefined'].$post` is not typed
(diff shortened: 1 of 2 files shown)


yusukebe:
@aidenlx 

Nice catch! Looks good. Thanks.

PR #4739, opened 2026-02-21
fix: improve cookie compatibility by adding expires to deleteCookie

### LINKED ISSUE #4740: Improve cookie compatibility
This PR improves cookie compatibility by adding an expires option to deleteCookie. The expires option is set to new Date(0), which ensures that the cookie is deleted immediately in all browsers, including older browsers that don't support the maxAge option.

luojiyin1987:
This PR improves cookie compatibility by adding an expires option to deleteCookie. The expires option is set to new Date(0), which ensures that the cookie is deleted immediately in all browsers, including older browsers that don't support the maxAge option.

Close #4740


yusukebe:
All modern browsers support Max-Aget. This fix is not needed. Closing.

PR #4750, opened 2026-02-22
fix(bearer-auth): escape regex metacharacters in bearer auth prefix option

otoneko1102:

- [x] Add tests
- [x] Run tests

---

"prefix" option in Bearer Auth middleware is interpolated directly into `new RegExp()` without escaping. This may cause SyntaxError crashes or unintended pattern matching when "prefix" contains regex metacharacters.

e.g., (, ), ., +, ...
- `prefix: 'Bearer|Token'` ...

### DIFF
--- a/src/middleware/bearer-auth/index.ts
+++ b/src/middleware/bearer-auth/index.ts
@@ -13,7 +13,6 @@ const TOKEN_STRINGS = '[A-Za-z0-9._~+/-]+=*'
 const PREFIX = 'Bearer'
 const HEADER = 'Authorization'
 
-const escapeRegExp = (str: string): string => str.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
 type MessageFunction = (c: Context) => string | object | Promise<string | object>
 type CustomizedErrorResponseOptions = {
   wwwAuthenticateHeader?: string | object | MessageFunction
@@ -113,9 +112,9 @@ export const bearerAuth = (options: BearerAuthOptions): MiddlewareHandler => {
   }
 
   const realm = options.realm?.replace(/"/g, '\\"')
-  const prefixRegexStr = options.prefix === '' ? '' : `${escapeRegExp(options.prefix)} +`
-  const regexp = new RegExp(`^${prefixRegexStr}(${TOKEN_STRINGS}) *$`, 'i')
-  const wwwAuthenticatePrefix = options.prefix === '' ? '' : `${options.prefix} `
+  const prefix = options.prefix
+  const tokenRegexp = new RegExp(`^${TOKEN_STRINGS}$`)
+  const wwwAuthenticatePrefix = prefix === '' ? '' : `${prefix} `
 
   const throwHTTPException = async (
     c: Context,
@@ -165,8 +164,19 @@ export const bearerAuth = (options: BearerAuthOptions): MiddlewareHandler => {
           'Unauthorized'
       )
     } else {
-      const match = regexp.exec(headerToken)
-      if (!match) {
+      let tokenValue: string | undefined
+
+      if (prefix === '') {
+        tokenValue = headerToken
+      } else {
+        const headerLower = headerToken.toLowerCase()
+        const prefixLower = prefix.toLowerCase()
+        if (headerLower.startsWith(prefixLower) && headerToken[prefix.length] === ' ') {
+          tokenValue = headerToken.slice(prefix.length).trimStart()
+        }
+      }
+
+      if (!tokenValue || !tokenRegexp.test(tokenValue)) {
         // Invalid Request
         await throwHTTPException(
           c,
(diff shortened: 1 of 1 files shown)


src/middleware/bearer-auth/index.ts
```diff
@@ -112,7 +114,7 @@ export const bearerAuth = (options: BearerAuthOptions): MiddlewareHandler => {
   }
 
   const realm = options.realm?.replace(/"/g, '\\"')
-  const prefixRegexStr = options.prefix === '' ? '' : `${options.prefix} +`
+  const prefixRegexStr = options.prefix === '' ? '' : `${escapeRegExp(options.prefix)} +`
```
EdamAme-x:
```suggestion

type MessageFunction = (c: Context) => string | object | Promise<string | object>
type CustomizedErrorResponseOptions = {
wwwAuthenticateHeader?: string | object | MessageFunction
@@ -112,7 +114,7 @@ export const bearerAuth = (options: BearerAuthOptions): MiddlewareHandler => {
}

const realm = options.realm?.replace(/"/g, '\\"')
  const prefixRegexStr = options.prefix === '' ? '' : `${RegExp.escape(options.prefix)} +`
```

You can use `RegExp.escape` instead, maybe

Ref: https://developer.mozilla.org/ja/docs/Web/JavaScript/Reference/Global_Objects/RegExp/escape

src/middleware/bearer-auth/index.ts
```diff
@@ -13,6 +13,8 @@ const TOKEN_STRINGS = '[A-Za-z0-9._~+/-]+=*'
 const PREFIX = 'Bearer'
 const HEADER = 'Authorization'
 
+const escapeRegExp = (str: string): string => str.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
```
EdamAme-x:
```suggestion

```

src/middleware/bearer-auth/index.ts
EdamAme-x: Based on node-server, it looks like Node.js v18 is a target for Hono. Because of this, RegExp.escape won't work, so I'm going to revert this for now

yusukebe:
Hey @otoneko1102

Thank you for the PR. I realized you don't need to use a regexp match for the prefix value. If so, the `escapeRegExp` in this PR is unnecessary, as shown by the following patch. The code is increased, but I think it is simple.

```diff
diff --git a/src/middleware/bearer-auth/index.ts b/src/middleware/bearer-auth/index.ts
index 3b9d225f..1a7e7854 100644
--- a/src/middleware/bearer-auth/index.ts
+++ b/src/middleware/bearer-auth/index.ts
@@ -13,7 +13,6 @@ const TOKEN_STRINGS = '[A-Za-z0-9._~+/-]+=*'
 const PREFIX = 'Bearer'
 const HEADER = 'Authorization'

-const escapeRegExp = (str: string): string => str.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
 type MessageFunction = (c: Context) => string | object | Promise<string | object>
 type CustomizedErrorResponseOptions = {
   wwwAuthenticateHeader?: string | object | MessageFunction
@@ -113,9 +112,9 @@ export const bearerAuth = (options: BearerAuthOptions): MiddlewareHandler => {
   }

   const realm = options.realm?.replace(/"/g, '\\"')
-  const prefixRegexStr = options.prefix === '' ? '' : `${escapeRegExp(options.prefix)} +`
-  const regexp = new RegExp(`^${prefixRegexStr}(${TOKEN_STRINGS}) *$`, 'i')
-  const wwwAuthenticatePrefix = options.prefix === '' ? '' : `${options.prefix} `
+  const prefix = options.prefix
+  const tokenRegexp = new RegExp(`^${TOKEN_STRINGS}$`)
+  const wwwAuthenticatePrefix = prefix === '' ? '' : `${prefix} `

   const throwHTTPException = async (
     c: Context,
@@ -165,8 +164,19 @@ export const bearerAuth = (options: BearerAuthOptions): MiddlewareHandler => {
           'Unauthorized'
       )
     } else {
-      const match = regexp.exec(headerToken)
-      if (!match) {
+      let tokenValue: string | undefined
+
+      if (prefix === '') {
+        tokenValue = headerToken
+      } else {
+        const headerLower = headerToken.toLowerCase()
+        const prefixLower = prefix.toLowerCase()
+        if (headerLower.startsWith(prefixLower) && headerToken[prefix.length] === ' ') {
+          tokenValue = headerToken.slice(prefix.length).trimStart()
+        }
+      }
+
+      if (!tokenValue || !tokenRegexp.test(tokenValue)) {
         // Invalid Request
         await throwHTTPException(
           c,
@@ -180,12 +190,12 @@ export const bearerAuth = (options: BearerAuthOptions): MiddlewareHandler => {
       } else {
         let equal = false
         if ('verifyToken' in options) {
-          equal = await options.verifyToken(match[1], c)
+          equal = await options.verifyToken(tokenValue, c)
         } else if (typeof options.token === 'string') {
-          equal = await timingSafeEqual(options.token, match[1], options.hashFunction)
+          equal = await timingSafeEqual(options.token, tokenValue, options.hashFunction)
         } else if (Array.isArray(options.token) && options.token.length > 0) {
           for (const token of options.token) {
-            if (await timingSafeEqual(token, match[1], options.hashFunction)) {
+            if (await timingSafeEqual(token, tokenValue, options.hashFunction)) {
               equal = true
               break
             }
```

@otoneko1102 @EdamAme-x @usualoma What do you think of this?

otnc:
I agree with you

yusukebe:
@otoneko1102 Thanks! Can you change the code?

otnc:
@yusukebe OK! I'll do it later.
Thanks for the review and advice!

yusukebe:
@otoneko1102 Thanks. Please ping me if it's done!

otnc:
@yusukebe Sorry for the delay! Fixed it!

An error has occurred for Node.js v20 test.
It is not a bug in the code, but the GitHub Actions call is failing.

EdamAme-x:
Looks good!

yusukebe:
@otoneko1102 @EdamAme-x 

The CI works well now. Looks good! Thanks!

PR #4751, opened 2026-02-22
feat(middleware): add Method Not Allowed middleware

### LINKED ISSUE #4633: Return `405 Method Not Allowed` when a path exists but the HTTP method is unsupported

rokasta12:
## Summary

Add a new `methodNotAllowed` middleware that returns `405 Method Not Allowed` with an `Allow` header when a request matches a registered route path but the HTTP method is not supported.

This implements the approach proposed by @usualoma in #4633 ([comment](https://github.com/honojs/hono/issues/4633#issuecomment-3770141489)) and confirmed by @yusukebe ([comment](https://github.com/honojs/hono/issues/4633#issuecomment-3776607069)).

Closes #4633

## Usage

```ts
import { Hono } from 'hono'
import { methodNotAllowed } from 'hono/method-not-allowed'

const app = new Hono()

app.get('/hello', (c) => c.text('Hello!'))
app.post('/hello', (c) => c.text('Posted!'))

// GET /hello  => 200 "Hello!"
// POST /hello => 200 "Posted!"
// PUT /hello  => 405 Method Not Allowed (Allow: GET, POST)
// GET /foo    => 404 Not Found
```

## Implementation Details

- Takes the `app` instance as an argument (same pattern as `methodOverride`)
- Lazily builds an internal `TrieRouter` from `app.routes` on first 404 encounter
- Checks if the request path matches any registered route with a different method
- Throws `HTTPException(405)` with `Allow` header listing valid methods (per RFC 7231)
- Skips routes registered with `app.all()` / `app.use()` (`METHOD_NAME_ALL`) when collecting methods
- Deduplicates methods in the `Allow` header

## Tests (16 passing)

| Test | Description |
|------|-------------|
| Valid GET/POST requests | Returns 200 as normal |
| Unsupported method on existing path | Returns 405 with "Method Not Allowed" body |
| Allow header | Lists permitted methods, excludes request method |
| Non-existent path | Returns 404 (not 405) |
| Parameterized routes (`:id`) | Correctly returns 405 with all allowed methods |
| Works alongside other middleware | `app.use()` middleware doesn't break 405 detection |
| Sub-applications | `app.route('/sub', sub)` routes are handled correctly |
| HEAD requests on GET routes | Returns 200 (Hono handles HEAD automatically) |
| OPTIONS when explicitly defined | Returns the defined response, not 405 |
| `app.all()` routes | No 405 — any method is valid |

## Checklist

- [x] Tests pass: 16/16
- [x] 100% code coverage (statements, branches, functions, lines)
- [x] TypeScript: `npx tsc --noEmit` — zero errors
- [x] Formatting: `bun run format:fix` — no changes needed
- [x] Linting: `npx eslint src/middleware/method-not-allowed/` — zero warnings
- [x] Export entries added to `package.json` (exports + typesVersions) and `jsr.json`
- [x] Export validation: `build/validate-exports.test.ts` passes

### DIFF
--- a/jsr.json
+++ b/jsr.json
@@ -45,6 +45,7 @@
     "./timeout": "./src/middleware/timeout/index.ts",
     "./timing": "./src/middleware/timing/timing.ts",
     "./logger": "./src/middleware/logger/index.ts",
+    "./method-not-allowed": "./src/middleware/method-not-allowed/index.ts",
     "./method-override": "./src/middleware/method-override/index.ts",
     "./powered-by": "./src/middleware/powered-by/index.ts",
     "./pretty-json": "./src/middleware/pretty-json/index.ts",
--- a/package.json
+++ b/package.json
@@ -230,6 +230,11 @@
       "import": "./dist/middleware/logger/index.js",
       "require": "./dist/cjs/middleware/logger/index.js"
     },
+    "./method-not-allowed": {
+      "types": "./dist/types/middleware/method-not-allowed/index.d.ts",
+      "import": "./dist/middleware/method-not-allowed/index.js",
+      "require": "./dist/cjs/middleware/method-not-allowed/index.js"
+    },
     "./method-override": {
       "types": "./dist/types/middleware/method-override/index.d.ts",
       "import": "./dist/middleware/method-override/index.js",
@@ -517,6 +522,9 @@
       ],
       "logger": [
         "./dist/types/middleware/logger"
+      ],
+      "method-not-allowed": [
+        "./dist/types/middleware/method-not-allowed"
       ],
       "method-override": [
         "./dist/types/middleware/method-override"
--- a/src/middleware/method-not-allowed/index.test.ts
+++ b/src/middleware/method-not-allowed/index.test.ts
@@ -1,0 +1,158 @@
+import { Hono } from '../../hono'
+import { methodNotAllowed } from '.'
+
+describe('Method Not Allowed Middleware', () => {
+  const app = new Hono()
+
+  app.use(methodNotAllowed({ app }))
+
+  app.get('/hello', (c) => c.text('Hello!'))
+  app.post('/hello', (c) => c.text('Posted!'))
+  app.get('/users/:id', (c) => c.text(`User ${c.req.param('id')}`))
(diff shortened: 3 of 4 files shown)


yashmakhija:
Clean implementation. One edge case: the `methodNotAllowedRouter` is built and cached on first 404. If routes are added dynamically after that, they won't be reflected in the cached router.

Probably fine for typical use (routes defined at startup), but worth a note in the docs that this middleware assumes routes are registered before the first request.

rokasta12:
Thanks for the careful read @yashmakhija — appreciate it.

I dug into the dynamic-route concern and it turns out Hono itself prevents the scenario: `SmartRouter.add` throws `MESSAGE_MATCHER_IS_ALREADY_BUILT` once the matcher has been built (i.e. after the first request matches), so users can't register routes after the first match anyway. I added a test for it and confirmed Hono raises before the middleware's cache is even consulted. So the cache being built once on first 404 is consistent with the rest of the framework — extra invalidation logic would be defensive against a scenario the runtime forbids.

That said, your review made me re-examine the middleware end-to-end and surfaced a few real correctness issues that I've now fixed in the latest force-push:

1. **Intentional 404s were being silently converted to 405.** If a registered handler returned `c.notFound()` (e.g. `GET /users/:id` for "user not found"), the middleware previously returned 405 with `Allow` listing the *other* methods, dropping the handler's 404. Fix: skip the transform when the request's method is registered for the path.

2. **Switched from `throw new HTTPException` to setting `c.res` directly.** 405 is a well-defined expected response, not an error condition — it shouldn't be routed through `app.onError`.

3. **Removed the `method !== c.req.method` filter on the `Allow` header.** It was dead code (the request method is never in the matched list when we reach this branch) and would have been wrong if it ever fired — RFC 9110 §15.5.6 wants `Allow` to list every method the resource supports.

4. Exported `MethodNotAllowedOptions` for typed wrappers, matching every other middleware in the package.

New tests cover the intentional-404 case. Thanks again for the prompt — without your comment I'd have stopped at "tests pass."

PR #4753, opened 2026-02-22
fix(validator): prevent type diff bug in form data parsing

EdamAme-x:
Use `Object.create(null)` instead of `{}` for the form data object and replace `key in form` with `Object.hasOwn(form, key)` to prevent `__proto__` keys in FormData from polluting the object prototype chain.

I think that this issue itself doesn't present a significant security risk and can be viewed as just a bug.

```ts
import {
    Hono
} from 'hono'
import {
    validator
} from 'hono/validator'
const app = new Hono()

app.post('/form', validator('form', (value) => value), (c) => {
    const data = c.req.valid('form')
    return c.json(data)
})

const form = new FormData()
form.append('toString', 'hello')
form.append('name', 'alice')

const res = await app.request('/form', {
    method: 'POST',
    body: form
})
const body = await res.json()

console.log(JSON.stringify(body, null, 2))
/***
     {
       "toString": [
         null,
         "hello"
       ],
       "name": "alice"
     }
***/
console.log(typeof body.toString) // object
console.log('toString', body.toString) // [ null, "hello" ]
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/validator/validator.test.ts
+++ b/src/validator/validator.test.ts
@@ -1434,9 +1434,7 @@ describe('Form validator prototype pollution prevention', () => {
 
     const res = await app.request('/form', { method: 'POST', body: form })
     const result = await res.json()
-    // Object.create(null) → __proto__ is stored as a regular property
     expect(result['__proto__']).toBe('evil')
-    // Object.hasOwn → toString is not mistaken for an existing key and stays a string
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @EdamAme-x 

That's a bug. Good fix. Thanks!

PR #4754, opened 2026-02-22
fix(jwt): use `Math.floor` instead of bitwise OR for safe timestamp

EdamAme-x:
The bitwise OR operator `| 0` truncates numbers to 32-bit signed integers, causing overflow after 2038-01-19 when Unix timestamps exceed 2^31-1. This would make expired JWT tokens pass validation.

Replace `(Date.now() / 1000) | 0` with `Math.floor(Date.now() / 1000)` and add a regression test that verifies expired tokens are correctly rejected when the current time is past the Y2038 boundary.

While this would typically be a severe vulnerability, it is not expected to occur for another 10 years, so I believe it poses no immediate issue.

### PoC
```ts
// PoC: JWT Y2038
const _originalDateNow = Date.now
Date.now = () => 2147483648 * 1000 // 2038-01-19 03:14:08 UTC

import { Jwt } from 'hono/utils/jwt'

async function main() {
  const secret = 'my-secret'
  const token = await Jwt.sign({ sub: 'user', exp: 1735689600 }, secret, 'HS256')

  try {
    const payload = await Jwt.verify(token, secret, 'HS256')
    console.log('[VULN]:', payload)
  } catch (e) {
    console.log('[OK]:', (e as Error).message)
  }
}

main().finally(() => { Date.now = _originalDateNow })
```

#### Result
image

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/utils/jwt/jwt.test.ts
+++ b/src/utils/jwt/jwt.test.ts
@@ -122,7 +122,7 @@ describe('JWT', () => {
     expect(authorized).toBeUndefined()
   })
 
-  it('JwtTokenExpired after Y2038 (regression: bitwise OR overflow)', async () => {
+  it('JwtTokenExpired after Y2038', async () => {
     const postY2038 = 2147483648 // 2038-01-19 03:14:08 UTC
     vi.useFakeTimers().setSystemTime(new Date(postY2038 * 1000))
 


EdamAme-x:
I saw an actual instance of the Y2038 bug for the first time.

yusukebe (approved):
LGTM!

yusukebe:
@EdamAme-x 

Hono will exist over 2038. Thanks!

PR #4756, opened 2026-02-23
fix: throw TypeError for non-JSON-serializable values in c.json()

veeceey:
Fixes #2343

When `c.json(undefined)` is called, `JSON.stringify(undefined)` returns `undefined` (not a string), which leads to inconsistent behavior across runtimes - Bun returns an empty string `""`, Cloudflare returns `"undefined"`, and Node throws an error.

This follows the same approach discussed in the issue thread, matching Deno's implementation: check the result of `JSON.stringify` and throw a `TypeError` if the value isn't serializable. This way the behavior is consistent across all runtimes instead of silently returning wrong responses.

Also moved the fast path to use `new Response(body, ...)` instead of `Response.json(object)` so the serialization check happens before sending the response regardless of which code path is taken.

Added a test to verify the behavior.

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -724,11 +724,7 @@ export class Context<
     return (
       this.#useFastPath() && !arg && !headers
         ? new Response(body, { headers: { 'content-type': 'application/json' } })
-        : this.#newResponse(
-            body,
-            arg,
-            setDefaultContentType('application/json', headers)
-          )
+        : this.#newResponse(body, arg, setDefaultContentType('application/json', headers))
     ) as JSONRespondReturn<T, U>
   }
 


veeceey:
### Test Results

Ran the full test suite - all context tests pass including the new test:

```
✓ |main| src/context.test.ts (58 tests) 17ms
  Tests  58 passed (58)
```

Full suite:
```
Test Files  1 failed | 143 passed (144)
     Tests  4 failed | 4089 passed | 33 skipped (4126)
```

The 4 pre-existing failures are in `runtime-tests/node/index.test.ts` (stream abort handler tests) and unrelated to this change.

**Manual verification:**
```js
// Before: JSON.stringify(undefined) → undefined → new Response(undefined) → "" (Bun) / "undefined" (CF)
// After: throws TypeError("Value is not JSON serializable")
```

yusukebe:
@veeceey 

Thank you for the PR. If this issue is introduced by https://github.com/honojs/hono/pull/4707, https://github.com/honojs/hono/pull/4757 fixed the problem.

veeceey:
Great, thanks for the heads up — #4757 addresses the root cause. Closing this one.

PR #4759, opened 2026-02-23
fix(jwt): prevent memory leak by avoiding mutation of options object

EdamAme-x:
The current implementation mutates the keys in the options object, which causes a memory leak.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/utils/jwt/jwt.ts
+++ b/src/utils/jwt/jwt.ts
@@ -233,7 +233,7 @@ export const verifyWithJwks = async (
     if (!Array.isArray(data.keys)) {
       throw new Error('invalid JWKS response. "keys" field is not an array')
     }
-    verifyKeys = verifyKeys ?? []
+    verifyKeys ??= []
     verifyKeys.push(...(data.keys as HonoJsonWebKey[]))
   } else if (!verifyKeys) {
     throw new Error('verifyWithJwks requires options for either "keys" or "jwks_uri" or both')


usualoma:
Hi @EdamAme-x 
Thanks for creating the pull request!

As you pointed out, I also think changing the value of the `options` passed as an argument is a bug that should be fixed. Thank you for noticing this important detail.

By the way, since a linear search is ultimately performed as follows, and the first found `.kid` key is used as a result, I think it's better to omit `mergeJwkKeys` and use the following approach to reduce the amount of code. What do you think?

```
verifyKeys.find((key) => key.kid === header.kid)
```

```diff
diff --git i/src/utils/jwt/jwt.ts w/src/utils/jwt/jwt.ts
index 16ba0209..1f2872f7 100644
--- i/src/utils/jwt/jwt.ts
+++ w/src/utils/jwt/jwt.ts
@@ -188,26 +188,6 @@ const symmetricAlgorithms: SymmetricAlgorithm[] = [
   AlgorithmTypes.HS512,
 ]
 
-const mergeJwkKeys = (
-  existingKeys: readonly HonoJsonWebKey[],
-  fetchedKeys: readonly HonoJsonWebKey[]
-): HonoJsonWebKey[] => {
-  const mergedKeys: HonoJsonWebKey[] = []
-  const seenKids = new Set<string>()
-
-  for (const key of [...existingKeys, ...fetchedKeys]) {
-    if (typeof key.kid === 'string') {
-      if (seenKids.has(key.kid)) {
-        continue
-      }
-      seenKids.add(key.kid)
-    }
-    mergedKeys.push(key)
-  }
-
-  return mergedKeys
-}
-
 export const verifyWithJwks = async (
   token: string,
   options: {
@@ -239,7 +219,7 @@ export const verifyWithJwks = async (
     throw new JwtAlgorithmNotAllowed(header.alg, options.allowedAlgorithms)
   }
 
-  let verifyKeys = options.keys ? [...options.keys] : undefined
+  const verifyKeys = options.keys ? [...options.keys] : []
 
   if (options.jwks_uri) {
     const response = await fetch(options.jwks_uri, init)
@@ -253,7 +233,7 @@ export const verifyWithJwks = async (
     if (!Array.isArray(data.keys)) {
       throw new Error('invalid JWKS response. "keys" field is not an array')
     }
-    verifyKeys = mergeJwkKeys(verifyKeys ?? [], data.keys as HonoJsonWebKey[])
+    verifyKeys.push(...data.keys as HonoJsonWebKey[])
   } else if (!verifyKeys) {
     throw new Error('verifyWithJwks requires options for either "keys" or "jwks_uri" or both')
   }
```

EdamAme-x:
I think so! thanks you!

usualoma:
Hi @EdamAme-x 
Thank you. LGTM!

yusukebe:
@EdamAme-x @usualoma Thank you both!

PR #4763, opened 2026-02-24
refactor(client): make properties readonly and adjust modifiers

BelovedYaoo:

- [ ] Add tests
- [ ] Run tests

mark properties as readonly and adjust access modifiers in client classes to improve immutability and encapsulation :)

### DIFF
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -31,9 +31,9 @@ const createProxy = (callback: Callback, path: string[]) => {
 }
 
 class ClientRequestImpl {
-  private url: string
+  private readonly url: string
   private method: string
-  private buildSearchParams: BuildSearchParamsFn
+  private readonly buildSearchParams: BuildSearchParamsFn
   private queryParams: URLSearchParams | undefined = undefined
   private pathParams: Record<string, string> = {}
   private rBody: BodyInit | undefined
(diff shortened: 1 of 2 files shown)


src/client/client.ts
```diff
@@ -31,9 +31,9 @@ const createProxy = (callback: Callback, path: string[]) => {
 }
 
 class ClientRequestImpl {
```
yusukebe, line 33: `ClientRequestImpl` is only used in internal. This refactor is not necessary right now.

src/client/fetch-result-please.ts
```diff
@@ -47,19 +47,19 @@ export class DetailedError extends Error {
   /**
    * Additional `message` that will be logged AND returned to client
    */
-  public detail?: any
+  detail?: any
```
yusukebe, line 50: This file is ported from another package https://github.com/honojs/hono/blob/b1df30439afcaac8205790ecad272f49caf0fa7e/src/client/fetch-result-please.ts#L4

The author intended marking as `public`. And it's not a problem making them `public`. This refactoring is not necessary.

yusukebe:
I think these refactorings are not necessary. Closing.

BelovedYaoo:
ok,sorry for making two unnecessary changes. :)

PR #4766, opened 2026-02-24
feat: extend basePath() helper to accept Hono instances

G4brym:
## Summary

Extends the existing `basePath()` helper in `hono/route` to also accept a `Hono` instance, allowing libraries to read a Hono app's base path at setup time — without relying on the private `_basePath` property.

## Motivation

Currently, the only way to read a Hono instance's base path outside of a request handler is by accessing the private `_basePath` property. This is fragile — it is not part of the public API and could break without notice.

Libraries that integrate with Hono (such as [chanfana](https://github.com/cloudflare/chanfana), an OpenAPI schema generator) need to detect the base path to correctly generate API documentation and route schemas. Right now, chanfana reads `_basePath` defensively with runtime type guards ([see PR](https://github.com/cloudflare/chanfana/pull/306)), but a proper public API would be much better.

## Changes

- **`src/helper/route/index.ts`** — Added an overload to `basePath()` that accepts a `HonoBase` instance and returns its base path directly
- **`src/helper/route/index.test.ts`** — Added 4 tests for the new overload:
  - Returns `"/"` for a fresh instance
  - Returns the base path after `basePath()` 
  - Returns the merged path after chained `basePath()` calls
  - Original instance is unaffected after `basePath()` creates a clone

## Usage

```ts
import { basePath } from 'hono/route'

const app = new Hono().basePath('/api')
basePath(app) // '/api'

const v1 = app.basePath('/v1')
basePath(v1)  // '/api/v1'
basePath(app) // '/api' (unchanged)
```

The existing `basePath(c: Context)` handler behaviour is fully preserved.

### DIFF
--- a/src/helper/route/index.test.ts
+++ b/src/helper/route/index.test.ts
@@ -1,4 +1,5 @@
 import { Context } from '../../context'
+import { Hono } from '../../hono'
 import { matchedRoutes, routePath, baseRoutePath, basePath } from '.'
 
 const defaultContextOptions = {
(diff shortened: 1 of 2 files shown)


yusukebe:
Hey @G4brym 

If you want to get a base path in a handler, you can use the `basePath()` in the Route Helper: https://hono.dev/docs/helpers/route#basepath

Have you checked it?

G4brym:
Hey @yusukebe, thanks for the pointer!                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                          I checked that, but it requires a Context (c) and only works inside a request handler
my use case is different, I need the base path at setup time, outside of a request

for example:
```js
const app = new Hono().basePath('/api')                                                                                                                                                                                                                                                                              
const router = fromHono(app)  // ← need to know "/api" here, no `c` exists yet 
```

G4brym:
or maybe the basePath helper function could be updated to also work with Hono apps, like:
```js
import {
  basePath,
} from 'hono/route'

const subApp = new Hono()
const path = basePath(subApp) // '/api'

app.route('/:sub', subApp)
```
i can update the pr if you agree with this change

G4brym:
Hey there, just updated the pr to match the comment above

yusukebe:
Hi @G4brym 

Sorry for being late. I think this is good.

We have to discuss whether we should use overload. We can also use such as `basePathFromApp`.

Hey @usualoma What do you think of this PR?

PR #4767, opened 2026-02-24
fix(types): preserve response inference for Promise.then handlers

### LINKED ISSUE #4765: BUG: RPC types broken when using arrow function with a Promise

4.12.2

# Reproduction

```ts
import { Hono } from "hono";

const app = new Hono().get("/", (c) =>
  Promise.resolve({ hello: "world" }).then((d) => c.json(d)),
);

const api = hc<typeof app>("");

// data has unknown type
const data = await api.index.$get().then((r) => r.json());
```

### But the same code without arrow funcion works:
```ts
import { Hono } from "hono";

const app = new Hono().get("/", (c) => {
  const resp = Promise.resolve({ hello: "world" }).then((d) => c.json(d));
  return resp;
});

const api = hc<typeof app>("");

// data has { hello: string; } type
const data = await api.index.$get().then((r) => r.json());
```

Types should be inferred in both cases

### What do you see instead?

RPC json response has `unknown` type

EdamAme-x:
Close: https://github.com/honojs/hono/issues/4765

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -620,8 +620,9 @@ describe('Infer the response/request type', () => {
       const req = client.index.$get
 
       type Actual = InferResponseType<typeof req>
-      type Expected = { hello: string }
-      type verify = Expect<Equal<Expected, Actual>>
+      type verify = Expect<
+        Actual extends { hello: string } ? ({ hello: string } extends Actual ? true : false) : false
+      >
     })
 
     it('Should infer request type the type correctly', () => {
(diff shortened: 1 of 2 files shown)


EdamAme-x:
~~The implementation itself works with this, but it will result in a large diff if we try to fix all patterns (app.|any|(handler x |any|))~~

valerii15298:
I don't know specifics here but curious if Awaited would have been helpful https://www.typescriptlang.org/docs/handbook/utility-types.html#awaitedtype

Issue #3312, opened 2024-08-22
Interface 'ClientResponse<T, U, F>' incorrectly extends interface 'Response'.


osameyy:
### What version of Hono are you using?

4.5.8

### What runtime/platform is your app running on?

Cloudflare Worker

### What steps can reproduce the bug?

hono: 4.5.8

### Additional information

`node_modules/.pnpm/hono@4.5.8/node_modules/hono/dist/types/client/types.d.ts:46:18 - error TS2430: Interface 'ClientResponse<T, U, F>' incorrectly extends interface 'Response'.
  The types returned by 'clone().json()' are incompatible between these types.
    Type 'Promise<unknown>' is not assignable to type 'Promise<T>'.
      Type 'unknown' is not assignable to type 'T'.
        'T' could be instantiated with an arbitrary type which could be unrelated to 'unknown'.

46 export interface ClientResponse<T, U extends number = StatusCode, F extends ResponseFormat = ResponseFormat> extends globalThis.Response {
                    ~~~~~~~~~~~~~~


Found 1 error in node_modules/.pnpm/hono@4.5.8/node_modules/hono/dist/types/client/types.d.ts:46`

osameyy:
@yusukebe any help here?

yusukebe:
Hi @osameyy 

I never understood your problem with just your information. Please provide a minimal project to reproduce it.

yusukebe:
Closing this issue. If you will have still problem, feel free to reopen.

Issue #1200, opened 2023-06-25
using hono requires skipLibCheck


aroman:
Hi,

I've just tried adding your library to my project. However, simply running `new Hono();` breaks my typescript project due to numerous type issues with your bundled types; some of your types expect the global `Request` to be a DOM request object, but this is not true in the workers environment.

Passing `skipLibCheck: true` to my tsconfig suppresses the errors, which I notice is what the [hono repo itself does in its own tsconfig](https://github.com/honojs/hono/blob/4b2dfb824ba68c6dd03ea93ba60649a95d370e80/tsconfig.json#L10).

However, this is definitely not a good thing to do, as it disables ALL type checking for all dependencies.

Here is a demonstration. Any ideas? I've also included my tsconfig below. 

```
~/P/server main• 1s ❱ tsc --noEmit --project .
~/P/server main• ❱ echo "Note: no type errors in my project prior to adding hono"
Note: no type errors in my project prior to adding hono
~/P/server main• ❱ npm install hono

added 1 package, and audited 915 packages in 883ms

152 packages are looking for funding
  run `npm fund` for details

1 moderate severity vulnerability

To address all issues, run:
  npm audit fix

Run `npm audit` for details.
~/P/server main• 1s ❱ echo -e "import { Hono } from 'hono';\n const app = new Hono();" >> src/worker.ts
~/P/server main• ❱ tsc --noEmit --project .
../node_modules/hono/dist/types/client/types.d.ts:22:18 - error TS2430: Interface 'ClientResponse<T>' incorrectly extends interface 'Response'.
  The types returned by 'json()' are incompatible between these types.
    Type 'Promise<T>' is not assignable to type 'Promise<T>'. Two different types with this name exist, but they are unrelated.
      Type 'T' is not assignable to type 'T'. Two different types with this name exist, but they are unrelated.
        'T' could be instantiated with an arbitrary type which could be unrelated to 'T'.

22 export interface ClientResponse<T> extends Response {
                    ~~~~~~~~~~~~~~

  ../node_modules/hono/dist/types/client/types.d.ts:22:33
    22 export interface ClientResponse<T> extends Response {
                                       ~
    This type parameter might need an `extends T` constraint.

../node_modules/hono/dist/types/request.d.ts:51:21 - error TS2304: Cannot find name 'RequestRedirect'.

51     get redirect(): RequestRedirect;
                       ~~~~~~~~~~~~~~~

[...]

PR #4768, opened 2026-02-24
fix(types): remove DOM type dependencies from ClientResponse and request method

### LINKED ISSUE #3750: Type 'ClientResponse<T, U, F>' is not assignable to type 'void | Response'.

4.6.13

Cloudflare Workers

reproduction: https://github.com/satoshun00/hono-client-response-repro

```ts
import { Hono } from "hono";
import app from "./index";

const client = hc<typeof app>("http://localhost:8787");

const app2 = new Hono()
  .post("/", () => {
    return client.index.$post();
  });
```

It should be possible to return the response (`ClientResponse`) from hono/client as the response (`globalThis.Response`) of the handler.

### What do you see instead?

TypeScript Error
```
    Argument of type '() => Promise<ClientResponse<{ message: string; }, StatusCode, "json">>' is not assignable to parameter of type 'H<BlankEnv, "/", BlankInput, HandlerResponse<any>>'.
      Type '() => Promise<ClientResponse<{ message: string; }, StatusCode, "json">>' is not assignable to type 'MiddlewareHandler<BlankEnv, "/", BlankInput>'.
        Type 'Promise<ClientResponse<{ message: string; }, StatusCode, "json">>' is not assignable to type 'Promise<void | Response>'.
          Type 'ClientResponse<{ message: string; }, StatusCode, "json">' is not assignable to type 'void | Response'.
            Type 'ClientResponse<{ message: string; }, StatusCode, "json">' is not assignable to type 'Response'.
              The types returned by 'json()' are incompatible between these types.
                Type 'Promise<{ message: string; }>' is not assignable to type 'Promise<T>'.
                  Type '{ message: string; }' is not assignable to type 'T'.
                    'T' could be instantiated with an arbitrary type which could be unrelated to '{ message: string; }'.ts(2769)
```

### Additional information

**Additional information**

- Related issue: #3312
- Currently, the following type casting is used as a temporary workaround to avoid type errors:  
  ```ts
  const app2 = new Hono()
    .post("/", () => {
      return client.index.$post() as Promise<Response>;
    });
  ```

YevheniiKotyrlo:
## Summary

- Remove `extends globalThis.Response` from `ClientResponse` interface in `src/client/types.ts`, adding `redirected`, `type`, and `bytes()` properties inline so the interface remains structurally compatible with `Response`
- Replace `RequestInfo | URL` with `Request | string | URL` in `Hono.request()` method signature in `src/hono-base.ts` (semantically identical since `RequestInfo = Request | string` in DOM lib)

## Motivation

Hono's type declarations depend on DOM-only types (`globalThis.Response`, `RequestInfo`) that are unavailable in non-DOM TypeScript environments (e.g. Bun with `types: ["bun"]`, Node without DOM lib). Projects using `skipLibCheck: false` in these environments hit `TS2304` errors when importing from hono.

Related issues: #1200, #3312, #3750

### `src/client/types.ts`

`ClientResponse` already redeclares all `Response` members inline. Removing the `extends globalThis.Response` clause and adding three previously-inherited properties (`redirected`, `type`, `bytes()`) makes the interface self-contained and structurally compatible with `Response` without requiring DOM types.

### `src/hono-base.ts`

`RequestInfo` is defined as `Request | string` in `lib.dom.d.ts`. Expanding it inline to `Request | string | URL` removes the DOM dependency while being semantically identical.

## Checklist

- [x] Added tests (no new runtime behavior; existing 4097 tests all pass)
- [x] Ran tests (`vitest --run` — 144/144 files, 4097/4097 tests pass)
- [x] `bun run format:fix && bun run lint:fix` (0 errors, 50 warnings — identical to baseline)
- [x] `tsc --noEmit` — 0 errors
- [x] `bun run build` — CJS + ESM + declarations all build successfully
- [x] `editorconfig-checker` — pass

Closes #3750

### DIFF
--- a/src/client/types.ts
+++ b/src/client/types.ts
@@ -129,8 +129,7 @@ export interface ClientResponse<
   T,
   U extends number = StatusCode,
   F extends ResponseFormat = ResponseFormat,
->
-  extends globalThis.Response {
+> {
   readonly body: ReadableStream | null
   readonly bodyUsed: boolean
   ok: U extends SuccessStatusCode
@@ -138,12 +137,15 @@ export interface ClientResponse<
     : U extends Exclude<StatusCode, SuccessStatusCode>
       ? false
       : boolean
+  redirected: boolean
   status: U
   statusText: string
+  type: 'basic' | 'cors' | 'default' | 'error' | 'opaque' | 'opaqueredirect'
   headers: Headers
   url: string
   redirect(url: string, status: number): Response
   clone(): Response
+  bytes(): Promise<Uint8Array<ArrayBuffer>>
   json(): F extends 'text' ? Promise<never> : F extends 'json' ? Promise<T> : Promise<unknown>
   text(): F extends 'text' ? (T extends string ? Promise<T> : Promise<never>) : Promise<string>
   blob(): Promise<Blob>
--- a/src/hono-base.ts
+++ b/src/hono-base.ts
@@ -491,7 +491,7 @@ class Hono<
    * @see https://hono.dev/docs/api/hono#request
    */
   request = (
-    input: RequestInfo | URL,
+    input: Request | string | URL,
     requestInit?: RequestInit,
     Env?: E['Bindings'] | {},
     executionCtx?: ExecutionContext


YevheniiKotyrlo:
> I have a local branch which disables `lib: dom`. That is a much larger change, but I think this is a good start and is the right direction to take 👍🏻

Happy to apply all reviewer changes for this quick fix to pass

yusukebe:
Hi @YevheniiKotyrlo 

I also think removing DOM type dependencies is a good idea. Thanks!

PR #4774, opened 2026-02-26
fix(types): correct middleware types

### LINKED ISSUE #4773: `c.req.valid` produces invalid type for `c.req.valid` in route middlewares

4.12.2

Cloudflare Workers

```
import { Hono } from 'hono';
import { sValidator } from '@hono/standard-validator';
import { type } from 'arktype';

const app = new Hono()
    .get('/', sValidator('query', type({
        test: 'string'
    })), async (c, next) => {
        const { test } = c.req.valid('query'); // This expression is not callable.
    }, async (c) => {
        const { test } = c.req.valid('query'); // Works normally
        return c.json({ ok: true })
    });
```

`c.req.valid` still supported in middleware

### What do you see instead?

_No response_

### Additional information

_No response_

hmnd:
fixes #4773 

Found that #4393 used any's for all middlewares, rather than passing through the appropriate `In` (`I`) type to `H<...>`.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -155,7 +155,7 @@ export interface HandlerInterface<
     R extends HandlerResponse<any> = any,
     E2 extends Env = E,
     E3 extends Env = IntersectNonAnyTypes<[E, E2]>,
-    M1 extends H<E2, P, any> = H<E2, P, any>,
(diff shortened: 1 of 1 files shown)


yusukebe:
Hey @hmnd 

Good catch! Thanks!

PR #4777, opened 2026-02-28
fix(client): preserve route schema in ApplyGlobalResponse

agumy:
Fixes a type inference issue where `ApplyGlobalResponse` loses route paths for apps composed with `.route()`.

## Problem

`Hono.route()` can produce a union schema (`MergeSchemaPath<...> | S`).  
`ApplyGlobalResponse` was mapping over the inferred schema directly, and in this union case route keys could collapse, causing `hc<ApplyGlobalResponse<...>>` to be inferred as `unknown`

## Changes

- `src/client/types.ts`
  - Use `ExtractSchema<App>` inside `ApplyGlobalResponse` before applying `ModSchema`.
- `src/client/client.test.ts`
  - Add regression test: `Should keep route() paths when global responses are applied`.

## Checklist

- [x] `bun run test`
- [x] `bun run lint`
- [x] `bun run format`
- [x] `bun run build`

### DIFF
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -1842,5 +1842,4 @@ describe('ApplyGlobalResponse Type Helper', () => {
 
     type verify = Expect<Equal<ResponseType, Expected>>
   })
-
 })


yusukebe:
Hi @agumy 

Nice fix. Thank you!

Issue #1977, opened 2024-01-14
SSG/JSX: SSG supports `Suspense`


yusukebe:
### What is the feature you are proposing?

Currently, the SSG Helper does not fully support `Suspense`.

For example, consider the following example, which returns JSX with AsyncComponet wrapped in `Suspense`:

```tsx
app.get('/shops', async (c) => {
  return c.render(
    <div>
      <Suspense fallback={'loading...'}>
        <Component />
      </Suspense>
    </div>
  )
})
```

The HTML exported by SSG is as follows:

```html
<html>
    <head>
        <meta charset="UTF-8"/>
        <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
        <link href="/static/style.css" rel="stylesheet"/>
    </head>
    <body>
        <header>
            <a href="/">top</a>
            <a href="/about">about</a>
            <a href="/shops">shops</a>
        </header>
        <main>
            <div>
                <template id="H:1"></template>
                loading...
                <!--/$-->
            </div>
        </main>
        <footer>
            <address>
                <a href="https://github.com/yusukebe/ramen-api">powered by Ramen API</a>
            </address>
        </footer>
    </body>
</html>
<template>
    <ul>
        <li>
            <a href="/shops/yoshimuraya">yoshimuraya</a>
        </li>
        <li>
            <a href="/shops/sugitaya">sugitaya</a>
        </li>
        <li>
            <a href="/shops/takasagoya">takasagoya</a>
        </li>
    </ul>
</template>
<script>
    ((d,c,n)=>{
        c = d.currentScript.previousSibling
        d = d.getElementById('H:1')
        if (!d)
            return
        do {
            n = d.nextSibling;
            n.remove()
        } while (n.nodeType != 8 || n.nodeValue != '/$')d.replaceWith(c.content)
    }
    )(document)
</script>
```

I would like to show only the list of shops without ` loading...`. This means that the SSG should be executed with the AsyncComponent in `Suspense` resolved.

However, if we need to add a lot of code to resolve this issue, we can just write in the documentation that "SSG does not support Suspense completely".

EdamAme-x:
I believe this issue should be resolved.
Am I correct in understanding that this implementation will resolve it first when **SSG** Async components?

aretrace:
Any update on this?

yusukebe:
Currently, there are no updates on this matter.

3w36zj6:
I may be overlooking something, but what do you think of this approach?
[...]

PR #4780, opened 2026-03-03
feat(jsx-renderer): support function-based options

3w36zj6:
cf. https://github.com/honojs/hono/issues/1977#issuecomment-3728740345

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/jsx-renderer/index.ts
+++ b/src/middleware/jsx-renderer/index.ts
@@ -31,8 +31,14 @@ type ComponentWithChildren = (
 ) => HtmlEscapedString | Promise<HtmlEscapedString>
 
 const createRenderer =
-  (c: Context, Layout: FC, component?: Component, options?: RendererOptions) =>
+  (
+    c: Context,
+    Layout: FC,
+    component?: Component,
+    options?: RendererOptions | ((c: Context) => RendererOptions)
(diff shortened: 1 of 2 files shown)


src/middleware/jsx-renderer/index.ts
```diff
@@ -107,9 +113,9 @@ const createRenderer =
  * })
  * ```
  */
-export const jsxRenderer = (
+export const jsxRenderer = <E extends Env = Env>(
   component?: ComponentWithChildren,
-  options?: RendererOptions
+  options?: RendererOptions | ((c: Context<E>) => RendererOptions)
```
3w36zj6, line 118: To provide type inference for `Env`, I implemented it based on the example at:

https://github.com/honojs/hono/blob/6a0607a929d888893f0c91d92dce2fcfdb3662a3/src/middleware/serve-static/index.ts#L35

It would be convenient if the type of `env` is inherited from the Hono instance without needing to explicitly specify type arguments, but I think this would require changes to `src/types.ts`.

yusukebe:
Hey @3w36zj6 

This is good! This is a new feature, but it's slight, so I'll merge it into the main immediately and release it as a patch release. Thanks!

PR #4782, opened 2026-03-04
fix(lambda-edge): avoid callback handler deprecation on NODEJS_24_X

### LINKED ISSUE #4772: handler() from hono/lambda-edge triggers Runtime.CallbackHandlerDeprecated on NODEJS_24_X

4.12.2

AWS/CloudFront Lambda@edge Node 24

The handle() function exported from hono/lambda-edge returns an async function with three parameters:
`// node_modules/hono/dist/adapter/lambda-edge/handler.js
var handle = (app) => {
  return async (event, context, callback) => {
    // ...
  }
}`

Because the returned function has .length === 3, the AWS Lambda runtime identifies it as a callback-style handler. Starting with NODEJS_24_X, AWS Lambda has deprecated callback-based handlers, producing:

This happens even though the function is async and returns a Promise — the runtime uses handler.length (i.e., the number of formal parameters) to determine the handler type, not whether it's actually async.

Reproduction:

Create a Lambda@Edge function using hono/lambda-edge
Deploy with runtime: lambda.Runtime.NODEJS_24_X
Invoke the function — it fails with Runtime.CallbackHandlerDeprecated

(more claude opus output...)

The handler returned by handle(app) should work on all supported Node.js Lambda runtimes (NODEJS_18_X through NODEJS_24_X) without requiring a wrapper. Since callback is already optional and the function is already async, the returned function should have .length <= 2 so the Lambda runtime correctly identifies it as an async handler. On older runtimes that still support callbacks, the behavior would be unchanged — callback can still be passed positionally and used via c.env.callback() — but it should not be a formal parameter that inflates Function.length.

### What do you see instead?

_No response_

### Additional information

_No response_

t0waxx:
Closes #4772

## Problem

`handle()` from `hono/lambda-edge` returns an async handler with `Function.length === 3` because `callback` is a formal parameter. AWS Lambda NODEJS_24_X determines handler type by `Function.length` — when it's `>= 3`, the runtime treats it as a callback-style handler and throws `Runtime.CallbackHandlerDeprecated`.

## Solution

Replace the three formal parameters `(event, context, callback)` with `(event, ...args)` and destructure `context` and `callback` internally. This reduces `Function.length` to `1` while preserving positional callback compatibility for older runtimes.

### Before

```ts
return async (event: CloudFrontEdgeEvent, context?: CloudFrontContext, callback?: Callback) => {
  // Function.length === 3 → NODEJS_24_X treats as callback-style
}
```

### After

```ts
return async (event: CloudFrontEdgeEvent, ...args: [CloudFrontContext?, Callback?]) => {
  const [context, callback] = args
  // Function.length === 1 → NODEJS_24_X treats as async-style
}
```

## Tests Added

- Verify `handler.length <= 2` (regression test for the root cause)
- Verify positional callback still works via `handler(event, undefined, callback)`

```ts
it('Should expose async handler arity compatible with NODEJS_24_X', () => {
  expect(handler.length).toBeLessThanOrEqual(2)
})

it('Should preserve positional callback compatibility', async () => {
  type Env = { Bindings: { callback: Callback } }
  const app = new Hono<Env>()
  const callback = vi.fn()

  app.get('/test-path', (c) => {
    c.env.callback?.(null, {
      status: '200',
      headers: {
        'x-test': [{ key: 'x-test', value: 'ok' }],
      },
    })
    return c.text('ok')
  })

  const handler = handle(app)
  await handler(cloudFrontEdgeEvent, undefined, callback)

  expect(callback).toHaveBeenCalledWith(null, {
    status: '200',
    headers: {
      'x-test': [{ key: 'x-test', value: 'ok' }],
    },
  })
})
```

## Testing Note

- This fix directly addresses the `handler.length` issue, which is the documented cause of the `Runtime.CallbackHandlerDeprecated` error in the NODEJS_24_X runtime.
- Since this involves AWS infrastructure and I don't have an environment ready for immediate end-to-end testing, verification on a real deployment would be greatly appreciated.

---

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/adapter/lambda-edge/handler.ts
+++ b/src/adapter/lambda-edge/handler.ts
@@ -121,10 +121,7 @@ export const handle = (
   context?: CloudFrontContext,
   callback?: Callback
 ) => Promise<CloudFrontResult>) => {
-  return async (
-    event,
-    ...args: [context?: CloudFrontContext, callback?: Callback]
-  ) => {
+  return async (event, ...args: [context?: CloudFrontContext, callback?: Callback]) => {
     const [context, callback] = args
     const res = await app.fetch(createRequest(event), {
       event,


yusukebe:
@t0waxx Thanks!

PR #4791, opened 2026-03-05
feat(client): add `PickResponseByStatusCode` type

### LINKED ISSUE #4590: Since PR #4393, success response types can no longer be resolved without explicit status/ok checks
PR https://github.com/honojs/hono/pull/4393 improved the ability to get accurate types for specific status codes. While this is a great improvement on its own, it is causing an issue in my project.

In my project, we centrally manage 4xx and 5xx errors using a customized fetch function. Our design allows individual UI components to handle only the success cases. Previously, this meant our UI components were type-safe without needing to explicitly check the response code.

However, since v4.10.0 (where https://github.com/honojs/hono/pull/4393 was released), we can no longer resolve the types without checking status or ok in each UI component. As a result, we are forced to add these checks to every single API call site. This negatively affects the codebase by reducing readability (due to increased boilerplate) and increasing the bundle size.

Is there a way to maintain type safety without manually checking status or code, just as we were able to do before? Any advice would be appreciated.

_Originally posted by @baseballyama in https://github.com/honojs/hono/discussions/4575_

yusukebe:
Fixes #4590

Adding `PickResponseByStatusCode` to limit the status code for the Hono client in case of using a custom fetch that only returns a 200 response.

Usage:

```ts
type AppSuccessOnly = PickResponseByStatusCode<typeof app, 200>
const client = hc<AppSuccessOnly>('http://localhost')
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -15,6 +15,7 @@ import type {
   InferRequestType,
   InferResponseType,
   ApplyGlobalResponse,
+  PickResponseByStatusCode,
 } from './types'
 
 class SafeBigInt {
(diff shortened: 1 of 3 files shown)


yusukebe:
@baseballyama Thank you for reviewing. I'll merge this change and include it in the next patch version.

PR #4792, opened 2026-03-05
fix(jsx): align link hoisting and dedupe with React 19

### LINKED ISSUE #4789: link rel=canonical is silently removed when it shares the same href as link rel=alternate hreflang

usualoma:
fixes #4789

### Summary

Because React 19 appears to dedupe only link elements that are stylesheet + precedence, this PR aligns Hono JSX runtime link hoisting/dedupe behavior with that model.

It fixes over-deduplication by href (for canonical / alternate / preload / icon / preconnect, etc.) and narrows dedupe to stylesheet links with precedence only.

As shown below, the list of links with issues reported in the issue will also remain without being deduplicated.

https://github.com/honojs/hono/compare/fix-hoisting-link-element-3?expand=1#diff-7077a729697655835dd10a0638c9e8ffaad5aef95597d9f409cc4e3f3fb9f848R374-R384

### Compatibility

- No public API or type changes.
- Changes are limited to internal JSX runtime `link` hoisting/dedupe behavior and tests.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/jsx/dom/intrinsic-element/components.test.tsx
+++ b/src/jsx/dom/intrinsic-element/components.test.tsx
@@ -981,6 +981,42 @@ describe('intrinsic element', () => {
         expect(onError).toBeCalledTimes(1)
       })
 
+      it('should not register load handlers when the tag has no de-duplication key', () => {
+        const addEventListener = vi.fn<HTMLElement['addEventListener']>()
+        const onLoad = vi.fn()
+        const onError = vi.fn()
+
+        const App = () => {
+          return (
+            <div>
+              <title
+                ref={(e: HTMLTitleElement) => {
+                  if (!e) {
+                    return
+                  }
+                  const originalAddEventListener = e.addEventListener.bind(e)
+                  addEventListener.mockImplementation((...args) =>
+                    originalAddEventListener(...args)
+                  )
+                  e.addEventListener = addEventListener
+                }}
+                onLoad={onLoad}
+                onError={onError}
+              >
+                Document Title
+              </title>
+              Content
+            </div>
+          )
+        }
+        render(<App />, root)
+        expect(document.head.innerHTML).toBe('<title>Document Title</title>')
+        expect(root.innerHTML).toBe('<div>Content</div>')
+        expect(addEventListener).not.toBeCalled()
+        expect(onLoad).not.toBeCalled()
+        expect(onError).not.toBeCalled()
+      })
+
       it('should be blocked by blocking attribute', async () => {
         const Component = () => {
           return (


usualoma:
@yusukebe
Would you please review this?

usualoma:
Oops

usualoma:
@yusukebe 
My bad. I've fixed it.

src/jsx/intrinsic-element/common.ts
```diff
@@ -9,3 +11,13 @@ export const deDupeKeyMap: Record<string, string[]> = {
 export const domRenderers: Record<string, Function> = {}
 
 export const dataPrecedenceAttr = 'data-precedence'
+
+export const isStylesheetLinkWithPrecedence = (props: Props): boolean =>
+  props.rel === 'stylesheet' && 'precedence' in props
+
+export const shouldDeDupeByKey = (tagName: string, props: Props, supportSort: boolean): boolean => {
```
yusukebe: Can you apply the flowing patch to remove the lint error though CI does not fail?

```diff
diff --git a/src/jsx/intrinsic-element/common.ts b/src/jsx/intrinsic-element/common.ts
index aaa2ce0c..937c0974 100644
--- a/src/jsx/intrinsic-element/common.ts
+++ b/src/jsx/intrinsic-element/common.ts
@@ -15,7 +15,11 @@ export const dataPrecedenceAttr = 'data-precedence'
 export const isStylesheetLinkWithPrecedence = (props: Props): boolean =>
   props.rel === 'stylesheet' && 'precedence' in props

-export const shouldDeDupeByKey = (tagName: string, props: Props, supportSort: boolean): boolean => {
+export const shouldDeDupeByKey = (
+  tagName: string,
+  _props: Props,
+  supportSort: boolean
+): boolean => {
   if (tagName === 'link') {
     return supportSort
   }
```

usualoma:
@yusukebe 
Fixed in 9d63042 and improved test coverage in 2b0d6d3.

yusukebe:
@usualoma 

Thank you! Looks good.

PR #4800, opened 2026-03-11
fix(utils/mime): Normalize input extension to lowercase before MIME check

TheEssem:
When a filename with an upper/mixed-case extension is provided to getMimeType, it currently returns an undefined value which causes the static middleware to return a Content-Type of `application/octet-stream`. This PR normalizes upper/mixed-case file extensions during the check so that they can be properly matched to the MIME type map.

- [x] Add tests
- [x] Run tests


yusukebe:
@TheEssem Thanks! Good fix.

PR #4802, opened 2026-03-14
fix: buffer response body in request() to prevent async leak detection

andrew-t-james-wc:
## Problem

Test runners with async-leak detection (e.g. `vitest --detect-async-leaks`) flag every `app.request()` call as having a leaked Promise. The leak is the internal `ReadableStream` body created by `new Response(body, init)` inside `createResponseInstance` (context.js:11).

Every `c.json()`, `c.text()`, `c.html()` call creates a Response whose body stream is never consumed when tests only check `.status` or read the body via `.json()` (which internally consumes the stream but the original Promise from the Response constructor persists in the leak detector).

This affects any project using Hono + vitest with `--detect-async-leaks` enabled.

## Solution

Since `request()` is explicitly a **test helper** (the JSDoc shows a test case), buffering the response body before returning is safe and eliminates the false positives.

The fix wraps the `this.fetch()` result through `bufferResponse()` which:
- Reads the response body into an `ArrayBuffer` via `res.arrayBuffer()`
- Creates a new `Response` with the buffered body (no pending `ReadableStream`)
- Preserves `status`, `statusText`, and `headers`
- Falls back to the original `Response` if body consumption fails (e.g. stream already locked)
- Preserves synchronous throw behavior for error cases (non-async path)

## Impact

- All 201 existing `hono.test.ts` tests pass without modification
- `request()` return type is unchanged: `Response | Promise<Response>`
- Sync throw behavior is preserved for handler errors
- Negligible overhead since `request()` is only used in tests

### DIFF
--- a/.vitest.config/setup-vitest.ts
+++ b/.vitest.config/setup-vitest.ts
@@ -24,7 +24,8 @@ class MockCache {
   }
 
   async match(key: Request | string): Promise<Response | null> {
-    return this.store.get(key) || null
+    const stored = this.store.get(key)
+    return stored ? stored.clone() : null
   }
 
   async keys() {
--- a/src/hono-base.ts
+++ b/src/hono-base.ts
@@ -95,6 +95,37 @@ type MountOptions =
       replaceRequest?: MountReplaceRequest | false
     }
 
+/**
+ * Buffer a non-streaming Response so its internal ReadableStream body is fully consumed.
+ * Prevents async-leak detection tools (e.g. vitest --detect-async-leaks)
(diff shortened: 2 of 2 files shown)


andrew-t-james-wc:
@yusukebe address failure on main. should be good to go now.

yusukebe:
@andrew-t-james-wc 

This should not be implemented in Hono's core. The async leak detection issue is specific to the test runner, so it should be handled on the test side. not in the framework itself.  Closing.

andrew-t-james-wc:
@yusukebe will addres in a new pr, thanks for the feedback

PR #4807, opened 2026-03-18
fix(request): remove `parseBody` from bodyCache to prevent TypeError

### LINKED ISSUE #4806: parseBody() breaks subsequent text()/json() calls with TypeError

4.12.8

Node.js v22 (also reproducible on any runtime)

Calling `parseBody()` followed by `text()` or `json()` on the same request throws `TypeError: bodyCache[anyCachedKey].then is not a function`.

This happens when `Content-Type` is **not** `multipart/form-data` or `application/x-www-form-urlencoded` (e.g., `application/json` or no Content-Type header).

**Root cause:**
`parseBody()` stores a resolved plain object in `bodyCache.parsedBody` (via `??= await parseBody(...)`). When Content-Type is not a form type, `parseBody()` returns `{}` immediately without going through `formData()`, so `parsedBody` is the only key in `bodyCache`.

`#cachedBody()` then picks it up via `Object.keys(bodyCache)[0]` and calls `.then()` on it, assuming every cache entry is a `Promise`.

When Content-Type **is** a form type, `parseBody()` internally calls `request.formData()` which goes through `#cachedBody('formData')`, caching a `Promise` under `bodyCache.formData` first — so `Object.keys()[0]` is `'formData'` and `.then()` succeeds.

**Reproduction:**

```typescript
import { HonoRequest } from 'hono/request'

const req = new HonoRequest(
    method: 'POST',
    body: JSON.stringify({ event: 'push' }),
  })
)
await req.parseBody()  // returns {}, caches plain object in bodyCache.parsedBody
await req.text()       // TypeError: bodyCache[anyCachedKey].then is not a function
```

`text()` should return the raw request body text regardless of whether `parseBody()` was called first.

### What do you see instead?

`TypeError: bodyCache[anyCachedKey].then is not a function`

### Additional information

**Use case:** A middleware that tries to extract form data if present (without checking Content-Type upfront), then a downstream handler needs the raw body for signature verification:

```typescript
  const body = await c.req.parseBody()
  if (Object.keys(body).length > 0) {
    c.set('formData', body)
  }
  await next()
})

app.post('/webhook', async (c) => {
  const raw = await c.req.text()  // TypeError — even though parseBody() returned {}
  verify(raw, c.req.header('X-Signature'))
})
```

Note: This is related to but distinct from #3881 — that issue did not identify the root cause (non-Promise value in bodyCache) or the Content-Type dependency.

yusukebe:
Fixes #4806

I realized `bodyCache.parsedBody` in Hono Request is unnecessary, and this change will fix #4806.

`parseBody` uses `formData` inside, so it might already be cached:

https://github.com/honojs/hono/blob/fe689eceb7834db653a64ecae5f8d203d9c23b9a/src/utils/body.ts#L125

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/request.test.ts
+++ b/src/request.test.ts
@@ -354,6 +354,48 @@ describe('Body methods with caching', () => {
       expect(async () => await req.blob()).not.toThrow()
     })
 
+    describe('should not break body methods after parseBody() with non-form content-type', () => {
+      const createReq = () =>
+        new HonoRequest(
+          new Request('http://localhost', {
+            method: 'POST',
+            body: JSON.stringify({ event: 'push' }),
+            headers: { 'Content-Type': 'application/json' },
+          })
+        )
+
+      it('text()', async () => {
+        const req = createReq()
+        await req.parseBody()
+        expect(await req.text()).toBe(JSON.stringify({ event: 'push' }))
+      })
+
+      it('json()', async () => {
+        const req = createReq()
+        await req.parseBody()
+        expect(await req.json()).toEqual({ event: 'push' })
+      })
+
+      it('arrayBuffer()', async () => {
+        const req = createReq()
+        await req.parseBody()
+        expect(await req.arrayBuffer()).toBeInstanceOf(ArrayBuffer)
+      })
+
+      it('blob()', async () => {
+        const req = createReq()
+        await req.parseBody()
+        expect(await req.blob()).toBeInstanceOf(Blob)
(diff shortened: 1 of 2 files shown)


yusukebe:
Hello @usualoma !

Can you review this?

yusukebe:
I think this is good! All tests passed. I'll merge.

PR #4808, opened 2026-03-19
feat(basic-auth): add passUsername option to support context injection (#4192)

### LINKED ISSUE #4192: Pass username to context from basicAuth

ramanverse:
This PR adds a `passUsername` option to the `basicAuth` middleware, allowing the authenticated username to be automatically passed to the context. This addresses a highly requested feature for accessing the username in subsequent route handlers without re-parsing headers.

### Changes:
- Added `passUsername?: boolean` to `BasicAuthOptions`.
- Implemented automatic context injection: `c.set('basicAuthUsername', username)`.
- Added module augmentation for `ContextVariableMap` to ensure type-safety.
- Updated JSDoc with examples.
- Added comprehensive tests in `src/middleware/basic-auth/index.test.ts`.

Resolves #4192

### DIFF
--- a/src/middleware/basic-auth/index.test.ts
+++ b/src/middleware/basic-auth/index.test.ts
@@ -511,4 +511,56 @@ describe('Basic Auth with passUsername', () => {
     expect(res.status).toBe(200)
     expect(await res.text()).toBe(username)
   })
+
+  it('should NOT pass username to context when auth fails and passUsername is true', async () => {
+    const app = new Hono()
+    let errorContext: any
+
+    app.use(
+      '/*',
+      basicAuth({
+        username,
+        password,
+        passUsername: true,
+      })
+    )
+    app.onError((err, c) => {
+      errorContext = c
+      return c.text('unauthorized', 401)
+    })
+    app.get('/', (c) => c.text('ok'))
+
+    const credential = Buffer.from('wrong:wrong').toString('base64')
+    await app.request('/', {
+      headers: { Authorization: `Basic ${credential}` },
+    })
+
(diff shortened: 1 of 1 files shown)


ramanverse:
It only sets the username after a successful match. If auth fails, it throws a 401 before it ever hits that code. I've just added a couple of tests to the PR to confirm this. Ready for review!

yusukebe:
This PR is unnecessary. You can use `onAuthSuccess` to set the username. The option is useless. Closing.

PR #4809, opened 2026-03-19
feat(service-worker): ensure fire() inherits default 404 fallback behavior (#4632)

### LINKED ISSUE #4632: The default behavior on 404 in handle() and fire() seems inconsistent and not documented

ramanverse:
This PR restores consistency to the `fire()` function in the Service Worker adapter. Currently, `fire()` disables the 404 network fallback by default, which contradicts its documentation and the behavior of the manual `handle()` function.

### Changes:
- Updated `fire()` in `src/adapter/service-worker/index.ts` to use the default `HandleOptions` (which enables 404 fallback).
- Improved JSDoc in `src/adapter/service-worker/handler.ts` to clearly document the fallback fetch behavior.
- Added unit tests to verify that `fire()` correctly triggers a fallback fetch when a route is missing.

Resolves #4632

### DIFF
--- a/src/adapter/service-worker/handler.test.ts
+++ b/src/adapter/service-worker/handler.test.ts
@@ -1,5 +1,5 @@
 import { Hono } from '../../hono'
-import { handle } from './handler'
+import { handle, fire } from '.'
 import type { FetchEvent } from './types'
 
 beforeAll(() => {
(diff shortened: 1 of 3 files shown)


ramanverse:
Hey @yusukebe can you review this and tell me any change needed!

yusukebe:
@ramanverse I know it. Please wait.

yusukebe:
I can't accept this PR. There are non-related changes. Closing.

PR #4810, opened 2026-03-19
fix(ssg): pass SSG_CONTEXT to forGetInfoURLRequest

yuintei:
# fix(ssg): pass SSG_CONTEXT to forGetInfoURLRequest

## Summary
Pass `SSG_CONTEXT` to `forGetInfoURLRequest` fetch (in `fetchRoutesContent`) so that `isSSGContext(c)` returns `true` and `disableSSG()` works as intended.
```ts
await pool.run(() => app.fetch(forGetInfoURLRequest, { [SSG_CONTEXT]: true })) // Pass SSG_CONTEXT
```

## Details
In `fetchRoutesContent`, `SSG_CONTEXT` is missing in the `forGetInfoURLRequest` fetch, while it is passed correctly in the `replacedUrlParam` fetch.

As a result, during the build process,
- `isSSGContext(c)` incorrectly returns `false`.
- `disableSSG()` falls back to `next()` instead of `c.notFound()`.
- Execution leaks into middlewares/handlers.
- In Cloudflare Workers projects, `c.env` is `undefined` during the Vite build, causing the following errors in every route:
```
TypeError: Cannot read properties of undefined (reading 'ADMIN_PASSWORD')
TypeError: Cannot read properties of undefined (reading 'd1')
```

I had to add manual guards like `if (!c.env) return next()` or `if (!c.env) return c.notFound()` to every `app.use()` and `app.get()` that accesses `c.env`, which is what `disableSSG()` should be handling automatically.

With this fix, I verified that the errors above are resolved and the manual guards are no longer needed. All existing tests pass with this change.

To reproduce,
1. Add a middleware that reads `c.env` (e.g. `basicAuth` with `c.env.ADMIN_PASSWORD`)
2. Add a route handler with `disableSSG()` that accesses `c.env` bindings (e.g. `c.env.d1`)
3. Run vite build with `@hono/vite-ssg` or `toSSG`

### DIFF
--- a/src/helper/ssg/ssg.test.tsx
+++ b/src/helper/ssg/ssg.test.tsx
@@ -647,6 +647,30 @@ describe('disableSSG/onlySSG middlewares', () => {
   })
 })
 
+describe('isSSGContext with disableSSG', () => {
+  it('Should work correctly when used together', async () => {
+    const app = new Hono()
+
+    app.use('*', async (c, next) => {
+      if (!isSSGContext(c)) {
+        return next()
+      }
+      await next()
+    })
+    app.get('/guarded', disableSSG(), (c) => c.html('<h1>should be skipped</h1>'))
+    app.get('/page', (c) => c.html('<h1>hello</h1>'))
+
+    const fsMock: FileSystemModule = {
+      writeFile: vi.fn(() => Promise.resolve()),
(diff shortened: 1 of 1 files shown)


yusukebe:
@yuintei Thanks!

PR #4813, opened 2026-03-19
fix(cors): reflect request origin when credentials is true with wildcard

ghost:
## Summary

Fixes the CORS middleware sending `Access-Control-Allow-Origin: *` together with `Access-Control-Allow-Credentials: true`, which browsers reject per the Fetch spec.

## Problem

When configured with `origin: '*'` and `credentials: true`:

```ts
cors({ origin: '*', credentials: true })
```

The middleware sets both:
- `Access-Control-Allow-Origin: *`
- `Access-Control-Allow-Credentials: true`

The [Fetch spec](https://fetch.spec.whatwg.org/#http-access-control-allow-origin) requires that `Access-Control-Allow-Origin` must not be `*` when credentials mode is `include`. Browsers reject these responses.

## Fix

When `credentials` is true and origin is `'*'`, reflect the actual request `Origin` header instead of the wildcard. Also set `Vary: Origin` since the response now varies by request.

This matches the behavior of other CORS libraries (e.g., Express cors middleware).

Fixes #4811

### DIFF
--- a/src/middleware/cors/index.test.ts
+++ b/src/middleware/cors/index.test.ts
@@ -397,4 +397,25 @@ describe('CORS by Middleware', () => {
     expect(res.headers.has('Access-Control-Allow-Origin')).toBeFalsy()
     expect(res.headers.get('Access-Control-Allow-Credentials')).toBe('true')
   })
+
+  it('Reflect different origins correctly for credentials with wildcard', async () => {
+    const res1 = await app.request('http://localhost/api10/abc', {
+      headers: { Origin: 'http://example.com' },
+    })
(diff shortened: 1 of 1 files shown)


ghost:
Added tests. Three cases covering the credentials + wildcard origin behavior:

1. GET request reflects the request `Origin` header instead of `*`, sets `Vary: Origin`
2. Preflight (OPTIONS) does the same
3. When no `Origin` header is present, `Access-Control-Allow-Origin` is omitted (nothing to reflect)

yusukebe:
@ctonneslan Thanks!

PR #4814, opened 2026-03-20
fix(cors): reflect request origin when origin:"*" is combined with credentials:true

guoyangzhen:
### Problem

The CORS middleware sets `Access-Control-Allow-Origin: *` and `Access-Control-Allow-Credentials: true` when configured with `origin: "*"` (default) and `credentials: true`. Per the [Fetch spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin#examples), browsers reject responses where `Access-Control-Allow-Origin` is `*` combined with credentials mode.

### Root Cause

In `middleware/cors/index.ts`, the `findAllowOrigin` function unconditionally returns `"*"` when `origin === "*"`, regardless of the `credentials` setting. When the Fetch spec requires credentials mode, the wildcard origin is invalid.

### Fix

When `origin === "*"` and `credentials === true`, the middleware now reflects the request `Origin` header back instead of `*`. When no Origin header is present (same-origin requests), it falls back to `*`. The `Vary: Origin` header is also set in both preflight and actual responses when this combination is used.

### Tests Added

- Reflects request origin when credentials + wildcard origin are combined
- Preflight reflects origin and sets Vary when credentials is true
- Falls back to `*` when no Origin header is present

### Verification

All 17 tests pass (14 existing + 3 new).

```sh
npx vitest run src/middleware/cors/index.test.ts
# ✓ 17 tests passed
```

### DIFF
--- a/src/middleware/cors/index.test.ts
+++ b/src/middleware/cors/index.test.ts
@@ -349,4 +349,45 @@ describe('CORS by Middleware', () => {
     expect(res2.headers.get('Access-Control-Allow-Origin')).toBe('*')
     expect(res2.headers.get('Access-Control-Allow-Methods')).toBe('GET,HEAD')
   })
+
+  describe('CORS with origin * and credentials', () => {
+    const app = new Hono()
+
+    app.use(
+      '/api/*',
+      cors({
+        credentials: true,
+      })
+    )
+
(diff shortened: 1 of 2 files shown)


yusukebe:
The issue was fixed by #4813

PR #4818, opened 2026-03-20
feat(cookie): add AES-256-GCM encrypted cookie support

### LINKED ISSUE #4817: feat: Add AES-256-GCM encrypted cookie support
### Summary

The current cookie helper only supports **signed cookies** (HMAC-SHA256), which provides integrity but not confidentiality — the cookie value remains readable in plaintext.

This request adds **encrypted cookie** support using **AES-256-GCM** via the Web Crypto API, providing both **confidentiality and integrity** for cookie values.

### Motivation

- Signed cookies (`setSignedCookie`) prove the value wasn't tampered with, but the value is still visible (e.g., `session=user-id-123.signature`).
- Encrypted cookies make the value opaque — useful for storing sensitive data like session tokens, user preferences, or any data that should not be readable client-side.
- AES-256-GCM is the standard choice: authenticated encryption with a single pass.

### Proposed API

Mirror the existing signed cookie API:

```ts
import {
  setEncryptedCookie,
  getEncryptedCookie,
  generateEncryptedCookie,
} from 'hono/cookie'

// Set
await setEncryptedCookie(c, 'session', 'sensitive-data', secret)

// Get (returns string | undefined | false)
const value = await getEncryptedCookie(c, secret, 'session')

// Generate without setting on response
const cookie = await generateEncryptedCookie('session', 'sensitive-data', secret)
```

### Design

- **Algorithm**: AES-256-GCM via `crypto.subtle` (multi-runtime: Node, Deno, Bun, CF Workers)
- **Key derivation**: HKDF-SHA256 from `string | BufferSource` secret (same type as signed cookies)
- **IV**: Random 12 bytes per encryption via `crypto.getRandomValues()`
- **AAD**: Cookie name bound as additional authenticated data (prevents copying encrypted values between cookie names)
- **Cookie format**: `base64(iv || ciphertext || authTag)`, URL-encoded
- **Prefix support**: Full `__Secure-` / `__Host-` prefix support
- **Return semantics**: `false` on decryption failure (tampered/wrong key), same as `getSignedCookie`

### Checklist

- [x] API mirrors existing signed cookie pattern
- [x] Uses Web Crypto API for multi-runtime compatibility
- [x] Supports cookie prefixes (`secure`, `host`)
- [x] All standard cookie options (domain, path, maxAge, expires, httpOnly, sameSite, etc.)

brunorodmoreira:
Closes #4817

## Summary

- Add `setEncryptedCookie`, `getEncryptedCookie`, `generateEncryptedCookie` helpers using AES-256-GCM encryption via Web Crypto API
- Add `serializeEncrypted`, `parseEncrypted` low-level utils with `EncryptedCookie` type
- API mirrors existing signed cookie pattern (`setSignedCookie` / `getSignedCookie`)

### Crypto design

| Concern | Choice |
|---|---|
| Algorithm | AES-256-GCM (`crypto.subtle`) |
| Key derivation | HKDF-SHA256 with app-specific salt |
| IV | Random 12 bytes per encryption |
| Auth tag | 128-bit (full strength) |
| AAD | Cookie name bound as additional data |
| Cookie format | `base64(iv \|\| ciphertext \|\| authTag)` |

- [x] Add tests
- [x] Run tests

## Test plan

- [x] Round-trip set/get encrypted cookie (single key and all cookies)
- [x] Missing cookie returns `undefined`
- [x] Tampered cookie returns `false`
- [x] Wrong secret returns `false`
- [x] Copying encrypted value to different cookie name returns `false` (AAD binding)
- [x] Secure and host prefix round-trips
- [x] Cookie options (domain, httpOnly, maxAge, expires, sameSite)
- [x] `generateEncryptedCookie` with and without options
- [x] All 29 existing cookie tests still pass

### DIFF
--- a/src/helper/cookie/index.test.ts
+++ b/src/helper/cookie/index.test.ts
@@ -526,6 +526,24 @@ describe('Cookie Middleware', () => {
         expect(await res.text()).toBe('INVALID')
       })
 
+      it('should return false when encrypted value is copied to a different cookie name', async () => {
+        const app2 = new Hono()
+        app2.get('/get', async (c) => {
+          const value = await getEncryptedCookie(c, secret, 'other_cookie')
+          return c.text(value === false ? 'INVALID' : 'UNEXPECTED')
+        })
+
+        const setRes = await app.request('http://localhost/set-encrypted-cookie')
+        const setCookieHeader = setRes.headers.get('Set-Cookie')
+        // take the encrypted value from secret_cookie and assign it to other_cookie
+        const encryptedValue = setCookieHeader!.split(';')[0].split('=').slice(1).join('=')
+
+        const getRes = await app2.request('http://localhost/get', {
+          headers: { Cookie: `other_cookie=${encryptedValue}` },
+        })
+        expect(await getRes.text()).toBe('INVALID')
+      })
+
       it('should return false for wrong secret', async () => {
         const app2 = new Hono()
         app2.get('/get', async (c) => {
--- a/src/helper/cookie/index.test.ts
+++ b/src/helper/cookie/index.test.ts
@@ -2,10 +2,13 @@ import { Hono } from '../../hono'
 import {
   deleteCookie,
   getCookie,
+  getEncryptedCookie,
   getSignedCookie,
   setCookie,
+  setEncryptedCookie,
   setSignedCookie,
   generateCookie,
+  generateEncryptedCookie,
   generateSignedCookie,
 } from '.'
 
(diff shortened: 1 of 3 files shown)


themavik (commented):
Binding AES-GCM `additionalData` to the cookie name in `encrypt`/`decrypt` is the right move. nit: `getEncryptionKey` runs on every `serializeEncrypted`/`parseEncrypted` — caching the derived `CryptoKey` per secret would cut work if apps decrypt many cookies per request.

travisbreaks (commented):
Crypto implementation in a framework that runs on Cloudflare Workers deserves a careful look. Some observations:

1. **IV uniqueness**: 12 random bytes from `crypto.getRandomValues()` is standard for AES-GCM and provides strong uniqueness guarantees. Good.

2. **HKDF salt**: Is the salt hardcoded or configurable? If hardcoded, two different applications using the same master key would derive the same encryption key. A configurable salt (or using the cookie name as salt input) would provide domain separation.

3. **AAD binding**: If the cookie name is used as AAD (Additional Authenticated Data), this correctly prevents cookie swapping attacks (moving an encrypted value from cookie A to cookie B). Worth explicitly calling this out in the docs/comments if it is the case.

4. **Workers shared isolate concern**: CryptoKey objects cached across requests in Workers share the same isolate. This is fine for encryption keys derived from a static secret, but worth confirming the cache key includes all relevant derivation parameters so different configurations don't accidentally share a CryptoKey.

5. **Timing side-channels**: The base64 encoding/decoding path should be constant-time or at least not leak information about the plaintext length beyond what the ciphertext length already reveals. Standard `btoa`/`atob` implementations are generally fine here.

6. **Error handling**: What happens if decryption fails (tampered ciphertext, wrong key)? AES-GCM will throw, and the behavior should be clearly documented: does it return `undefined`, throw, or return the raw encrypted value?

Solid implementation overall. The main thing I'd want to see addressed is the salt configurability and the decryption failure behavior.

PR #4819, opened 2026-03-21
feat(testing): add bufferResponse helper for async leak detection

andrew-t-james-wc:
## Summary

Adds an opt-in `bufferResponse` test helper to `hono/testing` that prevents async leak detection false positives in test runners like `vitest --detect-async-leaks`.

This follows up on #4802, addressing the feedback that the fix should be **on the test side, not in Hono's core**.

## Problem

Test runners with async leak detection flag the unconsumed `ReadableStream` body of Response objects created by `c.json()`, `c.text()`, etc. as leaked Promises. This affects any project using Hono + vitest with `--detect-async-leaks`.

## Solution

A new `bufferResponse` helper exported from `hono/testing` that users can opt into:

```ts
import { bufferResponse } from 'hono/testing'

const res = await bufferResponse(app.request('/api'))
expect(res.status).toBe(200)
```

The helper:
- Reads the response body into an `ArrayBuffer` (no pending `ReadableStream`)
- Creates a new `Response` with the buffered body
- Preserves `status`, `statusText`, and `headers`
- Handles edge cases: no body, already consumed body, non-Promise responses

## What changed

- `src/helper/testing/index.ts` — Added `bufferResponse` export
- `src/helper/testing/index.test.ts` — 5 new tests covering text, JSON, custom status/headers, empty body, and plain Response inputs

## Impact

- **Zero core changes** — test-side helper only
- All existing tests pass
- 100% coverage on new code

### DIFF
--- a/src/helper/testing/index.test.ts
+++ b/src/helper/testing/index.test.ts
@@ -1,5 +1,5 @@
 import { Hono } from '../../hono'
-import { testClient } from '.'
+import { testClient, bufferResponse } from '.'
 
 describe('hono testClient', () => {
   it('Should return the correct search result', async () => {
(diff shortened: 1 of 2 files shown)


yusukebe:
Hmmm. I think this should not be included in the `hono` package.

yusukebe:
The `bufferResponse` receives a standard `Response` and returns a `Promise<Response>`. This means this is not a Hono-specific function.  Of course, there are some non-Hono-specific functions in `hono`, but this method's use case is rare. I don't like this feature. Closing.

andrew-t-james-wc:
@yusukebe help me understand what you would prefer here, more than happy to make the changes but I can no longer run tests against hono because of these leaks, dont want to add hacks or workarounds.

yusukebe:
@andrew-t-james-wc You can create `bufferResponse()` in your code by yourself.

PR #4821, opened 2026-03-23
fix(service-worker): make `fire()` fallback behavior consistent with `handle()`

### LINKED ISSUE #4632: The default behavior on 404 in handle() and fire() seems inconsistent and not documented

yusukebe:
Fixes #4632

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/src/adapter/service-worker/index.ts
+++ b/src/adapter/service-worker/index.ts
@@ -27,9 +27,7 @@ import type { HandleOptions } from './handler'
  */
 const fire = <E extends Env, S extends Schema, BasePath extends string>(
   app: Hono<E, S, BasePath>,
-  options: HandleOptions = {
-    fetch: undefined,
-  }
+  options?: HandleOptions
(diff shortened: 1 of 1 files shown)


src/adapter/service-worker/index.ts
```diff
@@ -27,9 +27,7 @@ import type { HandleOptions } from './handler'
  */
 const fire = <E extends Env, S extends Schema, BasePath extends string>(
   app: Hono<E, S, BasePath>,
-  options: HandleOptions = {
-    fetch: undefined,
```
yusukebe, line 31: I can't remember why I made the default value of `fetch` `undefined`, or whether it was intended. But the JSDoc said `fire` means `addEventListener('fetch', handle(app, options))`. So to make it consistent, this PR's change is meaningful.

Issue #3809, opened 2025-01-07
Default logger middleware doesn't check that stdout is a TTY


gpanders:
### What version of Hono are you using?

4.6.16

### What runtime/platform is your app running on? (with version if possible)

Cloudflare Workers

### What steps can reproduce the bug?

Use the default logger middleware provided by Hono https://hono.dev/docs/middleware/builtin/logger

Observe that emitted logs contain ANSI escape sequences when the output is not a TTY:

![image](https://github.com/user-attachments/assets/a438d946-bad1-474a-8151-d8875db9ddd3)

This screenshot is taken from Cloudflare Workers logs, but the same would be true for any log output which is not a terminal.

### What is the expected behavior?

Hono should check that stdout is a TTY when determining whether or not to emit ANSI escape sequences for colors: https://github.com/honojs/hono/blob/2ead4d8faa58d187bf7ec74bac2160bab882eab0/src/utils/color.ts#L13

In Node this can be done with `Boolean(process.stdout.isTTY) === true` https://nodejs.org/api/tty.html#tty

Deno has `Deno.stdout.isTerminal()` https://docs.deno.com/api/deno/~/Deno.stdout.isTerminal

yusukebe:
HI @gpanders 

Thank you for raising the issue. There is a similar issue https://github.com/honojs/hono/issues/3751. We have [the hacky way](https://github.com/honojs/hono/issues/3751#issuecomment-2550192019), but using checking TTY is a good idea.

Hey @ryuapp! What do you think of checking TTY for the logger?

ryuapp:
Hi

> Hey @ryuapp! What do you think of checking TTY for the logger?

It seems like one good idea.
One concern is that some, such as GitHub Actions, don't have a TTY, which makes it difficult to control.
Then someone might ask for `FORCE_COLOR` and make things complicate.
Hono is generally not executed with GitHub Actions, so the benefits of checking TTY look great for me.

yusukebe:
Since version 4.8.0 includes #4094 (thank you, @ryuapp ), the logger looks at `NO_COLOR` in Cloudflare Workers environment variables. So, developers who want to disable coloring can set the value to `NO_COLOR` in the production environment.

It's good, but ideally, we want the logger to disable coloring without setting a variable, by checking the TTY. Hey @ryuapp, should we try to implement that feature? What do you think of this?

ryuapp:
Sorry for the late reply.

[...]

PR #4823, opened 2026-03-24
Implement TTY detection for Hono's logger middleware

Goldyvaiiii:
Currently, the logger middleware always applies ANSI color codes regardless of whether
the output destination is a terminal. This causes garbled color escape sequences in
logs when output is piped to files, CI systems, or other non-TTY destinations.

This PR adds cross-runtime TTY detection to getColorEnabled() in src/utils/color.ts:
- Node.js / Bun: checks process.stdout.isTTY
- Deno: checks Deno.stdout.isTerminal() (with legacy isatty() fallback)
- Serverless / Workers: defaults to true (unchanged behavior)

The NO_COLOR env variable continues to take highest precedence.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,15631 @@
+{
+  "name": "hono",
+  "version": "4.12.9",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "hono",
+      "version": "4.12.9",
+      "license": "MIT",
+      "devDependencies": {
+        "@hono/eslint-config": "^2.1.0",
+        "@hono/node-server": "^1.13.5",
+        "@types/glob": "^9.0.0",
+        "@types/jsdom": "^21.1.7",
+        "@types/node": "^24.3.0",
+        "@types/ws": "^8.18.1",
+        "@typescript/native-preview": "7.0.0-dev.20260210.1",
+        "@vitest/coverage-v8": "^3.2.4",
+        "arg": "^5.0.2",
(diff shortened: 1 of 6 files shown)


Goldyvaiiii:
Adds TTY detection to getColorEnabled() to avoid emitting ANSI colors in non-TTY environments (e.g., CI, piped logs). Supports Node/Bun (process.stdout.isTTY) and Deno (stdout.isTerminal() with fallback). Preserves existing serverless behavior. Updates tests and adds TTY coverage.

yusukebe:
Related to #3017 #3809

yusukebe:
Hey @ryuapp @NamesMT 

Is this a good time to think of this matter? If we start implementing TTY detection, I think using #3017 is better.

Goldyvaiiii:
Hey @yusukebe — thanks for the suggestion!

I've updated this PR to align with the centralized TTY detection approach from #3017.

- Added `src/utils/flags.ts` for environment-based detection
- Refactored `getColorEnabled()` to use these flags
- Ensured correct precedence for `NO_COLOR` and `FORCE_COLOR`
- Updated tests to evaluate the flags dynamically

All tests are passing locally. Please let me know if this direction looks good or if you'd like any further adjustments.

ryuapp:
Hi.
> Is this a good time to think of this matter? If we start implementing TTY detection, I think using https://github.com/honojs/hono/pull/3017 is better.

I agree with the approach #3017  for now. In most cases, it's better to prevent problems from occurring by default.

PR #4824, opened 2026-03-24
chore: update typescript to v6

Moshyfawn:
This PR updates TypeScript to v6 and adjusts config for the new defaults.

- Add explicit `types` fields (v6 defaults to `types: []`)
- Constrain `BufferSource` type to `ArrayBuffer`
- Drop `esModuleInterop` true by default

> [!NOTE]
> TS v6 benchmarks ran a bit slower, which is expected - v6 is mostly a preparations release with stricter checks ahead of v7.
>
> Benchmarks using `tsgo` and today's TS v7 release show roughly 2× lower memory usage and almost 3× faster type-checking on Mac M3.

### TS 7.0 readiness
- `tsc --stableTypeOrdering --noEmit` - passes with no issues
- `tsgo --noEmit` with TS _7.0.0-dev.20260324.1_ - passes with no issues (full type-check with the native Go compiler)
- `tsgo -p tsconfig.build.json` - passes with no issues (successfully generates `d.ts` files)
- `tsgo --build` - 4 errors in tests code regarding overload resolution

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/benchmarks/routers/tsconfig.json
+++ b/benchmarks/routers/tsconfig.json
@@ -1,7 +1,6 @@
 {
   "compilerOptions": {
     "allowImportingTsExtensions": true,
-    "esModuleInterop": true,
     "module": "NodeNext"
   },
   "include": ["./src"]
--- a/bun.lock
+++ b/bun.lock
@@ -1,6 +1,5 @@
 {
   "lockfileVersion": 1,
-  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "hono",
@@ -26,7 +25,7 @@
         "pkg-pr-new": "^0.0.53",
         "prettier": "3.7.4",
         "publint": "0.3.15",
-        "typescript": "^5.9.2",
+        "typescript": "^6.0.0",
         "undici": "^6.21.3",
         "vite-plugin-fastly-js-compute": "^0.4.2",
         "vitest": "^3.2.4",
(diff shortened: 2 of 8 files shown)


travisbreaks (commented):
TypeScript major version bumps in framework code need careful review. A few things:

1. **BufferSource narrowing to ArrayBuffer**: This is the most impactful change. Downstream consumers passing TypedArray or DataView to Hono APIs that accept BufferSource will break at the type level. Worth checking if any middleware or common patterns rely on passing Uint8Array directly (common in Workers/crypto code).

2. **esModuleInterop removal**: If downstream projects importing Hono have esModuleInterop enabled in their tsconfig (which many do as a default), this should be fine. But if Hono's own code relied on esModuleInterop for default imports from CJS modules, those imports would need to change to namespace imports. Worth verifying no CJS dependencies are imported with default-import syntax.

3. **TS7 pre-testing**: Thorough approach testing against tsgo. The 4 build errors in tests: are these known tsgo gaps or real issues that TS7 would surface?

4. **Peer dependency range**: Will the package.json typescript peer dep be updated to include v6? Users on v5 shouldn't be forced to upgrade until they're ready.

Good migration work. The author's diligence on pre-testing TS7 is noted.

hlobil (approved):

yusukebe:
The `typescript` is already v6.

PR #4828, opened 2026-03-26
fix(context): throw for non-serializable json responses

berketok-git:
## Summary
- make `c.json()` throw a `TypeError` when `JSON.stringify()` returns `undefined`
- avoid silently returning an empty string body for non-serializable values such as `undefined`
- add regression tests for `undefined` and preserve valid `null` serialization

Fixes #2343

## Test plan
- `bun x vitest run --config ./vitest.config.ts --project main src/context.test.ts`
- `bun x eslint src/context.ts src/context.test.ts`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/context.test.ts
+++ b/src/context.test.ts
@@ -70,6 +70,16 @@ describe('Context', () => {
     expect(res.headers.get('X-Custom')).toBe('Message')
   })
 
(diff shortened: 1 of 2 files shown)


yusukebe:
@berketok-git This issue will be handled by maintainers

PR #4831, opened 2026-03-27
fix(trie-router): treat ** wildcard the same as *

### LINKED ISSUE #4623: Inconsistent behavior of /** syntax in TrieRouter vs other routers
## Description

## Code sample

```typescript
import { Hono } from 'hono'

showRoutes(app)

serve({
  fetch: app.fetch,
  port: 3001
}, (info) => {
    if (res.ok){
      console.log("OK!")
    } else {
    }
  }
  )
})
```

Output:
```
POST  /auth/**
GET   /auth/**
TrieRouter
Error! 404
```

#### RegExpRouter

```typescript
```

Output:

```
POST  /auth/**
GET   /auth/**
RegExpRouter
OK!
```

#### PatternRouter

```typescript
```

Output:

```
POST  /auth/**
GET   /auth/**
PatternRouter
OK!
```

#### LinearRouter

```typescript
```

Output:

```
POST  /auth/**
GET   /auth/**
LinearRouter
OK!
```

```typescript
```

Output:

```
POST  /auth/*
GET   /auth/*
TrieRouter
OK!
```

## Proposal

```
TrieRouter:    404 Not Found
RegExpRouter:  200 OK
PatternRouter: 200 OK
LinearRouter:  200 OK
```

Ishiezz:
Fixes #4623

## Problem

TrieRouter returns 404 for routes defined with `/**` (double-star wildcard), while RegExpRouter, PatternRouter, and LinearRouter all match them correctly as equivalent to `/*`.

For example:
```ts
app.on(['POST', 'GET'], '/auth/**', (c) => c.json({}))
// GET /auth/hello → 404 ❌

### DIFF
--- a/src/router/trie-router/node.test.ts
+++ b/src/router/trie-router/node.test.ts
@@ -814,3 +814,73 @@ describe('Node with initial method and handler', () => {
     expect(res[0][0]).toEqual('hello handler')
   })
 })
+
+describe('Double star wildcard (**)', () => {
+  describe('Basic', () => {
+    const node = new Node()
+    node.insert('get', '/auth/**', 'auth handler')
+    it('/auth/hello', () => {
+      const [res] = node.search('get', '/auth/hello')
+      expect(res.length).toBe(1)
+      expect(res[0][0]).toEqual('auth handler')
+    })
+    it('/auth/hello/world', () => {
+      const [res] = node.search('get', '/auth/hello/world')
+      expect(res.length).toBe(1)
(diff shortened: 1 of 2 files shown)


usualoma:
Hi @Ishiezz 

As I mentioned in my comment below, I don’t believe that simply changing the behavior of TrieRouter is the best solution.
https://github.com/honojs/hono/issues/4623#issuecomment-3736714348

I’m sorry to say this, but I’m leaning toward rejecting this pull request.

yusukebe:
Let's close this.

Issue #4444, opened 2025-10-05
Calling `getCookie(c, 'key')` after `setCookie(c, 'key')` results in old data


NamesMT:
### What is the feature you are proposing?

Not sure if this should be considered a bug, but would be great if we have something that can get the latest data after `setCookie`, it would help a lot to implement things like session managers.

yusukebe:
Hi @NamesMT 

It's not a bug, it's designed so. This is a simple question. Isn't it enough to call `getCookie` after `setCookie` manually?

NamesMT:
Hi @yusukebe, it is not enough to just call `getCookie` after `setCookie` currently, i.e.:
```js
setCookie('k', 'val')
getCookie('k') // Is `undefined`, not `'val'`
```

yusukebe:
@NamesMT 

Ah, completely misunderstood. Yes. The feature you suggest should be implemented. Can you create a PR to do it?

IAmSSH:
@yusukebe Is it actually an expected behaviour? I feel the current implementation is correct, because `setCookie` sets the header attribute on the response headers, while `getCookie` tries to get cookie from the request, hence the `undefined` because it is not yet present on the request. If we try to hit the same endpoint that sets the cookie (and prints it) one more time, we'll see that `getCookie` will print the cookie value since it is now present in the request.

When we set a cookie, we should be setting it on the response, not on the request. Does it actually makes sense to set it also on the request?

NamesMT:
@IAmSSH Thats why I said not sure if this should be a feature request or bug, technically the current implementation is more correct when thinking as a request/response server, but DX-wise, its a bit not perfect and doesn't match client-side (frontend) cookies I/O.

I think, this issue (and a few related: #4445, #4446, #4447), if implemented, can probably be a new (set of) opt-in util(s), leaving the current utils which is highly optimized for lighter usecases / smaller servers intact.

Consider the issues "Request For Comments" for now.

PR #4832, opened 2026-03-27
fix(cookie): return updated cookie values after setCookie/setSignedCookie in same request

Ishiezz:
### Summary
This PR fixes an issue where `getCookie()` and `getSignedCookie()` would return stale data (from the incoming request headers) even after `setCookie()` or `setSignedCookie()` had been called during the current request cycle.

### Changes
- Modified `getCookie` and `getSignedCookie` to check the response's `Set-Cookie` headers for any newly set cookies.
- Implemented a `getSetCookieString` helper to extract and format cookies from response headers, ensuring they are correctly merged with request cookies.
- Response-set cookies now take precedence over request cookies, matching expected behavior in a single request-response cycle.
- Added comprehensive tests covering:
    - `getCookie` after `setCookie`
    - `getSignedCookie` after `setSignedCookie`
    - Overriding existing request cookies with new values
    - Merging multiple cookies from both request and response

### Verification Results
- Executed `bun run test src/helper/cookie/index.test.ts`
- Result: 36 tests passed (including all new validation cases).

### DIFF
--- a/src/helper/cookie/index.test.ts
+++ b/src/helper/cookie/index.test.ts
@@ -498,4 +498,121 @@ describe('Cookie Middleware', () => {
       )
     })
   })
+
+  describe('getCookie after setCookie', () => {
+    it('should return the newly set cookie value', async () => {
+      const app = new Hono()
+
+      app.get('/test', (c) => {
+        setCookie(c, 'session', 'abc123')
+        const value = getCookie(c, 'session')
+        return c.text(value || 'not found')
+      })
+
+      const res = await app.request('http://localhost/test')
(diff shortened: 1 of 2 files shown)


usualoma:
Hi @Ishiezz, thank you for creating the pull request.

This is the issue discussed below.

https://github.com/honojs/hono/issues/4444

Personally, I agree with the opinion below and believe that the current specification (where `getCookie` and `setCookie` are unrelated) is better for this issue.

https://github.com/honojs/hono/issues/4444#issuecomment-3424547391

In any case, I think sufficient discussion is necessary before implementation.

PR #4836, opened 2026-03-29
Fix/issue 1837

### LINKED ISSUE #1837: Capture Group Tests Appear to use non-capture group and are duplicate of non-capture group tests.
https://github.com/honojs/hono/blob/a1c7144d26bf1a13a2678f8ff031e4244e51d833/src/router/reg-exp-router/router.test.ts#L579-L601

Goldyvaiiii:
Fix #1837: Correct "Simple capturing group" test case

This PR updates the failing "Simple capturing group" test in common.case.test.ts to align with expected router behavior.
The issue was caused by incorrect test expectations rather than implementation logic.
Tests were executed using the proper Vitest configuration to resolve missing globals.
All test cases now pass successfully after the fix.
This ensures more accurate test coverage and prevents false failures in the suite.

### DIFF
--- a/src/router/common.case.test.ts
+++ b/src/router/common.case.test.ts
@@ -777,7 +777,7 @@ export const runTest = ({
     describe('Capture Group', () => {
       describe('Simple capturing group', () => {
         beforeEach(() => {
-          router.add('get', '/foo/:capture{(?:bar|baz)}', 'ok')
+          router.add('get', '/foo/:capture{(bar|baz)}', 'ok')
         })
 
         it('GET /foo/bar', () => {


Goldyvaiiii:
Hi @yusukebe 
I've fixed issue #1837 by correcting the "Simple capturing group" test case in common.case.test.ts.
The test expectations are now aligned with the actual behavior.
All tests are passing with the proper Vitest configuration.

Would appreciate your review when you get a chance. Thanks!

yusukebe:
Your PR has unnecessary changes. We can't accept it.

PR #4843, opened 2026-03-30
docs: fix impaired -> inspired typo in benchmark READMEs

Abhi3975:
Fixes a recurring typo where 'inspired' was misspelled as 'impaired' in the benchmarks documentation.


Abhi3975:
Thank you so much.

PR #4846, opened 2026-03-30
fix(context): return 'null' for c.json(undefined)

JBR-0100:
Calling `c.json(undefined)` currently returns an empty string (`""`). This PR ensures it returns the valid JSON literal `null` instead.

Closes #2343


yusukebe:
We, the maintainers, will handle the issue. Closing.

PR #4847, opened 2026-03-30
fix(jsx/dom): apply select value after children are rendered

### LINKED ISSUE #4837: [jsx/dom] Select input value not applied when options and value are set simultaneously

4.12.9

Bun (with Vite)

Create a component using `hono/jsx/dom` with a select element where both options and value are set in the same `useEffect`:

```tsx
import { useEffect, useState } from "hono/jsx";

export function App() {
  const [options, setOptions] = useState<string[]>([]);
  const [value, setValue] = useState<string>("");

  useEffect(() => {
    const newOptions = ["option1", "option2", "option3"];
    setOptions(newOptions);
    setValue(newOptions[1]); // Attempting to set value to "option2"
  }, []);

  return (
    <div>
      <h1>Select Input Race Condition Test</h1>
      <select value={value} onChange={(e) => setValue((e.target as HTMLSelectElement).value)}>
        {options.map((option) => (
          <option key={option} value={option}>
            {option}
          </option>
        ))}
      </select>
      <p>Selected value: {value}</p>
    </div>
  );
}
```

The select element should display "option2" as selected (the value set in state).

### What do you see instead?

The select element displays "option1" (the first option) as selected, even though the state value is "option2". The text below correctly shows "Selected value: option2", confirming the state is set correctly, but the select UI does not reflect this.

Image

### Additional information

The issue can be worked around by setting the value in a separate useEffect that depends on the options being populated:

```tsx
useEffect(() => {
  const newOptions = ["option1", "option2", "option3"];
  setOptions(newOptions);
}, []);

useEffect(() => {
  if (options.length > 0) {
    setValue(options[1]);
  }
}, [options]);
```

usualoma:
fixes #4837

### What's included in this fix

* 81e69f5db6d9375dc0e862a8349ae4e095ce6487 : Fix `select[value]` not being correctly applied when child options are specified
* 027a0ef21504fca1c58a28805607b3574d295384 : Fix `select[multiple]` incorrectly selecting the first option when it should remain unselected (not directly related to #4837, but included in this PR as a select-related fix)
* 4fb796edad279e0be5ce1513714df95255ce4a0d : Performance improvement

### Related Preact issues

https://github.com/preactjs/preact/issues/446
https://github.com/preactjs/preact/issues/761

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/jsx/dom/render.ts
+++ b/src/jsx/dom/render.ts
@@ -418,7 +418,7 @@ const applyNodeObject = (node: NodeObject, container: Container, isNew: boolean)
 
         applyProps(el as HTMLElement, child.props, child.pP)
         applyNodeObject(child, el as HTMLElement, isNewLocal)
-        if ((el as HTMLElement).nodeName === 'SELECT') {
+        if (child.tag === 'select') {
           applySelectValue(el as HTMLSelectElement, child.props)
         }
       }


usualoma:
Hi @yusukebe,

yusukebe:
@usualoma 

Nice to solve the difficult problem! Thanks.

PR #4848, opened 2026-03-31
fix(compress): convert strong ETag to weak ETag when compressing

### LINKED ISSUE #4845: compress() middleware should change etags to weak

4.12.9

n/a

```
import {compress} from 'hono/compress';

const app = new Hono()
  .use(compress())
  .get('/', async(c) => {
    c.header('ETag', '"strong-etag"');
    return c.text('<response long enough to trigger gzip compression>')
  });
```

The `compress` middleware is compressing a response which has an assigned `ETag`. Since it modifies the original response, it's no longer byte-to-byte equivalent. It's a functional equivalent.

For such cases, the response should switch to a __weak__ `ETag` instead.

In the example above, this header:
```
ETag: "strong-etag"
```
should be automatically changed into:
```
ETag: W/"strong-etag"
```

Otherwise it breaks range requests, since the response is no longer a byte-to-byte equivalent.

### What do you see instead?

The `compress` middleware incorrectly leaves the custom `ETag` as is, despite modifying the response.

### Additional information

_No response_

usualoma:
fixes #4845

Convert strong ETag to weak ETag (W/ prefix) when the compress middleware compresses the response.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/compress/index.test.ts
+++ b/src/middleware/compress/index.test.ts
@@ -193,6 +193,58 @@ describe('Compress Middleware', () => {
     })
   })
 
+  describe('ETag Handling', () => {
+    const app = new Hono()
+    app.use('*', compress())
+    app.get('/strong-etag', (c) => {
+      c.header('Content-Type', 'text/plain')
+      c.header('Content-Length', '1024')
+      c.header('ETag', '"strong-etag"')
+      return c.text('a'.repeat(1024))
+    })
+    app.get('/weak-etag', (c) => {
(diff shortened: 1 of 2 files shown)


usualoma:
@sebastian-nowak's suggestion makes sense, and I think we should address it.

usualoma:
Hi @yusukebe,

yusukebe:
@usualoma 

I see! The issue should be fixed. Thanks.

PR #4853, opened 2026-03-31
feat(middleware): allow overriding timeout in subroutes

ismailkarsli:
This PR allows users to override upstream route timeouts. This enables users to set a general, application-wide timeout while configuring specific exceptions for long-running routes.

### Key Changes
- Uses `context.get/set` to store `timeoutTimer`, ensuring any existing timer is cleared before a new one is created.
- Updates the middleware handler type to include `Variables.timeoutTimer`.
- Adds a test route and assertions to verify that the previous timeout is successfully overridden.

### Questions/comments
- I'm not sure about exposing the timer with variables. It allows users manipulate it but at the same time it can pollute variable space. Do I need to change it to an internal namespace?
- Do I need to fetch timeoutTimer from variables after next() function?

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/timeout/index.test.ts
+++ b/src/middleware/timeout/index.test.ts
@@ -18,6 +18,7 @@ describe('Timeout API', () => {
   const exception500: HTTPExceptionFunction = (context: Context) =>
     new HTTPException(500, { message: `Internal Server Error at ${context.req.path}` })
   app.use('/slow-endpoint/error', timeout(1200, exception500))
+  app.use('/slow-endpoint/replace', timeout(1300))
   app.use('/normal-endpoint', timeout(1000))
 
   app.get('/slow-endpoint', async (c) => {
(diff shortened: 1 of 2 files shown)


yusukebe:
Hey @ismailkarsli 

Thank you for the PR. I can understand what you want to do in rough terms, but can you provide use cases for this feature? What is the situation in which the user wants to modify the timer?

ismailkarsli:
Hello, thanks for the consideration @yusukebe.

For my use case, I wanted to add a generic timeout to all routes. For example, a safe timeout of 30 seconds for CRUD operations.

However, sometimes a deeply nested route might need to take a long time, and an exception might be required only for that route, as in the example below.

```ts
const app = new Hono();
app.use(timeout(30000)); // Set a generic timeout for all routes

app.use("/slightly-slow-route", timeout(60000), async (c, next) => {
	await someLongRunningFunction();
	return c.text("Completed within 60 seconds");
});

const user = new Hono();
app.use("/user", user);
// Allow up to 2 minutes for user export in a nested route.
user.get("/export", timeout(120000), async (c) => {
	await someLongRunningFunction();
	return c.text("Completed within 120 seconds");
});
```

So I don't want to sacrifice the convenience of `app.use(timeout(30000))` to create an exception only for a specific route. Therefore, I added the existing timeout timer to the context so that it can be removed/modified in a future middleware.

To be honest, I don't particularly like the timeout timer being shared in the context, but I'm not sure if there's another way to share variables with next middlewares.

ismailkarsli:
Stale

PR #4854, opened 2026-04-01
fix(utils/cookie): accept ':' in cookie names when parsing

### LINKED ISSUE #3189: Cookie not properly parsed. `validCookieNameRegEx` seems too strict.

4.5.1

Cloudflare workers

Create a middleware with the following code

```js
app.use(async (c, next) => {
  const allGetCookie = getCookie(c);
  console.log("getCookie", allGetCookie);
  console.log("allCookies", c.req.header("Cookie"));
  await next()
});
```

Send a curl request providing a cookie to your endpoint

```
curl --location --request POST 'http://localhost:8787/' \
--header 'Cookie: paraglide:lang=en;test=ok'
```

Your console should print
```
getCookie { test: 'OK' }
allCookies paraglide:lang=en;test=OK
```

getCookie should print `{'paraglide:lang': 'en', test: 'OK'}`
Cookie names containing `:` should be accepted.

### What do you see instead?

According to [this popular SO topic](https://stackoverflow.com/a/1969339), all alphanums plus ``!#$%&'()*+-./:<=>?@[]^_`{|}~`` characters should be considered OK though not entirely complying with rfc6265.

The most popular cookie parsing libraries don't enforce cookie name checking (ex: cookie, cookie-js...) and unfortunately, some external libraries or services can sometime set exotic cookie names 😢 ...which makes the use of `getCookie` not ideal since it can omit a lot of the cookies we see in the real world.

Don't you think we could also skip the validation of cookie name or be less strict about it?

### Additional information

If so, I'd be happy to submit a PR

pranav-nag:
## Problem                                                                                                                                                                 
                                                            
  `parse()` in `src/utils/cookie.ts` validates every cookie name against a regex based on strict RFC 6265 token definition, which excludes `:`. This silently drops any      
  cookie whose name contains a colon — for example `paraglide:lang` (set by @inlang/paraglide-sveltekit) and cookies issued by AWS Cognito. Users calling `getCookie(c,
  'paraglide:lang')` always received `undefined` even when the cookie was present in the request.                                                                            
                                                            
  ## Solution

  Introduced a separate `parseCookieNameRegEx` that extends the allowlist with `:`, used only in the `parse()` function (incoming cookie parsing). The fix follows the same  
  principle as the popular `jshttp/cookie` npm package: be strict when writing, permissive when reading.
                                                                                                                                                                             
  ## Changes                                                

  - Added `parseCookieNameRegEx` with `:` and escaped hyphen for clarity                                                                                                     
  - Updated `parse()` to use the new permissive regex
  - Added 3 new test cases covering colon names in regular and signed cookies                                                                                                
  - All 4230+ tests pass; 0 regressions                                                                                                                                      
   
  ## Reproduction                                                                                                                                                            
                                                            
  ```ts                                                                                                                                                                      
                                                                                                                                                                             
  app.get('/', (c) => {                                     
    // Request carries: Cookie: paraglide:lang=en
    const lang = getCookie(c, 'paraglide:lang') // was undefined, now 'en'                                                                                                   
    return c.text(lang ?? 'not found')                                                                                                                                       
  })                                                                                                                                                                         
 ```                                                                                                                                                                            
  Fixes #3189.

### DIFF
--- a/src/utils/cookie.test.ts
+++ b/src/utils/cookie.test.ts
@@ -73,6 +73,32 @@ describe('Parse cookie', () => {
     expect(cookie['']).toBeUndefined()
   })
 
+  it('Should parse cookie names containing colons', () => {
+    // Colons are used by real-world libraries such as @inlang/paraglide-sveltekit
+    // (e.g. 'paraglide:lang') and AWS Cognito. They are technically outside the
+    // strict RFC 6265 token definition but browsers send them without issue.
+    const cookieString = 'paraglide:lang=en; other_cookie=value; ns:sub:key=deep'
+    const cookie: Cookie = parse(cookieString)
+    expect(cookie['paraglide:lang']).toBe('en')
+    expect(cookie['other_cookie']).toBe('value')
+    expect(cookie['ns:sub:key']).toBe('deep')
+  })
+
+  it('Should parse a single cookie with a colon in its name when specified by name', () => {
+    const cookieString = 'paraglide:lang=en; other_cookie=value'
+    const cookie: Cookie = parse(cookieString, 'paraglide:lang')
+    expect(cookie['paraglide:lang']).toBe('en')
+    expect(cookie['other_cookie']).toBeUndefined()
+  })
+
+  it('Should parse signed cookies with colon in name', async () => {
+    const secret = 'test-secret'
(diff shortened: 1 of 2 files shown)


yusukebe:
This PR lacks the discussion of #3189. Closing.

PR #4862, opened 2026-04-06
feat(trailing-slash): add `skip` option

yusukebe:
This PR adds a `skip` option to Trailing Slash Middleware. This option is useful if the user wants to add the rule to skip the redirect depending on the path. For example, to skip the redirect if the path has an extension, you can write like this:

```ts
app.use(
  appendTrailingSlash({
    skip: (path) => /\.\w+$/.test(path),
  })
)
```

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/trailing-slash/index.test.ts
+++ b/src/middleware/trailing-slash/index.test.ts
@@ -255,6 +255,63 @@ describe('Resolve trailing slash', () => {
     })
   })
 
+  describe('appendTrailingSlash middleware with skip option', () => {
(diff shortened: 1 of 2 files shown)


yusukebe:
Hey @dai-shi ! What do you think of this? Can this satisfy your needs?

dai-shi (approved):
LGTM

yusukebe:
@dai-shi Thanks!

Hey @usualoma Can you review this? I think this feature is good.

PR #4871, opened 2026-04-11
fix(cache): log "caches is not defined" only once

yuintei:
The log is repeated for every `cache()` call when `globalThis.caches` is not available. This PR ensures the message is logged only once.

```bash
$ vite

  VITE v8.0.0  ready in 675 ms

  ➜  Local:   http://localhost:5173/
  ➜  Network: use --host to expose
  ➜  press h + enter to show help
Cache Middleware is not enabled because caches is not defined.
Cache Middleware is not enabled because caches is not defined.
Cache Middleware is not enabled because caches is not defined.
Cache Middleware is not enabled because caches is not defined.
Cache Middleware is not enabled because caches is not defined.
Cache Middleware is not enabled because caches is not defined.
Cache Middleware is not enabled because caches is not defined.
Cache Middleware is not enabled because caches is not defined.
(repeats...)
Using vars defined in .env
```

- [x] Add tests
- [x] Run tests
- [ ] Add [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code (N/A)

### DIFF
--- a/src/middleware/cache/index.test.ts
+++ b/src/middleware/cache/index.test.ts
@@ -403,18 +403,31 @@ describe('Cache Middleware', () => {
     expect(res.headers.get('cache-control')).not.toBe('max-age=10')
   })
 
-  it('Should not be enabled if caches is not defined', async () => {
+  it('Should not be enabled if caches is not defined and should log only once', async () => {
     vi.stubGlobal('caches', undefined)
+    const spy = vi.spyOn(console, 'log').mockImplementation(() => {})
+
     const app = new Hono()
     app.use(cache({ cacheName: 'my-app-v1', cacheControl: 'max-age=10' }))
+    cache({ cacheName: 'extra-1' })
+    cache({ cacheName: 'extra-2' })
     app.get('/', (c) => {
       return c.text('cached')
     })
+
     expect(caches).toBeUndefined()
     const res = await app.request('/')
     expect(res).not.toBeNull()
     expect(res.status).toBe(200)
     expect(res.headers.get('cache-control')).toBe(null)
+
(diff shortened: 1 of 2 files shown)


yusukebe:
Hi @yuintei 

This PR is not bad. BUT, the `cache` function is called only once in the initialization phase, so usually the log shows only one. So, in your environment, showing the log multiple times may be a little weird. Can you share the minimal project to reproduce it?

yuintei:
Hi @yusukebe, thanks for the review. Here's a minimal reproduction: [Link](https://github.com/yuintei/hono-cache-log-repro)

I use cache() on 20+ routes, so the log fills up the terminal on every dev server start, which is why I made this PR.

yusukebe:
@yuintei 

Thank you for the response and repro! I considered that declaring the variable with `let` and using it like globally is not a good design. I have another idea and will implement it later.

yusukebe:
Hi @yuintei 

I created PR #4876 using a different approach. In this way, the setting will be a bit verbose, but you can control the behavior in detail. And it's not necessary to define `cacheNotDefinedLogged` as a global condition. It prevents unintended side effects. What do you think of it?

yuintei:
@yusukebe 
I tested #4876 and it works as expected. The approach is great. Per-route control and custom logging are strict improvements over the current behavior.

Opt-in would be more convenient, but I understand it needs to be opt-out for backward compatibility. 
I'll just add `onCacheNotAvailable: false` to all my routes.

Thank you for working on this! Closing this PR.

I can open a docs PR on honojs/website once this is merged.

PR #4872, opened 2026-04-12
fix(aws-lambda): infer response type from event via overloads

Danny-Devs:
## Summary

While reviewing the AWS Lambda adapter, I noticed `handle()` used a conditional return type guarded by `@ts-expect-error FIXME: Fix return typing`. Tracing the FIXME back to commit 0e4dfe58 (#3883), the author's goal was to *"Infer response type from event"* — narrowing to `WithHeaders` for events that cannot produce multi-value headers.

## The issue

The conditional `L extends { multiValueHeaders: Record<string, string[]> } ? WithMultiValueHeaders : WithHeaders` is vacuous. Every `LambdaEvent` member declares `multiValueHeaders` as optional (v1, ALB) or omits/undefines it (v2 which has `?: undefined`, Lattice which has no such field), so none satisfy the required-field extends constraint. The conditional always resolves to `WithHeaders` regardless of the input event type, and the `@ts-expect-error` on the implementation hides the mismatch between the declared return type and the actual value produced by `processor.createResult` (which at runtime may return either branch, based on `'multiValueHeaders' in event && event.multiValueHeaders`).

## The fix

Replace the conditional with per-event overloads that encode the real runtime guarantee. The overloads are ordered for resolution:

1. **v2 / Lattice** → `APIGatewayProxyResult & WithHeaders` — these event types never produce multi-value headers
2. **Any event with `multiValueHeaders` present** → `APIGatewayProxyResult & WithMultiValueHeaders` — the runtime uses multi-value headers when the event carries them
3. **Any event with `multiValueHeaders` absent/undefined** → `APIGatewayProxyResult & WithHeaders` — the runtime uses regular headers otherwise
4. **Generic `LambdaEvent` fallback** → the full `APIGatewayProxyResult` union — for callers whose event type is not statically known

This models the exact runtime behavior of `getProcessor` and `createResult`, so callers get the most precise return type the type system can express from their call-site argument.

The overload pattern matches `HonoRequest.param` in `src/request.ts:94-104`. The `as LambdaHandlerFunction` cast is the idiomatic TypeScript pattern for giving a `const` arrow function multiple call signatures — arrow functions cannot use stacked `function` declaration overloads.

No change in runtime behavior. The `@ts-expect-error FIXME: Fix return typing` directive is removed.

## Verification

- `bun tsc --build` — clean (the runtime-tests/lambda test file compiles without changes, since the overloads are precise enough for TypeScript to infer the correct return type from each test's inline event literal)
- `bun run test` — 4255 passed, 33 skipped, 0 failed
- `bun run build` — green (includes declaration emission and publint)
- `bun run format:fix && bun run lint:fix` — no changes needed
- Type-level narrowing also verified locally with an `@ts-expect-error`-style test file that imported the real `handle` from this branch and asserted the narrowing contract for all four event types (v2/Lattice narrow to `WithHeaders` without a discriminant; v1/ALB correctly require one when multiValueHeaders presence is unknown). Happy to contribute that as a follow-up if the team would like a regression guard in the test suite — kept this PR focused on the fix itself.

- [ ] Add tests (type-level verification done locally; see note above)
- [x] Run tests
- [ ] Add [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code (no new public API surface)

### DIFF
--- a/src/adapter/aws-lambda/handler.ts
+++ b/src/adapter/aws-lambda/handler.ts
@@ -197,16 +197,23 @@ type HandleOptions = {
 }
 
 type LambdaHandlerFunction = {
-  (event: APIGatewayProxyEvent, lambdaContext?: LambdaContext): Promise<APIGatewayProxyResult>
   (
     event: APIGatewayProxyEventV2,
     lambdaContext?: LambdaContext
   ): Promise<APIGatewayProxyResult & WithHeaders>
-  (event: ALBProxyEvent, lambdaContext?: LambdaContext): Promise<APIGatewayProxyResult>
   (
     event: LatticeProxyEventV2,
     lambdaContext?: LambdaContext
   ): Promise<APIGatewayProxyResult & WithHeaders>
+  (
+    event: LambdaEvent & { multiValueHeaders: Record<string, string[]> },
+    lambdaContext?: LambdaContext
+  ): Promise<APIGatewayProxyResult & WithMultiValueHeaders>
+  (
+    event: LambdaEvent & { multiValueHeaders?: undefined },
+    lambdaContext?: LambdaContext
+  ): Promise<APIGatewayProxyResult & WithHeaders>
+  (event: LambdaEvent, lambdaContext?: LambdaContext): Promise<APIGatewayProxyResult>
 }
 
 /**
--- a/src/adapter/aws-lambda/handler.ts
+++ b/src/adapter/aws-lambda/handler.ts
@@ -196,6 +196,19 @@ type HandleOptions = {
   isContentTypeBinary: ((contentType: string) => boolean) | undefined
 }
 
+type LambdaHandlerFunction = {
+  (event: APIGatewayProxyEvent, lambdaContext?: LambdaContext): Promise<APIGatewayProxyResult>
+  (
+    event: APIGatewayProxyEventV2,
+    lambdaContext?: LambdaContext
+  ): Promise<APIGatewayProxyResult & WithHeaders>
+  (event: ALBProxyEvent, lambdaContext?: LambdaContext): Promise<APIGatewayProxyResult>
+  (
+    event: LatticeProxyEventV2,
+    lambdaContext?: LambdaContext
+  ): Promise<APIGatewayProxyResult & WithHeaders>
+}
+
 /**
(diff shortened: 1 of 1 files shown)


Danny-Devs:
The "Type & Bundle size Check on PR" failed on the first push — `tsc --build` caught 12 `TS18048` errors in `runtime-tests/lambda/index.test.ts` where the test code accessed `.headers` or `.multiValueHeaders` without a check.

This was the original overloads being *correct but imprecise*: v1/ALB returned the full `APIGatewayProxyResult` union, so TypeScript correctly flagged `.headers` as possibly `undefined`. The test file was silently relying on the old (vacuous) type that always resolved to `WithHeaders`.

Rather than adding non-null assertions to the test file, I refined the overloads to also narrow based on `multiValueHeaders` presence in the event — this models the exact runtime behavior of `getProcessor`/`createResult`, so all 12 errors resolve without any test changes. Updated the PR description with details.

yusukebe:
@Danny-Devs 

It's good to remove `FIXME: Fix return typing`, but is there any use case to use the type returned from `handle`?

Danny-Devs:
Hi Yusukebe — sorry for the delay in responding, I didn't see your message until now.

I went and checked: 30+ real-world consumers of `hono/aws-lambda` on GitHub (SST, openauth, bknd, saleor, vikejs/bati, and app-level users). All use `export const handler = handle(app)` with AWS as the consumer. Nobody narrows on the return type externally. Inside the repo, only `runtime-tests/lambda/index.test.ts` does.

So the overloads fulfill what #3883 was trying to express, but nobody outside the test suite actually exercises that narrowing. `streamHandle` returns `Handler` with no narrowing, which looks like it may be the right precedent.

Happy to collapse this to `Promise<APIGatewayProxyResult>` — kills the FIXME, matches `streamHandle`, 3 lines instead of 26. What do you think?

yusukebe:
Since the return type of handle is only used internally, I don't think it's worth defining a precise type just to suppress the @ts-expect-error. There are other places in Hono that take the same approach. I'll pass on this PR. Thanks.

Danny-Devs:
Hi @yusukebe — no worries at all. Thanks for reviewing. I'll come back with something more useful next time!

PR #4873, opened 2026-04-12
Throw on non-serializable JSON responses

ShipItAndPray:
## Summary
- make `c.json(undefined)` throw a consistent `TypeError`
- preserve the existing JSON response behavior for all serializable values
- add a regression test covering the undefined case

Fixes #2343

## Validation
- bunx vitest run src/context.test.ts
- bunx eslint src/context.ts src/context.test.ts

## Notes
- `bun install` was required in this checkout to restore the local vitest dependency tree

### DIFF
--- a/src/context.test.ts
+++ b/src/context.test.ts
@@ -70,6 +70,12 @@ describe('Context', () => {
     expect(res.headers.get('X-Custom')).toBe('Message')
   })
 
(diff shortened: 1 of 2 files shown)


yusukebe:
This issue will be handled by the maintainers. Closing.

PR #4874, opened 2026-04-12
Treat compressed lambda-edge responses as binary

### LINKED ISSUE #4254: hono/lambda-edge adapter should treat compressed response body as binary
https://github.com/honojs/hono/blob/5149aa421bef92d3a495287f2f67bca4382bad39/src/adapter/lambda-edge/handler.ts#L140-L141

When I use the `hono/lambda-edge` adapter with the `hono/compress` middleware, no response body is returned. This is because the gzipped response isn't being Base64-encoded by the adapter.

The above lines of code should look at the `Content-Encoding` header to detect a compressed response.

ShipItAndPray:
## Summary
- treat non-identity `Content-Encoding` values as binary in the lambda-edge adapter
- base64-encode compressed responses so CloudFront receives valid payloads
- add a regression test that exercises the compress middleware through the adapter

Fixes #4254

## Validation
- bunx vitest run src/adapter/lambda-edge/handler.test.ts
- bunx eslint src/adapter/lambda-edge/handler.ts src/adapter/lambda-edge/handler.test.ts
- git diff --check

### DIFF
--- a/src/adapter/lambda-edge/handler.test.ts
+++ b/src/adapter/lambda-edge/handler.test.ts
@@ -1,9 +1,10 @@
-import { describe } from 'vitest'
+import { describe, expect, it, vi } from 'vitest'
 import { setCookie } from '../../helper/cookie'
 import { Hono } from '../../hono'
-import { encodeBase64 } from '../../utils/encode'
+import { compress } from '../../middleware/compress'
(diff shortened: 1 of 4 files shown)


yusukebe:
This PR includes not related changes. Slop PR. Closing.

PR #4875, opened 2026-04-12
fix(logger): disable ANSI when stdout is not a tty

### LINKED ISSUE #3809: Default logger middleware doesn't check that stdout is a TTY

ShipItAndPray:
Fixes #3809.

Summary:
- avoid ANSI escape sequences when stdout is not a TTY
- keep colored logs when stdout is a TTY
- add regression coverage for the non-TTY case

Validation:
- env -u NO_COLOR bunx vitest run src/middleware/logger/index.test.ts src/utils/color.test.ts
- bunx eslint src/middleware/logger/index.ts src/middleware/logger/index.test.ts
- git diff --check

### DIFF
--- a/src/adapter/lambda-edge/handler.test.ts
+++ b/src/adapter/lambda-edge/handler.test.ts
@@ -1,9 +1,10 @@
-import { describe } from 'vitest'
+import { describe, expect, it, vi } from 'vitest'
 import { setCookie } from '../../helper/cookie'
 import { Hono } from '../../hono'
-import { encodeBase64 } from '../../utils/encode'
+import { compress } from '../../middleware/compress'
+import { decodeBase64, encodeBase64 } from '../../utils/encode'
 import type { Callback, CloudFrontEdgeEvent } from './handler'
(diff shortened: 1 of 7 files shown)


yusukebe:
This is slop. Closing.

PR #4887, opened 2026-04-22
feat(cacherouter) add routing cache

bulebrainbrand:
## Overview
Add CacheRouter for cache routing Result.
- CacheRouter take router instance  on constructer
- this is LRU

## changes
- made cache-router directory and router.ts, index.ts, router.test.ts files.

## impact
- None

## Tests
- src/router/common.case.test.ts

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/router/cache-router/index.ts
+++ b/src/router/cache-router/index.ts
@@ -1,0 +1,6 @@
+/**
+ * @module
+ * CacheRouter for Hono.
+ */
+
+export { CacheRouter } from './router'
--- a/src/router/cache-router/router.test.ts
+++ b/src/router/cache-router/router.test.ts
@@ -1,0 +1,14 @@
+import { runTest } from '../common.case.test'
+import { TrieRouter } from '../trie-router'
+import { CacheRouter } from './router'
+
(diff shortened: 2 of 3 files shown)


usualoma:
Hi @bulebrainbrand,
You’re right—wrapping TrieRouter would likely make it faster.

However, since the routers currently built into Hono itself can handle a wide range of use cases and are sufficiently fast, I don’t think there’s a need to add a router that requires maintenance to the core Hono code.

I think it would be best to release it as a third-party router separately, rather than including it in the core.

yusukebe:
@bulebrainbrand 

Thanks!

I have also considered using a caching mechanism in routers. But I don't introduce it because Hono is basically made for "edge" environments, such as Cloudflare Workers, where the process will die with one of a few requests. In the environment, caching is not effective.

For the reason @usualoma mentioned, we don't have to include it in the `hono` core. I think it's a good idea to create an independent project.

bulebrainbrand:
thank you for valuable replies.
that makes sense. i close pullrequest and make 3rd party package. thank you!

PR #4891, opened 2026-04-23
feat: add built-in rate limiter middleware

codebybilal18:
## Summary

Adds a built-in `rate-limiter` middleware to Hono. Rate limiting is one of the most commonly needed features in any production web application, and until now users had to rely on third-party packages or write their own. This contribution brings it in as a first-class, zero-dependency middleware consistent with Hono's existing patterns.

## Why

Every public-facing API needs protection against abuse, brute-force attacks, and accidental overload. Without a built-in rate limiter, Hono users must:
- Wire up third-party npm packages with varying APIs
- Write their own store + counter logic
- Handle response headers manually

This middleware follows the same design language as `body-limit`, `timeout`, and `ip-restriction` — simple defaults, escape hatches for power users.

### New files
- `src/middleware/rate-limiter/index.ts` — middleware implementation
- `src/middleware/rate-limiter/index.test.ts` — 17 tests (all passing)

### Updated files
- `package.json` — added `./rate-limiter` export
- `jsr.json` — added `./rate-limiter` export for Deno/JSR

### Implementation details

- **Algorithm**: Fixed-window counter — simple, predictable, low overhead
- **Default store**: `MemoryStore` (in-process `Map`) — works out of the box for single-instance deployments
- **Pluggable store**: Implement the `RateLimitStore` interface to plug in Redis, Cloudflare KV, Durable Objects, etc. for multi-instance deployments
- **Standard headers**: Sets `RateLimit-Limit`, `RateLimit-Remaining`, `RateLimit-Reset`, and `RateLimit-Policy` per the IETF RateLimit header fields draft
- **Context variable**: Exposes `c.get('rateLimit')` typed via `ContextVariableMap` augmentation — downstream handlers can inspect limit state
- **Custom key generator**: Defaults to the first IP in `X-Forwarded-For`; can be any sync or async function
- **Custom error handler**: Override the default `429 Too Many Requests` response

### Basic

```ts
import { Hono } from 'hono'
import { rateLimiter } from 'hono/rate-limiter'

const app = new Hono()

app.use(rateLimiter({
  windowMs: 60_000, // 1 minute
  limit: 100,       // 100 requests per minute per IP
}))

app.get('/', (c) => c.text('Hello!'))


rafaell-lycan:
Hey @codebybilal18 👋

FYI there's a lib for that already called [hono-rate-limiter](https://github.com/rhinobase/hono-rate-limiter) which works fine.

My 2 cents are: I'm not 100% sure if this middleware should live inside of Hono's core, I'd consider moving it to the middleware repository instead which can allow you evolve in a different way, with the in-memory solution at first and later adding [adapters](https://express-rate-limit.mintlify.app/reference/stores) for other stores like Redis, etc.

codebybilal18:
> Hey @codebybilal18 👋
> 
> FYI there's a lib for that already called [hono-rate-limiter](https://github.com/rhinobase/hono-rate-limiter) which works fine.
> 
> My 2 cents are: I'm not 100% sure if this middleware should live inside of Hono's core, I'd consider moving it to the middleware repository instead which can allow you evolve in a different way, with the in-memory solution at first and later adding [adapters](https://express-rate-limit.mintlify.app/reference/stores) for other stores like Redis, etc.

Hey @rafaell-lycan, thanks for your feedback. 

I got your points. But I think rate limiting is foundational enough to belong in core (like `body-limit` and `timeout` already there), but I'm open to the middleware repo if that's preferred. The pluggable store interface means users can bring their own Redis/KV logic anyway. Curious what the maintainers think!

yusukebe:
Hi @codebybilal18 

As @rafaell-lycan, we already have [hono-rate-limiter](https://github.com/rhinobase/hono-rate-limiter). Plus, we actually have many built-in middleware now, and maintenance is a mess. Honestly, I don't want to add middleware if there's already a feature that does the same. So, we can't accept this PR. Thanks.

PR #4893, opened 2026-04-24
fix(jsx): normalize SVG attributes on the <svg> root element

kfly8:
### Summary

`<svg strokeWidth="1.5">` was output as-is by SSR, but the DOM renderer
converted it to `stroke-width`. This broke icon libraries that set stroke/fill
attributes on the `<svg>` root, and it caused SSR and DOM to produce different
HTML for the same JSX (hydration mismatch).

After this PR, both SSR and DOM output `<svg stroke-width="1.5" ...>`.

### Reproduction

```tsx
app.get('/', (c) => {

  const icon = (
    <svg stroke="currentColor" strokeWidth="1.5" viewBox="0 0 32 32">
      <path d="M1 1L8 8" />
    </svg>
  )

  return c.render(<span>{icon}</span>)
})
```

###  Investigation

- Icon libraries put presentation attributes on the <svg> root so child paths can inherit them.
  -  Example from Heroicons:
https://github.com/tailwindlabs/heroicons/blob/master/optimized/24/outline/pencil-square.svg
- React also converts strokeWidth → stroke-width on the root
  -  checked with react-dom/server@18.3.1 renderToStaticMarkup
- HTML spec: when the parser sees a <svg> start tag, it inserts "a foreign element for the token, with SVG namespace". The <svg> element
itself is in the SVG namespace, not only its children:
https://html.spec.whatwg.org/multipage/parsing.html#parsing-main-inbody:adjust-svg-attributes

###  Root cause

The factory put only the children of <svg> inside the namespace provider:  `svg > Provider > children`

When SSR rendered the <svg>'s own attributes, useContext(nameSpaceContext) returned the default "" instead of "svg", so toSVGAttributeName was skipped:

https://github.com/honojs/hono/blob/f774f8df49e7ec7e205f15c5076a37132c515ebf/src/jsx/base.ts#L179-L182

###  Fix

Put the <svg> element itself inside the provider too: `Provider > svg > children`

---

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/jsx/base.test.tsx
+++ b/src/jsx/base.test.tsx
@@ -34,3 +34,13 @@ describe('cloneElement', () => {
     expect(element.toString()).toBe('<hr/>')
   })
 })
+
+describe('createElement', () => {
+  it('should preserve the SVG element shape', () => {
+    const ref = { current: null }
+    const element = jsx('svg', { ref }) as unknown as JSXNode
+    expect(element.tag).toBe('svg')
+    expect(element.type).toBe('svg')
+    expect(element.ref).toBe(ref)
+  })
+})
--- a/src/jsx/base.ts
+++ b/src/jsx/base.ts
@@ -335,7 +335,20 @@ export const jsxFn = (
       props,
       children
     )
-  } else if (tag === 'svg' || tag === 'head') {
+  } else if (tag === 'svg') {
+    nameSpaceContext ||= createContext('')
+    return new JSXFunctionNode(
+      nameSpaceContext,
+      {
+        value: tag,
+        // The `<svg>` element itself belongs to the SVG namespace it
+        // establishes, so its own attributes must render inside the provider.
+        // Otherwise SSR skips kebab-case normalization on e.g. `<svg strokeWidth>`.
+        children: new JSXNode(tag, props, children),
+      },
+      []
+    )
+  } else if (tag === 'head') {
     nameSpaceContext ||= createContext('')
     return new JSXNode(tag, props, [
       new JSXFunctionNode(
--- a/src/jsx/base.ts
+++ b/src/jsx/base.ts
@@ -177,7 +177,7 @@ export class JSXNode implements HtmlEscaped {
     buffer[0] += `<${tag}`
 
     const normalizeKey: (key: string) => string =
-      nameSpaceContext && useContext(nameSpaceContext) === 'svg'
+      tag === 'svg' || (nameSpaceContext && useContext(nameSpaceContext) === 'svg')
         ? (key) => toSVGAttributeName(normalizeIntrinsicElementKey(key))
         : (key) => normalizeIntrinsicElementKey(key)
     for (let [key, v] of Object.entries(props)) {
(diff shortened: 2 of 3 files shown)


usualoma:
Hi @kfly8 
Thank you for creating the pull request! I understand the situation.
Please give me a little time while I figure out how to fix this.

kfly8:
Thank you!
Please let me know if there is anything I should do to help you with your review.

usualoma:
Hi @kfly8 

I apologize for the delayed response; I had some errands to run over the long weekend.
I understand the issue.

133c914c31f2239ce468d242192e28cc0636314a is a valid approach, but with the current implementation, `jsx(‘svg’, ...)` returns a `JSXFunctionNode` for `nameSpaceContext` instead of a `JSXNode` whose tag is `svg`.

This can break compatibility with the existing JSX node structure. For example, `element.tag` and `element.type` are no longer `‘svg’`, and `element.ref` can no longer be read from the `<svg>` props. It may also affect cases like `cloneElement(<svg ...>, props)`, where the additional props would be applied to the Provider node instead of the actual `<svg>` element.

So I’d prefer to keep the `<svg>` node structure unchanged, and instead treat the `<svg>` root element as SVG during attribute normalization at render time.

Could you consider implementing a change like this?

```diff
diff --git c/src/jsx/base.test.tsx w/src/jsx/base.test.tsx
index e12a424e..27ab7a59 100644
--- c/src/jsx/base.test.tsx
+++ w/src/jsx/base.test.tsx
@@ -34,3 +34,13 @@ describe('cloneElement', () => {
     expect(element.toString()).toBe('<hr/>')
   })
 })
+
+describe('createElement', () => {
+  it('should preserve the SVG element shape', () => {
+    const ref = { current: null }
+    const element = jsx('svg', { ref })
+    expect(element.tag).toBe('svg')
+    expect(element.type).toBe('svg')
+    expect(element.ref).toBe(ref)
+  })
+})
diff --git c/src/jsx/base.ts w/src/jsx/base.ts
index d2156d68..65ba04d7 100644
--- c/src/jsx/base.ts
+++ w/src/jsx/base.ts
@@ -177,7 +177,7 @@ export class JSXNode implements HtmlEscaped {
     buffer[0] += `<${tag}`
 
     const normalizeKey: (key: string) => string =
-      nameSpaceContext && useContext(nameSpaceContext) === 'svg'
+      tag === 'svg' || (nameSpaceContext && useContext(nameSpaceContext) === 'svg')
         ? (key) => toSVGAttributeName(normalizeIntrinsicElementKey(key))
         : (key) => normalizeIntrinsicElementKey(key)
     for (let [key, v] of Object.entries(props)) {
@@ -335,20 +335,7 @@ export const jsxFn = (
       props,
       children
     )
-  } else if (tag === 'svg') {
-    nameSpaceContext ||= createContext('')
-    return new JSXFunctionNode(
-      nameSpaceContext,
-      {
-        value: tag,
-        // The `<svg>` element itself belongs to the SVG namespace it
-        // establishes, so its own attributes must render inside the provider.
-        // Otherwise SSR skips kebab-case normalization on e.g. `<svg strokeWidth>`.
-        children: new JSXNode(tag, props, children),
-      },
-      []
-    )
-  } else if (tag === 'head') {
+  } else if (tag === 'svg' || tag === 'head') {
     nameSpaceContext ||= createContext('')
     return new JSXNode(tag, props, [
       new JSXFunctionNode(
```

kfly8:
@usualoma Thank you for catching that!

I checked and confirmed that it does break compatibility with cloneElement and the like. 
I've applied the fix as suggested.

usualoma:
Thanks for the update! I think it looks good.

Though, well, it doesn’t really matter either way, but is it appropriate to list Claude Opus as a Co-Authored-By for f1004ed? (It’s fine if my name isn’t included, but seeing Claude Opus listed there feels a bit off. After all, Opus didn’t write that code, did he?)

kfly8:
@usualoma I am very sorry, please excuse me. I have added usualoma as a co-author.
It is a fact that I am using Opus, so I will leave that in as well.

Side note: Actually, the idea you proposed was one that also came up when I was working on it myself, but I rejected it at the time because the branching for the SVG tag felt messy.
Even though usualoma suggested it, I ended up committing it based on my own arbitrary context.
I am sorry. Thank you for pointing this out.

usualoma:
@kfly8 Thank you for the explanation. I understand that it was added because of your sincere attitude!

kfly8:
Thank you for your understanding. I look forward to continuing to work with you.

yusukebe (approved):
LGTM!

yusukebe:
@kfly8 @usualoma Thank you both!

PR #4894, opened 2026-04-24
fix(method-override): handle Content-Type with charset parameter

### LINKED ISSUE #4892: method-override: form-based override fails when Content-Type includes charset parameter

4.12.15

The bug is runtime-agnostic, but I encountered it on: Cloudflare Workers

When a request to a `method-override` middleware has a `Content-Type` header with a charset parameter (e.g. `application/x-www-form-urlencoded;charset=UTF-8`), the method override does **not** take effect and the request is handled as the original POST.

This is the Content-Type that the Fetch API automatically sets when a `URLSearchParams` instance is passed as the request body. It is also the default for libraries such as [Hotwired Turbo](https://turbo.hotwired.dev/), which is how I originally hit this.

### Minimal reproduction

```ts
import { Hono } from 'hono'
import { methodOverride } from 'hono/method-override'

const app = new Hono()
app.use('/posts', methodOverride({ app }))
app.delete('/posts', (c) => c.text('deleted'))

const body = new URLSearchParams()
body.append('_method', 'DELETE')

// Note: no explicit Content-Type header — fetch/Request sets
// `application/x-www-form-urlencoded;charset=UTF-8` automatically
// when the body is a URLSearchParams instance.
const res = await app.request('/posts', {
  method: 'POST',
  body,
})

console.log(res.status) // 404 (expected 200)
```

The form-based override should fire for any `Content-Type` whose media type is `application/x-www-form-urlencoded`, regardless of parameters like `charset`. The request above should be routed to the DELETE handler and return 200.

### What do you see instead?

The override does not fire. The request is handled as the original POST. In the repro above, since there is no POST handler for `/posts`, the request falls through to a 404.

### Root cause

In [`src/middleware/method-override/index.ts`](https://github.com/honojs/hono/blob/main/src/middleware/method-override/index.ts#L92), the URL-encoded branch uses strict string equality:

```ts
if (contentType === 'application/x-www-form-urlencoded') {
```

So `application/x-www-form-urlencoded;charset=UTF-8` fails the check.

The `multipart/form-data` branch in the same function (line 74) already uses `startsWith`, and the rest of the codebase consistently tolerates media-type parameters:

- [`src/utils/body.ts` L105](https://github.com/honojs/hono/blob/main/src/utils/body.ts#L105) uses `contentType?.startsWith('application/x-www-form-urlencoded')`
- [`src/validator/validator.ts` L26](https://github.com/honojs/hono/blob/main/src/validator/validator.ts#L26) uses a regex that accepts parameters
- [`src/middleware/csrf/index.ts` L30](https://github.com/honojs/hono/blob/main/src/middleware/csrf/index.ts#L30) uses a regex that accepts parameters

The same class of bug was previously fixed in the validator middleware in #3297 (`URL Encoded Data could not be validated if the Content-Type included charset`).

### Proposed fix

Change line 92 to `startsWith`, matching the multipart branch above it and `utils/body.ts`:

```ts
if (contentType?.startsWith('application/x-www-form-urlencoded')) {
```

I would be happy to open a PR with this fix and tests if this makes sense.

mixelburg:
Fixes #4892

The `method-override` middleware checked Content-Type with an exact match (`=== 'application/x-www-form-urlencoded'`), but browsers and libraries like Hotwired Turbo automatically append a charset (e.g. `application/x-www-form-urlencoded;charset=UTF-8`) when sending URLSearchParams as the request body.

Changed to `startsWith` to match the pattern already used for `multipart/form-data`.

### DIFF
--- a/src/middleware/method-override/index.test.ts
+++ b/src/middleware/method-override/index.test.ts
@@ -95,6 +95,23 @@ describe('Method Override Middleware', () => {
         expect(data.contentType).toBe('application/x-www-form-urlencoded')
       })
 
+      it('Should override POST to DELETE - with charset in Content-Type', async () => {
+        const params = new URLSearchParams()
+        params.append('message', 'Hello')
+        params.append('_method', 'DELETE')
+        const res = await app.request('/posts', {
+          body: params,
+          headers: {
+            'Content-Type': 'application/x-www-form-urlencoded;charset=UTF-8',
+          },
+          method: 'POST',
+        })
+        expect(res.status).toBe(200)
+        const data = await res.json()
+        expect(data.method).toBe('DELETE')
+        expect(data.message).toBe('Hello')
+      })
+
       it('Should override POST to PATCH - not found', async () => {
         const form = new FormData()
         form.append('message', 'Hello')
--- a/src/middleware/method-override/index.ts
+++ b/src/middleware/method-override/index.ts
@@ -89,7 +89,7 @@ export const methodOverride = (options: MethodOverrideOptions): MiddlewareHandle
         }
       }
       // Content-Type is `application/x-www-form-urlencoded`
-      if (contentType === 'application/x-www-form-urlencoded') {
+      if (contentType?.startsWith('application/x-www-form-urlencoded')) {
         const params = await parseBody<Record<string, string>>(clonedRequest)
         const method = params[methodFormName]
         if (method) {


mast1ff:
Thanks for picking this up @mixelburg. For the record, I had explicitly mentioned in #4892 that I'd be happy to open a PR with the fix and tests, so it would have been nice to be given a chance to do so — but the fix itself looks correct and matches what I outlined in the issue.

One small suggestion: the added test only covers `application/x-www-form-urlencoded;charset=UTF-8` (no space after `;`). It might be worth adding a case with a space (`application/x-www-form-urlencoded; charset=UTF-8`) since RFC 7231 allows whitespace before parameters, and some clients send it that way. The `startsWith` check covers both, so it's just a matter of pinning down the behavior with a regression test.

yusukebe (approved):
LGTM!

yusukebe:
@mixelburg Thank you.

@mast1ff I think just adding the minimal test is okay. This looks good to me. I don't know @mixelburg referred your code, but I'll add you as a co-author. Thanks.

mast1ff:
Thanks @yusukebe, really appreciate the co-author credit and the thoughtful handling. Glad I could contribute to Hono — looking forward to using it (and contributing more) in the future 🙏

PR #4898, opened 2026-05-01
Universal cache middleware

saintparish4:
## Summary

- Refactors `src/middleware/cache/` to accept a pluggable `KVLike` storage
  adapter, so the middleware works on any runtime - not just where the Web
  Cache API is available.
- Ships two adapters in core: `cacheApi()` (Workers / Deno, zero-overhead
  pass-through of `Response`) and `memoryStore()` (Node / Bun / anywhere,
  bounded LRU + lazy TTL).
- 100% backwards compatible: `cacheName` and `wait` keep working unchanged;

## Why

Today `src/middleware/cache/index.ts` hard-codes the Web Cache API
(`globalThis.caches.open(name).match(key)`). On Node and Bun the middleware
disables itself entirely; on Cloudflare Workers users who want to use Workers
KV or another store have no path. This PR introduces a small `KVLike`
interface so any adapter can plug in, while preserving the zero-allocation
fast path on the runtime where Hono is most heavily deployed.

## Status / what's in this draft right now

This PR is opened as a **Draft**. The work is broken into phases; each phase
ends with the test suite green before the next is started. Current state on
this branch:

- [x] **Phase 1 - Pure types and parser utilities** - `types.ts`, `utils.ts`,
  `utils.test.ts` (commit `5f8d2941`)
- [x] Phase 2 - Refactor `index.ts` to consume utils (no behavior change)
- [x] Phase 3 - Extract adapter modules (`cache-api.ts`, `memory.ts`,
  shared `contract.ts`)
- [ ] Phase 4 - Vary-aware keying with two-phase warm-up rewrite
- [ ] Phase 5 - `onStoreError` hook + `FaultInjectingStore` test helper
- [ ] Phase 6 - Backwards-compat shims (`cacheName`, `wait`)
- [ ] Phase 7 - RFC 7234 field-list bug fix (test flip + new positive tests)
- [ ] Phase 8 - Runtime tests (`workerd`, `deno`, `bun`, `node`) +
  `benchmarks/cache.ts`
- [ ] Phase 9 - JSDoc, exports verification, lint/format pass

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ sandbox
 # Cloudflare Workers
 worker
 .wrangler
+.cursor 
 
 # Logs
 logs
--- a/.vitest.config/setup-vitest.ts
+++ b/.vitest.config/setup-vitest.ts
@@ -34,6 +34,10 @@ class MockCache {
   async put(key: Request | string, response: Response): Promise<void> {
     this.store.set(key, response)
   }
+
+  async delete(key: Request | string): Promise<boolean> {
+    return this.store.delete(key)
+  }
 }
 
 const globalStore: Map<string | Request, Response> = new Map()
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,5 +4,9 @@
   "editor.codeActionsOnSave": {
     "source.fixAll.eslint": "explicit"
   },
-  "typescript.tsdk": "node_modules/typescript/lib"
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "cSpell.words": [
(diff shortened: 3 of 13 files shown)


yusukebe:
@saintparish4 

Regarding this feature, we don't want to accept it. I know you can make this simple so that we can use any storage. But we don't increase the code of this middleware. I'll close it. Please do it somewhere else. Thanks.

saintparish4:
Thanks for the clarity on core. Where would you like pluggable cache storage to land instead --- @hono/middleware, a separate package, or mostly left to user land/docs ?

PR #4899, opened 2026-05-02
fix(ssg): add `atom+xml` and `rss+xml` to `defaultExtensionMap`

yuintei:
I came across this issue while creating a custom Atom feed for my SSG website.
This adds `application/atom+xml` and `application/rss+xml` to `defaultExtensionMap`.

- `toSSG` supports a custom `extensionMap`, but `@hono/vite-ssg` does not expose it, so `application/atom+xml` routes like `atom.xml` fall back to `atom.xml.html`.
- Exposing `extensionMap` from `@hono/vite-ssg` would work around this, but Atom and RSS are common XML feed formats for SSG sites and should be handled correctly by default.
- `@hono/ssg-plugins-essential` writes its own `rss.xml`, but it should remain optional. Custom route-based Atom/RSS feeds still need correct default extension mapping.

This keeps the default mapping scoped to the two most common feed media types, without trying to cover every XML-based media type.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/helper/ssg/ssg.test.tsx
+++ b/src/helper/ssg/ssg.test.tsx
@@ -541,6 +541,18 @@ describe('saveContentToFile function', () => {
       mimeType: htmlMimeType,
     }
 
+    const atomData = {
+      routePath: '/atom',
+      content: yamlContent,
+      mimeType: 'application/atom+xml',
+    }
+
+    const rssData = {
+      routePath: '/rss',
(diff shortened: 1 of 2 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@yuintei 

Thanks! Looks good.

PR #4905, opened 2026-05-04
fix(cors): make origin optional in CORSOptions

### LINKED ISSUE #4904: CORS middleware: `origin` should be optional
Currently `origin` is a required member of `CORSOptions`, but according to [the documentation](https://hono.dev/docs/middleware/builtin/cors#origin-string-string-origin-string-c-context-string) and the fact that it has default value, it should be optional.

truffle-dev:
The CORS middleware sets `origin: '*'` when not provided, and the JSDoc already documents `[options.origin='*']`. But `CORSOptions.origin` was required, so this failed to typecheck even though the runtime works:

```ts
cors({ allowMethods: ['GET', 'POST'] })
//   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
// Property 'origin' is missing in type ...
```

Make `origin` optional so the type matches the documented runtime default. No runtime change.

Test verifies that omitting `origin` still produces `Access-Control-Allow-Origin: *` and that an explicit `allowMethods` is honored.

- [x] Add tests
- [x] Run tests
- [x] Add [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code (existing JSDoc already says `[options.origin='*']`)

Closes #4904

### DIFF
--- a/src/middleware/cors/index.test.ts
+++ b/src/middleware/cors/index.test.ts
@@ -418,4 +418,22 @@ describe('CORS by Middleware', () => {
     expect(res.headers.get('Access-Control-Allow-Origin')).toBe('null')
     expect(res.headers.get('Access-Control-Allow-Credentials')).toBe('true')
   })
+
+  it('Options without origin fall back to wildcard default', async () => {
+    const subApp = new Hono()
+    subApp.use('/api/*', cors({ allowMethods: ['GET', 'POST'] }))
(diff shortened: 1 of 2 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@truffle-dev 

Thanks! Looks good.

PR #3852, opened 2025-01-25
fix(types): missing response type on `OnHandlerInterface`

sor4chi:
### The author should do the following, if applicable

- [x] Add tests
- [x] Run tests
- [x] `bun run format:fix && bun run lint:fix` to format the code
- [x] Add [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code

Issue #3746, opened 2024-12-12
MiddlewareHandler response type inference


FOZERY:
### What is the feature you are proposing?

When we use middleware, it can return a Response object or a Typed Response, for example, @hono/validator returns a MiddlewareHandler that can return a response with a 400 status code.

 The validator function returns a middleware handler type with a probably response returning type:
<img width="595" alt="image" src="https://github.com/user-attachments/assets/fbd6d29a-5b23-4e51-9e0c-ceac557e6b49" />

But then, if we use a validator or custom middleware that returns a response object, Hono can't infer the returning type for the Hono instance, and RPC also. Example from official docs:
<img width="848" alt="image" src="https://github.com/user-attachments/assets/75e5c9d2-4420-4d3e-99a5-9fe8026a32e8" />
We can see that hono infer only 201 response status code. And we can't see 400 code that return zValidator in middleware.

We can also create a custom validator and see the same thing - in the validator we return the 400 code, but we don't see it in the types:
<img width="1130" alt="image" src="https://github.com/user-attachments/assets/8e0c2467-2dd2-4977-a8dd-4cfcbf50c2a4" />

So we can't handle error if status code === 400.

We can solve this by using the res.status() method before returning the res.json() and get "StatusCode" type in statusCode field.
<img width="1026" alt="image" src="https://github.com/user-attachments/assets/ffad62ae-5979-41c5-ae8b-96132496549c" />

But I think that's not how it's supposed to work.

So I've go in types.ts file in source code of Hono.dev and start working on type inference for MiddlewareHandlers.

You can see what I have come up with in the my fork of Hono
[GitHub Repo](https://github.com/FOZERY/hono/blob/main/src/types.ts#L121). However, I have gotten stuck and am still unable to make progress with type inference from the validator function, as well as with type inference for functions such as app.get(...handlers: H[]), in the HandlerInterface.

lewisedc:
This has been brought a few times before, #2719, #580, #3170.
The devs seem to be aware of it but sadly don't think there's been any update in about 6 months.

FOZERY:
> This has been brought a few times before, #2719, #580, #3170. The devs seem to be aware of it but sadly don't think there's been any update in about 6 months.

So now the only way to get around the typing error in rpc is to use c.status() before returning the response, like in my example?

EdamAme-x:
To begin with, the status code may be changed by external factors such as reverse proxies, so perhaps the status code for the Client response type should be abstract.

[...]

PR #4906, opened 2026-05-04
fix(types): propagate middleware response types to app.on overloads

T4ko0522:
## Summary

PR #4393 introduced `MergeMiddlewareResponse<M_k>` to capture middleware responses (e.g. `c.json({
error }, 401)` returned from a preceding handler) into the inferred RPC schema. The change was applied
to `HandlerInterface` (`app.get/post/...`) but **`OnHandlerInterface` was not updated**, so the same
route declared with `app.on(method, path, ...)` or `app.on(method[], path, ...)` silently dropped
middleware responses from the schema.

This is the same shape of asymmetry that #3852 and #4865 already fixed for narrower cases (`R` discard)
 — `OnHandlerInterface` keeps drifting away from `HandlerInterface` whenever the handler-side overloads
 gain new type parameters.

## Affected overloads

18 overloads in total:

- `app.on(method, path, handler x2..x10)` — 9 overloads
- `app.on(method[], path, handler x2..x10)` — 9 overloads

The single-handler variants (`x1`) and the rest-array fallback overloads are unaffected because they
have no preceding middleware position to capture.

## Reproduction

```ts
import { Hono } from 'hono'

const auth = createMiddleware(async (c, next) => {
  if (!authorized) return c.json({ error: 'unauthorized' }, 401)
  await next()
})

// Works correctly (already covered by #4393)
const a = new Hono().get('/secret', auth, (c) => c.json({ ok: true }, 200))

// Bug: `output` collapses to `{ ok: true }` only, dropping the 401 response
const b = new Hono().on('GET', '/secret', auth, (c) => c.json({ ok: true }, 200))

const client = hc<typeof b>('/')
const res = await client.secret.$get()
if (res.status === 401) {
  // Before this PR: TypeScript narrows to `never`, so this branch is unreachable
  // After this PR:  `res.json()` resolves to `{ error: string }`
  const err = await res.json()
}
```

Switching `app.on('GET', ...)` ↔ `app.get(...)` flips the inference even though the routes are
semantically identical. The same gap exists for the `method[]` form.

## Fix

For each of the 18 affected overloads, add `M1..M(n-1)` middleware type parameters mirroring
`HandlerInterface`, intersect them into the handler tuple positions before the final response handler,
and union `MergeMiddlewareResponse<M_k>` into the schema's response type.

The same shape is applied for handler counts 3..10 (with `M1..M(n-1)`) and for the `methods: M[]`
overloads.

## Test plan

Added regression tests under the existing `describe('RPC supports Middleware responses', ...)` block in
 `src/types.test.ts`, mirroring the corresponding `app.get` tests. Coverage:

- `app.on('GET', '/test', ...)` with 1, 2, 5, and 9 middlewares before the response handler
- `app.on(['GET'], '/test', ...)` with 1, 2, 5, and 9 middlewares before the response handler

Each test asserts that every middleware status code (`400`, `401`, `403`, ...) is reachable in the
inferred RPC client and that `res.json()` is narrowed correctly per status. These tests fail on `main`
(status union collapses to just `200`, so every middleware-status branch becomes `never`) and pass
after this fix.

- [x] `bun run test` (full type check + vitest, 3400 passed)
- [x] Added regression tests for both single-method and `method[]` variants

## Related history

- **#4393** (2025-10-15) — feat(types): passing middleware types
  Introduced `MergeMiddlewareResponse<M_k>` for `HandlerInterface` only.
- **#3852** (2025-01-25) — fix(types): missing response type on `OnHandlerInterface`
  Earlier `R` discard fix for `OnHandlerInterface` (different from this one).
- **#4865** (2026-04-08) — fix(types): infer response type from last handler in app.on 9-/10-handler
overloads
  Closer cousin: same kind of `OnHandlerInterface` drift, narrower scope.

This PR completes the alignment so that `OnHandlerInterface` mirrors `HandlerInterface` for both `R`
and `M_k` propagation. May also alleviate the open feedback in #3746.

### DIFF
--- a/src/types.test.ts
+++ b/src/types.test.ts
@@ -3498,6 +3498,282 @@ describe('RPC supports Middleware responses', () => {
       type verify = Expect<Equal<Expected, Actual>>
     })
   })
+
+  describe('Merge responses from app.on handlers, single method', () => {
+    test('merge responses from 1 middleware', async () => {
+      const middleware = createMiddleware(async (c) => c.json({ '400': true }, 400))
+
+      const app = new Hono().on('GET', '/test', middleware, (c) => c.json({ '200': true }, 200))
+      const client = hc<typeof app>('http://localhost', { fetch: app.request })
+      const res = await client.test.$get()
+
+      if (res.status === 200) {
+        expectTypeOf(await res.json()).toEqualTypeOf<{ 200: true }>()
+      }
+      if (res.status === 400) {
+        expectTypeOf(await res.json()).toEqualTypeOf<{ 400: true }>()
+      }
+    })
+
+    test('merge responses from 2 middlewares', async () => {
+      const middleware1 = createMiddleware(async (c) => c.json({ '400': true }, 400))
+      const middleware2 = createMiddleware(async (c) => c.json({ '401': true }, 401))
+
+      const app = new Hono().on('GET', '/test', middleware1, middleware2, (c) =>
+        c.json({ '200': true }, 200)
+      )
+      const client = hc<typeof app>('http://localhost', { fetch: app.request })
+      const res = await client.test.$get()
+
+      if (res.status === 200) {
+        expectTypeOf(await res.json()).toEqualTypeOf<{ 200: true }>()
+      }
+      if (res.status === 400) {
+        expectTypeOf(await res.json()).toEqualTypeOf<{ 400: true }>()
+      }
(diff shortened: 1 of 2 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@T4ko0522 

Nice catch. This should be fixed! Thanks.

PR #4909, opened 2026-05-05
fix(utils/mime): extend charset to xml-based mime types

### LINKED ISSUE #4510: Proposal to change to specifying `charset` parameter individually for each MIME type

Currently, MIME types parameter are set through mechanical determination.

https://github.com/honojs/hono/blob/fa8eef707990e80a593fa32c2f67713a2ecd2e9b/src/utils/mime.ts#L16-L18

This has several problems.

- There are other items besides `text/*` for which I want to specify a `charset` parameter (`image/svg+xml`, `application/xhtml+xml`, `application/xml`)
- According to [IANA specifications](https://www.iana.org/assignments/media-types/media-types.xhtml), not all `text/*` formats have a `charset` parameter

Therefore, we will eliminate this processing and instead specify the `charset` parameter individually using the `_baseMimes` variable. (e.g. `css: 'text/css; charset=utf-8',`)

Furthermore, in the `getExtension()` function, we will modify the logic to perform the check after removing the parameter instead.

kasharkzy:
The getMimeType() function was using startsWith('text') to decide when to append ; charset=utf-8 to a MIME type, that heuristic misses types like image/svg+xml, application/xhtml+xml and application/xml, which are all textual formats where charset matters for correct decoding.

This replaces the prefix check with a condition that covers three cases:
  text/ prefix, per RFC 2046
  +xml structured syntax suffix, per RFC 6839
  application/xml explicitly, per RFC 7303 §8.5

The change is a single-line diff in mime.ts, with three new assertions in
mime.test.ts covering the previously untested types.

Fixes #4510

### DIFF
--- a/src/utils/mime.test.ts
+++ b/src/utils/mime.test.ts
@@ -16,6 +16,9 @@ describe('mime', () => {
     expect(getMimeType('goodmorninghellogif')).toBeUndefined()
     expect(getMimeType('indexjs.abcd')).toBeUndefined()
     expect(getMimeType('IMAGE.JPG')).toBe('image/jpeg')
+    expect(getMimeType('logo.svg')).toBe('image/svg+xml; charset=utf-8')
+    expect(getMimeType('page.xhtml')).toBe('application/xhtml+xml; charset=utf-8')
+    expect(getMimeType('feed.xml')).toBe('application/xml; charset=utf-8')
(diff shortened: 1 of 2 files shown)


yusukebe:
#4510 was fixed by #4912. Closing.

PR #4912, opened 2026-05-07
fix(mime): specify charset parameter per MIME type instead of mechanical detection

### LINKED ISSUE #4510: Proposal to change to specifying `charset` parameter individually for each MIME type

renatograsso10:
Closes #4510

## Summary

Move the `; charset=utf-8` suffix from a runtime check in `getMimeType` into the `_baseMimes` data table itself, and extend it to `image/svg+xml`, `application/xhtml+xml`, and `application/xml` as the issue requested.

## Why

Per #4510:
1. Other entries besides `text/*` benefit from a `charset` parameter (XML-based formats).
2. Per [IANA media types](https://www.iana.org/assignments/media-types/media-types.xhtml), `charset` is not universal across all `text/*` formats — keeping it as data lets us be precise per entry rather than running a generic `startsWith("text")` check.

The output of `getMimeType()` for previously-charseted text types is unchanged (still `text/<sub>; charset=utf-8`), so this is backward-compatible at the call site.

## Changes

`src/utils/mime.ts`
- `_baseMimes`: append `; charset=utf-8` to `css`, `csv`, `htm`, `html`, `ics`, `js`, `mjs`, `txt`, `svg`, `xhtml`, `xml`.
- `getMimeType`: drop the `if (mimeType.startsWith("text"))` block; data now carries the charset.
- `getExtension`: compare the base type (split on `;`) so callers can pass either `"text/html"` or `"text/html; charset=utf-8"` and still get the right extension. No behavior change for callers passing the raw type.

`src/utils/mime.test.ts`
- 2 new `it` blocks: charset values for the new XML-based formats and the additional text types; `getExtension` round-trip with and without parameters.

## Test plan

- [x] `npx vitest run src/utils/mime.test.ts` — 5/5 pass (3 existing + 2 new)
- [x] `npx vitest run src/middleware/compress src/middleware/jsx-renderer src/middleware/serve-static` — 56/56 pass (covers the consumers I could trace: `getMimeType` is used by `serve-static`, `BaseMime` is referenced by `context.ts`)
- [x] `npx tsc --noEmit` clean

The pre-existing `stream`/`streamSSE` `onAbort` and a few other failures on `main` are unrelated to this change (verified via `git stash` + re-run).

### DIFF
--- a/src/utils/mime.test.ts
+++ b/src/utils/mime.test.ts
@@ -18,6 +18,16 @@ describe('mime', () => {
     expect(getMimeType('IMAGE.JPG')).toBe('image/jpeg')
   })
 
+  it('getMimeType returns charset parameter for text and XML-based formats', () => {
+    expect(getMimeType('icon.svg')).toBe('image/svg+xml; charset=utf-8')
+    expect(getMimeType('page.xhtml')).toBe('application/xhtml+xml; charset=utf-8')
+    expect(getMimeType('feed.xml')).toBe('application/xml; charset=utf-8')
+    expect(getMimeType('events.ics')).toBe('text/calendar; charset=utf-8')
+    expect(getMimeType('script.mjs')).toBe('text/javascript; charset=utf-8')
+    expect(getMimeType('style.css')).toBe('text/css; charset=utf-8')
+    expect(getMimeType('data.csv')).toBe('text/csv; charset=utf-8')
+  })
+
   it('getMimeType with custom mime', () => {
     expect(getMimeType('morning-routine.m3u8', mime)).toBe('application/vnd.apple.mpegurl')
     expect(getMimeType('morning-routine1.ts', mime)).toBe('video/mp2t')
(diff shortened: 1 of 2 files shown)


yusukebe:
@renatograsso10 Thank you for the PR.

Hey @SaekiTominaga! Sorry for being super late. I think this is not bad. What do you think of this?

SaekiTominaga:
I think this is fine. Thank you!

yusukebe (approved):
LGTM!

PR #4915, opened 2026-05-08
feat(cache): key cache entries by configured vary headers

usualoma:
### Summary

- Improves cache efficiency when the `Vary` header is in use.
- Until now, `options.vary` only affected the response headers. With this PR, it also contributes to the cache key.
- This is also a follow-up to the v4.12.18 vulnerability fix (568c2ecc), which made `Vary`-bearing responses uncacheable. The change preserves that fix while letting users opt back into caching where it's safe.

### Behavior changes

The behavior changes when `options.vary` is configured:

| Situation | Before | After this PR |
| --- | --- | --- |
| `options.vary` set / response has no `Vary` | All requests share a single cache entry (requests that should return distinct responses can collide) | Stored as separate variants keyed by the configured request header values |
| `options.vary` set / response `Vary` is a subset of `options.vary` | Cache skipped (since v4.12.18) | Stored as separate variants keyed by the configured request header values |
| Response `Vary` includes a header outside `options.vary` | Cache skipped | Cache skipped (unchanged) |
| Response `Vary: *` | Cache skipped | Cache skipped (unchanged) |

Because of this behavior change, requests that were previously (unintentionally) sharing the same entry under `options.vary` are now split per header value, which can lower the hit rate. That said, this aligns with the intended semantics of `Vary` and should be a desirable change in virtually every deployment. On the flip side, paths that produced `Vary`-bearing responses — which were not cached at all after v4.12.18 — become cacheable again as long as their `Vary` is covered by `options.vary`, which raises the hit rate there.

### Design note: interaction with `options.keyGenerator`

Users who already supply `options.keyGenerator` may already be folding the same request headers into the key themselves. With this change, those headers will be appended to the key a second time.

I considered skipping the `options.vary`-based key extension when `options.keyGenerator` is provided, but decided against it:

- If a user wrote their code assuming "configuring `options.vary` extends the cache key", silently skipping that extension would violate their assumption.
- A double append is a safe failure mode: it can only over-split variants — it cannot return the wrong response.

So when both are configured, `options.vary` still contributes to the cache key unconditionally.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/cache/index.test.ts
+++ b/src/middleware/cache/index.test.ts
@@ -379,6 +379,24 @@ describe('Cache Middleware', () => {
 })
 
 describe('Cache Skipping Logic', () => {
+  const stubStoreBackedCache = () => {
+    const store = new Map<string | Request, Response>()
+    const mockCache = {
+      match: vi.fn((key: string | Request) => Promise.resolve(store.get(key)?.clone())),
+      put: vi.fn((key: string | Request, response: Response) => {
+        store.set(key, response.clone())
+        return Promise.resolve()
+      }),
+      keys: vi.fn(() => Promise.resolve(Array.from(store.keys()))),
+    }
+
+    vi.stubGlobal('caches', {
+      open: vi.fn().mockResolvedValue(mockCache),
+    })
+
+    return { mockCache }
+  }
+
   let putSpy: ReturnType<typeof vi.fn>
   let matchSpy: ReturnType<typeof vi.fn>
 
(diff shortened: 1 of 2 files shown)


usualoma:
Hi @yusukebe,

yusukebe (approved):
LGTM!

yusukebe:
Hey @usualoma 

I considered whether there are patterns does not work, but it looks good! I'll merge this later and may include it in the next patch. Thanks!

PR #4917, opened 2026-05-09
fix(cookie): return last value for duplicate cookie names when name param is used

### LINKED ISSUE #4916: bug: getCookie(c, name) and getCookie(c)[name] return different values for duplicate cookie names

4.12.18

Bun

`getCookie(c, name)` and `getCookie(c)[name]` return different values when the request `Cookie` header contains the same name twice.

Minimal reproduction:

```ts
import { Hono } from 'hono'

const app = new Hono()

app.get('/probe', (c) => {
  const all = getCookie(c)              // bulk form
  const byKey = getCookie(c, 'a')       // key form
  return c.json({
    raw: c.req.raw.headers.get('Cookie'),
    'getCookie(c)[a]': all['a'],
    "getCookie(c, 'a')": byKey,
  })
})

export default app
```

Request:

```sh
curl http://localhost:3000/probe -H 'Cookie: a=first; a=last'
```

Actual response:

```json
{
  "raw": "a=first; a=last",
  "getCookie(c)[a]": "last",
  "getCookie(c, 'a')": "first"
}
```

The same `Cookie` header is interpreted as `last` by one call and `first` by the other.

Root cause is in `src/utils/cookie.ts` `parse(cookie, name?)`:

- When `name` is provided, the function `break`s out of the loop on the first match → first-wins.
- When `name` is omitted, the loop runs to completion and later pairs overwrite earlier ones in the returned object → last-wins.

Both forms are exposed via `getCookie(c, name)` and `getCookie(c)` in `src/helper/cookie/index.ts`, so both code paths are reachable from a normal application.

Both call shapes should return the same value for the same `Cookie` header.

I prefer both last-wins (matches Express's `cookie`, Node `http.cookies`, and most browsers' DOM-side `document.cookie` enumeration order). Implementation: remove the early `break` in `parse(cookie, name)` so the loop always processes every pair.

### What do you see instead?

_No response_

### Additional information

_No response_

mixelburg:
Fixes #4916

## Problem

When the  header contains duplicate names (e.g. ), the two call shapes of  return different values:

-  → `first` (first-wins, due to early `break`)
-  → `last` (last-wins, later pairs overwrite)

## Root cause

In `src/utils/cookie.ts`, the `parse(cookie, name?)` function had an early `break` when a matching name was found, returning the first match. The bulk form (no name param) let the loop run to completion so later pairs won.

## Fix

Removed the early `break` so both code paths use last-wins semantics, matching Express's `cookie` parser, Node `http.cookies`, and browser `document.cookie` enumeration order.

## Changes

- `src/utils/cookie.ts`: Removed `break` + the conditional that triggered it
- `src/utils/cookie.test.ts`: Added test verifying duplicate cookies return last value for both call shapes

## Tests

All existing 30 tests pass, including the new one.

### DIFF
--- a/src/utils/cookie.test.ts
+++ b/src/utils/cookie.test.ts
@@ -44,6 +44,15 @@ describe('Parse cookie', () => {
     expect(cookie['no_such_cookie']).toBeUndefined()
   })
 
+  it('Should return last value for duplicate cookie names with name param (last-wins)', () => {
+    const cookieString = 'a=first; a=last'
+    const byName: Cookie = parse(cookieString, 'a')
+    const bulk: Cookie = parse(cookieString)
+    expect(byName['a']).toBe('last')
+    expect(bulk['a']).toBe('last')
+    expect(byName['a']).toBe(bulk['a'])
+  })
+
   it('Should parse cookies with no value', () => {
     const cookieString = 'yummy_cookie=; tasty_cookie = ; best_cookie= ; last_cookie=""'
     const cookie: Cookie = parse(cookieString)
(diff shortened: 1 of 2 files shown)


yusukebe:
Issue #4916 has already been closed.

PR #4918, opened 2026-05-10
fix: allow middleware to catch downstream errors when onError is regi…

### LINKED ISSUE #4895: Allow middleware to intercept handler errors
## Problem

When `app.onError` is registered, middleware that wraps `await next()` in a `try/catch` never sees errors thrown downstream. Compose intercepts them at the inner dispatch and routes them to `onError` before the await unwinds:

```ts
// compose.ts
try {
} catch (err) {
  if (err instanceof Error && onError) {
    context.error = err
    res = await onError(err, context)   // swallowed here
    isError = true
  } else {
    throw err
  }
}
```

This blocks legitimate middleware patterns — for example:

- **Error recovery**: catch a downstream failure and serve a fallback response (cached value, degraded mode, feature-flag fallback).
- **Per-group error transformation**: a sub-router wraps/tags errors and delegates final shaping to a parent `app.onError`. Sub-Hono's own `onError` can't help — it must end the chain with a `Response`.

## Proposal

Add an optional argument to `next`:

```ts
type NextOptions = { rethrow?: boolean };
type Next = (options?: NextOptions) => Promise<void>;
```

Default = current behavior. When `{ rethrow: true }` is passed, the dispatch lets the error propagate out of `await next()` so the caller can `try/catch` it.

Fully backward-compatible: `next()` with no argument behaves exactly as today.

Rohitsingh385:
Closes #4895

**Problem**
When `app.onError` is registered, middleware wrapping `await next()` in a 
try/catch never sees downstream errors. `compose.ts` catches them inside 
`dispatch` and routes them to `onError` before the error can bubble up.

**Fix**
Add an optional `{ rethrow?: boolean }` argument to `next`. When 
`rethrow: true` is passed, the catch block skips `onError` and rethrows 
the error, allowing the calling middleware's try/catch to handle it.

Default behavior is unchanged — fully backward compatible.

**Usage**
app.use(async (c, next) => {
  try {
    await next({ rethrow: true })
  } catch (err) {
    return c.text('fallback', 500)
  }
})

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/compose.test.ts
+++ b/src/compose.test.ts
@@ -748,4 +748,33 @@ describe('Compose', function () {
     expect(ctx.get('middleware')).toEqual(1)
     expect(ctx.get('next')).toEqual(1)
   })
+
+})
+describe('compose with rethrow option', () => {
+  it('should rethrow error to middleware when next({ rethrow: true }) is called', async () => {
+    const middleware: MiddlewareTuple[] = []
+    let middlewareCaughtError = false
+
+    const mHandler = async (_c: Context, next: Next) => {
+      try {
+        await next({ rethrow: true })
+      } catch {
+        middlewareCaughtError = true
+      }
+    }
+
+    const handler = () => {
+      throw new Error('downstream error')
+    }
+
+    const onError = (_err: Error, c: Context) => c.text('onError', 500)
+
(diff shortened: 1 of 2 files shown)


yusukebe:
It's not well discussed. This PR is currently not needed.

PR #4922, opened 2026-05-11
fix(cookie): return the first cookie when there are multiple cookies with the same name

### LINKED ISSUE #4916: bug: getCookie(c, name) and getCookie(c)[name] return different values for duplicate cookie names

usualoma:
fixes #4916

- eee6fa1ec9dcbff25a97a186a6f6ff5abadbd505: Align both code paths to first-wins semantics.
- e8a6fcfabfe8464c7dc5bfd89acd3f467d624498 : Additionally, use `Object.create(null)` to avoid collisions with `Object.prototype` properties.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/utils/cookie.test.ts
+++ b/src/utils/cookie.test.ts
@@ -118,6 +118,26 @@ describe('Parse cookie', () => {
     expect(parse(cookieString)['session']).toBe(parse(cookieString, 'session')['session'])
   })
 
+  it('Should parse cookies whose names collide with Object.prototype properties', () => {
+    const cookieString = 'toString=foo; hasOwnProperty=bar; constructor=baz'
+    const cookie: Cookie = parse(cookieString)
+    expect(cookie['toString']).toBe('foo')
+    expect(cookie['hasOwnProperty']).toBe('bar')
+    expect(cookie['constructor']).toBe('baz')
+  })
+
+  it('Should parse a cookie whose name collides with Object.prototype when parsed by name', () => {
+    const cookieString = 'toString=foo'
+    const cookie: Cookie = parse(cookieString, 'toString')
+    expect(cookie['toString']).toBe('foo')
+  })
+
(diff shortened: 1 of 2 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@usualoma 

Thanks. I think we can merge this now.

PR #4925, opened 2026-05-11
fix(etag): make stream digests independent of chunk boundaries

### LINKED ISSUE #4071: ETag changes every time when combine `etag` and `cache` middlewares

4.7.6

Cloudflare Workers

Repository: https://github.com/rwv/hone-etag-large-payload-test  
Live demo: https://hono-etag-large-payload-test.rwv.dev/

Steps:
1. Load the page
2. Refresh
3. Observe that the `ETag` header changes each time

For identical response bodies, the ETag should remain stable across requests, regardless of how the body is chunked.

### What do you see instead?

When combining the `etag()` and `cache()` middleware on Cloudflare Workers, the ETag value changes on every request—even when the response body remains the same. This breaks 304 behavior and prevents effective caching.

### Suspected Cause

On Cloudflare Workers, responses returned from the cache expose `.body` as a `ReadableStream`. The chunk structure of this stream is not consistent across requests.

This breaks the assumption made in [#3604](https://github.com/honojs/hono/pull/3604):

> “It is unlikely that the same architecture (server) will have different data divisions.”

**The above is speculative, based on observed behavior.**

### Suggestion

PR [#3832](https://github.com/honojs/hono/pull/3832) introduced the `generateDigest(body: Uint8Array)` option, which is useful but not sufficient in this case.

```ts
generateETag?: (res: ClonedResponse or stream?) => string | Promise<string>
```

This would:

- Native use of `DigestStream` in Cloudflare Workers for consistent hashing
- Better semantic clarity: digesting a stream should take a stream, not a fully merged `Uint8Array`
- Avoids confusion: the current `(body: Uint8Array) => ArrayBuffer` signature suggests that Hono reads the full body first, which it doesn’t. The chunked-digest logic is hidden from users and may cause incorrect assumptions.

raashish1601:
## Summary

- Accumulate the original response stream bytes before generating an ETag digest.
- Avoid feeding the previous digest into the next chunk, which made the final ETag depend on stream chunk boundaries.
- Add a regression test proving identical stream bytes produce the same ETag when split differently.

Fixes #4071.

## Validation

- `npx vitest --run src/middleware/etag/index.test.ts`
- `npx prettier --check src/middleware/etag/digest.ts src/middleware/etag/index.test.ts`
- `npx eslint src/middleware/etag/digest.ts src/middleware/etag/index.test.ts`
- `npx tsc --noEmit`
- `git diff --check`

### DIFF
--- a/src/middleware/etag/digest.ts
+++ b/src/middleware/etag/digest.ts
@@ -1,5 +1,5 @@
 const mergeBuffers = (
-  buffer1: ArrayBuffer | undefined,
+  buffer1: Uint8Array<ArrayBuffer> | undefined,
   buffer2: Uint8Array<ArrayBuffer>
 ): Uint8Array<ArrayBuffer> => {
   if (!buffer1) {
@@ -8,7 +8,7 @@ const mergeBuffers = (
   const merged = new Uint8Array<ArrayBuffer>(
     new ArrayBuffer(buffer1.byteLength + buffer2.byteLength)
   )
-  merged.set(new Uint8Array(buffer1), 0)
+  merged.set(buffer1, 0)
   merged.set(buffer2, buffer1.byteLength)
   return merged
 }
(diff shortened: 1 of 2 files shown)


usualoma:
Please do not have AI automatically generate PRs without reviewing the discussion on the issue.

PR #4933, opened 2026-05-12
fix(context): throw error for undefined values in c.json()

ruguoba:
## Description

This PR fixes the issue where `c.json()` silently converts `undefined` values to an empty string instead of throwing an error. This addresses the confusion raised in #2343.

### Problem

When passing `undefined` to `c.json()`:

```typescript
app.get('/api', (c) => {
  return c.json(undefined) // Returns empty string ""
})
```

The current behavior:
1. `JSON.stringify(undefined)` returns `undefined` (not a string)
2. `new Response(undefined)` converts it to an empty string
3. The client receives `""` instead of an error

This is confusing because:
- Developers expect `c.json()` to only accept valid JSON values
- Silent failures make debugging difficult
- Different runtimes may handle this differently

### Solution

Added validation in the `json` method to check if `JSON.stringify` returns `undefined`:

```typescript
json: JSONRespond = <...>(object: T, arg?: U | ResponseOrInit<U>, headers?: HeaderRecord): JSONRespondReturn<T, U> => {
  const stringified = JSON.stringify(object)
  if (stringified === undefined) {
    throw new TypeError('Cannot serialize undefined value to JSON. Use null instead.')
  }
  return this.#newResponse(
    stringified,
    arg,
    setDefaultContentType('application/json', headers)
  ) as any
}
```

### Benefits

1. **Explicit Error**: Developers immediately know when they pass `undefined`
2. **Consistent Behavior**: Same behavior across all runtimes
3. **Better Debugging**: Clear error message helps fix the issue
4. **Type Safety**: Encourages proper typing and null checks

### Breaking Change

This is a **breaking change** for code that currently relies on `c.json(undefined)` returning an empty string. However, such code is likely buggy and should be fixed.

### Migration Guide

If you have code that passes `undefined` to `c.json()`:

**Before:**
```typescript
app.get('/api', (c) => {
  const value = getValue() // might be undefined
  return c.json(value) // silently returns ""
})
```

**After:**
```typescript
app.get('/api', (c) => {
  const value = getValue() // might be undefined
  return c.json(value ?? null) // explicitly handle undefined
})
```

### Related Issues

Closes #2343

### Testing

Added unit tests to verify:
- `c.json(undefined)` throws TypeError
- `c.json(null)` works correctly
- `c.json({ key: undefined })` works correctly (undefined values in objects are omitted)

---

**Note**: This change makes `c.json()` behavior explicit and helps developers catch bugs early, improving the overall developer experience.

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -713,8 +713,12 @@ export class Context<
     arg?: U | ResponseOrInit<U>,
     headers?: HeaderRecord
   ): JSONRespondReturn<T, U> => {
+    const stringified = JSON.stringify(object)
+    if (stringified === undefined) {
+      throw new TypeError('Cannot serialize undefined value to JSON. Use null instead.')
+    }
     return this.#newResponse(
-      JSON.stringify(object),
+      stringified,
       arg,
       setDefaultContentType('application/json', headers)
     ) /* eslint-disable @typescript-eslint/no-explicit-any */ as any


yusukebe:
It will be handled by the maintainers. Closing.

PR #4934, opened 2026-05-12
fix(serveStatic): make options parameter optional in all adapters

### LINKED ISSUE #4911: Make options in serveStatic middleware from adapters optional

Hello! I'm taking a look at this [middleware](https://github.com/honojs/hono/blob/main/src/adapter/deno/serve-static.ts#L9) and for TS to check it successfully I need to provide an empty object:

```ts
import { Hono } from "@hono/hono";
import { serveStatic } from "@hono/hono/deno";

const app = new Hono();

app.use("/static/*", serveStatic({}));

Deno.serve(app.fetch);
```

It would be nice the example above could be written this way:

```ts
import { Hono } from "@hono/hono";
import { serveStatic } from "@hono/hono/deno";

const app = new Hono();

app.use("/static/*", serveStatic());

Deno.serve(app.fetch);
```

Notice there is not `{}` in serveStatic.

If you agree I can happily create a PR to support that! Please let me know 🙏🏼

mixelburg:
## Description

All properties in `ServeStaticOptions` are already optional, but the `serveStatic()` function in the Deno, Bun, and Cloudflare Workers adapters requires an explicit `{}` argument.

This change makes the `options` parameter optional with a default of `{}`, allowing:

```ts
serveStatic() // instead of serveStatic({})
```

Applied to all three adapters: Deno, Bun, and Cloudflare Workers.

Fixes #4911

### DIFF
--- a/src/adapter/bun/serve-static.ts
+++ b/src/adapter/bun/serve-static.ts
@@ -6,7 +6,7 @@ import type { ServeStaticOptions } from '../../middleware/serve-static'
 import type { Env, MiddlewareHandler } from '../../types'
 
 export const serveStatic = <E extends Env = Env>(
-  options: ServeStaticOptions<E>
+  options: ServeStaticOptions<E> = {}
 ): MiddlewareHandler => {
   return async function serveStatic(c, next) {
     const getContent = async (path: string) => {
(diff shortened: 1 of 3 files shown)


yusukebe (approved):
LGTM!

yusukebe:
Thanks!

PR #4944, opened 2026-05-17
fix(jsx): widen children type in jsx/jsxFn to Child[]

### LINKED ISSUE #4943: `jsx`/`jsxFn` children signature is narrower than `JSXNode` constructor and `Child` type
## Summary

In `hono/jsx`, the factory functions `jsx` / `jsxFn` (and therefore `createElement`, which is an alias of `jsx`) accept a narrower `children` type than what the `JSXNode` constructor and the `Child` type itself allow. This forces consumer libraries that call `createElement` directly with `JSXNode` or `Child` values to either cast or fall back to `hono/jsx/dom`'s `createElement`.

## Current declarations (`src/jsx/base.ts`)

```ts
export type Child = string | Promise<string> | number | JSXNode | null | undefined | boolean | Child[];

class JSXNode {
  constructor(tag: string | Function, props: Props, children: Child[]); // Child[] ✅
}

export const jsx: (
  tag: string | Function,
  props: Props | null,
  ...children: (string | number | HtmlEscapedString)[] // ❌ narrower than Child[]
) => JSXNode;

export const jsxFn: (
  tag: string | Function,
  props: Props,
  children: (string | number | HtmlEscapedString)[] // ❌ same
) => JSXNode;

export const cloneElement: (
  ...children: Child[] // ✅ correct
) => T;
```

Only `jsx` / `jsxFn` (and `createElement` via the alias) constrain children to a much smaller subset than `Child` allows. `JSXNode` constructor and `cloneElement` both accept `Child[]`.

## Reproduction

```ts
import { createElement } from 'hono/jsx'
import type { Child } from 'hono/jsx'

declare const myChild: Child
createElement('div', null, myChild)
// TS2345: Argument of type 'Child' is not assignable
// to parameter of type 'string | number | HtmlEscapedString'.
```

The same call against `hono/jsx/dom`'s `createElement` succeeds.

## Why this matters

Client-side libraries built on `hono/jsx` (e.g. `@ts-76/inertia-hono-jsx`) currently import `createElement` from `hono/jsx/dom` to get the wide signature. However, runtime imports from `hono/jsx/dom` produce plain VNodes that break `hono/jsx/dom/server.renderToString` during SSR. The intuitive fix — switch the runtime import to `hono/jsx` — works at runtime but fails `tsc` because of this narrower `createElement` signature.

If `jsx` / `jsxFn` accepted `Child[]` (matching `JSXNode` and `cloneElement`), the workaround would not be needed.

## Proposed fix

\`\`\`diff
- export const jsx: (tag, props, ...children: (string | number | HtmlEscapedString)[]) => JSXNode;
+ export const jsx: (tag, props, ...children: Child[]) => JSXNode;
- export const jsxFn: (tag, props, children: (string | number | HtmlEscapedString)[]) => JSXNode;
+ export const jsxFn: (tag, props, children: Child[]) => JSXNode;
\`\`\`

This is a **widening** change (subset → superset), so it should not break existing consumers.

## Question

Is the current narrower signature intentional (e.g. for performance reasons in the SSR string-concatenation path), or is it an oversight? Happy to follow up with a PR if the wider signature is acceptable.

mixelburg:
Fixes #4943

`jsx` and `jsxFn` accepted `(string | number | HtmlEscapedString)[]` for children, which is narrower than the `Child` type, `JSXNode`'s constructor signature, and `cloneElement`. This caused TS2345 errors when passing a `Child`-typed value to `createElement`.

Widened both to `Child[]` to match the constructor and `cloneElement`.

### DIFF
--- a/src/jsx/base.ts
+++ b/src/jsx/base.ts
@@ -294,7 +294,7 @@ export class JSXFragmentNode extends JSXNode {
 export const jsx = (
   tag: string | Function,
   props: Props | null,
-  ...children: (string | number | HtmlEscapedString)[]
+  ...children: Child[]
 ): JSXNode => {
   props ??= {}
   if (children.length) {
@@ -313,7 +313,7 @@ let initDomRenderer = false
 export const jsxFn = (
   tag: string | Function,
   props: Props,
-  children: (string | number | HtmlEscapedString)[]
+  children: Child[]
 ): JSXNode => {
   if (!initDomRenderer) {
     for (const k in domRenderers) {
--- a/src/jsx/base.ts
+++ b/src/jsx/base.ts
@@ -291,11 +291,7 @@ export class JSXFragmentNode extends JSXNode {
   }
 }
 
-export const jsx = (
-  tag: string | Function,
-  props: Props | null,
-  ...children: Child[]
-): JSXNode => {
+export const jsx = (tag: string | Function, props: Props | null, ...children: Child[]): JSXNode => {
   props ??= {}
   if (children.length) {
     props.children = children.length === 1 ? children[0] : children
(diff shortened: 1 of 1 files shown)


usualoma:
Hi @mixelburg,
As you can see from the referenced issue, the original poster is simply engaging in a polite discussion and intends to create the pull request themselves. I believe we should not create a pull request for such an issue while ignoring the discussion.

PR #4950, opened 2026-05-19
feat(compress): add shouldCompress option for custom content-type filtering

### LINKED ISSUE #4946: Support custom Content-Type in compress middleware

I want to compress responses with `Content-Type: application/vnd.msgpack`. (FYI it's a MIME type registered with IANA: https://github.com/msgpack/msgpack/issues/194)
But Hono's compress middleware only compresses responses with known Content-Type defined here:
https://github.com/honojs/hono/blob/7e62bcd22fa4e8f0e83cb564bac85e32f5434dd3/src/utils/compress.ts#L9-L10

I would like an option for custom conditions for which Content-Types are allowed to be compressed.
However, I'm not sure what format the ideal option should take. (For example, a function that returns a boolean, a list of Content-Types, or a regex? Also, should it be possible to add or remove from a default list of Content-Types, or should all of them have to be specified manually?)

mixelburg:
Closes #4946.

Adds an optional `shouldCompress` function to `CompressionOptions` that lets callers override the built-in content-type check entirely. When provided it replaces the default `COMPRESSIBLE_CONTENT_TYPE_REGEX` test, so any content-type (e.g. `application/vnd.msgpack`, custom vendor types) can be included or excluded.

```ts
app.use(compress({
  shouldCompress: (res) => {
    const type = res.headers.get('Content-Type') ?? ''
    return COMPRESSIBLE_CONTENT_TYPE_REGEX.test(type) || type.startsWith('application/vnd.msgpack')
  }
}))
```

`COMPRESSIBLE_CONTENT_TYPE_REGEX` is now re-exported from the middleware module so callers can compose it in their own predicate.

Two tests added: one verifying a custom type is compressed when `shouldCompress` returns `true`, one verifying nothing is compressed when it always returns `false`. All 21 tests pass.

### DIFF
--- a/src/middleware/compress/index.test.ts
+++ b/src/middleware/compress/index.test.ts
@@ -216,6 +216,48 @@ describe('Compress Middleware', () => {
   })
 })
 
+describe('shouldCompress option', () => {
+  it('should compress custom content types when shouldCompress returns true', async () => {
+    const app = new Hono()
+    app.use(
+      '*',
+      compress({
+        shouldCompress: (res) => {
+          const type = res.headers.get('Content-Type') ?? ''
+          return type.startsWith('application/vnd.msgpack')
+        },
+      })
+    )
+    app.get('/msgpack', (c) => {
+      c.header('Content-Type', 'application/vnd.msgpack')
+      c.header('Content-Length', '1024')
+      return c.body(new Uint8Array(1024))
+    })
+
+    const req = new Request('http://localhost/msgpack', {
+      headers: { 'Accept-Encoding': 'gzip' },
+    })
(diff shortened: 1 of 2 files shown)


usualoma:
Hi @mixelburg,
I'm sorry, and I don't think you mean any harm, but could you please stop creating pull requests in the middle of an ongoing discussion on an issue—especially when you aren't even participating in that discussion? I think it's fair to say that this behavior amounts to spam.

yusukebe:
@mixelburg, I'll block you from our project if you do it again, creating a PR that looks like spam, such as ignoring the context of the discussion on the issue.

PR #4953, opened 2026-05-20
feat: allow next to rethrow downstream errors

### LINKED ISSUE #4895: Allow middleware to intercept handler errors

Gauravkachwaha:
Closes #4895

## Summary

This PR adds an opt-in option to `next()` that allows downstream errors to be rethrown back to the middleware that called `await next()`.

```ts
await next({ rethrow: true })

### DIFF
--- a/src/compose.test.ts
+++ b/src/compose.test.ts
@@ -1,7 +1,7 @@
 import { compose } from './compose'
 import { Context } from './context'
 import type { Params } from './router'
-import type { Next } from './types'
+import type { ErrorHandler, Next } from './types'
 
 type MiddlewareTuple = [[Function, unknown], Params]
 
@@ -48,6 +48,7 @@ describe('compose', () => {
     expect(context.get('log')).toBe('log message')
     expect(context.get('xxx')).toBe('yyy')
   })
+
   it('Response', async () => {
     const composed = compose(middleware)
     const context = await composed(new Context(new Request('http://localhost/')))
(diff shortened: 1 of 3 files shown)


yusukebe:
We should discuss more before creating a PR. This PR is ignoring it. Slop Closing.

Gauravkachwaha:
Thanks for the clarification @yusukebe.

I’m sorry for opening this PR too early before the maintainer-side design discussion was settled. That was my mistake.

I understand that I should have waited for maintainer agreement on the approach before submitting an implementation.

I’ll continue the discussion in #4895 and won’t open another PR unless the maintainers agree on the direction first.

PR #4956, opened 2026-05-21
feat(client): infer DetailedError response types

proto-atlas:


yusukebe:
The discussion is not finished. The PR, which ignores the context of the issue and relies mainly on AI, is not welcome. Closing.

PR #4961, opened 2026-05-23
feat(compress): add contentTypeFilter option and `COMPRESSIBLE_CONTENT_TYPE_REGEX` re-export

### LINKED ISSUE #4946: Support custom Content-Type in compress middleware

na-trium-144:
close #4946 

* Added `contentTypeFilter` option to compress middleware that allows users to set custom conditions for which Content-Types are allowed to be compressed.
* Added re-export of `COMPRESSIBLE_CONTENT_TYPE_REGEX` in middleware/compress/index.ts so users can import it from "hono/compress" to use with contentTypeFilter option.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/compress/index.test.ts
+++ b/src/middleware/compress/index.test.ts
@@ -245,6 +245,50 @@ describe('Compress Middleware', () => {
     })
   })
 
+  describe('contentTypeFilter', () => {
+    describe('RegExp', () => {
+      const app = new Hono()
+      app.use('*', compress({ contentTypeFilter: /^application\/json/, threshold: 0 }))
+      app.get('/json', (c) => c.json({ data: 'a'.repeat(1024) }))
+      app.get('/text', (c) => c.text('a'.repeat(1024)))
+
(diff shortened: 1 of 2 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@na-trium-144 Thank you!

PR #4963, opened 2026-05-23
refactor(bearer-auth): collapse throwHTTPException args into an options object

francisjohnjohnston-web:
…ns object

The internal `throwHTTPException` helper took two adjacent parameters of the same type — `wwwAuthenticateHeader` and `messageOption`, both `string | object | MessageFunction`. Positional arguments of identical type can be transposed by a future edit with no compiler error, silently swapping the WWW-Authenticate challenge with the response body. Grouping them into a named `AuthFailureOptions` object makes all three call sites self-documenting and removes that latent footgun.

Behaviour-preserving: pure signature refactor of a non-exported closure; no public API change. All 49 bearer-auth tests pass; tsc --noEmit, eslint, and prettier all clean.

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/bearer-auth/index.ts
+++ b/src/middleware/bearer-auth/index.ts
@@ -121,9 +121,11 @@ export const bearerAuth = <E extends Env = Env>(
 
   const throwHTTPException = async (
     c: Context,
-    status: ContentfulStatusCode,
-    wwwAuthenticateHeader: string | object | MessageFunction,
-    messageOption: string | object | MessageFunction
+    { status, wwwAuthenticateHeader, message }: {
+      status: ContentfulStatusCode
+      wwwAuthenticateHeader: string | object | MessageFunction
+      message: string | object | MessageFunction
+    }
   ): Promise<Response> => {
     const wwwAuthenticateHeaderValue: string | object =
       typeof wwwAuthenticateHeader === 'function'
@@ -138,8 +140,7 @@ export const bearerAuth = <E extends Env = Env>(
               .map(([key, value]) => `${key}="${value}"`)
               .join(',')}`,
     }
-    const responseMessage =
-      typeof messageOption === 'function' ? await messageOption(c) : messageOption
+    const responseMessage = typeof message === 'function' ? await message(c) : message
     const res =
       typeof responseMessage === 'string'
         ? new Response(responseMessage, { status, headers })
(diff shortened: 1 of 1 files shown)


src/middleware/bearer-auth/index.ts
```diff
@@ -19,6 +19,12 @@ type CustomizedErrorResponseOptions = {
   message?: string | object | MessageFunction
 }
 
+type AuthFailureOptions = {
```
yusukebe: It's a matter of personal taste, but I think `AuthFailureOptions` isn't needed since it is used only in one place.

yusukebe:
Hey @francisjohnjohnston-web 

I think this refactoring is good. I left one comment. Check it.

src/middleware/bearer-auth/index.ts
francisjohnjohnston-web: Good point — happy to remove the named type and inline it. Will update.

francisjohnjohnston-web: Done — the anonymous type is now inline in the destructuring signature. The force-push removed the named alias.

BlankParticle:
@yusukebe @usualoma 
I should let you know that the author of this PR is probably an autonomous AI agent from what I see
even if the PRs from this author are fine, they don't disclose this anywhere

For starters the @francisjohnjohnston-web has no history, but suddenly started making bunch of PRs to valibot, h3, hono etc
the commits are made under the name `Prodia Evolution Worker` with email `evo@prodia.dev`

https://patch-diff.githubusercontent.com/raw/honojs/hono/pull/4969.patch

https://prodia.dev/ looks like some kind of autonomous AI thing

I am letting you know about this because of how much spam PRs you guys are dealing with

yusukebe:
@BlankParticle

Thanks, Rahul! Regarding this, I think the account is "not bad" because the RP description following our PR template includes `The author should do the following, if applicable`. But I found the account creates a lot of PRs when I wake up. You are right, actually spams. I don't have to accept and should block it.

usualoma:
This is a tricky one.

I get the sense that @francisjohnjohnston-web ’s PRs so far have been created with significant AI assistance, but I feel they’ve been created to address the right issues. (There are some minor points that aren’t really necessary.)

For #4966, for example, the content is actually something that should be included as-is. If we close it (which is unavoidable since it’s OSS), we might end up having to implement the same thing ourselves later, which is a bit of a dilemma. (We could cherry-pick this commit and include it as-is, though.)

yusukebe:
@usualoma 

> but I feel they’ve been created to address the right issues.

I also thought this, as you do. But the commit by the account will be included in our repo, which is not welcome. So it is better to use a cherry-pick commit or create another PR with the same code on our side.

To do it. I started creating an AI Usage Policy to handle these AI-related matters. The first one is #4970

BlankParticle:
@usualoma I agree with the point that the PRs are fixing real problems, but we shouldn't encourage anyone to just open bunch of PRs even with help of AI

It's increasing burden on maintainers and makes it easier to overlook for mistakes or even sneaky malicious code (like in case xz, maintenaners burnt out)
We should be at minimum require AI disclosures and be strict about spam PRs and issues in my opinion

yusukebe:
@BlankParticle 

I think @usualoma already knows that problem well! This case is a little tricky.

And I agree with your opinion.

PR #4965, opened 2026-05-24
refactor: prefer startsWith over str[0] === 'c' in helper/util files

francisjohnjohnston-web:
## Summary

Replaces `str[0] === 'c'` / `str[0] !== 'c'` character-index comparisons with `str.startsWith('c')` in helper and utility files. This is the pattern flagged by ESLint's `prefer-string-starts-ends-with` rule (currently disabled in `eslint.config.mjs`).

## Changes

| File | Before | After |
|------|--------|-------|
| `src/helper/css/index.ts` | `className[0] === '@'` | `className.startsWith('@')` |
| `src/jsx/dom/css.ts` | `className[0] === '@'` | `className.startsWith('@')` |
| `src/utils/filepath.ts` | `root[0] !== '/'`, `path[0] === '/'` | `root.startsWith('/')`, `path.startsWith('/')` |
| `src/helper/ssg/utils.ts` | `paths[0][0] === '/'` | `paths[0].startsWith('/')` |
| `src/jsx/utils.ts` | `k[0] === '-'` | `k.startsWith('-')` |
| `src/utils/url.ts` | 4 index comparisons in `getPattern`/`mergePath` | `startsWith` equivalents |

All 6 files are utility/helper code — not request-dispatch hot paths. `tsc --noEmit` passes cleanly.

**Excluded intentionally:** `src/router/trie-router/node.ts` (per-request path offset — `str[0] === '/'` is marginally faster in V8 for that hot loop); `src/middleware/secure-headers/secure-headers.ts` (inside a `length === 1 &&` guard where partial rewrite would create mixed style in the same expression).

## Motivation

`str.startsWith('c')` expresses intent more clearly than `str[0] === 'c'` and avoids the silent `undefined` return on empty strings. The index approach silently returns `undefined` (falsy, so usually harmless) when the string is empty; `startsWith` returns `false`. For the cases here — CSS class names, file paths, URL segments — that difference matters for robustness.

### DIFF
--- a/src/helper/css/index.ts
+++ b/src/helper/css/index.ts
@@ -98,7 +98,7 @@ export const createCssContext = ({
         added[className] = true
         stylesStr += className.startsWith(PSEUDO_GLOBAL_SELECTOR)
           ? toAdd[className]
-          : `${className[0] === '@' ? '' : '.'}${className}{${toAdd[className]}}`
+          : `${className.startsWith('@') ? '' : '.'}${className}{${toAdd[className]}}`
       })
       contextMap.set(context, [{}, added])
 
--- a/src/helper/ssg/utils.ts
+++ b/src/helper/ssg/utils.ts
@@ -51,7 +51,7 @@ export const joinPaths = (...paths: string[]): string => {
   paths = paths.map(normalizePath)
   const resultPaths: string[] = []
   handleSegments(paths.join('/').split('/'), resultPaths)
-  return (paths[0][0] === '/' ? '/' : '') + resultPaths.join('/')
+  return (paths[0].startsWith('/') ? '/' : '') + resultPaths.join('/')
 }
 
 interface FilterStaticGenerateRouteData {
--- a/src/jsx/dom/css.ts
+++ b/src/jsx/dom/css.ts
@@ -105,7 +105,7 @@ export const createCssJsxDomObjects: CreateCssJsxDomObjectsType = ({ id }) => {
       addedStyles.add(className)
       ;(className.startsWith(PSEUDO_GLOBAL_SELECTOR)
         ? splitRule(styleString)
-        : [`${className[0] === '@' ? '' : '.'}${className}{${styleString}}`]
+        : [`${className.startsWith('@') ? '' : '.'}${className}{${styleString}}`]
       ).forEach((rule) => {
         sheet.insertRule(rule, sheet.cssRules.length)
       })
--- a/src/jsx/utils.ts
+++ b/src/jsx/utils.ts
@@ -172,7 +172,7 @@ export const styleObjectForEach = (
 ): void => {
   for (const [k, v] of Object.entries(style)) {
     const key =
-      k[0] === '-' || !/[A-Z]/.test(k)
+      k.startsWith('-') || !/[A-Z]/.test(k)
         ? k // a CSS variable or a lowercase only property
         : k.replace(/[A-Z]/g, (m) => `-${m.toLowerCase()}`) // a camelCase property. convert to kebab-case
     if (!isValidStylePropertyName(key)) {
--- a/src/utils/filepath.ts
+++ b/src/utils/filepath.ts
@@ -52,7 +52,7 @@ export const getFilePathWithoutDefaultDocument = (
   let path = root ? root + '/' + filename : filename
   path = path.replace(/^\.?\//, '')
 
-  if (root[0] !== '/' && path[0] === '/') {
+  if (!root.startsWith('/') && path.startsWith('/')) {
     return
   }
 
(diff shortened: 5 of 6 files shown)


usualoma:
Hi @francisjohnjohnston-web,
Thank you for creating a pull request.

I benchmarked `s[0] === '@'`, `s.startsWith('@')`, `s.charCodeAt(0) === 64`, and `s.charAt(0) === '@'` on Node.js, Bun, and Deno.

Environment:

- macOS arm64
- Node.js v24.14.1
- Bun 1.3.12
- Deno 2.7.14

Benchmark shape:

- pre-generated 1024 className-like strings
- 33,554,432 iterations
- 9 samples
- median ns/op

Mixed dataset result:

| runtime | fastest | `s[0] === '@'` | `s.startsWith('@')` | `s.charCodeAt(0) === 64` | `s.charAt(0) === '@'` |
|---|---:|---:|---:|---:|---:|
| Node.js | `startsWith` | 2.856 ns/op | **1.527 ns/op** | 2.734 ns/op | 3.506 ns/op |
| Bun | `charAt` | 1.328 ns/op | 3.398 ns/op | 1.943 ns/op | **1.171 ns/op** |
| Deno | `startsWith` | 1.820 ns/op | **1.115 ns/op** | 2.697 ns/op | 2.868 ns/op |

Full dataset summary:

| runtime | result |
|---|---|
| Node.js | `startsWith('@')` was fastest in all tested cases. |
| Bun | `startsWith('@')` was consistently the slowest; `s[0]` / `charAt(0)` were fastest. |
| Deno | `startsWith('@')` was fastest in all tested cases. |

So, on current V8, `s.startsWith('@')` does seem to be the fastest option.

However, on JavaScriptCore, it seems significantly slower. Since Bun is an important runtime for Hono, this is a bit of a trade-off.

I'm a little torn, but for now I think we don't need to replace this with `startsWith('@')`.

<details>
<summary>Benchmark script</summary>

```js
const N = Number(globalThis.Deno?.env.get('N') ?? globalThis.process?.env.N ?? 33_554_432)
const SAMPLES = Number(globalThis.Deno?.env.get('SAMPLES') ?? globalThis.process?.env.SAMPLES ?? 9)
const SIZE = 1024

function makeData(kind) {
  const data = new Array(SIZE)
  for (let i = 0; i < SIZE; i++) {
    if (kind === 'all false') data[i] = `css-${i.toString(36)}-hover:bg-red`
    else if (kind === 'all true') data[i] = `@media-${i.toString(36)}`
    else if (kind === 'mixed') data[i] = i % 2 === 0 ? `@supports-${i.toString(36)}` : `css-${i.toString(36)}-btn`
    else data[i] = i % 4 === 0 ? '' : i % 2 === 0 ? `@layer-${i.toString(36)}` : `css-${i.toString(36)}`
  }
  return data
}

const datasets = {
  'all false': makeData('all false'),
  'all true': makeData('all true'),
  mixed: makeData('mixed'),
  'with empty': makeData('with empty'),
}

const methods = [
  ['s[0] === "@"', (data, n) => {
    let count = 0
    for (let i = 0; i < n; i++) if (data[i & 1023][0] === '@') count++
    return count
  }],
  ['s.startsWith("@")', (data, n) => {
    let count = 0
    for (let i = 0; i < n; i++) if (data[i & 1023].startsWith('@')) count++
    return count
  }],
  ['s.charCodeAt(0) === 64', (data, n) => {
    let count = 0
    for (let i = 0; i < n; i++) if (data[i & 1023].charCodeAt(0) === 64) count++
    return count
  }],
  ['s.charAt(0) === "@"', (data, n) => {
    let count = 0
    for (let i = 0; i < n; i++) if (data[i & 1023].charAt(0) === '@') count++
    return count
  }],
]

function median(values) {
  return [...values].sort((a, b) => a - b)[Math.floor(values.length / 2)]
}

for (const data of Object.values(datasets)) {
  for (const [, fn] of methods) fn(data, 3_000_000)
}

console.log(`N=${N} SAMPLES=${SAMPLES} SIZE=${SIZE}`)
console.log(['dataset', 'method', 'median ns/op', 'ops/sec', 'relative to fastest'].join('\t'))

for (const [datasetName, data] of Object.entries(datasets)) {
  const rows = []
  for (const [methodName, fn] of methods) {
    const times = []
    for (let sample = 0; sample < SAMPLES; sample++) {
      fn(data, N)
      times.push(((performance.now() - start) * 1_000_000) / N)
    }
    rows.push({ methodName, ns: median(times) })
  }

  const fastest = Math.min(...rows.map((row) => row.ns))
  rows.sort((a, b) => a.ns - b.ns)

  for (const row of rows) {
    console.log([
      datasetName,
      row.methodName,
      row.ns.toFixed(3),
      Math.round(1_000_000_000 / row.ns),
      `${(row.ns / fastest).toFixed(2)}x`,
    ].join('\t'))
  }
}

francisjohnjohnston-web:
Withdrawing this one — on reflection it's a purely cosmetic readability change with marginal value, and I'd rather not add review noise. Thanks for your time!

PR #4966, opened 2026-05-24
fix: RFC 5952 §4.2.2 — do not compress a single 0 group to ::

francisjohnjohnston-web:
## Bug

`convertIPv6BinaryToString` compresses a **single** 16-bit zero field to `::`, violating RFC 5952 §4.2.2.

RFC 5952 §4.2.2 states:
> The longest run of consecutive 16-bit 0 fields MUST be shortened. **When there is an alternative choice in the placement of a "::", the first sequence of zero bits MUST be shortened. For example, 2001:db8:0:0:1:0:0:1 is rendered as 2001:db8::1:0:0:1 rather than 2001:db8:0:0:1::1.** The symbol "::" MUST NOT be used to shorten just one 16-bit 0 field.

Current code:
```ts
if (maxZeroStart !== -1) {
  sections.splice(maxZeroStart, maxZeroEnd - maxZeroStart, ':')
}
```

This fires even when the zero run is exactly one group, producing `1:0:2:3:4:5:6:7` → `1::2:3:4:5:6:7` instead of the correct unexpanded form `1:0:2:3:4:5:6:7`.

## Impact

`convertIPv6BinaryToString` is used to normalise IP addresses for ip-restriction middleware rule matching. An IPv6 address with a single interior zero group would be stored in the rule set in the wrong (RFC-violating) form, causing mismatches when the rule key is compared against an incoming address normalised by a compliant library.

## Fix

```ts
if (maxZeroStart !== -1 && maxZeroEnd - maxZeroStart > 1) {
```

Guard added: only compress runs of **more than one** consecutive zero field.

## Tests

Added four cases to the existing parameterised round-trip test:
- single interior zero (`1:0:2:3:4:5:6:7`) — must remain unexpanded
- leading single zero (`0:1:2:3:4:5:6:7`) — must remain unexpanded
- trailing single zero (`1:2:3:4:5:6:7:0`) — must remain unexpanded
- two-group run (`1:0:0:2:3:4:5:6`) — still compressed correctly (`1::2:3:4:5:6`)

All 53 existing tests pass.

### DIFF
--- a/src/utils/ipaddr.ts
+++ b/src/utils/ipaddr.ts
@@ -355,7 +355,7 @@ export const convertIPv6BinaryToString = (ipV6: bigint): string => {
       maxZeroEnd = 8
     }
   }
-  // RFC 5952 4.2.2: "::" MUST NOT be used to shorten just one 16-bit 0 field.
+  // RFC 5952 4.2.2: compress only zero runs of at least two 16-bit fields.
   if (maxZeroStart !== -1 && maxZeroEnd - maxZeroStart > 1) {
     sections.splice(maxZeroStart, maxZeroEnd - maxZeroStart, ':')
   }
--- a/src/utils/ipaddr.ts
+++ b/src/utils/ipaddr.ts
@@ -355,7 +355,8 @@ export const convertIPv6BinaryToString = (ipV6: bigint): string => {
       maxZeroEnd = 8
     }
   }
-  if (maxZeroStart !== -1) {
+  // RFC 5952 4.2.2: "::" MUST NOT be used to shorten just one 16-bit 0 field.
+  if (maxZeroStart !== -1 && maxZeroEnd - maxZeroStart > 1) {
     sections.splice(maxZeroStart, maxZeroEnd - maxZeroStart, ':')
   }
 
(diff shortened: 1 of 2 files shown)


src/utils/ipaddr.ts
```diff
@@ -355,7 +355,8 @@ export const convertIPv6BinaryToString = (ipV6: bigint): string => {
       maxZeroEnd = 8
     }
   }
-  if (maxZeroStart !== -1) {
+  // RFC 5952 4.2.2: "::" MUST NOT be used to shorten just one 16-bit 0 field.
```
usualoma: nit: Could we phrase this as the positive invariant while keeping the RFC reference?

```suggestion
  // RFC 5952 4.2.2: compress only zero runs of at least two 16-bit fields.
```

usualoma (commented):
Thank you for creating this pull request!

Yes, this is exactly right. I had meant to fix this and forgot about it. The fix looks good to me, but I left one comment, so please take a look.

One note: `ip-restriction` no longer depends on canonical string representations after the fix for the following issue, so this particular impact does not apply anymore:
https://github.com/honojs/hono/security/advisories/GHSA-xrhx-7g5j-rcj5

usualoma:
Thank you!

PR #4973, opened 2026-05-25
fix(utils/ipaddr): expand "::" to eight zero groups

ogyamada:
### What this fixes

`expandIPv6('::')` returns **14** zero groups instead of 8:

```ts
expandIPv6('::')
// actual:   '0000:0000:0000:0000:0000:0000:0000:0000:0000:0000:0000:0000:0000:0000'
// expected: '0000:0000:0000:0000:0000:0000:0000:0000'
```

`::` is the IPv6 unspecified / all-zeros address, so this is a valid input.

### Cause

`'::'.split(':')` produces three empty sections `['', '', '']`. In the expansion
loop, the empty branch removed only a **single** adjacent empty section before
filling with zeros:

```ts
sections[i + 1] === '' && sections.splice(i + 1, 1)
```

For `::` that leaves a second empty slot behind, which is then expanded again on
the next iteration, doubling the zero groups. Other inputs were unaffected
because a single `::` elsewhere in an address yields at most two adjacent
empties.

### Fix

Drain **all** consecutive empty sections before filling, so exactly one slot
remains for the zero groups.

- [x] Add tests
- [x] Run tests

> Verified locally with `tsc --noEmit`, `vitest --run src/utils/ipaddr.test.ts` (41 passed), Prettier, and ESLint. Existing `expandIPv6` cases (`::1`, `2001:2::`, `2001:0:0:db8::1`, `::ffff:127.0.0.1`, …) continue to pass; added regression tests for `::` and `fe80::`.

### DIFF
--- a/src/utils/ipaddr.test.ts
+++ b/src/utils/ipaddr.test.ts
@@ -28,6 +28,10 @@ describe('expandIPv6', () => {
     expect(expandIPv6('2001:0:0:db8::1')).toBe('2001:0000:0000:0db8:0000:0000:0000:0001')
     expect(expandIPv6('::ffff:127.0.0.1')).toBe('0000:0000:0000:0000:0000:ffff:7f00:0001')
   })
+
+  it('Should expand the unspecified address "::" to eight zero groups', () => {
+    expect(expandIPv6('::')).toBe('0000:0000:0000:0000:0000:0000:0000:0000')
+  })
 })
 describe('distinctRemoteAddr', () => {
   it('Should result be valid', () => {
--- a/src/utils/ipaddr.ts
+++ b/src/utils/ipaddr.ts
@@ -26,7 +26,10 @@ export const expandIPv6 = (ipV6: string): string => {
     if (node !== '') {
       sections[i] = node.padStart(4, '0')
     } else {
-      sections[i + 1] === '' && sections.splice(i + 1, 1)
+      // Keep a single empty slot for `::` zero expansion.
+      while (sections[i + 1] === '') {
+        sections.splice(i + 1, 1)
+      }
       sections[i] = new Array(8 - sections.length + 1).fill('0000').join(':')
     }
   }


src/utils/ipaddr.ts
```diff
@@ -26,7 +26,13 @@ export const expandIPv6 = (ipV6: string): string => {
     if (node !== '') {
       sections[i] = node.padStart(4, '0')
     } else {
-      sections[i + 1] === '' && sections.splice(i + 1, 1)
+      // Drain every consecutive empty section produced by "::". A lone "::"
+      // splits into ['', '', ''] (three empties), so removing only the next one
+      // would leave a second empty that gets expanded again, yielding too many
+      // groups. Removing them all leaves a single slot to fill with zeros.
```
usualoma: Nit: this comment feels a bit longer than the code needs. Maybe shorten it to something like “Keep a single empty slot for `::` zero expansion”; the `::` regression test already documents the detailed case.

src/utils/ipaddr.test.ts
```diff
@@ -28,6 +28,14 @@ describe('expandIPv6', () => {
...
     expect(expandIPv6('::ffff:127.0.0.1')).toBe('0000:0000:0000:0000:0000:ffff:7f00:0001')
   })
+
+  it('Should expand the unspecified address "::" to eight zero groups', () => {
+    expect(expandIPv6('::')).toBe('0000:0000:0000:0000:0000:0000:0000:0000')
+  })
+
+  it('Should expand link-local "fe80::" correctly', () => {
+    expect(expandIPv6('fe80::')).toBe('fe80:0000:0000:0000:0000:0000:0000:0000')
+  })
```
usualoma: Could you remove this test case? It is already covered by the existing trailing `::` expansion test (`2001:2::`), and `expandIPv6()` does not treat link-local addresses specially.

usualoma (commented):
Hi @youcefzemmar 
Thanks for opening this pull request. The change looks reasonable to me.
I left two comments. Could you take a look when you have a chance?

src/utils/ipaddr.ts
ogyamada: Done — shortened to a single line: `// Keep a single empty slot for `'::'` zero expansion.` Thanks!

src/utils/ipaddr.test.ts
ogyamada: Removed. Agreed — the trailing `::` path is already covered by `2001:2::`, and `expandIPv6()` does not treat link-local specially. Kept the `::` case since that is the one this PR fixes.

usualoma:
@youcefzemmar Thank you! LGTM.

yusukebe (approved):
LGTM!

yusukebe:
@youcefzemmar Thank you! I'll merge this later.

PR #4980, opened 2026-05-30
refactor(language): Test/improve tests on languages middleware

iNeoO:

- [x] Add tests
- [x] Run tests

Hello

I removed unreachable fallback code from the language detector.

The middleware already validates detector options before running detection, so `detectLanguage` does not need to guard against an unknown detector. Individual detector failures are already caught inside `detectLanguage`, so `languageDetector` does not need a second fallback catch around it.

Also removes the `try/catch` from path detection. `c.req.url` should be a valid URL, and `normalizeLanguage` already returns `undefined` when the path segment cannot produce a supported language.

- Remove the redundant `detectFromPath` fallback catch
- Remove the unreachable missing-detector guard in `detectLanguage`
- Remove the redundant outer fallback catch in `languageDetector`
- Add some test to reach 100% coverage on language.ts middleware

If you disagree with these removals, I can revert them and keep this PR focused on the tests only.

### DIFF
--- a/src/middleware/language/index.test.ts
+++ b/src/middleware/language/index.test.ts
@@ -166,6 +166,23 @@ describe('languageDetector', () => {
       })
       expect(await res.text()).toBe('en')
     })
+
+    it('should handle invalid header lookup key gracefully', async () => {
+      const app = createTestApp({
+        order: ['header'],
+        lookupFromHeaderKey: 'accept\nlanguage',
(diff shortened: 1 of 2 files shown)


yusukebe (approved):
LGTM!

yusukebe:
Hey @iNeoO !

This is a good refactoring and adding tests. Looks good. Thank you!

PR #4988, opened 2026-06-01
fix(utils/stream): re-acquire writer lock when pipe() throws

### LINKED ISSUE #4981: StreamingApi.pipe(): WritableStream writer lock permanently lost when pipeTo() throws
## Summary

`StreamingApi.pipe()` releases the `WritableStreamDefaultWriter` lock before calling `pipeTo()`, but only re-acquires a new writer on the success path. If `pipeTo()` rejects — for example when the client disconnects, the upstream `ReadableStream` errors, or an abort signal fires — the `getWriter()` call on line 81 is never reached. `this.writer` is left pointing to the already-released (stale) writer, and `this.writable` has no active writer attached to it.

Any subsequent call to `write()`, `writeln()`, or `close()` on the same `StreamingApi` instance then hits one of two failure modes:

* `TypeError: This WritableStream writer has been released and cannot be used to monitor the stream's state.` (from the stale writer), or
* `TypeError: Failed to execute 'getWriter' on 'WritableStream': The stream is locked to a writer.` if `pipeTo()` left the stream in a locked-but-abandoned state.

Both outcomes silently break the stream for the caller. The error handlers registered with `onAbort()` or the outer `stream()` wrapper do not restore the writer, so post-pipe writes are dead with no diagnostic.

## Affected Code

`src/utils/stream.ts`, lines 78-82 (v4.12.23, present since pipe was introduced):

```ts
  this.writer.releaseLock()                            // lock released here
  await body.pipeTo(this.writable, { preventClose: true })  // throws → jumps out
  this.writer = this.writable.getWriter()             // never reached on throw
}
```

## Root Cause

The method has no `try/finally` block. `releaseLock()` is called unconditionally before the `await`, but `getWriter()` is only called after the `await` succeeds. Any rejection from `pipeTo()` leaves the class in an inconsistent state: `this.writer` is a released writer object, and `this.writable` is unlocked but the class no longer holds its reference.

## Steps to Reproduce

```ts
import { Hono } from 'hono'
import { stream } from 'hono/streaming'

const app = new Hono()

app.get('/test', (c) =>
  stream(c, async (s) => {
    // An erroring ReadableStream simulates a client disconnect or upstream error
    const erroringStream = new ReadableStream({
      start(ctrl) { ctrl.error(new Error('upstream failure')) },
    })

    try {
      await s.pipe(erroringStream)    // throws – pipeTo rejects
    } catch {
      // writer is now permanently lost
    }

    // This write silently fails or throws because this.writer is stale
    await s.write('this line is never delivered')
    await s.close()
  })
)
```

Run with any Hono-compatible runtime; the response body will be empty or the server will log an unhandled `TypeError` depending on runtime error propagation.

## Impact

Any handler that calls `pipe()` on a body that can fail (network interruption, upstream service error, client abort) will silently lose all subsequent `write()` / `writeln()` / `close()` calls. This includes:

* SSE handlers that fall back to a text message on pipe failure.
* Streaming proxy responses where the upstream may time out.
* Handlers that write a trailing chunk or close notification after pipe returns.

The affected code path is reached on every client disconnect during an active `pipe()`, which is a normal browser behavior (tab close, navigation away). In production, this manifests as truncated or empty streaming responses with no error surfaced to the application layer.

## Suggested Fix

Wrap the `pipeTo` call in a `try/finally` and always re-acquire the writer lock before returning, regardless of success or failure:

```ts
  this.writer.releaseLock()
  try {
    await body.pipeTo(this.writable, { preventClose: true })
  } finally {
    // Always re-acquire the writer so subsequent write()/close() calls
    // work correctly even when pipeTo() rejects (client disconnect, etc.)
    this.writer = this.writable.getWriter()
  }
}
```

This mirrors the pattern already used in `write()` and `close()`, which both use `try/catch` to ensure the class stays in a consistent state after errors.

Sriharsha-dev369:
### Problem

`pipe()` calls `releaseLock()` before `await pipeTo()` but only re-acquires
the writer on the success path. If `pipeTo()` rejects (source error, client
disconnect), the writer is left released and subsequent `write()` / `close()`
calls are silently lost.

### Fix

- Wraps `pipeTo()` in `try/finally` so the writer is always re-acquired
- Adds `preventAbort: true` so the destination stays writable after a source
  error — without it, the re-acquired writer targets an aborted stream and
  writes are still dropped

### Tests

The added tests assert that a `write()` after a failed `pipe()` is delivered;
they fail without either part of the fix.

---

Fixes #4981

### Checklist

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/utils/stream.test.ts
+++ b/src/utils/stream.test.ts
@@ -57,6 +57,44 @@ describe('StreamingApi', () => {
     expect((await reader.read()).value).toEqual(new TextEncoder().encode('bar'))
   })
 
+  it('pipe() re-acquires writer lock when pipeTo() throws', async () => {
+    const { readable, writable } = new TransformStream()
+    const api = new StreamingApi(writable, readable)
+    const reader = api.responseReadable.getReader()
+
+    const erroringReadable = new ReadableStream({
+      start(controller) {
+        controller.error(new Error('upstream error'))
+      },
+    })
+
+    await expect(api.pipe(erroringReadable)).rejects.toThrow('upstream error')
+    api.write('after pipe')
+    expect((await reader.read()).value).toEqual(new TextEncoder().encode('after pipe'))
+  })
+
+  it('pipe() can be called again after a previous pipe() throws', async () => {
+    const { readable, writable } = new TransformStream()
+    const api = new StreamingApi(writable, readable)
+    const reader = api.responseReadable.getReader()
+
+    const erroringReadable = new ReadableStream({
+      start(controller) {
+        controller.error(new Error('upstream error'))
+      },
+    })
+    await expect(api.pipe(erroringReadable)).rejects.toThrow('upstream error')
+
+    const src = new ReadableStream({
+      start(controller) {
+        controller.enqueue(new TextEncoder().encode('second pipe'))
+        controller.close()
+      },
+    })
+    await api.pipe(src)
+    expect((await reader.read()).value).toEqual(new TextEncoder().encode('second pipe'))
+  })
+
   it('close()', async () => {
     const { readable, writable } = new TransformStream()
     const api = new StreamingApi(writable, readable)
--- a/src/utils/stream.ts
+++ b/src/utils/stream.ts
@@ -77,8 +77,11 @@ export class StreamingApi {
 
   async pipe(body: ReadableStream) {
     this.writer.releaseLock()
-    await body.pipeTo(this.writable, { preventClose: true })
-    this.writer = this.writable.getWriter()
+    try {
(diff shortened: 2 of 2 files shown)


Sriharsha-dev369:
Pushed an updated commit to fix the Codecov patch coverage — added a test covering the success path of finally. Could you approve the CI run? Thanks!

yusukebe (approved):
LGTM!

yusukebe:
Hi @Sriharsha-dev369 

I'm sorry to be late. Looks good. Thanks!

PR #4994, opened 2026-06-02
fix(context): preserve prepared Set-Cookie headers when assigning raw Response

### LINKED ISSUE #4992: set res() setter loses Set-Cookie headers when middleware sets cookies before next() and handler returns a raw Response

4.12.18

Node.js v24.1.0

When a middleware sets a `Set-Cookie` header via `c.header()` before `await next()`, and the downstream handler returns a **raw `Response`** object with its own `Set-Cookie` headers, one side's cookies are always lost — they are never merged correctly.

**Minimal reproduction** (no third-party dependencies):

```js
import { Hono } from 'hono'

const app = new Hono()

// Middleware: access c.res (e.g. CORS does this), then set a cookie
app.use(async (c, next) => {
  c.res.headers.set('X-Something', 'true') // triggers #res creation
  c.header('Set-Cookie', 'mw_cookie=hello; Path=/', { append: true })
  await next()
})

// Handler: returns a raw Response with its own Set-Cookie
app.get('/test', (c) => {
  const res = new Response('ok')
  res.headers.append('set-cookie', 'handler_cookie=world; Path=/; HttpOnly')
  return res
})

console.log(res.headers.getSetCookie())
// Actual:   ['mw_cookie=hello; Path=/']
// Expected: ['mw_cookie=hello; Path=/', 'handler_cookie=world; Path=/; HttpOnly']
```

Both `Set-Cookie` headers should be present in the final response:
- `mw_cookie=hello` (set by middleware via `c.header()`)
- `handler_cookie=world` (set by the handler's raw `Response`)

### What do you see instead?

Only one side's cookies survive, depending on whether `c.res` was accessed before `c.header()`:

| Scenario | `#res` created before `c.header()`? | Lost cookies |
|---|---|---|
| A: No middleware touches `c.res` early | No | `c.header()` cookies lost (merge skipped entirely) |
| B: Some middleware accesses `c.res` early (e.g. CORS) | Yes | Handler's cookies overwritten |

### Root cause

The bug is in the `set res()` setter in `src/context.ts`:

```ts
set res(_res) {
  if (this.#res && _res) {          // ← Scenario A: #res is falsy → entire merge is skipped
    _res = createResponseInstance(_res.body, _res)
    for (const [k, v] of this.#res.headers.entries()) {
      if (k === 'set-cookie') {
        const cookies = this.#res.headers.getSetCookie()
        _res.headers.delete('set-cookie')   // ← Scenario B: deletes handler's cookies
        for (const cookie of cookies) {
          _res.headers.append('set-cookie', cookie)  // ← only re-appends context's cookies
        }
      }
      // ...
    }
  }
  this.#res = _res
}
```

**Scenario A** — `this.#res` is `undefined` (no middleware accessed `c.res` before `c.header()`):
- `c.header()` writes to `#preparedHeaders`, but the `if (this.#res && _res)` guard is `false`
- `#preparedHeaders` are never merged into the handler's Response → middleware cookies lost

**Scenario B** — `this.#res` exists (e.g. CORS middleware called `c.res.headers.set()` earlier):
- `_res.headers.delete('set-cookie')` wipes ALL cookies from the handler's Response
- Only the context's cookies are re-appended → handler cookies lost

### Suggested fix

The merge logic should preserve cookies from **both** sides:

```ts
// this only fix Scenario B
if (k === 'set-cookie') {
  // Collect cookies from both sources before mutating
  const handlerCookies = _res.headers.getSetCookie()
  const contextCookies = this.#res.headers.getSetCookie()
  _res.headers.delete('set-cookie')
  for (const cookie of handlerCookies) {
    _res.headers.append('set-cookie', cookie)
  }
  for (const cookie of contextCookies) {
    _res.headers.append('set-cookie', cookie)
  }
}
```

**Additionally, Scenario A needs `#preparedHeaders` to be merged even when `#res` hasn't been created yet.**

### Real-world impact

This bug surfaces in common middleware combinations. For example, in our project:

```
CORS middleware (accesses c.res.headers)
  → languageDetector({ caches: true })   // calls setCookie() → c.header("Set-Cookie", ...)
    → better-auth handler                // returns raw Response with session Set-Cookie headers
```

With `caches: true`, the `languageDetector` sets a language cookie via `c.header()` before `next()`. When `better-auth`'s handler returns a raw `Response` with session cookies, the `set-cookie` merge in `set res()` causes cookies from one side to be dropped.

We currently work around this by setting `caches: false`, but this disables a legitimate feature. The issue is not specific to `languageDetector` or `better-auth` — any middleware that calls `c.header('Set-Cookie', ...)` before `next()` combined with any handler that returns a raw `Response` with `Set-Cookie` headers will trigger this bug.

xu91102:
Fixes https://github.com/honojs/hono/issues/4992

### What this fixes

When middleware prepares cookies before `next()` and the handler assigns a raw `Response` to `c.res`, the prepared `Set-Cookie` headers can be lost if `c.res` has not been initialized yet.

### Cause

The `Context.res` setter only merged headers from `this.#res`. In this path, cookies added by `setCookie(c, ...)` are still in `#preparedHeaders`, so assigning a raw `Response` skips them.

### Fix

- Preserve prepared `Set-Cookie` values when assigning a raw `Response` before `c.res` is initialized.
- Keep the existing overwrite behavior for ordinary prepared headers.
- Avoid duplicating cookies when `c.res` has already been read.
- Add regression tests for the lost-cookie path and the no-duplicate path.

- [x] Add tests
- [x] Run tests

Verified locally with:

- `npx vitest --run src/context.test.ts -t "raw Response"`
- `npx vitest --run src/hono.test.ts -t "Overwrite the response from middleware after next"`
- `npx vitest --run src/hono.test.ts src/context.test.ts`
- `npx eslint src/context.ts src/context.test.ts`
- `npx tsc --noEmit --pretty false`
- `npx prettier --check src/context.ts src/context.test.ts`
- `git diff --check`
- `Remove-Item Env:NO_COLOR -ErrorAction SilentlyContinue; npx vitest --run`

Full Vitest result: 144 files passed, 4525 tests passed, 33 skipped.

### DIFF
--- a/src/context.test.ts
+++ b/src/context.test.ts
@@ -436,6 +436,19 @@ describe('Context header', () => {
     expect(cookies.includes('foo2=bar2; Path=/')).toBe(true)
   })
 
+  it('Should keep prepared cookies when assigning a raw Response to c.res', () => {
+    setCookie(c, 'foo', 'bar', { path: '/' })
+    c.res = new Response('ok')
+    expect(c.res.headers.getSetCookie()).toEqual(['foo=bar; Path=/'])
+  })
+
+  it('Should not duplicate cookies when assigning a raw Response after reading c.res', () => {
+    setCookie(c, 'foo', 'bar', { path: '/' })
+    c.res
+    c.res = new Response('ok')
+    expect(c.res.headers.getSetCookie()).toEqual(['foo=bar; Path=/'])
+  })
+
   it('Should set set-cookie header values if c.res is already defined', () => {
     c.res = new Response(null, {
       headers: [
--- a/src/context.test.ts
+++ b/src/context.test.ts
@@ -442,11 +442,38 @@ describe('Context header', () => {
     expect(c.res.headers.getSetCookie()).toEqual(['foo=bar; Path=/'])
   })
 
+  it('Should keep raw Response cookies when assigning a Response with prepared cookies', () => {
+    setCookie(c, 'context', '1', { path: '/' })
+    const res = new Response('ok')
+    res.headers.append('set-cookie', 'handler=1; Path=/; HttpOnly')
+
+    c.res = res
+
+    expect(c.res.headers.getSetCookie()).toEqual([
+      'handler=1; Path=/; HttpOnly',
+      'context=1; Path=/',
+    ])
+  })
+
   it('Should not duplicate cookies when assigning a raw Response after reading c.res', () => {
     setCookie(c, 'foo', 'bar', { path: '/' })
     c.res
     c.res = new Response('ok')
     expect(c.res.headers.getSetCookie()).toEqual(['foo=bar; Path=/'])
+  })
+
+  it('Should keep raw Response cookies when assigning a Response after reading c.res', () => {
+    c.res.headers.set('X-Something', 'true')
+    setCookie(c, 'context', '1', { path: '/' })
+    const res = new Response('ok')
+    res.headers.append('set-cookie', 'handler=1; Path=/; HttpOnly')
+
+    c.res = res
+
+    expect(c.res.headers.getSetCookie()).toEqual([
+      'handler=1; Path=/; HttpOnly',
+      'context=1; Path=/',
+    ])
   })
 
   it('Should set set-cookie header values if c.res is already defined', () => {
--- a/src/context.ts
+++ b/src/context.ts
@@ -412,20 +412,23 @@ export class Context<
    * @param _res - The Response object to set.
    */
   set res(_res: Response | undefined) {
-    if (this.#res && _res) {
+    if ((this.#res || this.#preparedHeaders) && _res) {
       _res = createResponseInstance(_res.body, _res)
-      for (const [k, v] of this.#res.headers.entries()) {
-        if (k === 'content-type') {
-          continue
-        }
-        if (k === 'set-cookie') {
-          const cookies = this.#res.headers.getSetCookie()
-          _res.headers.delete('set-cookie')
-          for (const cookie of cookies) {
-            _res.headers.append('set-cookie', cookie)
+      const headers = this.#res?.headers ?? this.#preparedHeaders
+      if (headers) {
+        for (const [k, v] of headers.entries()) {
+          if (k === 'content-type' || (!this.#res && k !== 'set-cookie')) {
+            continue
           }
-        } else {
-          _res.headers.set(k, v)
+          if (k === 'set-cookie') {
+            const cookies = headers.getSetCookie()
+            _res.headers.delete('set-cookie')
+            for (const cookie of cookies) {
+              _res.headers.append('set-cookie', cookie)
+            }
+          } else {
+            _res.headers.set(k, v)
+          }
         }
       }
     }
--- a/src/context.ts
+++ b/src/context.ts
@@ -415,16 +415,23 @@ export class Context<
     if ((this.#res || this.#preparedHeaders) && _res) {
       _res = createResponseInstance(_res.body, _res)
       const headers = this.#res?.headers ?? this.#preparedHeaders
+      const isPreparedHeaders = !this.#res && !!this.#preparedHeaders
       if (headers) {
         for (const [k, v] of headers.entries()) {
-          if (k === 'content-type' || (!this.#res && k !== 'set-cookie')) {
+          if (k === 'content-type' || (isPreparedHeaders && k !== 'set-cookie')) {
             continue
           }
           if (k === 'set-cookie') {
-            const cookies = headers.getSetCookie()
-            _res.headers.delete('set-cookie')
-            for (const cookie of cookies) {
-              _res.headers.append('set-cookie', cookie)
+            if (isPreparedHeaders) {
+              _res.headers.append('set-cookie', v)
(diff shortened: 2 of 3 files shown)


techzealot:
Thanks for the fix! I tested it locally (hono 4.12.18, `dist/context.js` patched manually) and unfortunately **both scenarios still lose the handler's `Set-Cookie` headers**.

### Reproduction script

```js
import { Hono } from 'hono'

// ─── Scenario B: #res created before c.header() ───
{
  const app = new Hono()
  app.use(async (c, next) => {
    c.res.headers.set('X-Something', 'true') // triggers #res creation
    c.header('Set-Cookie', 'mw_cookie=hello; Path=/', { append: true })
    await next()
  })
  app.get('/test', (c) => {
    const res = new Response('ok')
    res.headers.append('set-cookie', 'handler_cookie=world; Path=/; HttpOnly')
    return res
  })
  console.log('[Scenario B]', res.headers.getSetCookie())
}

// ─── Scenario A: only #preparedHeaders, #res never created ───
{
  const app = new Hono()
  app.use(async (c, next) => {
    c.header('Set-Cookie', 'mw_cookie=hello; Path=/', { append: true })
    await next()
  })
  app.get('/test', (c) => {
    const res = new Response('ok')
    res.headers.append('set-cookie', 'handler_cookie=world; Path=/; HttpOnly')
    return res
  })
  console.log('[Scenario A]', res.headers.getSetCookie())
}
```

### Output with the official fix applied

```
[Scenario B] [ 'mw_cookie=hello; Path=/' ]          ❌ handler cookie lost
[Scenario A] [ 'mw_cookie=hello; Path=/' ]          ❌ handler cookie lost
```

### Root cause

The `set-cookie` branch in the fix still calls `headers.getSetCookie()` on the **context side only** (`this.#res.headers` or `this.#preparedHeaders`), then unconditionally `delete`s all set-cookie from `_res` — **`_res.headers.getSetCookie()` is never called**, so the handler's cookies are silently dropped:

```ts
if (k === 'set-cookie') {
  const cookies = headers.getSetCookie()   // ← only context's cookies
  _res.headers.delete('set-cookie')        // ← handler's cookies gone
  for (const cookie of cookies) {
    _res.headers.append('set-cookie', cookie) // ← only context's re-appended
  }
}
```

### Suggested correct fix

**Scenario B** — save handler's cookies before `delete`:

```ts
if (k === 'set-cookie') {
  const handlerCookies = _res.headers.getSetCookie()
  const contextCookies = this.#res.headers.getSetCookie()
  _res.headers.delete('set-cookie')
  for (const cookie of handlerCookies) _res.headers.append('set-cookie', cookie)
  for (const cookie of contextCookies) _res.headers.append('set-cookie', cookie)
}
```

**Scenario A** — use `append` instead of `delete` + `append`, since `#preparedHeaders` only holds context-side cookies:

```ts
if (k === 'set-cookie') {
  _res.headers.append('set-cookie', v)
}
```

Full corrected setter:

```ts
set res(_res: Response | undefined) {
  if (this.#res && _res) {
    _res = createResponseInstance(_res.body, _res)
    for (const [k, v] of this.#res.headers.entries()) {
      if (k === 'content-type') continue
      if (k === 'set-cookie') {
        const handlerCookies = _res.headers.getSetCookie()
        const contextCookies = this.#res.headers.getSetCookie()
        _res.headers.delete('set-cookie')
        for (const cookie of handlerCookies) _res.headers.append('set-cookie', cookie)
        for (const cookie of contextCookies) _res.headers.append('set-cookie', cookie)
      } else {
        _res.headers.set(k, v)
      }
    }
  } else if (this.#preparedHeaders && _res) {
    _res = createResponseInstance(_res.body, _res) 
    for (const [k, v] of this.#preparedHeaders.entries()) {
      if (k === 'content-type') continue
      if (k === 'set-cookie') {
        _res.headers.append('set-cookie', v)
      } else {
        _res.headers.set(k, v)
      }
    }
  }
  this.#res = _res
  this.finalized = true
}
```

### Output with the corrected fix

```
[Scenario B] [ 'handler_cookie=world; Path=/; HttpOnly', 'mw_cookie=hello; Path=/' ]  ✅
[Scenario A] [ 'handler_cookie=world; Path=/; HttpOnly', 'mw_cookie=hello; Path=/' ]  ✅
```

Both scenarios pass. Would appreciate a re-check on this — happy to provide any additional info.

xu91102:
Thanks for the detailed repro and suggested fix. I updated the PR in `d4975d1`.

What changed:
- For the `#preparedHeaders` path, `Set-Cookie` is now appended to the handler response instead of deleting/rebuilding it.
- For the `#res` path before the context is finalized, handler cookies are preserved first, then context cookies are appended.
- Existing finalized re-assignment behavior is kept, so the previous `c.res` cookie overwrite test still passes.

I also added regression coverage for both scenarios:
- `Context` unit tests for prepared headers and the `c.res`-already-created path.
- `Hono` request-level tests matching the raw `Response` handler cases.

Verified with:
- `npx vitest run src/context.test.ts`
- `npx vitest run src/hono.test.ts -t "Should keep raw response cookies"`
- `npx tsc --noEmit`
- `git diff --check`

PR #5005, opened 2026-06-04
feat(jwt,jwk): add realm option to customize WWW-Authenticate header

### LINKED ISSUE #4989: jwt and jwk middlewares hardcode req.url as WWW-Authenticate realm with no configurable override, unlike bearer-auth
## Description

The `jwt` and `jwk` built-in middlewares produce `401` responses with a `WWW-Authenticate` header whose `realm` is unconditionally set to the full request URL (`ctx.req.url`):

```
WWW-Authenticate: Bearer realm="https://api.example.com/v1/users/42?token=abc",error="invalid_token",error_description="token verification failure"
```

RFC 6750 defines `realm` as a human-readable string that identifies the protection space — not the request URL. Reflecting the full URL here is a poor default: it conflates the request path (which varies per resource) with the realm descriptor (which should be stable and descriptive), and it produces a different `realm` value for every distinct URL, which is inconsistent with how realm is supposed to work.

The `bearer-auth` middleware avoids this by exposing a `realm` option (defaulting to `""`) that lets developers supply a fixed, meaningful string:

```ts
// bearer-auth/index.ts — line 25-26
realm?: string
// line 117
```

`jwt` and `jwk` have no equivalent option; their `unauthorizedResponse()` helper is private and always reads `opts.ctx.req.url`. There is no workaround short of wrapping or replacing the middleware entirely. This API inconsistency means developers using `jwt`/`jwk` cannot produce a well-formed, stable `WWW-Authenticate` realm value without forking the middleware.

## Affected code

`src/middleware/jwt/jwt.ts` lines 160–173 (identical pattern at `src/middleware/jwk/jwk.ts` lines 170–183):

```ts
160 | function unauthorizedResponse(opts: {
161 |   ctx: Context
162 |   error: string
163 |   errDescription: string
164 |   statusText?: string
165 | }) {
166 |   return new Response('Unauthorized', {
167 |     status: 401,
168 |     statusText: opts.statusText,
169 |     headers: {
170 |       'WWW-Authenticate': `Bearer realm="${opts.ctx.req.url}",error="${opts.error}",error_description="${opts.errDescription}"`,
171 |     },
172 |   })
173 | }
```

## Fix recommendation

Add an optional `realm` field to both the `jwt()` and `jwk()` options objects, mirroring what `bearer-auth` already does. When `realm` is not provided, fall back to `""` rather than `ctx.req.url`:

```ts
// jwt options type — add:
realm?: string

// In unauthorizedResponse, replace:
`Bearer realm="${opts.ctx.req.url}"`
// with:
`Bearer realm="${opts.realm ?? ''}"`
```

This change is backwards-compatible. The same change should be applied to `src/middleware/jwk/jwk.ts`, which contains an identical copy of `unauthorizedResponse`.

## PoC

See poc_idx0.mjs / poc_output_idx0.txt

mixelburg:
Fixes #4989

The `jwt` and `jwk` middlewares previously hardcoded the request URL as the `realm` value in the `WWW-Authenticate` header. Per RFC 6750, `realm` should be a stable, human-readable string identifying the protection space — not the varying request URL.

This adds an optional `realm` option to both middlewares, matching the pattern already used by `bearer-auth`. When not set, the previous behavior (request URL) is preserved.

### DIFF
--- a/src/middleware/jwk/jwk.ts
+++ b/src/middleware/jwk/jwk.ts
@@ -25,6 +25,7 @@ import type { VerifyOptions } from '../../utils/jwt/jwt'
  * @param {boolean} [options.allow_anon] - If set to `true`, the middleware allows requests without a token to proceed without authentication.
  * @param {string} [options.cookie] - If set, the middleware attempts to retrieve the token from a cookie with these options (optionally signed) only if no token is found in the header.
  * @param {string} [options.headerName='Authorization'] - The name of the header to look for the JWT token. Default is 'Authorization'.
+ * @param {string} [options.realm] - The realm for the WWW-Authenticate header. If not set, defaults to the request URL.
  * @param {AsymmetricAlgorithm[]} options.alg - An array of allowed asymmetric algorithms for JWT verification. Only tokens signed with these algorithms will be accepted.
  * @param {RequestInit} [init] - Optional init options for the `fetch` request when retrieving JWKS from a URI.
  * @param {VerifyOptions} [options.verification] - Additional options for JWK payload verification.
@@ -55,6 +56,8 @@ export const jwk = (
       | { key: string; secret?: string | BufferSource; prefixOptions?: CookiePrefixOptions }
 
     headerName?: string
+
+    realm?: string
 
     alg: AsymmetricAlgorithm[]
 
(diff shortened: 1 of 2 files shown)


mixelburg:
Closing due to inactivity — over 40 days with no engagement. Happy to reopen if the team is interested.

PR #5014, opened 2026-06-09
fix: reject non-serializable json values

pupuking723:
Fixes #2343.

This makes `c.json()` throw a `TypeError` when the top-level value cannot be serialized by `JSON.stringify()`. Today those values produce `undefined` from `JSON.stringify()`, which gets passed into the response body and results in an empty response instead of a clear failure.

The check follows the Deno-style behavior discussed in the issue: serialize first, then reject values where JSON serialization returns `undefined`. This covers `undefined`, symbols, and functions while keeping existing serializable values on the same response path.

Verification:

- `/Users/mac/.bun/bin/bunx vitest --run src/context.test.ts`
- `/Users/mac/.bun/bin/bunx tsc --noEmit`
- `/Users/mac/.bun/bin/bun run lint`
- `/Users/mac/.bun/bin/bun run format`

### DIFF
--- a/src/context.test.ts
+++ b/src/context.test.ts
@@ -70,6 +70,12 @@ describe('Context', () => {
     expect(res.headers.get('X-Custom')).toBe('Message')
   })
 
+  it('c.json() throws for non-serializable values', () => {
+    expect(() => c.json(undefined)).toThrow(TypeError)
+    expect(() => c.json(Symbol('not-json'))).toThrow(TypeError)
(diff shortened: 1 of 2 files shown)


pupuking723:
Thanks for taking a look. Could you clarify whether this was closed because the behavior change is not desired, or because the implementation should take a different direction?

I do not want to reopen or resubmit without understanding the expected behavior. If there is a narrower fix that would be acceptable, I can adjust accordingly.

PR #5016, opened 2026-06-09
fix(aws-lambda): detect V2 events by request context, not rawPath alone

### LINKED ISSUE #5015: aws-lambda handler misinterprets V1 events as V2 events when using base path mappings

4.12.8

AWS API Gateway V1

1. deploy hono app as AWS lambda function
2. integrate the lambda as REST API in the API Gateway
3. add a base path mapping for a custom domain in the API Gateway
4. invoke the API via the custom domain

hono interprets the lambda event as an `APIGatewayProxyEvent`, using the `path` property as path for the route, thereby correctly stripping the base path from the base path mapping.

### What do you see instead?

hono sees a `rawPath` property, even though it's deployed as a REST API, thereby interpreting the lambda event as `APIGatewayProxyEventV2` and mistakenly using the `rawPath` as path for the route.

### Additional information

Though barely documented, when adding a base path mapping in AWS API Gateway, AWS adds a `rawPath` property to the APIGatewayProxyEvent for REST APIs, containing the original path before the base path mapping was applied.

This confuses hono into thinking it's an APIGatewayProxyEventV2, even though the app was integrated into a REST API (API Gateway V1) and not into a HTTP API (API Gateway V2), thereby using the wrong path for the routes (and potentially also misinterpreting other properties).

VihAMBR:
Fixes #5015

### What this fixes

When a Lambda is served through an API Gateway REST API (V1) with a custom domain base path mapping, API Gateway adds a `rawPath` to the event. `isProxyEventV2` only checked whether `rawPath` was present, so these V1 events were treated as V2 (HTTP API). The handler then routed on `rawPath` and read `requestContext.http`, which a V1 event does not have, so the request never reached its route.

Every genuine V2 event (HTTP API and Lambda function URLs) carries an `http` object on its request context; a V1 REST event does not. Checking for that alongside `rawPath` keeps real V2 events working and stops a V1 event with a base-path `rawPath` from being misread as V2.

### Test

Added a `handle` test that sends a V1 event carrying a base-path `rawPath` and asserts it routes by its `path`. It fails before the change (the event is handled as V2) and passes after.

---

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/adapter/aws-lambda/handler.test.ts
+++ b/src/adapter/aws-lambda/handler.test.ts
@@ -372,6 +372,24 @@ describe('EventProcessor.createRequest', () => {
 })
 
 describe('handle', () => {
+  it('Should route a V1 REST event by its path even when a base path mapping adds rawPath', async () => {
+    const app = new Hono()
+    app.get('/my/path', (c) => c.text('Hello'))
+    const handler = handle(app)
+
+    // A custom domain base path mapping makes API Gateway add a `rawPath` to the
+    // V1 (REST API) event, holding the path before the mapping was stripped. The
+    // event is still V1: it carries `path` and a V1 request context with no `http`.
+    const event: LambdaEvent = {
+      ...baseV1Event,
+      rawPath: '/base/my/path',
+    }
+
+    const result = await handler(event)
+    expect(result.statusCode).toBe(200)
+    expect(result.body).toBe('Hello')
+  })
+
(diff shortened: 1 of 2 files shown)


VihAMBR:
Reopening from my main account (VihaanAgarwal).

VihAMBR:
Reopened as #5033.

Issue #5010, opened 2026-06-08
Reliability issues in method-override, compress, and aws-lambda adapter (4.12.24)


m-be290108:
Hi @yusukebe — thanks for Hono. The zero-dependency, Web-Standard design is genuinely a pleasure to read through, and the way the security middleware is already hardened (cookie sanitize, CORS `Vary`, JWT alg confusion, timing-safe compare) made the surface for "easy wins" quite narrow.

Out of that constraint I focused on reliability and multi-runtime divergence on **v4.12.24** (Node 25 / `@hono/node-server`). Three observations:

**1. `methodOverride({ query })` returns 500 on any request with a body**

`src/middleware/method-override/index.ts:127` rebuilds the request from the raw `ReadableStream` without `duplex: 'half'`. Per Fetch spec, undici throws `TypeError: RequestInit: duplex option is required when sending a body`, surfaced as 500. The `form` (`:83`, `:98`) and `header` (`:113`) branches build non-stream bodies, so only `query` is affected. Since an HTML form always sends a body, every form `POST` to a query-override route returns 500. Current `Query` test (`index.test.ts:177`) only sends a bodyless `POST`, which passes. Same `duplex` family as #2627, different middleware.

**2. `compress()` encodes `206 Partial Content` and strips `Content-Length`**

`src/middleware/compress/index.ts:92` skips on `Content-Encoding` / `Transfer-Encoding` / HEAD / threshold / type, but not on status or `Content-Range`. A 206 ends up gzipped while `Content-Range` describes the *uncompressed* byte range — range clients reassemble corrupted data. nginx/apache skip compression on range responses for this reason.

**3. `aws-lambda` adapter ships without the `globalThis.crypto` polyfill that `lambda-edge` includes**

`src/adapter/lambda-edge/handler.ts` polyfills `globalThis.crypto` for older Node Lambda runtimes; `src/adapter/aws-lambda/handler.ts` does not. Functions that rely on `crypto.subtle` (JWT verify, hashing) throw at first request on Node 18 Lambda. Asymmetric across two adapters that share the runtime concern.

---

Both #1 and #2 look like one-line guards (`duplex: 'half'` ; skip when `status === 206`). #3 mirrors what `lambda-edge` already does.

A few of the patterns I surfaced felt a little too sensitive to lay out in public — happy to walk you through the rest privately, links are on my profile.

— Mathieu Bellot

sijie-Z:
[...]

PR #5020, opened 2026-06-11
fix(compress): do not compress 206 Partial Content responses

dr4gan0x:
This addresses the second of the three problems reported in #5010 (compress encodes 206 Partial Content).

The middleware bails out for already-encoded, chunked, HEAD, small, non-compressible and no-transform responses, but it never checks the status code. A 206 ends up gzipped while its Content-Range header still refers to the uncompressed byte range, and the Content-Length that matched the partial body gets dropped. A range client stitching parts together from such responses gets corrupted data. This is easy to hit with a handler that does its own range handling, or when proxying an upstream that serves ranges, since the proxy helper passes the upstream status through.

The fix skips compression when the status is 206.

The new test serves a 1024 byte text/plain part with Content-Range and asserts it comes back uncompressed with status, Content-Length and Content-Range intact. It fails without the guard.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/compress/index.test.ts
+++ b/src/middleware/compress/index.test.ts
@@ -58,6 +58,12 @@ describe('Compress Middleware', () => {
     c.header('Content-Length', '1024')
     return c.body(new Uint8Array(1024)) // Simulated chunked data
   })
+  app.get('/partial', (c) => {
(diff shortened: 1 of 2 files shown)


yusukebe (approved):
LGTM!

PR #5021, opened 2026-06-12
fix(concurrent): propagate rejections and release the pool slot on failure

sarathfrancis90:
### What

While reading `createPool` in `src/utils/concurrent.ts` I noticed `run()` only handles the success path:

```ts
const result = await fn()
// delete marker / resolve happen only here
```

When `fn()` rejects, two things go wrong:

1. **The pool slot leaks.** The marker is never removed, so once `concurrency` tasks have thrown, `pool.size >= concurrency` stays true forever and every later task just re-queues via `setTimeout` endlessly — the pool stops doing any work.
2. **A queued task's rejection is lost.** If a task rejects *while it was queued* (the pool was full when `run` was first called), the promise the caller is awaiting is never settled, and the rejection surfaces as an unhandled rejection. The caller hangs indefinitely instead of seeing the error.

This bites `hono/ssg`, which fetches every route through the pool — a couple of failing routes can wedge the whole generation run.

### How

- Moved the slot release into a `finally` so it runs on both the resolve and reject paths (the `interval`-delayed release is preserved unchanged).
- Threaded a `reject` callback through the queued re-dispatch so a queued task's rejection is forwarded to the original caller, and swallowed the re-dispatch's own (now unused) promise so it can't become an unhandled rejection.

### Tests

Added four cases covering: an immediately-running task that rejects, slot release after a rejection, a queued task whose rejection must reach the caller, and the pool continuing to process later tasks after a queued task rejects. Each fails on `main` (the queued ones hang / emit an unhandled rejection) and passes with the fix. Full suite, `tsc --noEmit`, and lint are green.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/utils/concurrent.test.ts
+++ b/src/utils/concurrent.test.ts
@@ -37,6 +37,82 @@ describe('concurrent execution', () => {
     expect(results).toEqual(expectedResults)
   })
 
+  it('Should reject when an immediately-running task rejects', async () => {
+    const pool = createPool({ concurrency: 2 })
+    await expect(pool.run(() => Promise.reject(new Error('boom')))).rejects.toThrow('boom')
+  })
+
+  it('Should release the slot when a task rejects', async () => {
+    const pool = createPool({ concurrency: 1 })
+
+    // The first task fails. If the slot is not released, the pool stays full forever.
+    await expect(pool.run(() => Promise.reject(new Error('boom')))).rejects.toThrow('boom')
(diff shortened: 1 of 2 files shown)


usualoma:
Closing this in accordance with our AI Usage Policy:
https://github.com/honojs/hono/blob/main/docs/CONTRIBUTING.md#ai-usage-policy

PR #5047, opened 2026-06-26
docs(MIGRATION): fix req.raw.headers reference (property, not method)

EduardF1:
In the v3.12 → v4.0 migration notes:

> - HonoRequest - `headers()`, `body()`, ... are obsolete. Use the methods in `req.raw` such as `req.raw.headers()`.

`req.raw` is a Fetch API [`Request`](https://developer.mozilla.org/en-US/docs/Web/API/Request), and its `headers`, `body`, `bodyUsed`, `integrity`, `keepalive`, `referrer` and `signal` are **properties (getters), not methods**. Copy-pasting `req.raw.headers()` produces `TypeError: req.raw.headers is not a function`.

Hono's own `src/request.ts` accesses it as a property (e.g. `this.raw.headers.get(name)`).

This PR fixes the example to `req.raw.headers` and refers to them as *properties* instead of *methods*.

Docs-only change.


yusukebe (approved):

yusukebe:
@EduardF1 Thanks!

PR #5049, opened 2026-06-27
feat: add first-class support for the HTTP QUERY method

BonchitoSky:
This PR adds QUERY as a first-class HTTP method in Hono, exposing app.query() alongside the existing built-in helpers (get, post, put, delete, options, patch).

Motivation

The HTTP QUERY method is an emerging standard (draft-ietf-httpbis-safe-method-w-body (https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-05.txt)) designed for safe, idempotent requests that carry a request body — filling the gap that GET cannot fill. As the HTTP ecosystem evolves toward adopting QUERY, having first-class support in Hono improves:

- Discoverability — app.query() is visible in autocomplete and type definitions, rather than hidden behind app.on()
- Consistency — matches the ergonomics of every other built-in method helper
- Type safety — full schema inference, path param typing, and chaining work out of the box
- Readability — intent is clear at a glance in route definitions

Changes

src/router.ts

Added 'query' to the METHODS array, which is the single source of truth for built-in method helpers.

export const METHODS = ['get', 'post', 'put', 'delete', 'options', 'patch', 'query'] as const

src/hono-base.ts

Added the query property declaration on the Hono class. The constructor already dynamically loops over METHODS to generate handler functions, so no additional runtime code was needed.

query!: HandlerInterface<E, 'query', S, BasePath, CurrentPath>

src/hono.test.ts

Added 4 tests covering:
- Basic routing via app.query()
- Routing via app.on('QUERY', ...) for backward compatibility
- 404 response for QUERY on an unregistered path
- Path parameter support with app.query()

Usage

const app = new Hono()

// First-class method helper
app.query('/search', (c) => {
  const body = await c.req.json()
  return c.json({ results: [] })
})

// Still works via app.on() for backward compatibility
app.on('QUERY', '/search', (c) => c.text('ok'))

Checklist

- [x] Tests added and passing (211/211)
- [x] TypeScript type check passes (tsc --noEmit)
- [ ] bun run format:fix && bun run lint:fix
- [ ] TSDoc/JSDoc added to the query property

### DIFF
--- a/src/hono-base.ts
+++ b/src/hono-base.ts
@@ -107,6 +107,7 @@ class Hono<
   delete!: HandlerInterface<E, 'delete', S, BasePath, CurrentPath>
   options!: HandlerInterface<E, 'options', S, BasePath, CurrentPath>
   patch!: HandlerInterface<E, 'patch', S, BasePath, CurrentPath>
+  query!: HandlerInterface<E, 'query', S, BasePath, CurrentPath>
   all!: HandlerInterface<E, 'all', S, BasePath, CurrentPath>
   on: OnHandlerInterface<E, S, BasePath>
   use: MiddlewareHandlerInterface<E, S, BasePath>
--- a/src/hono.test.ts
+++ b/src/hono.test.ts
@@ -2079,6 +2079,47 @@ describe('Multiple methods with `app.on`', () => {
   })
 })
 
+describe('HTTP QUERY method', () => {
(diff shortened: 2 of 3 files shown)


yusukebe:
We continue with the PR https://github.com/honojs/hono/pull/5070 for QUERY support.

PR #5058, opened 2026-06-28
feat(middleware): method not allowed

### LINKED ISSUE #4633: Return `405 Method Not Allowed` when a path exists but the HTTP method is unsupported

otnc:
### Why?

When a request hits a path that exists but with an unsupported HTTP method, Hono returns `404 Not Found`. Per HTTP semantics this should be `405 Method Not Allowed` with an `Allow` header, so clients can tell "route missing" apart from "method not supported".

In the discussion on #4633, the maintainers agreed this should ship as a standalone middleware that receives `app`, rather than a change to the Hono core. The existing PR #4637 takes the core approach (`app.methodNotAllowed()`); this PR implements the agreed middleware approach instead, picking up from where the issue discussion left off.

### Related Issue(s)

closes #4633

### Description

Adds a `methodNotAllowed({ app })` middleware (`hono/method-not-allowed`).

```ts
import { Hono } from 'hono'
import { methodNotAllowed } from 'hono/method-not-allowed'

const app = new Hono()

app.get('/', (c) => c.text('Hello World!'))
// `POST /` -> 405 Method Not Allowed, `Allow: GET`
// `GET /nonexistent` -> 404 Not Found (path does not exist)
```

- Implemented purely as middleware; no change to the Hono core.
- On the first 404, it builds a `path -> methods` lookup (a `TrieRouter`) from
  `app.routes` and caches it for subsequent requests.
- The requested method is excluded from `Allow`, and methods are de-duplicated
  so overlapping routes (e.g. `*` and an exact path) don't produce duplicates.
- Routes registered with `app.use()` / `app.all()` are ignored.
- Must be registered with `app.use()` before the handlers.

### Comment

Claude Code and Google Translate were used for part of the development and for drafting this PR description. All of the code, the tests, and this PR description were reviewed by me before submission.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/jsr.json
+++ b/jsr.json
@@ -46,6 +46,7 @@
     "./timing": "./src/middleware/timing/timing.ts",
     "./logger": "./src/middleware/logger/index.ts",
     "./method-override": "./src/middleware/method-override/index.ts",
+    "./method-not-allowed": "./src/middleware/method-not-allowed/index.ts",
     "./powered-by": "./src/middleware/powered-by/index.ts",
     "./pretty-json": "./src/middleware/pretty-json/index.ts",
     "./request-id": "./src/middleware/request-id/request-id.ts",
--- a/package.json
+++ b/package.json
@@ -235,6 +235,11 @@
       "import": "./dist/middleware/method-override/index.js",
       "require": "./dist/cjs/middleware/method-override/index.js"
     },
+    "./method-not-allowed": {
+      "types": "./dist/types/middleware/method-not-allowed/index.d.ts",
+      "import": "./dist/middleware/method-not-allowed/index.js",
+      "require": "./dist/cjs/middleware/method-not-allowed/index.js"
+    },
     "./powered-by": {
       "types": "./dist/types/middleware/powered-by/index.d.ts",
       "import": "./dist/middleware/powered-by/index.js",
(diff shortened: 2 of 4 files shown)


usualoma:
Hi @otnc,

I can tell you put a lot of effort into putting this together. However, I’m afraid I cannot accept a pull request that resolves an issue you haven’t participated in discussing and follows the same approach that’s already been proposed.

If you’d said something like, “I came up with a great solution using a different approach,” I’d be happy to review the details. Is there anything in your PR that wasn’t mentioned in the issue—something that could be considered the original part of this PR?

As a maintainer, I do feel bad that the issue has stalled. I’m sorry about that.

otnc:
Hi @usualoma, thanks for the honest and kind reply — that's completely fair.

You're right: I implemented essentially the approach you already outlined in
the issue, without having taken part in the discussion. The only deltas from
your snippet are minor (de-duplicating methods / excluding the requested method
in the `Allow` header), which isn't a meaningfully different approach. I
shouldn't have jumped straight to an implementation PR here.

Please feel free to close this. I'll engage in the issue discussion first next
time, or bring something genuinely new to the table. Thanks again for taking
the time to look.

PR #4955, opened 2026-05-21
fix(deno): echo negotiated WebSocket subprotocol in upgrade response

ATOM00blue:
## What

The Deno adapter's `upgradeWebSocket` does not set the `Sec-WebSocket-Protocol` header on the upgrade response. When a client opens a socket with a subprotocol (e.g. `new WebSocket(url, ["foobar"])`), Chrome rejects the connection because the negotiated subprotocol is missing from the 101 response. Firefox is more lenient, which is why this looked Chrome-specific.

## How

`Deno.upgradeWebSocket` only emits the `Sec-WebSocket-Protocol` response header when given a `protocol` option. This reads the first value from the request's `Sec-WebSocket-Protocol` header and passes it through as `protocol`. An explicitly provided `protocol` option still takes precedence, so existing behavior is preserved.

## Tests

Added unit tests that mock the `Deno.upgradeWebSocket` global and assert the `protocol` option is derived from the request header (and left unset when no subprotocol is requested). `tsc --noEmit`, eslint, and prettier all pass.

Closes #4439

PR #5059, opened 2026-06-28
fix(bun): report the requested subprotocol on WSContext.protocol

mahirhir:
The Bun WebSocket adapter sets `WSContext.protocol` to the request URL:

```ts
protocol: c.req.url,
```

`WSContext.protocol` is supposed to mirror the standard `WebSocket.protocol` property (the connection's subprotocol), so on Bun it currently returns something like `http://localhost/ws` instead of the subprotocol. The deno and cloudflare adapters read it from the platform socket (`socket.protocol` and `server.protocol`), so this is a Bun-only inconsistency.

This reads the first value of the `Sec-WebSocket-Protocol` header, the same way the deno adapter does (#4955), and stores it on the data that backs `WSContext.protocol`.

Added a test that upgrades with `Sec-WebSocket-Protocol: chat, superchat` and asserts `wsContext.protocol` is `chat`. Without this change it fails with `expected 'http://localhost/ws' to be 'chat'`.

- [x] Add tests
- [x] Run tests
- [x] Format with Prettier

### DIFF
--- a/src/adapter/bun/websocket.ts
+++ b/src/adapter/bun/websocket.ts
@@ -62,7 +62,9 @@ export const upgradeWebSocket: UpgradeWebSocket<any> = defineWebSocketHelper((c,
     data: {
       events,
       url: new URL(c.req.url),
-      protocol: c.req.url,
+      // The first requested subprotocol, exposed via WSContext.protocol to
+      // match the deno and cloudflare adapters.
+      protocol: c.req.header('sec-websocket-protocol')?.split(',')[0]?.trim() ?? '',
     },
   })
   if (upgradeResult) {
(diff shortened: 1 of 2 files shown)


src/adapter/bun/websocket.ts
yusukebe, line 1: Can you add a test to test that the value of `ws.protocol` will be `''` if you do not set the `sec-websocket-protocol` header?

mahirhir:
Added the no-protocol case, asserts ws.protocol is '' when the header is absent. Thanks.

yusukebe (approved):
LGTM!

yusukebe:
@greymoth-jp Thank you!

PR #5063, opened 2026-06-30
feat(jsx): add React-compatible overloads to useRef

### LINKED ISSUE #5056: Proposal: add overloads to `useRef` type to match React's API
## Context

When using `hono/jsx` hooks to write code that mirrors React-style patterns, `useRef` returns a ref whose `.current` is always typed as `T | null`, even when initialized with a non-null value.

```ts
const ref = useRef(new Map<string, number>())
ref.current.set('a', 1) // ❌ TS error: 'ref.current' is possibly 'null'.
```

This is because `useRef` currently has a single signature:

```ts
// hono/src/jsx/hooks/index.ts
export declare const useRef: <T>(initialValue: T | null) => RefObject<T>
// where RefObject<T> = { current: T | null }
```

The runtime implementation always returns `{ current: initialValue }`, so when a non-null initial value is passed, `.current` cannot actually be `null` at first read. The nullable type is purely a TypeScript-level limitation.

## React behavior (for comparison)

React provides overloads so that `useRef(value)` returns a non-nullable ref, while `useRef<T>(null)` keeps the nullable shape needed for DOM refs:

```ts
function useRef<T>(initialValue: T): MutableRefObject<T>
function useRef<T>(initialValue: T | null): RefObject<T>
function useRef<T = undefined>(): MutableRefObject<T | undefined>
```

## Proposal

Add matching overloads to `hono/jsx`'s `useRef` so that users get non-nullable `.current` when they pass a non-null initial value. Implementation does not need to change — this is a type-only fix.

I'm happy to submit a PR if this direction is welcome. Wanted to check first whether the single-overload signature was intentional (e.g. simpler types by design) before opening one.

ashunar0:
## Summary

Adds overloads to `useRef` in `hono/jsx` so that passing a non-null initial value returns a `MutableRefObject<T>` whose `.current` is non-nullable, matching React's API.

```ts
function useRef<T>(initialValue: T): MutableRefObject<T>
function useRef<T>(initialValue: T | null): RefObject<T>
function useRef<T = undefined>(): MutableRefObject<T | undefined>
```

Before this change, `useRef(new Map())` typed `.current` as possibly `null`, even though the runtime always initializes it. This is a type-only fix; the runtime implementation is unchanged.

Closes #5056

## Test plan

- [x] `tsc --noEmit` passes
- [x] Added type-level tests in `src/jsx/dom/index.test.tsx` covering: non-null initial value, null initial value, and no-argument forms

### DIFF
--- a/src/jsx/hooks/index.ts
+++ b/src/jsx/hooks/index.ts
@@ -316,10 +316,15 @@ export const useCallback = <T extends Function>(callback: T, deps: readonly unkn
 }
 
 export type RefObject<T> = { current: T | null }
-export const useRef = <T>(initialValue: T | null): RefObject<T> => {
+export type MutableRefObject<T> = { current: T }
+
+export function useRef<T>(initialValue: T): MutableRefObject<T>
+export function useRef<T>(initialValue: T | null): RefObject<T>
+export function useRef<T = undefined>(): MutableRefObject<T | undefined>
+export function useRef<T>(initialValue?: T | null): RefObject<T> | MutableRefObject<T | undefined> {
   const buildData = buildDataStack.at(-1) as [unknown, NodeObject]
   if (!buildData) {
-    return { current: initialValue }
+    return { current: initialValue } as RefObject<T>
   }
   const [, node] = buildData
 
--- a/src/jsx/hooks/index.ts
+++ b/src/jsx/hooks/index.ts
@@ -316,8 +316,6 @@ export const useCallback = <T extends Function>(callback: T, deps: readonly unkn
 }
 
 export type RefObject<T> = { current: T }
-/** @deprecated Use `RefObject<T | null>` instead. */
-export type MutableRefObject<T> = { current: T }
 
 export function useRef<T>(initialValue: T): RefObject<T>
 export function useRef<T>(initialValue: T | null): RefObject<T | null>
--- a/src/jsx/hooks/index.ts
+++ b/src/jsx/hooks/index.ts
@@ -322,9 +322,7 @@ export type MutableRefObject<T> = { current: T }
 export function useRef<T>(initialValue: T): RefObject<T>
 export function useRef<T>(initialValue: T | null): RefObject<T | null>
 export function useRef<T>(initialValue: T | undefined): RefObject<T | undefined>
-export function useRef<T>(
-  initialValue?: T | null
-): RefObject<T | null | undefined> {
+export function useRef<T>(initialValue?: T | null): RefObject<T | null | undefined> {
   const buildData = buildDataStack.at(-1) as [unknown, NodeObject]
   if (!buildData) {
     return { current: initialValue } as RefObject<T | null | undefined>
--- a/src/jsx/hooks/index.ts
+++ b/src/jsx/hooks/index.ts
@@ -315,16 +315,19 @@ export const useCallback = <T extends Function>(callback: T, deps: readonly unkn
   return callback
 }
 
-export type RefObject<T> = { current: T | null }
+export type RefObject<T> = { current: T }
+/** @deprecated Use `RefObject<T | null>` instead. */
 export type MutableRefObject<T> = { current: T }
 
-export function useRef<T>(initialValue: T): MutableRefObject<T>
-export function useRef<T>(initialValue: T | null): RefObject<T>
-export function useRef<T = undefined>(): MutableRefObject<T | undefined>
-export function useRef<T>(initialValue?: T | null): RefObject<T> | MutableRefObject<T | undefined> {
+export function useRef<T>(initialValue: T): RefObject<T>
+export function useRef<T>(initialValue: T | null): RefObject<T | null>
+export function useRef<T>(initialValue: T | undefined): RefObject<T | undefined>
+export function useRef<T>(
+  initialValue?: T | null
+): RefObject<T | null | undefined> {
   const buildData = buildDataStack.at(-1) as [unknown, NodeObject]
   if (!buildData) {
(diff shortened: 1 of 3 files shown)


ashunar0:
A note on which React version this matches:

These overloads follow **React 18** (`MutableRefObject<T>` for non-null initial values, `RefObject<T>` with nullable `.current` for the null case).

React 19 reshaped this API — `RefObject<T>` became mutable with non-null `current`, `MutableRefObject` was removed, and nullability is now expressed via the type parameter (e.g. `RefObject<T | null>`).

I went with the React 18 shape because hono's existing `RefObject<T> = { current: T | null }` already matches React 18, so this stays consistent with the current type and is non-breaking for users. Aligning with React 19 would be a separate, larger discussion since it would change the meaning of `RefObject` itself. Happy to revisit if you'd prefer the 19 shape instead.

usualoma:
Hi @ashunar0,

Thanks!

Hono generally tries to stay closer to React 19 semantics where possible, but given the current `RefObject<T>` shape in Hono, this React 18-style overload is a reasonable incremental improvement and keeps the change non-breaking.

If we decide to align `RefObject` itself with React 19 later, we can revisit this as part of that broader change.

yusukebe:
I think we can introduce `RefObject<T>`. It indeed makes a breaking change, but it's only in cases where users use `RefObject<T>` directly and add a type to it as a generic type parameter. The case is rare, and following React 19 is the true way for us.

@usualoma What do you think of it?

usualoma:
Hi @yusukebe and @ashunar0 

Sorry, I commented without reading it carefully, but I get it now!
There are a few breaking changes, but it doesn't seem like many users will be affected, so it looks like Hono should just align with React 19.

ashunar0:
Hi @usualoma @yusukebe,

Thanks for the discussion! I've updated the PR to align with React 19:

- `RefObject<T>` is now `{ current: T }` (non-null, mutable)
- `MutableRefObject<T>` kept as a `@deprecated` alias for backward compat
- `useRef` overloads match React 19 (the zero-arg form is gone — `useRef(undefined)` now)
- `createRef`, `forwardRef`, `useImperativeHandle` updated to use `RefObject<T | null>` where nullability applies
- Migrated the existing tests / internal usages to the new shape

`tsc --noEmit` passes locally. Let me know what you think!

src/jsx/hooks/index.ts
```diff
@@ -315,11 +315,17 @@ export const useCallback = <T extends Function>(callback: T, deps: readonly unkn
   return callback
 }
 
-export type RefObject<T> = { current: T | null }
-export const useRef = <T>(initialValue: T | null): RefObject<T> => {
+export type RefObject<T> = { current: T }
+/** @deprecated Use `RefObject<T | null>` instead. */
+export type MutableRefObject<T> = { current: T }
```
yusukebe: Is `MutableRefObject` needed??

src/jsx/hooks/index.ts
```diff
@@ -315,11 +315,17 @@ export const useCallback = <T extends Function>(callback: T, deps: readonly unkn
...
+
+export function useRef<T>(initialValue: T): RefObject<T>
```
yusukebe, line 320: With this code, if the argument is blank, like the following, it will throw a type error:

```ts
const ref = useRef<HTMLDivElement>()
```

So you have to change the overload to this, don't you?:

```ts
export function useRef<T>(initialValue: T): RefObject<T>
export function useRef<T>(initialValue: T | null): RefObject<T | null>
export function useRef<T = undefined>(): RefObject<T | undefined>
```

ashunar0:
Thanks for the review, @yusukebe!

I double-checked both points against `@types/react@19.2.17` and it turns out the current PR already matches React 19 exactly:

**`MutableRefObject`**
```ts
// @types/react@19.2.17 (react/index.d.ts)
/** @deprecated Use `RefObject` instead. */
interface MutableRefObject<T> {
    current: T;
}
```
React 19 keeps `MutableRefObject` as a `@deprecated` alias for migration compatibility, so I left it in the same form here.

**`useRef` overloads**
```ts
// @types/react@19.2.17
function useRef<T>(initialValue: T): RefObject<T>;
function useRef<T>(initialValue: T | null): RefObject<T | null>;
function useRef<T>(initialValue: T | undefined): RefObject<T | undefined>;
```
The PR already has these three overloads (lines 322-324). The zero-arg call `useRef<HTMLDivElement>()` is intentionally rejected in React 19 — you have to pass `undefined` explicitly (`useRef<HTMLDivElement>(undefined)`). Your suggested `useRef<T = undefined>(): RefObject<T | undefined>` form would allow the zero-arg call, which diverges from React 19.

That said, if Hono prefers a different stance here — e.g. drop `MutableRefObject` entirely, or allow the zero-arg call as an ergonomics improvement over React 19 — happy to adjust. Let me know which direction you'd like!

yusukebe:
@ashunar0 

Thank you for checking! You are right; the `useRef` overloads in this PR are correct. Regarding `MutableRefObject`, I don't think you need to add it because it's already `deprecated`, and hono/jsx never proved it before.

ashunar0:
Thanks @yusukebe! Agreed — I've removed the `MutableRefObject` alias in the latest commit. Since hono/jsx never exposed it before, there's no migration cost.

yusukebe (approved):
LGTM!

yusukebe:
Hi @ashunar0 

Looks good! I'll merge this later. I think it's okay to include this change in the next patch version. Anyway, thanks!

yusukebe:
I'll include this in v4.13.

PR #5064, opened 2026-06-30
fix(utils,validator): match Content-Type media types case-insensitively

### LINKED ISSUE #5060: parseBody and validator skip mixed-case Content-Type media types
## Summary

`parseBody()` and `validator()` match supported `Content-Type` media types with case-sensitive checks. Requests using valid mixed/upper-case media type values such as `Application/JSON` or `Application/X-WWW-Form-Urlencoded` can be parsed by the platform `Request`, but Hono's guards skip them and may pass an empty object downstream.

## Code path

Parser / validator / feedback anchors:

- `src/utils/body.ts:100-101`: `parseBody()` reads the raw `Content-Type` header value.
- `src/utils/body.ts:103-105`: form parsing is guarded by lower-case `startsWith()` literals.
- `src/utils/body.ts:110`: non-matching content types return `{}`.
- `src/validator/validator.ts:24-26`: JSON/form regexes use lower-case media type literals.
- `src/validator/validator.ts:95` and `src/validator/validator.ts:107`: non-matching content types skip the parsing/validation branch.

## Steps to reproduce

Validation level: dynamic platform reproduction plus source-control-flow validation.

```js
const req = new Request('http://localhost/post', {
  method: 'POST',
  headers: { 'Content-Type': 'Application/JSON' },
  body: '{"message":"hello"}',
})

console.log(await req.clone().json()) // { message: 'hello' }
```

The platform parser accepts the mixed/upper-case media type. However, Hono's current guards evaluate the same header value with lower-case-only string/regex checks. The local branch probe showed:

```text
Application/JSON -> platform JSON parsed, Hono json validator guard false
Application/X-WWW-Form-Urlencoded -> platform form parsed, Hono parseBody guard false
```

## Expected behavior

Supported media type matching should be case-insensitive for the media type portion. Parameter values such as multipart boundaries should remain intact.

## Actual behavior

Case-only variants of supported content types are silently skipped by Hono parser/validator logic, so handlers can receive `{}` or unvalidated empty data instead of the request body.

## Existing coverage

I searched current issues and PRs for `Content-Type parseBody validator case`, `Application JSON validator`, `parseBody`, and broader `Content-Type` terms. Related items such as #4062, #3707, and #4109 discuss validator/content-type behavior, but they do not appear to cover media type value case normalization.

## Suggested fix

Normalize or parse the media type portion before matching, for example by lowercasing only the part before `;`. Avoid lowercasing parameter values that may be case-sensitive.

## Suggested tests

- `parseBody()` handles `Application/X-WWW-Form-Urlencoded`.
- `validator('json')` handles `Application/JSON`.
- `validator('form')` handles mixed-case urlencoded/multipart media types while preserving multipart boundary behavior.

---
Submitted with Codex.

Ram-blip:
### What changed

* Match supported `Content-Type` media types case-insensitively in `validator()`.
* Normalize the media type portion of `Content-Type` in `parseBody()` before checking form content types.
* Add regression tests for mixed-case JSON and URL-encoded form media types.

### Why

`Content-Type` media types are case-insensitive, but mixed-case values such as `Application/JSON` and `Application/X-WWW-Form-Urlencoded` were skipped by `validator()` and `parseBody()`.

Fixes #5060.

### Tests

- `bunx vitest --run src/utils/body.test.ts src/validator/validator.test.ts`
- `bun run test`
- `bun run lint`

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/utils/body.test.ts
+++ b/src/utils/body.test.ts
@@ -42,6 +42,17 @@ describe('Parse Body Util', () => {
     expect(await parseBody(req)).toEqual({ message: 'hello' })
   })
 
+  it('should parse mixed-case `x-www-form-urlencoded` content type', async () => {
+    const searchParams = new URLSearchParams()
+    searchParams.append('message', 'hello')
+
+    const req = createRequest(SEARCH_URL, 'POST', searchParams, {
+      'Content-Type': 'Application/X-WWW-Form-Urlencoded',
+    })
+
+    expect(await parseBody(req)).toEqual({ message: 'hello' })
+  })
+
   it('should not parse multiple values in default', async () => {
     const data = new FormData()
     data.append('file', 'bbb')
--- a/src/utils/body.ts
+++ b/src/utils/body.ts
@@ -100,10 +100,9 @@ export const parseBody: ParseBody = async (
   const headers = request instanceof HonoRequest ? request.raw.headers : request.headers
   const contentType = headers.get('Content-Type')
 
-  if (
-    contentType?.startsWith('multipart/form-data') ||
-    contentType?.startsWith('application/x-www-form-urlencoded')
-  ) {
+  const mediaType = contentType?.split(';')[0].trim().toLowerCase()
+
+  if (mediaType === 'multipart/form-data' || mediaType === 'application/x-www-form-urlencoded') {
     return parseFormData(request, { all, dot })
   }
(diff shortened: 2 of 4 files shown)


yusukebe:
The issue was closed by another PR.

PR #5070, opened 2026-07-02
feat: add first-class QUERY method support

shellhaki:
## Summary

Adds built-in `app.query(...)` support for HTTP `QUERY` while keeping `app.on('QUERY', ...)` fully compatible.

## Changes

- Add `query` to shared method list.
- Add `query` handler interface on core Hono class.
- Include `$query` in client method expansion coverage.
- Add runtime tests for `app.query(...)` and `app.on('QUERY', ...)` across `RegExpRouter` and `TrieRouter`.

## Files

- src/router.ts
- src/hono-base.ts
- src/client/types.ts
- src/client/types.test.ts
- src/hono.test.ts

## Compatibility

- No breaking changes.
- Existing `app.on('QUERY', ...)` remains supported.

## Verification

```bash
bunx vitest --run src/hono.test.ts src/client/types.test.ts
```

Result: passing.

Manual check:

`curl` is used here intentionally because many REST/API client tools do not yet support the `QUERY` method token reliably.

```bash
curl -i -X QUERY 'http://localhost:4000/echo' \
  -H 'Content-Type: application/json' \
  --data '{"foo":"hello","fek":"bar"}'
```

Expected:

```http
HTTP/1.1 200 OK
Content-Type: application/json
```

```json
{"fek":"bar","foo":"hello"}
```

## Scope Note >-<>

Cache middleware behavior for `QUERY` was intentionally not changed in this PR.

### DIFF
--- a/src/client/types.test.ts
+++ b/src/client/types.test.ts
@@ -117,6 +117,7 @@ describe('app.all()', () => {
     expectTypeOf(client['all-route']).toHaveProperty('$delete')
     expectTypeOf(client['all-route']).toHaveProperty('$options')
     expectTypeOf(client['all-route']).toHaveProperty('$patch')
+    expectTypeOf(client['all-route']).toHaveProperty('$query')
   })
 
   it('should have correct return type for expanded methods', async () => {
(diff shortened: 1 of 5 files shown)


usualoma:
Thanks for implementing it.
I think that's a good first step 👍

yusukebe (approved):
LGTM!

PR #5069, opened 2026-07-02
fix(request,utils/body): break request.ts <-> utils/body.ts import cycle

momomuchu:
## Summary

Fixes #5068. `src/utils/body.ts` imported the `HonoRequest` class from `../request` at runtime solely to run `instanceof HonoRequest` checks, while `src/request.ts` imports `parseBody` from `./utils/body` at runtime. This creates a circular import between the two compiled modules, which bundlers like Rollup flag:

```
(!) Circular dependency
dist/request.js -> dist/utils/body.js -> dist/request.js
```

Reproduced against a real `bun run build` output with `rollup` — the warning is gone after this change.

## Root cause

Two-way runtime dependency:
- `request.ts` → `utils/body.ts` (imports `parseBody`)
- `utils/body.ts` → `request.ts` (imports `HonoRequest` for `instanceof`)

## Fix

Duck-type `HonoRequest` instead of `instanceof`, following the same pattern already used in this codebase for `GET_MATCH_RESULT` in `src/request/constants.ts` (a symbol-only module with no other dependencies).

- Added `HONO_REQUEST: unique symbol` to `src/request/constants.ts`.
- `HonoRequest` now exposes `readonly [HONO_REQUEST] = true as const`.
- `utils/body.ts` imports `HonoRequest` as a **type-only** import (erased at compile time — no runtime edge) and checks `HONO_REQUEST in request` instead of `instanceof HonoRequest`.

The marker is only present on `HonoRequest` instances, never on a plain `Request`, so discrimination for the `HonoRequest | Request` union in `parseBody`/`parseFormData` is exact.

## Test evidence

- Added a regression test in `request.test.ts` for the behavior the `instanceof` check protected (after `req.parseBody()` consumes the body, `req.formData()` must reuse the cached FormData). Verified it fails under a negative control (cache-guard disabled) and passes before and after the refactor.
- Full suite: `bun run test` → 145/145 files, 4579/4579 tests passing.
- `bun run lint` / `bun run format`: clean, no new warnings.
- Rebuilt `dist/` and re-ran Rollup: circular-dependency warning gone.

Closes #5068

PR #5071, opened 2026-07-02
fix: avoid circular dependency between body.ts and request.ts

### LINKED ISSUE #5068: Circular dependency warning

4.12.27

npm 11.11.0

I'm building my Hono project with Rollup:
```js
import commonjs from '@rollup/plugin-commonjs'
import resolve from '@rollup/plugin-node-resolve'
import typescript from '@rollup/plugin-typescript'
import del from 'rollup-plugin-delete'
import dotenv from 'dotenv'

dotenv.config()

export default {
  input: 'src/server.ts',
  output: [
    {
      file: 'dist/server.js',
      format: 'esm',
      inlineDynamicImports: true,
    },
  ],
  plugins: [
    del({ targets: 'dist/*' }),
    resolve(),
    commonjs(),
    typescript({ tsconfig: './tsconfig.json', rootDir: 'src' }),
  ],
}
```

Where I get the following warning:
Image

It does build, so this is a low priority bug but I'm reporting it either way cause it's not right.

The warning is right.

### What do you see instead?

(!) Circular dependency
node_modules/hono/dist/request.js -> node_modules/hono/dist/utils/body.js -> node_modules/hono/dist/request.js

### Additional information

My tsconfig.json:
```ts
{
  "compilerOptions": {
    // Environment setup & latest features
    "typeRoots": ["./node_modules/@types", "./src/cloudflow.d.ts"],
    "types": ["bun"],
    "target": "es2016",
    "module": "preserve",
    "moduleResolution": "bundler",

    // Best practices
    "strict": true,
    "skipLibCheck": true,
    "noFallthroughCasesInSwitch": true,
    "noUncheckedIndexedAccess": true,
    "noImplicitOverride": true,
    "sourceMap": false,
    "declaration": false,
    "removeComments": false,
    "pretty": true,

    "rootDir": "src",
    "outDir": "dist"
  },
  "include": ["src/**/*"]
}
```

My package.json (relevant parts):

```json
"scripts": {
    "dev": "rollup -c -w",
    "build": "rollup -c && prettier --write dist/*"
  },
  "devDependencies": {
    "@rollup/plugin-commonjs": "^29.0.3",
    "@rollup/plugin-node-resolve": "^16.0.3",
    "@rollup/plugin-replace": "^6.0.3",
    "@rollup/plugin-typescript": "^12.3.0",
    "@types/bun": "^1.3.14",
    "dotenv": "^17.4.2",
    "prettier": "^3.9.4",
    "rollup": "^4.62.2",
    "rollup-plugin-delete": "^3.0.2",
    "tslib": "^2.8.1",
    "typescript": "^6.0.3"
  },
  "dependencies": {
    "@google/genai": "^2.10.0",
    "hono": "^4.12.27",
    "hono-rate-limiter": "^0.5.3"
  }
```

usualoma:
fixes #5068

This resolves the circular dependency issue.

Although PR #5069 has also been submitted, that approach is too complex, and I don’t think Hono can accept the overhead of adding properties to `HonoRequest`.

Since it’s likely that `HonoRequest` will never have a `headers` property, I believe this solution will be stable under these conditions.

The bundle size should actually be slightly smaller, since there are fewer imports.

- [x] Run tests

### DIFF
--- a/src/utils/body.ts
+++ b/src/utils/body.ts
@@ -3,7 +3,7 @@
  * Body utility.
  */
 
-import { HonoRequest } from '../request'
+import type { HonoRequest } from '../request'
 import { bufferToFormData } from './buffer'
 
 type BodyDataValueDot = { [x: string]: string | File | BodyDataValueDot }
@@ -74,6 +74,8 @@ export type ParseBodyOptions = {
   dot: boolean
 }
 
+const isRawRequest = (request: HonoRequest | Request): request is Request => 'headers' in request
+
 /**
  * Parses the body of a request based on the provided options.
  *
@@ -98,7 +100,7 @@ export const parseBody: ParseBody = async (
 ) => {
   const { all = false, dot = false } = options
 
-  const headers = request instanceof HonoRequest ? request.raw.headers : request.headers
+  const headers = isRawRequest(request) ? request.headers : request.raw.headers
   const contentType = headers.get('Content-Type')
 
   const mediaType = contentType?.split(';')[0].trim().toLowerCase()
@@ -122,10 +124,10 @@ async function parseFormData<T extends BodyData>(
   request: HonoRequest | Request,
   options: ParseBodyOptions
 ): Promise<T> {
-  const headers = request instanceof HonoRequest ? request.raw.headers : request.headers
+  const headers = isRawRequest(request) ? request.headers : request.raw.headers
   const arrayBuffer = await (request as Request).arrayBuffer()
   const formDataPromise = bufferToFormData(arrayBuffer, headers.get('Content-Type') || '')
-  if (request instanceof HonoRequest) {
+  if (!isRawRequest(request)) {
     // Cache so that a later `c.req.formData()` reuses the already-consumed body
     request.bodyCache.formData = formDataPromise as unknown as FormData
   }
--- a/src/utils/body.ts
+++ b/src/utils/body.ts
@@ -74,7 +74,7 @@ export type ParseBodyOptions = {
   dot: boolean
 }
 
-const isRawRequest = (request: HonoRequest | Request): request is Request => 'headers' in request
+const isRawRequest = (request: HonoRequest | Request): request is Request => 'headers' in request // 'headers' method exists only on Request, not on HonoRequest.
 
 /**
  * Parses the body of a request based on the provided options.


src/utils/body.ts
```diff
@@ -74,6 +74,8 @@ export type ParseBodyOptions = {
   dot: boolean
 }
 
+const isRawRequest = (request: HonoRequest | Request): request is Request => 'headers' in request
```
yusukebe: This is clever!!

BlankParticle:
> Since it’s likely that HonoRequest will never have a headers property, I believe this solution will be stable under these conditions.

wouldn't it be better to use `req.constructor.name === "HonoRequest"` for the condition?

usualoma:
Hi @BlankParticle,
Thanks for the comment!

I’d avoid `req.constructor.name` because constructor names can change during minification, so this check may become unstable in bundled/minified output.

BlankParticle:
> Hi @BlankParticle, Thanks for the comment!
> 
> I’d avoid `req.constructor.name` because constructor names can change during minification, so this check may become unstable in bundled/minified output.

makes sense, right now its stable but will cause issues in minification, 
but `'headers' in request` looks very random, and there is no explanation attached either
I think its better to target known properties on `HonoRequest` instead of the other way
a better check would probably be `"addValidatedData" in request` (or any other specific propertly on HonoRequest) to validate the object is a hono request, with some comment added on top to explain the reasoning

usualoma:
Hi @BlankParticle,
Thanks for the comment.

I understand what you're saying. However, `global.Request` could be overridden or extended. So, there's a slight chance (though I don't think that's actually the case) that `global.Request` might have an `addValidatedData` property.

I think it would be more reliable to check for “properties that `HonoRequest` doesn't have” instead.

BlankParticle:
I guess I will agree with that, but we should probably use a better property, Request has `headers` and HonoRequest has `header`, it will be confusing as the properties are named similarly

file-fba35cf74e44a157160600bf8a65f5d5

usualoma:
Well, it’s true that it can be hard to tell the difference when you’re tired. That said, I don’t think it’ll be a problem when someone reads this code in the future.

Since `headers` is the method that’s mainly called after conditional checks within this file, I think there’s some logic to using that as a criterion.

CleanShot 2026-07-02 at 22 42 00@2x

For now, I’ve just added a comment.
https://github.com/honojs/hono/pull/5071/changes/8d01de5c603aa8d3cd65693946a2349d0124193b

BlankParticle:
hmm, given the use-case, I think its fine to use headers then

yusukebe:
I also thought using `headers` would be confused with `header`. If we do it more strictly, I think we will need to add a dedicated property (tag) to distinguish HonoRequest from Request.

```ts
class HonoRequest {
  readonly isHonoRequest = true
  // ...
}
```

But in this case, it's okay to use `headers` now. We will be able to consider whether it's needed.

ro80t:
I think `existsHeadersField` or `hasHeadersField` would be more appropriate names for the function than `isRawRequest`.

I apologize if this sounds like nitpicking, but I would appreciate your thoughts on this.

usualoma:
I really appreciate everyone's input, but I don't think a change is necessary.

yusukebe (approved):
LGTM!

yusukebe:
@usualoma Thank you so much! Great idea.

PR #5073, opened 2026-07-03
ci: add CVE Lite dependency audit workflow

sonukapoor:
This PR adds a CVE Lite CLI dependency audit workflow to Hono's CI pipeline.

A scan of the current `bun.lock` found **13 high-severity** and **7 medium-severity** advisories, including several direct dependencies with known vulnerabilities:

**Direct high-severity findings (copy-run fix commands available):**
- `minimatch@3.1.2` - ReDoS via crafted glob pattern (GHSA-f8q6-p94x-37v3) - `bun add minimatch@10.2.3`
- `picomatch@2.3.1` - ReDoS via crafted patterns (GHSA-9h6g-pr28-7cqp) - appears twice with different fix paths
- `undici@6.x` (two versions) - request smuggling / injection (GHSA-3wzh-ggcw-3hpv, GHSA-cqmj-92xf-r6r9)
- `vite@5.x` - server-side request forgery (GHSA-64vr-g452-qvp3)
- `wrangler@3.x` - path traversal via dev server (GHSA-cm93-56gx-m49q)
- `ws@8.x` - DoS via crafted HTTP upgrade request (GHSA-3h5q-q39x-f9x2)

The workflow runs on push, pull request, and weekly schedule. Findings are uploaded to GitHub's Security tab as SARIF so maintainers can triage inline on PRs.

**Tool:** [CVE Lite CLI](https://github.com/OWASP/cve-lite-cli) is an OWASP Lab Project. It reads the lockfile locally, queries the OSV database, and produces actionable fix commands for direct dependencies.

PR #5080, opened 2026-07-03
fix: upgrade undici, wrangler, ws, np, and vitest to patched versions

sonukapoor:
CVE Lite CLI (PR #5073) found several direct devDependencies at vulnerable versions: `undici`, `wrangler`, `ws`, `np`, and `vitest`. This upgrades each to the latest version where the advisories are patched.


yusukebe:
I will do it manually.

Issue #4326, opened 2025-07-30
Context missing if JSXNode between Context and consumer


danelowe:
### What version of Hono are you using?

4.8.9

### What runtime/platform is your app running on? (with version if possible)

deno 2.2.12 (stable, release, aarch64-apple-darwin)

### What steps can reproduce the bug?

In the example below, the `ExtraDiv` component does not propagate the context. 

It appears that `toString()` is called on `MyConsumer` before the MyContext.Provider function is invoked, so the value hasn't yet been injected to the context.

```ts
import {createContext, FC, PropsWithChildren, useContext} from "hono/jsx";

const MyContext = createContext<{ value: string } | null>(null);

const ExtraDiv: FC<PropsWithChildren> = () => {
    return (<MyContext.Provider value={{greeting: "I have an extra div"}}>
        <div>
            <MyConsumer/>
        </div>
    </MyContext.Provider>)
}

const NoExtraDiv: FC<PropsWithChildren> = () => {
    return (<MyContext.Provider value={{greeting: "I do not have an extra div"}}>
        <MyConsumer/>
    </MyContext.Provider>)
}

const ExtraDivAsComponent: FC = () => {
    return (
        <MyContext.Provider value={{greeting: "I have an extra div as a component"}}>
            <Box>
                <MyConsumer/>
            </Box>
        </MyContext.Provider>
    );
}

const Box = ({children}: PropsWithChildren) => {
    return <div class="box">{children}</div>
}


const MyConsumer: FC = () => {
    const context = useContext(MyContext);
    if (!context) {
        return <div>No context provided</div>;
    }
    return <div>{context.greeting}</div>;
};

const MyComponent: FC = () => {
    return (
        <div>
            <NoExtraDiv/>
            <ExtraDivAsComponent/>
            <ExtraDiv/>
        </div>
    );
}


console.log(MyComponent().toString());
```

### What is the expected behavior?

Outputs `<div><div>I have do not have an extra div</div><div class="box"><div>I have an extra div as a component</div></div><div><div>I have an extra div</div></div></div>`

### What do you see instead?

Outputs `<div><div>I have do not have an extra div</div><div class="box"><div>I have an extra div as a component</div></div><div><div>No context provided</div></div></div>`

### Additional information

It appears that `toString()` is called on `MyConsumer` before the MyContext.Provider function is invoked, so the value hasn't yet been injected to the context.

lemorage:
[...]

PR #5082, opened 2026-07-03
fix(jsx): preserve context through intermediate DOM elements in precompile mode

momomuchu:
## Summary

Fixes the context loss in JSX **precompile** mode reported in #4326: when a plain DOM element sits between a `<Context.Provider>` and its consumer, the consumer received the default value instead of the provided one. `react-jsx` mode was already correct; only precompile was affected.

## Root cause

Deno's precompile transform emits an intermediate `<div>` as `jsxTemplate(['<div>','</div>'], jsx(Consumer, null))`. Because that call is the `children` **argument** of `jsx(Ctx.Provider, …)`, JS evaluates it first. `jsxTemplate` was aliased to the `html` helper, which **eagerly** stringifies interpolated values — so the consumer rendered *before* the Provider pushed its value onto the context stack, and `useContext` returned the default. The no-wrapper and component-wrapper cases worked because their `children` stay lazy `JSXNode`s, rendered later inside the Provider's push/pop window. Only the plain-DOM case took the eager `html` path.

## Fix

Make `jsxTemplate` build a lazy `JSXFragmentNode` from the template strings + interpolated values instead of stringifying eagerly, so rendering defers until the surrounding tree stringifies (inside the Provider window) — mirroring the classic (`react-jsx`) runtime. The public `html` helper is untouched, and interpolated values still flow through the exact same shared escaping primitives (`escapeToBuffer`, `raw`, `childrenToStringToBuffer`), so escaping is unchanged.

## Before / after (issue's `ExtraDiv` case)

`<div><div>No context provided</div></div>` → `<div><div>I have an extra div</div></div>`

## Test evidence

Regression test added in `runtime-tests/deno-jsx/jsx.test.tsx` covering no-wrapper / component-wrapper / plain-DOM shapes, run under **both** the precompile and react-jsx configs. Verified: precompile 13/13, react-jsx 13/13, full vitest suite 1902 passed / 0 failed, `tsc` clean, lint clean. Negative control confirmed (the test fails on `main` without the fix).

## Note

First contribution to Hono — I decoupled `jsxTemplate` from `html` (rather than changing `html`) to preserve the public helper's eager semantics. Happy to adjust the layering if you'd prefer a different approach. Refs #4326.

### DIFF
--- a/runtime-tests/deno-jsx/jsx.test.tsx
+++ b/runtime-tests/deno-jsx/jsx.test.tsx
@@ -1,7 +1,9 @@
 /** @jsxImportSource ../../src/jsx */
 import { assertEquals } from '@std/assert'
 import { Style, css } from '../../src/helper/css/index.ts'
+import { createContext, useContext } from '../../src/jsx/context.ts'
 import { Suspense, renderToReadableStream } from '../../src/jsx/streaming.ts'
+import type { FC, PropsWithChildren } from '../../src/jsx/types.ts'
 import type { HtmlEscapedString } from '../../src/utils/html.ts'
 import { HtmlEscapedCallbackPhase, resolveCallback } from '../../src/utils/html.ts'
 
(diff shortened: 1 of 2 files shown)


usualoma:
The issue might be fixed, but if we start accepting pull requests that simply have the AI fix existing issues, things will get out of hand, so I think it’s appropriate to close this pull request.

yusukebe:
Closing

Issue #4975, opened 2026-05-27
Router: /** wildcard stops matching when routes with overlapping parameterized segments exist


SSardorf:
## Version

`hono@4.12.23`

## What is the expected behavior?

A `/**` wildcard route should match regardless of what other routes are registered on the same app.

## What is the actual behavior?

A `/**` wildcard route returns 404 when `SmartRouter` falls back from `RegExpRouter` to `TrieRouter` (e.g., when overlapping parameterized segments exist).

## Minimal reproduction

```ts
import { Hono } from 'hono';

const app = new Hono();
app.all('/auth/**', (c) => c.text('auth'));
app.get('/pkg/:name/:version', (c) => c.text('pkg'));
app.get('/pkg/:name/versions', (c) => c.text('versions'));

const r1 = await app.fetch(new Request('http://localhost/auth/test'));
const r2 = await app.fetch(new Request('http://localhost/pkg/lodash/1.0'));
const r3 = await app.fetch(new Request('http://localhost/pkg/lodash/versions'));

console.log('/auth/test:', r1.status);      // 404 ← BUG (expected 200)
console.log('/pkg/lodash/1.0:', r2.status);  // 200
console.log('/pkg/lodash/versions:', r3.status); // 200
```

## Root Cause

Two issues in `TrieRouter`'s handling of `**`:

1. **`getPattern('**')` returns `null`** (`src/utils/url.ts`) — only `'*'` is recognized as a wildcard, so `**` is treated as a literal path segment
2. **`node.insert()` stores the child under key `'**'`** (`src/router/trie-router/node.ts`) — but `search()` looks up `children['*']`, so the wildcard node is never found

Most apps don't hit this because `SmartRouter` defaults to `RegExpRouter` (which handles `**`). The bug surfaces when `RegExpRouter` can't handle the route set (overlapping param segments like `/:name/:version` + `/:name/versions`) and falls back to `TrieRouter`.

**Direct reproduction on TrieRouter:**
```ts
import { Hono } from 'hono';
import { TrieRouter } from 'hono/router/trie-router';

const app = new Hono({ router: new TrieRouter() });
app.all('/auth/**', (c) => c.text('auth'));
app.get('/other', (c) => c.text('other'));

await app.fetch(new Request('http://localhost/auth/test'));   // 404 ← always fails
await app.fetch(new Request('http://localhost/other'));       // 200
```

## Fix

PR: #4977 — two-line fix:
[...]

PR #5088, opened 2026-07-05
fix(router/trie-router): match trailing /* wildcard on empty remainder after regexp param

### LINKED ISSUE #5087: TrieRouter: /* wildcard does not match empty remainder after a regexp-constrained param
### Description

### Steps to reproduce
```ts
```
Matrix:
```
```

### Root cause

### Suggested fix

tsushanth:
## What

The RegExp-match branch of `TrieRouter.search()` (`src/router/trie-router/node.ts`) did not collect `child.#children['*']` for an empty remainder. As a result, `/:id{[0-9]+}/*` failed to match `/12`, even though the plain-param equivalent `/:id/*` matched it.

This mirrors the plain-param branch's empty-wildcard collection, gated on the regexp consuming the full remainder (`m[0].length === restPathString.length`), so the trailing `/*` is picked up when nothing follows the matched segment.

## Repro matrix (route `/:id{[0-9]+}/*`)

| Path | Before | After |
|------|--------|-------|
| `/12` | no match | matches ✅ |
| `/12/y` | matches | matches (unchanged) |
| `/abc` | no match | no match (unchanged) |

## Testing

- Added a regression test in `src/router/trie-router/node.test.ts` covering all three cases above.
- Full router test suite passes: `npx vitest --run src/router` → 523 passed, 20 skipped. `tsc --noEmit` clean. trie-router `node.ts` coverage now 100%.

Note: this is distinct from the `/**` double-star wildcard issues (#4623 / #4975) — it concerns a single trailing `/*` after a regexp param on an empty remainder.

Closes #5087

### DIFF
--- a/src/router/common.case.test.ts
+++ b/src/router/common.case.test.ts
@@ -550,6 +550,46 @@ export const runTest = ({
       })
     })
 
+    describe('Trailing wildcard on empty remainder', () => {
+      beforeEach(() => {
+        router.add('GET', '/static/*', 'static')
+        router.add('GET', '/plain/:id/*', 'plain')
+        router.add('GET', '/regexp/:id{[0-9]+}/*', 'regexp')
+      })
+
+      it('GET /static', () => {
+        const res = match('GET', '/static')
+        expect(res.length).toBe(1)
+        expect(res[0].handler).toEqual('static')
+      })
+
+      it('GET /plain/12', () => {
+        const res = match('GET', '/plain/12')
+        expect(res.length).toBe(1)
+        expect(res[0].handler).toEqual('plain')
+        expect(res[0].params['id']).toBe('12')
+      })
+
+      it('GET /regexp/12', () => {
+        const res = match('GET', '/regexp/12')
+        expect(res.length).toBe(1)
+        expect(res[0].handler).toEqual('regexp')
+        expect(res[0].params['id']).toBe('12')
+      })
+
+      it('GET /regexp/12/y', () => {
+        const res = match('GET', '/regexp/12/y')
+        expect(res.length).toBe(1)
+        expect(res[0].handler).toEqual('regexp')
+        expect(res[0].params['id']).toBe('12')
+      })
+
+      it('GET /regexp/abc', () => {
+        const res = match('GET', '/regexp/abc')
+        expect(res.length).toBe(0)
+      })
+    })
+
     describe('Capture complex multiple directories', () => {
       beforeEach(() => {
         router.add('GET', '/:first{.+}/middle-a/:reference?', '1')
--- a/src/router/linear-router/router.test.ts
+++ b/src/router/linear-router/router.test.ts
@@ -11,6 +11,11 @@ describe('LinearRouter', () => {
           'Multi match > `params` per a handler > GET /entry/123/show',
           'Capture regex pattern has trailing wildcard > GET /foo/bar/file.html',
           'Complex > Parameter with {.*} regexp',
+          'Trailing wildcard on empty remainder > GET /static',
+          'Trailing wildcard on empty remainder > GET /plain/12',
+          'Trailing wildcard on empty remainder > GET /regexp/12',
+          'Trailing wildcard on empty remainder > GET /regexp/12/y',
+          'Trailing wildcard on empty remainder > GET /regexp/abc',
         ],
       },
       {
--- a/src/router/trie-router/node.test.ts
+++ b/src/router/trie-router/node.test.ts
@@ -424,6 +424,26 @@ describe('Multi match', () => {
       expect(res[0][0]).toEqual('middleware a')
     })
   })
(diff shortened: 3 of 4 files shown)


usualoma:
There's no doubt this is a bug that needs to be fixed.
It's a shame that it appears to have been generated by AI, but it's not as if “someone else's bug was simply fixed by AI.” I'm not sure whether I should close this one for now and fix it separately.

usualoma:
If we're going to merge, we should put the tests in `common.case.test.ts`.

```diff
diff --git i/src/router/common.case.test.ts w/src/router/common.case.test.ts
index de3bc693..a94df2ea 100644
--- i/src/router/common.case.test.ts
+++ w/src/router/common.case.test.ts
@@ -550,6 +550,46 @@ export const runTest = ({
       })
     })
 
+    describe('Trailing wildcard on empty remainder', () => {
+      beforeEach(() => {
+        router.add('GET', '/static/*', 'static')
+        router.add('GET', '/plain/:id/*', 'plain')
+        router.add('GET', '/regexp/:id{[0-9]+}/*', 'regexp')
+      })
+
+      it('GET /static', () => {
+        const res = match('GET', '/static')
+        expect(res.length).toBe(1)
+        expect(res[0].handler).toEqual('static')
+      })
+
+      it('GET /plain/12', () => {
+        const res = match('GET', '/plain/12')
+        expect(res.length).toBe(1)
+        expect(res[0].handler).toEqual('plain')
+        expect(res[0].params['id']).toBe('12')
+      })
+
+      it('GET /regexp/12', () => {
+        const res = match('GET', '/regexp/12')
+        expect(res.length).toBe(1)
+        expect(res[0].handler).toEqual('regexp')
+        expect(res[0].params['id']).toBe('12')
+      })
+
+      it('GET /regexp/12/y', () => {
+        const res = match('GET', '/regexp/12/y')
+        expect(res.length).toBe(1)
+        expect(res[0].handler).toEqual('regexp')
+        expect(res[0].params['id']).toBe('12')
+      })
+
+      it('GET /regexp/abc', () => {
+        const res = match('GET', '/regexp/abc')
+        expect(res.length).toBe(0)
+      })
+    })
+
     describe('Capture complex multiple directories', () => {
       beforeEach(() => {
         router.add('GET', '/:first{.+}/middle-a/:reference?', '1')
diff --git i/src/router/linear-router/router.test.ts w/src/router/linear-router/router.test.ts
index 396d3c56..1c5c50a1 100644
--- i/src/router/linear-router/router.test.ts
+++ w/src/router/linear-router/router.test.ts
@@ -11,6 +11,11 @@ describe('LinearRouter', () => {
           'Multi match > `params` per a handler > GET /entry/123/show',
           'Capture regex pattern has trailing wildcard > GET /foo/bar/file.html',
           'Complex > Parameter with {.*} regexp',
+          'Trailing wildcard on empty remainder > GET /static',
+          'Trailing wildcard on empty remainder > GET /plain/12',
+          'Trailing wildcard on empty remainder > GET /regexp/12',
+          'Trailing wildcard on empty remainder > GET /regexp/12/y',
+          'Trailing wildcard on empty remainder > GET /regexp/abc',
         ],
       },
       {
diff --git i/src/router/trie-router/node.test.ts w/src/router/trie-router/node.test.ts
index 4e4ca87f..fa22b01c 100644
--- i/src/router/trie-router/node.test.ts
+++ w/src/router/trie-router/node.test.ts
@@ -424,26 +424,6 @@ describe('Multi match', () => {
       expect(res[0][0]).toEqual('middleware a')
     })
   })
-  describe('Regexp param with trailing wildcard on empty remainder', () => {
-    const node = new Node()
-    node.insert('get', '/:x{[0-9]+}/*', 'wildcard')
-    it('/12 matches the wildcard on empty remainder', () => {
-      const [res] = node.search('get', '/12')
-      expect(res.length).toBe(1)
-      expect(res[0][0]).toEqual('wildcard')
-      expect(res[0][1]['x']).toBe('12')
-    })
-    it('/12/y still matches the wildcard', () => {
-      const [res] = node.search('get', '/12/y')
-      expect(res.length).toBe(1)
-      expect(res[0][0]).toEqual('wildcard')
-      expect(res[0][1]['x']).toBe('12')
-    })
-    it('/abc does not match', () => {
-      const [res] = node.search('get', '/abc')
-      expect(res.length).toBe(0)
-    })
-  })
   describe('Trailing slash', () => {
     const node = new Node()
     node.insert('get', '/book', 'GET /book')
```

yusukebe:
@usualoma 

By the way, of course, It's no problem for you to create a PR by yourself to fix the issue!

tsushanth:
Moved the tests to `common.case.test.ts` as suggested, and added the LinearRouter skip entries for the 5 new cases (it doesn't support regexp params). Also removed the `node.test.ts` block.

usualoma:
Let's close this and reimplement it later.

PR #5090, opened 2026-07-07
fix(client): compose async header functions with per-request headers in deepMerge

### LINKED ISSUE #5089: RPC headers not merging when function

4.12.21

Node 24.15.0

Create a RPC client, passing in headers using a function
```
const client = hc<AppType>('/api', {
  headers: async () => {
    return {
      Authorization: await getToken(),
    }
  },
})
```
Call a api using the client, passing in additional headers
```
await client.search.$get(
  {},
  {
    headers: {
      'X-User-Agent': 'hc',
    },
  }
)
```

Both headers should be added to the api call

### What do you see instead?

Only the headers passed directly in are added to the api call

### Additional information

Looking at the source code, it seems like this bug is caused by the following `deepMerge` call: https://github.com/honojs/hono/blob/a05813c0586cd8ee7873dab8158c08a5f772a9ed/src/client/client.ts#L219

The `deepMerge` seems to only merge objects, not functions: https://github.com/honojs/hono/blob/a05813c0586cd8ee7873dab8158c08a5f772a9ed/src/client/utils.ts#L74-L75

Not sure if the `deepMerge` function should be updated to support async functions or if the function should be called before merging

AmariahAK:
## PR Description

Fixes #5089

i) What was the issue
When an RPC client is created with headers as an async function (e.g.
async () => ({ Authorization: await getToken() })) and a per-request
call passes headers as an object, the client-level function headers
were silently discarded. Only the per-request object headers were sent.

ii) Where was the issue
In deepMerge (src/client/utils.ts), the if/else checks isObject() on
both values before composing. Since isObject() returns false for
functions, when the target headers is a function and the source
headers is an object, the else branch overwrites the function with
the object instead of composing them. The call site is line 219 in
src/client/client.ts.

iii) How it was fixed
Added a branch in deepMerge that detects when the target value is a
function and the source value is an object. Instead of overwriting,
it composes a new async function that resolves the original and
merges its result with the per-request headers (per-request keys
take priority on conflicts).

iv) Why this method over others
- Fixed in deepMerge (single source of truth) rather than the call
  site — prevents the bug anywhere deepMerge is used
- Keeps headers lazy — the composed function is still called per-
  request in ClientRequestImpl.fetch, preserving dynamic values
  like tokens and timestamps
- Preserves all existing merge behaviors (function+function →
  overwrite, object+object → deep merge)
- Minimal diff — no changes to the proxy chain or async refactoring

v) How to test it locally
bun run test
The new test "Should merge async client headers with per-request
headers" in src/client/client.test.ts verifies both x-hono (from
the async function) and x-dynamic (from the per-request object)
arrive at the server.
```

## Checklist

- [x] Add tests
- [x] Run tests
- [x] Add TSDoc/JSDoc to document the code

### DIFF
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -1209,8 +1209,10 @@ describe('Dynamic headers', () => {
   const server = setupServer(
     http.post('http://localhost/posts', async ({ request }) => {
       const requestDynamic = request.headers.get('x-dynamic')
+      const requestHono = request.headers.get('x-hono')
       const payload = {
         requestDynamic,
+        requestHono,
       }
       return HttpResponse.json(payload)
     })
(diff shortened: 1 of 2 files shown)


AmariahAK:
@usualoma @yusukebe 
Kindly review

PR #5095, opened 2026-07-07
fix: make query() return type include undefined for missing params

### LINKED ISSUE #3426: `ctx.req.query()` should return `Record<string, string | undefined>`

4.6.2

Not matter

I think the return type of `c.req.query()` should be `Record<string, string | undefined>` instead of `Record<string, string>`

The latter will cause the following code to compile without errors, but to actually run with errors:
```ts
const { name } = c.req.query();
return c.text(name.toUpperCase());
```

The return type of `c.req.query()` should be `Record<string, string | undefined>`

### What do you see instead?

`Record<string, string>`

### Additional information

_No response_

Nabeel-akk:
Closes #3426

The `query()` method returns an object from `getQueryParam()`, which can have undefined values when a query parameter is missing. But the return type says `Record<string, string>`, which lets code like this compile but crash at runtime:

```
const { name } = c.req.query();
return c.text(name.toUpperCase());
```

Changing the return type to `Record<string, string | undefined>` forces users to handle the undefined case before using the value.

This is a one-line type change and is backward compatible since widening the value type is safe.

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -146,7 +146,7 @@ export class HonoRequest<P extends string = '/', I extends Input['out'] = {}> {
    * ```
    */
   query(key: string): string | undefined
-  query(): Record<string, string>
+  query(): Record<string, string | undefined>
   query(key?: string) {
     return getQueryParam(this.url, key)
   }


yusukebe:
This introduces a breaking change. We plan to fix the issue in the next major release. This is not needed now.

PR #5102, opened 2026-07-09
fix(trie-router): match empty wildcard remainder after regexp param

### LINKED ISSUE #5087: TrieRouter: /* wildcard does not match empty remainder after a regexp-constrained param

usualoma:
fixes #5087

* The implementation ultimately matches #5088, as this is the simplest and clearest way to express the fix.
* Corrected an apparent typo in src/router/trie-router/node.test.ts.
* Added only the minimal common-router test needed to cover this regression.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/router/trie-router/node.test.ts
+++ b/src/router/trie-router/node.test.ts
@@ -418,10 +418,11 @@ describe('Multi match', () => {
       expect(res[1][0]).toEqual('regexp')
       expect(res[1][1]['id']).toBe('123')
     })
-    it('/regexp-abc/123', () => {
-      const [res] = node.search('get', '/regex-abc/123/ghi')
+    it('/regex-abc/123', () => {
+      const [res] = node.search('get', '/regex-abc/123')
       expect(res.length).toBe(1)
       expect(res[0][0]).toEqual('middleware a')
+      expect(res[0][1]['id']).toBe('123')
     })
   })
   describe('Trailing slash', () => {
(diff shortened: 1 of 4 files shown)


usualoma:
Hi @yusukebe,

src/router/trie-router/node.test.ts
```diff
@@ -418,10 +418,11 @@ describe('Multi match', () => {
...
       expect(res[1][1]['id']).toBe('123')
     })
-    it('/regexp-abc/123', () => {
-      const [res] = node.search('get', '/regex-abc/123/ghi')
+    it('/regex-abc/123', () => {
+      const [res] = node.search('get', '/regex-abc/123')
       expect(res.length).toBe(1)
       expect(res[0][0]).toEqual('middleware a')
+      expect(res[0][1]['id']).toBe('123')
     })
```
yusukebe, line 432: How about adding the following test?

```ts
it('/regex-abc/123/ghi', () => {
  const [res] = node.search('get', '/regex-abc/123/ghi')
  expect(res.length).toBe(1)
  expect(res[0][0]).toEqual('middleware a')
  expect(res[0][1]['id']).toBe('123')
})
```

yusukebe:
@usualoma 

Thank you so much! As you said, the tests in `src/router/trie-router/node.test.ts` indeed should be fixed. I've left a comment.

src/router/trie-router/node.test.ts
usualoma: Thanks for pointing that out. You're absolutely right!

While I was at it, I also added a test with the same conditions to src/router/common.case.test.ts.

71c99170

usualoma: Oh, and I renamed the test in commit f73f2d6e.

yusukebe (approved):
LGTM!

PR #5103, opened 2026-07-10
fix(types): strip extra properties from array types in JSONParsed

### LINKED ISSUE #3677: Types not inferred correctly for Arrays with additional properties anymore

6.6.10

Browser

```
import type { JSONParsed } from 'hono/utils/types';

type TestWorking = { foo: 'bar' }[];
type TestBroken = { foo: 'bar' }[] & { meta: 'v1' };
type ResultWorking = JSONParsed<TestWorking>;
type ResultBroken = JSONParsed<TestBroken>;
```

```
type ResultWorking = {
    foo: "bar";
}[]
```

### What do you see instead?

```
type ResultBroken = {
    [x: number]: {
        foo: "bar";
    };
    length: number;
    toString: null;
    toLocaleString: null;
    pop: null;
    push: {};
    concat: {};
    join: {};
    reverse: null;
    shift: null;
    slice: {};
    sort: {};
    splice: {};
    unshift: {};
    indexOf: {};
    ... 27 more ...;
    meta: "v1";
}
```

### Additional information

We are using `postgres.js` and sometimes want to return a RowList, which is an Array, extended with some additional properties.
When returning an Array with additional properties in hono, it will not be resolved as a simple Array (even though the JSON transformation will remove the additional properties). 

So in this case, the frontend code will expect the returned Response to have a property `meta`, even though it will be stripped in transport. This was working as expected in earlier version. See https://github.com/peak-15/peak/blob/bbb61321c8e94fa91a18ed9b1b5c74dbf99edf07/apps/frontend/package.json.

Arman-Luthra:
Fixes #3677

When an array type is intersected with additional properties (e.g. postgres.js's `RowList`, which is `Row[] & { count: number, ... }`), `JSONParsed` mapped over all keys of the intersection — including array methods and the extra properties — producing a broken object type instead of an array:

```ts
type Broken = JSONParsed<{ foo: 'bar' }[] & { meta: 'v1' }>
// { [x: number]: { foo: "bar" }; length: number; pop: null; push: {}; ...; meta: "v1" }
```

The homomorphic mapped type in the `ReadonlyArray` branch (`src/utils/types.ts:66`) only preserves array-ness when `T` is exactly an array or tuple type, not for intersections. Since `JSON.stringify` serializes such values as plain arrays and drops the extra properties, the parsed type should be a plain array too:

```ts
type Fixed = JSONParsed<{ foo: 'bar' }[] & { meta: 'v1' }>
// { foo: 'bar' }[]
```

The fix keeps the mapped type result when it is still an array (plain arrays and tuples are unchanged, including element conversion and readonly modifiers) and only falls back to element-wise mapping over `T[number]` when the mapped result lost array-ness, i.e. when extra properties were intersected in.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/utils/types.test.ts
+++ b/src/utils/types.test.ts
@@ -116,6 +116,16 @@ describe('JSONParsed', () => {
       type Expected = { key: readonly number[] }
       expectTypeOf<Actual>().toEqualTypeOf<Expected>()
     })
+    it('should convert T[] with additional properties to T[]', () => {
+      type Actual = JSONParsed<{ foo: 'bar' }[] & { meta: 'v1' }>
+      type Expected = { foo: 'bar' }[]
+      expectTypeOf<Actual>().toEqualTypeOf<Expected>()
+    })
+    it('should convert elements of T[] with additional properties', () => {
+      type Actual = JSONParsed<(number | undefined)[] & { meta: 'v1' }>
+      type Expected = (number | null)[]
+      expectTypeOf<Actual>().toEqualTypeOf<Expected>()
+    })
   })
 
   describe('tuple', () => {
--- a/src/utils/types.ts
+++ b/src/utils/types.ts
@@ -63,7 +63,11 @@ export type JSONParsed<T, TError = bigint | ReadonlyArray<bigint>> = T extends {
     : T extends InvalidJSONValue
       ? never
       : T extends ReadonlyArray<unknown>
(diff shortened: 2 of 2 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@Arman-Luthra Thanks.

PR #5111, opened 2026-07-13
feat(etag): support conditional requests for the QUERY method

Cherry:
Similar to https://github.com/expressjs/express/issues/7365, but for Hono. Extends the etag middleware etag support conditional validation for QUERY methods as well as GET/HEAd.

`QUERY` is a safe, idempotent, cacheable method whose responses explicitly support conditional revalidation, as per datatracker.ietf.org/doc/rfc10008, so it should behave like GET/HEAD here.

Will pair nicely with https://github.com/honojs/hono/pull/5070 once that lands. Includes tests.

### DIFF
--- a/src/middleware/etag/index.test.ts
+++ b/src/middleware/etag/index.test.ts
@@ -226,6 +226,53 @@ describe('Etag Middleware', () => {
     expect(res.status).toBe(304)
   })
 
+  it('Should handle conditional QUERY requests', async () => {
+    const app = new Hono()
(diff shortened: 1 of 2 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@Cherry 

Thanks, James! Looks good. I'll merge this later and include the next minor release v4.13

Cherry:
Thank you! 🧡

Issue #5048, opened 2026-06-27
Add first-class support for the HTTP QUERY method


shellhaki:
## Title
Add first-class support for the HTTP `QUERY` method

### Summary
I’d like to propose adding `QUERY` as a first-class HTTP method in Hono, similar to the built-in helpers for `GET`, `POST`, `PUT`, `DELETE`, `OPTIONS`, and `PATCH`.

Hono already supports arbitrary methods through `app.on('QUERY', ...)`, so this is not about immediate compatibility. The goal is to standardize `QUERY` into the main method API for better ergonomics, discoverability, and long-term consistency as the HTTP ecosystem evolves.

### Motivation
My understanding is that `app.on(...)` is the general escape hatch for custom or newer methods, which is useful for immediate support. But once a method becomes standardized and widely recognized, it feels better to expose it in the same way as the other core HTTP methods.

In practice, first-class support would make `QUERY`:
- easier to discover in the public API
- consistent with Hono’s built-in method helpers
- more obvious in docs and examples
- easier to type-check and use with chaining patterns

### Proposed change
If the maintainers think this is appropriate, I’d suggest:
- adding `query` to the built-in method helpers
- ensuring routing/matching works correctly for `QUERY`
- adding tests for `app.query(...)` and `app.on('QUERY', ...)`
- updating docs if needed, including migration/docs references where appropriate

### Questions
1. Is Hono open to standardizing `QUERY` as a built-in method helper now?
2. If yes, is the preferred API `app.query(...)` or another naming convention?
3. Should this be implemented only for the core app API, or also reflected in types/docs/route helper patterns?

### Notes
I’m happy to work on the implementation if this is considered a good fit for the project.

yusukebe:
Hi @shellhaki 

Thank you for the proposal.

Yes. "Query" is now standardized. So, let's make Hono support it.

> 2. If yes, is the preferred API app.query(...) or another naming convention?

It's confusing because we already have `c.req.query()`, but there is no better name, so I think `app.query()` is okay.

> 3. Should this be implemented only for the core app API, or also reflected in types/docs/route helper patterns?

Yes. To keep consistency, we should change them.

Hey @usualoma ! What do you think of introducing "Query" things?

shellhaki:
Thanks! I’d be happy to work on the implementation once there’s consensus. I’ll wait for @usualoma thoughts before opening a PR.

usualoma:
Hi @shellhaki and @yusukebe,
[...]

PR #5115, opened 2026-07-14
feat(cors): allow QUERY by default as a first-class method

usualoma:
#5048

### Important behavioral change

> [!IMPORTANT]
> This PR adds `QUERY` to the default `allowMethods` list used by the CORS middleware.
>
> As a result, applications using `cors()` without an explicit `allowMethods` option will advertise `QUERY` in the `Access-Control-Allow-Methods` response header. This expands the default set of methods permitted for cross-origin requests.

This is not a breaking API change in the strict sense, but it is a security-relevant change to the default behavior and should be highlighted in the release notes. Applications that explicitly configure `allowMethods` are not affected.

### Rationale

Hono now treats `QUERY` as a first-class HTTP method. For that support to be consistent across the framework, the built-in CORS middleware should allow `QUERY` by default, just as it does for the other first-class methods.

Requiring every application to add `QUERY` to `allowMethods` manually would make the support incomplete and less discoverable. Expanding the default list is therefore considered appropriate for first-class `QUERY` support.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/cors/index.test.ts
+++ b/src/middleware/cors/index.test.ts
@@ -145,13 +145,26 @@ describe('CORS by Middleware', () => {
   })
 
   it('Preflight default', async () => {
-    const req = new Request('https://localhost/api/abc', { method: 'OPTIONS' })
-    req.headers.append('Access-Control-Request-Headers', 'X-PINGOTHER, Content-Type')
+    const req = new Request('https://localhost/api/abc', {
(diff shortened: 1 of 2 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@usualoma Thanks!

PR #5127, opened 2026-07-17
fix(sse): validate only string fields and emit retry: 0

kridaydave:
Fixes the SSE validation loop treating the numeric retry field as a string. Now only event and id are checked for \r\n, and retry: 0 is correctly emitted instead of being skipped.


kridaydave:
@yusukebe why was this pr closed ? Just curious

BlankParticle:
> why was this pr closed ? Just curious

I cant speak for yusukebe, but probably related to this, 
https://x.com/yusukebe/status/2079520915667492949

PR #5132, opened 2026-07-17
feat(middleware): add method-not-allowed middleware

usualoma:
#4633

Introduce an opt-in `methodNotAllowed` middleware. It returns `405 Method Not Allowed` with an `Allow` header when the path exists for another method.

Compared with the implementation discussed in the issue, this version:

- accepts `{ app, onMethodNotAllowed? }` to support custom responses
- assigns the response to `c.res` instead of throwing an `HTTPException`, allowing outer middleware to observe the final status without invoking `onError`
- follows Hono's routing behavior for HEAD/GET, custom methods, and case-sensitive method matching
- supports middleware installed in sub-apps composed with `app.route()`
- preserves error-handler 404 responses and removes stale representation headers when replacing a 404

### Why assign the response to `c.res` instead of throwing an `HTTPException`?

This follows the existing response-transforming middleware pattern used by `trailingSlash`, `etag`, and `prettyJSON`, which replace `c.res` after `await next()`.

A `405 Method Not Allowed` response here is a normal routing outcome derived from a downstream `404 Not Found`, rather than an exceptional failure. Assigning it to `c.res` preserves normal middleware unwinding, lets outer middleware observe the final `405` status, and avoids invoking the application's `onError` handler.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/jsr.json
+++ b/jsr.json
@@ -45,6 +45,7 @@
     "./timeout": "./src/middleware/timeout/index.ts",
     "./timing": "./src/middleware/timing/timing.ts",
     "./logger": "./src/middleware/logger/index.ts",
+    "./method-not-allowed": "./src/middleware/method-not-allowed/index.ts",
     "./method-override": "./src/middleware/method-override/index.ts",
     "./powered-by": "./src/middleware/powered-by/index.ts",
     "./pretty-json": "./src/middleware/pretty-json/index.ts",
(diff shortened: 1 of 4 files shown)


usualoma:
Hi @yusukebe,

yusukebe (commented):
LGTM!

yusukebe:
Hey @usualoma !

I think this is perfect! Great. Thanks!

PR #5122, opened 2026-07-16
perf(context): skip Headers creation when there are no headers to merge

yusukebe:
This PR is one of the performance improvements.

This approach is the same as https://github.com/honojs/hono/pull/5112

Avoid creating a `Headers` when possible and pass the plain record to `Response` directly. This is fast on Bun, which consumes the record natively. On Node.js there is no difference since it builds `Headers` internally anyway. This affects only `body POST /json` on Bun.

Regarding [the problem of "Case-insensitive header overrides"](https://github.com/honojs/hono/pull/5112#issuecomment-4969693658) as @usualoma pointed out: records with a single entry cannot have case-insensitive key conflicts, so only those take the fast path. Anything else still goes through Headers.

### Bun

```
$ ROUNDS=5 ./compare.sh
round 1/5: main -> dev
round 2/5: dev -> main
round 3/5: main -> dev
round 4/5: dev -> main
round 5/5: main -> dev
• ping GET /
  main             median  290.70 ns  rounds: [290.60 ns, 296.60 ns, 290.67 ns, 290.70 ns, 295.50 ns]
  dev              median  293.41 ns  rounds: [293.41 ns, 298.21 ns, 291.92 ns, 273.94 ns, 300.55 ns]
  summary: dev is 1.01x slower than main (median of round avgs)

• query GET /id/1?name=bun
  main             median  759.86 ns  rounds: [757.80 ns, 763.35 ns, 759.86 ns, 756.65 ns, 760.02 ns]
  dev              median  744.66 ns  rounds: [772.45 ns, 729.72 ns, 737.35 ns, 744.66 ns, 745.54 ns]
  summary: dev is 1.02x faster than main (median of round avgs)

• body POST /json
  main             median    1.30 µs  rounds: [1.29 µs, 1.30 µs, 1.30 µs, 1.31 µs, 1.32 µs]
  dev              median    1.12 µs  rounds: [1.12 µs, 1.12 µs, 1.12 µs, 1.14 µs, 1.15 µs]
  summary: dev is 1.16x faster than main (median of round avgs)
```

### Node.js

```
$ ROUNDS=5 RUNTIME=node ./compare.sh
round 1/5: main -> dev
round 2/5: dev -> main
round 3/5: main -> dev
round 4/5: dev -> main
round 5/5: main -> dev
• ping GET /
  main             median  905.37 ns  rounds: [915.77 ns, 912.07 ns, 905.37 ns, 890.70 ns, 895.37 ns]
[...]

PR #5121, opened 2026-07-16
perf(url/request): replace regex tests with `indexOf`

yusukebe:
This PR is one of the performance improvements.

I measured that `indexOf('%')` is faster than `/\%/.test()`. So replaced them in param/query decoding. This affects `query GET /id/1?name=bun`.

### Bun

```
$ ROUNDS=5 ./compare.sh
round 1/5: main -> dev
round 2/5: dev -> main
round 3/5: main -> dev
round 4/5: dev -> main
round 5/5: main -> dev
• ping GET /
  main             median  291.99 ns  rounds: [291.99 ns, 298.62 ns, 282.07 ns, 292.98 ns, 282.14 ns]
  dev              median  297.78 ns  rounds: [291.81 ns, 304.84 ns, 297.78 ns, 294.95 ns, 297.89 ns]
  summary: dev is 1.02x slower than main (median of round avgs)

• query GET /id/1?name=bun
  main             median  754.64 ns  rounds: [736.16 ns, 757.47 ns, 757.85 ns, 754.64 ns, 746.27 ns]
  dev              median  713.22 ns  rounds: [699.01 ns, 713.22 ns, 711.30 ns, 733.74 ns, 725.09 ns]
  summary: dev is 1.06x faster than main (median of round avgs)

• body POST /json
  main             median    1.29 µs  rounds: [1.29 µs, 1.31 µs, 1.32 µs, 1.28 µs, 1.29 µs]
  dev              median    1.30 µs  rounds: [1.30 µs, 1.29 µs, 1.30 µs, 1.29 µs, 1.31 µs]
  summary: dev is 1.00x slower than main (median of round avgs)
```

### Node.js

```
$ ROUNDS=5 RUNTIME=node ./compare.sh
round 1/5: main -> dev
round 2/5: dev -> main
round 3/5: main -> dev
round 4/5: dev -> main
round 5/5: main -> dev
• ping GET /
  main             median  912.56 ns  rounds: [862.88 ns, 941.91 ns, 912.56 ns, 892.33 ns, 941.43 ns]
  dev              median  892.78 ns  rounds: [875.07 ns, 882.54 ns, 897.88 ns, 892.78 ns, 932.69 ns]
  summary: dev is 1.02x faster than main (median of round avgs)

• query GET /id/1?name=bun
  main             median    1.59 µs  rounds: [1.59 µs, 1.65 µs, 1.61 µs, 1.59 µs, 1.59 µs]
[...]

PR #5134, opened 2026-07-18
perf: Low-hanging optimizations

kibertoad:
This PR removes several per-request allocations from Hono's core request/response cycle, reducing `app.fetch()` overhead by **7-11%** across text, JSON, and param/query workloads. It also fixes a Windows path bug that made the
`benchmarks/fetch` harness unrunnable on Windows, and extends it with two new cases that isolate the paths touched here.

No public API or observable behavior changes. All unit tests, the Node runtime tests, and `tsc` pass unchanged.

## Benchmark results

Measured with the existing `benchmarks/fetch` harness (`compare.sh`), which runs each variant in a fresh process and reverses variant order every round to cancel JIT/GC warm-up bias. Values are medians of per-round averages over 7 rounds.

Environment: Node.js 24.15.0 (x64-win32), i7-13700HX. `RUNTIME=node ROUNDS=7 ./compare.sh`

| Case | `main` | this PR | Change |
| --- | ---: | ---: | --- |
| `ping GET /` (plain text) | 8.90 µs | 8.04 µs | **1.11x faster** |
| `query GET /id/1?name=bun` (param + query + header) | 12.68 µs | 11.82 µs | **1.07x faster** |
| `json GET /user` (`c.json()` response) | 10.42 µs | 9.51 µs | **1.10x faster** |
| `body POST /json` (`req.json()` + `c.json()`) | 33.43 µs | 30.74 µs | **1.09x faster** |
| `middleware GET /mw/hello` (compose path) | 8.69 µs | 8.80 µs | neutral (within noise) |

### 1. Fast path for `c.json()` and `c.html()` (`src/context.ts`)

`c.text()` already short-circuits to a bare `new Response(text)` when no headers, status, or init have been set. `c.json()` and `c.html()` had no such path: every call allocated a header-record object (`setDefaultContentType`), a `Headers` instance, an `Object.entries` array of `[key, value]` pairs, mutated the `Headers`, and then constructed the `Response` from it, even for the overwhelmingly common `return c.json(obj)` case with no customization.

They now use the same guard as `c.text()`:

```ts
!this.#preparedHeaders && !this.#status && !arg && !headers && !this.finalized
  ? new Response(body, { headers: { 'content-type': 'application/json' } })
  : /* existing slow path, unchanged */
```

Passing a plain-object headers init lets the runtime fill its internal header list directly, instead of Hono building and mutating an intermediate `Headers` wrapper first.

Safety: the guard is copied verbatim from the battle-tested `c.text()` fast path. `#res` can only be set after `#preparedHeaders` or `finalized` is set (via the `res` getter/setter or `c.header()`), so no code path can reach the fast path with pending response state. Header-name casing is irrelevant - `Headers` normalizes to lowercase in both paths.

### 2. Remove the throwaway `env = {}` field initializer (`src/context.ts`)

`Context` declared `env: E['Bindings'] = {}` as a field initializer, so **every request allocated an object that `#dispatch` immediately overwrote** with the real `env`. The initializer only ever mattered for `new Context(req)` without options; that case now gets its `{}` in the constructor's `else` branch instead. Semantics are identical, including `env` being `undefined` when `options.env` is `undefined` (as before).

This change benefits every request regardless of route or response type, which is why even the plain-text `ping` case improved by ~11%.

### 3. Lazy `#validatedData` (`src/request.ts`)

`HonoRequest` allocated `#validatedData = {}` in its constructor, paying for the validator feature on every request whether or not validators are used. It is now allocated on first `addValidatedData()` call; `valid()` reads through optional chaining and still returns `undefined` for unvalidated targets, exactly as before.

### 4. Allocation-free cache probe in `#cachedBody` (`src/request.ts`)

`Object.keys(bodyCache)[0]` allocated a fresh array on every body read (`req.json()`, `req.text()`, …) just to find the first cached key. Replaced with a `for ... in` loop which has same insertion-order semantics, but no allocation.
`bodyCache` is a plain object literal, so there are no enumerable prototype keys to worry about.

### 5. `indexOf` instead of regex in param decoding (`src/request.ts`)

`c.req.param()` used `/\%/.test(param)` to decide whether a value needs percent-decoding. `param.indexOf('%') !== -1` performs the same check without regex-engine overhead on typically short strings.

### 6. `Object.keys` instead of `Object.entries` in `#newResponse` (`src/context.ts`)

The slow-path header merge iterated `Object.entries(headers)`, allocating one pair-array per header. `Object.keys` allocates a single array and reads values by key, with the same own-enumerable iteration semantics.

## Benchmark harness changes (`benchmarks/fetch/bench.mts`)

- **Windows fix:** the default src path was computed with `URL.pathname` (`/C:/sources/hono/src`) and then passed through `pathToFileURL`, producing the broken `C:\C:\sources\hono\src`. Now uses `fileURLToPath`, which is correct on all platforms.
- **New case `json GET /user`:** isolates the `c.json()` response path.
  Previously JSON serialization was only measured inside `body POST /json`, mixed with request-body parsing.
- **New case `middleware GET /mw/hello`:** covers the multi-handler `compose` dispatch path, which none of the existing cases exercised.

- [ ] Add tests
- [X] Run tests

### DIFF
--- a/benchmarks/fetch/bench.mts
+++ b/benchmarks/fetch/bench.mts
@@ -37,6 +37,11 @@ const makeApp = async (src: string): Promise<any> => {
       c.header('x-powered-by', 'benchmark')
       return c.text(`${id} ${name}`)
     })
+    .get('/user', (c: any) => c.json({ id: 123, name: 'Alice', roles: ['admin', 'editor'] }))
+    .use('/mw/*', async (_c: any, next: any) => {
+      await next()
+    })
+    .get('/mw/hello', (c: any) => c.text('mw'))
   return app
 }
 
@@ -44,6 +49,8 @@ const app = await makeApp(src)
 
 const ping = new Request('http://localhost/')
 const query = new Request('http://localhost/id/1?name=bun')
+const user = new Request('http://localhost/user')
+const mw = new Request('http://localhost/mw/hello')
 
 let sink: unknown
 
@@ -51,6 +58,8 @@ let sink: unknown
 const cases: [string, (app: any) => Promise<unknown>][] = [
   ['ping GET /', (app) => app.fetch(ping)],
   ['query GET /id/1?name=bun', (app) => app.fetch(query)],
+  ['json GET /user', (app) => app.fetch(user)],
+  ['middleware GET /mw/hello', (app) => app.fetch(mw)],
   [
     'body POST /json',
     (app) =>
--- a/benchmarks/fetch/bench.mts
+++ b/benchmarks/fetch/bench.mts
@@ -15,9 +15,9 @@
 // Runs on both Bun and Node.
 import './ts-resolve.mjs'
 import { run, bench } from 'mitata'
-import { pathToFileURL } from 'node:url'
+import { fileURLToPath, pathToFileURL } from 'node:url'
 
-const src = process.env.HONO_SRC ?? new URL('../../src', import.meta.url).pathname
+const src = process.env.HONO_SRC ?? fileURLToPath(new URL('../../src', import.meta.url))
 const label = process.env.HONO_LABEL ?? 'hono'
 const asJson = process.env.HONO_JSON === '1'
 
--- a/src/context.ts
+++ b/src/context.ts
@@ -312,7 +312,7 @@ export class Context<
    * })
    * ```
    */
-  env: E['Bindings'] = {}
+  env: E['Bindings']
   #var: Map<unknown, unknown> | undefined
   finalized: boolean = false
   /**
@@ -357,6 +357,8 @@ export class Context<
       this.#notFoundHandler = options.notFoundHandler
       this.#path = options.path
       this.#matchResult = options.matchResult
+    } else {
+      this.env = {}
     }
   }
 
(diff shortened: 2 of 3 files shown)


yusukebe:
@kibertoad 

Thank you for the PR. There are great ideas. Before reviewing, we need to sort something out. I'll inform. Please wait.

yusukebe:
Hey @kibertoad ! Again, this PR is great. I have some things I want you to do.

We plan to include these "perf" improvements in the next minor version v4.13.0, and we've created [the next branch](https://github.com/honojs/hono/tree/next) for it.

Actually, I was working on similar optimizations at the same time. Three of your six improvements are already on next: *1. Fast path for c.json() and c.html()* is covered by #5122, *5. indexOf instead of regex* by #5121, and *6. Object.keys instead of Object.entries* by #5118.

The remaining three (2. env initializer, 3. Lazy #validatedData, 4. Allocation-free cache probe) are all good ideas and I'd love to take them. The benchmark changes are also meaningful. So could you split this PR as follows, with the base changed to next?

1. First, a PR with only the benchmark changes (the Windows fix and the new cases). I'll merge it right away — then you can measure the rest against next with the same harness.
2. Then PRs for 2, 3, and 4, based on next.

kibertoad:
@yusukebe Done! All four PRs are up. Please let me know if anything is missing.

yusukebe:
@kibertoad Thank you!! I'll check them.

yusukebe:
This PR is no longer needed. Thanks!

PR #5158, opened 2026-07-23
perf(urls): refactor `tryDecodeURIComponent`

yusukebe:
The bundle size decreased by `47B` compared to the `next` branch.

- [ ] Add tests
- [x] Run tests

### DIFF
--- a/src/utils/url.ts
+++ b/src/utils/url.ts
@@ -205,12 +205,15 @@ export const checkOptionalParameter = (path: string): string[] | null => {
   return results.filter((v, i, a) => a.indexOf(v) === i)
 }
 
+export const tryDecodeURIComponent = (str: string): string =>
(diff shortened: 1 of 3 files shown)


src/utils/url.ts
```diff
@@ -205,12 +205,15 @@ export const checkOptionalParameter = (path: string): string[] | null => {
   return results.filter((v, i, a) => a.indexOf(v) === i)
 }
 
+export const tryDecodeURIComponent = (str: string): string =>
+  str.indexOf('%') !== -1 ? tryDecode(str, decodeURIComponent_) : str
+
 // Optimized
-const _decodeURI = (value: string) => {
+const _decodeURI = (value: string): string => {
```
yusukebe, line 212: Added return type `string`.

yusukebe:
I think this is a good improvement. Not just refactoring, but reducing bundle size. No need to review. I'll merge right now.

PR #5161, opened 2026-07-24
fix: use `Object.create(null)` when parsing query, headers, and params

yusukebe:
A `__proto__` key in a request could change the prototype of the returned object. I use `Object.create(null)` to keep `__proto__` as a normal key. This is limited to that single object and does not touch `Object.prototype`, so it is not prototype pollution. This is a code-quality fix, not a security fix.

By the way, I am planning to create a utility function to do `Object.create(null)`  to reduce the bundle size.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/request.test.ts
+++ b/src/request.test.ts
@@ -236,6 +236,12 @@ describe('headers', () => {
     expect(req.header('Content-Type')).toBe('application/json')
     expect(req.header('ApiKey')).toBe('abc')
   })
+
+  test('req.header() is not affected by a `__proto__` header name', () => {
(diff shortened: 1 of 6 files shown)


yusukebe:
Hey @usualoma !

Can you review this? As I wrote in the description, this should not be a security fix.

usualoma:
@yusukebe
Thank you!
Yes, I also understand that there are currently no issues in terms of security, but just to be safe, I agree that we should use `Object.create(null)`!

yusukebe:
@usualoma Thank you! Merging.

PR #5164, opened 2026-07-25
fix(cookie): relax name validation when parsing Cookie header

### LINKED ISSUE #3189: Cookie not properly parsed. `validCookieNameRegEx` seems too strict.

yusukebe:
Fixes #3189

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/utils/cookie.test.ts
+++ b/src/utils/cookie.test.ts
@@ -80,6 +80,32 @@ describe('Parse cookie', () => {
     expect(cookie['']).toBeUndefined()
   })
 
+  it('Should parse cookies whose names are not RFC 6265 tokens', () => {
+    const cookieString =
+      'paraglide:lang=en; CognitoIdentityServiceProvider.abc.user@example.com.idToken=token; round(brackets)=ok; square[brackets]=ok; curly{brackets}=ok'
+    const cookie: Cookie = parse(cookieString)
+    expect(cookie['paraglide:lang']).toBe('en')
+    expect(cookie['CognitoIdentityServiceProvider.abc.user@example.com.idToken']).toBe('token')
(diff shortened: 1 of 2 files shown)


yusukebe:
Hey @BlankParticle 

Can you review this? As I wrote in #3189, I think this patch can be acceptable.

BlankParticle (approved):

yusukebe:
@BlankParticle Thanks!

PR #5166, opened 2026-07-26
fix(jsx): handle useSyncExternalStore subscription and snapshot changes

### LINKED ISSUE #5162: `useSyncExternalStore` doesn't resubscribe when `subscribe` function changes

4.12.31

chrome 150.0.7871.181

1. Clone repo https://github.com/pajecawav/hono-usesyncexternalstore-subscribe-repro
2. `pnpm install && pnpm dev` and open http://localhost:4173/
3. Click "Increment" 
4. Click "Reset"
5. New store is created but Hono is subscribed to the previous one so nothing happens

`useSyncExternalStore` accepts a `subscribe` function as a first argument to subscribe to the store change events, but doesn't resubscribe when the function changes. I believe the problem is that it uses `useEffect` without dependencies to subscribe to the store https://github.com/honojs/hono/blob/26d8e42bfff7eb78fc323a01e1d02d63b9dbcc72/src/jsx/hooks/index.ts#L415-L422

Hono demo:

https://github.com/user-attachments/assets/be7fa615-136b-4c77-8632-2de3007b27de

`useSyncExternalStore` resubscribes to the store when `subscribe` changes like react does

React demo from the same repo:

https://github.com/user-attachments/assets/955f8a6c-63bd-45b1-be4c-5114ad12de86

### What do you see instead?

_No response_

### Additional information

_No response_

usualoma:
fixes #5162

The reported issue can be addressed by making the subscription effect depend on `subscribe`. This PR goes slightly further and also handles changes to `getSnapshot`:

- Re-subscribe when `subscribe` changes.
- Keep the previous subscription active until the new subscription is installed during the effect flush.
- Read the snapshot during render so a changed `getSnapshot` is reflected without waiting for a store notification.
- Re-check the latest snapshot after subscribing to avoid missing an update between render and effect execution.
- Add regression tests for changing `subscribe` and `getSnapshot`.

While investigating this, I also explored fixing the underlying deferred effect and cleanup behavior. Since that would affect all effect hooks and needs broader review, it is intentionally kept out of this PR; see the draft PoC in #5165.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/jsx/hooks/dom.test.tsx
+++ b/src/jsx/hooks/dom.test.tsx
@@ -674,7 +674,7 @@ describe('Hooks', () => {
       let count = 0
       const unsubscribe = vi.fn()
       const subscribe = vi.fn(() => unsubscribe)
-      const getSnapshot = vi.fn(() => count++)
+      const getSnapshot = vi.fn(() => count)
       const SubApp = () => {
         const count = useSyncExternalStore(subscribe, getSnapshot)
         return <div>{count}</div>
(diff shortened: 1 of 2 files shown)


usualoma:
Hi @yusukebe,

yusukebe (approved):
LGTM!

yusukebe:
@usualoma 

Thank you!

PR #5171, opened 2026-07-30
feat(reg-exp-router): throw UnsupportedPathError during route registration

usualoma:
Unlike the other routers, `RegExpRouter` currently defers `UnsupportedPathError` until the matcher is built by the first `match()` call. This PR brings it in line with the other routers by reporting unsupported paths early, during route registration.

Although this changes when the error is thrown, the implementation is otherwise largely a restructuring of the existing matcher-building process. The generated matcher, routing behavior, steady-state matching path, and post-build memory cleanup remain unchanged.

`PreparedRegExpRouter` also continues to use the same matcher format and runtime behavior.

### Changes

- Throw `UnsupportedPathError` during route registration
- Preserve existing matching behavior and matcher representation
- Add regression tests for route ordering and method inheritance

### Performance

Local Bun benchmark results:

| Scenario | Before | After |
|---|---:|---:|
| Register 12 routes | 1.47 µs | 5.16 µs |
| Register 12 routes + first match | 17.77 µs | 14.45 µs |
| Register 500 routes + first match | 0.620 ms | 0.541 ms |
| Steady-state match | 45.3 ns | 46.0 ns |

Validation is moved from the first match to route registration. Registration alone is therefore slower, while registration plus matcher construction is faster and steady-state matching remains effectively unchanged.

<details>
<summary>Benchmark methodology and scripts</summary>

- Bun 1.3.12
- Before: `224d2f5cb`
- After: `18cde15bc`
- Measurements were run in clean worktrees on the same machine.
- Reported values are medians after warmup.

#### 12 routes and steady-state matching

```ts
import { RegExpRouter } from './src/router/reg-exp-router/router.ts'

const routes = [
  ['GET', '/user'],
  ['GET', '/user/comments'],
  ['GET', '/user/avatar'],
  ['GET', '/user/lookup/username/:username'],
  ['GET', '/user/lookup/email/:address'],
  ['GET', '/event/:id'],
  ['GET', '/event/:id/comments'],
  ['POST', '/event/:id/comment'],
  ['GET', '/map/:location/events'],
  ['GET', '/status'],
  ['GET', '/very/deeply/nested/route/hello/there'],
  ['GET', '/static/*'],
] as const

let sink = 0

const createRouter = (match = false) => {
  const router = new RegExpRouter<number>()
  for (let i = 0; i < routes.length; i++) {
    router.add(routes[i][0], routes[i][1], i)
  }
  if (match) {
    sink += router.match('GET', '/event/abcd1234/comments')[0].length
  }
  return router
}

const median = (values: number[]) =>
  values.sort((a, b) => a - b)[values.length >> 1]

const measure = (name: string, iterations: number, fn: () => void) => {
  for (let i = 0; i < Math.max(100, iterations / 20); i++) {
    fn()
  }

  const samples: number[] = []
  for (let sample = 0; sample < 9; sample++) {
    const start = Bun.nanoseconds()
    for (let i = 0; i < iterations; i++) {
      fn()
    }
    samples.push((Bun.nanoseconds() - start) / iterations)
  }

  console.log(`${name}\t${median(samples).toFixed(2)} ns/op`)
}

measure('add', 20_000, () => {
  sink += createRouter().name.length
})

measure('add+first-match', 20_000, () => {
  createRouter(true)
})

const router = createRouter(true)
measure('steady-match', 2_000_000, () => {
  sink += router.match('GET', '/event/abcd1234/comments')[0].length
})
```

#### 500 routes

```ts
import { RegExpRouter } from './src/router/reg-exp-router/router.ts'

const routes: [string, string][] = []
for (let i = 0; i < 250; i++) {
  routes.push(['GET', `/static-${i}`])
  routes.push(['GET', `/resource-${i}/:id`])
}

let sink = 0

const createRouter = (match = false) => {
  const router = new RegExpRouter<number>()
  for (let i = 0; i < routes.length; i++) {
    router.add(routes[i][0], routes[i][1], i)
  }
  if (match) {
    sink += router.match('GET', '/resource-249/value')[0].length
  }
  return router
}

const median = (values: number[]) =>
  values.sort((a, b) => a - b)[values.length >> 1]

  createRouter(true)
}

const samples: number[] = []
for (let sample = 0; sample < 7; sample++) {
  const start = Bun.nanoseconds()
  for (let i = 0; i < 250; i++) {
    createRouter(true)
  }
  samples.push((Bun.nanoseconds() - start) / 250)
}

console.log(`${(median(samples) / 1_000_000).toFixed(3)} ms/router`)
```

</details>

### Bundle size

Measured using the repository's minified bundle-size check (uncompressed).

| Before | After | Difference |
|---:|---:|---:|
| 19,032 B | 18,989 B | -43 B (-0.23%) |

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/router/reg-exp-router/node.ts
+++ b/src/router/reg-exp-router/node.ts
@@ -25,9 +25,9 @@ function compareKey(a: string, b: string): number {
     return 1
   }
 
-  // wildcard
+  // wildcard: the only wildcard (.*) precedes the tail wildcard
   if (a === ONLY_WILDCARD_REG_EXP_STR || a === TAIL_WILDCARD_REG_EXP_STR) {
-    return 1
+    return b === TAIL_WILDCARD_REG_EXP_STR ? -1 : 1
   } else if (b === ONLY_WILDCARD_REG_EXP_STR || b === TAIL_WILDCARD_REG_EXP_STR) {
     return -1
   }
--- a/src/router/reg-exp-router/node.ts
+++ b/src/router/reg-exp-router/node.ts
@@ -43,6 +43,7 @@ function compareKey(a: string, b: string): number {
 }
 
 export class Node {
+  // handler index of a dynamic path, or -1 for a static path terminal
   #index?: number
   #varIndex?: number
   #children: Record<string, Node> = Object.create(null)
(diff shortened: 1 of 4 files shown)


usualoma:
Hi @yusukebe,

Until now, RegExpRouter was designed to throw an `UnsupportedPathError` during the first `match()` call, which made it difficult to use with a single router; however, this PR addresses that issue.

What do you think?

yusukebe:
@usualoma 

The direction is good! I'll review the details later.

yusukebe (approved):
LGTM!

yusukebe:
@usualoma 

This is great! The behavior of throwing errors during the registration phase is ideal. And the performance and file size changes are interesting for me. I want to introduce this into the next minor release. To do this, it has to be merged into `next`. I'll merge this later, but there may be some conflicts; if so, please help me with it.

Anyway, thanks!

PR #5173, opened 2026-07-30
chore(benchmarks): correct src path on Windows, add json and middleware cases

kibertoad:
Split out of #5134 as requested in https://github.com/honojs/hono/pull/5134#issuecomment-5058932909. This is part 1 of 4: benchmark harness only, no `src/` changes. The three remaining optimizations (env initializer, lazy `#validatedData`, allocation-free cache probe) follow as separate PRs against `next`.

## Windows fix

The default src path was computed with `URL.pathname`:

```ts
const src = process.env.HONO_SRC ?? new URL('../../src', import.meta.url).pathname
```

On Windows that yields `/C:/sources/hono/src`, and passing it back through `pathToFileURL` produces `C:\C:\sources\hono\src`, so `benchmarks/fetch` could not run at all. `fileURLToPath` returns the correct native path on every platform.

## New cases

- `json GET /user` isolates the `c.json()` response path. Previously JSON serialization was only measured inside `body POST /json`, mixed together with request body parsing, so a change to either one moved the same number.
- `middleware GET /mw/hello` covers the multi-handler `compose` dispatch path, which none of the existing cases exercised.

Verified with `node bench.mts` on Node.js 24.16.0 (x64-win32); all five cases report.

- [ ] Add tests
- [X] Run tests

### DIFF
--- a/benchmarks/fetch/bench.mts
+++ b/benchmarks/fetch/bench.mts
@@ -15,9 +15,9 @@
 // Runs on both Bun and Node.
 import './ts-resolve.mjs'
 import { run, bench } from 'mitata'
-import { pathToFileURL } from 'node:url'
+import { fileURLToPath, pathToFileURL } from 'node:url'
 
-const src = process.env.HONO_SRC ?? new URL('../../src', import.meta.url).pathname
+const src = process.env.HONO_SRC ?? fileURLToPath(new URL('../../src', import.meta.url))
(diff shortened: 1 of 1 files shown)


yusukebe (approved):
LGTM!

yusukebe:
Hey @kibertoad 

Thank you! Looks good. I'll merge this into `next` immediately and review the rest.

Issue #5177, opened 2026-07-31
A function component returning an array fails with "str.search is not a function"


natsuki-engr:
### What version of Hono are you using?

4.12.32

### What runtime/platform is your app running on? (with version if possible)

Node.js v24.13.0, also reproduced on Bun 1.3.12

### What steps can reproduce the bug?

A function component that returns an array fails to render.

```tsx
import { Hono } from 'hono'

const app = new Hono()

const Items = () => ['a', 'b'].map((x) => <li>{x}</li>)

app.get('/', (c) => c.html(<Items />))

export default app
```

### What is the expected behavior?

Either of these, depending on which direction you prefer:

1. Returning an array stays unsupported, but the failure says so, instead of surfacing `str.search` from an internal helper. At minimum the message should name what to change.
2. `<li>a</li><li>b</li>` is rendered, the same as wrapping the elements in a Fragment. `hono/jsx/dom` already renders the array-returning version this way, so today the two renderers disagree on what a component may return.

### What do you see instead?

The request responds 500, and the terminal shows a message that names an internal helper:

```
TypeError: str.search is not a function
```

`JSXFunctionNode.toStringToBuffer()` in `src/jsx/base.ts` branches on the component's return value: boolean, `null`, `Promise`, `JSXNode`, number, escaped string. An array matches none of them, so it falls through to the final `else`, which calls `escapeToBuffer()` in `src/utils/html.ts`. That helper starts with `str.search(escapeRe)`, and an array has no such method.

### Additional information

TypeScript rejects the same code with `TS2786: 'Items' cannot be used as a JSX component. Its return type 'Element[]' is not a valid JSX element`, so this only reaches the runtime when types are not checked. React has allowed components to return arrays since v16, so the pattern is easy to arrive at when porting.

I have implemented both directions and will open a PR for each, so they can be compared side by side. They are alternatives rather than a pair: only one should be merged, and I will close the other.

**1. Keeping it unsupported.** A single branch in `JSXFunctionNode.toStringToBuffer()` throws `A function component must not return an array. Wrap the elements in a Fragment: <>...</>`. No type changes are needed, because the type layer already rejects the code. This leaves `hono/jsx/dom` and the string renderer disagreeing, now with an explicit message rather than a confusing one.

[...]

PR #5178, opened 2026-07-31
fix(jsx): report a clear error when a component returns an array

natsuki-engr:
Refs #5177.

This is direction 1 in the issue: returning an array stays unsupported, and the failure says so.

The alternative, supporting the array return, is #5179. **The two are alternatives, not a pair.** Only one should be merged, and I will close the other.

## Problem

A function component that returns an array reaches `escapeToBuffer()` in `src/utils/html.ts`, which starts with `str.search(escapeRe)`. The request responds 500 with `TypeError: str.search is not a function`, naming an internal helper rather than the code that has to change.

## Solution

`JSXFunctionNode.toStringToBuffer()` gets a branch for arrays, which throws instead:

```
A function component must not return an array. Wrap the elements in a Fragment: <>...</>
```

No type changes are needed. The type layer already rejects the code with `TS2786`, so this only makes the runtime agree with what the types already say.

`hono/jsx/dom` renders an array-returning component correctly, and this PR leaves that difference in place. Removing it is what the other direction does.

## Tests

`src/jsx/index.test.tsx` asserts the thrown message. The test needs `@ts-expect-error` on the JSX expression, which doubles as a check that the type layer still rejects it.

### DIFF
--- a/src/jsx/base.ts
+++ b/src/jsx/base.ts
@@ -289,6 +289,12 @@ class JSXFunctionNode extends JSXNode {
         buffer.callbacks ||= []
         buffer.callbacks.push(...res.callbacks)
       }
+    } else if (Array.isArray(res)) {
+      // Reaching `escapeToBuffer()` with an array fails with an unrelated message,
+      // so report what the caller actually has to change.
+      throw new Error(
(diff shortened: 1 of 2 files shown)


yusukebe:
We chose https://github.com/honojs/hono/pull/5179. Closing.

PR #5179, opened 2026-07-31
fix(jsx): allow a function component to return an array

natsuki-engr:
Refs #5177.

This is direction 2 in the issue: a function component may return an array, as `hono/jsx/dom` already allows.

The alternative, keeping it unsupported behind an explicit error, is #5178. **The two are alternatives, not a pair.** Only one should be merged, and I will close the other.

## Solution

**1. `fix(jsx): allow a function component to return an array`**

Handle arrays in `JSXFunctionNode.toStringToBuffer()` with `childrenToStringToBuffer()`, the path already used for array children.

**2. `fix(jsx): accept an array return in the type layer`**

The runtime fix alone is unreachable from TypeScript, which reports `TS2786: ... Its return type 'Element[]' is not a valid JSX element`. `Child[]` is added to the call signature of `FC`, and `JSX.ElementType` is defined with the same return type. Without `JSX.ElementType`, TypeScript checks the return type against `JSX.Element`; with it, it checks the tag instead.

Both types are given the same return type, so nothing beyond arrays becomes newly valid:

| | before | after |
| --- | --- | --- |
| `() => [<a />, <b />]` | error | **ok** |
| `() => 42`, `() => 'x'`, `() => undefined` | error | error |

**3. `fix(jsx): render an array resolved from a promise`**

A component may be async, so the array can arrive through a promise. Without this the feature would only be half there, since `async () => [<li>a</li>]` would still be rejected. It renders through `toString()` and through `Suspense` / `renderToReadableStream()` alike.

`Child` therefore carries `Promise<string \| Child[]>`, and a resolved array is wrapped in a fragment so it renders like any other node. Widening `Child` surfaced the same path for children, so both call sites share `resolveArrayToFragment()`. The helper also attaches the suspended context, which let the promise branch drop its own `then()` chain.

## A note on the cast in `resolveArrayToFragment()`

The helper returns `Promise<string>` but resolves to a node, so it ends with `as unknown as string`. The buffer already held nodes before this PR:

```ts
res.then((childRes) => {
  if (childRes instanceof JSXNode) { // the buffer already received nodes here
    childRes.suspendedContext = suspendedContext
  }
  return childRes
})
```

That branch type-checked only because `res` was `any`, and `stringBufferToString()` is written for it too, calling `toString()` on an object entry. What is inaccurate is the declaration, `StringBuffer = (string | Promise<string>)[]`.

Typing it honestly is a separate change. It needs `stringBufferToString()` to settle how it treats `String` objects versus primitives, and it exports a new type from `hono/utils/html`, which is used outside JSX. It is left out here on purpose, and the cast is kept in one place with a comment instead of spread across the three call sites.

### DIFF
--- a/src/jsx/base.ts
+++ b/src/jsx/base.ts
@@ -26,7 +26,7 @@ import {
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type Props = Record<string, any>
 export type FC<P = Props> = {
-  (props: P): HtmlEscapedString | Promise<HtmlEscapedString> | Child[] | null
+  (props: P): HtmlEscapedString | Child[] | Promise<HtmlEscapedString | Child[]> | null
   defaultProps?: Partial<P> | undefined
   displayName?: string | undefined
 }
@@ -39,7 +39,7 @@ export namespace JSX {
   // returning an array is accepted as a tag.
   export type ElementType =
     | string
-    | ((props: never) => HtmlEscapedString | Promise<HtmlEscapedString> | Child[] | null)
+    | ((props: never) => HtmlEscapedString | Child[] | Promise<HtmlEscapedString | Child[]> | null)
   export interface ElementChildrenAttribute {
     children: Child
   }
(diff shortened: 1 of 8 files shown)


usualoma:
Hi @natsuki-engr,
Thank you for opening the issue and putting together this PR.

Between the two proposed approaches, I think supporting array-returning function components in this PR is preferable to keeping them unsupported with a more descriptive error.

However, extending the general `Child` type to support `Promise<Child[]>` introduces a much broader set of consistency concerns around contexts, callbacks, streaming, metadata hoisting, and related APIs.

I’ve opened a follow-up PR against this branch that keeps async array support scoped specifically to function component results and addresses the issues found during additional testing: https://github.com/natsuki-engr/hono/pull/3.

Could you take a look and, if the changes look good to you, merge that follow-up into this branch? With those changes included, I think this PR would be ready to merge into Hono.

yusukebe:
@natsuki-engr Thanks!

@usualoma Can you review this again? If you approve, I will merge.

natsuki-engr:
Thanks! On the current state of this branch: @usualoma sent a follow-up PR to my fork (natsuki-engr/hono#3), and I have one open question there about `JSXNode.suspendedContext`. Once that is settled I will merge it into this branch, so the diff in this PR will be updated then.

yusukebe:
@natsuki-engr Okay!

natsuki-engr:
@yusukebe @usualoma Merged usualoma's follow-up into this branch. Ready for another look.

usualoma:
LGTM!

yusukebe (approved):

yusukebe:
@natsuki-engr @usualoma 

Thanks! I won't merge this immediately, but will do it later.

PR #5180, opened 2026-07-31
fix(stream): re-acquire writer after pipe() failure

pacocartones:
## What

`StreamingApi.pipe()` releases the writer lock, awaits `body.pipeTo()`, then re-acquires the writer — but with no `try/finally`:

```ts
  this.writer.releaseLock()
  await body.pipeTo(this.writable, { preventClose: true })
  this.writer = this.writable.getWriter()   // skipped if pipeTo() rejects
}
```

If `pipeTo()` rejects (upstream stream error, abort / client disconnect), the re-acquire line is skipped and `this.writer` is left pointing at a **released** writer for the lifetime of the instance. Subsequent `write()`/`close()` then run against that released writer and throw `Writer is not bound to a WritableStream`. Because `write()`/`close()` swallow errors by design, the throw is hidden — but the `StreamingApi` is left in an inconsistent state (its `writer` no longer bound to its `writable`).

Refs #4981.

## Change

Wrap `pipeTo()` in `try/finally` so the writer is always re-acquired, however the pipe terminates:

```ts
  this.writer.releaseLock()
  try {
    await body.pipeTo(this.writable, { preventClose: true })
  } finally {
    this.writer = this.writable.getWriter()
  }
}
```

**Scope / honest note:** this restores the `writer`↔`writable` invariant; it does **not**, on its own, make the destination usable again after a rejected pipe. With the default `preventAbort: false`, a rejecting `pipeTo()` aborts the destination writable, so the stream is torn down regardless — the value here is instance-state consistency (no lingering released writer / `TypeError`), not resurrecting the aborted stream. Making the destination reusable across a failed pipe would need `preventAbort: true`, which is a separate behavior change I've deliberately left out to keep this diff minimal — happy to follow up if you'd want that.

## Test

Added a regression test in `src/utils/stream.test.ts`: pipe an erroring source, assert `pipe()` rejects, then assert the writer is re-bound (writable re-locked) and `write()`/`close()` stay safe. The existing success-path `pipe()` test and the `helper/streaming` abort-signal test continue to pass (10/10 in `stream.test.ts`). `prettier --check` and `eslint` clean.

---
_AI-assisted; reviewed and verified by the author._


yusukebe:
This was closed https://github.com/honojs/hono/pull/4988

PR #5185, opened 2026-08-03
perf(context): set headers on a finalized response without recreating it

kibertoad:
`c.header()` on a finalized response (the "set headers after `await next()`" middleware pattern) recreated the entire `Response` on every call. CPU profiling on Node shows `new Response()` is the most expensive operation in the request path (ReadableStream setup plus TextEncoder account for roughly 25% of in-process request time), so each post-`next()` header set paid for a full extra Response construction.

The recreation only exists because some responses have immutable headers (e.g. one returned by `fetch()`, whose headers undici marks immutable). This change mutates the headers in place and falls back to recreating the Response only when the mutation throws.

Measured with a new `post-mw headers GET /hdr/hello` case (middleware that sets two headers after `next()`) added to `benchmarks/fetch`, two independent 5-round `compare.sh` runs on Node:

- 5.04 µs -> 3.45 µs (1.46x faster)
- 5.42 µs -> 3.60 µs (1.51x faster)

All other cases stayed within the noise band.

Notes on behavior:

- When the finalized response's headers are mutable (any Response constructed by hono or by handler code), `c.res` now keeps its object identity across `c.header()` calls instead of being replaced.
- The fallback path was verified against a real `fetch()`ed upstream response, and a regression test uses `Response.redirect()`, which has the same immutable-headers guard without needing a network.
- `Response.error()` behaves exactly as before (recreation is attempted and throws, since status 0 cannot be passed to the Response constructor).

- [x] Add tests
- [x] Run tests

### DIFF
--- a/benchmarks/fetch/bench.mts
+++ b/benchmarks/fetch/bench.mts
@@ -15,9 +15,9 @@
 // Runs on both Bun and Node.
 import './ts-resolve.mjs'
 import { run, bench } from 'mitata'
-import { pathToFileURL } from 'node:url'
+import { fileURLToPath, pathToFileURL } from 'node:url'
 
-const src = process.env.HONO_SRC ?? new URL('../../src', import.meta.url).pathname
+const src = process.env.HONO_SRC ?? fileURLToPath(new URL('../../src', import.meta.url))
 const label = process.env.HONO_LABEL ?? 'hono'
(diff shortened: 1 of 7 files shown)


yusukebe:
@kibertoad 

Thanks! Let's consider merging the PRs after releasing the next minor version.

PR #5189, opened 2026-08-04
fix(trie-router): count every slash a pattern consumes

Jaybhade:
When a `:param{regexp}` match spans several path parts, `Node#search` requeues the child into `curNodesQueue[componentCount]` so it resumes at the part *after* the match. `componentCount` came from:

```js
const componentCount = m[0].match(/\//)?.length ?? 0
```

Without the `g` flag, `String.prototype.match` returns a single-match array, so `.length` is `1` for **any** number of slashes. A pattern spanning three or more parts resumed one part too early and matched a part it had already consumed.

Two parts happen to give the right answer (one slash → `1`), which is why the existing `/foo/bar/file.html` case passes and this went unnoticed.

## Symptoms

Both reproduce on a plain `new Hono()` — `RegExpRouter` reports `UnsupportedPathError` for these routes, so `SmartRouter` falls through to `TrieRouter`.

**A route stops matching once the pattern spans 3+ parts.** `/:dirs{.+}/file.html` is the repo's own `Capture simple multiple directories` route:

| path | before | after |
| --- | --- | --- |
| `/foo/file.html` | 200 | 200 |
| `/foo/bar/file.html` | 200 | 200 |
| `/foo/bar/baz/file.html` | **404** | 200 |
| `/a/b/c/d/file.html` | **404** | 200 |

**A handler is returned twice, so middleware runs twice per request:**

```ts
const app = new Hono()
let runs = 0
app.use('/:path{.*}/*', async (c, next) => { runs++; await next() })
app.get('/a/b/c', (c) => c.text('ok'))

await app.request('/a/b/c')
// before: runs === 2
// after:  runs === 1
```

At the raw router level:

```ts
const r = new TrieRouter()
r.add('GET', '/:file{.*}/*', 'handler')
r.match('GET', '/a/b/c')
// before: [[['handler', {file:'a/b/c'}], ['handler', {file:'a/b/c'}]]]
// after:  [[['handler', {file:'a/b/c'}]]]
```

`RegExpRouter` and `PatternRouter` already return the correct result in every case above; this brings `TrieRouter` in line.

## Fix

Count with a global regexp so the offset reflects every slash the pattern consumed.

## Tests

- `common.case.test.ts` — extended `Capture simple multiple directories` with 3- and 4-directory paths. It runs for every router, so it pins the behaviour across all of them; all pass unchanged, no new skip entries needed.
- `trie-router/node.test.ts` — added `Pattern spanning multiple parts`, covering both the duplicate handler and the missed match at increasing depth.

Without the one-line change these fail with `expected 2 to be 1` and `expected +0 to be 1`.

I also fuzzed ~197k random route sets across all four routers looking for any handler returned twice from a single `match()`: 16 distinct reproductions before, 0 after.

- [x] Add tests
- [x] Run tests — `bun run test`: 146 files, 4758 passed / 35 skipped
- [x] `bun run format:fix && bun run lint:fix` to format the code — prettier clean, 0 lint errors
- [ ] Add [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code — n/a, internal branch in `Node#search`; left an inline comment on the counting logic instead

### DIFF
--- a/src/router/common.case.test.ts
+++ b/src/router/common.case.test.ts
@@ -535,6 +535,20 @@ export const runTest = ({
         expect(res[0].handler).toEqual('file.html')
         expect(res[0].params['dirs']).toEqual('foo/bar')
       })
+
+      it('GET /foo/bar/baz/file.html', () => {
+        const res = match('GET', '/foo/bar/baz/file.html')
+        expect(res.length).toBe(1)
+        expect(res[0].handler).toEqual('file.html')
+        expect(res[0].params['dirs']).toEqual('foo/bar/baz')
+      })
+
+      it('GET /a/b/c/d/file.html', () => {
+        const res = match('GET', '/a/b/c/d/file.html')
+        expect(res.length).toBe(1)
+        expect(res[0].handler).toEqual('file.html')
+        expect(res[0].params['dirs']).toEqual('a/b/c/d')
+      })
     })
 
     describe('Capture regex pattern has trailing wildcard', () => {
--- a/src/router/trie-router/node.test.ts
+++ b/src/router/trie-router/node.test.ts
@@ -812,6 +812,35 @@ describe('The same name is used for path params', () => {
   })
 })
 
+describe('Pattern spanning multiple parts', () => {
+  describe('followed by a wildcard', () => {
+    const node = new Node()
+    node.insert('get', '/:file{.*}/*', 'file')
+
+    it('should return the handler once per matching part count', () => {
+      for (const path of ['/a', '/a/b', '/a/b/c', '/a/b/c/d']) {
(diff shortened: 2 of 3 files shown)


usualoma:
Hi @Jaybhade,
This is just a simple bug. Thanks for fixing it!

Since Hono doesn't include code comments for this level of processing, I think you should just delete that part.

Jaybhade:
Done in 41d9b67 — comment removed, the change is just the `g` flag now.

yusukebe (approved):
LGTM!

yusukebe:
@Jaybhade Thank you!

PR #5192, opened 2026-08-05
feat(etag): add cacheableStatusCodes option

### LINKED ISSUE #5191: ETag middleware should not generate ETag on 4xx or 5xx

Current ETag middleware always generates `ETag` and returns 304 if `If-None-Match` matches, regardless of the response status.
If a route returns the same response body (like `{}`) for different statuses, and the client cached the response and ETag, the client cannot distinguish between the response statuses.

So I want ETag middleware to generate etag only for 200 response,
and to maintain backward compatibility, I propose adding `cacheableStatusCode: number[] | undefined` option like cache middleware.
(The default would be all status codes)

na-trium-144:
close #5191

I copied the snippet from the cache middleware.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/etag/index.ts
+++ b/src/middleware/etag/index.ts
@@ -4,12 +4,14 @@
  */
 
 import type { MiddlewareHandler } from '../../types'
+import type { StatusCode } from '../../utils/http-status'
 import { generateDigest } from './digest'
 
 type ETagOptions = {
   retainedHeaders?: string[]
   weak?: boolean
   generateDigest?: (body: Uint8Array<ArrayBuffer>) => ArrayBuffer | Promise<ArrayBuffer>
+  cacheableStatusCodes?: StatusCode[] | undefined
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/middleware/etag/index.ts
```diff
@@ -4,12 +4,14 @@
...
 
 import type { MiddlewareHandler } from '../../types'
+import type { StatusCode } from '../../utils/http-status'
 import { generateDigest } from './digest'
 
 type ETagOptions = {
   retainedHeaders?: string[]
   weak?: boolean
   generateDigest?: (body: Uint8Array<ArrayBuffer>) => ArrayBuffer | Promise<ArrayBuffer>
+  cacheableStatusCodes?: StatusCode[] | undefined
```
yusukebe, line 14: You can remove the `undefined` and it should be:

```ts
cacheableStatusCodes?: StatusCode[]
```

PR #5196, opened 2026-08-06
fix(etag): skip unsafe methods or error responses on non-* case

### LINKED ISSUE #5191: ETag middleware should not generate ETag on 4xx or 5xx

na-trium-144:
fix #5191

Previously, it checked request methods and response status only when if-none-match is `*`.
According to spec, it should not return 304 on unsafe methods or error response regardless of if-none-match value.
see https://github.com/honojs/hono/issues/5191#issuecomment-5205914907

If this gets merged, I think #5192 is not needed.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/etag/index.test.ts
+++ b/src/middleware/etag/index.test.ts
@@ -273,13 +273,13 @@ describe('Etag Middleware', () => {
     expect(res.status).toBe(304)
   })
 
-  it('Should not return 304 for `If-None-Match: *` on unsafe methods or error responses', async () => {
+  it('Should not return 304 on unsafe methods or error responses', async () => {
     const app = new Hono()
     app.use('/etag/*', etag())
-    app.put('/etag/put', (c) => c.text('created', 201))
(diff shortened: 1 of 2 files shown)


usualoma:
Hi @na-trium-144,
I see. That does seem like the right approach.

yusukebe (approved):
LGTM!

yusukebe:
@na-trium-144 

This is right. The behavior was like a bug that should be fixed. Thanks!

ababove20-jpg (approved):

ababove20-jpg (approved):

ababove20-jpg (approved):

PR #5197, opened 2026-08-07
 fix(secure-headers): output standard empty parentheses () instead of none for disabled Permissions-Policy directives

spellsaif:
Historically, Hono's `secureHeaders` middleware serialized disabled or `none` value Permissions-Policy directives as `none` (e.g.,
  `camera=none`). However, according to the modern W3C Structured Fields specification for the `Permissions-Policy` header, the
  keyword `none` is not a valid value. Instead, the correct syntax to disable a feature for all origins is an empty structured list
  represented by empty parentheses `()`.

This PR updates the serialization in `getPermissionsPolicyDirectives` to correctly output `()` instead of `none` when a directive
  value is `false` or `['none']`, ensuring strict compliance with the standard and preventing modern browsers from failing to parse
  the header.

    - [x] Add tests
    - [x] Run tests

### DIFF
--- a/src/middleware/secure-headers/index.test.ts
+++ b/src/middleware/secure-headers/index.test.ts
@@ -187,7 +187,7 @@ describe('Secure Headers Middleware', () => {
 
     const res = await app.request('/test')
     expect(res.headers.get('Permissions-Policy')).toEqual(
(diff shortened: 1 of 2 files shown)


yusukebe (approved):

spellsaif:
Thanks for reviewing and merging this, @yusukebe! Glad we could get the Permissions-Policy serialization aligned with the current specification.

PR #5198, opened 2026-08-07
fix(cookie): return false for malformed or unsigned cookies when parsing signed cookie by name

spellsaif:
When calling `getSignedCookie(c, secret, 'cookie_name')`, Hono is expected to return the verified cookie value, `undefined` if the cookie is not present, or `false` if verification fails.

However, if the cookie exists in the request but has a malformed signature (e.g., incorrect signature length or does not contain a signature dot delimiter), the helper silently skipped it and returned `undefined` instead of `false`. This allows clients to bypass verification logging or custom checks (which inspect for `=== false`) simply by sending a malformed signature length.

This PR updates `parseSigned` to ensure that when a specific cookie key is requested by name, any structural signature
  verification failure returns `false` explicitly instead of falling back to `undefined`.

    - [x] Add tests
    - [x] Run tests

### DIFF
--- a/src/utils/cookie.test.ts
+++ b/src/utils/cookie.test.ts
@@ -232,6 +232,20 @@ describe('Parse cookie', () => {
     expect(cookie['tasty_cookie']).toBe(false)
   })
 
+  it('Should parse one signed cookie specified by name and return "false" if cookie has no signature', async () => {
+    const secret = 'secret ingredient'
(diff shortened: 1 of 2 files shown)


yusukebe:
This is intended. `false` means a signature failed verification; `undefined` means the value isn't a signed cookie. Closing.

PR #5199, opened 2026-08-07
 fix(etag): resolve incorrect incremental hashing for chunked responses

### LINKED ISSUE #4401: `etag` middleware might produce different ETags for same content because of incremental hashing

4.9.6

Node.js 24.7.0

```js
import { Hono } from 'hono';
import { etag } from 'hono/etag';
import { stream } from 'hono/streaming';

const app = new Hono;
app
    .get('/test1', etag(), c => stream(c, async (s) => {
        for (let i = 0; i < 4; i++) {
            await s.write(new Uint8Array(1024));
        }
    }))
    .get('/test2', etag(), async c => c.body(new Uint8Array(4096)));

serve(({
    fetch: app.fetch,
    hostname: 'localhost',
    port: 4000,
}));
```

Send request to `/test1` and `/test2` and check the `etag` response header.

The ETags of the two endpoints should be the same because the response body is the same.

### What do you see instead?

The ETags are not same.

```
$ curl -Ss -v --output - http://localhost:4000/test1 | b3sum -
* Host localhost:4000 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:4000...
* Connected to localhost (::1) port 4000
* using HTTP/1.x
> GET /test1 HTTP/1.1
> Host: localhost:4000
> User-Agent: curl/8.15.0
> Accept: */*
>
< HTTP/1.1 200 OK
< etag: "cbcc719088aaae76d4a55abba3fe5a0fb440f498"
< content-type: text/plain; charset=UTF-8
< Date: Sat, 06 Sep 2025 04:41:30 GMT
< Connection: keep-alive
< Keep-Alive: timeout=5
< Transfer-Encoding: chunked
<
{ [4129 bytes data]
* Connection #0 to host localhost left intact
b6fb73fc46938c981e2b0b4b1ef282adcfc89854d01bfe3972fdc4785b41b2c7 *-

$ curl -Ss -v --output - http://localhost:4000/test2 | b3sum -
* Host localhost:4000 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:4000...
* Connected to localhost (::1) port 4000
* using HTTP/1.x
> GET /test2 HTTP/1.1
> Host: localhost:4000
> User-Agent: curl/8.15.0
> Accept: */*
>
< HTTP/1.1 200 OK
< etag: "1ceaf73df40e531df3bfb26b4fb7cd95fb7bff1d"
< content-type: text/plain; charset=UTF-8
< content-length: 4096
< Date: Sat, 06 Sep 2025 04:41:34 GMT
< Connection: keep-alive
< Keep-Alive: timeout=5
<
{ [4096 bytes data]
* Connection #0 to host localhost left intact
b6fb73fc46938c981e2b0b4b1ef282adcfc89854d01bfe3972fdc4785b41b2c7 *-
```

### Additional information

https://github.com/honojs/hono/blob/2f489b3562cd0d29075062b2ea0648ab92b88727/src/middleware/etag/digest.ts#L26-L34

The `etag` middleware uses incremental hashing. When a chunk of response is read, the middleware uses `generator` to calculate a new etag `hash(etag + chunk)`. However, if the same response body is divided into chunks in different ways, the resulting ETag will be different.

Is it necessary to use some buffering to ensure each chunk passed into `generator` has a fixed size? Or provide options for `init`, `update`, and `finalize` functions similar to many existing implementations of hash algorithms?

#4070 also mentioned this issue.

spellsaif:
Resolves #4401. Hono's `etag` middleware previously generated ETags for streams by hashing chunks incrementally using `result = generator(mergeBuffers(result, value))`. This computes `hash(hash(chunk1) + chunk2)` instead of `hash(chunk1 + chunk2)`.
Consequently, the generated ETag was mathematically incorrect and depended on how the response stream chunks were split across the network.

This PR fixes the digest generation logic in `generateDigest` to accumulate all stream chunks in memory first and perform a single SHA-1 hash on the full payload, ensuring mathematical correctness and consistency across chunk sizes. Additionally, the middleware now skips ETag generation for `text/event-stream` (SSE) responses to prevent infinite reader blocking and hangs.

 ###  Design Decisions & Rationale

* **Buffer-and-Hash for Correctness:** Native Web Crypto (`crypto.subtle.digest`) does not support streaming/incremental hashing. To calculate the correct SHA-1 hash of the response body, we must collect the stream chunks in memory and merge them. This ensures that identical response payloads always yield the same ETag, regardless of network chunk fragmentation or compression differences.

* **Targeted SSE Bypass:** Server-Sent Events (SSE) are infinite streams. Trying to read them to the end blocks the reader indefinitely. We added a zero-overhead check for the `Content-Type: text/event-stream` header to bypass the ETag pipeline safely.

* **Memory Trade-off:** While buffering chunks consumes memory proportional to the response body, the ETag middleware already clones and reads the entire body stream to hash it. Consolidating the chunks in memory is a minor overhead that is necessary to achieve cache specification compliance.

    - [x] Add tests
    - [x] Run tests

### DIFF
--- a/src/middleware/etag/digest.ts
+++ b/src/middleware/etag/digest.ts
@@ -1,18 +1,3 @@
-const mergeBuffers = (
-  buffer1: ArrayBuffer | undefined,
-  buffer2: Uint8Array<ArrayBuffer>
-): Uint8Array<ArrayBuffer> => {
-  if (!buffer1) {
-    return buffer2
-  }
-  const merged = new Uint8Array<ArrayBuffer>(
-    new ArrayBuffer(buffer1.byteLength + buffer2.byteLength)
-  )
-  merged.set(new Uint8Array(buffer1), 0)
-  merged.set(buffer2, buffer1.byteLength)
-  return merged
-}
-
 export const generateDigest = async (
   stream: ReadableStream<Uint8Array<ArrayBuffer>> | null,
   generator: (body: Uint8Array<ArrayBuffer>) => ArrayBuffer | Promise<ArrayBuffer>
@@ -21,7 +6,8 @@ export const generateDigest = async (
     return null
   }
 
-  let result: ArrayBuffer | undefined = undefined
+  const chunks: Uint8Array<ArrayBuffer>[] = []
+  let totalLength = 0
 
   const reader = stream.getReader()
   for (;;) {
@@ -29,13 +15,22 @@ export const generateDigest = async (
     if (done) {
       break
     }
-
-    result = await generator(mergeBuffers(result, value))
+    chunks.push(value)
+    totalLength += value.byteLength
   }
 
-  if (!result) {
+  if (chunks.length === 0) {
     return null
   }
+
+  const merged = new Uint8Array<ArrayBuffer>(new ArrayBuffer(totalLength))
+  let offset = 0
+  for (const chunk of chunks) {
+    merged.set(chunk, offset)
+    offset += chunk.byteLength
+  }
+
+  const result = await generator(merged)
 
   return Array.prototype.map
     .call(new Uint8Array(result), (x) => x.toString(16).padStart(2, '0'))
--- a/src/middleware/etag/index.test.ts
+++ b/src/middleware/etag/index.test.ts
@@ -388,18 +388,6 @@ describe('Etag Middleware', () => {
     expect(res.headers.get('ETag')).toBe('"d104fafdb380655dab607c9bddc4d4982037afa1"')
   })
 
-  it('Should skip ETag generation for text/event-stream (SSE)', async () => {
-    const app = new Hono()
-    app.use('/etag/*', etag())
-    app.get('/etag/sse', (c) => {
-      return c.text('data: hello\n\n', 200, {
-        'Content-Type': 'text/event-stream',
-      })
-    })
-    const res = await app.request('http://localhost/etag/sse')
-    expect(res.headers.get('ETag')).toBeNull()
-  })
-
   describe('When crypto is not available', () => {
     let _crypto: Crypto | undefined
     beforeAll(() => {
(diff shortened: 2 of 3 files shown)


yusukebe:
Skipping SSE is not needed, since streaming responses are not limited to SSE. The user must handle the issue themselves.

yusukebe (approved):

usualoma:
Hi @yusukebe,

Could we revisit this change?

My understanding is that ETags depending on stream chunk boundaries has so far been a known limitation. The existing chunk-by-chunk implementation was also intentional: it avoided retaining every chunk on the hashing side and allocating an additional full-size contiguous buffer.

I understand that `Response.clone()` uses teeing and that unread chunks can already be retained by the original branch. However, that does not make the additional `chunks` and `merged` buffers free. This change can significantly increase peak memory usage, potentially by multiple times the response size depending on the runtime, which is particularly concerning for large or concurrent streaming responses.

If buffering the entire body is acceptable, we could simply use `await res.clone().arrayBuffer()` from the beginning. Manually consuming the stream and then merging all chunks provides no memory advantage in that case.

Also, an ETag is an opaque validator; it is not required to be the standard SHA-1 digest of the concatenated response body. If making ETags independent of chunk boundaries is now a requirement, I think we should use a different approach, such as a genuinely incremental hashing implementation, rather than introducing unbounded buffering.

Given this trade-off, are we sure we want to keep this change merged?

usualoma:
If generating the same ETag regardless of stream chunk boundaries is a requirement, I think we should use a bounded-memory approach like #5205 instead of buffering the entire response.

yusukebe:

As you said, I must not merge this. But I think the streaming chunk inconsistency that breaks the ETag should be fixed. Thank you for #5202, I'll check it.

spellsaif:
@usualoma Thanks for the review and for pointing this out. I understand the memory implications of buffering the entire response, and I agree that a bounded-memory incremental hashing approach is a better solution for streaming responses.

I’ll take a look at #5202/#5205 and learn from the approach there. Glad the underlying chunk-boundary inconsistency is being addressed.

Issue #4842, opened 2026-03-30
Observability through `TracingChannels` API


logaretm:
### What is the feature you are proposing?

I'd like to propose adding first-class [`TracingChannel`](https://nodejs.org/api/diagnostics_channel.html#class-tracingchannel) support to Hono, following the pattern established by [`undici`](https://github.com/nodejs/undici) in Node.js core and adopted by framework peers like [`h3`](https://github.com/h3js/h3/pull/1251), [`fastify`](https://github.com/fastify/fastify), and [`srvx`](https://github.com/h3js/srvx/pull/141).

`TracingChannel` is a higher-level API built on top of `diagnostics_channel`, specifically designed for tracing async operations. It provides structured lifecycle channels (`start`, `end`, `error`, `asyncStart`, `asyncEnd`) and handles async context propagation correctly, this is the missing piece that makes monkey-patching approaches fragile in real-world async applications.

We (at Sentry) are currently working on an SDK for hono, and right. now we have to monkey patch 9 methods to ensure all paths are covered. On certain runtimes this gets tricky since those patches need to be re-applied each time a request comes in.

The Sentry case is particularly instructive: a single APM vendor had to build **two completely separate instrumentation approaches** for the same framework. One for Node.js (IITM-based) and one for Cloudflare Workers (Proxy + middleware-based) and explicitly filters out the Node integration when running on Workers to avoid double-instrumentation. This is the kind of complexity that native TracingChannel support eliminates entirely.

With TracingChannel, a single `diagnostics_channel` subscription works identically across Node.js, Cloudflare Workers, and any other runtime that supports the API. No IITM, no Proxy hacks, no per-runtime code paths, no user code changes.

If Hono emits structured events through `TracingChannel`, instrumentation libraries become **subscribers**, not **patches**. Each tool listens independently with no ordering concerns, no clobbering, and no internal API dependency.

## Where TracingChannel Fits: Relationship to `@hono/otel` and Built-in Middleware

Hono has two existing observability mechanisms:

1. **`hono/timing`:** a middleware that adds `Server-Timing` headers with `setMetric`, `startTime`, `endTime`. Useful for browser DevTools but not for APM/distributed tracing.
2. **`@hono/otel`:**  - the official OTel middleware in `honojs/middleware`. It uses `@opentelemetry/api` directly (`tracer.startActiveSpan()`) to create a root server span and record request metrics. It is a **standard Hono middleware**, not a framework-level instrumentation.

[...]

PR #5200, opened 2026-08-07
feat: introduce `handlerWrapper`

yusukebe:
To implement the function to measure the performance per handler, there is no good way to do it. So I think the demand, like the follow-up PR, is raised:

https://github.com/honojs/hono/issues/4842

To do it, I don't like to write a hard-coded `TracingChannels` function, but introducing a flexible general method is a great design.

`handlerWrapper` will enable them to inject their own method to wrap each handler execution. The example:

```ts
const app = new Hono({
  handlerWrapper: async (c, next, route) => {
    try {
      return await route.handler(c, next)
    } finally {
      console.log(`${route.path} ${route.handler.name}: ${performance.now() - start}ms`)
    }
  },
})
```

Tracing channel:

```ts
const dc = (globalThis as any).process?.getBuiltinModule?.('node:diagnostics_channel')
const tc = dc.tracingChannel('hono:request')

tc.subscribe({
  start(m: any) {
    m._start = performance.now()
    console.log(
      `  [start] ${m.route.method} ${m.route.path} :: ${m.route.handler.name || '(anonymous)'}`
    )
  },
  asyncEnd(m: any) {
    const ms = (performance.now() - m._start).toFixed(2)
    console.log(`  [end]   ${m.route.handler.name || '(anonymous)'} — ${ms}ms`)
  },
  error(m: any) {
    console.log(`  [error] ${m.route.handler.name || '(anonymous)'}: ${m.error.message}`)
  },
})

const app = new Hono({
  handlerWrapper: (c, next, route) => tc.tracePromise(() => route.handler(c, next), { route, c }),
})
```

And there are other use cases.

You can know which handler the error is thrown in. Currently, you can't know where the error comes from in `app.onError()`.

```ts
const app = new Hono({
  handlerWrapper: async (c, next, route) => {
    try {
      return await route.handler(c, next)
    } catch (err) {
      throw new Error(`Error in ${route.method} ${route.path}`)
    }
  },
})
```

And you can change the handler dynamically.

```ts
const app = new Hono({
  handlerWrapper: (c, next, route) => {
    const impl = registry.get(route.method + route.path) ?? route.handler
    return impl(c, next)
  },
})
```

### Pain points

The bundle size will be increased. But I think it's acceptable:

CleanShot 2026-08-07 at 17 45 19@2x

- [ ] Add tests
- [ ] Run tests

### DIFF
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -1,5 +1,7 @@
 import type { Context } from './context'
-import type { Env, ErrorHandler, Next, NotFoundHandler } from './types'
+import type { Env, ErrorHandler, HandlerWrapper, Next, NotFoundHandler, RouterRoute } from './types'
+
+export const defaultHandlerWrapper: HandlerWrapper = (c, next, route) => route.handler(c, next)
 
 /**
  * Compose middleware functions into a single function based on `koa-compose` package.
@@ -15,7 +17,8 @@ import type { Env, ErrorHandler, Next, NotFoundHandler } from './types'
 export const compose = <E extends Env = Env>(
   middleware: [[Function, unknown], unknown][] | [[Function]][],
   onError?: ErrorHandler<E>,
-  onNotFound?: NotFoundHandler<E>
+  onNotFound?: NotFoundHandler<E>,
+  wrapper: HandlerWrapper = defaultHandlerWrapper
 ): ((context: Context, next?: Next) => Promise<Context>) => {
   return (context, next) => {
     let index = -1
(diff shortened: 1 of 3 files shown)


yusukebe:
Hey @usualoma !

What do you think of this approach?

usualoma:
Nice idea! I created #5201 as an alternative approach that solves this with a router.

I think each approach has its own pros and cons.

For an API that directly expresses what users want to do and is easy to discover, I think the `handlerWrapper` proposed in this PR is better.

The main advantage of the router-based approach is that it does not require any changes to Hono's existing internal APIs or handler execution paths for applications that do not use it. Using a router for this also feels aligned with Hono's design, since `Router<T>` is already a replaceable extension point.

Issue #4936, opened 2026-05-14
Introduce linters for GitHub Actions


3w36zj6:
I got this idea after seeing #4932. If this is already on your roadmap, please feel free to ignore it.

Managing GitHub Actions has its pitfalls. To help avoid known vulnerabilities, would we consider introducing linters?

The following tools are well-maintained and widely used:

- [actionlint](https://github.com/rhysd/actionlint)
- [ghalint](https://github.com/suzuki-shunsuke/ghalint)
- [zizmor](https://github.com/zizmorcore/zizmor)
- [pinact](https://github.com/suzuki-shunsuke/pinact)
  - Technically not a linter, but useful for pinning and updating action version hashes.

While their rules overlap somewhat, they don't conflict with each other, so, in my opinion, I suggest adopting all of them.

For tool management, I think it would be good to use `mise` with the existing `.tool-versions` or `mise.toml`. This ensures we use the exact same tool versions across both CI and local development environments[^1].

Additionally, since this setup is common, it might be useful to manage it via [Reusable workflows](https://docs.github.com/en/actions/concepts/workflows-and-actions/reusing-workflow-configurations) at the Organization level.

[^1]:  For context, BarryThePenguin appears to have personally been using mise for some time: https://github.com/honojs/middleware/pull/1366#issue-3298431440

yusukebe:
@3w36zj6 

Thank you for the proposals! Regarding linters, one big uncertainty for me is that they avoid the supply change attack, but if they are attacked, that would be a risk. What do you think of this?

yusukebe:
Fyi, TanStack starts using Zismor.

https://x.com/schanuelmiller/status/2055280077764452451?s=20

SeanCassiere:
> Fyi, TanStack starts using Zismor.
> 
> https://x.com/schanuelmiller/status/2055280077764452451?s=20

For TanStack, we run [Zizmor](https://github.com/zizmorcore/zizmor) in a separate workflow/action that has basically 0 permissions. Additionally, we recommend pinning the workflow using the SHA-hash commit so you also aren't vulnerable to floating version attacks either.

Paired with [CODEOWNERS](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) - then you'd only update these versions using something like Dependabot or Renovate, so these changes to the workflow will require your equivalent to a 'core maintainer' to review it.

3w36zj6:
[...]

PR #4932, opened 2026-05-12
ci: pin GitHub Actions to SHAs

yusukebe:
It's a dangerous world now. I pinned the GitHub Actions to SHAs. Plus, set least-privilege permissions.

### The author should do the following, if applicable

- [ ] Add tests
- [ ] Run tests
- [ ] `bun run format:fix && bun run lint:fix` to format the code
- [ ] Add [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code

PR #5203, opened 2026-08-08
ci: introduce linters for GitHub Actions

tarikermis:
Refs #4936

- [ ] Add tests
- [ ] Run tests

(Not applicable: this PR only touches `.github/**` workflow definitions — no TypeScript sources, so there are no tests/docs to add and no code paths that `bun run test`, `format`, or `lint` cover. `bun run editorconfig-checker`, which does cover YAML, passes.)

## What & why

Following up on #4932 (pin GitHub Actions to SHAs), #4936 proposes linting the workflows themselves. This PR adds a new `lint-actions` workflow running the two linters that came up in the issue discussion:

- [actionlint](https://github.com/rhysd/actionlint) v1.7.12 — workflow syntax/semantics lint. Installed from the official release tarball with SHA-256 checksum verification (same pattern as the existing `bombardier` install in `ci.yml`), so no new third-party action needs to be trusted.
- [zizmor](https://github.com/zizmorcore/zizmor) v1.29.0 — static security analysis for workflows, via the official SHA-pinned `zizmorcore/zizmor-action`, following the setup described for TanStack in the issue: separate workflow, read-only `contents: read` permission, `persist-credentials: false` on checkout, `advanced-security: false` + `annotations: true`. The zizmor version is pinned (`version: 1.29.0`) for reproducibility, mirroring how the repo pins everything else.

The workflow runs on pushes to `main`/`next` and on PRs that touch `.github/workflows/**` or `.github/actions/**`.

## Findings fixed in existing workflows

Running both linters against the current workflows (the "reproduction" step) surfaced the following, all fixed here so the new check is green from day one:

- `artipacked` (18 findings): every `actions/checkout` now sets `persist-credentials: false`. No job pushes over git — `autofix-ci/action` uploads the diff to the autofix.ci GitHub App instead of pushing, and npm/JSR publishing uses OIDC — so persisting credentials is unnecessary.
- `cache-poisoning` (4 findings): in the publish workflows (`release.yml`, `cr.yml`), `actions/setup-node`'s automatic package-manager cache is disabled via `package-manager-cache: false` and `oven-sh/setup-bun`'s cache via `no-cache: true`.
- `ref-version-mismatch` (28 findings, online audit): version comments on SHA-pinned actions now name the exact tag the pinned SHA points to (e.g. `# v6` → `# v6.0.2`, all verified against the upstream tag refs). I applied the same exact-tag style to the remaining, not-yet-flagged comments (`# v2` → `# v2.2.0`, etc.) for consistency — this also keeps the check deterministic, since a floating major tag moving upstream would otherwise flag the pin without any change in this repo. Happy to drop those if you prefer a smaller diff.
- `adhoc-packages` (1 finding): `npm install -g npm@11.17.0` in `release.yml` is kept with an inline `# zizmor: ignore[adhoc-packages]` justification — the npm version is exactly pinned and required for staged publishing.

actionlint itself was already clean; the new workflow file passes both linters (dogfooded).

## Verification

- `actionlint` v1.7.12 on the repo: exit 0. Run locally with shellcheck 0.9.0 on PATH, matching `ubuntu-latest`, so CI parity is confirmed (also verified clean with `-shellcheck=` disabled).
- `zizmor` v1.29.0 in online mode (same persona/audits as the CI job): `No findings to report. Good job! (1 ignored, 12 suppressed)`, exit 0.
- `bun install --frozen-lockfile` and `bun run editorconfig-checker`: pass.
- The full diff was reviewed with an AI code-review pass (kiro-cli, claude-opus-5); its findings were either verified against evidence (ignore-comment syntax, SHA↔tag matching — both enforced by zizmor's online audits) or addressed above.

### Testing limitations / notes

- The new workflow only executes on GitHub Actions after merge (or on this PR once it touches `.github/**`); I validated equivalent commands locally with the same pinned tool versions rather than via a live CI run.
- If these jobs ever become *required* status checks, the `paths` filter should be reconsidered, since skipped runs report no status.
- Disabling the default caches in `cr.yml`/`release.yml` makes those (infrequent) publish jobs slightly slower, which is the accepted trade-off for the cache-poisoning findings.
- Per the AI usage policy: this PR was prepared with AI assistance; every change was verified locally as described above.

### DIFF
--- a/.github/actions/perf-measures/action.yml
+++ b/.github/actions/perf-measures/action.yml
@@ -10,7 +10,7 @@ inputs:
 runs:
   using: 'composite'
   steps:
-    - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2
+    - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2.2.0
       with:
         bun-version-file: '.tool-versions'
     - run: |
@@ -44,7 +44,7 @@ runs:
 
     - name: Run octocov
       if: ${{ inputs.target-ref == 'auto' }}
-      uses: k1LoW/octocov-action@b3b6ee60482a667950f87553abf1df63217235d9 # v1
+      uses: k1LoW/octocov-action@b3b6ee60482a667950f87553abf1df63217235d9 # v1.5.1
       with:
         config: perf-measures/.octocov.consolidated.perf-measures.yml
       env:
@@ -54,7 +54,7 @@ runs:
 
     - name: Run octocov with custom target ref
       if: ${{ inputs.target-ref == 'main' }}
-      uses: k1LoW/octocov-action@b3b6ee60482a667950f87553abf1df63217235d9 # v1
+      uses: k1LoW/octocov-action@b3b6ee60482a667950f87553abf1df63217235d9 # v1.5.1
       with:
         config: perf-measures/.octocov.consolidated.perf-measures.main.yml
       env:
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -19,14 +19,16 @@ jobs:
     if: ${{ github.event_name == 'push' || !github.event.pull_request.draft }}
     steps:
       - name: Checkout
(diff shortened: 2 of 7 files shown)


yusukebe:
Discussion is not finished yet. This is not needed. Closing.

PR #5205, opened 2026-08-09
fix(etag): stabilize digest across stream chunks

usualoma:

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/etag/digest.ts
+++ b/src/middleware/etag/digest.ts
@@ -13,6 +13,8 @@ const mergeBuffers = (
   return merged
 }
 
+const CHUNK_SIZE = 64 * 1024
+
 export const generateDigest = async (
   stream: ReadableStream<Uint8Array<ArrayBuffer>> | null,
   generator: (body: Uint8Array<ArrayBuffer>) => ArrayBuffer | Promise<ArrayBuffer>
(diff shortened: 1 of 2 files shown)


yusukebe:
@usualoma Thanks!

I also think the same approach. At first glance, this seems to be fine. Is there any issue with performance?

usualoma:
Before this PR, the digest function was called once for each source stream chunk. Although chunk sizes depend on the runtime and network conditions, network responses commonly arrive in chunks ranging from a few KiB to a few dozen KiB. Even filesystem streams use 64 KiB by default in Node.js. Therefore, using a fixed 64 KiB chunk size would generally have comparable performance or be slightly faster in common cases.

That said, I think using up to 256 KiB here is also reasonable. A larger chunk size reduces the number of hash function calls, so this PR should provide a small performance improvement. The implementation also avoids eagerly allocating the entire 256 KiB buffer for small responses: a single small chunk is hashed directly, and the buffer grows only as needed.

I updated the chunk size to 256 KiB in commit 15eb50bf.

Could you please review it again with this change?

ababove20-jpg (approved):

yusukebe (approved):
LGTM!

yusukebe:
@usualoma 

Using `256 KiB` is reasonable. I checked; it looks good. Thank you!

ababove20-jpg (approved):

PR #5206, opened 2026-08-09
fix(types): infer handler response from last handler in path[] overload

### LINKED ISSUE #5144: type bug: no overload matches this call (multiple paths + middleware)

4.12.31

Node 26.4.0, TypeScript 7.0.2

This example using multiple paths and middleware in a route won't compile using the latest TypeScript 7.0.2.

Example: 
```ts
import type { HttpBindings } from '@hono/node-server';
import { type Context, Hono, type Next } from 'hono';

type Bindings = HttpBindings & {};

const app = new Hono<{ Bindings: Bindings }>();

const middleware = async (c: Context, next: Next) => {
	console.log('just an example');
	return next();
};

app.on('POST', ['/route1', '/route2'], middleware, (c) => {
	return c.json({ message: 'Hello from route1 or route2' });
});
```

Error:
```
% npm run build

> hono-test@1.0.0 build
> tsc --noEmit

index.ts:13:52 - error TS2769: No overload matches this call.
    Argument of type '(c: Context<any, "/route1" | "/route2", {}>) => JSONRespondReturn<{ message: string; }, ContentfulStatusCode>' is not assignable to parameter of type 'H<any, "/route1" | "/route2", {}, Promise<void>>'.
      Type '(c: Context<any, "/route1" | "/route2", {}>) => JSONRespondReturn<{ message: string; }, ContentfulStatusCode>' is not assignable to type 'Handler<any, "/route1" | "/route2", {}, Promise<void>>'.
        Type 'JSONRespondReturn<{ message: string; }, ContentfulStatusCode>' is missing the following properties from type 'Promise<void>': then, catch, [Symbol.toStringTag], finally

13 app.on('POST', ['/route1', '/route2'], middleware, (c) => {
                                                      ~~~~~~~~

  node_modules/hono/dist/types/types.d.ts:446:5 - The last overload is declared here.
```

tsconfig is default via `npx tsc --init`

Dependencies:
```
"dependencies": {
    "@hono/node-server": "2.0.10",
    "hono": "4.12.31"
  },
  "devDependencies": {
    "@types/node": "26.1.1",
    "typescript": "7.0.2"
  }
```

Compile successfully.

### What do you see instead?

_No response_

### Additional information

_No response_

pacocartones:
Fixes #5144

## Problem

`app.on('POST', ['/route1', '/route2'], middleware, handler)` failed to compile with:

```
  Argument of type '(c: Context<any, "/route1" | "/route2", {}>) => JSONRespondReturn<...>' is not assignable to parameter of type 'H<any, "/route1" | "/route2", {}, Promise<void>>'.
```

## Root cause

The `app.on(method | method[], path[], ...handlers[])` overload typed **every** handler as `H<..., I, R>[]`, so TypeScript inferred `R` from the first handler. An inline middleware `async (c, next) => { return next() }` has inferred return type `Promise<void>` (from `next()`), which forced `R = Promise<void>` — and then the final handler returning `c.json(...)` no longer matched.

The single-path and `method[]` overloads already handle this correctly by only giving the **last** handler the response type (`[...middleware, handler]`); the `path[]` overload was the odd one out.

## Fix

Rework the `path[]` overload to the same shape:

```ts
...handlers: [...H<E2, MergedPath, I>[], H<E3, MergedPath, I2, R>]
```

Only the last handler carries `R`; preceding handlers are middleware. Input/env types are still accumulated across all handlers.

## Testing

- New regression test reproducing the exact issue (inline middleware returning `next()`, multiple paths, JSON handler) — verified it **fails without the fix** (1 TS error) and **passes with it**.
- Full `types.test.ts` suite: 121/121 passing.
- `tsc -p tsconfig.spec.json` clean.

### DIFF
--- a/.changeset/fix-path-array-overload.md
+++ b/.changeset/fix-path-array-overload.md
@@ -1,0 +1,5 @@
+---
+'hono': patch
+---
+
+Fixed `app.on()` type inference when multiple paths are combined with middleware: an inline middleware whose inferred return type is `Promise<void>` (e.g. an `async` function returning `next()`) made the `path[]` overload infer `R = Promise<void>` for **every** handler, so the final handler returning a JSON response was rejected with "No overload matches this call". The `path[]` overload now mirrors the single-path and `method[]` overloads: only the last handler carries the response type, and preceding handlers are treated as middleware.
--- a/src/types.test.ts
+++ b/src/types.test.ts
@@ -2,6 +2,7 @@
 import { expectTypeOf } from 'vitest'
 import { hc } from './client'
 import { Context } from './context'
+import type { Next } from './types'
 import { createMiddleware } from './helper/factory'
 import { testClient } from './helper/testing'
 import { Hono } from './hono'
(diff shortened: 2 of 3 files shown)


yusukebe:
In the case of [the problem](https://github.com/honojs/hono/issues/5144), https://github.com/honojs/hono/issues/5144#issuecomment-5234811229 can fix it. You don't have to go that far.

PR #5209, opened 2026-08-10
fix(url): strip trailing question mark correctly for optional params with regex quantifiers

spellsaif:
Fixes `checkOptionalParameter` in `src/utils/url.ts` to properly strip the trailing `?` optional parameter indicator instead of stripping `?` quantifiers inside custom parameter regexes (e.g., `/:id{[0-9]?}?`).

### Motivation & Problem
When defining routes with optional parameters containing a RegExp with a `?` quantifier (such as `/:id{[0-9]?}?` or `/:lang{[a-z]{2}?}?`), `segment.replace('?', '')` replaces the **first** `?` encountered in the segment string.

This corrupts the inner RegExp (changing `[0-9]?` to `[0-9]`) and leaves the trailing `?` intact at the end of the segment (`/api/:id{[0-9]}?`), causing malformed route registration in the router matcher.

### Solution
Change `segment.replace('?', '')` to `segment.replace(/\?$/, '')` in `checkOptionalParameter` so that only the trailing optional parameter indicator `?` at the end of the path segment is removed.

### Testing
- Added unit test case `checkOptionalParameter('/api/:id{[0-9]?}?')` in `src/utils/url.test.ts`.
- Verified `npm test src/utils/url.test.ts` passes 100% cleanly.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/utils/url.test.ts
+++ b/src/utils/url.test.ts
@@ -253,6 +253,7 @@ describe('url', () => {
         '/api/:version/animal',
         '/api/:version/animal/:type',
       ])
+      expect(checkOptionalParameter('/api/:id{[0-9]?}?')).toEqual(['/api', '/api/:id{[0-9]?}'])
     })
   })
 
--- a/src/utils/url.ts
+++ b/src/utils/url.ts
@@ -193,7 +193,7 @@ export const checkOptionalParameter = (path: string): string[] | null => {
         } else {
           results.push(basePath)
         }
-        const optionalSegment = segment.replace('?', '')
+        const optionalSegment = segment.replace(/\?$/, '')
         basePath += '/' + optionalSegment
         results.push(basePath)
       } else {
--- a/src/utils/url.ts
+++ b/src/utils/url.ts
@@ -187,13 +187,13 @@ export const checkOptionalParameter = (path: string): string[] | null => {
     if (segment !== '' && !/\:/.test(segment)) {
       basePath += '/' + segment
     } else if (/\:/.test(segment)) {
-      if (/\?/.test(segment)) {
+      if (segment.charCodeAt(segment.length - 1) === 63) {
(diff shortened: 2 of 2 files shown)


usualoma:
Hi @spellsaif,

Thanks for catching this. This is indeed a bug, and your proposed change fixes the reported case.

While reviewing it, I found that the preceding condition was also incorrect:

```ts
if (/\?/.test(segment)) {
```

This treats any `?` in the segment, including quantifiers inside a custom RegExp, as an optional parameter marker. Could you update the fix to cover both the detection and removal, so that only a trailing `?` is treated as the marker? Something like this:

```diff
diff --git a/src/utils/url.ts b/src/utils/url.ts
index ea92ff93..12493648 100644
--- a/src/utils/url.ts
+++ b/src/utils/url.ts
@@ -187,13 +187,13 @@ export const checkOptionalParameter = (path: string): string[] | null => {
     if (segment !== '' && !/\:/.test(segment)) {
       basePath += '/' + segment
     } else if (/\:/.test(segment)) {
-      if (/\?/.test(segment)) {
+      if (segment.charCodeAt(segment.length - 1) === 63) {
         if (results.length === 0 && basePath === '') {
           results.push('/')
         } else {
           results.push(basePath)
         }
-        const optionalSegment = segment.replace('?', '')
+        const optionalSegment = segment.slice(0, -1)
         basePath += '/' + optionalSegment
         results.push(basePath)
       } else {
```

Thanks again for reporting this.

spellsaif:
Thanks @usualoma, I made necessary changes as you suggested. :)

usualoma:
@spellsaif I've confirmed it. I think it's good!

yusukebe (approved):
LGTM!

yusukebe:
@spellsaif @usualoma 

Thanks! Looks good!

PR #5221, opened 2026-08-12
feat(mount)!: introduce Mount Middleware and obsolete `app.mount`

yusukebe:
I realized that `app.mount()` doesn't need to be placed in `hono-base.ts`. The codebase of `app.mount()` is a little bit fat compared to other methods in `hono-base.ts`, though the lines are very few in the real world. Anyway, the unnecessary should be removed from `hono-base.ts`.

So, we can make it a middleware called Mount Middleware. The usage is below:

```ts
import { Router as IttyRouter } from 'itty-router'
import { Hono } from 'hono'
import { mount } from 'hono/mount'

const ittyRouter = IttyRouter()
ittyRouter.get('/hello', () => new Response('Hello from itty-router'))

const app = new Hono()
app.all('/itty-router/*', mount(ittyRouter.handle))
```

I think the API syntax is very Hono-like and love it.

As a result, the bundle size will be `501B` smaller (vs current main):

CleanShot 2026-08-12 at 21 31 59@2x

This PR will obsolete `app.mount()` completely, so this is a breaking change. Should be introduced in the next major release v5.

- [x] Add tests
- [x] Run tests


yusukebe:
Hey @usualoma !

What do you think of moving the mount feature from `hono-base.ts` to middleware? I want your opinion on the direction only (no implementation yet)!

usualoma:
@yusukebe 
That's a great change!

I think `mount` is a feature that symbolizes Hono, but I don't think it's a “characteristic of the Hono instance itself,” so I think it makes sense for it to be implemented as middleware (which will also make the Hono instance smaller).

PR #5222, opened 2026-08-12
fix(etag): match If-None-Match tags with optional whitespace before the comma

contactjawad:
Fixes the ETag middleware so `If-None-Match` matches when optional whitespace precedes the comma in the tag list.

## What

The ETag middleware returned `200` with the full body instead of `304 Not Modified` when a conditional request's `If-None-Match` header placed whitespace before a delimiting comma (e.g. `"a" , "b"`), even though the current ETag was present in the list.

## Why

`etagMatches` split the header with `/,\s*/`, which only consumes whitespace *after* each comma. RFC 9110 defines `If-None-Match` as a `1#entity-tag` list, and the `#` list rule permits optional whitespace (OWS) on **both** sides of the delimiter. When OWS preceded a comma, the preceding tag kept a trailing space, so `stripWeak('"abc" ') !== stripWeak('"abc"')` and the match silently failed. Any tag not in the last list position was affected, causing an unnecessary full-body transfer instead of a cheap `304`.

The fix trims the header and splits on `/\s*,\s*/`, so OWS on either side of the comma is consumed. A regression test covers a matching tag with a space before the following comma.

- [x] Add tests
- [x] Run tests
- [x] `format:fix && lint:fix` to format the code

### DIFF
--- a/src/middleware/etag/index.ts
+++ b/src/middleware/etag/index.ts
@@ -32,10 +32,7 @@ const stripWeak = (tag: string) => tag.replace(/^W\//, '')
 function etagMatches(etag: string, ifNoneMatch: string | null) {
   return (
     ifNoneMatch != null &&
-    ifNoneMatch
-      .trim()
-      .split(/\s*,\s*/)
-      .some((t) => stripWeak(t) === stripWeak(etag))
+    ifNoneMatch.split(',').some((t) => stripWeak(t.trim()) === stripWeak(etag))
   )
 }
 
--- a/src/middleware/etag/index.ts
+++ b/src/middleware/etag/index.ts
@@ -31,7 +31,11 @@ const stripWeak = (tag: string) => tag.replace(/^W\//, '')
 
 function etagMatches(etag: string, ifNoneMatch: string | null) {
   return (
-    ifNoneMatch != null && ifNoneMatch.split(/,\s*/).some((t) => stripWeak(t) === stripWeak(etag))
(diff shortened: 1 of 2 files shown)


src/middleware/etag/index.ts
```diff
@@ -31,7 +31,11 @@ const stripWeak = (tag: string) => tag.replace(/^W\//, '')
 
 function etagMatches(etag: string, ifNoneMatch: string | null) {
   return (
-    ifNoneMatch != null && ifNoneMatch.split(/,\s*/).some((t) => stripWeak(t) === stripWeak(etag))
+    ifNoneMatch != null &&
+    ifNoneMatch
+      .trim()
+      .split(/\s*,\s*/)
```
usualoma: Since `/\s*,\s*/` has a ReDoS issue, if this PR is merged, I think we need to use `split` on the `,` and then call `trim()`, as shown in the code below.


https://github.com/honojs/hono/blob/41bdc4288ed7023c24d35f32bd065bd963435e3e/src/middleware/serve-static/index.ts#L98-L100

contactjawad: Good catch, thanks. Updated to split on ',' and trim each token instead, so there's no regex backtracking — matches the serve-static pattern you linked. All the existing etag tests still pass.

yusukebe (approved):
LGTM!

yusukebe:
@contactjawad @usualoma Thanks!

PR #5224, opened 2026-08-13
fix(serve-static): skip non-GET/HEAD requests so methodNotAllowed fires (fix #5223)

### LINKED ISSUE #5223: When serveStatic and methodNotAllowed are used together, methods other than GET cannot be handled correctly

4.13.1

Node.js

with `get`:
```ts
const app = new Hono()
app.get("/*", serveStatic());
```

with `use`:
```ts
const app = new Hono()
app.use("/*", serveStatic());
```

- return 200 on GET /api,
- return 405 on POST /api,

### What do you see instead?

with `get(serveStatic)`:
- return 405 on POST /api,

with `use(serveStatic)`,
- return 405 on POST /api,

### Additional information

Alternatively,

Sskift:
## Description

Fixes #5223: `serveStatic` no longer serves files on non-GET/HEAD requests.

When `serveStatic` is registered with `use('/*', serveStatic())` (as shown in the docs), it intercepts every method. A `POST` to an existing static file returned `200` instead of falling through, so the `methodNotAllowed` middleware never got a chance to respond with `405`.

## Root cause

The `serveStatic` handler (src/middleware/serve-static/index.ts) did not filter by HTTP method. It returned a `200` for any request whose path resolved to an existing file, regardless of method.

## Fix

Skip non-GET/HEAD requests and call `next()` so downstream middleware (e.g. `methodNotAllowed`) can handle them:

```ts
if (c.req.method !== 'GET' && c.req.method !== 'HEAD') {
  return next()
}
```

## Testing

- Added 2 tests in `src/middleware/serve-static/index.test.ts`:
  - POST to an existing static file now falls through (404 / `next()`), `getContent` is not called.
  - Combined with `methodNotAllowed`, a POST to a path with a real GET route still returns `405` with `Allow: GET, HEAD`.
- TDD: both new tests failed before the fix (returned `200`), pass after.
- Regression: `serve-static` suite (22/22) and `method-not-allowed` + `path` suites (21/21) all pass.

## Note on expected behavior

For `use('/*', serveStatic())` with `methodNotAllowed`, a `POST` to a static file now returns `404` (not `405`) because `use()` does not register a GET route that `methodNotAllowed` can infer an `Allow` header from. This matches the "or maybe 404 is also accepted in this case?" option in the issue. Using `get('/*', serveStatic())` instead keeps the `405` behavior, since a GET route is registered.

Closes #5223

### DIFF
--- a/src/middleware/serve-static/index.test.ts
+++ b/src/middleware/serve-static/index.test.ts
@@ -1,4 +1,5 @@
 import { Hono } from '../../hono'
+import { methodNotAllowed } from '../method-not-allowed'
 import { serveStatic as baseServeStatic } from '.'
 
 describe('Serve Static Middleware', () => {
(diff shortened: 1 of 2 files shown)


usualoma:
Thanks for the PR. This may be a valid partial solution to #5223, but the expected behavior is still under discussion. If the issue author agrees with this approach, I’d prefer that they open a PR.

This PR also appears to be largely AI-generated. Under our [AI Usage Policy](https://github.com/honojs/hono/blob/main/docs/CONTRIBUTING.md#ai-usage-policy), I’m closing it for now. Apologies if I’m mistaken.

PR #5201, opened 2026-08-07
feat: introduce `TransformRouter`

usualoma:
This is an alternative approach to #5200.

This PR introduces `TransformRouter`, a generic router decorator. It transforms registered handlers once in `add()` and delegates matching to another router without changing Hono's handler execution path.

For Hono, the registered value is `[handler, route]`, so each handler can be wrapped like this:

```ts
import { Hono } from 'hono'
import { TransformRouter } from 'hono/router/transform-router'

const app = new Hono({
  router: new TransformRouter({
    transform: ([handler, route]) => [
      async (c, next) => {
        const start = performance.now()
        try {
          return await handler(c, next)
        } finally {
          console.log(
            `${route.path} ${route.handler.name}: ${performance.now() - start}ms`
          )
        }
      },
      route,
    ],
  }),
})
```

### Tracing channel

```ts
const dc = (globalThis as any).process?.getBuiltinModule?.(
  'node:diagnostics_channel'
)
const tc = dc.tracingChannel('hono:request')

tc.subscribe({
  start(m: any) {
    m._start = performance.now()
    console.log(
      `  [start] ${m.route.method} ${m.route.path} :: ${
        m.route.handler.name || '(anonymous)'
      }`
    )
  },
  asyncEnd(m: any) {
    const ms = (performance.now() - m._start).toFixed(2)
    console.log(`  [end]   ${m.route.handler.name || '(anonymous)'} — ${ms}ms`)
  },
  error(m: any) {
    console.log(
      `  [error] ${m.route.handler.name || '(anonymous)'}: ${m.error.message}`
    )
  },
})

const app = new Hono({
  router: new TransformRouter({
    transform: ([handler, route]) => [
      (c, next) =>
        tc.tracePromise(() => handler(c, next), {
          route,
          c,
        }),
      route,
    ],
  }),
})
```

### Other use cases

You can identify which handler threw an error before `app.onError()` handles it:

```ts
const app = new Hono({
  router: new TransformRouter({
    transform: ([handler, route]) => [
      async (c, next) => {
        try {
          return await handler(c, next)
        } catch (err) {
          throw new Error(`Error in ${route.method} ${route.path}`)
        }
      },
      route,
    ],
  }),
})
```

Handlers can also be selected dynamically:

```ts
const app = new Hono({
  router: new TransformRouter({
    transform: ([handler, route]) => [
      (c, next) => {
        const impl = registry.get(route.method + route.path) ?? handler
        return impl(c, next)
      },
      route,
    ],
  }),
[...]

Issue #5193, opened 2026-08-05
Middleware on notFound / onError


na-trium-144:
### What is the feature you are proposing?

Currently, `onError` and `notFound` only accept a single handler. It would be great if they could accept middlewares, similar to route definitions.

Use Case: running specific middleware (e.g., `languageDetector`) when rendering 404 page or error page, without registering it globally.
(For example you might not want to use `languageDetector` globally because it has the side effect of disabling the edge cache due to the `Set-Cookie`.)

```ts
const app = new Hono()

app.notFound(
  languageDetector(),
  (c) => {
    return c.html(render404Page(c.get("language")))
  }
)

app.onError(
  languageDetector(),
  (err, c) => {
    return c.html(renderErrorPage(c.get("language")))
  }
)
```

MonisMS:
This would be useful for my case too (X). @na-trium-144 are you planning to implement this? If not, I'd be happy to take it

na-trium-144:
I think this won't be a simple patch but requires consideration on changes to the internal structure, so I would wait for the maintainer.
Personally I don't mind who take this though.

yusukebe:
To implement this, we have to add code to `hono-base.ts`. This is not as easy as some think.

SnowingFox:
Opened a PR to address this: https://github.com/honojs/hono/pull/5211

usualoma:
While I think it would be overengineering if we viewed this as a solution intended solely for this issue, it might be worth considering something like the following.
https://github.com/honojs/hono/pull/5229

PR #5229, opened 2026-08-14
[PoC] feat: dispatch notFound and onError handlers as internal routes

usualoma:
Taken in isolation as a solution to #5193, this may look over-engineered. The broader goal of this PR is to model error and not-found handling using Hono's own routing system. The current proposal does this through the new `catch()` and `catchNotFound()` APIs while leaving the existing `onError()` and `notFound()` semantics unchanged. Existing applications should not require any changes.

The implementation increases the minified bundle size by 3.35%. In particular, however, the single-handler fast path used by successful requests without middleware remains structurally unchanged, and local benchmarks show no performance regression.

## Summary

This draft introduces routable middleware for error and not-found handling:

```ts
app.catchNotFound(
  languageDetector(),
  (c) => c.html(render404Page(c.get('language')))
)

app.catch(
  languageDetector(),
  (c) => c.html(renderErrorPage(c.get('language'), c.error), 500)
)
```

Unlike `onError()`, `catch()` uses the normal middleware signature `(c, next)`. The caught error is available as `c.error`.

The middleware can also be scoped explicitly by path:

```ts
api.catch('/items/*', (c) =>
  c.text(`items error: ${c.error?.message}`, 500)
)

api.catchNotFound('/items/*', (c) =>
  c.text('items not found', 404)
)
```

These definitions are registered with reserved internal methods, `@ERROR` and `@NOT_FOUND`, and selected through the configured Router.

The existing final handlers remain unchanged:

```ts
app.onError((error, c) => c.text(error.message, 500))
app.notFound((c) => c.text('Not Found', 404))
```

If all matching `catch()` or `catchNotFound()` middleware calls `next()`, execution reaches the corresponding legacy handler.

> [!IMPORTANT]
> The existing APIs remain compatible, but this changes internal fallback dispatch, adds reserved internal routes, and adds route-depth metadata. This is still a PoC and should be evaluated with a major version in mind.

## Why internal routes?

A definition such as:

```ts
app.catchNotFound(
  languageDetector(),
  renderNotFound
)
```

can be understood as a route definition that is internally dispatched when a not-found response is requested.

This gives fallback middleware the useful properties of normal route middleware:

- middleware composition
- path and base-path scoping
- sub-application flattening
- Router-level wrapping and instrumentation
- preservation of metadata describing where an error or explicit not-found originated

The internal dispatch reuses the existing `Context`, request path, and route metadata. It does not create another `Request` or restart the whole application middleware chain.

### `catch()`

`catch()` adds middleware that runs when an error is caught:

```ts
app.catch(async (c, next) => {
  console.error(c.error)
  await next()
})
```

A middleware may:

- call `next()` to continue through the matching error middleware
- modify the response after `next()`
- return a response without calling `next()` to stop the chain
- throw another error, which is passed to the legacy `onError()` handler

If every matching middleware calls `next()`, the current legacy `onError(error, c)` handler produces the final response.

### `catchNotFound()`

`catchNotFound()` adds middleware that runs when a not-found response is requested:

```ts
app.catchNotFound(async (c, next) => {
  c.header('x-not-found', 'true')
  await next()
})
```

It is used for both:

- implicit not-found handling
- a returned `c.notFound()` response

If every matching middleware calls `next()`, the current legacy `notFound(c)` handler produces the final response.

An error thrown by `catchNotFound()` enters the `catch()` / `onError()` flow.

### Path and base-path scoping

Without an explicit path, both APIs register at `*` within the current base path. With a path, they only run for matching requests:

```ts
const api = new Hono().basePath('/api')

api.catch('/items/*', errorMiddleware)
api.catchNotFound('/items/*', notFoundMiddleware)
```

Fallback matching is independent of the HTTP method of the original request.

Calling `catch()` or `catchNotFound()` does not change the current route path used by subsequent chained route methods.

## Sub-application ordering

Fallback routes are flattened into the parent Router together with normal routes when `route()` mounts a sub-application.

Each route records its application nesting depth. The depth is incremented whenever the route crosses a `route()` boundary.

Matching fallback middleware is composed in this order:

1. deeper sub-application routes first
2. registration order among routes at the same depth

For example:

```ts
app.catch(async (c) => c.text(`app: ${c.error?.message}`, 500))

sub.catch(async (_c, next) => {
  console.log('sub')
  await next()
})

app.route('/sub', sub)
```

For an error under `/sub`, the sub-application middleware runs first. If it calls `next()`, execution continues to the parent middleware.

For nested applications, the order is therefore:

```text
nested application -> sub-application -> parent application
```

This priority is based on application nesting, not path specificity. `basePath()` alone does not increase the depth.

Within the same application depth, matching definitions follow normal registration order:

```ts
app.catch('*', globalMiddleware)
app.catch('/items/*', itemsMiddleware)
```

For `/items/...`, `globalMiddleware` runs first. If it returns a response without calling `next()`, `itemsMiddleware` is not reached.

The same ordering rules apply to `catchNotFound()`.

## Flattening semantics

Fallback selection uses the original request path and the nesting depth of the matching fallback routes. It does not use the originating normal handler to establish an application ownership boundary.

For example, a parent route registered directly at `/api/example` may enter fallback middleware imported from an application mounted at `/api` because the request path matches that internal route.

Likewise, an error thrown by parent middleware while handling `/api/...` may enter the mounted API's error middleware.

This is consistent with the existing flattening model used by `route()` for paths and middleware.

Legacy `onError()` remains application-bound for compatibility. A sub-application with its own `onError()` continues to use that handler for errors thrown by its imported normal handlers.

## Error origin metadata

Internal dispatch does not overwrite the original route index.

Route parameters and origin-sensitive information such as `routePath(c)` therefore continue to describe the original route while fallback middleware runs.

If middleware throws after `await next()`, Hono restores the route index to the middleware that actually threw before dispatching the error middleware.

This separates three concerns:

- the request path determines which fallback routes match
- application depth determines sub-application priority
- the original route metadata describes where the fallback originated

## Internal routes and route metadata

Registrations made through `catch()` and `catchNotFound()` appear in `app.routes` with methods beginning with `@`.

The `@` prefix is reserved for internal routes. Internal methods are excluded from:

- `inspectRoutes()` / `showRoutes()`
- the `Allow` header generated by `methodNotAllowed()`

All routes created by Hono also carry depth metadata describing how many `route()` boundaries they have crossed.

Code that directly consumes `app.routes` may need to account for these additions.

If exposing internal routes through `app.routes` is undesirable, they can instead be kept in a completely separate internal collection. An earlier prototype showed that this was possible, but added roughly another 130 B to the minified bundle.

## Semantics preserved

- Existing `onError(error, c)` and `notFound(c)` signatures are unchanged.
- Repeated `onError()` and `notFound()` calls retain the existing last-assignment-wins behavior.
- Existing applications do not need to register `catch()` or `catchNotFound()`.
- Existing sub-application `onError()` behavior is preserved.
- `c.notFound()` is not a control-flow terminator. Its response must still be returned or awaited when it should become the route response.
- The default 404 and 500 responses are unchanged.
- Default `HTTPException#getResponse()` handling is unchanged.
- Internal dispatch reuses the original `Context`.
- Route parameters remain available while fallback middleware runs.
- Global `app.use()` middleware is not executed a second time during internal dispatch.
- An error thrown by a legacy error handler is still propagated.
- Successful normal routes do not pass through error or not-found middleware.

For example, this retains its existing behavior:

```ts
app.get('/', (c) => {
  c.notFound()
  return c.text('OK')
})
```

To use the not-found response, the handler must return it:

```ts
app.get('/', (c) => {
  return c.notFound()
})
```

## Router integration

Fallback middleware is registered using the configured Router's existing `add()` method and selected through `match()`.

This broadens the scope of router-decoration features such as the one proposed in #5201. A Router wrapper can wrap, instrument, or otherwise decorate `catch()` and `catchNotFound()` middleware alongside normal route handlers.

The legacy final `onError()` and `notFound()` handlers remain stored handlers rather than routes. A terminal `catch()` or `catchNotFound()` handler can provide the fully routable equivalent, while preserving the old APIs for compatibility.

## Performance

Internal routing is only entered when:

- normal routing does not produce a response
- a returned `c.notFound()` is dispatched
- a handler or middleware throws an error

It does not add another Router match to successful requests.

The existing successful-request execution structure is retained:

| Request path | Before | After |
| --- | --- | --- |
| Single matched handler | one Router match + direct fast path | one Router match + direct fast path |
| Multiple matched handlers | one Router match + `compose()` | one Router match + `compose()` |
| Not found / error | direct stored fallback handler | additional internal Router match |

The single-handler fast path remains in place, and no global fallback middleware layer is added to every request.

For successful requests involving middleware, `compose()` now evaluates one additional boolean condition when updating the route index. A very small regression is therefore theoretically possible.

A local 15-round microbenchmark measured:

| Successful request | `main` | This PR | Difference |
| --- | ---: | ---: | ---: |
| Single handler | 483.86 ns | 483.99 ns | +0.03% |
| With middleware | 772.21 ns | 769.75 ns | -0.32% |

These differences are within the observed benchmark noise.

Matching fallback routes are sorted by application depth before composition. This work occurs only during error or not-found dispatch and only when fallback middleware matched.

The existing wrapper used to preserve a mounted sub-application's custom `onError()` remains for compatibility. The new routable fallback middleware itself is flattened without adding wrappers to successful normal routes.

## Bundle size

Measured from clean builds based on `cf785287b` using the existing bundle-size configuration. Compressed sizes were generated deterministically with gzip level 9 and Brotli quality 11.

| | `main` | This PR | Difference |
| --- | ---: | ---: | ---: |
| Minified | 19,000 B | 19,636 B | +636 B (+3.35%) |
| gzip -9 | 7,836 B | 8,102 B | +266 B (+3.39%) |
| Brotli q11 | 7,148 B | 7,395 B | +247 B (+3.46%) |

The internal methods are defined inside `hono-base.ts`; they are not added to the public Router API.

## Appendix: Design alternatives and trade-offs

Several API, fallback-selection, and storage models were explored. The current branch favors compatibility with the existing handlers and predictable sub-application ordering over the smaller but less compatible initial implementation.

The size estimates below come from earlier prototypes and should be treated as directional. If a larger bundle-size increase or different semantics are acceptable, some of these designs can be revisited.

### Adding middleware directly to `onError()` and `notFound()`

The initial implementation allowed middleware to be passed directly to the existing APIs:

```ts
app.notFound(languageDetector(), renderNotFound)
app.onError(languageDetector(), renderError)
```

This looked compact, but changed repeated registrations from last-assignment-wins to route composition. It also made the meaning of chaining and the current route path ambiguous for existing code.

The current design leaves both existing APIs unchanged and introduces `catch()` and `catchNotFound()` for routable middleware.

### Deferring fallback registration

Another model interprets fallback registration as being deferred until the end of the current application or until the corresponding `route()` call.

This naturally places a sub-application's fallback before its parent's fallback.

However, actual deferred registration must account for nested applications, repeated mounts, registration after mounting, and the point at which the route table becomes final.

The current implementation achieves the intended ordering without buffering registrations. It records route depth during `route()` and sorts only the matched fallback routes before composition.

### Selecting fallbacks from the originating application

An earlier implementation preserved the base path of the normal route that initiated fallback dispatch.

It selected:

1. a fallback registered at the exact originating base path
2. the root fallback when no exact definition existed

This prevented a parent route whose URL happened to be below a mounted application from entering the mounted application's fallback.

It also meant that:

- an implicit `/api/unknown` used the root fallback because it had no originating route
- a parent handler registered directly at `/api/example` used the parent fallback
- a handler imported through `app.route('/api', subApp)` used the sub-application fallback

This provides a stronger application boundary, but `app.route()` otherwise flattens paths and middleware into the parent routing space. Using handler provenance only for fallback selection therefore introduced a separate ownership model that did not apply to normal routing.

The implementation also required carrying the originating base path through dispatch, selecting an exact scope manually, and forwarding that scope through error handling. The earlier prototype added approximately another 200 B to the minified bundle compared with the flattened model at that time.

The current design keeps origin metadata for inspection but does not use it to select fallback middleware.

### Selecting the most specific path scope

Another prototype selected only the most specific matching internal path scope, independently of registration order.

This makes a scoped fallback reliably override a root fallback:

```ts
app.catchNotFound('*', renderAppNotFound)
app.catchNotFound('/api/*', renderApiNotFound)
```

Under that model, `/api/unknown` would always use `renderApiNotFound`, regardless of which definition was registered first.

However, Hono normally composes matching routes in registration order. Adding an independent path-specificity rule only for fallback routes would introduce a second ordering model.

A segment-aware implementation also required scanning and comparing matched scopes. The earlier prototype added approximately 126 B to the minified bundle compared with composing matching internal routes directly.

The current implementation distinguishes application nesting from path specificity. Deeper `route()` applications run first, while definitions at the same application depth retain registration order.

### Keeping internal routes outside `app.routes`

Internal routes could be stored in a completely separate collection while still being registered with the configured Router.

This would prevent raw `app.routes` consumers from seeing methods such as `@NOT_FOUND` and `@ERROR`.

However, it would require separate storage and mounting logic for internal routes. The earlier prototype increased the minified bundle by roughly another 130 B.

The selected implementation keeps a single route collection and filters reserved methods from user-facing helpers and the `Allow` header.

### Making `c.notFound()` a control-flow terminator

A terminal model was also considered, where calling `c.notFound()` would immediately stop the current handler even if its result was not returned:

```ts
app.get('/', (c) => {
  c.notFound()
  return c.text('unreachable')
})
```

Implementing this required a special control-flow signal to survive middleware composition and asynchronous handlers. It also changed how user `try` / `catch` blocks and upstream middleware observed the call.

The additional machinery was relatively large, and the behavior would differ substantially from the existing function-call semantics.

The current implementation therefore preserves the requirement to return or await `c.notFound()`.

### Reusing the normal request dispatch path

Internal fallback could have been implemented by re-entering the complete normal dispatch path with a reserved method or internal URL.

That would reduce the conceptual distinction between normal and fallback routing, but would either:

- create another `Request` or `Context`
- execute global middleware a second time
- add conditionals or wrappers to the successful-request path

These consequences conflict with the goal of preserving the original context and normal-request performance.

The selected implementation performs a separate Router match only after fallback dispatch has already been requested.

### Removing per-route wrappers for legacy sub-application errors

A sub-application with a custom `onError()` currently wraps its imported normal handlers so errors can be delegated to that stored handler.

Removing these wrappers would simplify the flattening model and allow every error handler to be represented only as a route. However, it would change the existing sub-application `onError()` behavior.

The current implementation retains the wrapper for the legacy API and adds routable `catch()` middleware alongside it. A future major version could reconsider the legacy handler after a migration path exists.

- [x] Add tests
- [x] Run tests
- [x] Add TSDoc/JSDoc where applicable

### DIFF
--- a/src/adapter/cloudflare-pages/handler.test.ts
+++ b/src/adapter/cloudflare-pages/handler.test.ts
@@ -79,8 +79,8 @@ describe('Adapter for Cloudflare Pages', () => {
     })
 
     const handler = handle(app)
-    // It does reject with the error if app is NOT "subApp"
-    await expect(handler(eventContext)).rejects.toThrowError('Custom Error')
+    // It does throw the error if app is NOT "subApp"
+    expect(() => handler(eventContext)).toThrowError('Custom Error')
   })
 })
 
--- a/src/adapter/cloudflare-pages/handler.test.ts
+++ b/src/adapter/cloudflare-pages/handler.test.ts
@@ -79,8 +79,8 @@ describe('Adapter for Cloudflare Pages', () => {
     })
 
     const handler = handle(app)
-    // It does throw the error if app is NOT "subApp"
-    expect(() => handler(eventContext)).toThrowError('Custom Error')
+    // It does reject with the error if app is NOT "subApp"
+    await expect(handler(eventContext)).rejects.toThrowError('Custom Error')
   })
 })
 
--- a/src/adapter/vercel/handler.test.ts
+++ b/src/adapter/vercel/handler.test.ts
@@ -33,6 +33,6 @@ describe('Adapter for Next.js', () => {
 
     const handler = handle(app)
     const req = new Request('http://localhost/api/error')
-    await expect(handler(req)).rejects.toThrowError('Custom Error')
+    expect(() => handler(req)).toThrowError('Custom Error')
   })
 })
--- a/src/adapter/vercel/handler.test.ts
+++ b/src/adapter/vercel/handler.test.ts
@@ -33,6 +33,6 @@ describe('Adapter for Next.js', () => {
 
     const handler = handle(app)
     const req = new Request('http://localhost/api/error')
-    expect(() => handler(req)).toThrowError('Custom Error')
+    await expect(handler(req)).rejects.toThrowError('Custom Error')
   })
 })
--- a/src/compose.test.ts
+++ b/src/compose.test.ts
@@ -727,6 +727,28 @@ describe('Compose', function () {
     }
   })
 
+  it('should not update the route index when disabled', async () => {
+    const routeIndexes: number[] = []
+    const middleware = [
+      buildMiddlewareTuple(async (ctx: Context, next: Next) => {
+        routeIndexes.push(ctx.req.routeIndex)
+        await next()
+        routeIndexes.push(ctx.req.routeIndex)
+      }),
+      buildMiddlewareTuple((ctx: Context) => {
+        routeIndexes.push(ctx.req.routeIndex)
+        return ctx.text('OK')
+      }),
+    ]
+    const ctx = new Context(new Request('http://localhost/'))
+    ctx.req.routeIndex = 10
+
+    await compose(middleware, undefined, undefined, false)(ctx)
+
+    expect(routeIndexes).toEqual([10, 10, 10])
+    expect(ctx.req.routeIndex).toBe(10)
+  })
+
   it('should not get stuck on the passed in next', async () => {
     const middleware = [
       buildMiddlewareTuple((ctx: Context, next: Next) => {
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -9,13 +9,15 @@ import type { Env, ErrorHandler, Next, NotFoundHandler } from './types'
  * @param {[[Function, unknown], unknown][] | [[Function]][]} middleware - An array of middleware functions and their corresponding parameters.
  * @param {ErrorHandler<E>} [onError] - An optional error handler function.
  * @param {NotFoundHandler<E>} [onNotFound] - An optional not-found handler function.
+ * @param {boolean} [updateRouteIndex] - Whether to update the matched route index.
  *
  * @returns {(context: Context, next?: Next) => Promise<Context>} - A composed middleware function.
  */
 export const compose = <E extends Env = Env>(
   middleware: [[Function, unknown], unknown][] | [[Function]][],
   onError?: ErrorHandler<E>,
-  onNotFound?: NotFoundHandler<E>
+  onNotFound?: NotFoundHandler<E>,
+  updateRouteIndex: boolean = true
 ): ((context: Context, next?: Next) => Promise<Context>) => {
   return (context, next) => {
     let index = -1
@@ -37,31 +39,31 @@ export const compose = <E extends Env = Env>(
 
       let res
       let isError = false
-      let handler
+      const entry = middleware[i]
+      const handler = entry ? entry[0][0] : (i === middleware.length && next) || undefined
 
-      if (middleware[i]) {
-        handler = middleware[i][0][0]
-        context.req.routeIndex = i
-      } else {
-        handler = (i === middleware.length && next) || undefined
+      if (entry) {
+        if (updateRouteIndex) {
+          context.req.routeIndex = i
+        }
       }
 
       if (handler) {
         try {
           res = await handler(context, () => dispatch(i + 1))
         } catch (err) {
-          if (err instanceof Error && onError) {
-            context.error = err
-            res = await onError(err, context)
-            isError = true
-          } else {
+          if (updateRouteIndex) {
+            context.req.routeIndex = i
+          }
+          if (!(err instanceof Error) || !onError) {
             throw err
           }
+          context.error = err
+          res = await onError(err, context)
+          isError = true
         }
-      } else {
-        if (context.finalized === false && onNotFound) {
-          res = await onNotFound(context)
-        }
+      } else if (context.finalized === false && onNotFound) {
+        res = await onNotFound(context)
       }
 
       if (res && (context.finalized === false || isError)) {
--- a/src/helper/dev/index.test.ts
+++ b/src/helper/dev/index.test.ts
@@ -41,6 +41,8 @@ describe('inspectRoutes()', () => {
     const subApp = new Hono()
 
     subApp.get('/', (c) => c.json(0))
+    subApp.catch(async (_c, next) => next())
+    subApp.catchNotFound(async (_c, next) => next())
     subApp.onError((_, c) => c.json(0))
 
     const mainApp = new Hono()
--- a/src/helper/dev/index.ts
+++ b/src/helper/dev/index.ts
@@ -25,15 +25,17 @@ const handlerName = (handler: Function): string => {
 }
 
 export const inspectRoutes = <E extends Env>(hono: Hono<E>): RouteData[] => {
-  return hono.routes.map(({ path, method, handler }: RouterRoute) => {
-    const targetHandler = findTargetHandler(handler)
-    return {
-      path,
-      method,
-      name: handlerName(targetHandler),
-      isMiddleware: isMiddleware(targetHandler),
-    }
-  })
+  return hono.routes
+    .filter(({ method }) => method[0] !== '@')
+    .map(({ path, method, handler }: RouterRoute) => {
+      const targetHandler = findTargetHandler(handler)
+      return {
+        path,
+        method,
+        name: handlerName(targetHandler),
+        isMiddleware: isMiddleware(targetHandler),
+      }
+    })
 }
 
 export const showRoutes = <E extends Env>(hono: Hono<E>, opts?: ShowRoutesOptions): void => {
--- a/src/hono-base.ts
+++ b/src/hono-base.ts
@@ -295,7 +295,7 @@ class Hono<
         ;(handler as any)[COMPOSED_HANDLER] = r.handler
       }
 
-      subApp.#addRoute(r.method, r.path, handler, r.basePath)
+      subApp.#addRoute(r.method, r.path, handler, r)
     })
     return this
   }
@@ -451,14 +451,15 @@ class Hono<
     return this
   }
 
-  #addRoute(method: string, path: string, handler: H, baseRoutePath?: string): void {
+  #addRoute(method: string, path: string, handler: H, baseRoute?: RouterRoute): void {
     method = method.toUpperCase()
     path = mergePath(this._basePath, path)
     const r: RouterRoute = {
-      basePath: mergePath(this._basePath, baseRoutePath ?? '/'),
+      basePath: mergePath(this._basePath, baseRoute?.basePath ?? '/'),
       path,
       method,
       handler,
+      depth: (baseRoute?.depth ?? -1) + 1,
     }
     this.router.add(method, path, [handler, r])
     this.routes.push(r)
@@ -484,6 +485,7 @@ class Hono<
     if (!handlers.length) {
       return fallback(c)
     }
+    handlers.sort((a, b) => b[0][1].depth! - a[0][1].depth!)
     c.finalized = false
 
     const handleError = (err: unknown): Response | Promise<Response> => {
--- a/src/hono-base.ts
+++ b/src/hono-base.ts
@@ -15,15 +15,18 @@ import type {
   FetchEventLike,
   H,
   HandlerInterface,
+  IntersectNonAnyTypes,
   MergePath,
   MergeSchemaPath,
   MiddlewareHandler,
   MiddlewareHandlerInterface,
+  Next,
   NotFoundHandler,
   OnHandlerInterface,
   RouterRoute,
   Schema,
 } from './types'
+import { COMPOSED_HANDLER } from './utils/constants'
 import { getPath, getPathNoStrict, mergePath } from './utils/url'
 
 const METHOD_NAME_NOT_FOUND = '@NOT_FOUND'
@@ -52,6 +55,21 @@ const getResponse = (context: Context): Response => {
 }
 
 type GetPath<E extends Env> = (request: Request, options?: { env?: E['Bindings'] }) => string
+
+type FallbackHandlerInterface<
+  E extends Env,
+  S extends Schema,
+  BasePath extends string,
+  CurrentPath extends string,
+> = {
+  <E2 extends Env = E>(
+    ...handlers: MiddlewareHandler<E2, MergePath<BasePath, '*'>, any, any>[]
+  ): Hono<IntersectNonAnyTypes<[E, E2]>, S, BasePath, CurrentPath>
+  <Path extends string, E2 extends Env = E>(
+    path: Path,
+    ...handlers: MiddlewareHandler<E2, MergePath<BasePath, Path>, any, any>[]
+  ): Hono<IntersectNonAnyTypes<[E, E2]>, S, BasePath, CurrentPath>
+}
 
 export type HonoOptions<E extends Env> = {
   /**
@@ -122,6 +140,42 @@ class Hono<
   on: OnHandlerInterface<E, S, BasePath>
   use: MiddlewareHandlerInterface<E, S, BasePath>
 
+  /**
+   * `.catch()` adds middleware that runs when an error is caught.
+   * If every matching middleware calls `next()`, the error is passed to `.onError()`.
+   *
+   * @param {string} [path] - path to scope the error middleware
+   * @param {...MiddlewareHandler[]} handlers - middleware to run when handling an error
+   * @returns {Hono} changed Hono instance
+   *
+   * @example
+   * ```ts
+   * app.catch('/api/*', async (c, next) => {
+   *   console.error(c.error)
+   *   await next()
+   * })
+   * ```
+   */
+  catch: FallbackHandlerInterface<E, S, BasePath, CurrentPath>
+
+  /**
+   * `.catchNotFound()` adds middleware that runs when a not-found response is requested.
+   * If every matching middleware calls `next()`, the request is passed to `.notFound()`.
+   *
+   * @param {string} [path] - path to scope the not-found middleware
+   * @param {...MiddlewareHandler[]} handlers - middleware to run when handling not found
+   * @returns {Hono} changed Hono instance
+   *
+   * @example
+   * ```ts
+   * app.catchNotFound('/api/*', async (c, next) => {
+   *   c.header('x-not-found', 'true')
+   *   await next()
+   * })
+   * ```
+   */
+  catchNotFound: FallbackHandlerInterface<E, S, BasePath, CurrentPath>
+
   /*
     This class is like an abstract class and does not have a router.
     To use it, inherit the class and implement router in the constructor.
@@ -172,6 +226,11 @@ class Hono<
       return this as any
     }
 
+    this.catch = (...handlers: (string | H)[]) =>
+      this.#addRoutes(METHOD_NAME_ERROR, handlers) as any
+    this.catchNotFound = (...handlers: (string | H)[]) =>
+      this.#addRoutes(METHOD_NAME_NOT_FOUND, handlers) as any
+
     const { strict, ...optionsWithoutStrict } = options
     Object.assign(this, optionsWithoutStrict)
     this.getPath = (strict ?? true) ? (options.getPath ?? getPath) : getPathNoStrict
@@ -182,9 +241,15 @@ class Hono<
       router: this.router,
       getPath: this.getPath,
     })
+    clone.errorHandler = this.errorHandler
+    clone.#notFoundHandler = this.#notFoundHandler
     clone.routes = this.routes
     return clone
   }
+
+  #notFoundHandler: NotFoundHandler = notFoundHandler
+  // Cannot use `#` because it requires visibility at JavaScript runtime.
+  private errorHandler: ErrorHandler = errorHandler
 
   /**
    * `.route()` allows grouping other Hono instance in routes.
@@ -215,7 +280,23 @@ class Hono<
     app: Hono<SubEnv, SubSchema, SubBasePath, SubCurrentPath>
   ): Hono<E, MergeSchemaPath<SubSchema, MergePath<BasePath, SubPath>> | S, BasePath, CurrentPath> {
     const subApp = this.basePath(path)
-    app.routes.map((r) => subApp.#addRoute(r.method, r.path, r.handler, r.basePath))
+    app.routes.map((r) => {
+      let handler
+      if (app.errorHandler === errorHandler || r.method === METHOD_NAME_ERROR) {
+        handler = r.handler
+      } else {
+        handler = async (c: Context<E>, next: Next) => {
+          try {
+            return await r.handler(c, next)
+          } catch (err) {
+            return (c.res = await subApp.#handleError(err, c, app.errorHandler))
+          }
+        }
+        ;(handler as any)[COMPOSED_HANDLER] = r.handler
+      }
+
+      subApp.#addRoute(r.method, r.path, handler, r.basePath)
+    })
     return this
   }
 
@@ -245,8 +326,6 @@ class Hono<
    *
    * @see {@link https://hono.dev/docs/api/hono#error-handling}
    *
-   * @param {string} [path] - path to scope the error handler
-   * @param {...MiddlewareHandler[]} handlers - middleware to run before the error handler
    * @param {ErrorHandler} handler - request handler for error
    * @returns {Hono} changed Hono instance
    *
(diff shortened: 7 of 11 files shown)


usualoma:
Hi @yusukebe,

Could you take a look when you have a chance? This does increase the bundle size, and it is not intended to improve performance. Still, as a redefinition of the semantics of notFound() and onError() in terms of Hono’s routing model, I think it is an interesting and promising direction. What do you think?

yusukebe:
Hi @usualoma 

This is super interesting! Treating them as internal handlers is a good idea. And this change must be included in v5 if we accept this.

Looking at this PR, I thought it's a good idea to add "factory methods" `notFound` and `onError`  to add these handlers like this (let's consider where to place these methods: `hono` or others ) : 

```ts
import { Hono, notFound, onError } from 'hono'

const app = new Hono()

app.use(
  '*',
  onNotFound((e, c) => c.text('custom not found', 404))
)

app.use(
  '*',
  onError((e, c) => c.text('custom error', 500))
)
```

This seems closer to a normal handler, and we don't have to change the mental model between them and normal handlers.

And, if we make them enable distinguishing paths, you can write like this:

```ts
api.get(
  '/items',
  (c) => c.text('items list'),
  onError((e, c) => c.text('list db error', 500))
)
```

This is one I want to do in the next v5. I want to specify error handling per route to avoid writing many `try`/`catch` blocks, so this method would work. What do you think of this?

That's what I imagine after reading this PR. No blockers; the file size increase is acceptable. Let's go ahead. Hi

yusukebe:
Or without the `onError`, it works for me if I can write it like this with the expected behavior:

```ts
const api = new Hono()

api
  .get('/items', (c) => c.text('items list'))
  .onError((e, c) => c.text('list db error', 500))
```

Hmm. This is neater than using the `onError` method.

yusukebe:
One of the opinions https://x.com/schplitt/status/2088560850864882117

yusukebe:
I'm writing this in the train with my smartphone but this is cool

```tsx

import { createOnErrorHandler } from 'hono/factory'

const onRenderError = createOnErrorHandler((e, c) =>
  c.html(<ErrorPage error={e} />, 500)
)
const onApiError = createOnErrorHandler((e, c) =>
  c.json({ error: e.message }, 500)
)

app.get('/', onRenderError, async (c) =>
  c.html(<Home posts={await db.posts()} />)
)
app.get('/posts/:id', onRenderError, async (c) =>
  c.html(<Post post={await db.post(c.req.param('id'))} />)
)
app.get('/api/posts', onApiError, async (c) =>
  c.json(await db.posts())
)
```

usualoma:
@yusukebe 

Thank you for confirming! First, I’ll explain what “scoping” is.

Regarding scoping: the current implementation can scope fallback handlers, but the scope must currently be introduced through a sub-application:

```ts
const items = new Hono()

items.get('/', (c) => c.text('items list'))
items.onError((e, c) => c.text('list db error', 500))

api.route('/items', items)
```

This restriction is primarily for compatibility with the existing API.

This is not how the current PR behaves, but with an additional change to the existing chaining semantics, we could make the following `onError()` call scoped to `/items`:

```ts
api
  .get('/items', (c) => c.text('items list'))
  .onError((e, c) => c.text('list db error', 500))
```

In that alternative design, `onError()` would use the current path in the same way as chained `.get()` or `.post()` calls.

However, that would make the meaning of the existing API ambiguous:

```ts
api.get('/items', (c) => c.text('items list'))

api.onError((e, c) => {
  // ...
})
```

Existing applications can write this today, and the `onError()` call is understood as configuring a global error handler. If `onError()` started using the current path, the same code would unexpectedly register an error handler scoped to `/items`.

For that reason, at least in the current implementation, chaining `onError()` after `get()` does not make it route-scoped.

Since this change is intended for a major version, we could reconsider that behavior if changing this part of the existing semantics is acceptable.

Another option would be to introduce a separate method with explicitly route-local semantics, for example:

```ts
api
  .get('/items', (c) => c.text('items list'))
  .catch((e, c) => c.text('list db error', 500))
```

In that case, having `.catch()` use the current path would be unambiguous.

Alternatively, `onError()` could remain independent of the current path while accepting an explicit path:

```ts
api.onError('/items/*', (e, c) =>
  c.text('list db error', 500)
)
```

That may be a useful extension even if we preserve the current chaining semantics.

usualoma:
Let’s refer to the approach shown in https://github.com/honojs/hono/pull/5229#issuecomment-5301774467 as the “route-local middleware model” here.

I would like to make sure that I understand the intended semantics of this model.

In its simplest form, if we ignore details such as nested handlers, it seems that the current version of Hono could implement something similar without the implementation in this PR by defining the factory function as ordinary middleware:

```ts
const createOnErrorHandler = (
  handler: (error: Error, c: Context) => Response | Promise<Response>
) =>
    await next()

    if (c.error) {
      c.res = await handler(c.error, c)
    }
  })
```

In this version, the route-local handler runs after the existing global `onError`, using the error stored in the `Context`.

If the route-local handler should instead run at the same stage as the current global `onError`, another possible implementation might be to allow middleware to temporarily install a request-local `errorHandler` on the `Context` while executing `next()`.

At its simplest, support for a context-local error handler might look like this:

```diff
diff --git a/src/compose.ts b/src/compose.ts
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -50,9 +50,10 @@ export const compose = <E extends Env = Env>(
         try {
           res = await handler(context, () => dispatch(i + 1))
         } catch (err) {
-          if (err instanceof Error && onError) {
+          const errorHandler = context.onError || onError
+          if (err instanceof Error && errorHandler) {
             context.error = err
-            res = await onError(err, context)
+            res = await errorHandler(err, context)
             isError = true
           } else {
             throw err
diff --git a/src/context.ts b/src/context.ts
--- a/src/context.ts
+++ b/src/context.ts
@@ -2,6 +2,7 @@ import { HonoRequest } from './request'
 import type { Result } from './router'
 import type {
   Env,
+  ErrorHandler,
   FetchEventLike,
   H,
   Input,
@@ -332,6 +333,7 @@ export class Context<
    * ```
    */
   error: Error | undefined
+  onError: ErrorHandler<E> | undefined
 
   #status: StatusCode | undefined
```

The factory function could then install and restore the handler around `next()`:

```ts
const createOnErrorHandler = (
  handler: ErrorHandler
) =>
    const previous = c.onError
    c.onError = handler

    try {
      await next()
    } finally {
      c.onError = previous
    }
  })
```

Both approaches seem conceptually independent of the internal-route implementation in this PR.

Is either of these close to what you have in mind? At this point, I would mainly like to align our understanding of the route-local middleware model before discussing the implementation details.

usualoma:
I added support on this branch for explicitly scoping fallback handlers by path:

```ts
api.onError('/items/*', (e, c) =>
  c.text('list db error', 500)
)
```

The same form is also available for `notFound()`:

```ts
api.notFound('/items/*', (c) =>
  c.text('items not found', 404)
)
```

This is conceptually similar to `use()` in terms of scoping. The fallback is matched by path and does not depend on the HTTP method of the original request.

My current feeling is that the semantics in this PR work well if we assume the following:

- Error handlers and not-found responses seem more useful when they can be configured independently of HTTP methods, conceptually at a level similar to `METHOD_NAME_ALL`.
- In many applications, the natural scope for these handlers will be a sub-application.
- When a finer scope is needed, an explicit path such as `/items/*` should be sufficient.

I think the “route-local middleware model” is a separate, complementary model. If a route-local error handler should apply to every method under a path, it seems that we could write something like this:

```ts
app.use(
  '/items/*',
  createOnErrorHandler((e, c) =>
    c.text('list db error', 500)
  )
)
```

One remaining question is how the route-local middleware model should address #5193.

For the `onError()` part, `createOnErrorHandler()` could accept middleware that is executed only when handling an error:

```ts
app.use(
  '/items/*',
  createOnErrorHandler(
    languageDetector(),
    (e, c) => c.html(renderErrorPage(c.get('language')), 500)
  )
)
```

yusukebe:
Hi @usualoma !

I think making `app.notFound` and `app.onError` follow the API and behavior of `app.use()` is a good idea!

Regarding "route-local", it should be done in a separate PR. I also considered the `createOnErrorHandler()` middleware approach. But after rethinking, the best solution is the `.catch()` you suggested. The naming is easy for users to understand. But we have to guarantee that the inner error handler always wins over the outer one, which needs core support. Let's design that separately.

One concern with accepting this PR: "Registration order determines fallback ordering" will confuse users. For us, "the rule is the same as `app.use()`" is easy, but users used to the current last-wins behavior will be surprised. Especially since a default `onError` registered earlier can no longer be overridden.

That said, `.catch()` should reduce this confusion to some extent. Most cases where users say "my handler must handle this error" will be written with `.catch()`, which is order-independent. `onError()` then becomes just a scoped default response, so the registration order will matter much less in practice.

Hmm. Still, do you think it's worth introducing this in v5, given the confusion?

usualoma:
This is still a PoC, so I am intentionally trying a different API while exploring the semantics.

I added another commit (24e22b3) that restores `onError()` and `notFound()` to their existing behavior. This removes the need to address backward compatibility in those APIs. Instead, the commit introduces two new APIs, `catch()` and `catchNotFound()`, for routable fallback middleware.

With `catch()`, the error is available as `c.error` instead of being passed through the legacy `(error, c)` signature. A terminal handler can therefore be written as `async (c) => Response`, while middleware uses the usual `async (c, next)` signature:

```ts
app.catch(async (c) => {
  return c.text(`Error: ${c.error?.message}`, 500)
})
```

This makes error handling fit more naturally into Hono's existing routing and middleware model.

The fallback APIs are now separated into two models:

| API | Role | Signature | Multiple registrations |
|---|---|---|---|
| `onError()` | Legacy final error handler | `(error, c)` | Last registration wins |
| `notFound()` | Legacy final not-found handler | `(c)` | Last registration wins |
| `catch()` | Routable error middleware | `(c, next)` | Composed in registration order |
| `catchNotFound()` | Routable not-found middleware | `(c, next)` | Composed in registration order |

This restores the existing semantics of `onError()` and `notFound()`. In particular, registering either method multiple times continues to use the last registered handler.

The existing `onError()` API remains available for backward compatibility, but is expected to be deprecated after the new API is introduced.

The new middleware-capable behavior is provided by separate APIs:

```ts
app.catch(
  '/items/*',
  languageDetector(),
  async (c) => {
    return c.html(renderErrorPage(c.get('language'), c.error), 500)
  }
)

app.catchNotFound(
  '/items/*',
  languageDetector(),
  async (c) => {
    return c.html(render404Page(c.get('language')), 404)
  }
)
```

For example, both of the following register the error middleware for `/api/*`:

```ts
const api = new Hono()
  .basePath('/api')
  .catch(async (c) => {
    return c.text(`Error: ${c.error?.message}`, 500)
  })

api.get('/book', (c) => c.text('List Books'))
```

```ts
const api = new Hono().basePath('/api')

api.catch(async (c) => {
  return c.text(`Error: ${c.error?.message}`, 500)
})

api.get('/book', (c) => c.text('List Books'))
```

An explicitly specified path is relative to the current base path:

```ts
const api = new Hono().basePath('/api')

api.catch('/items/*', errorMiddleware)
```

In this case, `errorMiddleware` is registered for `/api/items/*`.

Multiple calls to `catch()` are composed in registration order:

```ts
app.catch(middlewareA, middlewareB)
app.catch(middlewareC)
app.catch(finalErrorHandler)
```

This is executed as:

```text
middlewareA -> middlewareB -> middlewareC -> finalErrorHandler
```

One trade-off is bundle size. Using the same current esbuild command:

- Previous implementation on this branch: `19,254 B`
- Revised implementation: `19,616 B`
- Difference from the previous implementation: `+362 B`
- Difference from the branch base: `+570 B`, approximately `2.99%`

The increase comes mainly from retaining the legacy application-boundary behavior while also adding the routable middleware model.

usualoma:
As for whether we should include this in v5, here are my thoughts.

I don’t think there’s enough motivation to make the introduction of this API a “top priority.” That said, if we were to deprecate some APIs, I think the plan would be to deprecate them in v5 and remove them in v6. Since missing this opportunity would mean waiting quite a while, I think we should consider including them in v5 if possible.

yusukebe:
@usualoma Thanks! The explanation is easy to understand. I'll sort my thoughts and reply later.

yusukebe:
@usualoma 

Thanks again for the proposal! Regarding this feature, let's sleep on it for several days. I plan to release v5 around [HonoConf 2026](https://honoconf.dev/2026) on Oct 11th, so we don't have to rush.

na-trium-144:
I am not deeply familiar with Hono's internal implementation, so please forgive me if I am mistaken, but I would like to share my thoughts after reading the description of this PR.

(The following discussion assumes that we are changing the semantics of `onError` and `notFound`, but the same interpretation would apply to `catch` if it were newly introduced.)

If my understanding is correct, even though `notFound` and `onError` internally follow the same registration-order rules as standard routes, their registration order is effectively separate from that of standard routes.
For instance, in the code below, I believe the execution order would be `middleware1`, `middleware2`, `handler`.

```ts
const app = new Hono();

app.use(middleware1);
app.notFound(handler);
app.use(middleware2);
```

If that is the case, this behavior could be interpreted to mean that 
"the registration of `notFound` and `onError` is deferred until the very end of the `app`."
In other words, regardless of when the `notFound` handler is registered, it becomes functionally equivalent to the following code.

```ts
const app = new Hono();

app.use(middleware1);
app.use(middleware2);

app.notFound(handler);
```

Furthermore, if we extend this interpretation to a rule stating that 
"the registration of `notFound` and `onError` within a sub-app is deferred until the end of that sub-app definition (i.e. the `route()` call)," 
the sub-app's error handler would naturally take precedence over the parent's, thereby resolving the inconsistency with the current `onError` behavior.
Essentially, the following two code snippets would become equivalent.

```ts
const app = new Hono();

app.onError(parentHandler);

const sub = new Hono();
sub.onError(subHandler);
sub.use(() => { throw new Error(); });

app.route("/sub", sub);
```

```ts
const app = new Hono();

const sub = new Hono();

sub.use(() => { throw new Error(); });

sub.onError(subHandler);  // registration deferred until route("/sub", sub)
app.route("/sub", sub);

app.onError(parentHandler);  // registration deferred until the end
```

I believe this offers another alternative approach to "Selecting fallbacks from the originating application" and "Selecting the most specific matching scope."
That said, I am unsure how much this would complicate the internal codebase compared to those.

usualoma:
Hi @na-trium-144,
Thank you for the comment. This was very helpful.

While exploring several approaches, I had ended up going in the wrong direction. What I actually wanted for the evaluation order was to prefer the sub-app, exactly as you pointed out.

I have updated the implementation accordingly.

Each route now records its application nesting depth, which is incremented when the route is imported through `route()`. Matching `catch()` and `catchNotFound()` routes are composed in descending depth order, while registration order is preserved among routes at the same depth.

For example:

```ts
app.catch(parentHandler)

sub.catch(subHandler)
app.route('/sub', sub)
```

For a request under `/sub`, `subHandler` is called first. If it calls `next()`, execution continues to `parentHandler`.

This provides the effective ordering described in your deferred-registration model without actually deferring route registration. The priority is based on application nesting rather than path specificity, and normal route dispatch is not reordered.

usualoma:
I updated the PR description.
In the current version, the increase in bundle size is a bit concerning.

usualoma:
In [1e867e7](https://github.com/honojs/hono/pull/5229/commits/1e867e75a668ec34faea4151b06225617eadb3e0), I added the application depth to the routing metadata.

This is useful for ordering fallback middleware in this PR, but it may also be useful for route introspection in general. For example, `showRoutes()` could use the depth together with `basePath` to display nested routing scopes:

```ts
const admin = new Hono().basePath('/admin').get('/users', (c) => c.text('users'))
const api = new Hono().basePath('/api').get('/books', (c) => c.text('books')).route('/', admin)
const app = new Hono().get('/', (c) => c.text('home')).route('/', api)

showRoutes(app)
```

```text
GET  /
/api
├── GET  /books
└── /admin
    └── GET  /users
```

The depth alone does not preserve the exact identity of each sub-application, so this would be a routing-scope view rather than a complete application tree. Still, I think the number of `route()` boundaries crossed is potentially useful route metadata beyond this PR.

usualoma:
For reference, I also revisited the direct `compose()`-based approach on top of the current Hono implementation.
https://github.com/honojs/hono/pull/5286

The implementation in #5211 would need a little more handling to make it complete. In particular, the fallback chain must be able to replace an already finalized response, and the original `routeIndex` must be preserved across the middleware chain and nested `next()` calls so that APIs such as `c.req.param()` continue to refer to the originating route.

A prototype including those details increases the minified bundle by about 602 B. This PR increases it by 636 B, so the difference is only 34 B.

Therefore, if we consider bundle size alone, both approaches are realistic options.

usualoma:
Since this is a PoC, to broaden the discussion a little, an approach like 07bb8dae could also allow `showRoutes()` to expose additional information, such as registered error and not-found handlers, on demand.

For example:

```ts
const app = new Hono()
  .get('/', home)
  .get('/health', healthCheck)

const api = new Hono()
  .get('/users', listUsers)
  .get('/users/:id', getUser)
  .post('/users', createUser)
  .catch(handleApiError)
  .catchNotFound(handleApiNotFound)

app
  .route('/api', api)
  .catch(handleAppError)
  .catchNotFound(handleAppNotFound)

showRoutes(app, {
  includeInternal: true,
  verbose: true,
  colorize: false,
})
```

This produces:

```text
GET         /
              home
GET         /health
              healthCheck
GET         /api/users
              listUsers
GET         /api/users/:id
              getUser
POST        /api/users
              createUser
@ERROR      /api/*
              handleApiError
@NOT_FOUND  /api/*
              handleApiNotFound
@ERROR      /*
              handleAppError
@NOT_FOUND  /*
              handleAppNotFound
```

PR #5231, opened 2026-08-14
fix(router): keep the prefix of a trailing wildcard

Kjubikstronk:
> **Stacked on #5230.** The first commit here is that PR; review only the second, `fix(router): keep the prefix of a trailing wildcard`. If #5230 lands first this rebases to a single commit, and I'll rebase on request either way.

Found while adding the shared router cases in #5230 — the new case had to be skipped for these two routers, and this removes those skips.

## What was happening

Both routers dropped **two** characters when a route ended in `*`:

```ts
// linear-router
const parts = (endsWithStar ? routePath.slice(0, -2) : routePath).split(splitByStarRe)

// pattern-router
if (endsWithWildcard) {
  path = path.slice(0, -2)
}
```

That's correct for `/a/b/*`, where the two characters are `/` and `*`. For `/a/b*` it also eats the `b`, leaving the prefix `/a/`. Both routers then match anything below `/a/`:

```ts
const router = new LinearRouter()
router.add('GET', '/a/b*', 'handler')
router.match('GET', '/a/c')    // matched
router.match('GET', '/a/zzz')  // matched
```

PatternRouter degrades further, because the collapsed prefix becomes an unanchored regex: `/a/b*` compiles to `^/a` and matches `/assets`, `/assetsfoo`, `/assets/app.js`. `/assets*` compiles to `^/asset` in both routers, so it matches `/asset`.

## The change

Slice one character, so the prefix keeps its last character. Two consequences follow:

- The prefix for `/a/b/*` now keeps its trailing slash, so LinearRouter needs an explicit check for the form matching `/a/b` itself. RegExpRouter and TrieRouter both match it, so this preserves that.
- PatternRouter's regex for `/assets/*` is now anchored with `(?:/|$)`. That also fixes it matching `/assetsfoo`, which RegExpRouter and TrieRouter reject.

## Verification

Cross-router matrix over 5 route shapes × 14 paths, comparing all four routers:

- **before:** 12 of 70 divergent
- **after:** 2 of 70

Both remaining divergences are the mid-path form `/x*/y`, which RegExpRouter supports and TrieRouter and PatternRouter do not, and where LinearRouter matches `/x/y` when RegExpRouter does not. That's the same non-trailing-wildcard gap I left alone in #5230, and it's untouched here.

Specifically fixed:

| Route | Path | before | after (all four agree) |
| --- | --- | --- | --- |
| `/a/b*` | `/a/c`, `/a/`, `/a/zzz` | Linear + Pattern matched | no match |
| `/a/b*` | `/assets`, `/assetsfoo`, `/assets/app.js`, `/asset` | Pattern matched | no match |
| `/assets*` | `/asset` | Linear + Pattern matched | no match |
| `/assets/*` | `/assetsfoo` | Linear + Pattern matched | no match |

`pnpm vitest --project main`: 122 files, 3808 passed, 28 skipped, 0 failures — against 3800 passed / 30 skipped on #5230, the difference being the shared case now running for these two routers plus the added slash-wildcard case. `tsc -p tsconfig.spec.json`, ESLint and Prettier are clean.

## Tests

The `Suffix wildcard > Matches below the root` case from #5230 now runs for all four routers, and its two `skip` entries are removed. Added `Does not run a slash wildcard past its segment` to the same shared block, covering `/assets/*` against `/assets`, `/assets/app.js` and `/assetsfoo` — all four routers agree on it.

### DIFF
--- a/src/router/common.case.test.ts
+++ b/src/router/common.case.test.ts
@@ -122,6 +122,14 @@ export const runTest = ({
         expect(match('GET', '/static/app.js').length).toBe(0)
       })
 
+      it('Does not run a slash wildcard past its segment', async () => {
+        router.add('GET', '/assets/*', 'get assets')
+
+        expect(match('GET', '/assets').length).toBe(1)
+        expect(match('GET', '/assets/app.js').length).toBe(1)
+        expect(match('GET', '/assetsfoo').length).toBe(0)
+      })
+
       it('Matches below the root', async () => {
         router.add('GET', '/a/b*', 'get b')
 
--- a/src/router/common.case.test.ts
+++ b/src/router/common.case.test.ts
@@ -106,6 +106,35 @@ export const runTest = ({
       })
     })
 
+    describe('Suffix wildcard', () => {
+      it('Matches the rest of the path', async () => {
+        router.add('GET', '/assets*', 'get assets')
+
+        for (const path of ['/assets', '/assets/app.js', '/assets/nested/app.js']) {
+          const res = match('GET', path)
+          expect(res.length).toBe(1)
+          expect(res[0].handler).toEqual('get assets')
+        }
+      })
+
+      it('Does not match another prefix', async () => {
+        router.add('GET', '/assets*', 'get assets')
+        expect(match('GET', '/static/app.js').length).toBe(0)
+      })
+
+      it('Matches below the root', async () => {
+        router.add('GET', '/a/b*', 'get b')
+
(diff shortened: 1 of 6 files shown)


Kjubikstronk:
Closing this — #5252 landed the same fix, so there is nothing left here.

For anyone finding this later: the cause was that both routers dropped two characters when a route ended in `*`, which took the last character of the segment along with the star, so `/a/b*` became the prefix `/a/` and matched everything below it.

The mid-path form is still open in #5243, which I will rebase onto current `main`.

PR #5232, opened 2026-08-14
fix(method-not-allowed): ignore trailing wildcard routes when inferring Allow

### LINKED ISSUE #5223: When serveStatic and methodNotAllowed are used together, methods other than GET cannot be handled correctly

DSeaStar:
## Summary

Fixes #5223.

`methodNotAllowed` currently treats a trailing-wildcard route such as `GET /*` as evidence that every path exists. That turns `POST /missing` into 405 instead of 404, which is the `serveStatic` + `methodNotAllowed` case reported in the issue.

A maintainer asked to ignore routes whose paths end in `*` when inferring allowed methods: a trailing wildcard matches an arbitrary remainder, so it does not show that a particular resource exists.

This change skips those routes when building the Allow map. Concrete paths such as `GET /api` still produce 405 for other methods. GET/HEAD still reach the wildcard handler itself.

## Test plan

- [x] Added a regression test for `GET /api` + `GET /*` + `GET /files*`: POST `/api` is 405, POST `/missing` and POST `/files/readme` stay 404, GET `/missing` still hits the wildcard.
- [x] `bunx vitest --run src/middleware/method-not-allowed/index.test.ts` — 21 passed

### DIFF
--- a/src/middleware/method-not-allowed/index.test.ts
+++ b/src/middleware/method-not-allowed/index.test.ts
@@ -197,6 +197,30 @@ describe('Method Not Allowed Middleware', () => {
     expect(res.headers.has('Allow')).toBe(false)
   })
 
+  it('ignores trailing wildcard routes when collecting allowed methods', async () => {
+    const app = new Hono()
+    app.use(methodNotAllowed({ app }))
+    app.get('/api', (c) => c.text('ok'))
+    app.get('/*', (c) => c.text('file'))
+    app.get('/files*', (c) => c.text('prefix'))
+
+    const postApi = await app.request('/api', { method: 'POST' })
+    expect(postApi.status).toBe(405)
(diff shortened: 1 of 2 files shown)


usualoma:
Sorry, but I will close all PRs created before the discussion in #5223 has concluded without merging them. Please bear with me.

PR #5233, opened 2026-08-14
feat(types): add Cloudflare Access context

bndnsmth:
## Summary

- add the optional Cloudflare Access context to Hono's `ExecutionContext`
- mirror Cloudflare's structural Access context and identity types without adding a runtime-specific dependency

## Problem

Cloudflare Workers now exposes an optional, readonly `access` property on the execution context for requests authenticated by Cloudflare Access. Hono's local `ExecutionContext` interface predates that property, so applications cannot access `c.executionCtx.access` without a cast or module augmentation even when the runtime provides it.

The Access context is not application-defined: Cloudflare specifies its audience value and `getIdentity()` result. Typing it as `any` would discard useful platform guarantees, while directly referencing Cloudflare's ambient generated types would make Hono's core declarations depend on those types being installed.

## Implementation

This change mirrors Cloudflare's canonical structural definitions as `CloudflareAccessContext` and `CloudflareAccessIdentity`. The identity includes the documented common fields and extends `Record<string, unknown>` for identity-provider-specific data. `ExecutionContext.access` is optional because the runtime only supplies it for applicable requests. The supporting types are not re-exported from the main `hono` entry point.

Canonical definitions: https://github.com/cloudflare/workerd/blob/main/types/defines/access.d.ts

## Validation

- `bun run test`
- `bun run lint`
- `bun run format`
- `bun run build`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -37,7 +37,13 @@ export type {
  * Types for context, context variable map, context renderer, and execution context.
  */
 export { Context } from './context'
-export type { ContextVariableMap, ContextRenderer, ExecutionContext } from './context'
+export type {
+  CloudflareAccessContext,
+  CloudflareAccessIdentity,
+  ContextVariableMap,
+  ContextRenderer,
+  ExecutionContext,
+} from './context'
 /**
  * Type for HonoRequest.
  */
--- a/src/index.ts
+++ b/src/index.ts
@@ -37,13 +37,7 @@ export type {
  * Types for context, context variable map, context renderer, and execution context.
  */
 export { Context } from './context'
-export type {
-  CloudflareAccessContext,
-  CloudflareAccessIdentity,
-  ContextVariableMap,
-  ContextRenderer,
-  ExecutionContext,
-} from './context'
+export type { ContextVariableMap, ContextRenderer, ExecutionContext } from './context'
 /**
  * Type for HonoRequest.
  */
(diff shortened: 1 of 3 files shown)


yusukebe:
We must not add platform-specific code to `context.ts`. So I can't accept this PR now. But we can consider how to handle the `access` property.

src/index.ts
```diff
@@ -37,7 +37,13 @@ export type {
  * Types for context, context variable map, context renderer, and execution context.
  */
 export { Context } from './context'
-export type { ContextVariableMap, ContextRenderer, ExecutionContext } from './context'
+export type {
+  CloudflareAccessContext,
```
yusukebe: What is the reason you export `CloudflareAccessContext` and `CloudflareAccessIdentity`?

bndnsmth: They could be removed to avoid the platform-specific leak.  

It's easy enough to augment ExecutionContext type locally in an app with access.  However, since it's a platform-specific context, a strategy to type it further would be nice.

bndnsmth: Removed the two root re-exports in 3a8f2637. The supporting types now stay out of the main `hono` entry point.

yusukebe (approved):
LGTM!

yusukebe:
Hey @brandon93s 

For this feature, I was wondering whether we should introduce it, since it's a platform-specific feature in `context.ts`. But after rethinking, I decided to approve it. The change is mainly types, and it doesn't affect other platforms.

I'll merge this later and include it in a patch release or the next minor or major release. This isn't a big feature, so it might be OK to release it as a patch.

Thanks!

PR #5234, opened 2026-08-14
fix(etag): avoid skipping headers when filtering 304 response headers

spellsaif:
Fixes an issue in the `etag` middleware where non-retained headers could be skipped during 304 response filtering due to mutating the `Headers` instance during `forEach` iteration.

 ### Problem
In `src/middleware/etag/index.ts`, non-retained headers are deleted inside `c.res.headers.forEach`:

```ts
    c.res.headers.forEach((_, key) => {
      if (retainedHeaders.indexOf(key.toLowerCase()) === -1) {
        c.res.headers.delete(key)
      }
    })
```

Under standard Headers (and Map) iterator semantics, calling .delete() on the underlying collection during .forEach() shifts internal iterator indices, causing the iterator to skip subsequent header entries.

### Solution

Collect the keys to delete in a temporary array first, and then delete them in a separate loop so the collection being iterated is not mutated in-flight.

- [X] Add tests
- [X] Run tests

### DIFF
--- a/src/middleware/etag/index.test.ts
+++ b/src/middleware/etag/index.test.ts
@@ -441,4 +441,33 @@ describe('Etag Middleware', () => {
       expect(res.headers.get('ETag')).toBeNull()
     })
   })
+
+  it('Should remove all non-retained headers on 304 without skipping any', async () => {
+    const app = new Hono()
+    app.use('/etag/*', etag())
+    app.get('/etag/multi-headers', (c) => {
+      c.header('X-Custom-1', 'val1')
+      c.header('X-Custom-2', 'val2')
+      c.header('X-Custom-3', 'val3')
+      c.header('X-Custom-4', 'val4')
+      c.header('Cache-Control', 'max-age=3600')
+      return c.text('Hono is hot')
+    })
+    const res1 = await app.request('http://localhost/etag/multi-headers')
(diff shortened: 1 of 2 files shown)


usualoma:
Hi @spellsaif,

Thanks for opening this PR. Nice catch! The proposed fix is functionally correct as written, and I don’t see any behavioral issue with it.

That said, how about using `Array.from()` to snapshot the header keys before deleting them? This avoids manually building `headersToDelete` and keeps the code shorter.

Both approaches allocate a temporary array. The performance difference may depend on the proportion of headers being deleted, but with realistic header counts, I expect it to be too small to measure in practice.

```ts
      for (const key of Array.from(c.res.headers.keys())) {
        if (retainedHeaders.indexOf(key.toLowerCase()) === -1) {
          c.res.headers.delete(key)
        }
      }
```

spellsaif:
@usualoma Hi! Thanks for the review and suggestion. :)

I've updated the implementation to use `Array.from(c.res.headers.keys())` as suggested. It keeps the code much cleaner and concise. All tests are passing green.

usualoma:
Thank you 👍

yusukebe (approved):
LGTM!

yusukebe:
@spellsaif @usualoma 

Ah, this was actually a bug. Thanks!

PR #5236, opened 2026-08-15
fix(trie-router): match suffix wildcard routes

### LINKED ISSUE #5219: Suffix wildcard routes (/foo*) return 404 when a sub-app has static + :param routes with the same method
## Version

hono 4.12.29 and 4.13.1 (latest)

## Expected behavior

A suffix wildcard route like `/assets*` should match `/assets/app.js` regardless of what else is registered on the app.

## Actual behavior

When a mounted sub-app registers a static route and a `:param` route as siblings (same method), every suffix wildcard route (`/assets*`) on the parent app silently returns 404. Slash-star (`/assets/*`) is unaffected.

## Minimal reproduction

```ts
import { Hono } from 'hono'

// control: works without the sub-app
const control = new Hono()
control.get('/assets*', (c) => c.text('ok'))
console.log((await control.request('http://x/assets/app.js')).status) // 200

// bug: same wildcard, but sub-app has static + :param siblings
const app = new Hono()
const sub = new Hono()
sub.post('/items', (c) => c.text('items'))
sub.post('/:slug', (c) => c.text('slug'))
app.route('/api', sub)
app.get('/assets*', (c) => c.text('ok'))
console.log((await app.request('http://x/assets/app.js')).status) // 404
```

## What I tried

- Removing either `/items` or `/:slug` makes it work again
- Different methods (`get('/items')` + `post('/:slug')`) is fine
- `{slug}` instead of `:slug` also fixes it
- Confirmed on 4.12.29 and 4.13.1

## Related issues

#4975, #4623 — same router-fallback family, but those cover `/**` with param-vs-param conflicts; this is suffix-`*` with a static-vs-param conflict.

usualoma:
Fixes #5219

This PR makes TrieRouter recognize a terminal segment ending in `*` as a suffix wildcard. The prefix is matched literally, so regular expression characters have no special meaning.

There are already two PRs addressing this issue (#5220 and #5230), but they either have a larger implementation footprint or do not cover all relevant cases. This approach is more compact and keeps the fix narrowly scoped to TrieRouter.

It slightly increases the bundle size, but planned follow-up PRs should more than offset that increase.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/hono.test.ts
+++ b/src/hono.test.ts
@@ -473,6 +473,20 @@ describe('Routing', () => {
     expect(await res.text()).toBe('get /add-path-after-route-call')
   })
 
+  it('Should match a suffix wildcard after falling back to TrieRouter', async () => {
+    const app = new Hono()
+    const sub = new Hono()
+    sub.post('/items', (c) => c.text('items'))
+    sub.post('/:slug', (c) => c.text('slug'))
+    app.route('/api', sub)
+    app.get('/assets*', (c) => c.text('asset'))
+
+    const res = await app.request('http://localhost/assets/app.js')
+    expect(res.status).toBe(200)
+    expect(await res.text()).toBe('asset')
+    expect(app.router.name).toBe('SmartRouter + TrieRouter')
+  })
+
   it('Nested route - subApp with basePath', async () => {
     const app = new Hono()
     const book = new Hono().basePath('/book')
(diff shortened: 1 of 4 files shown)


usualoma:
Hi @yusukebe,

I've merged the following two pull requests here.
* https://github.com/honojs/hono/pull/5237
* https://github.com/honojs/hono/pull/5238

Please review these changes!

yusukebe (approved):
LGTM!

yusukebe:
CleanShot 2026-08-17 at 09 37 13@2x

@usualoma Woooow great! Thanks!

PR #5238, opened 2026-08-15
refactor(trie-router): trade registration speed for bundle size

usualoma:
This is a follow-up to the previous TrieRouter bundle-size reduction.

> [!IMPORTANT]
> This PR intentionally trades approximately 5.5% slower route registration for a smaller bundle. Matching benchmarks showed no meaningful performance regression.

It further reduces the bundle size by storing pattern metadata on child nodes and simplifying the route-registration path.

| TrieRouter bundle | Before | After | Difference |
| --- | ---: | ---: | ---: |
| Minified | 3,064 B | 2,793 B | -271 B |
| Minified + gzip | 1,533 B | 1,458 B | -75 B |

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/router/trie-router/node.ts
+++ b/src/router/trie-router/node.ts
@@ -20,7 +20,8 @@ export class Node<T> {
   #methods: Record<string, HandlerSet<T>>[] = []
 
   #children: Record<string, Node<T>> = Object.create(null)
-  #patterns: (Pattern | string)[] = []
+  #patterns: Node<T>[] = []
+  #pattern?: Pattern | string
   #params: Record<string, string> = emptyParams
 
   insert(method: string, path: string, handler: T): void {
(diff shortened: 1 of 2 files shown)


usualoma:
Hi @yusukebe,

This change makes route registration approximately 5.5% slower, while our benchmarks show no regression in matching performance. For TrieRouter specifically, I think the 271-byte reduction is a worthwhile trade-off for Hono. Do you think this trade-off is acceptable?

yusukebe:
@usualoma 

The code I wrote a long time ago is being cleaned up, haha!

> For TrieRouter specifically, I think the 271-byte reduction is a worthwhile trade-off for Hono. Do you think this trade-off is acceptable?

Yes. File size reduction is the No.1 priority for TrieRouter.

Can you merge those PRs into #5236?

PR #5239, opened 2026-08-15
fix(etag): copy pending stream bytes

usualoma:
Fix ETag digest corruption when a `ReadableStream` producer reuses its backing buffer.

`generateDigest()` retained a `subarray()` across `reader.read()` calls. Because `subarray()` shares the producer-owned backing buffer, a subsequent pull could overwrite bytes that had not yet been digested. For example, a stream producing `[A, B]` with one reusable buffer could be hashed as `[B, B]`.

This regression was introduced in https://github.com/honojs/hono/pull/5205.

Pending bytes are now copied with `slice()` before the next read. This adds a small, bounded overhead—at most one additional allocation and a copy of less than 256 KiB per stream—but its impact should be negligible in practice. Regression coverage also verifies reusable buffers, patterned bodies across different chunk boundaries, and regular SHA-1 digests for bodies up to 256 KiB.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/etag/digest.test.ts
+++ b/src/middleware/etag/digest.test.ts
@@ -1,0 +1,21 @@
+import { generateDigest } from './digest'
+
+describe('generateDigest', () => {
+  it('Should copy pending bytes from a stream that reuses its buffer', async () => {
+    const scratch = new Uint8Array(4)
+    let pullCount = 0
+    const stream = new ReadableStream({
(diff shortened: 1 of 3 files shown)


yusukebe (approved):
LGTM!

yusukebe:
I see! Thanks @usualoma !

PR #5246, opened 2026-08-16
fix(cookie): allow parsing signed cookies with empty string values

spellsaif:
Fixes an issue in `parseSigned` where valid signed cookies with empty string values (`""`) were dropped during parsing.

### Problem

When serializing a signed cookie with an empty string value via `serializeSigned('key', '', secret)` (or `setSignedCookie(c, 'key', '', secret)`), the resulting cookie value is formatted as `.${signature}`.

In `src/utils/cookie.ts`:
```ts
    const signatureStartPos = value.lastIndexOf('.')
    if (signatureStartPos < 1) {
      continue
    }
```

When value is .${signature}, value.lastIndexOf('.') is 0. Because signatureStartPos < 1 evaluates to true, parseSigned assumes the cookie is malformed and drops it, returning an empty object (and causing getSignedCookie to return undefined).

### Solution

Change the condition from signatureStartPos < 1 to signatureStartPos < 0 so that signed cookies with a signature starting at index 0 (empty signed value "") are properly verified and parsed.

- [X] Add tests
- [X] Run tests

### DIFF
--- a/src/helper/cookie/index.test.ts
+++ b/src/helper/cookie/index.test.ts
@@ -67,8 +67,8 @@ describe('Cookie Middleware', () => {
       const res = new Response('Signed fortune cookie')
       if (typeof fortuneCookie !== 'undefined' && typeof fruitCookie !== 'undefined') {
         // just examples for tests sake
-        res.headers.set('Fortune-Cookie', fortuneCookie || 'INVALID')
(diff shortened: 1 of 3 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@spellsaif 

Ah I see! Thanks.

PR #5247, opened 2026-08-16
fix(utils/ipaddr): avoid truncation on embedded IPv4 addresses in expand IPv6

spellsaif:
## Problem

Fixes an issue in `expandIPv6` where IPv6 addresses with embedded IPv4 addresses in the `0.0.x.x` range (where the first two octets are `0`) were truncated into invalid 7-group IPv6 strings.

In `src/utils/ipaddr.ts`, the IPv4 segment was expanded via:

```ts
convertIPv6BinaryToString(convertIPv4ToBinary(sections.at(-1) as string))
  .substring(2)
  .split(':')
```

For IPv4 addresses such as `0.0.0.1`, `convertIPv6BinaryToString(1n)` compresses the leading zero groups to `::1`. Applying `.substring(2)` and splitting the result produces only a single group:

```ts
['1']
```

instead of the required two 16-bit groups:

```ts
['0', '1']
```

As a result, `sections.splice()` inserts only one group instead of two, producing an invalid 7-group IPv6 string.

For example:

```text
2001:db8:1:2:3:4:0.0.0.1
```

was incorrectly expanded to:

```text
2001:0db8:0001:0002:0003:0004:0001
```

## Solution

Directly convert the four IPv4 octets `a.b.c.d` into two 16-bit hexadecimal groups:

```ts
((a << 8) | b).toString(16)
((c << 8) | d).toString(16)
```

This guarantees that exactly two IPv6 groups are inserted, regardless of leading zeros in the IPv4 address.

## Verification

* [x] Add tests
* [x] Run tests
* [x] Run `bun run format:fix && bun run lint:fix` to format and lint the code
* [ ] Add [[TSDoc](https://tsdoc.org/)](https://tsdoc.org/)/[[JSDoc](https://jsdoc.app/about-getting-started)](https://jsdoc.app/about-getting-started) to document the code

### DIFF
--- a/src/utils/ipaddr.test.ts
+++ b/src/utils/ipaddr.test.ts
@@ -27,6 +27,9 @@ describe('expandIPv6', () => {
     expect(expandIPv6('2001:2::')).toBe('2001:0002:0000:0000:0000:0000:0000:0000')
     expect(expandIPv6('2001:0:0:db8::1')).toBe('2001:0000:0000:0db8:0000:0000:0000:0001')
     expect(expandIPv6('::ffff:127.0.0.1')).toBe('0000:0000:0000:0000:0000:ffff:7f00:0001')
+    expect(expandIPv6('::ffff:0.0.0.1')).toBe('0000:0000:0000:0000:0000:ffff:0000:0001')
+    expect(expandIPv6('2001:db8:1:2:3:4:0.0.0.1')).toBe('2001:0db8:0001:0002:0003:0004:0000:0001')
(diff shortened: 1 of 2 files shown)


usualoma:
👍

spellsaif:
Ohayo @usualoma! Do let me know if there is better way to do it 🙃

yusukebe (approved):
LGTM!

yusukebe:
@spellsaif Thanks!

PR #5250, opened 2026-08-16
fix(csrf): exempt OPTIONS request from CSRF validation

Jr-14:

- [x] Add tests
- [x] Run tests

[RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#safe.methods) defines `OPTIONS` as a safe method, and CORS uses it for preflight requests.
> Of the request methods defined by this specification, the [GET](https://www.rfc-editor.org/rfc/rfc9110.html#GET), [HEAD](https://www.rfc-editor.org/rfc/rfc9110.html#HEAD), [OPTIONS](https://www.rfc-editor.org/rfc/rfc9110.html#OPTIONS), and [TRACE](https://www.rfc-editor.org/rfc/rfc9110.html#TRACE) methods are defined to be safe.

### DIFF
--- a/src/middleware/csrf/index.test.ts
+++ b/src/middleware/csrf/index.test.ts
@@ -1,5 +1,6 @@
 import type { Context } from '../../context'
 import { Hono } from '../../hono'
+import { cors } from '../../middleware/cors'
 import { csrf } from '../../middleware/csrf'
 
 const simplePostHandler = vi.fn(async (c: Context) => {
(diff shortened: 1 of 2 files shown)


Jr-14:
I retract back my misunderstanding - the order for CSRF and CORS should matter https://docs.gofiber.io/blog/fiber-v3-security-middleware-stack/#why-order-matters. Whether we keep this PR to add `OPTIONS` as a safe method for CSRF, and downgrade the regression tests, happy to close out/refactor.

Otherwise, probably another documentation opportunity similar to above 👍🏼 

I learnt a lot opening this PR too 🙇🏼

usualoma:
Hi @Jr-14,

Thanks for opening this PR! I think keeping the change to treat `OPTIONS` as a safe method makes sense. Middleware ordering matters in general, but independently of that, `OPTIONS` should not be subject to CSRF validation. A focused unit test for the `OPTIONS` behavior should be sufficient without adding a dependency on the CORS middleware.

Jr-14:
Hi @usualoma Thanks so much for your review 🙇🏼

I've made the changes [here](https://github.com/honojs/hono/pull/5250/changes/2505a1f7e4180a9298bcd33283e0c8fe9eb77f36)

usualoma:
Thank you!

yusukebe (approved):
LGTM!

yusukebe:
@Jr-14 Nice. Thanks!

Jr-14:
Thank you both @usualoma @yusukebe  🙇

PR #5252, opened 2026-08-17
fix(pattern-router/linear-router): prevent prefix overmatch on wildcard routes

yusukebe:
I found that `/assets*` matches `/asset` and `/path/*` matches `/pathfoo` in PatternRouter. LinearRouter has the same bug. This PR will fix both.

The tests are added to `common.case.test.ts`, so all routers are now verified to behave the same for suffix and trailing wildcards.

Related to #5219

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/router/common.case.test.ts
+++ b/src/router/common.case.test.ts
@@ -370,6 +370,58 @@ export const runTest = ({
       })
     })
 
+    describe('Suffix wildcard', () => {
+      beforeEach(() => {
+        router.add('GET', '/assets*', 'assets')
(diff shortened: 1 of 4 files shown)


yusukebe:
Hey @usualoma 

Sorry for pinging you again. Can you review this when you have time?

usualoma:
Hi @yusukebe,

Thanks for the fix 👍  This is definitely a bug.

yusukebe:
@usualoma Thanks!

PR #5255, opened 2026-08-17
fix(accepts): support wildcard media types and specificity ordering in defaultMatch

adityajha77:
## Description

Fixes an issue in the `accepts` helper where `defaultMatch` failed to match wildcard media types (`*/*` and `type/*`, e.g., `application/*`) sent by clients in `Accept` headers against server-supported MIME types.

### Problem
`defaultMatch` previously used a strict `supports.includes(accept.type)` check. When clients sent standard browser wildcard headers like `Accept: application/*` or `Accept: */*;q=0.8`, `defaultMatch` failed to match supported MIME types like `application/json`, inappropriately falling back to the `default` MIME type.

### Solution
1. Added `matchType` helper function to handle exact matches, subtype wildcards (`application/*`), and global wildcards (`*/*`).
2. Added `getSpecificity` ordering so exact MIME matches take precedence over wildcard matches at the same quality factor (`q`) level per RFC 9110 HTTP Content Negotiation semantics.
3. Updated `defaultMatch` to sort preferences by quality factor (`q`) and specificity before resolving against supported server media types.

## Checklist
- [x] Added unit tests for `application/*`, `*/*`, and specificity ordering.
- [x] All unit tests pass (`npx vitest run src/helper/accepts/accepts.test.ts`).

### DIFF
--- a/src/helper/accepts/accepts.test.ts
+++ b/src/helper/accepts/accepts.test.ts
@@ -134,6 +134,22 @@ describe('accepts', () => {
     expect(result).toBe('application/json')
   })
 
+  test('should prefer subtype wildcard over global wildcard and exact match over subtype wildcard at same quality factor', () => {
+    const c = {
+      req: {
+        header: () => '*/*, text/*',
+      },
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } as any
+    const options: acceptsConfig = {
+      header: 'Accept',
+      supports: ['application/json', 'text/plain'],
+      default: 'application/json',
+    }
+    const result = accepts(c, options)
+    expect(result).toBe('text/plain')
+  })
+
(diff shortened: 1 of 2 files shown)


yusukebe:
@adityajha77 

Thank you for the PR! Supporting `application/*` is a nice fix.

But matching `*/*` changes the behavior for almost every browser request (they always send `*/*;q=0.8`), and the default would never be returned. So pushed a commit that keeps `*/*` and `*` falling through to default as before. Only subtype wildcards like `application/*` newly match.

yusukebe (approved):
LGTM!

adityajha77:
@yusukebe  thanks for the clarification and the fix and merging the PR. I’ll keep this in mind and make sure to avoid this mistake in future contributions.

PR #5256, opened 2026-08-17
fix(client): support custom buildSearchParams and filter undefined query in $ws()

adityajha77:
## Description

Fixes query parameter processing in `hc().$ws()` so that custom `buildSearchParams` functions are respected and `undefined` query parameter values are omitted.

### Problem
When calling `$ws()` on an `hc` client:
1. Any custom `buildSearchParams` function passed in `options.buildSearchParams` was ignored because `$ws()` manually iterated over `opts.args[0].query` using standard `URLSearchParams.set/append`.
2. `undefined` query parameter values were serialized as literal `"undefined"` strings (e.g., `?foo=undefined`), whereas `buildSearchParams` filters out `undefined` keys for HTTP endpoints.

### Solution
1. Updated `hc()`'s `proxyCallback` for `method === 'ws'` to route `opts.args[0].query` through `buildSearchParamsOption`.
2. Appended search parameter entries generated by `buildSearchParamsOption` to the WebSocket target URL, preserving custom search parameter encoders and filtering out `undefined` keys.

## Checklist
- [x] Added unit tests for `$ws()` with custom `buildSearchParams` and `undefined` query stripping in `src/client/client.test.ts`.
- [x] All unit tests pass (`npx vitest run src/client/client.test.ts`).

### DIFF
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -1787,6 +1787,13 @@ describe('Custom buildSearchParams', () => {
     expect(url.href).toBe('http://localhost/search?q=test&tags%5B%5D=tag1&tags%5B%5D=tag2')
   })
 
+  it('Should use default buildSearchParams in $url() when custom one is not provided', () => {
+    const client = hc<AppType>('http://localhost')
+    const url = client.search.$url({ query: { q: 'test', tags: ['tag1', 'tag2'] } })
+
+    expect(url.href).toBe('http://localhost/search?q=test&tags=tag1&tags=tag2')
+  })
+
   it('Should use custom buildSearchParams in $ws() method', () => {
     const webSocketMock = vi.fn()
     const client = hc<AppType>('http://localhost', {
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -1787,11 +1787,34 @@ describe('Custom buildSearchParams', () => {
     expect(url.href).toBe('http://localhost/search?q=test&tags%5B%5D=tag1&tags%5B%5D=tag2')
   })
 
-  it('Should use default buildSearchParams in $url() when custom one is not provided', () => {
-    const client = hc<AppType>('http://localhost')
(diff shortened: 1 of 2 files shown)


src/client/client.test.ts
```diff
@@ -1787,11 +1787,34 @@ describe('Custom buildSearchParams', () => {
     expect(url.href).toBe('http://localhost/search?q=test&tags%5B%5D=tag1&tags%5B%5D=tag2')
   })
 
-  it('Should use default buildSearchParams in $url() when custom one is not provided', () => {
```
yusukebe, line 1790: This "Should use default buildSearchParams in $url() when a custom one is not provided" test is removed in this PR. Can you not remove it?

adityajha77: "Restored the missing $url() test and verified all client unit tests pass!"

MahathirMohammadShuvo:
Just a heads-up in case it is useful — I opened #5272 for a related bit of this
and wanted to flag how the two fit together, so neither of us duplicates the
other.

Routing `$ws` through `buildSearchParams` the way you have here is the right
call, and it means `$ws` automatically picks up whatever that function does. The
one thing it does not fix on its own is an `undefined` *inside* an array:
`buildSearchParams` skips a top-level `undefined` but appends array entries
unconditionally, so `{ tag: ['a', undefined, 'b'] }` still serialises as
`tag=a&tag=undefined&tag=b`. #5272 is a three-line guard for exactly that, in
`utils.ts`.

So they are complementary rather than competing — once both land, `$ws` gets the
array case for free through your change. I deliberately left `client.ts`
untouched in mine so there is nothing to conflict with here.

yusukebe (approved):
LGTM!

yusukebe:
@adityajha77 Thanks!

adityajha77:
Welcome sir @yusukebe , happy to contribute .

PR #5257, opened 2026-08-17
fix(jsx): normalize SVG camelCase attribute keys to hyphenated names

adityajha77:
## Description

Fixes an issue in Hono JSX where SVG camelCase attribute keys (e.g., `strokeWidth`, `fillRule`, `strokeLinecap`) were not normalized to valid hyphenated HTML/SVG attribute names during rendering.

### Problem
`normalizeElementKeyMap` previously only mapped a few HTML attributes (`className`, `htmlFor`, etc.). When rendering SVG icon sets (such as Lucide or Heroicons React components) via Hono JSX / SSR, camelCase SVG attributes were preserved as-is (e.g. `strokeWidth="2"`), causing browsers to ignore the attributes and render icons incorrectly.

### Solution
Extended `normalizeElementKeyMap` in `src/jsx/utils.ts` to include standard SVG camelCase attribute mappings:
- `strokeWidth` -> `stroke-width`
- `strokeLinecap` -> `stroke-linecap`
- `strokeLinejoin` -> `stroke-linejoin`
- `strokeDasharray` -> `stroke-dasharray`
- `strokeDashoffset` -> `stroke-dashoffset`
- `strokeMiterlimit` -> `stroke-miterlimit`
- `strokeOpacity` -> `stroke-opacity`
- `fillRule` -> `fill-rule`
- `fillOpacity` -> `fill-opacity`
- `clipPath` -> `clip-path`
- `clipRule` -> `clip-rule`
- `stopColor` -> `stop-color`
- `stopOpacity` -> `stop-opacity`

## Checklist
- [x] Added unit tests in `src/jsx/utils.test.ts` for SVG camelCase attribute key normalization.
- [x] All JSX unit tests pass (`npx vitest run src/jsx/`).

### DIFF
--- a/src/jsx/utils.test.ts
+++ b/src/jsx/utils.test.ts
@@ -7,16 +7,29 @@ import {
 
 describe('normalizeIntrinsicElementKey', () => {
   test.each`
-    key                | expected
-    ${'className'}     | ${'class'}
-    ${'htmlFor'}       | ${'for'}
-    ${'crossOrigin'}   | ${'crossorigin'}
-    ${'httpEquiv'}     | ${'http-equiv'}
-    ${'itemProp'}      | ${'itemprop'}
-    ${'fetchPriority'} | ${'fetchpriority'}
-    ${'noModule'}      | ${'nomodule'}
-    ${'formAction'}    | ${'formaction'}
-    ${'href'}          | ${'href'}
+    key                  | expected
+    ${'className'}       | ${'class'}
+    ${'htmlFor'}         | ${'for'}
+    ${'crossOrigin'}     | ${'crossorigin'}
+    ${'httpEquiv'}       | ${'http-equiv'}
+    ${'itemProp'}        | ${'itemprop'}
+    ${'fetchPriority'}   | ${'fetchpriority'}
+    ${'noModule'}        | ${'nomodule'}
+    ${'formAction'}      | ${'formaction'}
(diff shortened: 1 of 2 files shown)


usualoma:
The current `main` branch already supports camelCase SVG attributes in both the SSR and DOM renderers.

Existing tests cover this behavior:

- [SSR](https://github.com/honojs/hono/blob/main/src/jsx/index.test.tsx#L1051-L1114)
- [DOM renderer](https://github.com/honojs/hono/blob/main/src/jsx/dom/index.test.tsx#L2779-L2845)
- [SVG root element in SSR](https://github.com/honojs/hono/blob/main/src/jsx/index.test.tsx#L978-L983)

The attributes added in this PR, including `strokeWidth`, `fillRule`, and `clipPath`, are already covered there.

If there is a case that does not work on the current `main` branch, please provide a minimal reproduction showing the input JSX, actual output, and expected output first.

adityajha77:
Thanks for pointing this out! I checked and confirmed main already supports this. Closing the PR.

usualoma:
If this is not addressing a problem you have actually encountered or functionality you need, please do not submit changes based solely on improvements suggested by AI.

Even when such suggestions are not false positives, reviewing, testing, and maintaining speculative changes creates additional work for maintainers. We generally do not welcome AI-generated improvement proposals without a concrete use case or a reproducible problem.

PR #5260, opened 2026-08-18
fix(reg-exp-router): match tail wildcard without a leading slash befo…

shiningsudipto:
…re *

buildWildcardRegExp() only recognized a wildcard suffix when it was preceded by a slash (e.g. "/assets/*"). A pattern like "/assets*" fell through to the generic metacharacter-escaping branch instead, so the "*" was escaped literally and the compiled regex became "^/assets\*$" — which only matches the literal string "/assets*" and never a real path such as "/assets/foo.js".

That regex is used by add() to associate an already-registered wildcard route/middleware with sibling paths added before or after it (via findMiddleware() and the re.test(p) cross-association loops). Because it never matched, middleware registered as app.use('/assets*', mw) silently never attached to routes added around it, even though a plain app.get('/assets*', handler) still dispatched correctly on its own (that path goes through the trie, not this regex).

Fix: treat a bare trailing "*" the same as the "/*" case, only without forcing a leading slash — replace it with ".*" instead of escaping it, while leaving the existing "/*" -> "(?:|/.*)" behavior untouched.

Adds regression tests covering both registration orders (wildcard added before and after the concrete path) for a pattern without a leading slash.

Fixes #5258

### The author should do the following, if applicable

- [x] Add tests
- [x] Run tests
- [x] `bun run format:fix && bun run lint:fix` to format the code
- [ ] Add [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code

PR #5261, opened 2026-08-18
fix(reg-exp-router): apply a trailing wildcard to paths using a different label name

edwinsatya:
## What this fixes

A middleware registered with a label before a trailing wildcard is silently skipped by
`RegExpRouter` whenever a route uses a **different label name** for the same segment.

```ts
const app = new Hono()

app.use('/:name/*', async (c, next) => {
  c.header('X-Name', c.req.param('name'))
  await next()
})
app.get('/:id', (c) => c.text('handler'))

const res = await app.request('/abc')
res.headers.get('X-Name') // null  -- expected 'abc'
```

This happens on the default router (`SmartRouter + RegExpRouter`), so it affects apps
out of the box. `TrieRouter` and `PatternRouter` both run the middleware here.

The behaviour depends only on the label *name*, which makes it easy to hit by accident:

| routes | `RegExpRouter` before | `TrieRouter` / `PatternRouter` |
| --- | --- | --- |
| `/:a/*` + `/:a` (same name) | middleware runs | middleware runs |
| `/:a/*` + `/:b` (different name) | **middleware skipped** | middleware runs |
| `/a/*` + `/a` (static) | middleware runs | middleware runs |

## Cause

`RegExpRouter` decides which registered paths a trailing wildcard is associated with by
testing the wildcard against the registered path *strings*, not against request paths.
`buildWildcardRegExp('/:name/*')` produces `^/:name(?:|/.*)$`, which is then tested
against the literal string `/:id` and does not match.

## The change

A plain `:label` matches any single path segment, so its name does not affect which paths
a wildcard covers. The association regexp now replaces plain labels with a pattern that
matches any label. Patterns with a regexp (`:label{[0-9]+}`) are kept as literals, because
they match a narrower set of segments and must not be broadened.

Because an associated path may now use different label names, the wildcard path's own
label names are carried alongside the handler and used to build the param map. This keeps
`c.req.param()` inside the middleware reporting the names it was registered with, matching
`TrieRouter` exactly.

Everything changed runs at registration and build time, so there is **no per-request cost**.

## Tests

- `src/router/common.case.test.ts` — a shared case, so it runs against every router.
  `LinearRouter` is skipped, as it already raises `UnsupportedPathError` for a label
  followed by a wildcard.
- `src/hono.test.ts` — an app-level case covering the reported symptom.

Without the source change these fail in three places: `RegExpRouter`,
`PreparedRegExpRouter` and the app-level test.

## Verification

I checked this with a differential test that registers the same routes in all four routers
and compares both the matched handlers and the resulting params.

- Across ~5,200 generated wildcard/label cases: **55 distinct divergences before, 2 after**.
- The 2 remaining divergences are the pre-existing duplicate-label-name limitation
  (`/:id/:id{[0-9]+}`), which is unrelated to this change and already documented in the
  routers' test skip lists.
- A separate order-sensitive sweep over 60,000 route/path combinations reports 0
  divergences after the change.

Full suite passes: 147 files, 4851 tests.

Note: this touches `buildWildcardRegExp`, which #5259 and #5260 also modify for a different
case (a trailing wildcard with no slash before it). Happy to rebase if either lands first.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/hono.test.ts
+++ b/src/hono.test.ts
@@ -487,6 +487,21 @@ describe('Routing', () => {
     expect(app.router.name).toBe('SmartRouter + TrieRouter')
   })
 
+  it('Should apply a wildcard middleware to a route using a different label name', async () => {
+    const app = new Hono()
+    app.use('/:name/*', async (c, next) => {
+      c.header('X-Name', c.req.param('name'))
+      await next()
+    })
+    app.get('/:id', (c) => c.text(`id=${c.req.param('id')}`))
+
+    const res = await app.request('http://localhost/abc')
+    expect(res.status).toBe(200)
+    expect(res.headers.get('X-Name')).toBe('abc')
+    expect(await res.text()).toBe('id=abc')
+    expect(app.router.name).toBe('SmartRouter + RegExpRouter')
+  })
+
   it('Nested route - subApp with basePath', async () => {
     const app = new Hono()
     const book = new Hono().basePath('/book')
--- a/src/router/common.case.test.ts
+++ b/src/router/common.case.test.ts
@@ -422,6 +422,29 @@ export const runTest = ({
       })
     })
 
+    describe('Trailing wildcard after a label', () => {
+      beforeEach(() => {
+        router.add('GET', '/:name/*', 'middleware')
+        router.add('GET', '/:id', 'handler')
+      })
+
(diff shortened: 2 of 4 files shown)


yusukebe:
Closing since #5266 was merged

PR #5263, opened 2026-08-18
fix(jsx/dom): run previous ref cleanup when the ref changes

### LINKED ISSUE #5262: fix(jsx/dom): execute previous ref cleanup when ref prop changes on re-render
### Problem Description

In Hono's client-side JSX DOM renderer (`src/jsx/dom/render.ts`), when a DOM element's `ref` prop reference changes during a re-render (e.g., `<div ref={condition ? ref2 : ref1} />`), Hono registers the new ref callback `ref2(container)` and updates `refCleanupMap`, but it bypasses the cleanup function returned by the previous ref (`cleanup1` or `ref1(null)`).

#### Root Cause
Currently, `ref` cleanup is only triggered when a prop key is completely removed from the element via the deletion loop (`!(key in attributes)`). When a `ref` prop simply switches identity from `ref1` to `ref2`, `'ref'` remains in `attributes`, so the cleanup loop is bypassed. Consequently, any teardown logic attached to `ref1` (such as unbinding observers or cleaning up event listeners) is leaked and never executed.

This leads to memory leaks in single-page applications when switching dynamic ref callbacks, conditional inline refs, or dynamic observers during component updates.

---

### Minimal Reproducible Example

Add the following test case to `src/jsx/dom/index.test.tsx` and run `npx vitest run src/jsx/dom/index.test.tsx`:

```tsx
it('executes ref cleanup when ref callback reference changes on re-render', async () => {
  const cleanup1 = vi.fn()
  const ref1 = vi.fn().mockReturnValue(cleanup1)
  const cleanup2 = vi.fn()
  const ref2 = vi.fn().mockReturnValue(cleanup2)

  const App = () => {
    const [useRef2, setUseRef2] = useState(false)
    return (
      <>
        <div ref={useRef2 ? ref2 : ref1} />
        <button onClick={() => setUseRef2(true)}>switch</button>
      </>
    )
  }

  render(<App/>, root)
  expect(ref1).toHaveBeenCalledTimes(1)
  expect(cleanup1).not.toHaveBeenCalled()

  // Switch ref from ref1 to ref2
  root.querySelector('button')?.click()
  await Promise.resolve()

  // Expect previous ref's cleanup function to have been invoked once
  expect(cleanup1).toHaveBeenCalledTimes(1)
  expect(ref2).toHaveBeenCalledTimes(1)
})

lazerg:
Fixes #5262

When a `ref` prop changes identity across a re-render, `applyProps` registers the new ref and overwrites `refCleanupMap`, so the cleanup returned by the previous ref (or the `ref(null)` call for a plain callback ref) never runs. The cleanup currently fires only when the `ref` key disappears from the props entirely, or when the node is removed.

The fix invokes the stored cleanup before the new ref is attached, matching how the event listener branch above already removes the old listener when the handler changes.

The removal branch now also drops the entry from `refCleanupMap` after running it. That element stays in the tree with only its `ref` gone, so without the delete a later re-added `ref` would run the stale cleanup a second time.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/jsx/dom/index.test.tsx
+++ b/src/jsx/dom/index.test.tsx
@@ -346,6 +346,30 @@ describe('DOM', () => {
       expect(ref).toBeCalledTimes(1)
       expect(cleanup).toBeCalledTimes(1)
     })
+
+    it('ref cleanup function on ref change', async () => {
+      const cleanup1 = vi.fn()
+      const ref1 = vi.fn().mockReturnValue(cleanup1)
+      const ref2 = vi.fn()
+      const App = () => {
+        const [swapped, setSwapped] = useState(false)
+        return (
+          <>
+            <div ref={swapped ? ref2 : ref1} />
+            <button onClick={() => setSwapped(true)}>switch</button>
+          </>
+        )
+      }
+      render(<App />, root)
+      expect(root.innerHTML).toBe('<div></div><button>switch</button>')
+      expect(ref1).toHaveBeenLastCalledWith(expect.any(dom.window.HTMLDivElement))
+      expect(cleanup1).not.toHaveBeenCalled()
+      root.querySelector('button')?.click()
+      await Promise.resolve()
+      expect(root.innerHTML).toBe('<div></div><button>switch</button>')
+      expect(cleanup1).toBeCalledTimes(1)
+      expect(ref2).toHaveBeenLastCalledWith(expect.any(dom.window.HTMLDivElement))
+    })
   })
 
   describe('child component', () => {
--- a/src/jsx/dom/render.ts
+++ b/src/jsx/dom/render.ts
@@ -187,6 +187,7 @@ const applyProps = (
       } else if (key === 'dangerouslySetInnerHTML' && value) {
         container.innerHTML = value.__html
       } else if (key === 'ref') {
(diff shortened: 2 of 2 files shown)


usualoma:
Closing this PR under our AI Usage Policy.
This PR was opened only 13 minutes after an issue was filed by another contributor, before maintainers had an opportunity to triage it or respond to the reporter’s offer to submit a fix. Together with #5220, this follows a recurring pattern of rapidly turning newly filed issues into PRs, creating unnecessary triage and review overhead for maintainers.
Whether or not AI was used, we do not want this style of contribution to continue. Please stop submitting PRs in this manner. Future contributions following the same pattern may be closed without review, and your account may be blocked.

lazerg:
Hmmm 🤨. So you mean "the code is correct, but I close this, because it is AI assisted"?

usualoma:
No. AI-assisted contributions are allowed. As explained above, this PR was closed because of the recurring contribution pattern and the unnecessary maintainer overhead it creates—not merely because AI may have been used.

This decision was made under our [AI Usage Policy](https://github.com/honojs/hono/blob/main/docs/CONTRIBUTING.md#ai-usage-policy), which explicitly applies regardless of whether AI was actually used. The PR will remain closed.

PR #5264, opened 2026-08-18
fix(jsx/dom): execute previous ref cleanup when ref prop changes on re-render

### LINKED ISSUE #5262: fix(jsx/dom): execute previous ref cleanup when ref prop changes on re-render

adityajha77:
Closes #5262

### Summary
Fixes an issue in `hono/jsx/dom` where updating a element's `ref` prop to a new callback/reference during a re-render did not execute the previous `ref` cleanup function.

### Changes
- Updated `applyProps` in `src/jsx/dom/render.ts`:
  - When `key === 'ref'` and `oldAttributes?.[key] && oldAttributes[key] !== value`, invoke the previous cleanup function (`refCleanupMap.get(container)?.()`) before setting up the new ref.
  - Guaranteed that if the `ref` reference has **not** changed across re-renders, the cleanup function is **not** called unnecessarily.
  - Added `refCleanupMap.delete(container)` when `ref` is removed in the attribute removal loop.
- Added unit tests in `src/jsx/dom/index.test.tsx` under `describe('DOM > attribute')`:
  - Verified `cleanup1` is called exactly once when switching from `ref1` to `ref2` on re-render, and `ref2` is invoked.
  - Verified that re-rendering with the same `ref` reference does not call the cleanup function.

### Verification
- Ran `npx vitest run src/jsx/dom/index.test.tsx` — all 843 tests passed across `jsx-runtime-dom`, `jsx-runtime-default`, and `main`.

### DIFF
--- a/src/jsx/dom/index.test.tsx
+++ b/src/jsx/dom/index.test.tsx
@@ -401,6 +401,30 @@ describe('DOM', () => {
       expect(ref).toHaveBeenCalledTimes(1)
       expect(cleanup).toHaveBeenCalledTimes(0)
     })
+
+    it('ref cleanup on ref prop removal', async () => {
+      const cleanup = vi.fn()
+      const ref = vi.fn().mockReturnValue(cleanup)
+
+      const App = () => {
+        const [hasRef, setHasRef] = useState(true)
+        return (
+          <>
+            <div {...(hasRef ? { ref } : {})} />
+            <button onClick={() => setHasRef(false)}>remove</button>
+          </>
+        )
+      }
+
+      render(<App />, root)
+      expect(ref).toHaveBeenCalledTimes(1)
+      expect(cleanup).toHaveBeenCalledTimes(0)
+
+      root.querySelector('button')?.click()
+      await Promise.resolve()
+
+      expect(cleanup).toHaveBeenCalledTimes(1)
+    })
   })
 
   describe('child component', () => {
--- a/src/jsx/dom/index.test.tsx
+++ b/src/jsx/dom/index.test.tsx
@@ -346,6 +346,61 @@ describe('DOM', () => {
       expect(ref).toBeCalledTimes(1)
       expect(cleanup).toBeCalledTimes(1)
     })
+
+    it('ref cleanup on ref prop change', async () => {
+      const cleanup1 = vi.fn()
+      const ref1 = vi.fn().mockReturnValue(cleanup1)
+      const cleanup2 = vi.fn()
+      const ref2 = vi.fn().mockReturnValue(cleanup2)
+
+      const App = () => {
+        const [currentRef, setCurrentRef] = useState(() => ref1)
+        return (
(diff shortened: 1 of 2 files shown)


usualoma:
Hi @adityajha77,

Thank you for the PR.

I opened [adityajha77/hono#1](https://github.com/adityajha77/hono/pull/1) to simplify the condition.

The outer condition in `applyProps` already ensures that this branch runs only when the `ref` changes or during the initial render, where the cleanup call is a no-op. The inner condition is therefore unnecessary.

Please review and merge it.

adityajha77:
"Merged your PR into this branch. Thanks @usualoma!"

usualoma:
Thank you!

yusukebe (approved):
LGTM!

yusukebe:
@adityajha77 @usualoma Thank you both!

PR #5267, opened 2026-08-18
perf(router): share null object creation

usualoma:
### Summary

This introduces an internal, non-exported `createNullObject()` helper in `src/internal/utils.ts` and uses it for the null-prototype objects created by `RegExpRouter` and `TrieRouter`.

Repeated `Object.create(null)` calls are large enough that sharing them behind a small internal helper produces a clear and measurable reduction in the minified bundle. The helper is internal implementation detail and is not added to Hono's public API.

### Bundle size

Measured with esbuild using `--bundle --minify --format=esm --target=es2022`, compared with the parent branch:

| Entry point | Before | After | Difference |
| --- | ---: | ---: | ---: |
| `src/index.ts` | 18,179 bytes | 17,973 bytes | **-206 bytes** |
| `src/router/reg-exp-router/index.ts` | 6,025 bytes | 5,861 bytes | **-164 bytes** |
| `src/router/trie-router/index.ts` | 2,793 bytes | 2,775 bytes | **-18 bytes** |

The minified-size benefit is concrete, but the runtime trade-off is less universal.

### Runtime considerations

Calling `createNullObject()` introduces a JavaScript function boundary around `Object.create(null)`. A JIT may inline it, but whether and when that happens can depend on the runtime, warm-up state, and call site. Hono supports multiple runtimes, so the result should not be assumed to be identical everywhere.

This PR applies the helper consistently within `RegExpRouter` and `TrieRouter`, including an allocation performed during TrieRouter matching. It does not attempt to replace every `Object.create(null)` in the project, nor does it establish that doing so would always be runtime-neutral.

Future replacements should be evaluated individually, considering:

- the measured minified-size reduction
- whether the call is on an initialization, registration, build, or request-time path
- possible differences between runtimes
- benchmarks when a performance-sensitive path is affected

## Important: this must not become a style rule

My biggest concern is that this helper could become a de facto style rule, leading to context-free automated or AI-generated PRs that mechanically replace every `Object.create(null)`. I strongly want to avoid that outcome; each use should remain a deliberate decision backed by size measurements and runtime considerations.

- [x] Run tests


yusukebe:
Hey @usualoma !

I think the direction of this optimization is good. And I can understand your concerns, including that AI PRs may replace every `Object.create(null)`, very well. Agreed.

One thing I noticed: I created `createResponseInstance` in `context.ts` for the same reason that reduces bundle size:

https://github.com/honojs/hono/blob/a10592f3e1864cbe277619633db953450a6b1e4b/src/context.ts#L288-L291

It's good to scope only `context.ts` if it's writing to `context.ts`. But moving `createResponseInstance` to `src/internal/utils.ts` and reusing it is also a good idea (maybe we can use it rather than `context.ts`).

usualoma:
There's also something like `decodeURIComponent`, isn't there?
https://github.com/honojs/hono/blob/fdcbafcce9de6e3c18d187110cf6cdeecf22da33/src/utils/url.ts#L319

I think there are other areas where we could optimize using the same approach, but once we start down that path, I’m afraid AI slop will just keep popping up endlessly.

yusukebe:
@usualoma 

> There's also something like `decodeURIComponent`, isn't there?

Ah, yeah. That uses the same pattern.

> I think there are other areas where we could optimize using the same approach, but once we start down that path, I’m afraid AI slop will just keep popping up endlessly.

You are right. And I think it's a good idea to narrow the scope for the optimization. Then, how about moving `createNullObject` from `src/internal/utils.ts` to `src/router/utils.ts`? This can indicate that the function is only used in the routers. And make this PR only for that, not for `createResponseInstance` or anything else. Let's think about them later. What do you think of it?

usualoma:
@yusukebe 
That sounds good! Let's go with that for now.

[6676b93](https://github.com/honojs/hono/pull/5267/commits/6676b93fa310c9cda508ce7f24504fa7e713a9a6) I've applied the proposal for src/router/utils.ts.

yusukebe (approved):
LGTM!

yusukebe:
@usualoma Thanks!

PR #5268, opened 2026-08-19
fix(request): handle params on unmatched requests

usualoma:
Prevents `req.param()` from throwing when the request has no matched route.

Keyed access returns `undefined`, while `req.param()` returns an empty object. Includes a regression test.

- [x] Add tests
- [x] Run tests


yusukebe (approved):
LGTM!

yusukebe:
@usualoma Thanks!

PR #5266, opened 2026-08-18
fix(reg-exp-router): associate wildcard middleware with matching routes

usualoma:
This fixes two inconsistencies in how `RegExpRouter` associates wildcard middleware with registered routes:

1. Suffix wildcards without a preceding slash now work correctly.

   A middleware such as `/assets*` is associated with `/assets/app.js`, regardless of registration order. Previously, the `*` was treated as a literal character during middleware association.

2. Trailing wildcards after parameters now work across different parameter names.

   For example, middleware registered as `/:name/*` is associated with a route registered as `/:id`, while each handler still receives parameters under its own registered name.

   This also supports equivalent custom patterns, including nested braces and regexp metacharacters:

   - `/:year{[0-9]{4}}/*` and `/:yr{[0-9]{4}}/comments`
   - `/:kind{(?:foo|bar)}/*` and `/:type{(?:foo|bar)}/detail`
   - `/:userId{[^/]+}/*` and `/:id/profile`

This also resolves the related wildcard association issues discussed in #5260 and #5261.

Fixes #5258

### Bundle size

Measured with esbuild using `--bundle --minify --format=esm --target=es2022`, compared with `origin/main`:

| Entry point | Before | After | Difference |
| --- | ---: | ---: | ---: |
| `src/index.ts` | 18,657 bytes | 18,179 bytes | **-478 bytes (-2.6%)** |
| `src/router/reg-exp-router/index.ts` | 6,501 bytes | 6,025 bytes | **-476 bytes (-7.3%)** |

The implementation was compacted so the additional wildcard behavior reduces the overall minified bundle size.

### Performance

The wildcard association work is performed only during route registration and matcher construction.

Route registration is slightly faster due to fewer redundant iterations and allocations. Steady-state `match()` performance is unchanged because the per-request matching path is not modified.

### The author should do the following, if applicable

- [x] Add tests
- [x] Run tests
- [x] `bun run format:fix && bun run lint:fix` to format the code

PR #5270, opened 2026-08-19
fix(reg-exp-router): apply middleware registered with a tail wildcard that has no preceding slash

### LINKED ISSUE #5258: Middleware wildcard without leading slash before * never matches (RegExpRouter/TrieRouter), unlike an equivalent route
## Descripción del bug

Un middleware registrado con un patrón wildcard sin `/` antes del `*` (por ejemplo `app.use('/assets*', mw)`) nunca se ejecuta con `RegExpRouter` ni con `TrieRouter`, aunque una ruta normal (`app.get('/assets*', handler)`) con el mismo patrón sí matchea correctamente contra el mismo path.

## Reproducción

Con `hono@4.13.2` (el último publicado en `latest`):

```ts
import { Hono } from 'hono'

// Ruta normal: SÍ matchea
const app1 = new Hono()
app1.get('/assets*', (c) => c.text('route-hit'))
const res1 = await app1.request('/assets/app.js')
console.log(res1.status) // 200

// Middleware con el mismo patrón: NUNCA se ejecuta
const app2 = new Hono()
app2.use('/assets*', async (c, next) => {
  c.header('X-MW', 'hit')
  await next()
})
app2.get('/assets/app.js', (c) => c.text('ok'))
const res2 = await app2.request('/assets/app.js')
console.log(res2.headers.get('x-mw')) // null, esperaría "hit"

// Para comparar, con slash antes del * sí funciona
const app3 = new Hono()
app3.use('/assets/*', async (c, next) => {
  c.header('X-MW', 'hit')
  await next()
})
app3.get('/assets/app.js', (c) => c.text('ok'))
const res3 = await app3.request('/assets/app.js')
console.log(res3.headers.get('x-mw')) // "hit"
```

Corrí exactamente este script contra el paquete `hono@4.13.2` instalado desde npm y confirmé los tres resultados de los comentarios.

## Causa raíz que encontré

En `src/router/reg-exp-router/router.ts`, `buildWildcardRegExp` es la función que asocia middleware ya registrado con wildcard a las rutas explícitas subsecuentes:

```ts
function buildWildcardRegExp(path: string): RegExp {
  return (wildcardRegExpCache[path] ??= new RegExp(
    path === '*'
      ? ''
      : `^${path.replace(/\/\*$|([.\+*[^\]$()])/g, (_, metaChar) =>
          metaChar ? `\${metaChar}` : '(?:|/.*)'
        )}$`
  ))
}
```

El primer alternativo de la regex, `\/\*$`, solo reconoce el wildcard cuando está precedido de `/` (o sea `/algo/*`). Para `/assets*` (sin `/` antes del `*`) esa alternativa no matchea, así que el `*` cae en la segunda alternativa (la clase de metacaracteres) y se escapa literalmente. El patrón final construido es `^/assets\*$`, que solo matchea el string literal `/assets*`, nunca `/assets`, `/assets-v2` ni `/assets/app.js`.

Confirmé el mismo comportamiento corriendo la regex construida directamente en Node:

```
buildWildcardRegExp('/assets*').source  →  ^\/assets\*$
buildWildcardRegExp('/assets*').test('/assets/app.js')  →  false
buildWildcardRegExp('/assets/*').source →  ^\/assets(?:|\/.*)$
buildWildcardRegExp('/assets/*').test('/assets/app.js') →  true
```

`TrieRouter` tiene el mismo problema para middleware (reproducido arriba, `res2` da `null` también con `TrieRouter` como router activo). `LinearRouter` y `PatternRouter` sí lo manejan bien.

Lo que hace esto más confuso es que el propio router de rutas normales (no el de middleware) sí soporta `/assets*` sin slash como wildcard válido, así que no es un patrón inválido en general dentro de Hono, es específicamente `findMiddleware`/`buildWildcardRegExp` el que no reconoce la misma forma.

Revisé los PRs #5236, #5252 y #5243 (mergeados esta semana, arreglan inconsistencias de wildcard en TrieRouter/LinearRouter/PatternRouter/RegExpRouter para el *matching de rutas*) y ninguno toca `reg-exp-router/router.ts`'s `buildWildcardRegExp`, que es un mecanismo aparte usado solo para resolver qué middlewares aplican a una ruta ya registrada.

## Impacto

Como `RegExpRouter` es el router por defecto (via `SmartRouter`), un patrón común como `app.use('/api*', authMiddleware)` (sin slash final antes del `*`, en vez de `/api/*`) deja de aplicar el middleware de forma completamente silenciosa, sin error ni warning.

## Entorno

- hono: 4.13.2 (npm, `latest`)
- Node: verificado también contra el código actual de `main` (`buildWildcardRegExp` sin cambios)

yigiterturk-dev:
Closes #5258.

- [x] Add tests
- [x] Run tests
- [x] `format:fix` / `lint:fix` (verified clean with `prettier --check` and `eslint`)
- [ ] Add TSDoc/JSDoc — not applicable, internal helper

---

### Problem

`findMiddleware` matches already-registered middleware against a path using `buildWildcardRegExp`, whose only wildcard branch is `\/\*$` — it requires a slash before the `*`. For a pattern like `/assets*` the `*` falls through to the metacharacter class and is escaped literally:

```
/assets*   ->  ^\/assets\*$          matches only the literal string "/assets*"
/assets/*  ->  ^\/assets(?:|\/.*)$
```

So `app.use('/assets*', mw)` never runs, while `app.get('/assets*', handler)` matches the same requests — route matching goes through a different code path. Since `RegExpRouter` is the default via `SmartRouter`, a common pattern like `app.use('/api*', authMiddleware)` silently applies no middleware, with no error or warning.

### Fix

Add a branch for a trailing `*` with no preceding slash and emit a prefix match:

```js
path.replace(/\/\*$|\*$|([.\\+*[^\]$()])/g, (match, metaChar) =>
  metaChar ? `\\${metaChar}` : match === '/*' ? '(?:|/.*)' : '.*'
)
```

Alternation order matters here: `\/\*$` is tried first, so `/foo/*` keeps its existing `(?:|/.*)` expansion and still does not match a sibling prefix like `/foo-v2`. A non-trailing `*` (e.g. `/a*b`) keeps being escaped as a literal, as before.

Prefix semantics for the no-slash form matches what the other routers already do for the same pattern as a route — `LinearRouter`, `PatternRouter` and `RegExpRouter`'s own route matching all treat `/assets*` as matching `/assets`, `/assets/app.js`, `/assets-v2` and `/assetsX`.

### Tests

Three cases added to `router.test.ts`:

- middleware registered as `/assets*` is applied to `/assets/app.js`
- the no-slash form matches a sibling prefix (`/assets*` against `/assets-v2/app.js`)
- regression guard: `/assets/*` still does **not** match `/assets-v2/app.js`

The first two fail on `main` and pass with this change; the third passes both ways and exists to pin the `/foo/*` semantics.

Full suite is green — 3861 passed, 28 skipped across 123 files — plus `tsc -p tsconfig.spec.json`, `prettier --check` and `eslint` clean.

### Scope

This only covers `RegExpRouter`'s middleware path. As noted in #5258, `TrieRouter` also does not match `/assets*` as a plain route while the other three routers do; that is a separate behaviour change and I left it out of this PR deliberately. Happy to follow up on it if you want it aligned.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Smwb1MRyiGRRMEEDVioV88

### DIFF
--- a/src/router/reg-exp-router/router.test.ts
+++ b/src/router/reg-exp-router/router.test.ts
@@ -158,6 +158,36 @@ describe('RegExpRouter', () => {
       expect(res[1][0]).toBe('wildcard')
     })
 
+    it('Should apply middleware registered with a tail wildcard that has no preceding slash', () => {
+      // https://github.com/honojs/hono/issues/5258
+      // `/assets*` is a valid route pattern, but `findMiddleware` used to escape the
+      // `*` literally, so middleware registered this way never matched any request.
+      const router = new RegExpRouter<string>()
+      router.add('ALL', '/assets*', 'middleware')
+      router.add('GET', '/assets/app.js', 'handler')
+
+      const [res] = router.match('GET', '/assets/app.js')
+      expect(res.map(([handler]) => handler)).toEqual(['middleware', 'handler'])
+    })
+
+    it('Should match a tail wildcard without a preceding slash as a prefix', () => {
+      const router = new RegExpRouter<string>()
+      router.add('ALL', '/assets*', 'middleware')
+      router.add('GET', '/assets-v2/app.js', 'handler')
+
+      const [res] = router.match('GET', '/assets-v2/app.js')
+      expect(res.map(([handler]) => handler)).toEqual(['middleware', 'handler'])
+    })
+
+    it('Should not let a tail wildcard with a preceding slash match a sibling prefix', () => {
+      const router = new RegExpRouter<string>()
+      router.add('ALL', '/assets/*', 'middleware')
+      router.add('GET', '/assets-v2/app.js', 'handler')
+
+      const [res] = router.match('GET', '/assets-v2/app.js')
+      expect(res.map(([handler]) => handler)).toEqual(['handler'])
+    })
+
     it('Should prioritize the only wildcard over the tail wildcard regardless of the order', () => {
       for (const paths of [
         ['/a*', '/a/*'],
--- a/src/router/reg-exp-router/router.ts
+++ b/src/router/reg-exp-router/router.ts
@@ -17,8 +17,8 @@ function buildWildcardRegExp(path: string): RegExp {
   return (wildcardRegExpCache[path] ??= new RegExp(
     path === '*'
       ? ''
-      : `^${path.replace(/\/\*$|([.\\+*[^\]$()])/g, (_, metaChar) =>
-          metaChar ? `\\${metaChar}` : '(?:|/.*)'
+      : `^${path.replace(/\/\*$|\*$|([.\\+*[^\]$()])/g, (match, metaChar) =>
+          metaChar ? `\\${metaChar}` : match === '/*' ? '(?:|/.*)' : '.*'
         )}$`
   ))
 }


yigiterturk-dev:
Superseded by #5266, which landed the same fix with a broader rewrite of the wildcard association logic. I checked `main` at `a10592f` — `app.use('/assets*', ...)` now applies on SmartRouter, RegExpRouter and TrieRouter, `/assets/*` still doesn't match a sibling prefix like `/assets-v2`, and the plain-route case that TrieRouter used to 404 (`app.get('/assets*')`) passes as well.

Nothing left for this to add — closing. Sorry for the duplicate, #5266 was already open when I started and I didn't check for linked PRs first.

PR #5271, opened 2026-08-19
perf(jws): reuse imported CryptoKey for string secrets

DenisCDev:
`jwt({ secret: '…', alg: 'HS256' })` and `Jwt.verify(token, secret, alg)` import the secret with `crypto.subtle.importKey` on every request. CryptoKey inputs already skip that. String secrets are the documented default.

On Node 22, `importKey` is about 30% of HMAC import+verify (14 µs vs 32 µs) and 42% of RSA SPKI import+verify (38 µs vs 53 µs). On Bun, HMAC `importKey` is cheaper; a full `app.fetch()` with `jwt()` + `c.json()` still spent about 7% of the authenticated request on it (84 µs vs 78 µs). JWT itself is ~90% of that request.

### What changed

`verifying()` caches the imported `CryptoKey` for `typeof publicKey === 'string'` (HMAC secrets and PEM). Keyed by `alg + secret`, cap 16, FIFO. The `CryptoKey` used for `subtle.verify` is a local binding, so eviction or a concurrent import cannot mix keys for an in-flight verify.

JWK objects are not cached. Mutating a JWK after the first verify would otherwise keep serving the old key. `signing()` is unchanged.

Factory-only cache inside `jwt()` would miss the documented Workers pattern that constructs the middleware per request from `c.env.JWT_SECRET`, and would miss standalone `Jwt.verify()`.

### Out of scope

- JWKS `fetch` on every `jwk({ jwks_uri })` request — users can already cache via `keys()`. That needs TTL and kid-miss refetch, which is a separate design.
- Signed cookies (`getCryptoKey`) — `parseSigned` already imports once per header.

### Testing

- `bunx vitest --run src/utils/jwt/jwt.test.ts src/middleware/jwt/index.test.ts` — 138 passed (6 new)
- `tsc -p tsconfig.spec.json` — OK
- New tests: same string is imported once; different string misses; signature mismatch still caches; JWK objects are not cached; two secrets in parallel do not cross; oldest of 16 is evicted

Not validated: Cloudflare Workers isolate (no local workerd run of this path).

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/utils/jwt/jws.ts
+++ b/src/utils/jwt/jws.ts
@@ -36,6 +36,12 @@ export async function signing(
   return await crypto.subtle.sign(algorithm, cryptoKey, data)
 }
 
+// String/PEM secrets are reused across requests; skip importKey after the first
+// call. Cap the map so Jwt.verify() cannot cache unbounded request-derived
+// material. JWK objects are not cached: mutation would keep the old CryptoKey.
+const MAX_CACHED_STRING_PUBLIC_KEYS = 16
+const cachedStringPublicKeys = new Map<string, CryptoKey>()
+
 export async function verifying(
   publicKey: SignatureKey,
   alg: SignatureAlgorithm,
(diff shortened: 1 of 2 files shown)


DenisCDev:
Hi — gentle follow-up on this one. The patch keeps JWK inputs uncached and bounds string-secret entries to 16; the focused JWT tests and typecheck are documented above. Is this caching direction something you'd be comfortable with in Hono? Happy to adjust or close if you prefer a different cache lifetime.

yusukebe:
We don't add in-process caches to the core. `jwt()` already accepts a `CryptoKey`, which skips `importKey` entirely, so users can use it to improve performance. This PR is not necessary. Closing. Thanks.

PR #5273, opened 2026-08-20
fix(method-not-allowed): ignore trailing wildcard routes when inferring allowed methods

### LINKED ISSUE #5223: When serveStatic and methodNotAllowed are used together, methods other than GET cannot be handled correctly

Viktor-Kolyshkin:
When a catch-all route like `app.get('/*', serveStatic())` is registered, `methodNotAllowed`
treats GET as allowed for every path, since the wildcard matches everything. A non-GET request
to a path that doesn't exist then returns 405 instead of 404 (#5223).

Skip routes whose path ends in `*` when collecting allowed methods, the same way `ALL` routes
are already skipped. A trailing wildcard isn't evidence that a specific resource exists. Middle
wildcards (`/a/*/b`) and regex catch-alls are left alone, matching the direction discussed in the
issue. The `serveStatic` part of the report is separate (#5224) and left untouched.

One side effect: when a wildcard GET also covers a path that has a concrete non-GET route, `Allow`
no longer lists GET. That is the intended trade-off, since a 404 on an otherwise-unhandled path is
better than a misleading 405.

Closes #5223

## Testing

- `vitest --run --project main`: 3920 passed, including the new method-not-allowed cases
- `tsc -p tsconfig.spec.json`, `eslint`, and `prettier --check` on the changed files

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/middleware/method-not-allowed/index.test.ts
+++ b/src/middleware/method-not-allowed/index.test.ts
@@ -197,6 +197,31 @@ describe('Method Not Allowed Middleware', () => {
     expect(res.headers.has('Allow')).toBe(false)
   })
 
+  it('ignores trailing wildcard routes when collecting allowed methods', async () => {
+    const app = new Hono()
+    app.use(methodNotAllowed({ app }))
+    app.get('/api', (c) => c.text('ok'))
+    // Catch-all GET routes, e.g. those registered by `app.get('/*', serveStatic())`
+    // or `app.get('/assets/*', serveStatic({ root: './public' }))`.
+    app.get('/*', (_c, next) => next())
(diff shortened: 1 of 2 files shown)


usualoma:

PR #5274, opened 2026-08-21
fix(utils/stream): do not let abort listeners crash abort()

tushardev-365:
`StreamingApi.abort()` invokes every abort subscriber bare:

```ts
this.abortSubscribers.forEach((subscriber) => subscriber())
```

`onAbort` accepts `() => void | Promise<void>`, and the first built-in subscriber is async (`await reader.cancel()`), so two failure modes exist on the client-disconnect path that triggers `abort()`:

- an async listener that rejects (cleanup RPC fails, resource already gone) becomes an unhandled promise rejection, which crashes the process under Node's default policy;
- a listener that throws synchronously stops the `forEach`, so the remaining listeners never run, including the built-in reader cancel when user listeners were registered before it runs.

The fix contains errors per listener: sync throws are caught, and a returned promise gets a no-op catch. Silently containing them matches the policy `write()` in the same class already applies ("Do nothing. If you want to handle errors, create a stream by yourself.").

Two tests added next to the existing `abort()` specs:

- a throwing listener must not prevent later listeners from running;
- a rejecting async listener must not produce an unhandled rejection (asserted via a `process.on('unhandledRejection')` capture).

Verification with the repo's vitest setup: with the fix, 13 passed (13); with current main's `stream.ts` and only the tests added, 2 failed | 11 passed, the two failures being exactly the new cases.

The author should do the following, if applicable

- [x] Add tests
- [x] Run tests
- [ ] Add [docs](https://github.com/honojs/website)

### DIFF
--- a/src/utils/stream.ts
+++ b/src/utils/stream.ts
@@ -95,7 +95,19 @@ export class StreamingApi {
   abort() {
     if (!this.aborted) {
       this.aborted = true
-      this.abortSubscribers.forEach((subscriber) => subscriber())
+      this.abortSubscribers.forEach((subscriber) => {
+        try {
+          const result = subscriber()
+          if (result instanceof Promise) {
+            result.catch(() => {
+              // Do nothing. A rejecting abort listener must not crash the runtime.
+            })
+          }
+        } catch {
+          // Do nothing. A throwing abort listener must not prevent the remaining
+          // listeners, including the built-in reader cancel, from running.
+        }
+      })
     }
   }
 }
--- a/src/utils/stream.ts
+++ b/src/utils/stream.ts
@@ -97,15 +97,9 @@ export class StreamingApi {
       this.aborted = true
       this.abortSubscribers.forEach((subscriber) => {
         try {
-          const result = subscriber()
-          if (result instanceof Promise) {
-            result.catch(() => {
-              // Do nothing. A rejecting abort listener must not crash the runtime.
-            })
-          }
+          void Promise.resolve(subscriber()).catch(() => {})
         } catch {
-          // Do nothing. A throwing abort listener must not prevent the remaining
-          // listeners, including the built-in reader cancel, from running.
+          // Ignore synchronous listener errors.
         }
       })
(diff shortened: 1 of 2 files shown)


usualoma:
Hi @tushardev-365,

Thanks for creating this PR. I agree that this should be fixed.

Could we avoid `instanceof Promise` here? It misses promises from another realm and generic thenables. Hono already explicitly supports thenable values in `ErrorBoundary` ([implementation](https://github.com/honojs/hono/blob/4b44abe1b5c7755aac8555d7943611555183450b/src/jsx/components.ts#L75-L93), [test](https://github.com/honojs/hono/blob/4b44abe1b5c7755aac8555d7943611555183450b/src/jsx/index.test.tsx#L1573-L1591)).

Could we normalize the listener result in the same spirit?

```ts
try {
  void Promise.resolve(subscriber()).catch(() => {})
} catch {
  // Ignore synchronous listener errors.
}
```

tushardev-365:
Good point about cross realm promises and thenables, adopted your shape as is. Also added a test with a rejecting thenable listener so the normalization is pinned. 14 passing locally.

usualoma:
Thank you!

yusukebe (approved):
LGTM!

yusukebe:
@tushardev-365 @usualoma Thanks!

tushardev-365:
Thank you!

src/utils/stream.ts
```diff
@@ -95,7 +95,13 @@ export class StreamingApi {
...
     if (!this.aborted) {
       this.aborted = true
-      this.abortSubscribers.forEach((subscriber) => subscriber())
+      this.abortSubscribers.forEach((subscriber) => {
+        try {
+          void Promise.resolve(subscriber()).catch(() => {})
+        } catch {
+          // Ignore synchronous listener errors.
+        }
+      })
```
acoBOYZ: For more simpler and readable equivalent could be used:

Promise.try(subscriber).catch({})

PR #5278, opened 2026-08-21
fix(body-limit): enforce the limit when Content-Length is not a number

pacocartones:
## Problem

In the `body-limit` middleware, the Content-Length-only fast path trusts `parseInt` without checking for `NaN`:

```ts
const contentLength = parseInt(c.req.raw.headers.get('content-length') || '0', 10)
return contentLength > maxSize ? onError(c) : next()
```

A non-numeric `Content-Length` (e.g. `Content-Length: abc`) makes `parseInt` return `NaN`, and `NaN > maxSize` is `false`, so the middleware calls `next()` and the handler reads the body with **no size limit** — the protection is silently bypassed. This only affects the branch where `Content-Length` is present and `Transfer-Encoding` is not (the streaming branch already checks the real byte count).

## Fix

When the parsed `Content-Length` is not a usable number, fall through to the existing streaming check, which enforces the limit against the actual bytes. Numeric Content-Length behavior is unchanged.

## Test

Adds a regression test that sends `Content-Length: not-a-number` with an oversized body and asserts `413`. Verified **red→green**: it fails on `main` (the body is accepted) and passes with the fix; the existing `body-limit` tests stay green (13/13).

AI was used to help find and verify this; the fix and test were reproduced end-to-end before opening.

### DIFF
--- a/src/middleware/body-limit/index.test.ts
+++ b/src/middleware/body-limit/index.test.ts
@@ -176,6 +176,30 @@ describe('Body Limit Middleware', () => {
       expect(await res.text()).toBe(smallContent)
     })
 
+    it('should not bypass the limit when Content-Length is not a number', async () => {
+      // A non-numeric Content-Length makes parseInt return NaN, and `NaN > maxSize`
+      // is false, which previously skipped the check and let an oversized body
+      // through. The limit must still be enforced against the actual bytes.
+      const largeContent = 'this is a large content that exceeds 10 bytes'
+      const stream = new ReadableStream({
+        start(controller) {
+          controller.enqueue(new TextEncoder().encode(largeContent))
+          controller.close()
+        },
+      })
+
+      const res = await app.request('/test', {
+        method: 'POST',
+        headers: {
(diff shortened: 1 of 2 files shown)


usualoma:
Thanks for the PR.

We have discussed this behavior before and intentionally decided that the `bodyLimit` middleware trusts the `Content-Length` value when it is present:
https://github.com/honojs/hono/pull/2077#issuecomment-1915778971

This behavior is also documented here:
https://hono.dev/docs/middleware/builtin/body-limit

A malformed `Content-Length` is an HTTP framing error and should be rejected by the runtime or HTTP parser before the request reaches Hono. The runtimes we tested either limit the body according to `Content-Length` or reject the request.

The regression test in this PR constructs a `Request` directly with an invalid `Content-Length` and a conflicting body, which does not represent an actual HTTP request handled by those runtimes.

Therefore, we don’t consider this a bug in `bodyLimit` and are going to close this PR. Thank you again for taking the time to investigate and submit it.

PR #5282, opened 2026-08-22
fix(request): drop stale content length for cloned FormData

Ersaoktaviannn:
## Summary

- remove the inherited `Content-Length` when `cloneRawRequest()` reserializes consumed `FormData`
- let the runtime determine representation headers for the newly encoded multipart body
- add a regression test using a valid multipart request with an explicit original length

## Impact

`cloneRawRequest()` supports forwarding requests after validation or body parsing. Multipart reserialization changes the boundary and can change the encoded size, so retaining the original `Content-Length` can leave downstream transports with invalid framing metadata.

## Regression proof

On the Node 22 runtime used for the regression test, the valid multipart fixture is 77 bytes. Before this fix, reserialization produced a 125-byte body while the clone still declared `Content-Length: 77`.

After this fix, the stale header is absent and the cloned form remains parseable, allowing the transport to determine the new length. This completes the representation-header handling introduced in #5133.

## Testing

- `vitest --run src/request.test.ts`: 43 passed
- `tsc -p tsconfig.spec.json --pretty false`
- `eslint src/request.ts src/request.test.ts`: 0 errors
- `prettier --check src/request.ts src/request.test.ts`
- Full `npm test`: 4,910 passed; local Windows `workerd` startup/WebSocket timed out. The `NO_COLOR`-dependent files passed 26/26 after unsetting the inherited variable.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/request.test.ts
+++ b/src/request.test.ts
@@ -593,6 +593,34 @@ describe('cloneRawRequest', () => {
     expect(formData.get('file')).toBeInstanceOf(File)
   })
 
+  test('drops stale content length when cloning consumed multipart request', async () => {
+    const boundary = 'boundary'
+    const body = [
+      `--${boundary}`,
+      'Content-Disposition: form-data; name="foo"',
+      '',
+      'bar',
+      `--${boundary}--`,
+      '',
+    ].join('\r\n')
+    const req = new HonoRequest(
+      new Request('http://localhost', {
+        method: 'POST',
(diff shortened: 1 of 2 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@Ersaoktaviannn Thanks!

PR #5283, opened 2026-08-22
fix(client): omit empty query delimiter

Ersaoktaviannn:
## Summary

`hc` currently appends `?` whenever query input is present, even when the configured serializer emits no query pairs. This makes an omitted optional query structurally different from an absent query.

This change centralizes query attachment so the delimiter is added only when the serialized query string is non-empty. It applies to request methods, `$url()`, and `$path()`; non-empty queries (including `q=`) are preserved.

## Before / after

For typed input `{ query: { q: undefined } }`:

| API | Before | After |
| --- | --- | --- |
| `$get()` request | `http://localhost/search?` | `http://localhost/search` |
| `$url().href` | `http://localhost/search?` | `http://localhost/search` |
| `$path()` | `/search?` | `/search` |

The same result is verified when a custom `buildSearchParams` returns an empty `URLSearchParams`.

## Impact

The URL Standard distinguishes an absent query from an empty query. Avoiding the dangling delimiter keeps generated client URLs canonical for exact comparisons, cache/signature keys, redirects, and logs, without changing queries that serialize at least one pair.

`$ws()` is unchanged because it already omits the delimiter when no pairs are appended. No public API or type changes are included.

## Verification

- `npx vitest --run src/client/client.test.ts` — 118 passed
- `npx prettier --check src/client/client.ts src/client/client.test.ts`
- `npx eslint src/client/client.ts src/client/client.test.ts` — no errors (one pre-existing warning elsewhere in the test file)
- Full `npm test` was attempted; Vitest aborted before assertions completed because its coverage temp directory was removed during the run. CI remains the authoritative full-suite check.

- [x] Add tests
- [x] Run tests
- [ ] Add TSDoc/JSDoc — not applicable; no public API added

### DIFF
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -1841,6 +1841,66 @@ describe('Query arrays containing undefined', () => {
   })
 })
 
+describe('Empty serialized query', () => {
+  const app = new Hono().get(
+    '/search',
+    validator('query', () => ({}) as { q?: string }),
+    (c) => c.json({ ok: true })
+  )
+  const input = { query: { q: undefined } }
+
+  it('Should omit the query delimiter when no query values are serialized', async () => {
+    let requestedUrl = ''
+    const client = hc<typeof app>('http://localhost', {
+      fetch: (input: RequestInfo | URL) => {
+        requestedUrl = input.toString()
+        return Promise.resolve(new Response('{}'))
+      },
+    })
+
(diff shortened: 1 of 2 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@Ersaoktaviannn Thanks!

PR #5289, opened 2026-08-23
fix(streaming): expose Last-Event-ID and make write-after-abort a no-op

### LINKED ISSUE #3765: SSE does not work again...

SynthLuvr:
Fixes #3765

## What

Two fixes to the SSE streaming lifecycle:

1. **`streamSSE` now exposes the `Last-Event-ID` request header as `stream.lastEventId`.** When an `EventSource` reconnects, the browser re-sends the `id` of the last event it received in this header, so handlers can resume the event stream instead of replaying everything. It is `undefined` when the header is absent.

2. **`StreamingApi.write()` and `close()` are no-ops after `abort()`.** Once the client is gone the writable may be errored, and touching the writer can throw or hang depending on the runtime. Both methods now return early when `aborted` is set, so a handler that keeps producing after a disconnect completes cleanly instead of hanging or crashing.

## Testing

- `src/utils/stream.test.ts`: `write()`/`close()` after `abort()` and after the client cancels the response stream resolve as no-ops without touching the writer.
- `src/helper/streaming/sse.test.tsx`: `lastEventId` is exposed from the header and `undefined` when absent; `writeSSE()` after a client disconnect resolves as a no-op and the handler finishes.
- `runtime-tests/node/index.test.ts`: end-to-end over a real Node server — resume from `Last-Event-ID` on reconnect, full replay without the header, and a handler finishing cleanly after a mid-stream disconnect.

### DIFF
--- a/runtime-tests/node/index.test.ts
+++ b/runtime-tests/node/index.test.ts
@@ -213,8 +213,8 @@ describe('streamSSE lifecycle (Last-Event-ID + write-after-abort)', () => {
   let handlerDone = false
   let writesAfterAbort = 0
 
-  // Resume-capable feed: replays only the events after the Last-Event-ID cursor,
-  // exactly like an EventSource reconnect would expect.
+  // Replays only the events after the Last-Event-ID cursor, as an EventSource
+  // reconnect would expect.
   app.get('/feed', (c) =>
     streamSSE(c, async (stream) => {
       const cursor = Number(stream.lastEventId ?? 0)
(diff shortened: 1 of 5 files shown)


yusukebe:
https://github.com/honojs/hono/issues/3765 has been closed. I don't really know the details of this PR, but the PR intended to close the issue should be closed.

SynthLuvr:
@yusukebe you keep closing PRs with no explanations. Feels like you're not really interested in maintaining Hono. It's a shame, because there are several upstream SSE bugs and it makes me consider using a different framework

yusukebe:
@SynthLuvr 

That's wrong. I am maintaining Hono by finding free time in my work and life. Plus, planning to release the next major version. The problem is that we have many PRs to review, including AI-generated ones. So we have to close PRs that ignore the context of the issues.

PR #5291, opened 2026-08-24
fix(client): normalize root WebSocket URLs

Ersaoktaviannn:
## Summary

- normalize the Hono Client `index` alias before constructing WebSocket URLs
- keep protocol translation, query serialization, and custom WebSocket providers unchanged
- preserve real path parameter values equal to `index`

## Before / after

| Typed route | Before | After |
| --- | --- | --- |
| `client.index.$ws()` for `/` | `ws://host/index` | `ws://host/` |
| nested root route | `ws://host/api/index` | `ws://host/api` |
| `/:id` with `id: "index"` | `ws://host/index` | unchanged |

## Impact

`index` is the client-side alias for a root route. `$get()`, `$url()`, and `$path()` already remove that synthetic segment, but `$ws()` did not. As a result, a typed WebSocket client for `/` attempted its handshake against `/index` (and a nested root against `/api/index`), which targets a different route and commonly returns `404`.

Normalization runs before path parameter substitution, so a genuine dynamic value named `index` is not removed.

## Verification

- `npx vitest --run src/client/client.test.ts` — 119 passed
- `npx vitest --run src/client/types.test.ts --coverage.enabled=false` — 13 passed
- Prettier, ESLint, and `git diff --check` on the changed files

- [x] Add tests
- [x] Run focused tests and type checking
- [ ] Add TSDoc/JSDoc — no public API added

### DIFF
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -1548,25 +1548,37 @@ describe('WebSocket URL Protocol Translation', () => {
   it('Translates HTTP to ws', async () => {
     const client = hc<AppType>('http://localhost')
     client.index.$ws()
-    expect(webSocketMock).toHaveBeenCalledWith('ws://localhost/index')
+    expect(webSocketMock).toHaveBeenCalledWith('ws://localhost/')
   })
 
   it('Translates HTTPS to wss', async () => {
     const client = hc<AppType>('https://localhost')
     client.index.$ws()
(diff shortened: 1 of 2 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@Ersaoktaviannn Thanks!

PR #5292, opened 2026-08-24
fix(aws-lambda): preserve empty query parameters

Ersaoktaviannn:
## Summary

- preserve present-but-empty query values when reconstructing API Gateway v1 and ALB requests
- continue omitting absent (`undefined`) entries
- add focused coverage for both affected AWS event formats

## Before / after

| Event input | Before | After |
| --- | --- | --- |
| v1 `{ flag: "" }` | query omitted | `?flag=` |
| ALB `{ flag: "" }` | query omitted | `?flag=` |
| `{ missing: undefined }` | omitted | unchanged |

## Impact

AWS event types distinguish an empty string from an absent value, but the single-value v1 and ALB processors filtered both through a truthiness check. This made `?flag=` indistinguishable from no `flag`, changing `URLSearchParams.has()`, request validation, feature-flag, and logging behavior before the request reached the Hono application.

The fix uses the declared `undefined` boundary. API Gateway v2 raw queries, multi-value queries, Lattice events, v1 encoding, and ALB serialization remain unchanged. AWS documents the relevant event mappings for [API Gateway proxy integrations](https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html) and [ALB Lambda targets](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html).

## Verification

- `npx vitest --run src/adapter/aws-lambda/handler.test.ts` — 40 passed
- Prettier, ESLint, and `git diff --check` on the changed files

- [x] Add tests
- [x] Run focused tests and type checking
- [ ] Add TSDoc/JSDoc — no public API added

### DIFF
--- a/src/adapter/aws-lambda/handler.test.ts
+++ b/src/adapter/aws-lambda/handler.test.ts
@@ -231,6 +231,45 @@ describe('EventProcessor.createRequest', () => {
     expect(url.searchParams.get('ampersand')).toBe('a&b&c')
   })
 
+  it('Should preserve empty query parameters for version 1.0', () => {
+    const event: LambdaEvent = {
+      ...baseV1Event,
+      queryStringParameters: {
+        empty: '',
+        present: '0',
+        omitted: undefined,
+      },
+    }
+
+    const request = getProcessor(event).createRequest(event)
+
+    expect(request.url).toBe(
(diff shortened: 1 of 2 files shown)


yusukebe (approved):
LGTM!

PR #5300, opened 2026-08-25
fix(types): allow symbol keys in Context<any> get and set fallbacks

### LINKED ISSUE #5298: Widen Context<any>'s untyped get/set fallback to accept symbol keys

`Context<E>`'s untyped fallback (when `E` is `any`, i.e. plain `new Hono()` / `Context` with no `Variables` generic) resolves `get`/`set` to a `Record<string, any>`-keyed shape:

```ts
// context.d.ts
set: Set<IsAny<E> extends true ? {
    Variables: ContextVariableMap & Record<string, any>
    ...
```

`Record<string, any>`'s key type only covers `string | number`, so calling `c.set`/`c.get` with a `symbol` key fails to compile — the same `Argument of type 'unique symbol' is not assignable to parameter of type 'string'` shape as #2484, but for `symbol` instead of a plain string.

The runtime has no such restriction: `set`/`get` store into a `Map` internally, which accepts a `symbol` key without any issue. This is purely a type-level gap, not a real limitation.

**Motivating case:** small frameworks building on top of Hono (mine included — a thin DI/routing layer) often want to stash an internal per-request marker on the context using a `Symbol()` key specifically *because* it can't collide with any string key a user of the framework might set. Today that requires an `as never` cast to work around the type, even though nothing at runtime objects.

**Proposed change:** widen the `IsAny<E>` fallback so it also accepts `symbol` keys, e.g.:

```ts
Variables: ContextVariableMap & Record<string, any> & { [key: symbol]: any }
```

This only changes the *untyped* `Context<any>` branch — anyone using `Hono<{ Variables: {...} }>()` with a concrete `Variables` map is on a different branch of the conditional type and is unaffected, so this shouldn't weaken any existing typed usage.

Happy to open a PR for this if the direction looks right to you.

lx3133584:
## Problem

Fixes #5298.

Calling `c.set(sym, val)` or `c.get(sym)` with a `symbol` key on an untyped `Context<any>` (the default when no generic `Variables` map is passed) fails type checking because the fallback was typed as `Record<string, any>`, which rejects `symbol` keys.

## Cause

`Context<E>`'s untyped fallback in `src/context.ts` typed `Variables` as `ContextVariableMap & Record<string, any>` and parameter keys as `string`. At runtime, `this.#var` is a `Map` that natively accepts `symbol` keys.

## Fix

- Widen the untyped `Context<any>` fallback variable map to `Record<string | symbol, any>` and allow `string | symbol` parameter keys for `get` and `set`.
- Add unit test in `src/context.test.ts` verifying `c.set` and `c.get` with symbol keys.

## Testing

- Verified type check with `./node_modules/.bin/tsc -p tsconfig.spec.json --noEmit`.
- Verified tests with `bun x vitest run src/context.test.ts` (60/60 passing).

### DIFF
--- a/src/context.test.ts
+++ b/src/context.test.ts
@@ -144,12 +144,11 @@ describe('Context', () => {
     expect(foo).toBe('Bar, Buzz')
   })
 
-
-  it("c.set() and c.get() with symbol key", () => {
-    const sym = Symbol("key")
+  it('c.set() and c.get() with symbol key', () => {
+    const sym = Symbol('key')
     expect(c.get(sym)).toBe(undefined)
-    c.set(sym, "value")
-    expect(c.get(sym)).toBe("value")
+    c.set(sym, 'value')
+    expect(c.get(sym)).toBe('value')
   })
 
   it('c.set() and c.get()', async () => {
--- a/src/context.test.ts
+++ b/src/context.test.ts
@@ -144,6 +144,14 @@ describe('Context', () => {
     expect(foo).toBe('Bar, Buzz')
   })
 
+
+  it("c.set() and c.get() with symbol key", () => {
+    const sym = Symbol("key")
+    expect(c.get(sym)).toBe(undefined)
+    c.set(sym, "value")
+    expect(c.get(sym)).toBe("value")
+  })
+
   it('c.set() and c.get()', async () => {
     expect(c.get('foo')).toBe(undefined)
     c.set('foo', 'bar')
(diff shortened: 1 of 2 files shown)


BarryThePenguin:
The [`PropertyKey`](https://www.totaltypescript.com/concepts/propertykey-type) type is also an option

lx3133584:
Updated to use `PropertyKey` instead of `string | symbol`. Thanks for the suggestion!

yusukebe (approved):
LGTM!

yusukebe:
@lx3133584 @BarryThePenguin 

I have never known `PropertyKey`! Thanks!

PR #5304, opened 2026-08-26
fix(timeout): prevent unhandled rejection on late downstream failures

spellsaif:
When `timeoutPromise` rejects in `Promise.race([next(), timeoutPromise])`, the middleware exits immediately without attaching a rejection handler to `next()`. If the background handler encounters an error later (e.g. database disconnect or rejected fetch), an `unhandledRejection` event is emitted on the process, which terminates modern Node.js processes.

### Solution
Attached a silent `.catch(() => {})` handler to `nextPromise` inside the timeout trigger. Downstream errors occurring *before* the timeout continue to be caught and handled normally by `Promise.race`.

- [X] Add tests
- [X] Run tests

### DIFF
--- a/src/middleware/timeout/index.test.ts
+++ b/src/middleware/timeout/index.test.ts
@@ -1,3 +1,4 @@
+import { vi } from 'vitest'
 import type { Context } from '../../context'
 import { Hono } from '../../hono'
 import { HTTPException } from '../../http-exception'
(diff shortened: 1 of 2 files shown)


usualoma:
Hi @spellsaif,

I applied only the added tests to the current `main` branch, without the implementation change, and they all passed.

Could you clarify the circumstances in which the reported `unhandledRejection` occurs? A minimal reproduction or a test that fails on `main` would be helpful.

spellsaif:
Konnichiwa @usualoma san,

Thanks for verifying and pointing this out!

You are completely right. I looked closely at src/compose.ts and realized that dispatch() wraps downstream execution in try ... catch and routes errors through onError. Because of that, the promise returned by next() resolves with the context rather than rejecting, so no unhandledRejection occurs under standard usage.

Since Hono's composition pipeline already handles this, this PR is unnecessary. I'll go ahead and close it.

Thanks again for your time and review!

PR #5305, opened 2026-08-26
fix(jwt): reject non-JSON JWKS responses

Amaresh:

- [x] Add tests
- [x] Run tests

### Description

`verifyWithJwks` already throws a clear error when the JWKS HTTP response is not `ok`, when `keys` is missing, and when `keys` is not an array. A 200 body that is not JSON (`response.json()` rejecting) still threw a raw `SyntaxError` with no indication that the JWKS URI was the source.

This wraps `response.json()` and rethrows `failed to parse JWKS JSON from ${jwks_uri}`, same shape as the existing fetch failure.

`bunx vitest --run src/utils/jwt/jwt.test.ts` — 89 passed.

### DIFF
--- a/src/utils/jwt/jwt.test.ts
+++ b/src/utils/jwt/jwt.test.ts
@@ -1514,6 +1514,35 @@ describe('verifyWithJwks key handling', () => {
 
     expect(localKeys).toEqual(originalKeys)
   })
+
+  it('Should throw when JWKS response is not valid JSON', async () => {
+    const originalFetch = globalThis.fetch
+    const header = Buffer.from(
+      JSON.stringify({ alg: 'RS256', typ: 'JWT', kid: 'unknown-key' })
+    ).toString('base64url')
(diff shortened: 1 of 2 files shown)


MILLERMARRU (commented):
Confirmed this against current `main`: `verifyWithJwks` already threw clear errors for a non-`ok` response and for a missing/non-array `keys` field, but `response.json()` rejecting on a non-JSON 200 body bubbled up as a bare `SyntaxError` with no indication the JWKS fetch was the source.

The fix wraps just that one `await response.json()` call and rethrows with the same message shape as the existing `!response.ok` branch (`failed to parse JWKS JSON from ${jwks_uri}`), so callers get a consistent error surface no matter which of the three ways JWKS fetching can fail. Minimal, doesn't touch the success path, and the new test actually exercises a 200 response with a non-JSON body (not just a network failure), which is the specific gap being closed. Ran the referenced `jwt.test.ts` file locally against the diff, all 90 pass.

Good to merge from my side.

PR #5307, opened 2026-08-27
fix(routing): collapse a trailing ** into a single * before routing

### LINKED ISSUE #4623: Inconsistent behavior of /** syntax in TrieRouter vs other routers

yfwmaniish:
Fixes #4623.

## What

`TrieRouter` and `LinearRouter` both 404 on a route registered with a trailing `**` (e.g. `.get('/auth/**', ...)`), while `RegExpRouter` and `PatternRouter` happen to accept it — matching the issue's own cross-router repro. A trailing `*` already matches every remaining path segment in Hono, so `**` isn't meant to be a semantically distinct pattern; it should just behave like `*`.

I re-ran the issue's own repro against current `main` before touching anything, since the issue only reported `TrieRouter` as broken: it turns out `LinearRouter` *also* 404s on `main` today (only `RegExpRouter`/`PatternRouter` currently work), so the fix needed to cover both.

## Root cause

Same class of bug in each router, different mechanism:
- `TrieRouter`'s `getPattern()` only recognizes the exact string `'*'` as a wildcard pattern. For a `'**'` segment it returns `null`, so `insert()` never registers that child as a pattern node — it's just a dead literal-string child that nothing can match against.
- `LinearRouter`'s `hasStar` branch strips exactly one trailing `*` character (`routePath.slice(0, -1)`) before splitting on `*`. For `/auth/**` that leaves one `*` behind, which then gets treated as an ordinary mid-path wildcard segment expecting a following `/` — which doesn't exist, so the match fails.

## Fix

Rather than patching two different routers' internal matching logic separately, I normalized once at the single place every router implementation already passes through: `Hono#addRoute()`, where the request path is finalized right before `router.add()`. Added `collapseWildcard()` next to the existing `mergePath()` in `utils/url.ts` (same small-pure-function convention), and applied it there.

```ts
export const collapseWildcard = (path: string): string => path.replace(/\*{2,}$/, '*')
```

## Testing

- `src/utils/url.test.ts`: unit tests for `collapseWildcard` (collapses 2+ trailing `*`, leaves a single trailing `*` alone, leaves non-trailing occurrences alone).
- `src/hono.test.ts`: a parametrized test registering `/auth/**` and requesting `/auth/hello` against all four routers (`RegExpRouter`, `TrieRouter`, `PatternRouter`, `LinearRouter`) — the issue's own repro, generalized.
- **Negative control**: reverted just `hono-base.ts`/`utils/url.ts` (kept the new tests) and reran — exactly the `TrieRouter` and `LinearRouter` cases fail with the reported 404, `RegExpRouter`/`PatternRouter` keep passing on their own. Restored the fix, all green again.
- Full suite (`vitest run`): 4956/4967 passing. The 11 failing are pre-existing and unrelated — ANSI color-code string assertions in `middleware/logger/index.test.ts`, confirmed identical with and without this change via `git stash`.
- `tsc -p tsconfig.spec.json`: clean.
- `eslint` on the four changed files: clean.

## 🤖 Agent context

Since `docs/CONTRIBUTING.md`'s AI Usage Policy asks that AI-assisted contributions not waste maintainer time regardless of disclosure, I'm disclosing anyway: this fix was developed with AI assistance (Claude Code) and reviewed before submission, with the verification steps above actually run (built/reverted/rerun, not just described) before opening this PR, including catching that `LinearRouter` was also affected — something the original issue didn't report.

### DIFF
--- a/src/hono-base.ts
+++ b/src/hono-base.ts
@@ -26,7 +26,7 @@ import type {
   Schema,
 } from './types'
 import { COMPOSED_HANDLER } from './utils/constants'
-import { getPath, getPathNoStrict, mergePath } from './utils/url'
+import { collapseWildcard, getPath, getPathNoStrict, mergePath } from './utils/url'
 
 const notFoundHandler: NotFoundHandler = (c) => {
   return c.text('404 Not Found', 404)
@@ -386,7 +386,7 @@ class Hono<
   }
 
   #addRoute(method: string, path: string, handler: H, baseRoutePath?: string): void {
-    path = mergePath(this._basePath, path)
+    path = collapseWildcard(mergePath(this._basePath, path))
     const r: RouterRoute = {
       basePath:
         baseRoutePath !== undefined ? mergePath(this._basePath, baseRoutePath) : this._basePath,
--- a/src/hono.test.ts
+++ b/src/hono.test.ts
@@ -7,6 +7,8 @@ import { Hono } from './hono'
 import { HTTPException } from './http-exception'
 import { logger } from './middleware/logger'
 import { poweredBy } from './middleware/powered-by'
+import { LinearRouter } from './router/linear-router'
+import { PatternRouter } from './router/pattern-router'
 import { RegExpRouter } from './router/reg-exp-router'
 import { SmartRouter } from './router/smart-router'
 import { TrieRouter } from './router/trie-router'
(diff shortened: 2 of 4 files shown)


MILLERMARRU (commented):
Reproduced the issue's own repro against current `main` before reading the fix: `TrieRouter` 404s on a route registered with `/auth/**`, and confirmed `LinearRouter` also 404s today even though the linked issue only mentioned `TrieRouter`, so the PR description is accurate that both need the fix, not just one.

I like that the fix normalizes once at `Hono#addRoute()` instead of patching `TrieRouter.getPattern()` and `LinearRouter`'s `hasStar` branch separately, since those two routers fail via genuinely different mechanisms (`TrieRouter` doesn't recognize `'**'` as a pattern at all and treats it as a dead literal child, `LinearRouter` strips only one trailing `*` and treats the leftover one as a mid-path segment expecting a following `/`). `collapseWildcard`'s regex (`/\*{2,}$/`) is trailing-only by design, which is correct since the bug report and the fix are both specifically about a route ending in `**`, a non-trailing occurrence like `/foo/**/bar` is a different (and much weirder) case that's out of scope here and the tests correctly don't claim to cover it.

Ran the parametrized test across all four routers locally, passes. Good fix.

PR #5311, opened 2026-08-28
fix(accepts, language): skip accept entries with quality 0 when matching

### LINKED ISSUE #5310: accepts() and the language detector match entries with q=0 that the client explicitly rejected
### Description

### Steps to Reproduce

```ts
const app = new Hono()

app.get('/', (c) =>
  c.text(
    accepts(c, {
      header: 'Accept',
      default: 'text/html',
    })
  )
)

})
```

```ts
```

### Expected Behavior

### Environment

LuckTerence:
`accepts()` (`defaultMatch`) and the language detector's header detection still match entries with quality `0`, which RFC 9110 §12.5.1/§12.5.4 defines as "explicitly not acceptable" — e.g. `Accept: application/json;q=0` was served `application/json`.

This change:

- `src/helper/accepts/accepts.ts` — `defaultMatch` filters out `q=0` entries before matching
- `src/middleware/language/language.ts` — `detectFromHeader` skips `q=0` languages

This aligns both consumers with the rule the `compress` middleware already applies (`q > 0` in `getEncoding`). `parseAccept` / `parseAcceptLanguage` are left unchanged, so custom `match` functions and direct callers still see the full list.

- [x] Add tests
- [x] Run tests
- [ ] Add [TSDoc](https://tsdoc.org/)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code (no public API change)

Tests: full suite `vitest --run` → 147 files, 4964 passed; `tsc -p tsconfig.spec.json` clean; new regression tests cover both `accepts()` and `languageDetector`.

Fixes #5310

### DIFF
--- a/src/helper/accepts/accepts.test.ts
+++ b/src/helper/accepts/accepts.test.ts
@@ -190,6 +190,38 @@ describe('accepts', () => {
     const result = accepts(c, options)
     expect(result).toBe('application/xml')
   })
+
+  test('should not match entries explicitly rejected with q=0', () => {
+    const c = {
+      req: {
+        header: () => 'application/json;q=0',
+      },
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } as any
+    const options: acceptsConfig = {
+      header: 'Accept',
+      supports: ['application/json'],
+      default: 'text/html',
+    }
+    const result = accepts(c, options)
+    expect(result).toBe('text/html')
+  })
+
+  test('should skip q=0 entries and match the remaining ones', () => {
+    const c = {
+      req: {
+        header: () => 'text/html;q=0.5, application/json;q=0',
+      },
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } as any
+    const options: acceptsConfig = {
+      header: 'Accept',
+      supports: ['application/json'],
+      default: 'application/octet-stream',
+    }
+    const result = accepts(c, options)
+    expect(result).toBe('application/octet-stream')
+  })
 })
 
 describe('Usage', () => {
--- a/src/helper/accepts/accepts.test.ts
+++ b/src/helper/accepts/accepts.test.ts
@@ -222,6 +222,22 @@ describe('accepts', () => {
     const result = accepts(c, options)
     expect(result).toBe('application/octet-stream')
   })
+
+  test('should not match a subtype wildcard the client rejected with q=0', () => {
+    const c = {
+      req: {
+        header: () => 'application/*;q=0',
+      },
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } as any
+    const options: acceptsConfig = {
+      header: 'Accept',
+      supports: ['application/json'],
+      default: 'text/html',
+    }
+    const result = accepts(c, options)
+    expect(result).toBe('text/html')
+  })
 })
 
 describe('Usage', () => {
--- a/src/helper/accepts/accepts.ts
+++ b/src/helper/accepts/accepts.ts
@@ -45,12 +45,16 @@ const getSpecificity = (type: string): number => {
 
 export const defaultMatch = (accepts: Accept[], config: acceptsConfig): string => {
   const { supports, default: defaultSupport } = config
-  const sortedAccepts = accepts.slice().sort((a, b) => {
-    if (b.q !== a.q) {
-      return b.q - a.q
-    }
-    return getSpecificity(b.type) - getSpecificity(a.type)
-  })
+  // A quality value of 0 means "not acceptable" (RFC 9110 §12.5.1), so such
+  // entries must never match, the same rule the compress middleware applies.
+  const sortedAccepts = accepts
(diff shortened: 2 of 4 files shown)


MILLERMARRU (commented):
Checked this against RFC 9110 directly: §12.5.1 says a qvalue of `0` means "not acceptable" for `Accept`, and §12.5.4 says the same for `Accept-Language`. Both `defaultMatch` in `src/helper/accepts/accepts.ts` and `detectFromHeader` in `src/middleware/language/language.ts` were sorting `q=0` entries by weight instead of excluding them, so a client explicitly saying "never send me `application/json`" via `Accept: application/json;q=0` could still get served it if nothing else in `supports` matched better.

The fix is scoped correctly: it only changes the two consumer functions (`defaultMatch`, `detectFromHeader`), and leaves `parseAccept`/`parseAcceptLanguage` themselves untouched, so a custom `match` callback still sees the raw list with `q=0` entries included if it wants to handle that case itself. That matches how `compress`'s `getEncoding` already treats `q=0` (`q > 0` filter), so this brings the other two consumers in line with an already-established convention in the codebase rather than inventing a new one. Test coverage for both the plain-reject case and the "skip q=0, fall through to the next candidate" case looks right.

LuckTerence:
No checks have run on this PR yet — likely waiting on first-time-contributor workflow approval. Could a maintainer approve the run?

@MILLERMARRU thanks for checking this against RFC 9110 §12.5.1 and §12.5.4. Your note on scope is the reason the filter sits in `defaultMatch` and `detectFromHeader` rather than in `parseAccept`/`parseAcceptLanguage` — a custom `match` callback still receives the raw list with `q=0` entries and can decide how to treat them.

WhoamiI00:
Your PR came first (and #5321 after) — I duplicated this in #5319 before checking for existing PRs on the issue, which was my mistake. Closing mine in favour of this one.

One test case from mine that might be worth folding in, since neither of the other two PRs covers it — a **subtype wildcard** rejected with `q=0`:

```ts
test('should not match a subtype wildcard the client rejected with q=0', () => {
  const c = {
    req: {
      header: () => 'application/*;q=0',
    },
  } as any
  const options: acceptsConfig = {
    header: 'Accept',
    default: 'text/html',
  }
  const result = accepts(c, options)
  expect(result).toBe('text/html')
})
```

This one genuinely discriminates: `matchType` returns `true` for `application/*` against `application/json` (unlike `*/*`, which it rejects outright), so without the `q > 0` filter it resolves to `application/json`. Worth having alongside the exact-type case.

Also worth noting for whoever reviews: `Accept: application/json;q=0, application/*;q=0.5` still resolves to `application/json` after this fix, because the wildcard matches it and nothing implements most-specific-range precedence. That needs a precedence rule rather than a `q` filter, so it's reasonably out of scope here — just flagging it so it isn't mistaken for a gap in the fix.

LuckTerence:
@WhoamiI00 Thanks for the suggestion! Folded that test case into the PR — confirms that subtype wildcards like `application/*;q=0` are properly skipped and fall back to default as expected.

WhoamiI00:
Thanks for folding it in — placement looks right.

One thing worth having in the record for whoever reviews: that case genuinely discriminates, which is not obvious from the diff. `matchType` returns `false` outright for `*/*` and `*`, so a `*/*;q=0` test would pass with or without the filter and prove nothing. `application/*` takes the `endsWith('/*')` branch and *does* match `application/json`, so without the `q > 0` filter that test resolves to `application/json` instead of the default. I verified it fails that way on `main` before suggesting it.

Still no checks here after nine days — the run is sitting on first-time-contributor approval. Might be worth pinging a maintainer again, since #5321 covers the same ground and this one has both the earliest timestamp and the fullest test coverage of the three.

gaijindev:
One extra regression case for `src/middleware/language/index.test.ts`: when the header contains only a rejected language, detection should continue to the next source and cache that result.

```ts
it('should continue to the next detector after a zero-quality language', async () => {
  const app = createTestApp({
    supportedLanguages: ['fr', 'en', 'ja'],
    fallbackLanguage: 'en',
    order: ['header', 'querystring'],
  })

  const res = await app.request('/?lang=ja', {
    headers: { 'accept-language': 'fr;q=0' },
  })
  expect(await res.text()).toBe('ja')
  expect(res.headers.get('set-cookie')).toContain('language=ja')
})
```

I ran this locally against main at `eebdf7b`: it returns `fr` and fails. Adding the same zero-quality guard in `detectFromHeader` makes it pass. Feel free to use it alongside the header-only fallback cases.

LuckTerence:
@gaijindev Thank you for the suggestion! I have added that regression test case to `src/middleware/language/index.test.ts` to ensure detection continues to the subsequent detector and caches the result when the header contains only a zero-quality language.

Friendly ping to the maintainers: could someone please approve the CI workflow run when you have a moment? Thank you!

yusukebe (approved):
LGTM!

yusukebe:
@LuckTerence Thank you!

PR #5314, opened 2026-08-28
fix(jsx): preserve self-closing void elements in cloneElement

adityajha77:
### Summary
Fixes an issue where `cloneElement()` on elements with `undefined` children (such as HTML void elements ``, `<input />`, `<br />`, `<hr />`, `<link />`) wrapped `undefined` into `[undefined]`, causing void elements to lose self-closing tag rendering in SSR.

### Problem
In `src/jsx/base.ts`, when `cloneElement` was called without new children, it extracted the existing element's `props.children`:

```ts
const c = (element as JSXNode).props.children
childrenToClone = Array.isArray(c) ? c : [c]

When an element has no children (e.g. <img src="foo.png" />), props.children is undefined. Because c was undefined, Array.isArray(c) ? c : [c] evaluated to [undefined], passing an array of length 1 containing undefined to jsx().

During SSR stringification (JSXNode.prototype.toStringToBuffer), the self-closing check emptyTags.includes(tag) && children.length === 0 failed because children.length was 1. This resulted in invalid non-self-closing HTML tags like <img src="foo.png"></img> instead of <img src="foo.png"/>.

###Solution

Check if c !== undefined before wrapping it into an array:

childrenToClone = c !== undefined ? (Array.isArray(c) ? c : [c]) : []

###Verification
Added a unit test in src/jsx/base.test.tsx verifying that cloning a void element (e.g. cloneElement(<img src="foo.png" />, { alt: 'bar' })) preserves <img src="foo.png" alt="bar"/> self-closing HTML output.

### DIFF
--- a/src/jsx/base.test.tsx
+++ b/src/jsx/base.test.tsx
@@ -36,7 +36,7 @@ describe('cloneElement', () => {
   })
 
   it('should preserve self-closing empty tag when cloned', () => {
-    const element = <img src="foo.png" />
+    const element = <img src='foo.png' />
     const clonedElement = cloneElement(element, { alt: 'bar' })
     expect(element.toString()).toBe('<img src="foo.png"/>')
     expect(clonedElement.toString()).toBe('<img src="foo.png" alt="bar"/>')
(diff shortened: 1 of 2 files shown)


usualoma:
Looks good to me. The fix makes sense and the test covers the issue.

adityajha77:
@usualoma thanks for reviewing!

adityajha77:
Rebased on latest main and added unit test coverage for array children in cloneElement to satisfy Codecov.

PR #5319, opened 2026-08-29
fix(accepts,language): never match an entry the client rejected with q=0

### LINKED ISSUE #5310: accepts() and the language detector match entries with q=0 that the client explicitly rejected

WhoamiI00:
Closes #5310.

- [x] Add tests
- [x] Run tests

## What

A quality value of `0` means the range is *explicitly not acceptable* (RFC 9110 §12.5.1, and §12.5.4 for `Accept-Language`). The `compress` middleware already honours this with its `q > 0` check in `getEncoding`, but the other two consumers of `parseAccept` did not.

I confirmed both against `main` before changing anything — each new test fails on the current code with exactly the reported symptom:

```
accepts:  AssertionError: expected 'application/json' to be 'text/html'
language: AssertionError: expected 'fr' to be 'en'
```

**`defaultMatch`** (`src/helper/accepts/accepts.ts`) sorted by quality but never excluded `q=0`, so `Accept: application/json;q=0` returned `application/json` — the one type the client said it would not take.

**`detectFromHeader`** (`src/middleware/language/language.ts`) destructured only `lang`, discarding `q` entirely, so `Accept-Language: fr;q=0` detected `fr` rather than falling through to the next source or the fallback.

## Fix

Both drop `q=0` entries, so all three `parseAccept` consumers now agree.

## Tests

Five added — three for `accepts`, two for the language detector:

| test | fails without the fix |
| --- | --- |
| `application/json;q=0` returns the default | yes |
| `application/*;q=0` returns the default | yes |
| `application/json;q=0, text/html;q=0.5` still returns `text/html` | no — guard |
| `fr;q=0` falls back to `en` | yes |
| `fr;q=0, es;q=0.5` detects `es` | no — guard |

The two marked "no" pass either way by design: `parseAccept` sorts by quality, so a `q=0` entry already loses to an acceptable one. They are there to catch over-correction — a fix that dropped too much would fail them.

```
vitest --run                 -> 147 files, 4966 passed | 44 skipped
prettier --check <4 files>   -> All matched files use Prettier code style
eslint src/helper/accepts src/middleware/language -> 0 errors
```

`tsc -p tsconfig.spec.json` reports 500 errors on my machine both with and without this change (`Type 'Uint8Array' is not generic`, etc.) — pre-existing local TS/lib skew, none in the files touched here.

## Notes

I have deliberately kept this to the reported behaviour. One adjacent case is out of scope: `Accept: application/json;q=0, application/*;q=0.5` still resolves to `application/json`, because the wildcard matches it and nothing implements most-specific-range precedence. That needs a precedence rule rather than a `q` filter, so it felt like a separate change — happy to follow up if you want it.

AI was used to help write this change; I have verified the behaviour, the RFC references, and the tests myself.

### DIFF
--- a/src/helper/accepts/accepts.test.ts
+++ b/src/helper/accepts/accepts.test.ts
@@ -70,6 +70,54 @@ describe('accepts', () => {
     expect(result).toBe('text/html')
   })
 
+  test('should not match a type the client rejected with q=0', () => {
+    const c = {
+      req: {
+        header: () => 'application/json;q=0',
+      },
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } as any
+    const options: acceptsConfig = {
+      header: 'Accept',
+      supports: ['application/json'],
+      default: 'text/html',
+    }
+    const result = accepts(c, options)
+    expect(result).toBe('text/html')
+  })
+
+  test('should fall through a q=0 type to the next acceptable one', () => {
+    const c = {
+      req: {
+        header: () => 'application/json;q=0, text/html;q=0.5',
+      },
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } as any
+    const options: acceptsConfig = {
+      header: 'Accept',
+      supports: ['application/json', 'text/html'],
+      default: 'application/xml',
+    }
+    const result = accepts(c, options)
+    expect(result).toBe('text/html')
+  })
+
+  test('should not match a subtype wildcard the client rejected with q=0', () => {
+    const c = {
+      req: {
+        header: () => 'application/*;q=0',
+      },
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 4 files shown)


WhoamiI00:
Closing this as a duplicate — #5311 was opened about a day before it and covers the same fix in the same two places. I didn't check for existing PRs against the issue before starting, which is on me.

I've left the one test case from here that neither #5311 nor #5321 covers as a comment on #5311 (a subtype wildcard rejected with `q=0`, which discriminates where `*/*` doesn't), so the useful part isn't lost. Sorry for the extra review noise.

PR #5326, opened 2026-08-31
refactor(on-handler): use forEach for consistent handler iteration

### LINKED ISSUE #5325: Refactor: Replace unused map() with forEach() in Hono.on() route registration
 Small cleanup in `Hono.on()` (`src/hono-base.ts`). 
Route handlers are registered with `handlers.map(...)`, but the return value is unused. The callback only calls `#addRoute`. In the same constructor, `get` / `use` already use `forEach` for the same kind of side-effect loop. 
### Proposed change 
```ts 

// before 
handlers.map((handler) => { 
       this.#addRoute(m.toUpperCase(), this.#path, handler) 
}) 

// after
 
const methodName = m.toUpperCase()  //-> to avoid the redundant uppercase calls 

handlers.forEach((handler) => { 
        this.#addRoute(methodName, this.#path, handler) 
})
```

### Why

- forEach matches intent (side effects, not transform)
- consistent with sibling registration code in the same file
- no API or behavior change

shk0Abdullah:

- [ ] Add tests
- [x] Run tests

Closes #5325

### DIFF
--- a/src/hono-base.ts
+++ b/src/hono-base.ts
@@ -146,8 +146,9 @@ class Hono<
       for (const p of [path].flat()) {
         this.#path = p
         for (const m of [method].flat()) {
-          handlers.map((handler) => {
-            this.#addRoute(m.toUpperCase(), this.#path, handler)
+          const methodName = m.toUpperCase()
+          handlers.forEach((handler) => {
+            this.#addRoute(methodName, this.#path, handler)
           })
         }
       }
--- a/src/hono-base.ts
+++ b/src/hono-base.ts
@@ -129,13 +129,14 @@ class Hono<
     const allMethods = [...METHODS, METHOD_NAME_ALL_LOWERCASE]
     allMethods.forEach((method) => {
       this[method] = (args1: string | H, ...args: H[]) => {
+        const methodName = method.toUpperCase()
         if (typeof args1 === 'string') {
           this.#path = args1
         } else {
-          this.#addRoute(method, this.#path, args1)
+          this.#addRoute(methodName, this.#path, args1)
         }
         args.forEach((handler) => {
-          this.#addRoute(method, this.#path, handler)
+          this.#addRoute(methodName, this.#path, handler)
         })
         return this as any
       }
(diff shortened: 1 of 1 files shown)


yusukebe:
Hi @usualoma 

What do you think of this?  I think this makes sense. It's only for the initialization phase, but it should be as fast as possible.

usualoma:
Hi @yusukebe,

Yeah, this makes sense to me. If we are going to optimize this initialization path, I wonder if it would be worth taking it a little further:

- `for...of` is likely preferable to `forEach` from a performance perspective, as it avoids the callback overhead.
- Since `#addRoute()` calls `.toUpperCase()` internally, the method name is currently normalized again for every handler. We could normalize it once at each call site and remove `.toUpperCase()` from `#addRoute()`.

Something along these lines:

```diff
diff --git a/src/hono-base.ts b/src/hono-base.ts
--- a/src/hono-base.ts
+++ b/src/hono-base.ts
@@ -129,13 +129,14 @@ class Hono<
     const allMethods = [...METHODS, METHOD_NAME_ALL_LOWERCASE]
     allMethods.forEach((method) => {
       this[method] = (args1: string | H, ...args: H[]) => {
+        const methodName = method.toUpperCase()
         if (typeof args1 === 'string') {
           this.#path = args1
         } else {
-          this.#addRoute(method, this.#path, args1)
+          this.#addRoute(methodName, this.#path, args1)
         }
         args.forEach((handler) => {
-          this.#addRoute(method, this.#path, handler)
+          this.#addRoute(methodName, this.#path, handler)
         })
         return this as any
       }
@@ -147,9 +148,9 @@ class Hono<
         this.#path = p
         for (const m of [method].flat()) {
           const methodName = m.toUpperCase()
-          handlers.forEach((handler) => {
+          for (const handler of handlers) {
             this.#addRoute(methodName, this.#path, handler)
-          })
+          }
         }
       }
       return this as any
@@ -385,7 +386,6 @@ class Hono<
   }

   #addRoute(method: string, path: string, handler: H, baseRoutePath?: string): void {
-    method = method.toUpperCase()
     path = mergePath(this._basePath, path)
```

shk0Abdullah:
Thanks for pointing this out @usualoma 🙌 I agree with your suggestions switched the handler iteration to for...of since it is more performant than forEach, moved the toUpperCase() normalization to the call sites so it only happens once, and removed the redundant normalization.

yusukebe (approved):
LGTM!

yusukebe:
@shk0Abdullah @usualoma 

Thanks!

PR #5327, opened 2026-08-31
fix(accepts, language): ignore entries rejected with q=0

### LINKED ISSUE #5310: accepts() and the language detector match entries with q=0 that the client explicitly rejected

safwenourabi:
Per RFC 9110 §12.5.1 (§12.5.4 for `Accept-Language`), `q=0` means the range is explicitly not acceptable. The `compress` middleware already honors this with its `q > 0` check in `getEncoding`, but the two other `parseAccept` consumers did not — `accepts()` matched `q=0` types instead of returning the configured default, and `languageDetector`'s header detection returned `q=0` languages instead of falling through.

This adds the same guard to both call sites, making all three consumers consistent.

One note for review: `parseQuality` maps the literal string `'NaN'` to `0`, so `q=NaN` now counts as rejected too. That seemed right for a malformed value, but happy to narrow it if you'd prefer.

Closes #5310

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/helper/accepts/accepts.test.ts
+++ b/src/helper/accepts/accepts.test.ts
@@ -234,6 +234,40 @@ describe('Usage', () => {
     expect(req4.headers.get('Content-Encoding')).toBeNull()
   })
 
+  test('should not match a type the client rejected with q=0', async () => {
+    const app = new Hono()
+    app.get('/', (c) =>
+      c.text(
+        accepts(c, {
+          header: 'Accept',
+          supports: ['application/json'],
+          default: 'text/html',
+        })
+      )
+    )
+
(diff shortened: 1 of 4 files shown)


safwenourabi:
Closing — I missed that #5311 already covers this (same fix, same two call sites), opened three days ago by the reporter of #5310. Sorry for the duplicate noise; #5311 is the one to review.

Issue #5306, opened 2026-08-27
jsx/dom: improve keyed children reconciliation for large list updates


kotarotaniguchi0523:
## What is the feature you are proposing?

I would like to improve the reconciliation performance of large keyed child lists in `hono/jsx/dom`.

The current reconciliation logic searches the existing children array with `findIndex()` and removes matched children with `splice()`.

This is not an issue for small lists, but it can scale poorly when large tables or log viewers handle selection, filtering, sorting, or reordering.

## Use case

For example, a parent component may manage both table rows and selection state:

```tsx
const [rows, setRows] = useState(initialRows)
const [selectedIds, setSelectedIds] = useState<string[]>([])

return (
  <>
    <button onClick={() => setSelectedIds([])}>
      Clear selection
    </button>

    <table>
      <tbody>
        {rows.map((row) => (
          <Row key={row.id} row={row} />
        ))}
      </tbody>
    </table>
  </>
)
```

Changing only the selection state causes the parent component to run again, so `rows.map()` creates the keyed children again.

Reordering also changes the DOM order in real applications:

```tsx
setRows((rows) => [...rows].reverse())
```

These patterns occur in data tables, log viewers, file lists, chat histories, and drag-and-drop interfaces.

## Current behavior

In a local Bun benchmark with 5,000 keyed elements, the current implementation produced:

- approximately 12.5 million array movements even when the order did not change
- approximately 12.5 million key comparisons when reversing the list
- approximately 12.5 million array movements when swapping the first and last elements

These numbers do not represent browser wall-clock time directly, but they show that the amount of reconciliation work can grow quadratically for large keyed lists.

## Expected behavior

Please reduce unnecessary linear searches and array operations during keyed-child reconciliation while preserving existing DOM elements and component state.

Possible approaches include:

- using an index structure that can efficiently find existing children by key
- avoiding repeated `splice()` operations on the old children array
- reusing the child at the current position when its key already matches
- keeping the additional overhead low for small lists

The implementation approach is not intended to be prescriptive. The important requirements are to preserve current DOM identity and state behavior.

## Compatibility requirements

The following behavior should remain unchanged:

- DOM element identity
- component state
- `ref` and effect cleanup
- focus and selection
- fragments
- controlled form values
- reconciliation behavior for unkeyed children
- current behavior for duplicate keys

This is a performance improvement for updates involving large keyed child lists, not a behavior change.

PR #5329, opened 2026-09-01
perf(jsx/dom): optimize matching-head child lookup during reconciliation

lx3133584:
## Problem
During JSX DOM children reconciliation in `hono/jsx/dom/render.ts`, each child in `children` searches `oldVChildren` using `findIndex()` and removes the matched node via `splice()`. When reconciling large child lists where the order is preserved or appended, this produces $O(N^2)$ linear searches and array shift operations.

## Root Cause
Reconciliation did not check the head of `oldVChildren` before falling back to full-array `findIndex()` and `splice()`.

## Fix
Check if `oldVChildren[0]` matches the current child (by string type, key+tag, or tag) and consume it directly via `oldVChildren.shift()`, reducing lookup from $O(N)$ to $O(1)$ for in-order children, while retaining `findIndex()` and `splice()` fallback for reordered or interspersed items.

Related to #5306

## Testing
1. Added unit test in `src/jsx/dom/index.test.tsx` verifying reconciliation and reordering across 100 keyed items.
2. Ran `npm test -- src/jsx/dom/` (all 24 test files pass, 1296 passed).

### DIFF
--- a/src/jsx/dom/index.test.tsx
+++ b/src/jsx/dom/index.test.tsx
@@ -1707,6 +1707,59 @@ describe('DOM', () => {
     )
   })
 
+  it('reconciles large keyed list updates and reordering efficiently', async () => {
+    let findIndexCalls = 0
+    const originalFindIndex = Array.prototype.findIndex
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    Array.prototype.findIndex = function (this: unknown[], ...args: [any]) {
+      findIndexCalls++
+      return originalFindIndex.apply(this, args)
+    }
+
+    try {
+      const ListApp = () => {
+        const [items, setItems] = useState(() => Array.from({ length: 100 }, (_, i) => `item-${i}`))
+        return (
+          <div>
+            <ul>
+              {items.map((item) => (
+                <li key={item}>{item}</li>
+              ))}
+            </ul>
+            <button id='reverse' onClick={() => setItems((prev) => [...prev].reverse())}>
+              reverse
+            </button>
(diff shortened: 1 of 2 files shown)


usualoma:
Most of the performance issues described in #5306 remain unresolved by this PR, so I think we should keep that issue open after merging.
That said, this change adds little complexity and provides a modest performance improvement, so I’m happy to merge it as an incremental improvement.

lx3133584:
Agreed, keeping #5306 open makes complete sense. Updated the PR description to decouple automatic issue closing so #5306 remains tracked for broader improvements.

usualoma:
Hi @yusukebe,

Sorry for not explaining that clearly. The optimization addressing the major issues pointed out in #5306 was #5340.
However, this PR, while touching the same area, involved a different optimization. I believe this is a useful improvement, even if the scope of the enhancement isn't huge. I think it would be best to merge the `main` branch, have it rewritten, and then have it reopened.
What do you think?

yusukebe:
@usualoma Thanks for the clarification! Makes sense. I'll reopen this.

@lx3133584 Sorry for closing this too early. #5340 has landed and changed this area, so could you merge `main` and rework this? Then we can review it again.

lx3133584:
Merged `main` and adapted the matching-head check to work alongside the key index introduced in #5340. Tests and benchmarks pass cleanly.

usualoma:
@lx3133584 Thank you. LGTM!

yusukebe (approved):
LGTM!

yusukebe:
@lx3133584 Thanks!

PR #5340, opened 2026-09-06
perf(jsx/dom): reduce lookup work for large keyed updates

### LINKED ISSUE #5306: jsx/dom: improve keyed children reconciliation for large list updates

usualoma:
fixes #5306.

Reduce repeated linear searches during large keyed child updates while preserving existing node reuse, component state, and cleanup order.

- Keep the existing forward `findIndex()` / `splice()` path for inexpensive updates.
- Try building a key index once accumulated search work exceeds a linear budget.
- Use the index only when the remaining old children have unique keys, including at most one `undefined` key.
- Preserve strict key and tag matching, distinguish text nodes from elements, and retain the original removal order.
- Fall back to array matching for new unkeyed elements so they can still reuse keyed old children with the same tag.

## Scope and relationship to #5329

This targets expensive searches during large keyed reorders. It does not make all reconciliation linear: in-order array shifts remain, and duplicate-key or unkeyed fallback paths can still require quadratic work. DOM placement is unchanged.

This branch is based directly on main and does not include #5329. That PR optimizes matching-head updates; this one addresses expensive searches. If #5329 lands first, this branch can be rebased to preserve its fast path alongside the indexed path.

## Benchmark

```sh
bun benchmarks/jsx-dom/benchmark.ts
```

The benchmark measures `build()` only, excluding JSX creation and DOM updates. It covers 100, 1,000, and 5,000 keyed children.

The same benchmark script was run against main (`eebdf7be`) and this branch (`0130dd16`) using **Bun 1.3.12 on an Apple M5 Pro**. Results below are for **5,000 keyed children**, with one additional unkeyed child in the footer case.

| Update | main (ms/build) | This PR (ms/build) |
| --- | ---: | ---: |
| Same order | 1.389 | 1.428 |
| Remove the second child | 1.564 | 1.586 |
| Adjacent swap | 1.362 | 1.413 |
| Reverse | 38.586 | 0.753 |
| Reverse with an unkeyed footer | 38.812 | 0.765 |

Values are medians of three separate process runs, with the comparison order reversed in the second run. Each run reports the median of five fresh-node batches, each with 10 warm-up updates and 20 measured updates.

These results demonstrate the benefit for large keyed reorders, rather than a general speedup for every update. They are local microbenchmark results, not browser rendering times.

- [x] Add tests
- [x] Run tests

### DIFF
--- a/benchmarks/jsx-dom/benchmark.ts
+++ b/benchmarks/jsx-dom/benchmark.ts
@@ -1,0 +1,65 @@
+// Run from the repository root: bun benchmarks/jsx-dom/benchmark.ts
+// Measures build() only, excluding JSX creation and DOM updates.
+import { jsx } from '../../src/jsx/dom/jsx-runtime'
+import type { NodeObject } from '../../src/jsx/dom/render'
+import { build, buildNode } from '../../src/jsx/dom/render'
+
+for (const size of [100, 1000, 5000]) {
+  for (const kind of [
+    'same-order',
+    'remove-second',
+    'adjacent-swap',
+    'reverse',
+    'reverse-unkeyed-footer',
+  ]) {
+    const ids = Array.from({ length: size }, (_, i) => i)
+    const hasFooter = kind === 'reverse-unkeyed-footer'
+    const createChildren = (ids: number[]) => {
+      const children = ids.map((id) => jsx('p', {}, String(id)))
+      if (hasFooter) {
+        children.push(jsx('p', {}))
+      }
+      return children
+    }
+    const nextIds = [...ids]
+    if (kind === 'remove-second') {
+      nextIds.splice(1, 1)
+    } else if (kind === 'adjacent-swap') {
+      ;[nextIds[1], nextIds[2]] = [nextIds[2], nextIds[1]]
+    } else if (kind.startsWith('reverse')) {
+      nextIds.reverse()
+    }
+    const samples: number[] = []
+    // Use fresh nodes in short batches instead of mixing long-lived node states.
+    for (let batch = 0; batch < 5; batch++) {
+      const parent = buildNode(jsx('div', { children: createChildren(ids) })) as NodeObject
+      build([], parent)
+      const initialChildren = parent.vC
+      const update = (children: ReturnType<typeof createChildren>) => {
+        parent.vC = initialChildren
+        parent.props.children = children
+        build([], parent)
+      }
+      for (let i = 0; i < 10; i++) {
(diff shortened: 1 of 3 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@usualoma 

Thank you! (This week, I am a little bit busy, so there is a little time for reviewing)

PR #5351, opened 2026-09-08
fix(aws-lambda): respect backpressure when streaming the response body

### LINKED ISSUE #5333: AWS Lambda streamHandle ignores writable backpressure and resolves before the response finishes

4.13.5

AWS Lambda response streaming with the Node.js 24 runtime.

`streamHandle` manually reads the response body and calls `writer.write()` without checking its boolean return value or waiting for the `drain` event:

```ts
const streamToNodeStream = async (
  reader: ReadableStreamDefaultReader<Uint8Array>,
  writer: NodeJS.WritableStream
): Promise<void> => {
  let readResult = await reader.read()
  while (!readResult.done) {
    writer.write(readResult.value)
    readResult = await reader.read()
  }
  writer.end()
}
```

When the response is larger than the destination stream's `highWaterMark`, `writer.write()` returns `false`. The current loop continues reading and writing anyway.

The function also resolves after calling `writer.end()`, without waiting for the writable stream to finish.

This standalone reproduction uses a slow writable stream with a 16 KiB high-water mark and a 2 MiB Hono response:

```ts
import { Writable } from 'node:stream'
import { Hono } from 'hono'
import { streamHandle } from 'hono/aws-lambda'

const awsLambdaRuntime = {
  streamifyResponse: (handler) => handler,
  HttpResponseStream: {
    from: (responseStream) => responseStream,
  },
}

Reflect.set(globalThis, 'awslambda', awsLambdaRuntime)

const chunkCount = 32
const chunk = new Uint8Array(64 * 1024)
let emittedChunks = 0

const app = new Hono()

app.get(
  '/',
  () =>
    new Response(
      new ReadableStream({
        pull(controller) {
          if (emittedChunks < chunkCount) {
            emittedChunks += 1
            controller.enqueue(chunk)
          } else {
            controller.close()
          }
        },
      })
    )
)

let writtenBytes = 0

const responseStream = new Writable({
  highWaterMark: 16 * 1024,
  write(buffer, _encoding, callback) {
    setImmediate(() => {
      writtenBytes += buffer.length
      callback()
    })
  },
})

const event = {
  version: '2.0',
  routeKey: '$default',
  rawPath: '/',
  rawQueryString: '',
  headers: { host: 'example.test' },
  requestContext: {
    domainName: 'example.test',
    http: { method: 'GET' },
  },
  body: null,
  isBase64Encoded: false,
}

const handler = streamHandle(app)
await handler(event, responseStream, {})

console.log({
  writableFinished: responseStream.writableFinished,
  writtenBytes,
  expectedBytes: chunkCount * chunk.length,
})
```

With the current implementation, the handler can resolve while `responseStream.writableFinished` is still `false`. The destination has not necessarily processed all buffered data when the handler promise resolves.

`streamHandle` should:

1. Respect writable-stream backpressure.
2. Stop reading from the web stream while the destination cannot accept more data.
3. Resolve only after the destination stream finishes.
4. Propagate read, write, and pipeline errors correctly.

For the reproduction above, after awaiting the handler:

```ts
responseStream.writableFinished === true
writtenBytes === chunkCount * chunk.length
```

### What do you see instead?

The implementation ignores the return value of `writer.write()` and continues reading the entire response into the writable stream's internal buffer.

Calling `writer.end()` also does not mean that the stream has finished. As a result, `streamHandle` can resolve before the destination emits `finish`.

In AWS Lambda response streaming, this can increase memory usage for large or slow responses and risks the Lambda invocation completing before the response has been fully delivered.

### Additional information

We currently work around this by converting the web `ReadableStream` to a Node.js readable and using Node's `pipeline`:

```ts
import { Readable } from 'node:stream'
import { pipeline } from 'node:stream/promises'

const streamToNodeStream = (
  stream: ReadableStream<Uint8Array>,
  writer: NodeJS.WritableStream
): Promise<void> => pipeline(Readable.fromWeb(stream), writer)
```

The call site then passes the response body directly:

```ts
if (res.body) {
  await streamToNodeStream(res.body, responseStream)
}
```

`pipeline` handles backpressure, waits for the writable stream to finish, and propagates stream errors.

We tested this workaround with a writable stream whose `highWaterMark` is smaller than each response chunk. The test verifies that the handler does not resolve until all 2 MiB have been written and `writableFinished` is `true`.

speedsharmaai:
Fixes #5333.

## Problem

`streamToNodeStream` in the AWS Lambda adapter ignored the return value of `writer.write()`:

```ts
while (!readResult.done) {
  writer.write(readResult.value)
  readResult = await reader.read()
}
writer.end()
```

Two consequences:

1. **No backpressure.** Once the response exceeds the destination's `highWaterMark`, `write()` returns `false`, but the loop keeps reading and writing. The whole body accumulates in the writable's internal buffer.
2. **Resolves too early.** `writer.end()` only signals that no more data is coming — it does not mean the stream has finished. `streamHandle` could resolve while `responseStream.writableFinished` was still `false`, so a Lambda invocation could complete before the response had actually been delivered.

## Fix

Read the body through an async generator and hand it to `pipeline` from `node:stream/promises`. `pipeline` paces reads against the destination, waits for `finish`, and propagates errors from either side — which is the behaviour the issue asks for, without hand-rolling the stream lifecycle.

`node:` imports are already used elsewhere in the Lambda adapters (`src/adapter/lambda-edge/handler.ts` imports `node:crypto`), and `stream/promises` is available on every Node version that supports Lambda response streaming.

## Tests

`runtime-tests/lambda/stream-backpressure.test.ts` streams 2 MiB in 16 KiB chunks into a slow writable with a 16 KiB `highWaterMark` and asserts that:

- every byte is written by the time the handler resolves,
- `writableFinished` is `true`,
- the source never runs more than a bounded number of chunks ahead of the sink.

Against `main` the first assertion fails with `expected +0 to be 2097152` — the handler resolves before a single chunk has been flushed.

A second test covers a body that errors mid-stream, confirming the handler still logs and resolves rather than rejecting, so the error contract is unchanged.

Full suite passes: 148 files, 5026 tests.

### DIFF
--- a/runtime-tests/lambda/stream-backpressure.test.ts
+++ b/runtime-tests/lambda/stream-backpressure.test.ts
@@ -1,0 +1,107 @@
+import { Writable } from 'node:stream'
+import { streamHandle } from '../../src/adapter/aws-lambda/handler'
+import type { LambdaContext } from '../../src/adapter/aws-lambda/types'
+import { Hono } from '../../src/hono'
+
+const awslambda = {
+  streamifyResponse: <T>(handlerFunc: T): T => handlerFunc,
+  HttpResponseStream: {
+    from: (stream: Writable): Writable => stream,
+  },
+}
+
+vi.stubGlobal('awslambda', awslambda)
+
+type StreamingHandler = (
+  event: unknown,
+  responseStream: Writable,
+  context: LambdaContext
+) => Promise<void>
+
+const CHUNK_SIZE = 16 * 1024
+const CHUNK_COUNT = 128
+
+const event = {
+  headers: {},
+  rawPath: '/',
+  rawQueryString: '',
+  body: null,
+  isBase64Encoded: false,
+  requestContext: { http: { method: 'GET' } },
+}
+
+describe('streamHandle backpressure', () => {
+  it('should pace reads against the destination and resolve only once it has finished', async () => {
+    let emittedChunks = 0
+    let writtenBytes = 0
+    let maxChunksAhead = 0
+
+    const app = new Hono()
+    app.get(
+      '/',
+      () =>
+        new Response(
+          new ReadableStream({
+            pull(controller) {
+              if (emittedChunks < CHUNK_COUNT) {
+                emittedChunks++
+                controller.enqueue(new Uint8Array(CHUNK_SIZE))
+              } else {
+                controller.close()
+              }
+            },
+          })
+        )
+    )
+
+    const responseStream = new Writable({
+      highWaterMark: CHUNK_SIZE,
+      write(buffer: Buffer, _encoding, callback) {
+        maxChunksAhead = Math.max(maxChunksAhead, emittedChunks - writtenBytes / CHUNK_SIZE)
+        setImmediate(() => {
+          writtenBytes += buffer.length
+          callback()
+        })
+      },
+    })
+
+    const handler = streamHandle(app) as unknown as StreamingHandler
+    await handler(event, responseStream, {} as LambdaContext)
+
+    expect(writtenBytes).toBe(CHUNK_SIZE * CHUNK_COUNT)
+    expect(responseStream.writableFinished).toBe(true)
+    expect(maxChunksAhead).toBeLessThan(CHUNK_COUNT / 2)
+  })
+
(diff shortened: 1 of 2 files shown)


yusukebe (commented):
LGTM!

PR #5356, opened 2026-09-09
fix(aws-lambda): preserve empty request headers

Ersaoktaviannn:
## Summary

- preserve present-but-empty request headers in API Gateway v1, API Gateway v2, and ALB events
- keep omitting entries whose value is `undefined`

## Before / after

| Event input | Before | After |
| --- | --- | --- |
| `headers: { "x-empty": "" }` | header absent | `x-empty` is present with an empty value |
| `headers: { "x-empty": undefined }` | header absent | unchanged |

## Impact

The event types distinguish `""` from `undefined`. The previous truthiness checks dropped `""`, so middleware and handlers observed the header as absent. This change preserves that distinction without changing multi-value header precedence or non-ASCII sanitization.

## Verification

- focused regression test failed before the change (`headers.get("x-empty")` was `null`)
- `npx vitest --run src/adapter/aws-lambda/handler.test.ts` — 39 passed
- Prettier, ESLint, and `git diff --check`

## Checklist

- [x] Tests added
- [x] Focused tests run
- [x] Typecheck, formatting, and lint run
- [x] No public API change
- [ ] CI — workflows await maintainer approval for this fork

### DIFF
--- a/src/adapter/aws-lambda/handler.test.ts
+++ b/src/adapter/aws-lambda/handler.test.ts
@@ -269,6 +269,30 @@ describe('EventProcessor.createRequest', () => {
     })
   })
 
+  it('Should preserve empty header values', () => {
+    const v1Event: LambdaEvent = {
+      ...baseV1Event,
+      headers: { 'x-empty': '' },
+      multiValueHeaders: undefined,
+    }
+    const v2Event: LambdaEvent = {
+      ...baseV2Event,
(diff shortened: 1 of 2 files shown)


usualoma:
Closing this as AI slop. The description misleadingly suggests that API Gateway and ALB can send `undefined` header values. These events are JSON; a TypeScript type allowing `undefined` is not evidence that AWS sends it.

PR #5358, opened 2026-09-10
fix(lambda-edge): fail with a descriptive error on a malformed event

### LINKED ISSUE #4417: try catch cloudfront event -> request mapping for lambda@edge handler to avoid undefined access errors

As is, this mapper assumes the event is well formed and valid. 

When testing the lambda@edge via the aws console using the prefilled example json, which is just an array of foobar type dummy ids, the handler throws with "cannot access undefined when reading "0" as it tries to index into an array that doesnt exist. 

Ideally it should try catch and return an error message along the lines of "Unable to map cloudfront request to request object when handling request: malformed cloudfront edge event"

problematic code: 
```
const createRequest = (event: CloudFrontEdgeEvent): Request => {
  const queryString = event.Records[0].cf.request.querystring
  const host =
    event.Records[0].cf.request.headers?.host?.[0]?.value ||
    event.Records[0].cf.config.distributionDomainName
  const urlPath = `https://${host}${event.Records[0].cf.request.uri}`
  const url = queryString ? `${urlPath}?${queryString}` : urlPath

  const headers = new Headers()
  Object.entries(event.Records[0].cf.request.headers).forEach(([k, v]) => {
    v.forEach((header) => headers.set(k, header.value))
  })

  const requestBody = event.Records[0].cf.request.body
  const method = event.Records[0].cf.request.method
  const body = createBody(method, requestBody)

  return new Request(url, {
    headers,
    method,
    body,
  })
}
```

curl-kawsar:
Closes #4417

## Problem

`handle()` and `createRequest()` dereference `event.Records[0].cf.request` nine times with no guard. When the function is invoked with something that isn't a Lambda@Edge event, the failure surfaces as:

```
TypeError: Cannot read properties of undefined (reading '0')
```

That names neither the adapter nor the field that was missing, so there's nothing in the CloudWatch log pointing at the event shape as the cause.

## Change

Read the record once through a small `getCloudFrontRecord()` helper that throws a `TypeError` naming the expected field, and pass it down to `createRequest()`:

```
TypeError: Unable to map the CloudFront event to a Request: expected `Records[0].cf.request` in the Lambda@Edge event.
```

The function still throws, so the invocation still fails and Lambda still records the error — only the message changes. Behaviour for well-formed events is unchanged.

While in here, two crashes of the same class on *otherwise valid* events:

- `Object.entries(request.headers)` threw `Cannot convert undefined or null to object` when a request carried no `headers` key → `request.headers ?? {}`.
- `cf.config.distributionDomainName` was read without a guard on the host fallback path → `cf.config?.distributionDomainName`.

## Note on the open question in the issue

@watany-dev asked whether the payload the Lambda console prefills matches the documented Lambda@Edge event schema. It doesn't — so this PR deliberately doesn't try to *support* that payload. It only makes the resulting error attributable. That holds regardless of how the event came to be malformed, and the two `headers`/`config` fixes above are reachable from genuinely valid CloudFront events.

## Side effect on bundle size

Hoisting the repeated `event.Records[0].cf.request` chain into a local removes eight duplicated property-access chains from the minified output.

## Tests

Four new cases in `src/adapter/lambda-edge/handler.test.ts`:

- `Records` holding no CloudFront record (the console-style payload)
- `Records` empty, and absent entirely
- `cf` present but carrying no `request`
- a well-formed request with `headers` omitted → still resolves 200 and falls back to `distributionDomainName`

Verified red/green: with `handler.ts` at its pre-change state all four fail (three with the opaque property-access error, one with `Cannot convert undefined or null to object`); with the change all 21 tests in the file pass.

- `vitest --run --project main` → 4051 passed, 37 skipped, 123 files
- `bun run test:lambda-edge` → 15 passed
- `tsc -p tsconfig.spec.json` → clean

- [x] Add tests
- [x] Run tests

### DIFF
--- a/src/adapter/lambda-edge/handler.test.ts
+++ b/src/adapter/lambda-edge/handler.test.ts
@@ -365,3 +365,73 @@ describe('handle', () => {
     expect(res).not.toHaveProperty('bodyEncoding')
   })
 })
+
+describe('handle with a malformed event', () => {
+  const invalidEventMessage =
+    'Unable to map the CloudFront event to a Request: expected `Records[0].cf.request` in the Lambda@Edge event.'
+
+  it('Should reject with a descriptive error when Records holds no CloudFront record', async () => {
+    const app = new Hono()
+    const handler = handle(app)
+
+    // The payload the AWS Lambda console prefills is an array of dummy ids.
+    const event = { Records: ['foo', 'bar'] } as unknown as CloudFrontEdgeEvent
+
+    await expect(handler(event)).rejects.toThrow(TypeError)
+    await expect(handler(event)).rejects.toThrow(invalidEventMessage)
+  })
+
+  it('Should reject with a descriptive error when Records is empty or absent', async () => {
+    const app = new Hono()
+    const handler = handle(app)
+
+    for (const event of [{ Records: [] }, {}] as unknown as CloudFrontEdgeEvent[]) {
+      await expect(handler(event)).rejects.toThrow(invalidEventMessage)
+    }
+  })
+
+  it('Should reject with a descriptive error when cf carries no request', async () => {
+    const app = new Hono()
+    const handler = handle(app)
+
+    const event = {
+      Records: [{ cf: { config: { distributionDomainName: 'd111111abcdef8.cloudfront.net' } } }],
+    } as unknown as CloudFrontEdgeEvent
+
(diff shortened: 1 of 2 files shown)


yusukebe (approved):
LGTM!

yusukebe:
@curl-kawsar 

Thank you!

PR #5366, opened 2026-09-12
fix(request): keep the request media type when reusing a cached body

### LINKED ISSUE #5365: c.req.formData() throws if the body was already read with c.req.text()

4.13.5

Reading the body as text first, then parsing it as form data, throws.

```ts
import { Hono } from 'hono'

const app = new Hono()

app.post('/', async (c) => {
  const raw = await c.req.text() // read the raw body first, e.g. to verify a signature
  const form = await c.req.formData() // throws here
  return c.json({ foo: form.get('foo') })
})
```

```bash
curl -X POST http://localhost:3000/ \
  -H 'content-type: application/x-www-form-urlencoded' \
  -d 'foo=bar'
```

It reproduces without a server too:

```js
import { Hono } from 'hono'

const app = new Hono()
app.post('/', async (c) => {
  await c.req.text()
  return c.json({ foo: (await c.req.formData()).get('foo') })
})

await app.request('/', {
  method: 'POST',
  headers: { 'content-type': 'application/x-www-form-urlencoded' },
  body: 'foo=bar',
})
```

`c.req.formData()` should return the parsed form, the same as when it is the first call. The body is already cached, so reading it a second way should still work.

### What do you see instead?

```
TypeError: Content-Type was not one of "multipart/form-data" or "application/x-www-form-urlencoded".
```

### Additional information

For a `application/x-www-form-urlencoded` or `multipart/form-data` request:

| first call | then `formData()` |
| --- | --- |
| (nothing) | works |
| `text()` | throws |
| `arrayBuffer()` | throws |
| `bytes()` | throws |
| `blob()` | works |
| `parseBody()` | works |

`blob()` works because a `Blob` carries its own type, and `parseBody()` works because it reads the media type from `req.raw.headers`. So the media type is available — it just is not carried over when the cached body is rebuilt in `#cachedBody()`.

Happy to send a PR if this looks worth fixing.

xia-chao:
Fixes #5365

- [x] Add tests
- [x] Run tests

### What this fixes

Reading the body one way and then calling `c.req.formData()` threw, even though the body was already cached:

```ts
app.post('/', async (c) => {
  const raw = await c.req.text() // e.g. verify a signature over the raw body first
  const form = await c.req.formData() // TypeError
})
```

```
TypeError: Content-Type was not one of "multipart/form-data" or "application/x-www-form-urlencoded".
```

Same for `arrayBuffer()` and `bytes()` first, and same for `multipart/form-data`. It works when `formData()` is the first call, so the same request behaved differently depending on which accessor happened to run first.

### Why it happened

`#cachedBody()` rebuilds a missing representation from a cached one with a bare `Response`:

```ts
return new Response(body)[key]()
```

That `Response` has no `Content-Type`, so `Response.formData()` has nothing to decide the parser with. The original media type is still on `this.raw.headers`, it just was not passed along.

`c.req.blob()` was unaffected (a `Blob` keeps its own type) and `c.req.parseBody()` was unaffected (it reads the media type from `req.raw.headers`), which is why the inconsistency was easy to miss.

### The change

Carry the request's `Content-Type` over when rebuilding the body. `FormData` is excluded on purpose, because `Response` has to generate a fresh multipart boundary for it.

```ts
const contentType = body instanceof FormData ? undefined : raw.headers.get('content-type')
return new Response(body, {
  headers: contentType ? { 'Content-Type': contentType } : undefined,
})[key]()
```

### Tests

Added to `src/request.test.ts`, under *Body methods with caching*:

- `formData()` after `text()` / `arrayBuffer()` / `bytes()` / `blob()` for a urlencoded body
- `formData()` after `text()` for a `multipart/form-data` body
- the cached representation itself is still returned unchanged

Four of them fail before the change and pass after. The `blob()` one passes either way, since a `Blob` already carries its type.

- `vitest --run src/request.test.ts`: 50 passed
- full suite: 5125 passed, 44 skipped. The 12 failures in `color` / `logger` / `dev` / `compress(node)` are colour- and TTY-related and fail the same way on an unmodified checkout
- `tsc -p tsconfig.spec.json --noEmit`: clean

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -227,7 +227,15 @@ export class HonoRequest<P extends string = '/', I extends Input['out'] = {}> {
         if (anyCachedKey === 'json') {
           body = JSON.stringify(body)
         }
-        return new Response(body)[key]()
+        // Rebuilding the body through a bare `Response` loses the request's media
+        // type, so a representation that needs it (e.g. `formData()`) can no longer
+        // be produced even though the bytes are still available. Carry the original
+        // `Content-Type` over, except for `FormData`, where `Response` must generate
+        // a fresh multipart boundary of its own.
+        const contentType = body instanceof FormData ? undefined : raw.headers.get('content-type')
+        return new Response(body, {
+          headers: contentType ? { 'Content-Type': contentType } : undefined,
+        })[key]()
       })
     }
 
--- a/src/request.ts
+++ b/src/request.ts
@@ -232,7 +232,8 @@ export class HonoRequest<P extends string = '/', I extends Input['out'] = {}> {
         // be produced even though the bytes are still available. Carry the original
         // `Content-Type` over, except for `FormData`, where `Response` must generate
         // a fresh multipart boundary of its own.
-        const contentType = body instanceof FormData ? undefined : raw.headers.get('content-type')
+        const contentType =
+          anyCachedKey === 'formData' ? undefined : raw.headers.get('content-type')
         return new Response(body, {
           headers: contentType ? { 'Content-Type': contentType } : undefined,
         })[key]()
--- a/src/request.test.ts
+++ b/src/request.test.ts
@@ -377,6 +377,58 @@ describe('Body methods with caching', () => {
     expect(async () => await req.blob()).not.toThrow()
   })
 
+  describe('formData() after another representation has been cached', () => {
+    const urlencoded = 'application/x-www-form-urlencoded'
+    const body = 'foo=bar&baz=qux'
+
+    for (const first of ['text', 'arrayBuffer', 'bytes', 'blob'] as const) {
+      test(`req.formData() after req.${first}()`, async () => {
+        const req = new HonoRequest(
+          new Request('http://localhost', {
+            method: 'POST',
+            headers: { 'Content-Type': urlencoded },
+            body,
+          })
+        )
+        await req[first]()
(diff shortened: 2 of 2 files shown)


src/request.ts
```diff
@@ -227,7 +227,16 @@ export class HonoRequest<P extends string = '/', I extends Input['out'] = {}> {
         if (anyCachedKey === 'json') {
           body = JSON.stringify(body)
         }
-        return new Response(body)[key]()
+        // Rebuilding the body through a bare `Response` loses the request's media
+        // type, so a representation that needs it (e.g. `formData()`) can no longer
+        // be produced even though the bytes are still available. Carry the original
+        // `Content-Type` over, except for `FormData`, where `Response` must generate
+        // a fresh multipart boundary of its own.
+        const contentType =
```
yusukebe, line 235: Switched to `anyCachedKey === 'formData'` for consistency with the `'json'` check above.

yusukebe (approved):
LGTM!

yusukebe:
@xia-chao Thank you for the issue and PR! Looks good.

xia-chao:
> @xia-chao感谢您提交的问题和公关稿！看起来不错。

Hono!
Next, we will frequently "meet" (here only issues, PR),
Because in the future, JavaScript servers will no longer be exclusive to Node.js, but will belong to a unified web runtime.

PR #5371, opened 2026-09-12
fix(trie-router): don't run a handler twice for a literal `*` path segment

### LINKED ISSUE #5370: TrieRouter runs a handler twice when a path segment is literally `*`
A handler runs twice when the last segment of the request path is literally `*`.

```ts
import { Hono } from 'hono'
import { TrieRouter } from 'hono/router/trie-router'

let calls = 0
app.use('/a/*', async (c, next) => { calls++; await next() })
app.all('*', (c) => c.text('ok'))

await app.request('http://localhost/a/b')   // calls === 1
calls = 0
await app.request('http://localhost/a/*')   // calls === 2
```

`/a/b` runs the middleware once. `/a/*` runs it twice.

This is reachable without choosing TrieRouter explicitly. SmartRouter falls back to it whenever RegExpRouter throws `UnsupportedPathError`, so this app on the default router hits it too:

```ts
const app = new Hono()
app.use('/a/*', async (c, next) => { calls++; await next() })
app.get('/a/b', (c) => c.text('static'))
app.get('/a/:p', (c) => c.text('param'))
// getRouterName(app) === 'SmartRouter + TrieRouter'
// GET /a/*  ->  middleware runs twice
```

### Cause

In `insert()` a pattern node is stored twice: under its own key in `#children`, and in `#patterns`.

```ts
const child = (curNode.#children[key] ||= new Node())
if (pattern && !child.#pattern) {
  child.#pattern = pattern
  curNode.#patterns.push(child)
}
```

In `search()`, a request part of `*` finds that node through `node.#children[part]` and pushes its handlers, then the `#patterns` loop reaches the same node and pushes them again. Any other part misses the `#children` lookup and only takes the pattern branch, which is why only a literal `*` duplicates.

The other three routers return the handler once for this path.

I have a fix that skips the pattern branch when it resolves to the node the `#children` lookup already handled, limited to the last segment so a wildcard still matches the rest of the path. Full suite passes and I added the case to `common.case.test.ts`, where all four routers pass it. Happy to open the PR.

harshit-d3v:
Fixes #5370.

A handler runs twice when the last segment of the request path is literally `*`. Requesting `/a/*` against a route `/a/*` ran the matching middleware two times, every other path ran it once.

It is reachable on a default app, not only when TrieRouter is chosen explicitly, because SmartRouter falls back to TrieRouter whenever RegExpRouter throws `UnsupportedPathError`.

### Cause

`insert()` stores a pattern node twice, under its own key in `#children` and in `#patterns`. In `search()` a request part of `*` finds it through the `#children` lookup and pushes its handlers, then the `#patterns` loop reaches the same node and pushes them again. Any other part misses the `#children` lookup and only takes the pattern branch, which is why only a literal `*` duplicates.

### Fix

Skip the pattern branch when it resolves to the node the `#children` lookup already handled. Limited to the last segment, so a wildcard still matches the rest of the path for earlier segments. Without that limit `/a/*` stops matching `/a/*/b`.

Added the case to `common.case.test.ts`. All four routers pass it.

### DIFF
--- a/src/router/common.case.test.ts
+++ b/src/router/common.case.test.ts
@@ -370,6 +370,24 @@ export const runTest = ({
       })
     })
 
+    describe('Request path segment that is literally a star', () => {
+      beforeEach(() => {
+        router.add('GET', '/a/*', 'wildcard')
+      })
+
+      it('GET /a/* matches once, not twice', () => {
+        const res = match('GET', '/a/*')
+        expect(res.length).toBe(1)
+        expect(res[0].handler).toEqual('wildcard')
+      })
+
+      it('GET /a/*/b still matches', () => {
+        const res = match('GET', '/a/*/b')
+        expect(res.length).toBe(1)
+        expect(res[0].handler).toEqual('wildcard')
+      })
+    })
+
     describe('Suffix wildcard', () => {
       beforeEach(() => {
         router.add('GET', '/assets*', 'assets')
--- a/src/router/trie-router/node.ts
+++ b/src/router/trie-router/node.ts
@@ -119,6 +119,15 @@ export class Node<T> {
         }
 
         for (const child of node.#patterns) {
+          // A pattern node is also stored under its own key in `#children`, so a
+          // final request part that is literally the pattern (e.g. `*`) already
+          // had its handlers pushed above. Pushing them again would run the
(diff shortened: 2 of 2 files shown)


yusukebe:
This PR duplicates #5318. Closing. Thanks.

PR #5377, opened 2026-09-12
fix(pretty-json): don't break responses with unparseable JSON bodies

ndycode:

- [x] Add tests
- [x] Run tests
- [x] Add [TSDoc](https://tsdoc.org)/[JSDoc](https://jsdoc.app/about-getting-started) to document the code

## Problem

When prettification is active (`?pretty` or `force: true`), `prettyJSON` turns a successful response into a `500 Internal Server Error` if the response has a JSON `Content-Type` but a body that cannot be parsed as JSON. For example, an endpoint proxying an upstream API that returns a `204` with a JSON content-type:

```ts
const app = new Hono()
app.use('*', prettyJSON({ force: true }))
app.delete('/x', (c) => {
  c.header('Content-Type', 'application/json')
  return c.body(null, 204) // DELETE /x -> 500 (expected: 204)
})
```

Empty bodies, invalid JSON bodies passed through from an upstream, and responses whose body was already consumed by another middleware fail the same way. Since `?pretty` is controlled by the client, any caller can trigger the 500.

Additionally, when the original response carries a `Content-Length` (e.g. proxied from an upstream), the prettified response keeps the stale length after the body is replaced, because the `Context` `res` setter merges the old headers onto the new response. The prettified body is longer than the original, so the header no longer matches the body.

## Fix

- Parse `c.res.clone().json()` instead of `c.res.json()`, so the original response stays intact when the body is not parseable as JSON. On failure the response passes through unchanged. Reading from a clone follows the same pattern as the `etag` middleware.
- Delete `Content-Length` after assigning the prettified response, since the length of the body has changed.

## Tests

Six new cases in `src/middleware/pretty-json/index.test.ts`: `204` and `304` with a JSON content-type, an empty body, an invalid JSON body (status and headers preserved), a body already consumed by another middleware, and stale `Content-Length` removal. All six fail on `main` and pass with this change.

`bun run test`, `bun run lint`, `bun run build`, `bun run test:node`, `bun run test:bun`, `bun run test:workerd`, `bun run test:lambda`, `bun run test:lambda-edge`, and `bun run test:fastly` pass locally. In `bun run test:deno`, the two `deno-jsx` suites pass (12/12 each); the main Deno suite has one pre-existing failure in `Serve Static middleware` caused by this Windows checkout's CRLF line endings in a static fixture (reproduced on unmodified `main` under the same conditions; CI's Linux LF checkout is unaffected).

Follow-up maintenance run (2026-09-13, Windows 11, Node 24.18.0, bun 1.3.14): on head `eb67c3d2`, `bun run test` (including `tsc`), `bun run lint` (0 errors), `bun run build`, `bun run test:bun` (26/26), and the `node`, `workerd`, `lambda`, `lambda-edge`, and `fastly` vitest projects all pass. The catch is narrowed to the clone-and-parse operation only, the stale `Content-Length` fixture uses the body's actual byte length (19), pass-through preserves `Content-Length`, and the consumed-body case states its guarantee precisely: the response passes through with its body still consumed. `bun run test:deno` passes 12/12 in all three parts on an LF checkout, as does `jsr publish --dry-run`; on this Windows CRLF working tree the main Deno suite's `Serve Static` fixture assertion still fails, reproduced on unmodified `main` (EOL artifact, not code).

---

### DIFF
--- a/src/middleware/pretty-json/index.test.ts
+++ b/src/middleware/pretty-json/index.test.ts
@@ -119,4 +119,72 @@ describe('JSON pretty by Middleware', () => {
     const jsonSeqRes = await app.request('http://localhost/json-seq')
     expect(await jsonSeqRes.text()).toBe('{"message":"Hono!"}')
   })
+
+  it('Should not break 204 responses with a JSON content-type', async () => {
+    const app = new Hono()
+    app.use('*', prettyJSON({ force: true }))
+    app.delete('/x', (c) => {
+      c.header('Content-Type', 'application/json')
+      return c.body(null, 204)
+    })
+
+    const res = await app.request('http://localhost/x', { method: 'DELETE' })
+    expect(res.status).toBe(204)
+    expect(res.body).toBeNull()
+  })
+
+  it('Should not break empty responses with a JSON content-type', async () => {
+    const app = new Hono()
+    app.use('*', prettyJSON())
+    app.get('/x', () => new Response('', { headers: { 'Content-Type': 'application/json' } }))
+
+    const res = await app.request('http://localhost/x?pretty')
+    expect(res.status).toBe(200)
+    expect(await res.text()).toBe('')
+  })
+
+  it('Should pass through a response with an invalid JSON body unchanged', async () => {
+    const app = new Hono()
+    app.use('*', prettyJSON({ force: true }))
+    app.get(
+      '/x',
+      () =>
+        new Response('not json', { status: 200, headers: { 'Content-Type': 'application/json' } })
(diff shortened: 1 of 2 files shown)


ndycode:
The updated head narrows the catch to the JSON parse step, uses the correct byte length in the stale Content-Length fixture, and states the consumed-body guarantee precisely. All local checks pass (test, lint, build, bun-native, node/workerd/lambda/fastly projects, and the Deno suites on an LF checkout).

yusukebe (approved):

PR #5378, opened 2026-09-12
fix(client): don't override a user-specified Content-Type when sending json

ndycode:

- [x] Add tests
- [x] Run tests

## Problem

When `args.json` is set, the RPC client sets `Content-Type: application/json` after merging the caller's headers, so an explicitly specified content type is silently replaced:

```ts
client.posts.$post(
  { json: payload },
  { headers: { 'Content-Type': 'application/merge-patch+json' } }
)
// server receives Content-Type: application/json
```

This contradicts the client's own convention that caller-supplied headers win over client defaults (see the note on `init` in `ClientRequestOptions`), and makes JSON-subtype media types like `application/merge-patch+json` or `application/vnd.api+json` unusable with `json` bodies unless the whole `init.headers` is replaced.

## Fix

Only apply the derived `application/json` when the caller did not already specify a `Content-Type` header, comparing header names case-insensitively (a lowercase `content-type` previously lost to the derived value when both were set on the `Headers` object). This covers both per-request `headers` (including the function form) and route-typed `header` args.

## Tests

Six cases in `src/client/client.test.ts`, each verifying the resulting request `Content-Type` and JSON body through the test server: exact and lowercase header names, the function form of the `headers` option, a route-typed `header` argument, an explicitly empty `Content-Type` value, and `init.headers` taking highest priority. The five override cases fail on `main` with `expected 'application/json' to be 'application/merge-patch+json'`; the `init.headers` precedence case passes on both, documenting unchanged precedence. All pass with this change.

Follow-up maintenance run (2026-09-13, Windows 11, Node 24.18.0, bun 1.3.14): on head `e1879ff7`, `bun run test` (including `tsc`, which type-checks the typed header argument), `bun run lint` (0 errors), `bun run build`, `bun run test:bun` (26/26), and the `node`, `workerd`, `lambda`, `lambda-edge`, and `fastly` vitest projects all pass. `bun run test:deno` passes 12/12 in all three parts on an LF checkout, as does `jsr publish --dry-run`; on this Windows CRLF working tree the main Deno suite's `Serve Static` fixture assertion and the jsr dirty-check still fail, both reproduced on unmodified `main` (EOL artifacts, not code).

---

### DIFF
--- a/src/client/client.test.ts
+++ b/src/client/client.test.ts
@@ -154,6 +154,38 @@ describe('Basic - JSON', () => {
     expect(data.requestBody).toEqual(payload)
   })
 
+  it('Should not override a user-specified Content-Type when sending json', async () => {
+    const res = await client.posts.$post(
+      {
+        json: payload,
+        header: { 'x-message': 'foobar' },
+        cookie: { debug: 'true' },
+      },
+      { headers: { 'Content-Type': 'application/merge-patch+json' } }
+    )
+
+    expect(res.ok).toBe(true)
+    const data = await res.json()
+    expect(data.requestContentType).toBe('application/merge-patch+json')
+    expect(data.requestBody).toEqual(payload)
+  })
+
+  it('Should not override a user-specified Content-Type when sending json (lowercase header name)', async () => {
+    const res = await client.posts.$post(
+      {
+        json: payload,
+        header: { 'x-message': 'foobar' },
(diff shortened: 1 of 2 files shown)


ndycode:
The updated head adds Content-Type tests for function-form headers, route-typed header arguments, empty values, and init.headers precedence. All local checks pass.

PR #5379, opened 2026-09-12
fix(jwt): throw JwtTokenInvalid when the signature is not valid base64url

ndycode:

- [x] Add tests
- [x] Run tests

## Problem

`JWT.verify()` decoded the signature part with `decodeBase64Url()` outside any `try`/`catch`, so a token whose third part contains characters outside the base64url alphabet (e.g. `header.payload.@@@@`) rejected with a raw `InvalidCharacterError` `DOMException` instead of a Hono JWT error:

```ts
await JWT.verify('eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJtZXNzYWdlIjoiaGVsbG8gd29ybGQifQ.@@@@', secret, 'HS256')
// Uncaught DOMException: Failed to execute 'atob'...
```

The header and payload parts already throw `JwtTokenInvalid` for unparseable input (see `decode()`), so the signature part was inconsistent. Code that branches on `e instanceof JwtTokenInvalid` (or on Hono JWT error types generally) misses this failure mode. `verifyWithJwks()` is affected through its call to `verify()`. The `jwt` middleware itself maps any error to a 401, so it is unaffected.

## Fix

Decode the signature inside `verify()` in a `try`/`catch` and throw `JwtTokenInvalid(token)`, matching the precedent in `decode()`. `verifying()` remains outside the `try`, so crypto errors and the signature-mismatch path are untouched. An empty signature part still produces `JwtTokenSignatureMismatched` (unchanged).

## Tests

Two cases in `src/utils/jwt/jwt.test.ts`. A signature that is not valid base64url must reject with a `JwtTokenInvalid` carrying the documented message; this fails on `main` (the thrown error is a `DOMException`) and passes with this change. A control asserts an empty signature still reports `JwtTokenSignatureMismatched`; it passes on both, documenting the unchanged behavior.

Follow-up maintenance run (2026-09-13, Windows 11, Node 24.18.0, bun 1.3.14): on head `68c9509b`, `bun run test` (including `tsc`), `bun run lint` (0 errors), `bun run build`, `bun run test:bun` (26/26), and the `node`, `workerd`, `lambda`, `lambda-edge`, and `fastly` vitest projects all pass. `bun run test:deno` passes 12/12 in all three parts on an LF checkout, as does `jsr publish --dry-run`; on this Windows CRLF working tree the main Deno suite's `Serve Static` fixture assertion and the jsr dirty-check still fail, both reproduced on unmodified `main` (EOL artifacts, not code).

---

### DIFF
--- a/src/utils/jwt/jwt.test.ts
+++ b/src/utils/jwt/jwt.test.ts
@@ -92,6 +92,21 @@ describe('JWT', () => {
     expect(authorized).toBeUndefined()
   })
 
+  it('JwtTokenInvalid for a signature that is not valid base64url', async () => {
+    const tok = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJtZXNzYWdlIjoiaGVsbG8gd29ybGQifQ.@@@@'
+    const secret = 'a-secret'
+    let err: JwtTokenInvalid
+    let authorized
+    try {
+      authorized = await JWT.verify(tok, secret, AlgorithmTypes.HS256)
+    } catch (e) {
+      err = e as JwtTokenInvalid
+    }
+    // @ts-ignore
+    expect(err).toEqual(new JwtTokenInvalid(tok))
+    expect(authorized).toBeUndefined()
+  })
+
   it('JwtTokenNotBefore', async () => {
     const tok =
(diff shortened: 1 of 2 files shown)


ndycode:
The updated head replaces the ts-ignore pattern with type-safe assertions on the JwtTokenInvalid class and its message, plus a control for the unchanged empty-signature behavior. All local checks pass.

yusukebe (approved):

PR #5380, opened 2026-09-12
fix(jsx): replace Suspense and ErrorBoundary content across newlines

ndycode:

- [x] Add tests
- [x] Run tests

## Problem

The buffer-phase replacement regexes used by `Suspense` and `ErrorBoundary` matched their fallback/content region with `.*?`, which cannot match across newlines. When the region contained a line break, the replacement was a silent no-op: for `Suspense` the resolved content was dropped entirely (it only reaches the buffer through that replacement), and for `ErrorBoundary` the `<template>`/comment markers stayed in the output. Example through the `writeSSE` Stringify phase:

```tsx
stream.writeSSE({
  data: (
    <Suspense fallback={<div>{'Loading...\nPlease wait'}</div>}>
      <AsyncComponent />
    </Suspense>
  ),
})
```

On `main` the SSE message contains the fallback plus `<template id="H:0"></template>...<!--/$-->` markers, and the async component's output never appears.

## Fix

Use `[\s\S]*?` instead of `.*?` in both regexes (`src/jsx/streaming.ts` for `Suspense`, `src/jsx/components.ts` for `ErrorBoundary`) so the regions match regardless of line breaks. The markers anchor each match, and the lazy quantifier keeps stopping at the boundary's own end marker, so the replacement scope is unchanged for single-line regions. The DOM (script-based) replacement paths are unaffected; only the Stringify-phase buffer replacements changed.

## Tests

Four cases in `src/helper/streaming/sse.test.tsx`: a `Suspense` with a multi-line fallback, an `ErrorBoundary` wrapping a `Suspense` whose resolved content is multi-line, sibling `Suspense` boundaries with multi-line fallbacks (each replaces only its own region; the sibling's content is not consumed), and nested `Suspense` boundaries with multi-line fallback and content. All four fail on `main` with the template markers leaking into the SSE output, and pass with this change. The last two read the stream to completion.

Follow-up maintenance run (2026-09-13, Windows 11, Node 24.18.0, bun 1.3.14): on head `380293d4`, `bun run test` (including `tsc`), `bun run lint` (0 errors), `bun run build`, `bun run test:bun` (26/26), and the `node`, `workerd`, `lambda`, `lambda-edge`, and `fastly` vitest projects all pass. `bun run test:deno` passes 12/12 in all three parts on an LF checkout, as does `jsr publish --dry-run`; on this Windows CRLF working tree the main Deno suite's `Serve Static` fixture assertion still fails, reproduced on unmodified `main` (EOL artifact, not code).

---

### DIFF
--- a/src/helper/streaming/sse.test.tsx
+++ b/src/helper/streaming/sse.test.tsx
@@ -1,6 +1,6 @@
 /** @jsxImportSource ../../jsx */
 import { Context } from '../../context'
-import { ErrorBoundary } from '../../jsx'
+import { ErrorBoundary, Suspense } from '../../jsx'
 import { streamSSE } from '.'
 
 describe('SSE Streaming helper', () => {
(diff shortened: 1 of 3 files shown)


ndycode:
The updated head adds sibling and nested multiline Suspense boundary controls, reading the stream to completion. All local checks pass.

yusukebe (approved):
LGTM!

PR #5384, opened 2026-09-13
fix(request): keep the body readable when form data parsing fails

### LINKED ISSUE #5383: A failed c.req.formData() makes all other body methods throw

4.13.7 (`main`, `edd138ee`)

When `c.req.formData()` fails, the failure does not stay local to that call — every other body representation of the same request throws afterwards.

The shortest reproduction is a request whose body is not a form at all:

```ts
import { Hono } from 'hono'

const app = new Hono()

app.post('/', async (c) => {
  try {
    await c.req.formData() // JSON body, so this throws
  } catch {}

  const body = await c.req.json() // throws too
  return c.json({ body })
})

await app.request('/', {
  method: 'POST',
  body: '{"event":"push"}',
})
```

The shape that made me notice it is trying one parser and falling back to another — which is the entire point of the `try`/`catch`:

```ts
let body
try {
  body = await c.req.formData()
} catch {
  body = await c.req.json() // never works
}
```

#### Tracing it with the debugger

**1.** The first call, `c.req.formData()` on a `Content-Type: application/json` request. `bodyCache` is still empty, and line 243 stores `raw.formData()`'s promise into it unconditionally — including when that promise is about to reject.

src/request.ts:243 — the first call, bodyCache is still empty, key = "formData"

`src/request.ts:243` — `return (bodyCache[key] = raw[key]())`. Variables: `key = "formData"`, `bodyCache = {}`, `cachedBody = undefined`. Watch: `"formData requested | cached: []"`. Call stack: `#cachedBody:243 ← formData:335 ← <anonymous> debug-src.ts:19`.

**2.** The second call comes from the `catch` above: `c.req.json()` asks for `text`. The direct cache lookup misses, so it drops into the fallback loop — and the only entry there is the already-rejected `formData` promise, which line 226 hands straight back:

src/request.ts:226 — key = "text" but anyCachedKey = "formData", and the cached promise is rejected

`src/request.ts:226` — `return (bodyCache[anyCachedKey as keyof Body] as Promise<BodyInit>).then(...)`. Variables: `key = "text"`, `anyCachedKey = "formData"`, `bodyCache = {formData: Promise}`, `cachedBody = undefined`.

```
bodyCache.formData
  = Promise { result: TypeError [ERR_FORMDATA_PARSE_ERROR]:
              Can't decode form data from body because of incorrect MIME type/boundary,
              status: "rejected" }
```

Watch: `"text requested | cached: [formData]"`. Call stack: `#cachedBody:226 ← json:259 ← <anonymous> debug-src.ts:27` — the `json` frame is the `catch` fallback.

**3.** The fallback therefore rejects with the first parser's error, and the request ends as a 500:

debug console — FORM_DATA_REJECTED, then JSON_REJECTED, then STATUS 500

```
Debugger attached.
FORM_DATA_REJECTED: TypeError
JSON_REJECTED: TypeError | Can't decode form data from body because of incorrect MIME type/boundary
STATUS: 500
RESULT: {"ok":false,"error":"Can't decode form data from body because of incorrect MIME type/boundary"}
Debugger detached.
```

A failed `c.req.formData()` should fail only that call. The other representations should still be readable, exactly as they already are when the body is read with `c.req.parseBody()`:

```ts
const req = new HonoRequest(
    method: 'POST',
    body: '{"event":"push"}',
  })
)

await req.parseBody()
await req.text() // works today — the bytes were kept
```

The same request, going through `c.req.formData()` instead:

```ts
await req.formData() // throws, as expected
await req.text()     // throws as well — this is the bug
```

### What do you see instead?

```
TypeError: Can't decode form data from body because of incorrect MIME type/boundary
```

for `text()`, `json()`, `arrayBuffer()`, `bytes()` and `blob()` — all five, after a single failed `formData()`.

### Additional information

`c.req.parseBody()` does not have this problem because it reads the bytes first and parses them afterwards, so the raw body survives a failed parse (`src/utils/body.ts`, `parseFormData()`). `c.req.formData()` goes straight to `raw.formData()`, which consumes the stream and caches only the result — a rejected promise when the media type does not match.

On Node this is worse than it looks: undici marks the body as used even when it rejects before parsing, so `raw.text()` is no longer possible either (`Body is unusable: Body has already been read`). Bun happens not to mark it used, so the two runtimes disagree on how recoverable the request is. The cache is what both share.

Related: #4806 / #4807 fixed the same class of problem for `parseBody()`, where a cached entry made later `text()` / `json()` calls throw.

xia-chao:
Fixes #5383

- [x] Add tests
- [x] Run tests

### What was wrong

`c.req.formData()` keeps only the parsed result and throws the body away. If the body isn't a form, that result is a rejected promise — and it stays in the cache, so every other way of reading the same body gets handed the same rejection:

```ts
app.post('/', async (c) => {
  try {
    await c.req.formData() // JSON body, so this throws
  } catch {}

  const body = await c.req.json() // throws too
})
```

`text()`, `json()`, `arrayBuffer()`, `bytes()` and `blob()` all fail after that.

Node makes it worse than it looks: `raw.formData()` consumes the stream even when it rejects (undici marks the body as used before it checks the media type), so there is nothing left to read afterwards — `raw.text()` throws `Body is unusable`. The bytes have to be read *before* the parse, not after.

### The fix

`c.req.parseBody()` already gets this right: it reads the bytes first and parses them afterwards, so a failed parse leaves the body readable. This makes `formData()` follow the same order, reusing `bufferToFormData()` — which is still the runtime's own parser, just fed from the cached bytes instead of the live stream:

```ts
if (key === 'formData') {
  return (bodyCache.formData = this.arrayBuffer().then((buffer) =>
    bufferToFormData(buffer, raw.headers.get('content-type') ?? '').then((formData) => {
      // ...
      return formData
    })
  ))
}
```

### Keeping `bodyCache` the shape callers expect

`cloneRawRequest()` and the cache middleware both read `Object.keys(bodyCache)[0]` and expect `formData` there once a form has been parsed:

- `src/request.ts` — `cloneRawRequest()`
- `src/middleware/cache/index.ts`

So the original order is restored **only when the parse succeeds**. After a failure the raw bytes stay first, which is exactly what makes the other representations readable again. Both of those callers keep working with no changes.

### Tests

Added to `src/request.test.ts`, next to the existing `should not break body methods after parseBody() with non-form content-type`:

- `text()` / `json()` / `arrayBuffer()` / `bytes()` / `blob()`, each after a failed `formData()`
- a second `formData()` still reports the parse failure
- trying another representation after a failed parse (the `try`/`catch` fallback)

Six of the seven fail without the change and pass with it.

- `vitest --run src/request.test.ts src/middleware/cache/index.test.ts src/utils/body.test.ts` — 253 passed
- full suite — 5141 passed, 44 skipped. The 12 failures in `color` / `logger` / `dev` / `compress(node)` are colour- and TTY-dependent and fail the same way on an unmodified checkout
- `tsc -p tsconfig.spec.json --noEmit` — clean
- `bun run format:fix && bun run lint:fix` — no further changes

### Out of scope

After a *successful* `formData()`, `text()` still returns the re-serialized form instead of the original bytes. Same root cause, different trigger — fixing it would change the cache shape on the success path too, and with it the two callers above, so it belongs in its own change.


xia-chao:
@yusukebe 
Please review

yusukebe:
@xia-chao 

I want to consider this. Please wait.

xia-chao:
> @xia-chao
> 
> 我想考虑一下。请稍等。

OK! You continue

xia-chao:
@yusukebe 
How's it going?

yusukebe:
@xia-chao Don't rush me

PR #5387, opened 2026-09-13
fix(accepts): match media types and language tags case-insensitively

### LINKED ISSUE #5375: accepts() matches media types and language tags case-sensitively
`accepts()` compares the accept header entry and the supported value as raw strings, so a client that varies the case gets the default instead of a match.

```ts
const app = new Hono()
app.get('/', (c) =>
  c.text(accepts(c, { header: 'Accept', supports: ['text/html'], default: 'application/json' }))
)

// Accept: text/html  -> text/html
// Accept: TEXT/HTML  -> application/json
```

| Header | Value | supports | result |
| --- | --- | --- | --- |
| Accept | `text/html` | `['text/html']` | `text/html` |
| Accept | `TEXT/HTML` | `['text/html']` | default |
| Accept | `Text/Html` | `['text/html']` | default |
| Accept-Language | `en-US` | `['en-US']` | `en-US` |
| Accept-Language | `en-us` | `['en-US']` | default |
| Accept-Language | `EN` | `['en']` | default |

Both are case-insensitive. Media types by RFC 9110 §8.3.1, language tags by RFC 4647 §2.1. `en-us` is the one I would expect in the wild, since the uppercase region subtag is only a convention.

### Cause

```ts
// src/helper/accepts/accepts.ts
const matchType = (acceptType: string, supportedType: string): boolean => {
  if (acceptType === supportedType) {
    return true
  }
  ...
  return acceptMain === supportedMain
}
```

Nothing case-folds. The subtype wildcard path has the same problem, so `TEXT/*` does not match `text/html` either.

This is the same bug #5060 reported for `Content-Type`. #5064 fixed it for `parseBody()` and `validator()`, but `accepts()` was not part of that change.

PR is up. `defaultMatch()` returns the caller's own `supports` entry, so the casing an application registered is preserved.

nikolas-sapa:
Fixes #5375.

`accepts()` compared the accept header entry and the supported value as raw strings, so a client that varies the case got the default instead of a match:

| Header | supports | result before | result after |
| --- | --- | --- | --- |
| `Accept: TEXT/HTML` | `['text/html']` | default | `text/html` |
| `Accept: Text/Html` | `['text/html']` | default | `text/html` |
| `Accept: TEXT/*` | `['text/html']` | default | `text/html` |
| `Accept-Language: en-us` | `['en-US']` | default | `en-US` |
| `Accept-Language: EN` | `['en']` | default | `en` |

Media types match case-insensitively per RFC 9110 §8.3.1 and language tags per RFC 4647 §2.1, so `matchType()` now case-folds both sides — the same treatment `parseBody()`/`validator()` received for `Content-Type` in #5064. The subtype-wildcard path folds too, so `TEXT/*` matches `text/html`.

Tests: 5 new cases mirroring the issue's table; falsified by reverting `matchType()` (all 5 fail), green with the fix (21/21 in the file). `tsc -p tsconfig.spec.json`, eslint, and prettier all clean.


yusukebe:
The issue was closed.

PR #5390, opened 2026-09-13
fix(linear-router): do not match a :label against an empty segment

### LINKED ISSUE #5372: LinearRouter matches an empty path segment and returns an empty param
`LinearRouter` matches a path with an empty segment and hands the handler an empty param. The other three routers return 404.

```ts
import { Hono } from 'hono'
import { LinearRouter } from 'hono/router/linear-router'

const app = new Hono({ router: new LinearRouter() })
app.get('/u/:id/posts', (c) => c.json({ id: c.req.param('id') }))

await app.request('http://localhost/u/42/posts')  // 200 { id: "42" }
await app.request('http://localhost/u//posts')    // 200 { id: "" }   <- expected 404
```

| Router | `/u//posts` |
| --- | --- |
| RegExpRouter | 404 |
| TrieRouter | 404 |
| PatternRouter | 404 |
| LinearRouter | 200, `id` is `""` |

A handler that expects an id to be present gets an empty string instead.

### Cause

In `match()`, the plain label branch already rejects an empty segment at the end of the path but not one in the middle:

```ts
let endValuePos = path.indexOf('/', pos + 1)
if (endValuePos === -1) {
  if (pos + 1 === path.length) {
    continue ROUTES_LOOP      // trailing empty segment, rejected
  }
  endValuePos = path.length
}
value = path.slice(pos + 1, endValuePos)
```

For `/u//posts` the `indexOf` returns `pos + 1`, so `value` is `''` and the match carries on. That is why `/a/:p` correctly 404s on `/a/` but `/u/:id/posts` matches `/u//posts`.

Fix is to reject `endValuePos === pos + 1` the same way. Full suite passes and I added the case to `common.case.test.ts`, where all four routers pass it. Happy to open the PR.

nikolas-sapa:
Fixes #5372.

`LinearRouter.match()` extracted an empty value for a `:label` when the path contained an empty segment mid-route — `/u//posts` against `/u/:id/posts` returned 200 with `params.id === ''`, while RegExpRouter, TrieRouter, and PatternRouter all return no match. A handler expecting an id to be present received an empty string instead.

The plain-label branch already rejects a *trailing* empty segment (`pos + 1 === path.length`); this adds the same rejection for `endValuePos === pos + 1` (the `indexOf('/')` landed exactly on the segment start). The `:label{pattern}` branch is unaffected — it already rejects zero-length matches.

Verification:

- new common case `Named Param does not match an empty segment` in `common.case.test.ts` runs against all four routers
- falsified by reverting the fix: LinearRouter alone fails the case (returns one handler with `id: ''`), the other three pass
- with the fix: all 7 router test files green (685 passed)


nikolas-sapa:
Follow-up: verified the cross-router consistency claim empirically instead of leaving it asserted, and extended the shared suite with the root-level shape.

**Matrix** (fresh router instance per pattern, all built-in routers):

| path | RegExpRouter | TrieRouter | PatternRouter | LinearRouter (this PR) |
|---|---|---|---|---|
| `/u/john/posts` vs `/u/:id/posts` | 1 match | 1 match | 1 match | 1 match |
| `/u//posts` vs `/u/:id/posts` | **0** | **0** | **0** | **0** |
| `/abc` vs `/:id` | 1 match | 1 match | 1 match | 1 match |
| `//` vs `/:id` | **0** | **0** | **0** | **0** |

So rejecting empty segments aligns `LinearRouter` with the other three routers on both the mid-route and root-level shapes, and the shared `common.case.test.ts` suite now pins both cases for every router (new test: "Named Param does not match an empty root segment"). All four router test files pass (403 passed, 24 skipped for the documented `UnsupportedPathError` differences).

yusukebe:
The issue was closed.

PR #5393, opened 2026-09-13
feat(adapter): add Ant adapter

theMackabu:
Adds hono/ant for the Ant runtime (https://github.com/theMackabu/ant), which exposes fetch(request, server), so the adapter is thin glue like hono/deno.

- upgradeWebSocket via server.upgradeWebSocket(request, { protocol })
- getConnInfo via server.requestIP
- serveStatic and toSSG on node:fs/promises
- getAntServer to read the server from c.env or c.env.server
- "ant" runtime key matched from the Ant/ user agent

- [x] Add tests
- [x] Run tests


yusukebe:
@theMackabu 

Thank you for the PR! Ant is a super interesting project for me.

For this adapter, I am considering how to treat it. We will release the next major version, v5, soon (maybe October?). In this version, we separate adapters from the core package `hono`.

For example, [`hono/bun`](https://github.com/honojs/hono/tree/main/src/adapter/bun) will be released as `@hono/bun`. So, I think it's better to add this Ant adapter as `@hono/ant` when v5 is released.

Anyway, whether it will be `@hono/ant` or `hono/ant`, this PR's code will be used. Please wait a while to decide.

theMackabu:
@hono/ant makes more sense for v5 anyways, the smaller the package the better

PR #5394, opened 2026-09-14
fix(reg-exp-router): apply middleware registered with a mid-path wildcard

nikolas-sapa:

- [x] Add tests
- [x] Run tests

---

Middleware registered on a path with a wildcard in the middle never runs on `RegExpRouter`, and therefore never runs on the default `SmartRouter` either.

```ts
const app = new Hono()

app.use('/*/admin/*', requireAdmin)
app.get('/:tenant/admin/users', (c) => c.text('secret'))

await app.request('/acme/admin/users')
// requireAdmin never runs; responds 200 "secret"
```

`TrieRouter`, `LinearRouter` and `PatternRouter` all run the middleware here. Only `RegExpRouter` skips it, and since `RegExpRouter` is what `SmartRouter` picks first, the default configuration is affected.

The failure is silent. `add()` does not throw, so `SmartRouter` has no reason to fall back to `TrieRouter` — it keeps `RegExpRouter` and quietly returns a result with the middleware missing. When the middleware is an auth guard, as above, the request reaches the handler unauthenticated.

## Cause

`buildWildcardRegExp` compiles a middleware path into a pattern that is then tested against each **registered route path** to decide which routes the middleware covers.

Its alternation only recognises a `*` that ends the path (`\/?\*$`). A `*` anywhere else falls through to the meta-character branch and is escaped into a literal asterisk:

```
'/*/admin/*'  →  /^\/\*\/admin(?:|\/.*)$/
```

That pattern only matches a path whose first part is the two characters `/*`, which no ordinary route is. So at

https://github.com/honojs/hono/blob/main/src/router/reg-exp-router/router.ts#L106

`re.test(p)` is false for every registered route and the handler is never pushed onto any of them.

`node.ts` already handles the same path correctly when it inserts it into the trie — a non-trailing `*` becomes `LABEL_REG_EXP_STR` there (node.ts#L69). The two disagree, and this change brings `buildWildcardRegExp` in line with the trie.

## Change

One alternative, `(\/\*)(?=\/)`, matching a `/*` that is followed by `/`, mapped to `/[^/]+` — one part, not the rest of the path. Ordering matters: it sits after `\/?\*$` so a trailing `*` still takes the tail-wildcard branch.

Behaviour of every path form that already worked is unchanged:

| middleware | before | after |
| --- | --- | --- |
| `/*` | `(?:\|/.*)` | unchanged |
| `/a/*` | `/a(?:\|/.*)` | unchanged |
| `/:x/*` | `/:[^/]+(?:\|/.*)` | unchanged |
| `/*/*` | `\/\*(?:\|/.*)` ❌ | `/[^/]+(?:\|/.*)` |
| `/*/admin/*` | `\/\*\/admin(?:\|/.*)` ❌ | `/[^/]+\/admin(?:\|/.*)` |

The new pattern discriminates correctly rather than over-matching: `/*/admin/*` covers `/:tenant/admin/users` and `/acme/admin/users`, and does not cover `/x/notadmin/y` or `/a/b`.

## Tests

Added to `common.case.test.ts` so all routers are held to it, next to the existing "Trailing wildcard after a middle wildcard" block. The new case adds the dimension that block was missing — a route with a **label** in the wildcard's position — plus a negative case so the fix cannot pass by over-matching.

Falsified: reverting the source change fails `RegExpRouter`, `PreparedRegExpRouter` and `SmartRouter` with `expected 1 to be 2` (the middleware missing), while `TrieRouter`, `LinearRouter` and `PatternRouter` pass either way — which is the asymmetry being fixed.

With the fix, all 7 router test files pass (691 tests). Full suite: 5144 passed. The two failures in `middleware/compress` streaming are pre-existing and reproduce on a clean `main` with this change stashed.

## Relationship to #5243

Not a duplicate, and they do not conflict — different files, different defect.

#5243 fixes *prefixed* wildcards (`/x*/y`, `/a/b*/c`) in the trie-insertion path in `node.ts`; this fixes middleware coverage matching in `buildWildcardRegExp` in `router.ts`, which #5243 does not touch. I checked out #5243 and re-ran this reproduction against it: `/*/admin/*` still silently skips the middleware. It also changes `use('/*/*') + get('/a/b')` from throwing `UnsupportedPathError` (safe — `SmartRouter` falls back and the middleware runs) to silently skipping it. Whichever lands first, the other still applies.


yusukebe:
The bug is real. But we'll go with https://github.com/honojs/hono/pull/5415.

PR #5399, opened 2026-09-15
fix(request): isolate failed formData cache so other body methods can succeed

### LINKED ISSUE #5383: A failed c.req.formData() makes all other body methods throw

tahiraltundag:
Fixes #5383

**Problem:** `c.req.formData()` stored its rejected promise in `bodyCache` unconditionally (`bodyCache[key] = raw[key]()`). The fallback that reuses any cached body (`for (...in bodyCache) return bodyCache[anyCachedKey].then(... Response[ key]())`) then handed the same rejection to every other body method. A `try { await formData } catch { await json }` fallback was impossible and the request ended 500.

Raw `Request` body is also marked `bodyUsed` after the first read, so even deleting the rejected entry is not enough — `raw.json()` would still fail with `Body is unusable`.

**Fix:** 
- Do not keep a rejected entry: `delete bodyCache[key]` on rejection and make the fallback loop skip a rejected entry.
- Preserve raw bytes via a pre-read `raw.clone().arrayBuffer()` so a later representation can be rebuilt from `arrayBuffer` even after `bodyUsed` becomes true (clone is taken before the first read).

Verified with reproduction (JSON body, `Content-Type: application/json`):
```ts
await c.req.formData() // throws as expected
await c.req.json() // now returns {event:'push'} instead of throwing
// app returns 200 instead of 500
```
Existing `src/request.test.ts` baseline unchanged (46 pass, 4 pre-existing fails due to bun fetch differences).


yusukebe:
I will handle the issue. Closing.

PR #5405, opened 2026-09-17
Fix multiline CSS handling in style replacement

uniho:
## Summary

Fix an issue where `Style` fails to append generated CSS when the existing `<style>` element contains a newline.

## Reproduction

The following works as expected:

```tsx
<Style>{css`/* comment（コメントが１行ならOK！） */`}</Style>
```

However, adding a newline inside the comment causes the generated CSS not to be appended:

```tsx
<Style>{css`
/* comment
コメントが複数行に渡る場合、おそらく改行が残るためNG?
another line */
`}</Style>
```

This also occurs with multiline comments in otherwise valid CSS.

## Cause

The style replacement regex currently uses `.*?`:

```ts
const replaceStyleRe = new RegExp(
  `(<style id="${id}"(?: nonce="[^"]*")?>.*?)(</style>)`
)
```

In JavaScript regular expressions, `.` does not match line terminators by default. Therefore, the expression cannot match the complete `<style>` element when its contents contain a newline.

## Fix

Use `[\s\S]*?` instead of `.*?` so that multiline style contents are matched:

```diff
- const replaceStyleRe = new RegExp(`(<style id="${id}"(?: nonce="[^"]*")?>.*?)(</style>)`)
+ const replaceStyleRe = new RegExp(`(<style id="${id}"(?: nonce="[^"]*")?>[\\s\\S]*?)(</style>)`)
```

This allows the existing style contents to be matched regardless of line breaks.


uniho:
By tracing this back, it also looks like there may be an issue with the comment removal in `minifyCssRe` in `common.ts`, where comments containing newlines are not always handled correctly.

This seems to be a separate issue from this PR, but it may be worth looking into as well.

usualoma:
Hi @uniho, thanks for opening this pull request.

I think this can be handled separately from the comment removal issue.

Could you use the `s` flag here?

```ts
const replaceStyleRe = new RegExp(
  `(<style id="${id}"(?: nonce="[^"]*")?>.*?)(</style>)`,
  's'
)
```

Please also add a regression test confirming that generated CSS is appended when `<Style>` contains a multiline comment. With that and CI passing, this should be ready to merge.

PR #5414, opened 2026-09-18
fix(utils): release the pool slot when a task rejects

### LINKED ISSUE #5413: utils: `createPool()` never releases the slot of a rejected task, and the parked caller's promise never settles

4.13.8 (`main`, `098e1191`)

### What runtime/platform/OS are you using?

`bun` 1.4.3 on macOS. The bug is in plain TypeScript, not runtime-specific.

A pool with one slot. The first task throws. The second task never runs:

```ts
import { createPool } from 'hono/utils/concurrent'

const pool = createPool({ concurrency: 1 })

pool.run(async () => {
  throw new Error('task-1-boom')
}).catch((e) => console.log('task-1 rejected:', e.message))

// Never throws, and never runs: the slot from task 1 was never given back.
const probe = pool.run(() => 'PROBE-RAN')
probe.then((v) => console.log('task-2 resolved:', v))
```

```
[repro] task-1 -> rejected: task-1-boom
[repro] ---- waited 1500ms ----
[repro] task-2 (probe) : pending
[repro] RESULT         : HANG — the pool never runs another task
```

When a task finishes — success or fail — its slot should be freed so the next task can run. `run()` should also reject if the task threw.

### What do you see instead?

Two things:

1. If a task throws, the slot is never given back. Later tasks wait forever.
2. A queued task's promise never settles — it neither resolves nor rejects.

Debugger trace on `src/utils/concurrent.ts`:

**Step 1** — first task enters `run()`. The pool is empty:

The pool is empty when the first task enters run()

**Step 2** — the marker is added, so the slot is taken. Next line throws, and `pool.delete(marker)` never runs:

The marker has been added, the slot is occupied

**Step 3** — second task enters `run()`. `pool.size` is still `1`, so it gets queued instead of running:

The second task sees a full pool because the slot was never released

**Step 4** — 1.5 s later, the second task's promise is still `pending`. It never runs, and it never rejects:

The parked promise never settles

xia-chao:
Fixes #5413

- [x] Add tests
- [x] Run tests

## What was wrong

If a task in `createPool()` throws, two things go wrong:

1. The pool never frees that slot. After enough failures, later tasks wait forever.
2. A task that was waiting in line never finishes — its promise neither resolves nor rejects.

That's why `toSSG()` can hang when one route fails.

## Fix

Always free the slot when the task is done, even if it failed. If a waiting task fails, reject its promise instead of leaving it hanging.

## Tests

Added cases for a failing task (with and without `interval`) and for a queued task that fails.


usualoma:
Hi @xia-chao, thanks for the PR. This issue has been discussed several times, and #5021 and #5312 were closed for somewhat different reasons.

Of the proposed fixes, #5021 looks the most complete. However, we don’t consider this an urgent issue, so we’d prefer to revisit it later if the maintainers decide a change is needed.

xia-chao:
By default, onError turns a thrown Error into a 500, so app.fetch doesn't reject. Thus a normal throwing route won't hang in toSSG. (Verified locally.)

Only two paths cause rejection:

Custom onError that rethrows.

Route/middleware throws a non-Error value (throw 'oops' / Promise.reject({...})). This happens with zero config, because hono-base.ts:400 (#handleError) and compose.ts:54 only handle err instanceof Error; anything else is rethrown.

@usualoma 
Yes, this is not an urgent issue. Thank you for your response, I just need a response

PR #5417, opened 2026-09-19
fix(serve-static): handle only GET/HEAD and stop wildcard routes from turning 404s into 405s

### LINKED ISSUE #5223: When serveStatic and methodNotAllowed are used together, methods other than GET cannot be handled correctly

nikolas-sapa:
## Description

Fixes #5223, implementing the direction agreed in [usualoma's comment](https://github.com/honojs/hono/issues/5223#issuecomment-5289772000).

Two related failures when `serveStatic` and `methodNotAllowed` are combined:

**1. `app.use('/*', serveStatic())` served files for any method.** The core middleware had no method check, so `POST <existing file>` answered with the file's content (200) — `methodNotAllowed` never saw a 404 to convert into a 405.

**Fix:** the handler now returns `next()` for any method other than GET/HEAD **before** the path is resolved or `getContent` is called — as specified in the discussion, deliberately without a file-existence check to distinguish 404 vs 405, and without touching the filesystem for other methods. All runtime adapters (Deno/Bun/Cloudflare) delegate to this core factory, so one change covers them.

**2. `app.get('/*', serveStatic())` made every unknown path answer 405 instead of 404.** `methodNotAllowed` inferred allowed methods from all registered routes, including trailing-wildcard routes — but a wildcard is no evidence a resource exists at any specific path, so `POST <non-existent file>` became 405.

**Fix:** `methodNotAllowed` now skips routes whose path ends in `*` when building the method-inference router, per the same comment. Specific routes (`/api` etc.) still yield 405 with a correct `Allow` header.

## Tests

- `serve-static`: POST to an existing file falls through with `getContent` never called (404); HEAD still serves (200).
- `method-not-allowed`: wildcard routes don't contribute to inference — unknown path stays 404 without `Allow`, specific path still 405 with `Allow: GET, HEAD`, wildcard route itself still serves.

All 43 tests in both suites pass; full vitest run shows only the pre-existing flaky `stream > Should call onAbort` timeout, which fails identically on `main` without these changes.


HeyITGuyFixIt:
It looks like cloudflare pages needs to be handled differently. It doesn't delegate to the core factory as the environment is substantially different.

PR #5419, opened 2026-09-19
fix(accepts): treat `*/*` and `*` wildcards as matching any supported type

### LINKED ISSUE #5418: accepts(): `Accept: */*` never matches a supported type and always returns the default

4.13.8 (`main`)

### What steps reproduce the bug?

`defaultMatch` in `src/helper/accepts/accepts.ts` treats `*/*` (and `*`) as a non-matching entry — `matchType` returns `false` for it — so with a wildcard-only `Accept` header every supported entry is skipped and the configured `default` is returned:

```ts
import { accepts } from 'hono/accepts'

const result = accepts(c, {
  header: 'Accept',
  supports: ['text/html', 'application/json'],
  default: 'text/plain',
})
// with `Accept: */*` → 'text/plain' (the default), not 'text/html'
```

Verified against `main`: `Accept: */*` returns `text/plain`; `Accept: text/*` correctly returns `text/html`.

Per RFC 9110 §12.5.1, `*/*` means every media type is acceptable, and the server should respond with its most-preferred supported representation — here `text/html`, the first entry in `supports`. Falling back to `default` is only correct when nothing in `supports` is acceptable, which is not the case for `*/*` (and `type/*` already demonstrates the intended wildcard matching).

### Suggested direction

Treat `*/*` (and `*`) as matching any supported entry with the lowest specificity (it already sorts last among equal-q entries via `getSpecificity`), so a wildcard-only header selects the server's preferred supported type instead of bypassing negotiation. The same question applies to `Accept-Language: *` for the language path.

Found while working on the case-insensitivity fix in #5387; filing separately to keep that PR scoped to case folding.

nikolas-sapa:
## Description

Fixes #5418 (issue I filed while working on #5387, kept separate to scope that PR to case folding).

`matchType` in `src/helper/accepts/accepts.ts` explicitly returned `false` for `*/*` and `*`, so a wildcard-only `Accept` header bypassed negotiation entirely and always fell through to `default` — even though RFC 9110 §12.5.1 says the wildcard makes every type acceptable and the server should answer with its most-preferred supported representation. `type/*` already matched correctly, which highlights the inconsistency. `getSpecificity()` also already ranks wildcards lowest (1) — dead logic while `matchType` rejected them unconditionally.

## Changes

- `src/helper/accepts/accepts.ts`: `matchType` now returns `true` for `*/*` and `*` against any supported entry. Because wildcards sort last at equal q (existing `getSpecificity`), a header like `*/*, application/json` still prefers the concrete type, and q=0 wildcards remain non-acceptable via the existing q>0 filter.
- `src/helper/accepts/accepts.test.ts`:
  - Updated three tests that encoded the old fall-through behavior (they relied on `*/*`/`*` never matching to reach `default`); the no-match test now uses a wildcard-free header so it still exercises the default path.
  - Added: wildcard-only `*/*` selects the first supported type (server preference), `*` matches `Accept-Encoding`/`Accept-Language`, and `*/*;q=0` still returns the default.

## Behavior change note

This changes observable behavior for wildcard-only headers: e.g. the compress middleware with `Accept-Encoding: *` will now negotiate `gzip` instead of returning the `identity` default — which is the RFC-correct reading. Flagging it explicitly since existing apps relying on the old fall-through would see different negotiation results.

All 51 tests in the accepts suites pass; full `vitest --run` suite green.


yusukebe:
Should be closed since the issue was closed.

PR #5421, opened 2026-09-20
fix(types): expose context returned by next

### LINKED ISSUE #2421: middleware next() return type is inaccurate

4.1.4

Cloudflare workers

Use a bare bones middleware, and log in the middleware

```typescript
import { Hono } from 'hono';

const app = new Hono()

app.use(async (ctx, next) => {
  // typed as void, but actual value is a Context object.
  const nextResult = await next();
  console.log('nextResult', nextResult);
});
```

I expect the result of `await next()` to be `undefined`, since the return type is `void`.

### What do you see instead?

I see the actual value as a full Hono `Context` object.

### Additional information

- Also tested on Hono `4.0.1`.
- 💭 The returned Context object enables a lot of additional use cases. ie. checking `ctx.error` for if a wrapped middleware `threw`. Otherwise, middlewares are unable to detect if any wrapped middleware has `throw`n, since `await next()` does not appear to ever `throw`.

icecold009:
## Summary

- type `Next` with its actual `Context` return value
- preserve generic return typing for compatibility
- update internal handlers and add regression coverage for `await next()`

Fixes #2421

## Verification

- Focused Vitest suite: 153 tests passed across `src/types.test.ts` and `src/compose.test.ts`
- Prettier check passed
- `git diff --check` passed
- Full runtime suite reached 5,163 passing tests; remaining failures were environment/baseline issues in the Windows checkout, including Workerd filesystem access and runtime-specific ANSI/Undici behavior
- Repository-wide TypeScript validation is currently affected by dependency resolution from a pnpm install against the repository’s Bun lockfile


yusukebe:
This should be discussed. Until then, we can't handle this PR.

PR #5426, opened 2026-09-21
fix(lambda-edge): sync content type detection with aws-lambda

sary8:
`lambda-edge` still uses the old content-type check, so binary types such as `application/vnd.openxmlformats-officedocument.wordprocessingml.document` are treated as text because `openxmlformats` contains `xml`. `createResult()` then reads the body with `res.text()`.

`aws-lambda` has the updated check from #4469 and #5424. This applies the same check and test cases to `lambda-edge`. Six of those test cases fail with the current `lambda-edge` implementation.

Related: #4900, #5412.

- [x] Add tests
- [x] Run tests


yusukebe (approved):
LGTM!

yusukebe:
@sary8 Looks good. Thanks!

